PR #2, opened 2015-06-12
check for window object

felixzapata:
It checks the window object to avoid an undefined error when you use from a Node application.

### DIFF
--- a/lib/utils/respondable.js
+++ b/lib/utils/respondable.js
@@ -88,6 +88,8 @@
 		};
 	}
 
+if (typeof window !== 'undefined') {
+
 	window.addEventListener('message', function (e) {
 
 		if (typeof e.data !== 'string') {
@@ -111,7 +113,8 @@
 
 		publish(e, data);
 	}, false);
-
+	
+}
 	exports.respondable = respondable;
 
 }(utils));


dylanb:
Felix, can you explain the scenario under which you would use the library without a window object?

felixzapata:
I am trying to make a gulp plugin to use the axel core in a project instead of with the tests. This way you can check the rules meanwhile you are developing some templates. 

If I use your library inside of my plugin (work in progress) when I launch my tests i have errors because the window object is not defined. 

My gulp plugin should read the some files from a source and pass to your library.

Regards.

dylanb:
The way the rules engine works is it evaluates the HTML code using the DOM APIs. None of the rules will work if there is no DOM. If there is a DOM, it must have CSS enabled for certain of the rules to work, I think what you are trying to do is not within the scope of what this library was designed to do.

felixzapata:
I try to load the source file and parse it as a DOM to pass to your function. This way I think is more easy to use during the development instead to configure the test.

dylanb:
@felixzapata Instead of making this change, users who are doing what you are doing should add the following to their code base prior to calling the axe-core library

```
globals.window = {};
globals.window.addEventListener = function () {};
```

felixzapata:
Well, thanks.. I will try to make some attempt.

PR #25, opened 2015-06-17
Add tests for jQuery-like objects as context

dsturley:


dsturley:
Good old PhantomJS issue:

`typeof NodeList#length === 'function'`

PR #32, opened 2015-06-18
Add sauce labs

dsturley:
Might work... let's see.


dsturley:
Sauce tests cannot be run on forks.  "Security vulnerability" as someone could theoretically submit a PR that reads those environment variables and send them somewhere.

Can't tell if this will work until merged, so doing the naughty thing and merging myself.

PR #47, opened 2015-07-02
Remove Sauce Labs & combine integration tests

dsturley:


dsturley:
Sauce taking 30+ minutes to run all tests and intermittently failing tests for due to "Bad Gateway".  May want to leverage Jenkins to run these tests in Selenium as Sauce doesn't seem stable enough.

PR #58, opened 2015-07-08
Developer Guide for Rule Authoring

dsturley:
Can be viewed here:
https://github.com/dsturley/axe-core/blob/dev-guide/doc/developer-guide.md


dylanb:
@skotov I think you need to sign the CLA

PR #64, opened 2015-07-13
Adding support for ARIA 1.1 role='switch'

yzen:


dsturley:
Any change I could convince you to add the other roles in ARIA 1.1?

PR #66, opened 2015-07-15
Adding initial support for ARIA 1.1 roles: none, cell, table, searchb…

yzen:
…ox, text.


lib/commons/aria/index.js
```diff
@@ -194,24 +212,24 @@ lookupTables.role = {
...
+		attributes: {
+			allowed: ['aria-colindex', 'aria-colspan', 'aria-rowindex', 'aria-rowspan']
 		},
 		owned: null,
 		nameFrom: ['author', 'contents'],
-		context: null,
-		implicit: ['input[type="checkbox"]']
+		context: ['row']
 	},
-	'switch': {
```
yzen: I just moved switch down to preserve the alphabetical order in this file. Hope that's OK.

yzen:
I was also wondering, if this PR is good, what your process around bumping the axe-core versions (in terms of using it as an npm module)? Thanks

dsturley:
There is one more issue that needs to be in 1.1.0; #67.  When it will be available is a question for @dylanb 

Thanks again for your contribution.

yzen:
Ah good to know, thanks, I'll keep an eye on it.

dylanb:
@yzen full disclosure - we are still working on how/when and how frequently we release the modules that are dependent on this repo...

PR #99, opened 2015-08-02
create a build task for generating the valid major language list, add…

### LINKED ISSUE #23: Extend the default list of allowable languages to include all major languages
The list currently only supports Spanish, English and Japanese

https://github.com/dequelabs/axe-core/blob/master/lib/checks/language/valid-lang.json#L3

dylanb:
… this list to the valid-lang check. Fixes #23


lib/checks/language/valid-lang.tpl
```diff
@@ -0,0 +1,12 @@
+{
```
dsturley: This file isn't really necessary, you can rather easily read the JSON file, change options and then save it again.  If you decide to keep this file, then the valid-lang.json file should be removed from source control.

dylanb: The reason I left it there is so we could generate the list of languages again if we wanted to...but I also did not want the list of languages to have to be created with each build in case IANA moves that page. That is why I did not make the target part of the build target. Should I make it build every time?

dsturley: Its probably not a good idea to require this task to be run every time; it would be better to run once every other release or so.  However, if this is the case we should get rid of the `.tpl` file and just:

1.  Read JSON file - `var check = grunt.file.readJSON(path)` [grunt.file.readJSON](http://gruntjs.com/api/grunt.file#grunt.file.readjson)
2.  Assign options -  `check.options = langs`
3.  Write JSON file - `grunt.file.write(path, JSON.stringify(check, null, '  '))` [grunt.file.write](http://gruntjs.com/api/grunt.file#grunt.file.write)

Grunt handles all the pathing issues for you, so its preferable to use grunt file methods where possible.

If we keep the .tpl file and JSON file in source control, my fear is they will quickly become out of sync.  Part of the reason these files are still JSON is this kind of portability.

dylanb: That is not a bad solution.... :-)

dylanb:
Should be good to go now

lib/checks/language/valid-lang.tpl
```diff
@@ -0,0 +1,12 @@
...
+  "id": "valid-lang",
+  "options": <%= options %>,
+  "evaluate": "valid-lang.js",
+  "metadata": {
+    "impact": "serious",
+    "messages": {
+      "pass": "Value of lang attribute is included in the list of allowable languages",
+      "fail": "Value of lang attribute not included in the list of allowable languages"
```
dsturley: We should clarify this #103

Issue #100, opened 2015-08-03
question: Results API -> Node


jasonkarns:
The structure of the Results object has an array of Nodes `{html, target, ...}`. And each of the any/all/none Checks objects have an array of RelatedNodes `{html, target}`.  However, the higher level Nodes' `target` property is an array of selectors, whereas a RelatedNode's `target` seems to just be a single selector.

Is this intentional? As a consumer, it would be easier to have a data contract wherein the api of a `Node` type was the same. (Even though the higher level Node list has other properties in addition to `html` and `target`, I'm attempting to model them as where Node has `html` String and `target` Array, with `CheckedNode` as subclass of `Node` with the additional properties (`impact`, `any`, `all`, `none`)

dsturley:
`target` should alway be an `Array` of CSS selectors representing the path to the element, with all but the last item representing the path(s) to frames and the last item representing the selector for the actual element.  They are sourced from the same Constructor; DqElement.

Here is an example pulled from this page of a color contrast violation:
```json
{
  "target": [
    ".is-copy-enabled > .logged_in.env-production.macintosh.vis-public > .wrapper > .header.header-logged-in.true > .container.clearfix > .site-search.repo-scope.js-site-search > .js-site-search-form > .js-chromeless-input-container.form-control > .scope-badge"
  ],
  "html": "<div class=\"scope-badge\">This repository</div>",
  "impact": "critical",
  "any": [
    {
      "id": "color-contrast",
      "impact": "critical",
      "message": "Element has insufficient color contrast of 3.91 (foreground color: #767676, background color: #eeeeee, font size: 9.0pt, font weight: normal)",
      "data": {
        "fgColor": "#767676",
        "bgColor": "#eeeeee",
        "contrastRatio": "3.91",
        "fontSize": "9.0pt",
        "fontWeight": "normal"
      },
      "relatedNodes": [
        {
          "target": [
            ".is-copy-enabled > .logged_in.env-production.macintosh.vis-public > .wrapper > .header.header-logged-in.true > .container.clearfix > .site-search.repo-scope.js-site-search > .js-site-search-form > .js-chromeless-input-container.form-control > .scope-badge"
          ],
          "html": "<div class=\"scope-badge\">This repository</div>"
        }
      ]
    }
  ],
  "all": [],
  "none": []
}
```

Please let me know if you've found a reproducible case where `target` is not an Array, as this would be a bug.

jasonkarns:
[...]

PR #101, opened 2015-08-04
clarify that relatedNodes.target is an array

jasonkarns:


jasonkarns:
This was part of my confusion with #100

Issue #109, opened 2015-08-13
button-name misclassified as 1.1.1 Non-Text Content


dsturley:
4.1.2 Name, Role, Value

PR #110, opened 2015-08-14
Update messages and help URLs

dsturley:
Needs very thorough review.


dsturley:
Resolves #111 and #109

dsturley:
Also Resolves #109

PR #112, opened 2015-08-14
Exclude elements obscured by other elements from analysis (Color Contrast)

### LINKED ISSUE #77: Color contrast false positive on sticky headers
See https://gist.github.com/dsturley/4ce0bcaf8a83e931121e

We could probably ignore any element that is overlaid on top of the element tested.

dsturley:
Resolves #77


dylanb:
I wonder whether we should expose these as a warning (I know this is introducing a new concept) but there are at least two situations we can detect that fit into the category - elements on an image or gradient background and elements with something obscuring them.

PR #131, opened 2015-09-08
Lists with overridden roles

### LINKED ISSUE #129: False Positive: definition-list when role overriden
```html
<div role="list" id="target">
	<dl role="listitem">
	  <dd role="presentation">
			<a href="#somethin">Stuff</a>
		</dd>
	</dl>
</div>
```

Lists also have this problem:
```html
<ol role="presentation">
<span>hi</span>
</ol>
```

```html
<li role="presentation">
<span>hi</span>
</li>
```

dsturley:
Turn off list rules when role is overridden and allow an ARIA role=list to contain list items

Fixes #129


dylanb:
Looks to me like this will not select items with role="list"?

Issue #95, opened 2015-07-31
Disable color-contrast check when aria-disabled=true


dsturley:
Needs confirmation that aria-disabled=true satisifies the WCAG 1.4.3 requirement

dylanb:
Dave, as long as it is not used on an element supported by the "disabled" attribute, and as long as that element has a widget role, this attribute is sufficient for the WCAG 1.4.3 requirement

dsturley:
@dylanb What about readonly?

dylanb:
When working on this issue, look at the above pull request and close it if the issues raised in the pull request have not been addressed. If they have been addressed, please use the PR as a basis for completing this issue.

WilcoFiers:
Issue has been resolved for the next release.

PR #133, opened 2015-09-30
ignore color contrast requirements for disabled elements

BenMakesGames:
as per this page: http://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html

"Text or images of text that are part of an inactive user interface component, that are pure decoration, that are not visible to anyone, or that are part of a picture that contains significant other visual content, have no contrast requirement."

and, semi-relatedly, this: http://www.w3.org/WAI/PF/aria/states_and_properties#aria-disabled

"In addition to setting the aria-disabled attribute, authors SHOULD change the appearance (grayed out, etc.) to indicate that the item has been disabled."

### DIFF
--- a/lib/checks/color/color-contrast.js
+++ b/lib/checks/color/color-contrast.js
@@ -9,7 +9,7 @@ if (fgColor === null || bgColor === null) {
 
 var nodeStyle = window.getComputedStyle(node);
 
-if(node.getAttribute('disabled') === 'disabled' || node.getAttribute('aria-disabled') === 'true') {
+if(node.hasAttribute('disabled') || node.getAttribute('aria-disabled') === 'true') {
     return true;
 }
 
--- a/lib/checks/color/color-contrast.js
+++ b/lib/checks/color/color-contrast.js
@@ -9,7 +9,7 @@ if (fgColor === null || bgColor === null) {
 
 var nodeStyle = window.getComputedStyle(node);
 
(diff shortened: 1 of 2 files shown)


dylanb:
@BenMakesGames you must add tests for this new functionality

iandotkelly:
@BenMakesGames - sorry to be a pain, but this isn't the right place to put this.  Your additional checks against aria-disabled should probably go here:

https://github.com/dequelabs/axe-core/blob/master/lib/checks/color/color-contrast-matches.js

The way you've implemented this can also cause false negatives - where people incorrectly add disabled and aria-disabled to elements that shouldn't have them.  We need to be very specific about what elements actually support disabled and aria-disabled.  Also see https://github.com/dequelabs/axe-core/issues/95

dylanb:
@BenMakesGames are you working on fixing the issues raised?

dylanb:
@BenMakesGames we would like to have this fixed in the next release of axe-core. I have assigned this issue to the internal Deque backlog for aXe and when it gets pulled into a sprint, we will close this PR if the issues have not yet been addressed. Thank you for contributing to axe-core

BenMakesGames:
oh! sorry, I had forgotten all about this issue. (a co-worker submitted his own PR, and happened to notice my name here, so asked me about it :P) I barely remember the issue at this point, and am not terribly interested in getting back into it at this time.

if we happen to run up against this issue again in the future, perhaps I'll revisit it.

feel free to close.

PR #148, opened 2015-12-12
Fix viewport scroll bug

### LINKED ISSUE #147: Running the audit scrolls the viewport to the bottom of the page
When I run the audit, the viewport scrolls to the bottom of the page. This can be reproduced by running the axe extension in Firefox or running the javascript in console.

ckundo:
- Running the audit before would scroll the viewport to the bottom of the page.

[Closes #147]

### DIFF
--- a/lib/checks/color/color-contrast.js
+++ b/lib/checks/color/color-contrast.js
@@ -1,6 +1,7 @@
+var useScroll = !(options || {}).noScroll;
 var bgNodes = [],
-	bgColor = commons.color.getBackgroundColor(node, bgNodes),
-	fgColor = commons.color.getForegroundColor(node);
+	bgColor = commons.color.getBackgroundColor(node, bgNodes, useScroll),
+	fgColor = commons.color.getForegroundColor(node, useScroll);
 
 //We don't know, so we'll pass it provisionally
 if (fgColor === null || bgColor === null) {
--- a/lib/commons/color/get-background-color.js
+++ b/lib/commons/color/get-background-color.js
@@ -1,5 +1,5 @@
 /*global dom, color */
-/* jshint maxstatements: 31, maxcomplexity: 14 */
+/* jshint maxstatements: 32, maxcomplexity: 15 */
 //TODO dsturley: too complex, needs refactor!!
 
 /**
@@ -108,10 +108,11 @@ var getVisualParents = function(node, rect) {
  * there is no opaque ancestor element visually containing it, or because background images are used.
  * @param {Element} node
  * @param {Array} bgNodes array to which all encountered nodes should be appended
+ * @param {Boolean} useScroll
  * @return {Color}
  */
 //TODO dsturley; why is this passing `bgNodes`?
-color.getBackgroundColor = function(node, bgNodes) {
+color.getBackgroundColor = function(node, bgNodes, useScroll) {
 	var parent, parentColor;
 
 	var bgColor = getBackgroundForSingleNode(node);
@@ -122,7 +123,10 @@ color.getBackgroundColor = function(node, bgNodes) {
 		return bgColor;
 	}
 
-	node.scrollIntoView();
+	if(useScroll) {
+		node.scrollIntoView();
+	}
+
 	var rect = node.getBoundingClientRect(),
 		currentNode = node,
 		colorStack = [{
@@ -172,5 +176,6 @@ color.getBackgroundColor = function(node, bgNodes) {
 	while ((currColorNode = colorStack.pop()) !== undefined) {
 		flattenedColor = color.flattenColors(currColorNode.color, flattenedColor);
 	}
+
 	return flattenedColor;
 };
--- a/lib/rules/color-contrast.json
+++ b/lib/rules/color-contrast.json
@@ -1,5 +1,8 @@
 {
   "id": "color-contrast",
+  "options": {
+    "noScroll": false
+  },
   "selector": "*",
   "tags": [
     "wcag2aa",
--- a/test/commons/color/get-background-color.js
+++ b/test/commons/color/get-background-color.js
@@ -245,7 +245,7 @@ describe('color.getBackgroundColor', function () {
 		var shifted = fixture.querySelector('#shifted');
 		var parent = fixture.querySelector('#parent');
 		var bgNodes = [];
-		var actual = commons.color.getBackgroundColor(target, bgNodes);
+		var actual = commons.color.getBackgroundColor(target, bgNodes, true);
 		var expected = new commons.color.Color(0, 0, 0, 1);
 		if (commons.dom.supportsElementsFromPoint(document)) {
 			assert.deepEqual(bgNodes, [shifted]);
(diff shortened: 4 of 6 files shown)


ckundo:
Phantomjs is hanging locally, waiting to see what CI says.

dylanb:
@ckundo this call is required in order to make the color contrast checks work in all instances on all browsers. The only way to make this a little better would be to remember the scroll position at the beginning of an audit and then restore it at the end. This would make the browser scroll a bit but would restore it. If you'd like to make a change like that, it would be a welcome change...

ckundo:
Thanks @dylanb, I thought of returning focus. That'd be an improvement, but the jumpiness would still be distracting. Do you know which browsers/scenarios require this? I could work on a configuration option like `{ "browsers": ["IE-9+", "Firefox", ...] }` that we could use to toggle features like this.

dylanb:
@ckundo what is your use case? If we put browser-specific code into the library, then your use case would only "work" in the browsers that do not suffer from the problem, so the question would be, why not simply turn off the color contrast rule if you do not want this behavior?

ckundo:
@dylanb I'm including and running the audit automatically on a page. I was thinking that running the audit without the `scrollIntoView` could at least give results for browsers that don't require the node to be in the viewport, and silently pass on other browsers. That'd mean false negatives for some browsers.

dylanb:
@ckundo it could mean false positives too because of the nature of that issue, the end result will be the background always being calculated as white. This could lead to dark backgrounds causing lots of false positives.

I am not sure of the complete list of browsers that exhibit this problem. I know for a fact that Firefox does.

ckundo:
@dylanb got it, yes that makes sense about the false positives. What I'd like to do in that case is have the option to run the rule without scrolling, and then disable the rule in my app, except in a whitelisted set of browsers that I know will work without scroll. That means that axe-core would need to accept a configuration option to disable the scroll. Is that something you'd entertain?

dylanb:
@ckundo An option would be better how were you thinking of implementing and calling the option?

You could pass back `null` from `getBackgroundColor` which would result in false negatives - which this rule suffers from anyway due to background images and gradients.

ckundo:
@dylanb I updated the PR with a possible solution to accept a "noScroll" option to the rule: https://github.com/dequelabs/axe-core/commit/24b3879acb0bf25aac2b7de9a6b83af9868b5fc5

dylanb:
On a quick perusal of the changes that looks good to change the util function. You would need to make a change to the check to make use of this.

dylanb:
Here is an example of a check using options

https://github.com/dequelabs/axe-core/blob/master/lib/checks/language/valid-lang.js#L5

and the options in the .json file

https://github.com/dequelabs/axe-core/blob/master/lib/checks/language/valid-lang.json

lib/commons/color/get-background-color.js
```diff
@@ -108,11 +108,13 @@ var getVisualParents = function(node, rect) {
...
  * @param {Element} node
  * @param {Array} bgNodes array to which all encountered nodes should be appended
+ * @param {Object} options
  * @return {Color}
  */
 //TODO dsturley; why is this passing `bgNodes`?
-color.getBackgroundColor = function(node, bgNodes) {
+color.getBackgroundColor = function(node, bgNodes, options) {
 	var parent, parentColor;
+	var options = options || {};
```
dylanb: Should remove the var in this line as options is already defined

dylanb:
There are a bunch of lint problems with the code changes, you should fix those too

lib/checks/color/color-contrast.js
```diff
@@ -1,6 +1,12 @@
+var useScroll = true;
+
+if((options || []).indexOf('noScroll') !== -1) {
```
ckundo: @dylanb is this what you had in mind regarding options? The default will be to always scroll, unless the options array includes the string `"noScroll"`.

lib/commons/color/get-background-color.js
```diff
@@ -1,5 +1,5 @@
 /*global dom, color */
-/* jshint maxstatements: 31, maxcomplexity: 14 */
+/* jshint maxstatements: 32, maxcomplexity: 15 */
```
ckundo: I gave a shot at refactoring this to reduce complexity here: https://github.com/ckundo/axe-core/commit/2cac6de1dbcf2fdcbcebb2b3f636124cd6609619
I still had a failing test and gave up on it.

lib/rules/color-contrast.json
```diff
@@ -1,5 +1,6 @@
 {
   "id": "color-contrast",
+  "options": [],
```
ckundo: Should this default to empty?

ckundo:
@dylanb I went ahead with the changes you recommended, appeased jslint, and squashed the commits with a better commit message.

lib/rules/color-contrast.json
dylanb: I think it should default to false so that it is self-documenting

dylanb:
Here is an example of a custom call to configure

```
axe.configure({
	checks: [{
      id: "valid-lang",
      options: [ "dylan" ],
      evaluate: function(node, options) {
        var lang = (node.getAttribute("lang") || "").trim().toLowerCase();
        var xmlLang = (node.getAttribute("xml:lang") || "").trim().toLowerCase();
        var invalid = [];
        (options || []).forEach(function(cc) {
          cc = cc.toLowerCase();
          if (lang && (lang === cc || lang.indexOf(cc.toLowerCase() + "-") === 0)) {
            lang = null;
          }
          if (xmlLang && (xmlLang === cc || xmlLang.indexOf(cc.toLowerCase() + "-") === 0)) {
            xmlLang = null;
          }
        });
        if (xmlLang) {
          invalid.push('xml:lang="' + xmlLang + '"');
        }
        if (lang) {
          invalid.push('lang="' + lang + '"');
        }
        if (invalid.length) {
          this.data(invalid);
          return true;
        }
        return false;
      }
	}],
	rules:[]
})
```

lib/checks/color/color-contrast.js
dylanb: The options can be anything you specify. In this case, I think it would be better to make it an object with an attribute that is a boolean (truthy test), I would suggest changing this to:

```
if (!(options || {}).noScroll) {
```

and making the appropriate changes to the options below

test/commons/color/get-background-color.js
```diff
@@ -259,6 +259,19 @@ describe('color.getBackgroundColor', function () {
...
+		var targetEl = fixture.querySelector('#target');
+		var bgNodes = [];
+		var useScroll = false;
+		window.scroll(0, 0);
 
+		commons.color.getBackgroundColor(targetEl, bgNodes, useScroll);
 
+		assert.equal(window.pageYOffset, 0);
+		assert.equal(window.pageXOffset, 0);
+	});
```
dylanb: @ckundo you should probably add a test that tests to ensure that no scrolling occurs when the option is enabled - WDYT?

dylanb:
@ckundo just checkin to see if you are working on responses to the comments?

ckundo:
Hi Dylan, yes, I'll be updating the PR soon.

dylanb:
Awesome!!

dylanb:
@ckundo how is the PR going?

ckundo:
Oof, sorry for dropping this. I'm still interested in finishing it off, and have some time this week. Will work on it.

lib/checks/color/color-contrast.js
```diff
@@ -1,6 +1,12 @@
+var useScroll = true;
+
+if((options || {}).noScroll) {
```
ckundo: @dylanb you had suggested using negation here. Were you thinking initializing `useScroll` to false on line 1, then setting it to `true` if this expression is truthy?

ckundo:
@dylanb I made changes based on your comments. Do you want me to squash these commits so we can merge?

lib/checks/color/color-contrast.js
dylanb: I was suggesting the following to shorten this

```
var useScroll = !((options || {}).noScroll);
```

dylanb:
yeah, if you could squash these commits then I'll merge

ckundo:
@dylanb all set!

WilcoFiers:
@ckundo I've looked at your changes. The problem is that without scrolling, there are quite a few scenario's where we're not going to be able to find the correct ancestors. So please be aware that some of the 'passes' might come from not being able to look at the contrast difference, similar to how aXe doesn't look at contrast with background images. We are working on a feature which will improve on this.

ckundo:
thanks @wilco, that makes sense to me. Worth noting that in the case of phantomjs, I think it should work fine because one can load the whole page. But for client side injection I can see the concern for false negatives. Thanks!

ckundo:
@dylanb did this ever get released to npm? I'm not seeing the changes in axe.js in the node module.

seancurtis:
I'm also wondering why I'm still seeing this behaviour. We added some a11y checking to our React Storybook but now all story pages get scrolled to the bottom automatically. It's very frustrating.

seancurtis:
Seems like it was changed from useScroll to noScroll. Does that mean it only checks things within the viewport? Can't it just record the offset at the start of the check and set it back to that at the end?

dylanb:
@seancurtis we are working on a reset of the scroll and a new implementation of the scrolling in the first place. Stay tuned for that.

Also: we need both a reset and the "noScroll" option because dynamic analysis implementations (like react-axe) will want to avoid scrolling at all because of the flicker that can induce (even if it resets).

ndelangen:
I just found out about this, but it seems this disable scrolling feature is rather hidden away and undocumented.

Would a top level `noScroll` option make sense?
like `restoreScroll`?

Would an enum option like this make sense ?
```
scrollBehavior: 'disabled' | 'enabled' | 'restore'
```

PR #153, opened 2015-12-16
Take into account XHTML where there might be discrepancies with what …

yzen:
…case nodeName returns.


dylanb:
@yzen Thanks for this contribution!! It might be worthwhile trying to figure out how to actually test for this systematically. With this change, there are no tests, so this change will be a little fragile - especially to any new rules.

PR #168, opened 2016-02-26
Renaming await to awat to avoid collision with await reserved keyword…

deltakosh:
… (ES2016)


dylanb:
Thanks for the contribution!!

PR #172, opened 2016-03-03
Allow `include` to be specified, but empty

jasonkarns:
Right now, if context is given as `{include: [], exclude[]}`, the `include` piece does not get properly defaulted to `[document]`.

The following are already accepted and work as expected:
- `{include: null}`
- `{include: undefined}`
- `{include: ""}`
- `{exclude: []}`

I don't think there is an appreciable difference for the `include` key between `null`, `undefined`, `""` and `[]`, so I think the following should be supported as well:

`{include: []}`

(To be fair, right now any falsey value for `include` would default to `[document]`, but I'm not suggesting that `0` or `false` should be supported, even though they current work.)

(code hand-edited directly on github, so it probably doesn't conform to style-guidelines)

### DIFF
--- a/test/core/base/context.js
+++ b/test/core/base/context.js
@@ -316,6 +316,12 @@ describe('Context', function() {
 
 		});
 
+		it('should default empty-object include to document', function () {
+			var result = new Context({ include: {}, exclude: {} });
+			assert.lengthOf(result.include, 1);
+			assert.equal(result.include[0], document);
+		});
+
 	});
 
 	describe('initiator', function() {
@@ -342,7 +348,7 @@ describe('Context', function() {
 			var spec = document.implementation.createHTMLDocument('ie is dumb');
 			spec.hasOwnProperty = undefined;
 			var result = new Context(spec);
-			
+
 			assert.lengthOf(result.include, 1);
 			assert.equal(result.include[0], spec);
 
--- a/test/core/base/context.js
+++ b/test/core/base/context.js
@@ -316,8 +316,8 @@ describe('Context', function() {
 
 		});
 
-		it('should default empty-object include to document', function () {
-			var result = new Context({ include: {}, exclude: {} });
+		it('should default empty include to document', function () {
+			var result = new Context({ include: [], exclude: [] });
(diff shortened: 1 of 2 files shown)


dylanb:
@jasonkarns can you please fix the jshint failure.

jasonkarns:
Sure thing. I figured I'd verify that the new behavior was acceptable first
:)

jasonkarns:
fixed!

dylanb:
What is the purpose of the `+` sign in front of `context.include.length`?

jasonkarns:
It's a way to forcefully coerce the value to a `Number`. I'm personally not a fan of it, but I noticed it was used here: https://github.com/dequelabs/axe-core/pull/172/files#diff-f6abb6c181b36040217a018122420854R92 so I followed the same style.

Honestly, I don't really think it's all that necessary to ensure `context.include.length` is a `Number`. The likelihood that `context.include` has a `length` property that is defined *and* not falsey *and* not a number has got to be astronomically low. At which point, there are probably other errors popping up because `a11yCheck` is clearly not being invoked as expected.

But I would at least prefer all the `.length` checks to be consistent, so if we drop the number coercion, we should probably question why it's necessary here: https://github.com/dequelabs/axe-core/pull/172/files#diff-f6abb6c181b36040217a018122420854R92

dylanb:
Final question (should have asked earlier): could we have a test case to test for this specific use?

jasonkarns:
test added!

test/core/base/context.js
```diff
@@ -316,6 +316,12 @@ describe('Context', function() {
 
 		});
 
+		it('should default empty-object include to document', function () {
+			var result = new Context({ include: {}, exclude: {} });
```
dylanb: should these be arrays?

jasonkarns: Yep. Thanks. The empty-object case (though not "supported") is handled by the same logic as the empty array so the test was still passing and I didn't catch it.

jasonkarns:
Followup on this (since adding commits don't trigger notifications).

PR #173, opened 2016-03-04
Reorder Results/Parameters section in API doc

jasonkarns:
The API documentation originally had:

a11yCheck
  - purpose
  - description
  - synopsis
  - parameters
  - results
  - a11yCheck parameters
    - context
    - options
    - callback

The documentation for the a11yCheck parameters was split up. There was a brief 'overview' of the parameters after the synopsis, then the more detailed section about each individual parameter was *after* the return/results object documentation.

This change simply unifies the parameter documentation to a single location, following the same pattern as for the other documented public methods.

Additionally, with the simplified outline for the a11yCheck method, more deep-links were added to the TOC at the top. (deep linking to the parameters section itself, as well as to the details for each of the three parameters)

No changes to the actual documentation of the parameters or results have been made.

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -136,9 +136,9 @@ axe.a11yCheck(context, options, callback);
 
 #### Parameters
 
-* `context`: Defines the scope of the analysis - the part of the DOM that you would like to analyze. This will typically be the `document` or a specific selector such as class name, ID, selector, etc.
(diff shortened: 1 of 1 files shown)


dylanb:
Thanks for the improvement @jasonkarns

PR #177, opened 2016-03-09
Allow axe to optionally be require()'d or define()'d

nodren:
Currently, it's very difficult to use Axe with node.  While window and document are not normally defined, they can be defined with packages such as [jsdom](https://github.com/tmpvar/jsdom), which is useful when testing various front end frameworks.

This small change will keep axe's default behavior of being attached to window(in absence of a require type tool), but also allow it to be imported via require in node, as well as front end forms of importing packages.

### DIFF
--- a/build/tasks/nodeify.js
+++ b/build/tasks/nodeify.js
@@ -10,7 +10,9 @@ module.exports = function (grunt) {
 				return grunt.file.read(fn);
 			}).join('\n');
 
-			grunt.file.write(fileset.dest, 'exports.source = ' + JSON.stringify(source) + ';');
+			var file = grunt.file.read(fileset.dest);
+
+			grunt.file.write(fileset.dest, file + 'module.exports.source = ' + JSON.stringify(source) + ';');
 		});
 	});
 };
--- a/lib/outro.stub
+++ b/lib/outro.stub
@@ -1,3 +1,4 @@
 
 	axe.version = '<%= pkg.version %>';
+	if (typeof define === "function" && define.amd) this.axe = axe, define(axe); else if (typeof module === "object" && module.exports) module.exports = axe; else window.axe = axe;
 }(window, window.document));
--- a/test/core/index.js
+++ b/test/core/index.js
@@ -1,10 +1,3 @@
 describe('index', function () {
 	'use strict';
-
-	it('should redefine `define`', function () {
-		assert.equal(typeof define, 'undefined');
-	});
-	it('should redefine `require`', function () {
-		assert.equal(typeof require, 'undefined');
-	});
 });
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -16,12 +16,6 @@ module.exports = function (grunt) {
 	grunt.initConfig({
 		pkg: grunt.file.readJSON('package.json'),
 		clean: ['dist', 'tmp'],
-		nodeify: {
-			core: {
-				src: ['<%= concat.engine.dest %>'],
-				dest: 'dist/index.js'
-			}
-		},
 		'update-help': {
 			options: {
 				version: '<%=pkg.version%>'
@@ -238,7 +232,7 @@ module.exports = function (grunt) {
 	grunt.registerTask('default', ['build']);
 
 	grunt.registerTask('build', ['clean', 'validate', 'concat:commons', 'configure',
-		'concat:engine', 'copy', 'nodeify', 'uglify']);
+		'concat:engine', 'copy', 'uglify']);
 
 	grunt.registerTask('test', ['build',  'testconfig', 'fixture', 'connect',
 		'mocha', 'jshint']);
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -16,6 +16,12 @@ module.exports = function (grunt) {
 	grunt.initConfig({
 		pkg: grunt.file.readJSON('package.json'),
 		clean: ['dist', 'tmp'],
+		nodeify: {
+			core: {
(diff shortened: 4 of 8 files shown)


dylanb:
The spirit of this change is good but it undoes a requirement that the source of axe be available after a node style require to be able to use that source to inject into a remote browser (like Webdriver) without having to have a web server available to serve the axe.js file into the remote browser.

You can see how that is being done in this project https://github.com/dequelabs/axe-webdriverjs/blob/master/lib/inject.js#L4

Can you modify your PR so that both requirements can be met?

nodren:
@dylanb I've committed a change which should meet your requirement of require('axe-core').source to still work as before, while also now allowing require('axe-core') to return axe itself, best of both world's scenario, and doesn't require any changes to the webdriver plugin you mentioned to work.

test/core/index.js
```diff
@@ -1,10 +1,3 @@
 describe('index', function () {
 	'use strict';
-
```
dylanb: Are there tests we could add here to test that your changes are effective?

build/tasks/nodeify.js
```diff
@@ -10,7 +10,9 @@ module.exports = function (grunt) {
 				return grunt.file.read(fn);
 			}).join('\n');
 
-			grunt.file.write(fileset.dest, 'exports.source = ' + JSON.stringify(source) + ';');
+			var file = grunt.file.read(fileset.dest);
+
+			grunt.file.write(fileset.dest, file + 'module.exports.source = ' + JSON.stringify(source) + ';');
```
marcysutton: Including the source twice here makes the filesize jump to 432kb. Is there any way to minimize repeating it?

lib/outro.stub
```diff
@@ -1,3 +1,4 @@
 
 	axe.version = '<%= pkg.version %>';
+	if (typeof define === "function" && define.amd) this.axe = axe, define(axe); else if (typeof module === "object" && module.exports) module.exports = axe; else window.axe = axe;
```
marcysutton: Can you limit line length to 100 chars?

build/tasks/nodeify.js
WilcoFiers: Doesn't seem right indeed. How about instead we write the code to a different file, and have the current dest be a file that does two things: First it requires axe-core as a node module, and second it reads out that file so it can also expose the source.

lib/outro.stub
dylanb: @marcysutton should we change the jshint settings? This should really be picked up automatically.

build/tasks/nodeify.js
dylanb: @WilcoFiers can you elaborate? I don't understand what you mean by "reads out that file"

nodren:
@marcysutton there's absolutely options for doing this differently, however as @dylanb pointed out to me earlier, your webdriver plugin expects something at require('axe-core').source.  changing this would mean also having to change the webdriver plugin as well.

@WilcoFiers another option is to simply have webdriver read the file require()'d file using the [fs](https://nodejs.org/api/fs.html#fs_fs_readsync_fd_buffer_offset_length_position) node library, finding the location of that file using [require.resolve('axe-core')](https://nodejs.org/api/globals.html#globals_require_resolve).  Doing this we can completely drop that nodeify grunt plugin and nodeify task in gulp, as well as the index.js and have package.json simply point right to axe.js as bower.json does.  It will require a PR to the web driver plugin as well to get that sorted.

dylanb:
@marcysutton we are already extending the interface to axe-webdriverjs to allow for use with custom versions of axe-core. Could we kill two birds with one stone and get rid of the direct inclusion of axe-core in axe-webdriverjs altogether? This would make the use of it more consistent with the changes we have been contemplating.

marcysutton:
@dylanb I think that would be ok. I have already been demo'ing it with a separate axe-core install to support custom builds, so it makes sense.

dylanb:
@marcysutton can you coordinate that change with the changes to this PR so they work together well?

@nodren could you work with @marcysutton on this so that we commit both fixes at the same time?

nodren:
@dylanb @marcysutton I'm somewhat confused, am I just going to drop the exports.source from index.js in this PR like I suggested?

marcysutton:
@nodren yes, I believe so. We are still working out the changes required for axe-webdriverjs, I will coordinate with you once we decide on a direction.

dylanb:
ok, I have a solution that does not change the API for axe-core at all

```
(function axeFunction (window, document) {
	var axe = {};
	axe.version = "2.0.0";
	if (typeof define === "function" && define.amd) {
		this.axe = axe;
		define(axe);
	} else if (typeof module === "object" && module.exports) {
		axe.source = '(' + axeFunction.toString() + ')(this, this.document);'
		module.exports = axe;
	} else {
		window.axe = axe;
	}
})(this, this.document);
```

dylanb:
@nodren I had to make some more changes to get everything to work. Thanks for this commit. If you'd like to get credit for it, I could send you the full set of changes so you can modify your PR otherwise look for it in the 2.0.0 release.

nodren:
@dylanb No need, I don't need credit here, i'm more interested in the end result, which it looks like we've accomplished, looking forward to seeing this in 2.0!

nodren:
What's the timeframe for 2.0?  Axe in it's current form is unusuable for us with our workflow of requiring node modules.

marcysutton:
@nodren check the latest updates, we merged some code that should support your use case. Npm still needs updating, but you should be able to build from the Github repo.

PR #180, opened 2016-03-26
Babelify Axe

jessebeach:
Run browserify and the babel transform on the axe.js dist file.


WilcoFiers:
Awesome! Thanks Jesse

PR #187, opened 2016-04-15
Make project root the packable tarball

jasonkarns:
- Stop copying manifest files into dist
- main is now in dist/index.js relative to package.json
- Explicitly specify the files to be included in the tarball

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,8 @@
 tmp
-dist
 bower_components
 node_modules
+axe.js
+axe.min.js
 test/*/index.html
 test/integration/*/index.html
 .grunt
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -141,23 +141,6 @@ module.exports = function (grunt) {
 				}
 			}
 		},
-		copy: {
-			manifests: {
-				files: [{
-					src: ['package.json'],
-					dest: 'dist/'
-				}, {
-					src: ['README.md'],
-					dest: 'dist/'
-				}, {
-					src: ['bower.json'],
-					dest: 'dist/'
-				}, {
-					src: ['LICENSE'],
-					dest: 'dist/'
-				}]
-			}
-		},
 		watch: {
 			files: ['lib/**/*', 'test/**/*.js', 'Gruntfile.js'],
 			tasks: ['build', 'testconfig', 'fixture']
@@ -251,7 +234,7 @@ module.exports = function (grunt) {
 	grunt.registerTask('default', ['build']);
 
 	grunt.registerTask('build', ['clean', 'validate', 'concat:commons', 'configure',
-		'concat:engine', 'babel', 'copy', 'uglify']);
+		'concat:engine', 'babel', 'uglify']);
 
 	grunt.registerTask('test', ['build', 'testconfig', 'fixture', 'connect',
 		'mocha', 'jshint']);
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -7,7 +7,6 @@ module.exports = function (grunt) {
 	grunt.loadNpmTasks('grunt-contrib-clean');
 	grunt.loadNpmTasks('grunt-contrib-concat');
 	grunt.loadNpmTasks('grunt-contrib-connect');
-	grunt.loadNpmTasks('grunt-contrib-copy');
 	grunt.loadNpmTasks('grunt-contrib-jshint');
 	grunt.loadNpmTasks('grunt-contrib-uglify');
 	grunt.loadNpmTasks('grunt-contrib-watch');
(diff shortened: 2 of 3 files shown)


jasonkarns:
Assuming that `dist/axe.js` is node-requireable? At present, master doesn't have any `index.js` file at all to be required. Nothing in root nor in dist after a build. Assuming that #177 made dist/axe.js node-requireable?

`npm pack` to see the tarball that would be published to npm

Question: Is bower.json actually necessary in the published tarball? The current bower.json file makes no sense. It contains ignore patterns that are relative to the project root:

```
"**/.*",
"node_modules",
"bower_components",
"test",
"build"
```

This implies that bower will be reading the bower.json file at the root of the repo (via git, presumably). But if that's the case, why is bower.json copied to dist at all? That's only necessary if bower were reading from the npm registry. Further, from the package root, there is no `axe.js` file, so the bower.json file is invalid.

Alternatively, if bower is somehow instructed to only use the `dist` directory (which would explain why the bower.json is in there, and how the 'main' still finds axe.js -- though I don't know see anything that instructs bower to use dist), then the entire list of ignore patterns is unnecessary, since it's listing files that don't exist under `dist/`.

Clarification?

dylanb:
@jasonkarns can you make it build into the root directory instead of dist. That will make this change compatible with the 2.0 branch

jasonkarns:
Sure.

It's worth mentioning the downsides:
- it's trivial to gitignore dist/* but now we'll need to gitignore each generated file (not difficult but slightly brittle if any new files are generated)
- same as above for identifying files that need removed on `grunt clean`

Will update the PR. Will still need bower support checked, as I'm not sure of the bower release process.

dylanb:
@jasonkarns I think most of these issues have been resolved in the recent changes - possibly only the explicit listing of files to be included? Although I think excluding files through .npmignore is a better practice than explicitly including files. WDYT?

jasonkarns:
Actually, I strongly advise against `.npmignore` for a couple reasons.

If .npmignore is missing, npm will use an existing .gitignore in its place. Thus most node developers get in the habit of only keeping an eye on the git-excluded files. (And I'm not even certain whether npm uses the literal .gitignore or if it uses the calculated excludes: which would take into account system and user-level gitignores, as well as repo-level ignores from .git/info/exclude.)

So once developers suddenly need their gitignore and npmignore to differ, they typically just create an npmignore that has additional ignore patterns above and beyond the gitignores. But suddenly, now that .npmignore exists, npm no longer respects the .gitignore. So very soon after creating an npmignore file, I typically see a follow-up commit that copies the .gitignore into the .npmignore (along with the additional npm ignores). And now suddenly you have two different ignore files to maintain and keep in sync since they will likely have _very_ similar contents but with very minor (and hard to distinguish) differences. You're almost guaranteed to add an ignore pattern to one file or the other and forget to add it to both.

Which brings me to the `files` whitelist. npm will include `package.json`, `LICENSE{.*}` and `README{.*}` in the package tarball regardless if they are in the whitelist or blacklist. Which means, in 99% of cases, the only additional files one needs to include in the package is `index.js` and `lib/` (or `src/` or `dist/`, depending on build setup and target platform). Indeed, the npm team have indicated that they intend to begin including the `main` file automatically (and likely `bin` and `directories.{lib,bin}` too).

I find that it's much much simpler to define a (typically two-item) whitelist in package.json than to try and maintain two different ignore files that are eerily similar but not exactly the same. Especially because the whitelist is almost certainly shorter than the blacklist.

dylanb:
@jasonkarns I see your point and do think the npm approach of a standard format for a directory structure would be the ideal solution - however isn't the whitelist simply just a different 2nd list that needs to be maintained in addition to .gitignore?

jasonkarns:
So, since the files we want to track in git and the files we want to end up in the package aren't the same, there's definitely going to be two different "lists". My rationale for choosing the `files` whitelist instead of the `.npmignore` blacklist is:

- whitelist is shorter
- whitelist is less likely to change
- using the blacklist means making the same change in two files. using the whitelist, it's more likely that changes are only necessary to .gitignore since the whitelist rarely changes.
- `.npmignore` is usually forgotten by devs since most npm packages don't even have one (because they rely on npm's default gitignore behavior)

it's more common to continue adding directories to ignore (which would have to go in _both_ ignore files) but it's very rare to create a new directory that needs included in the package (since everything is usually just in `lib` or `dist`). Setting `files: ["index.js", "lib"]` is usually the last time the whitelist needs touched, but the blacklist gets additions regularly (for testing tools, development config files, generated reports, test data, editor config files, etc)

Either way, `npm pack` will create the tarball locally so the package contents can be verified without publishing.

dylanb:
ok, noted. I am going to close this PR because we have fixed it, just not in the way you suggested.

PR #193, opened 2016-06-17
Exclude DOM Nodes when testing for arrays in parseSelectorArray

BrianSipple:
I came across this issue while using `axe-core` in the `ember-a11y-testing` project, and I [made some notes](https://github.com/ember-a11y/ember-a11y-testing/issues/29) there that may make for a useful reference, but to elaborate...

When axe.a11yCheck is passed an HTMLFormElement node as its context, and when that element contains nested controls (for example: `HTMLInputElement` nodes), an error will be thrown within the `parseSelectorArray` function because it's [currently checking](https://github.com/dequelabs/axe-core/blob/4ef74435ac92a88c602690ac1e243c404f750d00/lib/core/base/context.js#L131) whether or not the `item` is an array with `if (item && item.length)`, and `length` also happens to be a [property of `HTMLFormElement`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/length). 

From `ember-a11y-testing` when I placed a `button` inside of a `form` (Although any [control type](https://www.w3.org/TR/html401/interact/forms.html#h-17.2.1) element will do the trick):
screen shot 2016-06-17 at 12 59 28 am

This PR proposes checking for Arrays by using `Array.isArray`, and also adds a passing test for this behavior that will indeed fail if it's run with the current implementation. Also, this is my first time really digging into the `axe-core` project itself, so if anyone has pointers on how this should truly be tested and whether or not the current test additions could be improved, I'd love to get feedback on that as well.

### DIFF
--- a/lib/core/base/context.js
+++ b/lib/core/base/context.js
@@ -128,7 +128,7 @@ function parseSelectorArray(context, type) {
 		if (typeof item === 'string') {
 			result = result.concat(utils.toArray(document.querySelectorAll(item)));
 			break;
-		} else if (item && item.length) {
+		} else if (item && item.length && !(item instanceof Node)) {
 
 			if (item.length > 1) {
 				pushUniqueFrameSelector(context, type, item);
--- a/test/core/base/context.js
+++ b/test/core/base/context.js
@@ -51,7 +51,33 @@ describe('Context', function() {
 
 		});
 
-		it('should accept an array of node reference', function() {
+              it('should accept a node reference consisting of nested divs', function() {
+                     var div1 = document.createElement('div');
+                     var div2 = document.createElement('div');
+
+                     div1.appendChild(div2);
+                     fixture.appendChild(div1);
+
+                     var result = new Context(div1);
+
+                     assert.deepEqual(result.include, [div1]);
+
+              });
+
+              it('should accept a node reference consisting of a form with nested controls', function() {
+                     var form = document.createElement('form');
+                     var input = document.createElement('input');
+
+                     form.appendChild(input);
+                     fixture.appendChild(form);
+
+                     var result = new Context(form);
+
+                     assert.deepEqual(result.include, [form]);
+
+              });
+
+		it('should accept an array of node references', function() {
 			fixture.innerHTML = '<div id="foo"><div id="bar"></div></div>';
 
 			var result = new Context([$id('foo'), $id('bar')]);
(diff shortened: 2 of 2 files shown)


lib/core/base/context.js
```diff
@@ -128,7 +128,7 @@ function parseSelectorArray(context, type) {
 		if (typeof item === 'string') {
 			result = result.concat(utils.toArray(document.querySelectorAll(item)));
 			break;
-		} else if (item && item.length) {
+		} else if (item && Array.isArray(item)) {
```
nathanhammond: Not a maintainer over here, but `Array.isArray` may be too new for the desired browser support of axe-core. I'd probably test it this way: `Object.prototype.toString.call(item) === '[object Array]';`

nathanhammond: (Or may add to `utils` an `isArray` function.)

nathanhammond: Retracted: `isArray` is used elsewhere in the project. LGTM.

dylanb:
Array.isArray works on all the required browsers (IE9+) http://kangax.github.io/compat-table/es5/#test-Array.isArray

test/core/base/context.js
```diff
@@ -51,7 +51,33 @@ describe('Context', function() {
 
 		});
 
-		it('should accept an array of node reference', function() {
+		it('should accept a node reference consisting of nested divs', function() {
+			var div1 = document.createElement('div');
+			var div2 = document.createElement('div');
+
+      div1.appendChild(div2);
```
dylanb: can you normalize the indentation

test/core/base/context.js
```diff
@@ -51,7 +51,33 @@ describe('Context', function() {
...
+
+			var result = new Context(div1);
+
+			assert.deepEqual(result.include, [div1]);
+
+		});
+
+		it('should accept a node reference consisting of a form with nested controls', function() {
+			var form = document.createElement('form');
+      var input = document.createElement('input');
```
dylanb: same

lib/core/base/context.js
dylanb: This test will not allow jQuery objects to work anymore, we should probably be excluding actual nodes here rather than testing for an array

BrianSipple:
@dylanb I updated the check to use `if (item && item.length && !(item instanceof Node))` as per the [suggestion](https://github.com/dequelabs/axe-core/pull/193#discussion_r67885039), and this seems to be effective as well. Also, indentation should be fixed.

dylanb:
Thanks for the contribution!!

Issue #194, opened 2016-06-21
Add tests for context with jQuery


dylanb:
Need some integration tests that ensure that jQuery objects will work as a context for a11yCheck

Also include and exclude

PR #198, opened 2016-07-01
Add tests for jQuery objects, include/exclude

marcysutton:
Closes https://github.com/dequelabs/axe-core/issues/194

### DIFF
--- a/test/integration/full/jquery/a11ycheck-object-exclude.js
+++ b/test/integration/full/jquery/a11ycheck-object-exclude.js
@@ -1,0 +1,13 @@
+/*global $ */
+
+describe('jQuery object as a11yCheck context', function () {
+	'use strict';
+	it('should find no violations', function (done) {
+		var target = $('#target')[0];
+		axe.a11yCheck({ exclude: [target] }, {}, function (results) {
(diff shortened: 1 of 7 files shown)


test/integration/full/jquery/a11ycheck-object-exclude.js
```diff
@@ -0,0 +1,13 @@
+/*global $ */
+
+describe('jQuery object as a11yCheck context', function () {
+	'use strict';
+	it('should find no violations', function (done) {
+		var target = $('#target')[0];
+		axe.a11yCheck({ exclude: [target] }, {}, function (results) {
+			assert.lengthOf(results.violations, 0, 'violations');
+			assert.lengthOf(results.passes, 11, 'passes');
```
marcysutton: @dylanb to confirm this is working as expected, should there be passes for an excluded target that has failures in it (meaning should they run at all)?

dylanb:
please merge into bitbucket develop and master

test/integration/full/jquery/a11ycheck-object-exclude.js
dylanb: @marcysutton I believe that they should not run at all

PR #200, opened 2016-07-08
Fix preserveComments in uglify which was not removing any comments

iandotkelly:
https://github.com/gruntjs/grunt-contrib-uglify/issues/366

This change very specifically defines which comments will be kept - which only includes those starting with /*!

This has been tested in the building of axe.min.js


marcysutton:
This looks good to me @iandotkelly

PR #205, opened 2016-08-04
Improve contrast detection for elements that are behind another element

mfairchild365:
I was seeing false negatives (not seeing contrast errors that should be detected) on pages where there was a sticky header attached to the viewport, while using the axe chrome extension. After diving into the issue, I discovered that it was due to axe scrolling to the element while testing a page. Scrolling to the element would place the element at the top of the page, which would also place the element behind a sticky header. Because axe tries to be smart and determine the background via visual order, it was finding the sticky header instead of the element with bad contrast.

For example, view https://mfairchild365.github.io/axe-contrast-bug/
That example has two contrast errors, but axe only finds one (on the sticky header).

Note: I'm not confident that the automated tests will detect if this is not working, I'm not sure if it is because phantomjs does not support `elementFromPoint()` or perhaps phantomjs does not need to scroll because it is a headless browser.

### DIFF
--- a/lib/commons/dom/elements-from-point.js
+++ b/lib/commons/dom/elements-from-point.js
@@ -21,7 +21,11 @@ dom.supportsElementsFromPoint = function (doc) {
  * @return {Array} Array of Elements
  */
 dom.elementsFromPoint = function (doc, x, y, targetNode) {
-	var elements = [], finalElements = [], previousPointerEvents = [], current, i, d, targetFound = false, floatingPositions = ['fixed', 'sticky'], style, currentNode, index;
+	var elements = [],
+		previousPointerEvents = [],
+		current,
+		i,
+		d;
 
 	targetNode = targetNode || false;
 
@@ -48,10 +52,26 @@ dom.elementsFromPoint = function (doc, x, y, targetNode) {
 		elements[i].style.setProperty('pointer-events', d.value ? d.value : '', d.priority);
 	}
 
+	elements = dom.reduceToElementsBelowFloating(elements, targetNode);
+
+	// return our results
+	return dom.reduceToFirstOpaque(elements);
+};
+
+/**
+ * Reduce an array of elements to only those that are below a 'floating' element.
+ * 
+ * @param {Array} elements
+ * @param {Element} targetNode
+ * @returns {Array}
+ */
+dom.reduceToElementsBelowFloating = function (elements, targetNode) {
+	var floatingPositions = ['fixed', 'sticky'], finalElements = [], targetFound = false, index, currentNode, style;
+	
 	// Filter out elements that are temporarily floating above the target
-	for (var index = 0; index < elements.length; ++index) {
+	for (index = 0; index < elements.length; ++index) {
 		currentNode = elements[index];
-		if (currentNode == targetNode) {
+		if (currentNode === targetNode) {
 			targetFound = true;
 		}
 
@@ -65,11 +85,19 @@ dom.elementsFromPoint = function (doc, x, y, targetNode) {
 
 		finalElements.push(currentNode);
 	}
-
-	elements = finalElements;
-	finalElements = [];
 	
-	//Now only return elements to the first opaque
+	return finalElements;
+};
+
+/**
+ * Reduce an array of elements to only those before the first opaque
+ * 
+ * @param {Array} elements array of elements to reduce
+ * @returns {Array}
+ */
+dom.reduceToFirstOpaque = function (elements) {
+	var finalElements = [], currentNode, index;
+	
 	for (index = 0; index < elements.length; ++index) {
 		currentNode = elements[index];
 
@@ -79,7 +107,6 @@ dom.elementsFromPoint = function (doc, x, y, targetNode) {
 			break;
 		}
 	}
-
-	// return our results
+	
 	return finalElements;
 };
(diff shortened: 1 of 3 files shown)


mfairchild365:
I realize some work might still need to be done with this. Please let me know if you want me to change anything, have any questions, or if I went about this all wrong. :)

dylanb:
@marcysutton can you review the logic of this change and make recommendations?

lib/commons/dom/elements-from-point.js
```diff
@@ -29,8 +36,6 @@ dom.elementsFromPoint = function (doc, x, y) {
 
 	// get all elements via elementFromPoint, and remove them from hit-testing in order
 	while ((current = doc.elementFromPoint(x, y)) && elements.indexOf(current) === -1 && current !== null) {
-
-		// push the element and its current style
```
marcysutton: Can you restore the comment here?

mfairchild365: Sure! I didn't realize that I had removed it. It has been restarted in https://github.com/dequelabs/axe-core/pull/205/commits/943644b69e3ec172ef81f8857d7b1b5363af385b

lib/commons/dom/elements-from-point.js
```diff
@@ -40,15 +47,68 @@ dom.elementsFromPoint = function (doc, x, y) {
...
+
+/**
+ * Reduce an array of elements to only those that are below a 'floating' element.
+ * 
+ * @param {Array} elements
+ * @param {Element} targetNode
+ * @returns {Array}
+ */
+dom.reduceToElementsBelowFloating = function (elements, targetNode) {
+	var floatingPositions = ['fixed', 'sticky'], finalElements = [], targetFound = false, index, currentNode, style;
```
marcysutton: Can you put these variable declarations onto separate lines?

lib/commons/dom/elements-from-point.js
```diff
@@ -40,15 +47,68 @@ dom.elementsFromPoint = function (doc, x, y) {
...
+	
+	// Filter out elements that are temporarily floating above the target
+	for (index = 0; index < elements.length; ++index) {
+		currentNode = elements[index];
+		if (currentNode === targetNode) {
+			targetFound = true;
+		}
+
+		style = window.getComputedStyle(currentNode);
```
marcysutton: Is it possible to get `style.position` directly instead of calling `window.getComputedStyle`? I want to make sure we aren't making performance worse than it already is for pages with lots of elements.

lib/commons/dom/elements-from-point.js
```diff
@@ -40,15 +47,68 @@ dom.elementsFromPoint = function (doc, x, y) {
...
+};
+
+/**
+ * Reduce an array of elements to only those before the first opaque
+ * 
+ * @param {Array} elements array of elements to reduce
+ * @returns {Array}
+ */
+dom.reduceToFirstOpaque = function (elements) {
+	var finalElements = [], currentNode, index;
```
marcysutton: Please put these declarations onto separate lines. Thanks!

mfairchild365: fixed in https://github.com/dequelabs/axe-core/pull/205/commits/23070e79f5ca57d8d71514ea7a7b673cd89308ba

lib/commons/dom/elements-from-point.js
mfairchild365: fixed in https://github.com/dequelabs/axe-core/pull/205/commits/23070e79f5ca57d8d71514ea7a7b673cd89308ba

lib/commons/dom/elements-from-point.js
mfairchild365: Unfortunately, I'm not aware of a better way than `window.getComputedStyle()`.

Something like `element.style.position` is possible, but only returns styles that were defined on the element inline. In the cases that I was testing under, the position styles were all defined in css files, and `element.style.position` returned an empty string.

It might be possible to cache style information in memory to reduce the calls to `window.getComputedStyle()`, but then I'd be worried about memory bloat and state changes.

If you know of a better way, I'd be happy to hear it. :) My js experience is limited.

dylanb: It is not possible to do this another way unfortunately.

test/checks/color/color-contrast.js
```diff
@@ -103,6 +103,34 @@ describe('color-contrast', function () {
 		assert.deepEqual(checkContext._relatedNodes, [target]);
 	});
 
+
+	it('should ignore position:fixed elements above the target', function () {
```
dylanb: @WilcoFiers can you see whether there are any other test cases we should add to this?

lib/commons/dom/elements-from-point.js
marcysutton: Perf benchmarks would really help us to determine how costly these additions are. I will make an issue internally to investigate it.

mfairchild365:
I just thought I'd follow up on this. Is there anything that I can do to move this forward?

marcysutton:
Hi! I will merge it. These changes will also have to be merged with our code internally, but @WilcoFiers said he was willing to take that on. 

We want to make substantial changes to improve performance for testing color contrast but as that's a bigger undertaking, it might take a while....so we should address this issue now.

Issue #209, opened 2016-08-09
Link rules?


badtant:
Hi,

I've created the following links and wanted to check if they are valid or not:
```
1. <a href=""></a>
2. <a href="">test2</a>
3. <a href="#">test3</a>
```

I only get an error on no 1. "Links must have discernible text". 
Are the other ones acceptable? I know that react-a11y and eslint-plugin-jsx-a11y check for no 3. Maybe that is something to include in axe too?

marcysutton:
@dylanb I'd be curious to hear your take. When I tested this issue recently every screen reader I tested used the link text and not the URL, even in the rotor/links list. http://codepen.io/marcysutton/pen/AXaAyg

dylanb:
Only anchor 1 is an accessibility problem. The other ones "may" be accessibility problems. For example if the JavaScript attached to those anchors is looking for anything other than a click event to do its magic, then the action might not be triggered using a screen reader or keyboard. There is no way that axe can tell that, so we don't flag it because it is most likely to be a false positive.

badtant:
Ok, so no 2 is fine. But doesn't no 3 indicate that you should use a button instead?

marcysutton:
As Dylan mentioned, it would likely be a false positive. Perhaps we could add a Best Practice rule instead, though.

badtant:
Sounds good to me :)

PR #210, opened 2016-08-11
add best practice rule href-no-hash

badtant:
New best practice rule as discussed here: https://github.com/dequelabs/axe-core/issues/209

### DIFF
--- a/doc/rule-descriptions.md
+++ b/doc/rule-descriptions.md
@@ -22,6 +22,7 @@
 | empty-heading | Ensures headings have discernible text | best-practice |
 | frame-title | Ensures &lt;iframe&gt; and &lt;frame&gt; elements contain a unique and non-empty title attribute | wcag2a, wcag241, section508, section508.22.i |
 | heading-order | Ensures the order of headings is semantically correct | best-practice |
+| href-no-hash | Ensures an anchors href value is not just # | best-practice |
 | html-has-lang | Ensures every HTML document has a lang attribute | wcag2a, wcag311 |
 | html-lang-valid | Ensures the lang attribute of the &lt;html&gt; element has a valid value | wcag2a, wcag311 |
 | image-alt | Ensures &lt;img&gt; elements have alternate text or a role of none or presentation | wcag2a, wcag111, section508, section508.22.a |
--- a/doc/rule-descriptions.md
+++ b/doc/rule-descriptions.md
@@ -1,56 +1,56 @@
-| Rule ID | Description | Tags |
-| :------- | :------- | :------- |
-| accesskeys | Ensures every accesskey attribute value is unique | wcag2a, wcag211 |
-| area-alt | Ensures &lt;area&gt; elements of image maps have alternate text | wcag2a, wcag111, section508, section508.22.a |
-| aria-allowed-attr | Ensures ARIA attributes are allowed for an element&apos;s role | wcag2a, wcag411, wcag412 |
-| aria-required-attr | Ensures elements with ARIA roles have all required ARIA attributes | wcag2a, wcag411, wcag412 |
-| aria-required-children | Ensures elements with an ARIA role that require child roles contain them | wcag2a, wcag131 |
-| aria-required-parent | Ensures elements with an ARIA role that require parent roles are contained by them | wcag2a, wcag131 |
-| aria-roles | Ensures all elements with a role attribute use a valid value | wcag2a, wcag131, wcag411, wcag412 |
-| aria-valid-attr-value | Ensures all ARIA attributes have valid values | wcag2a, wcag131, wcag411, wcag412 |
-| aria-valid-attr | Ensures attributes that begin with aria- are valid ARIA attributes | wcag2a, wcag411 |
-| audio-caption | Ensures &lt;audio&gt; elements have captions | wcag2a, wcag122, section508, section508.22.a |
-| blink | Ensures &lt;blink&gt; elements are not used | wcag2a, wcag222, section508, section508.22.j |
-| button-name | Ensures buttons have discernible text | wcag2a, wcag412, section508, section508.22.a |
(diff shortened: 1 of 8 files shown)


badtant:
It says "Contributor License Agreement is not signed yet" but when I click on Details it says "You have signed the CLA for dequelabs/axe-core" ??

dylanb:
@badtant despite what @marcysutton said in the issue referenced, I don't think this rule meets the criteria for a best practice axe-core rule.

In order for a rule to be included it must never flag things that are not problems. In this case many uses of an anchor with just a # as an href (and the accompanying JavaScript) will not be problems.

The difference between a best practice and a violation is that best practices are issues that can be found with 100% certainty but where the application would still be considered compliant with the issue (WCAG or Section 508). An example is: WCAG requires one mechanism to bypass repetitive navigation. If we don't find a mechanism at all that is a strict WCAG violation. If we find one mechanism but not all the mechanisms (e.g. a skip link, headings and landmark regions) then we flag this as a best practice.

Does that make sense?

badtant:
Hmm, ok. Don't quite understand...
If someone uses a link with only # as href. That link wont lead anywhere... so it shouldn't be a link? Therefor best practise should be to use a button instead so that the screen reader users don't get confused by being told it's a link. In my opinion that is a problem.

Or is there some case that I'm missing?

dylanb:
It depends on what is happening. If the anchor is acting on something e.g. deleting an item, popping up a dialog etc. then it should definitely be a button. If however it is navigating to some other portion of the app/page then it is ok for it to be a link. We cannot tell, so we there would be a high percentage of false positives

badtant:
But in those cases, shouldn't there be a url or hash+id as href and not just a hash?

dylanb:
If the app is written using a router, then it will have proper hashes, if not, then it depends on how its implemented. Take a look at jQuery mobile

http://demos.jquerymobile.com/1.3.0/docs/widgets/navbar/

It uses a LOT of anchors with # in its examples and anyone using that framework would get tons of false positives. This is just one example.

badtant:
And in my opinion that is an accessibility issue. Just because a framework promotes some code it doesn't mean that it's accessible. And also, the rule in the PR is disabled by default so only advanced users that put effort to configuration is affected.

Ringen:
I can see the problem with false positives but, as @badtant mentioned, if the rule is disabled by default and you have to manually enable it I don't think it's an issue.

dylanb:
@Ringen yes that is true, however it does go against our manifesto

badtant:
For example it is highlighted as a bad example here:
http://webaim.org/techniques/hypertext/

So I would argue that it isn't a false positive, just bad code :)

dylanb:
@badtant that is a bad example because of its use of `onmouseover` not because of the use of the `#` in the `href`

badtant:
"One of the most serious barriers is to create links that go nowhere."

"The link in this example goes nowhere. Its only purpose is to activate a JavaScript function."

I stil think it's best practice to use buttons instead. And was hopeing that Axe would promote that in order to get libraries such as jquery mobile (mentioned above) aware of accessibility issues that they are promoting. 

But I'm gonna stop argue now :)

dylanb:
ok, I think we could create a useful best practice rule if we clarify what we are pointing out a bit more precisely.

The best practice that is 100% never a false positive is that anchor links should be fragments or well formed URIs. I think this is what you are trying to achieve - correct?

If that is the case, the messages should be clearer and we should probably be checking for well-formed URIs when the anchor is not a fragment.

I will put message suggestions as comments in the diff

doc/rule-descriptions.md
```diff
@@ -22,6 +22,7 @@
 | empty-heading | Ensures headings have discernible text | best-practice |
 | frame-title | Ensures &lt;iframe&gt; and &lt;frame&gt; elements contain a unique and non-empty title attribute | wcag2a, wcag241, section508, section508.22.i |
 | heading-order | Ensures the order of headings is semantically correct | best-practice |
+| href-no-hash | Ensures an anchors href value is not just # | best-practice |
```
dylanb: How about "Ensures that href values are valid link references to promote only using anchors as links"

lib/rules/href-no-hash.json
```diff
@@ -0,0 +1,17 @@
+{
+  "id": "href-no-hash",
+  "selector": "a[href]",
+  "enabled": false,
+  "tags": [
+    "best-practice"
+  ],
+  "metadata": {
+    "description": "Ensures an anchors href value is not just #",
+    "help": "Anchors with href value that is just # should probably be a button"
```
dylanb: How about the help reading: "Anchors must only be used as links and must therefore have an href value that is a valid reference"

badtant:
Yes, that is what I'm trying to achieve :)
I've updated the texts.

dylanb:
Messages are good.

What about checking for well formed URIs as well?

badtant:
That's going to be quite tricky. Maybe should look for some third party code for that?

dylanb:
A fully generalized implementation is quite complex but I think we might be able to do a simple implementation using regexes that would cover most of what people would want to put into an anchor (for example, we probably don't need to handle `username:password@`)

If you accept the license again, I'll merge this PR, we can open another issue to improve it to do URI checking.

badtant:
When I click the details link I get the message "You have signed the CLA for dequelabs/axe-core" so don't know what more to sign.

dylanb:
click the details link next to the error above. It should take you to the CLA.

badtant:
When I click it it says "You have signed the CLA for dequelabs/axe-core"

badtant:
Finally got rid of the message :)

dylanb:
one more thing, could you add a test with an anchor that has no href at all?

badtant:
I've added a test for an empty href. Since the selector only looks for anchors with a href attribute there is no need for tests with anchors without href.

I've also updated rules-description.md with a column that shows which rules are enabled/disabled by default.

badtant:
Hmm, something happened with the Node 4 tests. Can it be run again?

Issue #183, opened 2016-03-27
Create Typescript definition file for Angular 2 testing


marcysutton:
To support Angular 2 development with axe-core, we should create a Typescript definition file for type annotations and hints. I still have to look into specifics, but I wanted to capture this now. Here is some documentation on creating a type definition file: 
- http://definitelytyped.org/guides/creating.html
- http://peter.grman.at/how-to-write-typescript-definition-files/

marcysutton:
Things I've found so far that we may want to address:

- We specify optional parameters before required ones in `axe.a11yCheck` and `axe._runRules`, which isn't allowed in Typescript. ( i.e. `context: any, options?:Object, callback:Function`).
- There are multiple methods and objects exposed in the global namespace that we may not want to be called directly. (i.e. `_audit`, `_load()`, `_runRules()`, `commons`, `constants`, `utils`, etc.)
- Arguments aren't always consistent, i.e. `callback` vs. `cb` in `axe.a11yCheck` and `axe.addReporter`, respectively.
- We could really use additional code comments and documentation for the newer parts of the API, such as the Plugin system and `axe.configure`.

I'm finding the exercise of creating a Typescript definition file to be super valuable, as it maps our API to interfaces and types. aXe is pretty loose with types in many places, taking full advantage of JavaScript's dynamic nature but making it more challenging to define the API in a definition file. We can work around this with the `any` type, but it seems like a good opportunity to look more closely at our code structure and the public API.

dylanb:
Any method starting with _ is not public for sure

commons and utils are intended to be used in the rules, so they are public but we may need to curate the list to make sure we are not exposing too much.

We should not change existing methods to preserve backwards compatibility. But we already plan to deprecate a11yCheck in favor of a run method that has a better signature.

Can you simply always call the callback the same thing in the TypeScript definition regardless of what it is called in our code? If not, then TypeScript is quite seriously flawed.

DanielRosenwasser:
Hey @marcysutton I saw [this tweet](https://twitter.com/marcysutton/status/755172506089066496). If you need, I'd be happy to help. Is this issue a good place for discussion, or is something like Gitter better?

marcysutton:
Hi @DanielRosenwasser! Either works. This is what I've got so far, I know there are a few kinks to be ironed out with enums and such but I couldn't even get it to work with a simple exported function. https://gist.github.com/marcysutton/a3e84fa229a90258fb6d86171b840d86

[...]

PR #224, opened 2016-09-07
feat: add typescript definition file

marcysutton:
Closes https://github.com/dequelabs/axe-core/issues/183

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -102,11 +102,12 @@ grunt build
 
 The TypeScript definition file for axe-core can be found in [typings/axe-core](./typings/axe-core).
 
-After running `npm install` to install TypeScript, you can run TS definition tests (with the optional `noImplicitAny` flag): 
+To develop with TypeScript you must first install it (globally recommended):
+```
+sudo npm -g install typescript
+```
 
+Once that's installed, you can run TypeScript definition tests (with the optional `--noImplicitAny` flag):
 ```
 tsc --noImplicitAny typings/axe-core/axe-core-tests.ts
 ```
-
-Coming soon: Angular 2 development tips.
-
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -97,3 +97,16 @@ To build the package:
 ```
 grunt build
 ```
+
+## Using aXe with TypeScript
+
+The TypeScript definition file for axe-core can be found in [typings/axe-core](./typings/axe-core).
+
+After running `npm install` to install TypeScript, you can run TS definition tests (with the optional `noImplicitAny` flag): 
+
+```
+tsc --noImplicitAny typings/axe-core/axe-core-tests.ts
+```
+
+Coming soon: Angular 2 development tips.
+
--- a/typings/axe-core/axe-core-tests.ts
+++ b/typings/axe-core/axe-core-tests.ts
@@ -5,6 +5,8 @@ var $fixture:any = {}
 
 // axe.a11yCheck config
 axe.a11yCheck(context, {}, (results) => {
+	// axe's results object
+	console.log(results.passes.length)
 	console.log(results.violations.length)
 });
 // axe.a11yCheck include/exclude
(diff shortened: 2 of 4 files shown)


typings/axe-core/axe-core-tests.ts
```diff
@@ -0,0 +1,87 @@
+/// <reference path="axe-core.d.ts" />
+
+var context:Object = document
```
DanielRosenwasser: I think you want `: any`

typings/axe-core/axe-core.d.ts
```diff
@@ -0,0 +1,152 @@
...
+	timestamp: string,
+	passes: Pass[],
+	violations: Violation[]
+}
+interface Pass {
+	description: string,
+	help: string,
+	helpUrl: string,
+	id: string,
+	impact: string, // "minor" | "moderate" | "serious" | "critical" | "null"
```
DanielRosenwasser: You should use the actual string literal values here

DanielRosenwasser: In other words, `impact: "minor | "moderate" | "serious" | "critical" | "null"`

Also, did you really mean `"null"` or `null`?

typings/axe-core/axe-core.d.ts
```diff
@@ -0,0 +1,152 @@
...
+	tags: string[], // "wcag2a" | "wcag2aa" | "section508" | "best-practice"
```
DanielRosenwasser: Probably should write `("wcag2a" | "wcag2aa" | "section508" | "best-practice")[]` here too

typings/axe-core/axe-core.d.ts
```diff
@@ -0,0 +1,152 @@
...
+
+	/**
+	 * Starts analysis on the current document and its subframes
+	 *
+	 * @param  {Object}   context  The `Context` specification object @see Context
+	 * @param  {Array}    options  Options passed into rules or checks, temporarily modifyint them.
+	 * @param  {Function} callback The function to invoke when analysis is complete.
+	 * @returns {Object}  results  The aXe results object
+	 */
+	a11yCheck(context: ElementContext, options: {runOnly?: RunOnly, rules?: Object}, callback: (results:any) => void): AxeResults
```
DanielRosenwasser: Is there a better type than `any` for the result parameter?

CONTRIBUTING.md
```diff
@@ -97,3 +97,16 @@ To build the package:
...
 ```
+
+## Using aXe with TypeScript
+
+The TypeScript definition file for axe-core can be found in [typings/axe-core](./typings/axe-core).
+
+After running `npm install` to install Typescript, you can run TS definition tests (with the optional `noImplicitAny` flag): 
+
+```
+ntsc --noImplicitAny typings/axe-core/axe-core-tests.ts
```
DanielRosenwasser: Do you mean `tsc`, or did you really mean `ntsc` for `ntypescript`?

CONTRIBUTING.md
```diff
@@ -97,3 +97,16 @@ To build the package:
 ```
 grunt build
 ```
+
+## Using aXe with TypeScript
+
+The TypeScript definition file for axe-core can be found in [typings/axe-core](./typings/axe-core).
+
+After running `npm install` to install Typescript, you can run TS definition tests (with the optional `noImplicitAny` flag):
```
DanielRosenwasser: nit: Typescript -> TypeScript

DanielRosenwasser:
Looks good apart from my nits!

One thing I'll note is that TypeScript 2.0 is coming out pretty soon. If you want to support the newer-style declaration format for UMD libraries (libraries that act as both globals and modules) at that time, let me know and I'd be happy to help with that.

typings/axe-core/axe-core.d.ts
marcysutton: Yep, good catch. I want to abstract these strings into a common place since they're used multiple times – same deal with the tag values above. Any tips on how to do that without depending on Typescript 2.0? I'd like for it to be usable in either version if possible...

typings/axe-core/axe-core.d.ts
marcysutton: It should probably be `AxeResults`, since it's a vehicle for the return object.

CONTRIBUTING.md
marcysutton: It should be `tsc`! I think I got confused with the many rewrites and library deprecations.

typings/axe-core/axe-core.d.ts
DanielRosenwasser: If you use the same strings, you can use a type alias like this:

```ts
export type ImpactValues = "minor | "moderate" | "serious" | "critical" | "null";
```

or something similar. String literal types've been around since 1.8, so you can feel free to use them here too.

typings/axe-core/axe-core.d.ts
DanielRosenwasser: Here you could write

```ts
export type TagValue = "wcag2a" | "wcag2aa" | "section508" | "best-practice";
```

and then write `tags: TagValue[]`

typings/axe-core/axe-core.d.ts
marcysutton: Thanks! Does it make things more complicated with `null` versus `"null"`?

marcysutton:
Ok, for some reason my module definition is now busted. I can't figure out what I changed that caused it to stop working. @DanielRosenwasser, any tips on exporting the module name? I saw this same issue happened with Observable and Visual Studio, but I still don't know how to solve it. I'm running TypeScript version 1.8.10.

```
typings/axe-core/axe-core-tests.ts(89,1): error TS2304: Cannot find name 'axe'.
typings/axe-core/axe-core.d.ts(158,16): error TS2664: Invalid module name in augmentation, module 'axe-core' cannot be found.
```

typings/axe-core/axe-core.d.ts
```diff
@@ -3,19 +3,27 @@
 // Definitions by: Marcy Sutton <https://github.com/marcysutton>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
+export type ImpactValue = "minor" | "moderate" | "serious" | "critical" | "null";
```
marcysutton: "null" should not be a string here, but TypeScript complains about mixed types.

DanielRosenwasser: Gotcha. Maybe when TS 2.0 comes out it would be a good idea to create a 1.8 `.d.ts` file, and make this friendlier to people using 2.0 with `--strictNullChecks`.

For now you can omit the `null` and the only people who might have a problem are those using 2.0 with `--strictNullChecks`.

PR #227, opened 2016-09-12
TypeScript namespace restructuring

DanielRosenwasser:
This PR moves the `.d.ts` file to use namespaces. The reason that you saw some issues with not being able to import/find certain things correctly is because as soon as you have a top-level `export` or `import`, the file becomes a module.

Once you have a module, your variables are no longer available in the global scope. So this moves all of your declarations into a namespace (which actually reflects the way that Axe itself was written!), and then use `export =` on that namespace to say that the module version has the same shape as the namespace.

I also fixed up the tests to use literal types. The problem is that strings only get literal types if they're getting assigned to something that's known to have a literal type. We're working on making this easier to use, but until then this will do it.


marcysutton:
Thank you so much! I've incorporated the changes in #224.

dylanb:
@DanielRosenwasser Thank you very much for this great contribution!!

PR #233, opened 2016-09-27
Angular ng click rule

marcysutton:
This rule checks click handlers in both Angular 1 and 2 to ensure they are bound to interactive elements, a big use case I tried (and failed) to cover with ngAria. 

It's disabled by default to only target Angular developers–but it made a lot of sense to use our infrastructure for test coverage and other check reuse, such as `focusable-no-name`. 

I added support for nested click events in case the developer has bound a click event to a DIV for a larger hit area but bound the same event handler to an accessible child element–that should not raise a violation.

### DIFF
--- a/doc/rule-descriptions.md
+++ b/doc/rule-descriptions.md
@@ -1,6 +1,7 @@
 | Rule ID | Description | Tags | Enabled by default |
 | :------- | :------- | :------- | :------- |
 | accesskeys | Ensures every accesskey attribute value is unique | wcag2a, wcag211 | true |
+| angular-ng-click | Ensures ng-click is used on accessible elements | wcag2aa, wcag21, wcag412 | false |
 | area-alt | Ensures &lt;area&gt; elements of image maps have alternate text | wcag2a, wcag111, section508, section508.22.a | true |
 | aria-allowed-attr | Ensures ARIA attributes are allowed for an element&apos;s role | wcag2a, wcag411, wcag412 | true |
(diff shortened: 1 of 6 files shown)


dylanb:
@marcysutton in principle, I think the rule is good. In practice, I would like to know if this will raise false positives. Could you run this against some of the Angular example applications and see whether you get any false positives? I want to know whether the false poitives we find (if we find any) would be considered bad Angular coding practices. Can you report your findings back here?

marcysutton:
I ran into a pretty big issue dogfooding this rule in axe-chrome for easy checks: `[ng-click]` and `[(click)]` are not valid DOM selectors. It may not be an issue with automated tools, but it's definitely an issue in the browser. I'll do some more testing on the automated side to see if it falls down in the same way–but it works in our test suite.

That said, I wrote this with years of Angular experience in mind. That's what the nested check is all about–looking to see if the same event callback function can be completed with the keyboard somewhere else in the DOM. That would make it so a high-level click event wouldn't be a violation if the same function could be done elsewhere with the keyboard.

dylanb:
`[ng-click]` is a valid selector but won't it (and `(click)` in Angular 2) be replaced in the rendered DOM?

Certainly when used inside repeaters (ng-repeat or ngFor) they do not end up in the DOM at all. So that raises the question of how valuable is this rule?

marcysutton:
That's a good point–I assumed they would remain in the DOM. It might need to be a part of a build step rather than looking at the compiled DOM. For click events, that seems okay as opposed to color contrast rules, which need styles to be applied. I'll think on that some more and close this PR for now.

PR #247, opened 2016-11-02
chore(deps): added grunt-cli as devDependencies

sarbbottam:
so that global grunt-cli installation is not required

---
`npm run build` expected to `grunt` in the `PATH` and failed as there was none. Installing grunt as a devDependencies solves the problem 

>In addition to the shell's pre-existing `PATH`, `npm run` adds `node_modules/.bin` to the `PATH` provided to scripts. Any binaries provided by `locally-installed` dependencies can be used without the `node_modules/.bin` prefix - [npm/run-script](https://docs.npmjs.com/cli/run-script#description)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,12 +3,10 @@ node_js:
   - '4'
   - '5'
 before_install:
-  - npm install -g grunt-cli
-  - npm install
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
 script:
-  - grunt test
+  - npm test
 env:
   global:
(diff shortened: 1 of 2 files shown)


sarbbottam:
@marcysutton hopefully the tiniest PR.

marcysutton:
The reason it isn't in the dependencies is Grunt recommends installing it globally in their own docs, so many developers have it installed globally: http://gruntjs.com/getting-started

Alternatively, we could call it out more clearly as a requirement in [contributing.md](https://github.com/dequelabs/axe-core/blob/master/CONTRIBUTING.md).

sarbbottam:
My 💵 0.02

>Grunt recommends installing it globally in their own docs, so many developers have it installed globally: http://gruntjs.com/getting-started

I don't know why Grunt still recommends that. Dev environment should be sandboxed/isolated as best as possible. In 2013/14, I installed `grunt-cli` globally too, but not evar again.

>we could call it out more clearly as a requirement in contributing.md.

Current workflow is abstracted via `npm script` and locally installed `grunt-cli` suffices the need.

If the process can be automated it should be, I guess. Sandbox everything, why to use something globally if that can be installed locally.

Instead of asking contributor to 

I believe

```
npm install 
```
is a better workflow than

```
npm install -g grunt-cli
npm install 
```
---
This PR doesn't:
- prevent anyone to install `grunt` globally
- add any extra abstraction 
- change the workflow
- affect the output
- affect the build performance 

_It just enhances workflow._

.travis.yml
```diff
@@ -3,12 +3,10 @@ node_js:
   - '4'
   - '5'
 before_install:
-  - npm install -g grunt-cli
-  - npm install
```
sarbbottam: travis runs `npm install` by default.

.travis.yml
```diff
@@ -3,12 +3,10 @@ node_js:
   - '4'
   - '5'
 before_install:
-  - npm install -g grunt-cli
```
sarbbottam: not required, as `grunt-cli` is part of devDependencies.

.travis.yml
```diff
@@ -3,12 +3,10 @@ node_js:
...
-  - npm install
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
 script:
-  - grunt test
+  - npm test
```
sarbbottam: use `npm` workflow as it is already defined in `package.json`

marcysutton:
Works for me!

sarbbottam:
Awesome! looking forward to it :)

dylanb:
Last I checked, you have to install grunt-cli globally in order to be able to type `grunt` at the command line.

If this is the case, then we need to add the line back suggesting that users do that.

If you think this is not the case, I want proof that local installation of grunt-cli will work on all platforms before I think this can be merged.

sarbbottam:
>I want proof that local installation of grunt-cli will work on all platforms before I think this can be merged.

It works on *nix system as the `travis` build has passed. If [appveyor](https://www.appveyor.com/) build was setup for windows, it should have verified that. Alternatively, you can clone this PR in a windows machine and verify it.

>Last I checked, you have to install grunt-cli globally in order to be able to type grunt at the command line.

This PR does not change that workflow, but enhances for the folks who does not want to have a global grunt-cli installation.

>This PR doesn't:
- prevent anyone to install `grunt` globally
- add any extra abstraction 
- change the workflow
- affect the output
- affect the build performance 

>_It just enhances workflow._

I am not sure, what holds you back from merging this PR?

.travis.yml
dylanb: Can you undo this change, then we can merge

dylanb:
@sarbbottam - merged - thanks for the contribution and apologies for the confusion. It has been a crazy week with the election and I was a bit frazzled :-)

Issue #252, opened 2016-11-10
Add an implicit role of "row" to the tr element


SiTaggart:
According to the [ARIA in HTML spec](https://www.w3.org/TR/html-aria/#tr), table rows come with an implicit role of "row".

Currently aXe causes false negatives on rows within an ARIA grid, that uses HTML table semantics. You shouldn't have to apply `<tr role="row">` to a table row element.

Yes, I agree it's strange that information is not supplied in the HTML 5 specification, so for good measure, I [filed an issue on the ARIA in HTML spec](https://github.com/w3c/html-aria/issues/47) too. 

I'm going to assume the implicit role is correct, but if it turns out it's a mistake in the ARIA in HTML spec then this issue can be closed.

dylanb:
@SiTaggart can you supply a test HTML snippet for this issue?

SiTaggart:
Is this useful @dylanb https://github.com/dequelabs/axe-core/pull/255 ?

SiTaggart:
I should click on things before I submit a PR. I see it's covered in #253. I guess my test my be of some worth. Feel free to close otherwise.

Thanks for adding these so quickly! 👍

dylanb:
Thanks, that test was useful

marcysutton:
Is this still an issue, or did it get handled with #253?

SiTaggart:
It's totally fixed as far as I can see in the latest release @marcysutton.

Thanks for being great at addressing these things! 🎉

marcysutton:
That one was all @dylanb! I will go ahead and close this issue. Thanks for contributing!

SiTaggart:
Well, just generally, as a team 👏

PR #253, opened 2016-11-12
add all missing implicit roles from HTML 5

dylanb:
Don't merge yet, need to add tests


dylanb:
closes https://github.com/dequelabs/axe-core/issues/252

dylanb:
@marcysutton @WilcoFiers this is now ready for review/merge

lib/commons/aria/index.js
```diff
@@ -473,7 +483,8 @@ lookupTables.role = {
 			one: ['menuitem', 'menuitemradio', 'menuitemcheckbox']
 		},
 		nameFrom: ['author'],
-		context: null
+		context: null,
+		implicit: ['menu[type="context"]']
```
marcysutton: Huh, I didn't know you could specify implicit ARIA roles that way (with an attribute selector).

test/checks/aria/allowed-attr.js
```diff
@@ -57,7 +57,7 @@ describe('aria-allowed-attr', function () {
 	});
 
 	it('should return true if there is no role', function () {
-		var node = document.createElement('div');
+		var node = document.createElement('picture');
```
marcysutton: After following https://github.com/w3c/html-aria/issues/52, it looks like `div` has no corresponding role (where I assumed it was `group`). Is that related?

test/mock/frames/recursive.html
```diff
@@ -21,6 +21,6 @@
 </head>
 <body>
 	<h1>Recursive iframe</h1>
-	<iframe src="recursive.html">
+	<iframe src="nested1.html">
```
marcysutton: What is the nature of this change? It would help to know as I'm learning more about recursive iframes.

dylanb: `recursive.html` was infinitely including iframes. The infinite recursion was not required for that test to do what it was doing, so I changed it to a single iframe.

PR #255, opened 2016-11-14
Implicit tr role

SiTaggart:
2 things contained in this PR:

- Firstly a test to assert that using a child with an implicit role, should pass the required child test.
  - This was a way to see a test fail when using `grid` and `tr` with no `role="row"`
- Add the `tr` element to the `row` role, as having an implicit role
  - Newly added test, now passes (yay!)


dylanb:
Is included in #253, closing

PR #256, opened 2016-11-16
Avoid calling `DqElement` unnecessarily when gathering results

christoffer:
This diff will modify Rule#run so that the `DqElement` wrapping of a node when gathering rule check results is deferred until it's been established that the result will actually be used.

This seems to have some performance impact since (at least in my own tests) the vast majority of runs ends up with a non empty aggregate result set, but with only empty results. This situation leads to a lot of work being spent finding the selector of a node unnecessarily.

### DIFF
--- a/test/core/base/rule.js
+++ b/test/core/base/rule.js
@@ -313,7 +313,95 @@ describe('Rule', function() {
 				}, {}, function(r) {
 					assert.lengthOf(r.nodes, 0);
 				}, isNotCalled);
+			});
 
+			describe('DqElement', function() {
+				var origDqElement;
+				var isDqElementCalled;
+
+				beforeEach(function() {
+					isDqElementCalled = false;
+					origDqElement = axe.utils.DqElement;
+					axe.utils.DqElement = function() {
+						isDqElementCalled = true;
+					};
+					fixture.innerHTML = '<blink>Hi</blink>';
+				});
+
+				afterEach(function() {
+					axe.utils.DqElement = origDqElement;
+				});
+
+				it('is created for matching nodes', function(done) {
+					var rule = new Rule({
+						all: ['cats']
+					}, {
+						checks: {
+							cats: new Check({
+								id: 'cats',
+								enabled: true,
+								evaluate: function() {
+									return true;
+								},
+								matches: function() {
+									return true;
+								}
+							})
+						}
+					});
+					rule.run({
+						include: [fixture]
+					}, {}, function() {
+						assert.isTrue(isDqElementCalled);
+						done();
+					}, isNotCalled);
+				});
+
+				it('is not created for disabled checks', function(done) {
+					var rule = new Rule({
+						all: ['cats']
+					}, {
+						checks: {
+							cats: new Check({
+								id: 'cats',
+								enabled: false,
+								evaluate: function() {},
+								matches: function() {
+									return true;
+								}
+							})
+						}
+					});
+					rule.run({
+						include: [fixture]
+					}, {}, function() {
+						assert.isFalse(isDqElementCalled);
+						done();
+					}, isNotCalled);
+				});
+
+				it('is not created for un-matching nodes', function(done) {
+					var rule = new Rule({
+						all: ['cats']
+					}, {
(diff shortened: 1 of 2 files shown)


dylanb:
@christoffer Thank you for the submission. Performance improvements are always very welcome :-)

Just to be sure, what your change is doing is to ensure that when ALL the checks for a rule return empty results due to being disabled or not matching the node, then this change will save time by not instantiating a DqElement instance?

Can you add a test that shows that the DqElement constructor is in fact not being called in that scenario (please test both the disabled and the non-matching cases) so we can avoid a regression?

christoffer:
> Just to be sure, what your change is doing is to ensure that when ALL the checks for a rule return empty results due to being disabled or not matching the node, then this change will save time by not instantiating a DqElement instance?

Correct. I'm not entirely sure what actually causes the results to be empty but your suggestion seems reasonable :)

> Can you add a test that shows that the DqElement constructor is in fact not being called in that scenario (please test both the disabled and the non-matching cases) so we can avoid a regression?

Great suggestion.

dylanb:
@christoffer took me a bit to remember what could cause this too

christoffer:
@dylanb Sorry for the late reply on this (holidays and seasonal sniffles and whatnot :P). 

I went a little outside my own style (which tends to move as much stuff as possible out of the `it` block) in an effort to fit the explicit style for the rest of the tests.

Please let me know if I should take another pass at it, or if there's anything wrong with the test approach itself. I verified that the last two cases fails with `rule.js` from `master` (and obviously that they pass with the new version).

test/core/base/rule.js
```diff
@@ -313,7 +313,95 @@ describe('Rule', function() {
...
+						done();
+					}, isNotCalled);
+				});
 
+				it('is not created for un-matching nodes', function(done) {
+					var rule = new Rule({
+						all: ['cats']
+					}, {
+						checks: {
+							cats: new Check({
```
dylanb: This line (and others) are failing the linter because `Check` is not defined. Go to line 1 in this file and add `Check` to the global defines to solve this problem

dylanb:
@christoffer Thanks! The tests look good. There are some linting problems (see comment) and you also need to accept the CLA

test/core/base/rule.js
christoffer: Ah, sorry -- silly mistake. Thanks for pointing it out.

christoffer:
> and you also need to accept the CLA

Yeah, not sure what's going on there. I clicked the link the first thing I did after creating the PR and saw: "You have signed the CLA for dequelabs/axe-core", but the badge never turned green. Figured it was just a delay, but today when I tried it again it showed me the unsigned screen. And after accepting (again) it still seems like there's some issue with it because it's still not green even though I've accepted it.

cla

Anyone got any experience with this happening before?

dylanb:
@christoffer go to https://cla-assistant.io/ and sign in with Github first, then go back to the signing page. That should sort things out.

christoffer:
@dylanb Unfortunately it seems like there's something broken with the CLA integration. I tried visiting the landing page (the one you linked) and then visit the signing page again, but with the same result. I also tried clearing cookies for both Github + cla-assistant.io and re-doing the process, and for good measure I also tried in an incognito window + unauthing the application and the re-authing it.

But I still get the same result: the signing page shows me that I've already signed, but the badge here is still yellow.

Any advice on how to proceed? I'm super happy to have this contribution applied unattributed, say as a regular commit rather than a PR -- if that's an option.

dylanb:
@christoffer I am not sure why its doing that.

If you could just comment with "I agree to the CLA" that should suffice.

christoffer:
I agree to the CLA

christoffer-dropbox:
@dylanb Sorry this was my mistake, but as the commenter on my Issue filed with CLA mentioned, I had made commits with another Github user than the user I filed the PR from -- and the CLA needs to be signed with the commiter (which makes sense).

Sorry for the confusion.

dylanb:
@christoffer-dropbox thank you for the contribution!!

PR #257, opened 2016-11-16
Include all language tags, not just two letter tags

mfairchild365:
While testing several sites, I came across one about the Omaha language. The site used the 'oma' language tag to mark up Omaha language on the site, which appears to conform to BCP47 and is listed in the IANA registry. However, aXe-core was flagging the language as an error because it is only pulling in two character language codes from the registry. I may be missing something, but I don't see anything in the HTML or WCAG standards that limit usage to 2 character codes.


mfairchild365:
Any thoughts on this?

dylanb:
@mfairchild365 where did you find the language code in the registry? Can you post a link here?

mfairchild365:
Sure, http://www.iana.org/assignments/language-subtag-registry/language-subtag-registry and this is the relevant record:

```
Type: language
Subtag: oma
Description: Omaha-Ponca
Added: 2009-07-29
```

mfairchild365:
@dylanb I added an integration test to cover this case too.

I should mention that I seriously doubt any AT would know how to handle the oma language, but that doesn't change that it is still a valid language code.

dylanb:
@mfairchild365 you will need to run the language generation task (it does not run every time). This will update the `valid-lang.json` file and then you need to submit that file as part of your PR. I would like to see how much bigger that makes our final code before deciding on whether to approve this change.

mfairchild365:
@dylanb already done

dylanb:
@mfairchild365 ah, I see, I missed it because the diff was not showing as it was too big - useful!

dylanb:
Wow, this is increasing the minified file size by 62% from ~163K to 264K. The API already has a way to add/change languages. I would prefer to add a capability to the UI of the extension to allow specific languages to be added rather than bloat the library for languages that are not supported by any ATs.

@WilcoFiers @marcysutton WDYT?

marcysutton:
I would also prefer to not add much to the filesize if we can enable language support through the API and an affordance in the extension UI–especially if the overhead doesn't gain much for AT. The bulk of the content added is in `valid-lang.json`–how do we decide which languages to include? The more common ones?

mfairchild365:
While I understand where you are coming from, I disagree. In my opinion (and I realize that I may not share the same opinions as the project maintainers), I see the file size of the library as not that important when compared to the items listed in the project manifesto, specifically a zero false positive test rate.

If the majority of users are using this library as a browser extension, they only need to download the library once (or whenever it is updated), so why not just include every valid language? The scenario of automated testing with the library (integration or a crawling method), might benefit more from a smaller library, but again, I don't think file size should get in the way of the core objectives of the library (but a small file size should a goal).

boris1558:
I wonder if loading a common language set and then doing a query to http://www.iana.org/assignments/language-subtag-registry/language-subtag-registry would accomplish both?

mfairchild365:
@boris1558 performing a query to the registry would add an external dependency, which is an extra point of failure and might not even be possible in closed environments. However, it is a possibility.

@dylanb I had a think about this some more and ended up doing some more research. What do you think about testing the structure of the language tag instead of testing the value against a predefined set? This could be accomplished through something like https://www.npmjs.com/package/bcp-47

Advantages to testing the structure:

- Much smaller file size impact
- Will not fail new languages that have been added to the registry since we last updated our code
- Can validate the entire language tag (language, scirpt, region, variants, etc). Although I'm not sure how much of an impact this would have on AT. Would AT toss out the entire language tag if the script part was in an invalid format?

Disadvantages to testing the structure

- Possibility of false negatives by not detecting typos (typed 'em' instead of 'en') 

I also did some research regarding how other tools handle this.

- The i18n JavaScript API appears to test the structure: http://ecma-international.org/ecma-402/1.0/#sec-6.2.2
- HTML_Codesniffer tests the structure: https://github.com/squizlabs/HTML_CodeSniffer/blob/9d6832f4f332f21c2b63899f4e9a804a381237e2/Standards/WCAG2AAA/Sniffs/Principle3/Guideline3_1/3_1_1.js#L70
- It looks like the w3c validator checks both structure and value: https://github.com/validator/validator/blob/fe88da45fd505a4add9a61d1e15c3a57cd8b30fe/src/nu/validator/datatype/Language.java#L109

dylanb:
@mfairchild365 I agree - external call is not the way we want to go for the reasons you mention PLUS it violates our manifesto in that it would introduce lag and failure modes we do not want to introduce.

@marcysutton given that our library is NOT really a run-time library for most use cases, I think I am leaning towards suggesting that we merge this change and if we get compelling evidence that it is causing problems, consider actions based on the details of those use cases. Before we do that, I would like to test to see what the actual performance impact is when the large library is included into our extension(s). Could you do that test (a before and after test)?

marcysutton:
I do have benchmarking on my plate as part of our performance improvements. But we have to weigh priority against the ES6 architecture spike I was about to start. Maybe I should try to finish the perf improvements before moving on to architecture?

dylanb:
yes, that sounds like the right priorities

mfairchild365:
Any word on this and when it might get merged?

dylanb:
@mfairchild365 still waiting for the analysis on performance from @marcysutton

marcysutton:
I ran 2 axe audits with our performance timer on different sites with and without this PR, and it was a very small increase. The filesize is honestly a bigger issue than what I found. If you have another page you want me to test, let me know. I'm still working on automating this process so we can have it run against PRs automatically.

## Mars Commuter page 
https://dequeuniversity.com/demo/mars/

**Before: all 4 frames**
Measure axe took 788.2050000000008ms

**After: all 4 frames**
Measure axe took 802.3150000000005ms

## MDN Promises page
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise

**Before: 2 frames**
Measure axe took 963.2800000000007ms

**After: 2 frames**
Measure axe took 996.3150000000005ms

marcysutton:
@dylanb wanted more data on the effect on injecting a bigger file into axe-chrome, so I ran some performance measurements on that code:

## Mars Commuter page
https://dequeuniversity.com/demo/mars/

```
PerformanceMeasure
duration: 155.60499999999956
entryType: "measure"
name: "measure-startup"

mfairchild365:
Interesting. I'm guessing the difference in start time is due to the larger file being loaded. I'm unsure of what exactly 'startTime' refers to. Is that the time is takes to start the tests after the 'Analyze' button is pressed?

marcysutton:
I marked when a call is made to initialize frames in axe-chrome and again in that initializer's callback function. Then I measured those two marks. The measure has a start time and a duration, and the former definitely did look impacted by the file size.

From the User Timing API: `The startTime attribute will return a DOMHighResTimeStamp with the measure's start mark ` - https://www.w3.org/TR/user-timing/

mfairchild365:
Please let me know if you would like me to make any changes :)

marcysutton:
We decided that the duration difference was minimal enough that it was worth the change, despite the filesize increase. And startTime is irrelevant. 

Thanks so much for your contribution!

PR #265, opened 2016-12-08
Sync with Attest/2.1.6

marcysutton:
This PR updates aXe to be in sync with our code on Bitbucket, which is at version 2.1.6.

Two major conflicts with this merge: 

- https://github.com/dequelabs/axe-core/pull/256
- https://github.com/dequelabs/axe-core/pull/205

The code in those PRs changed significantly on Bitbucket, so I merged them together.


dylanb:
Delete this test, it is not required because of changes we made in 2.1 https://github.com/dequelabs/axe-core/pull/256/files#diff-1fc85dd7e1a36ca051cce9b8cc418453R383

marcysutton:
Thanks, @dylanb. That works.

It looks like CI is having some issues with the Mocha tests. If we can enable logging in the test options on Travis, that will point me in the right direction. 

I'm also having trouble with the integration tests in Safari 10 with Selenium 2.53 and 3.0.1, and in multiple versions of Node. They didn't work for me in Safari 9 either. @WilcoFiers are you able to run them in any version?

dylanb:
just for the record....remote debugging race conditions is a **insert_expletive_of_choice_here**

PR #272, opened 2016-12-21
Fix a bug color.getBackgroundStack, incorrectly checking innerWidth instead of innerHight

mfairchild365:
I'm guessing this should be height, not width. I found this by testing at a mobile width.


mfairchild365:
Well, thats an odd Travis CI error. Any suggestions on how to fix it? Maybe we just need to re-run the tests?

marcysutton:
That's what it looks like. But I haven't encountered a Snyk 500 error on ci before.

dylanb:
Perhaps we can pull the grunt snyk task out of the targets that run on Travis because this is really only required for local development.

@mfairchild365 do you want to make that part of this PR?

mfairchild365:
@dylanb done :)

dylanb:
@mfairchild awesome - thanks for the contribution!!

@marcysutton @WilcoFiers: we don't have a test for this condition yet - should we create one?

mfairchild365:
So, I started to put together a test for this, but found that there are a ton of other tests that fail due to this bug when the browser is at a mobile width. So I wonder if the better solution is to run the test suite at both desktop and mobile widths...

I tried to do implement mobile-width testing via the grunt-mocha plugin which is using phantomjs. However, I have never been able to get `grunt test` to work locally on my Mac OS development machine for this project. I just see something like `Testing: http://localhost:9876/test/commons/` and it hangs indefinitely. I spent an hour trying to debug it, but got nowhere. I even tried to spin up a fresh centos server and ran into a different set of issues. :( Any ideas on how to fix that? Could it be because I'm running node v7.3.0? I think travis CI is only running versions 4 and 5, and the tests are working there...

In the meantime, perhaps someone else can add mobile-width testing to the suite. I was thinking of implementing it as a separate grunt task, something like:

```
'mocha-mobile': testConfig(grunt, {
  reporter: grunt.option('reporter') || 'Spec',
  view: '400x800' //tell phantomjs to use a mobile width
}),
```

Looking over various documentation, I think that should work, but I have no idea if it actually does because I can't get these phantomjs tests to work locally. :(

mfairchild365:
A further update on this. I was able to test my idea by turning on travis CI for my fork of the project and committing to a junk branch to experiment. In short, my above idea didn't work because I apparently have no idea how grunt works. :man_shrugging: 

Another idea that I had was to edit the `build/test/config.js` script to add a 'unit-mobile' block. Something like
```
'unit-mobile': {
  options: {
    log: true,
    urls: [
      'http://' + host + ':<%= connect.test.options.port %>/test/core/',
      'http://' + host + ':<%= connect.test.options.port %>/test/checks/',
      'http://' + host + ':<%= connect.test.options.port %>/test/rule-matches/',
      'http://' + host + ':<%= connect.test.options.port %>/test/commons/',
      'http://' + host + ':<%= connect.test.options.port %>/test/integration/rules/'
    ],
    run: true,
    mocha: {
      grep: grunt.option('grep')
    },
    page: {
      viewportSize: { //tell phantomjs to use a mobile width
        width: 400,
        height: 800
      }
    }
  }
},
```

I think that might have worked, because travis started to mark tests as failing, but then the tests stalled and travis ci times out after 10min. So... I'm not quite sure how to move forward with this.

mfairchild365:
@dylanb Hope you had good holidays :)

Please let me know if there is more that I should do here, or if you would rather tackle mobile width testing internally / another PR.

dylanb:
@WilcoFiers @marcysutton @mfairchild365 can we separate out the test creation into a separate ticket so that we can merge this PR?

marcysutton:
We haven't forgotten about this PR, we're just working towards an internal deadline. Stay tuned!

mfairchild365:
@marcysutton thanks for the update! I'm not in a big rush anymore; I was able to deploy a custom build for my needs. I'd rather not get into a habit of doing that, but I know you are busy.

Also, just an FYI, I don't know if I will be able to put any more time into this PR. I also have some other projects that I am focusing on right now.

marcysutton:
@mfairchild365 no prob, thanks for all your contributions already! We'll get it wrapped up as soon as we can.

marcysutton:
I started looking into this today. Grunt-mocha does not support viewportSize, as per this open issue from 2014: https://github.com/kmiyashiro/grunt-mocha/issues/86 

We'd have to write a custom Grunt task wrapping [mocha-phantomjs](https://github.com/nathanboktae/mocha-phantomjs) or run Webdriver tests instead. Any preference on that @dylanb?

As a side note: @mfairchild365, I had an issue a while back with tests hanging due to a Node.js misconfiguration on my machine, which I resolved by switching to [n](https://github.com/tj/n) instead of using [nvm](https://github.com/creationix/nvm).

dylanb:
@marcysutton I prefer Webdriver

marcysutton:
Going ahead and merging this, we'll track the mobile tests with #277. Thanks for your contribution!

Closes #292.

PR #291, opened 2017-03-05
Return optimally suited implicit role for a node

### LINKED ISSUE #290: Implicit node role ambiguously inferred when multiple options available
Currently, given a node where multiple implicit roles are available, such as a `th` with multiple valid implicit roles ('cell', 'gridcell', 'rowheader', 'columnheader'), the first available implicit role is returned.

This causes a problem in cases where the expected role is implicitly defined by the aria attributes present on the node.

Example:

`<th aria-sort="descending">`

While the above example is valid, currently `cell` is returned as the implicit role, and aXe correctly marks this as error, since a `cell` cannot have `aria-sort`.

In order to fix this, aXe needs to determine the optimal combination of valid aria-attributes and possible implicit roles, and return this role.

jkva:
This fixes #290.

This is my first PR to this repository; I would much appreciate a code review. See also code comments pertaining to helper functions.

In case of multiple implicit roles existing for a node, we need to infer the expected implicit context based on a combination of applied aria-attributes for the node, and which of these are applicable to the existing implicit roles.

Each implicit role is given a score based on the combination of aria-attributes for the role, and the list of aria-attributes for the node.

The highest scoring role is returned. When multiple equally highest scored roles exist, the first is returned.

### DIFF
--- a/lib/commons/aria/attributes.js
+++ b/lib/commons/aria/attributes.js
@@ -22,6 +22,7 @@ aria.allowedAttr = function (role) {
 	var roles = lookupTables.role[role],
 		attr = (roles && roles.attributes && roles.attributes.allowed) || [],
 		requiredAttr = (roles && roles.attributes && roles.attributes.required) || [];
+
 	return attr.concat(lookupTables.globalAttributes).concat(requiredAttr);
 };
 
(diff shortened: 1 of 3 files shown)


dylanb:
Hey @jkva thanks for this submit. It is a huge improvement over what we currently have. I originally thought that we should be returning all the roles, but I like this solution better because it will still catch problems where mismatches exist in attributes used (for example, use one attribute from one implicit role and another attribute from another implicit role)

dylanb:
Tested the changes on my machine to bypass the Snyk fail

jkva:
Thanks for the merge @dylanb. Agreed on your comment, I wanted the function itself to remain identical in its use - 'return most suitable implicit role' - while making sure that the surrounding validation checks can still catch other problems. Makes things simpler to test as well.

PR #300, opened 2017-04-06
Fix false positive for th[scope=col] in tfoot and th[scope=row] in last column

majornista:
When a `th[scope=col]` element is in the `tfoot` and has no `td` cells
following it in the column, or when a `th[scope=row]` element is the
last child in the row, the `th-has-data-cells` check fails because it
doesn’t check for preceding cells in the column or row.

### DIFF
--- a/lib/checks/tables/th-has-data-cells.js
+++ b/lib/checks/tables/th-has-data-cells.js
@@ -29,6 +29,13 @@ var headers = cells.filter(function (cell) {
 
 var tableGrid = tableUtils.toGrid(node);
 
+var isRowHeader = false;
+var dataCellIsNotEmpty = function (out, cell) {
+	var isHeader = isRowHeader ? tableUtils.isRowHeader : tableUtils.isColumnHeader;
+	return out || (cell.textContent.trim() !== '' &&
+		!isHeader(cell));
+};
+
 // Look for all the bad headers
 return headers.reduce(function (res, header) {
 	if (header.id && reffedHeaders.indexOf(header.id) !== -1) {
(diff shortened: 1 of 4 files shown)


lib/checks/tables/th-has-data-cells.js
```diff
@@ -41,19 +45,17 @@ return headers.reduce(function (res, header) {
...
 	}
 
 	// Look for any data cells or column headers that this might refer to
 	if (!hasCell && tableUtils.isRowHeader(header)) {
 		hasCell = tableUtils.traverse('right', pos, tableGrid)
-		.reduce((out, cell) => {
-			return out || (cell.textContent.trim() !== '' &&
-				!tableUtils.isRowHeader(cell));
-		}, false);
+		.reduce(dataCellIsNotEmpty, false) ||
```
marcysutton: The code called `tableUtils.isRowHeader` here before but the new function calls `isColumnHeader`. The tests still pass, but I wanted to make sure this wasn't an oversight

majornista: Oh, yeah. That looks like an error on my part, I'll fix it.

majornista:
The use case is when TH is repeated in footer of a scrolling table, or when the TH for a row is right aligned, which is a permitted, albeit unlikely use case, in a component library we have.

lib/checks/tables/th-has-data-cells.js
majornista: Should be fixed now.

test/checks/tables/th-has-data-cells.js
```diff
@@ -24,6 +24,8 @@ describe('th-has-data-cells', function () {
 			'<table>' +
 			'  <tr> <th>hi</th> <td>hello</td> </tr>' +
 			'  <tr> <th>hi</th> <td>hello</td> </tr>' +
+			'  <tr> <td>hello</td> <th>hi</th> </tr>' +
```
dylanb: Last time I checked, this technique did not work on all major screen reader - browser combinations.

What screen readers have you tested?

dylanb:
@majornista the philosophy of aXe is to only allow techniques that work on all "platforms" to pass. However, if a technique works on some "platforms" (depending on which and how many) we will sometimes break a fail out into a separate check so it can be removed by a custom rule or using the options.

Last time I checked, the technique of middle, or right positioned header cells in a table row did not work in all screen readers. Which screen readers have you tested?

test/checks/tables/th-has-data-cells.js
majornista: I agree that support for this markup is poor across browser screen reader combinations. I tested on Windows 10 using IE11/JAWS 17, IE11/NVDA 2017.1, Firefox/JAWS 17, Firefox/NVDA 2017.1, Chrome/JAWS 17, and Chrome/NVDA 2017.1 and only Chrome/NVDA 2017.1 read the right-aligned row headers.

On macOS, the right-aligned cells were announced with Chrome/VoiceOver, but not with Safari/VoiceOver.

The right-aligned cells were not announced with iOS/VoiceOver or either Chrome or Firefox on Android with Talkback.

The use case where I've seen this happening involves repetition of column headers in a TFOOT, or a row header repeated as the last cell in a row in long and wide data tables. So it's not like the cells aren't labelled, but I'm not sure if the test should account for this or not. Also a row header in the middle of the row would pass this test as originally written because we are just checking to make sure that at least one subsequent data cell in the same row as the TH has content.

marcysutton:
> On macOS, the right-aligned cells were announced with Chrome/VoiceOver, but not with Safari/VoiceOver.

> The right-aligned cells were not announced with iOS/VoiceOver or either Chrome or Firefox on Android with Talkback.

After talking with @dylanb, the lack of support for this table structure on mobile devices means that we can't merge this change as it doesn't fit in with the axe-core philosophy of flagging code that isn't fully supported. In the future, we want to add an Accessibility Supported API to axe-core that would allow us to be more detailed about support: at that time we could revisit this change.

PR #304, opened 2017-04-17
Add configuration options

alice:
- `iframes`: if false, only collect results from top frame
- `elementRef`: if true, return live element references for results from top frame
- `noSelectors`: if true, omit `target` for results from top frame.

### DIFF
--- a/lib/core/public/a11y-check.js
+++ b/lib/core/public/a11y-check.js
@@ -16,6 +16,10 @@ axe.a11yCheck = function (context, options, callback) {
 		options = {};
 	}
 
+	if (!('iframes' in options)) {
+		options.iframes = true;
+	}
+
 	var audit = axe._audit;
 	if (!audit) {
 		throw new Error('No audit configured');
@@ -30,4 +34,3 @@ axe.a11yCheck = function (context, options, callback) {
 		}
 	}, axe.log);
 };
-
--- a/lib/core/public/a11y-check.js
+++ b/lib/core/public/a11y-check.js
@@ -16,10 +16,6 @@ axe.a11yCheck = function (context, options, callback) {
 		options = {};
 	}
 
-	if (!('iframes' in options)) {
-		options.iframes = true;
-	}
-
 	var audit = axe._audit;
 	if (!audit) {
 		throw new Error('No audit configured');
--- a/lib/core/reporters/helpers/process-aggregate.js
+++ b/lib/core/reporters/helpers/process-aggregate.js
@@ -1,5 +1,5 @@
 /*global helpers */
-function normalizeRelatedNodes(node, xpath) {
+function normalizeRelatedNodes(node, options) {
 	'use strict';
 	['any','all','none'].forEach((type) => {
 		if (!Array.isArray(node[type])) {
@@ -9,16 +9,20 @@ function normalizeRelatedNodes(node, xpath) {
 		.forEach((checkRes) => {
 			checkRes.relatedNodes = checkRes.relatedNodes.map((relatedNode) => {
 				var res = {
-					html: relatedNode.source,
-					target: relatedNode.selector
+					html: relatedNode.source
 				};
-				if (xpath) {
+				if (options.elementRef && !relatedNode.fromFrame) {
+					res.element = relatedNode.element;
+				}
+				if (!options.noSelectors || relatedNode.fromFrame) {
+					res.target = relatedNode.selector;
+				}
+				if (options.xpath) {
 					res.xpath = relatedNode.xpath;
 				}
 				return res;
 			});
 		});
-
 	});
 }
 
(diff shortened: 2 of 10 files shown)


lib/core/reporters/helpers/process-aggregate.js
```diff
@@ -37,16 +41,23 @@ helpers.processAggregate = function (results, options) {
...
+						}
 						if (options.xpath) {
 							subResult.xpath = subResult.node.xpath;
 						}
 					}
 					delete subResult.result;
 					delete subResult.node;
 
-					normalizeRelatedNodes(subResult, options.xpath);
+					normalizeRelatedNodes(subResult,
```
dylanb: why not just pass in options?

test/integration/full/options/options.js
```diff
@@ -0,0 +1,190 @@
...
+	});
+
+	function $id(id) {
+		return document.getElementById(id);
+	}
+
+	describe('iframes', function() {
+		it('should include iframes by default', function(done) {
+			var config = {};
+            axe.a11yCheck(document, config, function(results) {
```
dylanb: There are a bunch of mixed spaces and tabs in this file, this is one of them

dylanb:
Should we create a rule that puts every iframe into a "needs review" result if there are iframes and `options.iframes === false`?

Pros: every user would be able to know/display-to-end-users the fact that there are parts of the page that are not being analyzed

Cons: We would have to pass this option into every check so that it could operate appropriately dependent on the options - this is not a huge con IMO because it would allow all checks to modify their behavior based on options

lib/core/public/a11y-check.js
```diff
@@ -16,6 +16,10 @@ axe.a11yCheck = function (context, options, callback) {
 		options = {};
 	}
 
+	if (!('iframes' in options)) {
```
marcysutton: We should make sure this is also supported by `axe.run`. Perhaps moving it into run-rules would cover both API methods more easily?

marcysutton:
The tests pass with our internal fork, where @dylanb made some changes to our tests. We're planning a release next week, so we can merge this at that time.

lib/core/public/a11y-check.js
dylanb: Actually you could get rid of this code simply by negating the test in `run-rules.js`

lib/core/reporters/helpers/process-aggregate.js
alice: Hm, I swear I had a reason but it seems to work fine. Fixed.

lib/core/public/a11y-check.js
alice: Done! How can I test it for `axe.run`?

test/integration/full/options/options.js
alice: Fixed. Sigh, tabs.

alice:
Tests are passing now! Maybe @marcysutton's suggestion to move the iframe option logic down into run-rules?

lib/core/public/a11y-check.js
marcysutton: We could put a test in `test/core/public/run-rules.js`?

marcysutton:
Tests are passing for me too! :clap:

alice:
Ok, I'll add a test in `test/core/public/run-rules.js` - anything else you'd like me to do?

dylanb:
I'd like to see some other opinions on the needs review question because if possible, I'd like to make this change at the same time...

marcysutton:
@dylanb I like your idea. Perhaps it could piggyback on our planned Needs Review rule for hidden regions (i.e. flagging content skipped due to `display: none` or `aria-hidden="true"`)? We could float up iframes that were skipped as well.

Writing a new rule would delay our release though, is that something we really want to take on now?

alice:
I don't think that's something I would be doing in this change, right?

marcysutton:
I think we should merge this and work on adding a new rule to the next release. Thanks so much for your contribution, @alice!!

alice:
Ok, I'll just write up that one test then it should be good to go :)

alice:
Just want to check: are you ok with `noSelectors` having the inverted sense, or would you prefer `selectors`?

marcysutton:
That's a great question. I do think `selectors` would be more straight forward from an API consumption perspective!

alice:
Changes made - I don't know what's up with the build :( Can't repro error locally

marcysutton:
I was able to repro it and debug with `grunt dev`, and found the reporter tests were failing looking for `results.violations[0].nodes`. The reporter tests mock runRules and set options explicitly, and I got them to pass by adding `selectors: true` to the options objects in our 4 reporter test files.

- On line 81 of `test/core/reporters/no-passes.js`: 
```var noPassOpt = { reporter: 'no-passes', selectors: true };```

- On line 144 of `test/core/reporters/v1.js`:
```var optionsV1 = { reporter: 'v1', selectors: true };```

- On line 86 of `test/core/reporters/v2.js`:
```var optionsV2 = { reporter: 'v2', selectors: true };```

- On line 96 of `test/core/reporters/na.js`:
```var naOption = { reporter: 'na', selectors: true };```

I could commit those to your fork if you [give me permission](https://help.github.com/articles/committing-changes-to-a-pull-request-branch-created-from-a-fork/), otherwise feel free to make those changes and I can merge it.

marcysutton:
I'm going to go ahead and merge this and make another commit once it lands, so we can keep moving forward with our release today. I've opened https://github.com/dequelabs/axe-core/issues/307 to track @dylanb's idea for iframes.

alice:
Sorry I didn't get a chance to come back to this. I hope you don't need to pass `selectors: true` to each test to get things to work; that should be the default if things are working properly.

PR #309, opened 2017-04-25
Check for explicit `selectors` and `iframes` in options instead of attempting to set default

alice:


marcysutton:
This works! Thanks for the contribution!

PR #314, opened 2017-04-26
Replace TravisCI with CircleCI

dylanb:
Also now adds Chrome testing to the CI, was not able to get FF testing working


marcysutton:
I've had nothing but trouble with Firefox lately too. I'd love to get that working, but it will definitely take more time.

PR #317, opened 2017-04-27
Shadow DOM implementation

dylanb:
**DO NOT MERGE**

Opening a pull request to allow discussion on implementation to happen.

Current state:

1. Implemented own querySelectorAll to resolve rule selectors across shadow DOM boundaries
1. Make sure all existing rules work with the changes to qsa
1. Documented design for result object target array with target selectors that work across shadow DOM boundaries
1. Updated utilities like findUp, text.label and text.accessibleText to support shadow DOM
1. Implemented test utilities to allow conditional execution of shadow DOM tests
1. Updating rules to support shadow DOM


doc/API.md
```diff
@@ -444,7 +452,7 @@ Each object returned in these arrays have the following properties:
 * `nodes` - Array of all elements the Rule tested
 	* `html` - Snippet of HTML of the Element
 	* `impact` - How serious the violation is. Can be one of "minor", "moderate", "serious", or "critical" if the test failed or `null` if the check passed
-	* `target` - Array of selectors that has each element correspond to one level of iframe or frame. If there is one iframe or frame, there should be two entries in `target`. If there are three iframe levels, there should be four entries in `target`.
+	* `target` - Array of either strings or Arrays of strings. If the item in the array is a string, then it is a CSS selector. If there are multiple items in the array each item corresponds to one level of iframe or frame. If there is one iframe or frame, there should be two entries in `target`. If there are three iframe levels, there should be four entries in `target`. If the item in the Array is an Array of strings, then it points to an element in a shadow DOM and each item (except the n-1th) in this array is a selector to a DOM element with a shadow DOM. The last element in the array points to the final shadow DOM node.
```
marcysutton: This is a confusing paragraph. I wonder if there's a way to explain it better? 

>If the item in the Array is an Array of strings, then it points to an element in a shadow DOM

How would you tell that apart from an iframe?

doc/API.md
```diff
@@ -527,6 +535,23 @@ axe.run(document, {
...
+  * `help` - `"Elements must have sufficient color contrast"`
+  * `helpUrl` - `"https://dequeuniversity.com/rules/axe/2.1/color-contrast?application=axeAPI"`
+  * `id` - `"color-contrast"`
+		* `nodes`
+			* `target[0][0]` - `"header > aria-menu"`
+			* `target[0][1]` - `"li.expanded"`
+
+* `violations[1]` ...
+
+As you can see the `target` array contains one item that is an array. This array contains two items, the first is a CSS selector string that finds the custom element `<aria-menu>` in the `<header>`. The second item in this array is the selector within that custom element's shadow DOM to find the `<li>` element with a class of `expanded`.
```
marcysutton: > As you can see the `target` array contains one item that is an array.

I see two items...? Perhaps this should be spelled out more completely since I wasn't able to "see" it?

lib/core/utils/composed-tree.js
```diff
@@ -0,0 +1,99 @@
...
+			res = res.concat(replacements);
+		}
+		return res;
+	}
+
+	if (node.documentElement) { // document
+		node = node.documentElement;
+	}
+	nodeName = node.nodeName.toLowerCase();
+	// for some reason Chrome's marquee element has an open shadow DOM
```
marcysutton: Did marquee return something different that required filtering it out here?

lib/core/utils/composed-tree.js
```diff
@@ -0,0 +1,99 @@
...
+	if (node.shadowRoot && nodeName !== 'marquee') {
```
marcysutton: Did `marquee` return something different that required filtering it out here?

lib/core/utils/composed-tree.js
```diff
@@ -0,0 +1,99 @@
...
+		// generate an ID for this shadow root and overwrite the current
+		// closure shadowId with this value so that it cascades down the tree
+		shadowId = 'a' + Math.random().toString().substring(2);
+		realArray = Array.from(node.shadowRoot.childNodes);
+		return realArray.reduce(reduceShadowDOM, []);
+	} else {
+		if (nodeName === 'content') {
+			realArray = Array.from(node.getDistributedNodes());
```
marcysutton: When I looked up `getDistributedNodes` it looked related to the [Shadow DOM v0 API](https://www.html5rocks.com/en/tutorials/webcomponents/shadowdom-301/). Should we make a note of that here?

lib/core/utils/composed-tree.js
```diff
@@ -0,0 +1,99 @@
...
+			return realArray.reduce(reduceShadowDOM, []);
+		} else if (nodeName === 'slot') {
+			realArray = Array.from(node.assignedNodes());
+			return realArray.reduce(reduceShadowDOM, []);
+		} else {
+			if (node.nodeType === 1) {
```
marcysutton: Does this run for regular non-shadowDOM nodes so they all have `children` and `actualNode` properties? What happens if there isn't a shadowId?

doc/API.md
dylanb: iframe: `['selector to iframe', 'selector within iframe']
shadowDOM: `[['selector of shadow DOM element', 'selector within shadow DOM']]`

lib/core/utils/composed-tree.js
dylanb: yes, its shadow DOM is open instead of closed

lib/core/utils/composed-tree.js
dylanb: It runs for everything. Only relevant node types (1 and 3) will be returned (not comment nodes for example)

dylanb:
Updated status:

1. Implemented the shadow DOM selector (as documented)
1. Implemented contains
1. Implemented isHidden

Started work on individual rules, starting with ARIA

doc/API.md
marcysutton: I still find these arrays confusing; users will have to be intimately familiar with our documentation to use them. Is there any way to put them in an object with named keys instead of relying on array depth?

lib/core/utils/get-selector.js
```diff
@@ -167,54 +168,73 @@ function getElmFeatures (elm, featureCount) {
...
+    }
+    addParent = (minDepth !== 0 || !isUnique);
+  }
 
-		node = node.parentNode;
-	}
+  const selectorParts = [selector, ...childSelectors];
 
-	return parts.join(' > ');
+  if (elm.parentElement && elm.parentElement.nodeType !== 11 &&
```
marcysutton: Perhaps a good spot for a code comment about filtering out DocumentFragments?

test/core/utils/is-hidden.js
```diff
@@ -1,3 +1,19 @@
+function createContentHidden() {
+	'use strict';
+	var group = document.createElement('div');
+	group.innerHTML = '<label id="mylabel">Label</label><input aria-labelledby="mylabel" type="text" />';
+	return group;
+}
+
+function makeShadowTreeHidden(node) {
```
marcysutton: What makes this tree hidden? I don't see any special CSS that would hide it, or a closed mode.

test/core/utils/get-selector.js
```diff
@@ -286,4 +302,35 @@ describe('axe.utils.getSelector', function () {
...
+	it('toRoot: should work with shadow DOM', function () {
+		var shadEl;
+
+		if (document.body && typeof document.body.attachShadow === 'function') {
+			// shadow DOM v1 - note: v0 is compatible with this code, so no need
+			// to specifically test this
+			fixture.innerHTML = '<div></div>';
+			makeShadowTreeGetSelector(fixture.firstChild);
+			shadEl = fixture.firstChild.shadowRoot.querySelector('input#myinput');
+			assert.deepEqual(axe.utils.getSelector(shadEl, { toRoot: true }), [
```
marcysutton: Can you describe what `toRoot` does? Should we add it to the docs now or later?

doc/API.md
dylanb: the disadvantage of that is it would break existing code - which will still work for everything that is not inside a shadow DOM

test/core/utils/get-selector.js
dylanb: @WilcoFiers can you add documentation for this?

test/core/utils/is-hidden.js
dylanb: For the shadow DOM test functions, I am using a convention of adding something unique to the end of the makeShadowTree - I used "hidden" because this is the test file for  the `isHidden` function and it seemed like overkill to have to write `makeShadowTreeIsHidden` - its just a test function though, so I don't think its a big deal

WilcoFiers:
Should we create a `develop/3.0` branch instead @dylanb? I don't think we need to keep this PR around, do we?

marcysutton:
I like that idea @WilcoFiers.

marcysutton:
I created a develop-2x branch and updated this branch from develop using a merge commit–rebasing was going to take hours, and I couldn't tell what changes to keep for each (of the many) conflicts. I'll merge this PR to develop and we can cut the 3.0.0-alpha release.

PR #330, opened 2017-05-13
Fix bugs with implicit labels and TRs for color contrast

marcysutton:
To address default implicit labels and inputs returning as Can't Tell (due to what it thinks is bgOverlap), this change adds some logic to the `calculateObscuringAlpha` function in `lib/commons/color/get-background-color.js` to check if an element's top left corner is actually overlapping in addition to returning the alpha value. I based it on the top left corner and not the center point because of the problem I was trying to solve–inputs being nested in implicit labels with the default styling causing an incomplete result.

I ran into issues with the integration tests because axe-core sometimes returns the label as the node, and sometimes the label and the input, depending on whether the tests are running in `grunt dev` or `grunt test-fast`. The runner used for our integration tests complicates debugging with one `axe.run` call for the entire fixture and flattening the results, so this test issue has taken more time than I'd like, hence submitting the code knowing there is a test failure. IMO aXe should just return the label and not the input, but I'm not quite sure where it's deciding on the relevant node when our test harness is being used.

I confirmed the fix in this PR works when I run it manually against a page like [this one](http://coc.kciprojects.com/analysis/implicit/), or [HHS.gov](hhs.gov) where it was originally reported.

Note: I also did some cleanup to the `color-contrast.html` file because it had a random DIV with a background image at the top but no closing tag. I'm surprised the tests worked at all.

### DIFF
--- a/lib/checks/color/color-contrast.json
+++ b/lib/checks/color/color-contrast.json
@@ -12,7 +12,7 @@
         "imgNode": "Element's background color could not be determined because element contains an image node",
         "bgOverlap": "Element's background color could not be determined because it is overlapped by another element",
         "fgAlpha" : "Element's foreground color could not be determined because of alpha transparency",
-        "elmOutsideParent": "Element's background color could not be determined because it extends beyond a parent node",
+        "elmPartiallyObscured": "Element's background color could not be determined because it's partially obscured by another element",
         "default": "Unable to determine contrast ratio"
       }
     }
--- a/lib/checks/color/color-contrast.json
+++ b/lib/checks/color/color-contrast.json
@@ -12,6 +12,7 @@
         "imgNode": "Element's background color could not be determined because element contains an image node",
         "bgOverlap": "Element's background color could not be determined because it is overlapped by another element",
         "fgAlpha" : "Element's foreground color could not be determined because of alpha transparency",
+        "elmOutsideParent": "Element's background color could not be determined because it extends beyond a parent node",
         "default": "Unable to determine contrast ratio"
       }
     }
--- a/lib/commons/color/get-background-color.js
+++ b/lib/commons/color/get-background-color.js
@@ -125,6 +125,7 @@ function includeMissingElements(elmStack, elm) {
 	const tagArray = elmStack.map((elm) => {
 		return elm.tagName;
 	});
+	let bgNodes = elmStack;
 	for (let candidate in elementMap) {
 		if (elementMap.hasOwnProperty(candidate)) {
 			// tagName matches key
@@ -134,17 +135,18 @@ function includeMissingElements(elmStack, elm) {
 					// found an ancestor not in elmStack, and it overlaps
 					let overlaps = axe.commons.dom.visuallyOverlaps(elm.getBoundingClientRect(), ancestorMatch);
 					if (overlaps) {
-						elmStack.splice(elmStack.indexOf(elm) + 1, 0, ancestorMatch);
+						bgNodes.splice(elmStack.indexOf(elm) + 1, 0, ancestorMatch);
 					}
 				}
 			}
 			// tagName matches value
 			// (such as LABEL, when matching itself. It should be in the list, but Phantom skips it)
 			if (elm.tagName === elementMap[candidate] && tagArray.indexOf(elm.tagName) === -1) {
-				elmStack.splice(tagArray.indexOf(candidate) + 1, 0, elm);
+				bgNodes.splice(tagArray.indexOf(candidate) + 1, 0, elm);
 			}
 		}
 	}
+	return bgNodes;
 }
 
 /**
@@ -152,8 +154,10 @@ function includeMissingElements(elmStack, elm) {
  * @private
  * @param {Array} elmStack
  */
-function consultDocumentBody(elmStack) {
+function sortPageBackground(elmStack) {
 	let bodyIndex = elmStack.indexOf(document.body);
+
+	let bgNodes = elmStack;
 
 	if (// Check that the body background is the page's background
 		bodyIndex > 1 && // only if there are negative z-index elements
@@ -161,12 +165,13 @@ function consultDocumentBody(elmStack) {
 		getBgColor(document.documentElement).alpha === 0
 	) {
 		// Remove body and html from it's current place
-		elmStack.splice(bodyIndex, 1);
-		elmStack.splice( elmStack.indexOf(document.documentElement), 1);
+		bgNodes.splice(bodyIndex, 1);
+		bgNodes.splice( elmStack.indexOf(document.documentElement), 1);
 
 		// Put the body background as the lowest element
-		elmStack.push(document.body);
+		bgNodes.push(document.body);
 	}
+	return bgNodes;
 }
 
 /**
@@ -190,10 +195,9 @@ color.getBackgroundStack = function(elm) {
 			window.innerHeight - 1);
 
 	let elmStack = document.elementsFromPoint(x, y);
-	includeMissingElements(elmStack, elm);
+	elmStack = includeMissingElements(elmStack, elm);
 	elmStack = dom.reduceToElementsBelowFloating(elmStack, elm);
-
-	consultDocumentBody(elmStack);
+	elmStack = sortPageBackground(elmStack);
 
 	// Return all elements BELOW the current element, null if the element is undefined
 	let elmIndex = elmStack.indexOf(elm);
--- a/lib/commons/color/get-background-color.js
+++ b/lib/commons/color/get-background-color.js
@@ -38,16 +38,27 @@ function getBgColor(elm, elmStyle) {
 	return bgColor;
 }
 
-function calculateObscuringAlpha(elmIndex, elmStack) {
+function contentOverlapped(targetElement) {
+	// Get dimensions of text content box.
+	// Subsequent clientRects are bounding boxes, if different than content box
+	var targetRect = targetElement.getClientRects()[0];
+	var elementAtPoint = document.elementFromPoint(targetRect.left, targetRect.top);
+	if (elementAtPoint !== null && elementAtPoint.innerHTML !== targetElement.innerHTML) {
+		return true;
+	}
+	return false;
+}
+
+function calculateObscuringAlpha(elmIndex, elmStack, originalElm) {
 	var totalAlpha = 0;
 
 	if (elmIndex > 0) {
-		// there are elements above our element, check if they are contribute to the background
+		// there are elements above our element, check if they contribute to the background
 		for (var i = elmIndex - 1; i >= 0; i--) {
 			let bgElm = elmStack[i];
 			let bgElmStyle = window.getComputedStyle(bgElm);
 			let bgColor = getBgColor(bgElm, bgElmStyle);
-			if (bgColor.alpha) {
+			if (bgColor.alpha && contentOverlapped(originalElm, bgElm)) {
 				totalAlpha += bgColor.alpha;
 			} else {
 				// remove elements not contributing to the background
@@ -102,7 +113,7 @@ color.getBackgroundStack = function(elm) {
 
 	// Return all elements BELOW the current element, null if the element is undefined
 	let elmIndex = elmStack.indexOf(elm);
-	if (calculateObscuringAlpha(elmIndex, elmStack) >= 0.99) {
+	if (calculateObscuringAlpha(elmIndex, elmStack, elm) >= 0.99) {
 		// if the total of the elements above our element results in total obscuring, return null
 		axe.commons.color.incompleteData.set('bgColor', 'bgOverlap');
 		return null;
--- a/lib/commons/color/get-background-color.js
+++ b/lib/commons/color/get-background-color.js
@@ -1,4 +1,5 @@
 /* global axe, color, dom */
+/*jshint maxstatements: 22 */
 const graphicNodes = [
 	'IMG', 'CANVAS', 'OBJECT', 'IFRAME', 'VIDEO', 'SVG'
 ];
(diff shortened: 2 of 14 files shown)


lib/commons/color/get-background-color.js
```diff
@@ -44,16 +45,27 @@ function getBgColor(elm, elmStyle) {
 	return bgColor;
 }
 
-function calculateObscuringAlpha(elmIndex, elmStack) {
+function contentOverlapped(targetElement) {
+	// Get dimensions of text content box.
+	// Subsequent clientRects are bounding boxes, if different than content box
+	var targetRect = targetElement.getClientRects()[0];
+	var elementAtPoint = document.elementFromPoint(targetRect.left, targetRect.top);
+	if (elementAtPoint !== null && elementAtPoint.innerHTML !== targetElement.innerHTML) {
```
WilcoFiers: Why can't this be a straight element comparison?

test/checks/color/color-contrast.js
```diff
@@ -157,4 +157,39 @@ describe('color-contrast', function () {
 		assert.equal(checkContext._data.missingData[0].reason, 'bgGradient');
 		assert.equal(checkContext._data.contrastRatio, 0);
 	});
+
+	it('should return undefined when there are elements overlapping', function () {
```
WilcoFiers: I'm not sure I understand why this case needs to be reviewed. The text is completely obscured, so there is no contrast issue, right?

test/checks/color/color-contrast.js
```diff
@@ -157,4 +157,39 @@ describe('color-contrast', function () {
...
+		fixture.innerHTML = '<div style="color: black; background-color: white; width: 200px; height: 100px; position: relative;" id="target">' +
+			'My text <div style="position: absolute; top:0; left: 0; background-color: white; width: 100%; height: 100%;"></div></div>';
+		var target = fixture.querySelector('#target');
+		assert.isUndefined(checks['color-contrast'].evaluate.call(checkContext, target));
+		assert.equal(checkContext._data.missingData[0].reason, 'bgOverlap');
+		assert.equal(checkContext._data.contrastRatio, 0);
+	});
+
+	it('should return true when a form wraps mixed content', function() {
```
WilcoFiers: Why would this form element ever be tested? Shouldn't the matcher throw this out because it doesn't have child nodes with text in them?

test/commons/color/get-background-color.js
```diff
@@ -190,6 +190,64 @@ describe('color.getBackgroundColor', function () {
...
+		var expected = new axe.commons.color.Color(243, 243, 243, 1);
+		assert.equal(actual.red, expected.red);
+		assert.equal(actual.green, expected.green);
+		assert.equal(actual.blue, expected.blue);
+		assert.equal(actual.alpha, expected.alpha);
+		assert.deepEqual(bgNodes, [parent]);
+	});
+
+	it('should count an implicit label as a background element', function () {
+		fixture.innerHTML = '<label id="target" style="background-color: #fff;">My label' +
```
WilcoFiers: Should use a different background value here. White is going to be the background color regardless of if this works or not.

test/commons/dom/visually-contains.js
```diff
@@ -47,4 +47,12 @@ describe('dom.visuallyContains', function () {
 		assert.isTrue(axe.commons.dom.visuallyContains(target, target.parentNode));
 	});
 
+	it('should return true when element is inline', function () {
```
WilcoFiers: Couldn't you use position:absolute or float to move inline elements outside their parent? Through clipping an child element can also be outside it's parent. There are probably some other ways to do it too. So I'm not sure the assumption you're making here is right.

lib/commons/color/get-background-color.js
marcysutton: The strict comparison was failing on `elementAtPoint` === `targetElement` so I changed it to   a loose equality check and jshint was complaining about it. So I changed it to innerHTML for both of those reasons.

test/checks/color/color-contrast.js
marcysutton: The code was already in there and it was occurring for some nodes, this added a test for pre-existing code.

test/checks/color/color-contrast.js
marcysutton: It was an extra check just to make sure the DOM walker didn't break in that case. We probably don't need it though

test/commons/color/get-background-color.js
marcysutton: I'd actually remove the background-color value here entirely after learning more about this edge case: if the label is inline and it doesn't have a background-color set on it, the element wouldn't get added to the stack.

test/commons/dom/visually-contains.js
marcysutton: I can add some more tests for other scenarios, but this was definitely a problem case.

marcysutton: Also: this is just for the `visuallyContains` method, which checks the element's bounding box against the parent bounding box and parent scrollArea: it doesn't work on inline elements at all. The scroll area is completely off cross-platform so it won't return consistently.

lib/checks/color/color-contrast.json
```diff
@@ -12,6 +12,7 @@
         "imgNode": "Element's background color could not be determined because element contains an image node",
         "bgOverlap": "Element's background color could not be determined because it is overlapped by another element",
         "fgAlpha" : "Element's foreground color could not be determined because of alpha transparency",
+        "elmOutsideParent": "Element's background color could not be determined because it extends beyond a parent node",
         "default": "Unable to determine contrast ratio"
```
dylanb: I know this is not a changed line, but I have seen this message appear many, many times. What are the cases where this would still be displayed?

lib/commons/color/get-background-color.js
```diff
@@ -1,4 +1,5 @@
 /* global axe, color, dom */
+/*jshint maxstatements: 22 */
```
dylanb: Seems to me like we should be thinking about refactoring this at this point

lib/checks/color/color-contrast.json
marcysutton: The default line comes up any time the check returns undefined but there isn't incompleteData passed through for it. The `elmOutsideParent` code path definitely caused it. I didn't see any others, but if we keep seeing it after this is merged we can identify and float up more reasons.

lib/commons/color/get-background-color.js
```diff
@@ -44,16 +45,27 @@ function getBgColor(elm, elmStyle) {
 	return bgColor;
 }
 
-function calculateObscuringAlpha(elmIndex, elmStack) {
+function contentOverlapped(targetElement) {
```
dylanb: This function is confusing to me:

1. It has one argument, but is called with two arguments.
1. It is only called when there is an element above the target element but it is trying to find the target element using document.elementFromPoint - which should return the topmost element - which will therefore never be the target element

I don't think this does what it is intended to do

lib/commons/color/get-background-color.js
```diff
@@ -65,9 +77,48 @@ function calculateObscuringAlpha(elmIndex, elmStack) {
 }
 
 function elmOutsideParent(elm, bgElm, bgColor) {
```
dylanb: This function makes no sense to me either - why do we go through the effort of using elementsFromPoint if we are then going to ignore elements positioned outside their parent?

marcysutton: I'm not actually sure–I was simply plumbing messaging through code that already existed. I abstracted the code both you and Wilco had worked on into a separate function to make it easier to follow, here's the commit where it originated: https://github.com/dequelabs/axe-core/commit/7252b57c74e0a3d1d8b731a4920756f731e961b1#diff-4d3897aa0085003d18376b4bb785364aR79

I could try removing it and seeing how it impacts real world scans.

lib/commons/color/get-background-color.js
```diff
@@ -65,9 +77,48 @@ function calculateObscuringAlpha(elmIndex, elmStack) {
...
+			reason: 'elmOutsideParent'
+		});
+	}
+	return outside;
 }
+
+// document.elementsFromPoint misses some elements we need
+// i.e. TR is missing from table elementStack and leaves out bgColor
+// https://github.com/dequelabs/axe-core/issues/273
+function includeMissingElements(elmStack, elm) {
```
marcysutton: In theory yes, but in practice the only elements I've seen it miss are `TR`, which can't be positioned; and `LABEL`, but only when it's `display: inline`. The reason I bring this up is I'm not sure how I'd write a test for it given those two scenarios.

marcysutton: I guess one more case is `pointer-events: none;`. But I wasn't going to add support for that in this PR.

marcysutton: In theory, yes. But in practice, it only occurs with `TR` elements, which can't be positioned, and inline `LABEL` elements. (It also occurs with `pointer-events: none;`, but I'm not addressing that here)

I could add a visuallyContains check to this but there's a wrinkle, in that `LABEL` ends up missing from the list when itself is being evaluated, so there's nothing to compare against. I think `elmOutsideParent` actually does have a use with this code in place.

lib/commons/color/get-background-color.js
dylanb: @WilcoFiers do you know why we do this at all?

lib/commons/color/get-background-color.js
dylanb: TR can be positioned absolute - even though that would be bad code - but that is not the point - any of the parents could be positioned absolute which means that the parents of the absolute parents might not be in the actual background and might be added

lib/commons/color/get-background-color.js
WilcoFiers: @marcysutton I think we should do is lift that `bgColor.alpha !== 0` check out of this function. That would make more sense for it.

The purpose of this check is to allow us to drop an element into cantTell if the parent giving it a background only partially overlaps the child element. E.g. that element may have multiple background colors, which we're not set up to test.

dylanb: Shouldn't we do that check for every overlapping element and not only the parent?

WilcoFiers: It doesn't. This loops over the entire stack of elements beneath it. The function name isn't all that clear though. These things don't even have to be ancestor elements. They can be any two elements.

marcysutton: I'll look over these functions to see they can be refactored to be smaller and still cover the necessary use cases. But speaking of those, @dylanb do you remember what the Capital One bug was you mention in this commit? That's where `bgColor.alpha` originated, so I want to make sure it doesn't regress. https://github.com/dequelabs/axe-core/commit/6557ff85f9a9a523f831c0642a20b335746d00cd

dylanb: i believe this codepen is the one http://codepen.io/wilcofiers/pen/wWJYrM

lib/commons/color/get-background-color.js
marcysutton: After a bit of wrangling this into a real test I'm going to opt for `<label>` instead of `<tr>`, as most browsers will "fix" a table when there is no TD. To keep it aligned with the rest of the changes in this PR I'll use an inline label so it won't get transformed by the browser.

lib/commons/color/get-background-color.js
marcysutton: I'm going to keep the logic in this function the same–it doesn't actually impact the case I was trying to address (inline labels triggering `bgOverlap`), and I don't want to regress other cases this was intended to fix. I'll rename it to `elmPartiallyObscured` so the missingData exposed is less confusing.

marcysutton:
Ok, I've made some updates to this. Develop was behind master when I pulled in the latest incompleteData changes so there are a few extra commits–I can remove those if necessary, but I'll leave them for now.

Here are the relevant changes to look at:
- `includeMissingElements` checks if ancestor nodes overlap, including TRs with absolutely-positioned TDs (valid HTML and CSS)
- `contentOverlapping` method checks the entire stack of elementsAtPoint in addition to the top one to see if any overlap. 
- Additional tests for absolutely-positioned elements to make sure this doesn't add elements that aren't applicable.
- Refactored some of the getBackgroundStack code into a separate function to reduce complexity. Ideally I wanted to move the coordinates checking into a separate function but it failed a disabled scrolling test in Phantom no matter what I tried, so I moved on.

lib/commons/color/get-background-color.js
```diff
@@ -38,16 +46,45 @@ function getBgColor(elm, elmStyle) {
...
+ * @param {Element} bgNode
+ * @return {Boolean}
+ */
+function contentOverlapping(targetElement, bgNode) {
+	// get content box of target element
+	// check to see if the current bgNode is overlapping
+	var targetRect = targetElement.getClientRects()[0];
+	var obscuringElements = document.elementsFromPoint(targetRect.left, targetRect.top);
+	if (obscuringElements) {
+		for(var i = 0; i < obscuringElements.length; i++) {
```
WilcoFiers: I'd prefer we use array.find() instead, it's easier to read and harder to make bugs in (IMO). Not a big deal, just a suggestion for future updates.

lib/commons/color/get-background-color.js
```diff
@@ -57,11 +94,81 @@ function calculateObscuringAlpha(elmIndex, elmStack) {
...
+ * Include nodes missing from initial gathering because
+ * document.elementsFromPoint misses some elements we need
+ * i.e. TR is missing from table elementStack and leaves out bgColor
+ * https://github.com/dequelabs/axe-core/issues/273
+ *
+ * @private
+ * @param {Array} elmStack
+ * @param {Element} elm
+ */
+function includeMissingElements(elmStack, elm) {
```
WilcoFiers: This function relies on mutation. It's just as easy to have this function return a new array and not mutate the elmStack. Generally, we shouldn't rely on mutation if we can avoid it. Pure functions are easier to test and use.

lib/commons/color/get-background-color.js
```diff
@@ -57,11 +94,81 @@ function calculateObscuringAlpha(elmIndex, elmStack) {
...
+		}
+	}
 }
+
+/**
+ * Look at document and body elements for relevant background information
+ * @private
+ * @param {Array} elmStack
+ */
+function consultDocumentBody(elmStack) {
```
WilcoFiers: This could do with a better name, something like `sortPageBackground`. Also, as before, I think this should be a pure function. Don't mutate elmStack, just return a new array. Much easier to test and less error prone.

lib/commons/color/get-background-color.js
```diff
@@ -38,16 +46,45 @@ function getBgColor(elm, elmStyle) {
...
 
-function calculateObscuringAlpha(elmIndex, elmStack) {
+/**
+ * Determines overlap of node's content with a bgNode. Used for inline elements
+ * @private
+ * @param {Element} targetElement
+ * @param {Element} bgNode
+ * @return {Boolean}
+ */
+function contentOverlapping(targetElement, bgNode) {
```
marcysutton: The DOM order/z-index of the elements matters when you do the comparison, and I couldn't get that to work for all scenarios.

lib/commons/color/get-background-color.js
marcysutton: Ok, I'll add that change back in to build up a new array instead.

marcysutton:
> We could maybe also add a check that the foreground and background colors are the same, have it return undefined in that scenario. We discussed that case with Birkir previously. Seems like a small enough addition.

I'd prefer to do that in a separate PR. But I can pick that up next.

marcysutton:
I've addressed Wilco's feedback by not mutating elmStack within functions. I'm not in love with how many times we overwrite elmStack in a row in `color.getBackgroundStack`, but trying to do all of that filtering in a single iterator would be difficult. I looked at using `array.reduce` but I couldn't make it work with the elm passed in, since that is what you're comparing against the elmStack in `includeMissingElements`. Array.reduce takes an initial value but we can't count on the order of the elements in the array. If anyone has suggestions on how to make those filter methods more elegant (but not overly complex at the cost of readability), we could work those into a future rewrite.

marcysutton:
@dylanb do you want to review this one more time or can I merge it?

PR #332, opened 2017-05-16
Adding checks to ensure no aria-hidden is on document.body

downzer0:
Splitting this up into two smaller pieces of work. The first is #332 and the second is forthcoming.

We'll close this once those two are in PR-form.


WilcoFiers:
I'm not too sure about these two rules. I think the aria-hidden-body one is too specific. It only triggers in one very specific scenario, and any minor deviation from that (such as having it on a main container) won't be caught.

The second rule, hidden-content, is going to flag WAAAYYY to many elements to be useful. It flags anything that's hidden, giving a users thousands of elements to sift through manually, probably isn't the way to go. I'm not sure what the best way would be to approach this. I suggest you experiment with it, try it out in a whole bunch of sites and see how well it works.

PR #333, opened 2017-05-17
feat: added new rule/check for aria-hidden on body

downzer0:
# [WWD-669](https://dequesrc.atlassian.net/browse/WWD-669)

Adds a new rule/check:

1. checks for `aria-hidden` on the `body` element and fails if true,

Tests are included.


lib/checks/aria/aria-hidden-body.js
```diff
@@ -0,0 +1,11 @@
+var aria = /^aria-hidden/;
+if (node && node.hasAttributes()) {
```
marcysutton: Can you check for the attribute with `node.hasAttribute()` to avoid having to iterate?

lib/rules/aria-hidden-body.js
```diff
@@ -0,0 +1,11 @@
+var aria = /^aria-hidden/;
+if (node && node.hasAttributes()) {
```
marcysutton: I think we don't need the matches function here since the code is identical to the check. I also don't see where it checks for the body element–perhaps that could go into the `selector` field in the rule json?

downzer0:
@marcysutton Thanks. I've made some updates.

downzer0:
circleci run

marcysutton:
@downzer0 the second test is failing, I think it needs a false result instead of true.

downzer0:
@marcysutton Well, this test is checking that it returns an error if it's present. I think false is correct, no?

lib/checks/aria/aria-hidden-body.js
```diff
@@ -0,0 +1,6 @@
+var aria = 'aria-hidden';
+if (node && node.hasAttribute(aria)) {
```
dylanb: You need to ensure that the value is `true` - a value of `false`, while dumb, does not cause any problems

lib/checks/aria/aria-hidden-body.json
```diff
@@ -0,0 +1,11 @@
+{
+  "id": "aria-hidden-body",
+  "evaluate": "aria-hidden-body.js",
+  "metadata": {
+    "impact": "critical",
+    "messages": {
+      "pass": "No aria-hidden attribute is present on document body",
+      "fail": "aria-hidden should not be present on the document body"
```
dylanb: add `="true"`

lib/rules/aria-hidden-body.json
```diff
@@ -0,0 +1,19 @@
...
+  "selector": "body",
+  "tags": [
+    "cat.aria",
+    "wcag2a",
+    "wcag411",
+    "wcag412"
+  ],
+  "metadata": {
+    "description": "Ensures aria-hidden is not present on the document body.",
+    "help": "Document body should not include the aria-hidden attribute."
```
dylanb: Perhaps describe the problem this causes? e.g. using `aria-hidden="true"` makes the entire page unusable by screen readers?

downzer0:
Thanks @dylanb. Feedback addressed.

lib/rules/aria-hidden-body.json
```diff
@@ -0,0 +1,19 @@
+{
+  "id": "aria-hidden-body",
+  "selector": "body",
+  "tags": [
+    "cat.aria",
+    "wcag2a",
+    "wcag411",
```
WilcoFiers: Remove this tag. This rule has nothing to do with parsing.

lib/checks/aria/aria-hidden-body.js
```diff
@@ -0,0 +1,6 @@
+var aria = 'aria-hidden';
```
WilcoFiers: You can simplify this quite a bit:

```
return node.getAttribute('aria-hidden') === 'true';
```

marcysutton:
@downzer0 can you check on the integration tests? They are still failing.

downzer0:
@WilcoFiers Odd that these integration tests pass locally, but fail on CI. Would you mind checking to see if you get the same results?

downzer0:
@dylanb @WilcoFiers @marcysutton I've added integration tests.

lib/rules/aria-hidden-body.json
```diff
@@ -0,0 +1,18 @@
+{
+  "id": "aria-hidden-body",
+  "selector": "body",
+  "excludeHidden": false,
+  "tags": [
+    "cat.aria",
+    "wcag2a"
```
WilcoFiers: Sorry, missed that, but you're missing a tag for the wcag criterion. i suggest wcag412.

PR #334, opened 2017-05-19
fix: remove circular dep from incompleteData

marcysutton:
The incompleteData API for color-contrast was causing a circular dependency by including nodes. They weren't being used, so they're gone with this commit.


dylanb:
Don't know why this slipped through the code review last time - thanks for fixing!!

PR #335, opened 2017-05-22
feat: added new rule/check notifying for unseen or hidden content

downzer0:
# [WWD-669](https://dequesrc.atlassian.net/browse/WWD-669)

Adds a new rule/check:

1. checks for content with `display: none`, `visibility: hidden`, or `aria-hidden` and flags it for review

Tests are included.

### DIFF
--- a/doc/rule-descriptions.md
+++ b/doc/rule-descriptions.md
@@ -23,6 +23,7 @@
 | frame-title-unique | Ensures &lt;iframe&gt; and &lt;frame&gt; elements contain a unique title attribute | cat.text-alternatives, best-practice | true |
 | frame-title | Ensures &lt;iframe&gt; and &lt;frame&gt; elements contain a non-empty title attribute | cat.text-alternatives, wcag2a, wcag241, section508, section508.22.i | true |
 | heading-order | Ensures the order of headings is semantically correct | cat.semantics, best-practice | false |
+| hidden-content | Informs users about hidden content. | experimental, review-item | false |
 | href-no-hash | Ensures that href values are valid link references to promote only using anchors as links | cat.semantics, best-practice | false |
 | html-has-lang | Ensures every HTML document has a lang attribute | cat.language, wcag2a, wcag311 | true |
 | html-lang-valid | Ensures the lang attribute of the &lt;html&gt; element has a valid value | cat.language, wcag2a, wcag311 | true |
--- a/lib/checks/visibility/hidden-content.js
+++ b/lib/checks/visibility/hidden-content.js
@@ -1,8 +1,15 @@
 let styles = window.getComputedStyle(node);
 
 if (axe.commons.dom.hasContent(node)) {
-  if (styles.getPropertyValue('display') === 'none' || styles.getPropertyValue('visibility') === 'hidden') {
+  if (styles.getPropertyValue('display') === 'none') {
     return undefined;
+  } else if (styles.getPropertyValue('visibility') === 'hidden') {
+    if (node.parentNode) {
+      var parentStyle = window.getComputedStyle(node.parentNode);
+    }
+    if (!parentStyle || parentStyle.getPropertyValue('visibility') !== 'hidden') {
+      return undefined;
+    }
   }
 }
 return true;
--- a/lib/checks/visibility/hidden-content.js
+++ b/lib/checks/visibility/hidden-content.js
@@ -1,0 +1,8 @@
+let styles = window.getComputedStyle(node);
+
+if (axe.commons.dom.hasContent(node)) {
+  if (styles.getPropertyValue('display') === 'none' || styles.getPropertyValue('visibility') === 'hidden') {
+    return undefined;
+  }
+}
+return true;
--- a/lib/checks/visibility/hidden-content.json
+++ b/lib/checks/visibility/hidden-content.json
@@ -1,0 +1,12 @@
+{
+  "id": "hidden-content",
+  "evaluate": "hidden-content.js",
+  "metadata": {
+    "impact": "minor",
+    "messages": {
+      "pass": "All content on the page has been analyzed.",
+      "fail": "There were problems analyzing the content on this page.",
+      "undefined": "There is hidden content on the page that was not analyzed. You will need to trigger the display of this content in order to analyze it."
+    }
+  }
+}
--- a/lib/rules/hidden-content.json
+++ b/lib/rules/hidden-content.json
@@ -1,0 +1,18 @@
+{
+  "id": "hidden-content",
+  "selector": "*",
+  "tags": [
+    "experimental",
+    "review-item"
+  ],
+  "metadata": {
+    "description": "Informs users about hidden content.",
+    "help": "Trigger the UI state that displays the content and then run the analysis again."
+  },
+  "all": [],
+  "any": [
+    "hidden-content"
+  ],
+  "none": [],
+  "enabled": false
+}
--- a/lib/rules/hidden-content.json
+++ b/lib/rules/hidden-content.json
@@ -1,6 +1,7 @@
 {
   "id": "hidden-content",
   "selector": "*",
+  "excludeHidden": false,
   "tags": [
     "experimental",
     "review-item"
--- a/test/checks/visibility/hidden-content.js
+++ b/test/checks/visibility/hidden-content.js
@@ -27,6 +27,12 @@ describe('hidden content', function () {
 		assert.isUndefined(checks['hidden-content'].evaluate.call(checkContext, node));
 	});
 
+	it('should return true with visibility:hidden and parent with visibility:hidden', function () {
+    fixture.innerHTML = '<div style="visibility: hidden;"><p id="target" style="visibility: hidden;">Some paragraph text.</p></div>';
+		var node = fixture.querySelector('#target');
+		assert.isTrue(checks['hidden-content'].evaluate.call(checkContext, node));
+	});
+
 	it('should return true with aria-hidden and no content', function () {
 		fixture.innerHTML = '<span id="target" class="icon" aria-hidden="true"></span>';
 		var node = fixture.querySelector('#target');
--- a/test/checks/visibility/hidden-content.js
+++ b/test/checks/visibility/hidden-content.js
@@ -1,0 +1,36 @@
+describe('hidden content', function () {
+	'use strict';
+
+  var fixture = document.getElementById('fixture');
+
+	var checkContext = {
+		_data: null,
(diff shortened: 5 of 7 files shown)


doc/rule-descriptions.md
```diff
@@ -23,6 +23,7 @@
 | frame-title-unique | Ensures &lt;iframe&gt; and &lt;frame&gt; elements contain a unique title attribute | cat.text-alternatives, best-practice | true |
 | frame-title | Ensures &lt;iframe&gt; and &lt;frame&gt; elements contain a non-empty title attribute | cat.text-alternatives, wcag2a, wcag241, section508, section508.22.i | true |
 | heading-order | Ensures the order of headings is semantically correct | cat.semantics, best-practice | false |
+| hidden-content | Informs users about hidden content. |  | true |
```
marcysutton: We should tag this as experimental, and possibly also best-practice.

lib/rules/hidden-content.json
```diff
@@ -0,0 +1,14 @@
...
+  "metadata": {
+    "description": "Informs users about hidden content.",
+    "help": "Manually check elements that ar ehidden with display: none, visibility: hidden, or aria-hidden: \"true\""
+  },
+  "all": [],
+  "any": [
+    "hidden-content"
+  ],
+  "none": []
+}
```
marcysutton: We should probably disable this rule by default in axe-core and enable it in the Attest extension.

test/checks/visibility/hidden-content.js
```diff
@@ -0,0 +1,48 @@
...
+	});
+
+	it('was present on the page with visibility: hidden', function () {
+    fixture.innerHTML = '<div id="target" style="visibility: hidden;"><p>Nothing here.</p></div>';
+		var node = fixture.querySelector('#target');
+		assert.isUndefined(checks['hidden-content'].evaluate.call(checkContext, node));
+	});
+
+	it('was present on the page within a span with aria-hidden', function () {
+    fixture.innerHTML = '<span id="target" aria-hidden="true">Nothing here.</span>';
```
marcysutton: I'd love to see a test for the inline span thing you mentioned. Perhaps we could exempt child elements with `aria-hidden` inside of links or buttons if there is an accessible name otherwise?

lib/rules/hidden-content.json
```diff
@@ -0,0 +1,14 @@
+{
+  "id": "hidden-content",
+  "matches": "hidden-content.js",
+  "tags": [],
+  "metadata": {
+    "description": "Informs users about hidden content.",
+    "help": "Manually check elements that ar ehidden with display: none, visibility: hidden, or aria-hidden: \"true\""
```
marcysutton: Typo: `that ar ehidden`

lib/checks/visibility/hidden-content.json
```diff
@@ -0,0 +1,12 @@
+{
+  "id": "hidden-content",
+  "evaluate": "hidden-content.js",
+  "metadata": {
+    "impact": "minor",
+    "messages": {
+      "pass": "Page does not have any suspicious hidden content.",
+      "fail": "Manually inspect hidden content on this page.",
+      "undefined": "One or more elements on this page are hidden from assistive technology. You should verify these elements manually."
```
marcysutton: We should add a little more detail so people know what they're looking for to resolve it. Perhaps something like this? 

> One or more elements on this page are hidden from assistive technology. You should verify it was intentional for these elements with manual debugging.

lib/rules/hidden-content.json
```diff
@@ -0,0 +1,14 @@
+{
+  "id": "hidden-content",
+  "matches": "hidden-content.js",
+  "tags": [],
```
marcysutton: Let's tag this as experimental. @WilcoFiers @dylanb should we also tag it as a best-practice?

lib/rules/hidden-content.json
downzer0: @marcysutton So to disable this here, do I move the `hidden-content` to the `none` section?

test/checks/visibility/hidden-content.js
downzer0: @marcysutton I think the last test here includes hidden content within hidden content. See line 42.

lib/checks/visibility/hidden-content.json
downzer0: 👍

lib/rules/hidden-content.json
downzer0: 👍

lib/rules/hidden-content.json
marcysutton: You should be able to add `enabled: false`.

test/checks/visibility/hidden-content.js
marcysutton: I don't think that test is specific enough to the scenarios where a developer might put `aria-hidden="true"` on an icon element inside of an anchor or button. I thought we'd discussed whitelisting those to minimize noise in Needs Review, since they are valid use cases. As written that test below still returns undefined.

dylanb:
This is the first of a new type of rule. Its intent is to just flag content for manual review. This means that it need to return a "review item" (this PR does that). However, it needs to be tagged under something other than `best-practice`

I would suggest that we use a new tag called something like `review-item`

downzer0:
@dylanb @marcysutton I've modified the new rule/check and the tests. In short, I'm checking for elements with no content (typically icon fonts) and not flagging them. Anything else that's hidden gets a review flag.

Does this line up with what we were thinking?

lib/checks/visibility/hidden-content.js
```diff
@@ -0,0 +1,17 @@
+if (!axe.commons.dom.isVisible(node)) {
```
dylanb: I think the test should do the following:

1. Look for all nodes that directly have `display:none` or `visibility:hidden`
1. Make sure those nodes have descendants with real content
1. Flag only the node that is causing the hide

The intent is to flag areas on the page that are not being covered by the analysis. `aria-hidden` content does get covered by some of the analysis, so it should not be flagged.

lib/checks/visibility/hidden-content.js
```diff
@@ -0,0 +1,17 @@
...
+  // not visible on-screen or via AT
+  // probably ok, return truthy
+  return true;
+} else if (node.hasAttribute('aria-hidden')) {
```
dylanb: I don't see the usefulness of this

lib/checks/visibility/hidden-content.json
```diff
@@ -0,0 +1,12 @@
+{
+  "id": "hidden-content",
+  "evaluate": "hidden-content.js",
+  "metadata": {
+    "impact": "minor",
+    "messages": {
+      "pass": "Page does not have any suspicious hidden content.",
+      "fail": "Manually inspect hidden content on this page.",
```
dylanb: These messages should reflect the intended use:

pass: all the content on the page has been analyzed
fail: there is hidden content on the page that is not being analyzed, you will need to trigger the display of this content to analyze it

lib/rules/hidden-content.json
```diff
@@ -0,0 +1,18 @@
+{
+  "id": "hidden-content",
+  "matches": "hidden-content.js",
+  "tags": [
+    "experimental",
+    "review-item"
+  ],
+  "metadata": {
+    "description": "Informs users about hidden content.",
+    "help": "Manually check elements that are hidden with display: none, visibility: hidden, or aria-hidden: \"true\""
```
dylanb: There are two ways to deal with this

1. trigger the UI state that displays the content and then analyze, or
1. manually assess it (not sure how you would do that without first doing the above)

lib/checks/visibility/hidden-content.js
WilcoFiers: It seems to me this check is overly complicated. We don't have to check if the element is visible. You've already done that in the matcher. All this should be doing is check if there is content or not. If there is, return undefined, if not, return true.

lib/rules/hidden-content.js
```diff
@@ -0,0 +1 @@
+return axe.commons.dom.isVisible(node);
```
WilcoFiers: This should check if the element has `display:none`, `visibility:hidden` or `aria-hidden=true`.

downzer0: This only checks for `aria-hidden` if the `true` parameter is set as the second parameter. Am I understanding the docs correctly?

WilcoFiers: I'm not sure what you are asking @downzer0

downzer0:
@dylanb @WilcoFiers How's this?

lib/checks/visibility/hidden-content.js
```diff
@@ -1,17 +1,9 @@
...
-  } else {
-    // probably not an icon font, flag it for review
+let styles = window.getComputedStyle(node);
+
+if (axe.commons.dom.hasContent(node)) {
+  if (styles.getPropertyValue('display') === 'none' || styles.getPropertyValue('visibility') === 'hidden') {
     return undefined;
   }
-} else {
   return true;
```
dylanb: unnecessary extra return, can fall through to the return at the bottom

test/checks/visibility/hidden-content.js
```diff
@@ -0,0 +1,36 @@
+describe('hidden content', function () {
```
WilcoFiers: Add tests to make sure you're not selecting descendants of elements with `display:none` / `visibility:hidden`.

downzer0:
@WilcoFiers Integration tests added!

lib/checks/visibility/hidden-content.js
```diff
@@ -1,8 +1,15 @@
...
 if (axe.commons.dom.hasContent(node)) {
-  if (styles.getPropertyValue('display') === 'none' || styles.getPropertyValue('visibility') === 'hidden') {
+  if (styles.getPropertyValue('display') === 'none') {
     return undefined;
+  } else if (styles.getPropertyValue('visibility') === 'hidden') {
+    if (node.parentNode) {
+      var parentStyle = window.getComputedStyle(node.parentNode);
+    }
+    if (!parentStyle || parentStyle.getPropertyValue('visibility') !== 'hidden') {
+      return undefined;
```
dylanb: What is this for?

downzer0: Ah, where to begin...

So, `display` is NOT inherited, however `visibility` IS inherited. We needed to add an additional layer here that checks child/parents to return appropriately. I worked with Wilco on this one earlier today.

PR #338, opened 2017-05-25
feat: add fallback fallback message

marcysutton:
@dylanb here's the logic we talked about. I'll have to talk with Wilco about how to localize the fallback message.

### DIFF
--- a/lib/core/utils/publish-metadata.js
+++ b/lib/core/utils/publish-metadata.js
@@ -7,20 +7,32 @@
  * @private
  */
 function getIncompleteReason(checkData, messages) {
+	function getDefaultMsg(messages) {
+		if (messages.incomplete && messages.incomplete.default) {
+			// fall back to the default message if no reason specified
+			return messages.incomplete.default;
+		} else {
+			// TODO: localize it
+			return 'aXe couldn\'t tell the reason. Time to break out the element inspector!';
+		}
+	}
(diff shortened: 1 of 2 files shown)


lib/core/utils/publish-metadata.js
```diff
@@ -7,20 +7,32 @@
  * @private
  */
 function getIncompleteReason(checkData, messages) {
+	function getDefaultMsg(messages) {
+		if (messages.incomplete && messages.incomplete.default) {
+			// fall back to the default message if no reason specified
+			return messages.incomplete.default;
+		} else {
+			// TODO: localize it
+			return 'aXe couldn\'t tell the reason. Time to break out the element inspector!';
```
marcysutton: This string could be anything. Feel free to suggest alternatives if it's not helpful enough / too cheeky.

lib/core/utils/publish-metadata.js
```diff
@@ -7,20 +7,32 @@
...
+		}
+	}
 	if (checkData && checkData.missingData) {
-		var missingReason;
 		try {
-			missingReason = checkData.missingData[0].reason;
-		} finally {
+			var msg = messages.incomplete[checkData.missingData[0].reason];
```
marcysutton: This throws an error in the wild (enough to break our extensions) but in testing it didn't for some reason, so I threw an error just to be sure.

PR #340, opened 2017-05-25
feat: localize incomplete fallback message

marcysutton:
I added the ability to localize the fallback message. This makes it so that if a check doesn't supply a default incomplete message or reasons, aXe will still return _something_ in the appropriate language.

### DIFF
--- a/test/core/reporters/helpers/incomplete-fallback-msg.js
+++ b/test/core/reporters/helpers/incomplete-fallback-msg.js
@@ -1,0 +1,21 @@
+
+describe('helpers.incompleteFallbackMessage', function() {
+	'use strict';
+	before(function() {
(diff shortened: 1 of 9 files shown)


test/core/reporters/helpers/incomplete-fallback-msg.js
```diff
@@ -0,0 +1,21 @@
...
+			rules: [],
+			data: {
+				incompleteFallbackMessage: function anonymous() {
+					return 'Dogs are the best';
+				}
+			}
+		});
+	});
+
+	it('should return a string', function() {
```
marcysutton: This test is sorta useless since we're not injecting data or transforming the string, but still nice to have.

PR #343, opened 2017-06-05
Rebuild getSelector

WilcoFiers:
Rebuild the getSelector method, to return selectors that are more readable, shorter and more durable. In addition, the absolutePaths option was added which will return full paths to elements, rather then the shortest possible selectors we could find.

### DIFF
--- a/lib/core/utils/dq-element.js
+++ b/lib/core/utils/dq-element.js
@@ -1,8 +1,6 @@
 /*exported DqElement */
 
 function truncate(str, maxLength) {
-	'use strict';
-
 	maxLength = maxLength || 300;
 
 	if (str.length > maxLength) {
@@ -14,8 +12,6 @@ function truncate(str, maxLength) {
 }
 
 function getSource (element) {
-	'use strict';
-
 	var source = element.outerHTML;
 	if (!source && typeof XMLSerializer === 'function') {
 		source = new XMLSerializer().serializeToString(element);
@@ -29,12 +25,13 @@ function getSource (element) {
  * @param {HTMLElement} element The element to serialize
  * @param {Object} spec Properties to use in place of the element when instantiated on Elements from other frames
  */
-function DqElement(element, spec) {
-	'use strict';
-	
+function DqElement(element, options, spec) {
 	this._fromFrame = !!spec;
 
 	this.spec = spec || {};
+	if (options && options.absolutePaths) {
+		this._options = { toRoot: true };
+	}
 
 	/**
 	 * The generated HTML source code of the element
@@ -56,7 +53,9 @@ DqElement.prototype = {
 	 * @return {String}
 	 */
 	get selector() {
-		return this.spec.selector || [axe.utils.getSelector(this.element)];
+		return this.spec.selector || [
+			axe.utils.getSelector(this.element, this._options)
+		];
 	},
 
 	/**
@@ -64,7 +63,9 @@ DqElement.prototype = {
 	 * @return {String}
 	 */
 	get xpath() {
-		return this.spec.xpath || [axe.utils.getXpath(this.element)];
+		return this.spec.xpath || [
+			axe.utils.getXpath(this.element)
+		];
 	},
 	
 	/**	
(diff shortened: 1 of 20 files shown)


lib/core/utils/dq-element.js
```diff
@@ -1,8 +1,6 @@
 /*exported DqElement */
 
 function truncate(str, maxLength) {
-	'use strict';
```
dylanb: Why this deletion?

lib/core/utils/get-selector.js
```diff
@@ -1,104 +1,218 @@
...
-	}
-
-	return index;
+const escapeSelector = axe.utils.escapeSelector;
+
+function isUncommonClassName (className) {
+  return ![
+    'focus', 'hover',
+    'hidden', 'visible',
+    'dirty', 'touched', 'valid', 'disable',
```
dylanb: duplicate `disable`

lib/core/utils/get-selector.js
```diff
@@ -1,104 +1,218 @@
...
+      });
+    }
+  }, Array.from(elm.classList).filter(isUncommonClassName));
+}
 
-	if (!siblings) {
-		return false;
-	}
+const commonNodes = [
+  'div', 'span', 'p',
```
dylanb: How did you choose these?

lib/core/utils/get-selector.js
```diff
@@ -1,104 +1,218 @@
...
-	var length = siblings.length;
+const createSelector = {
+  // Get ID properties
+  getElmId (elm) {
+  	if (!elm.id) {
+  		return;
+  	}
+  	const id = '#' + escapeSelector(elm.id || '');
+    if (
+    	// Don't include youtube's uid values, they change  on reload
```
dylanb: Isn't this a bit specific? Is this for embedded videos?

lib/core/utils/get-selector.js
```diff
@@ -1,104 +1,218 @@
...
-		if (element.nodeName.toUpperCase() === type) {
-			index++;
-		}
-		element = element.previousElementSibling;
-	}
-
-	return index;
+const escapeSelector = axe.utils.escapeSelector;
+
+function isUncommonClassName (className) {
```
dylanb: Why not look for how many other elements on the page have the class?

lib/core/utils/get-selector.js
```diff
@@ -1,104 +1,218 @@
...
+    distinctClassList: getDistinctClassList(elm)
+  };
+
+  return [
+    // go through feature selectors in order of priority
+    createSelector.getCustomElm,
+    createSelector.getElmRoleProp,
+    createSelector.getUncommonElm,
+    createSelector.getElmNameProp,
+    createSelector.getDistinctClass,
```
dylanb: Should we demote classes even further? They often get added after events like focus.

lib/core/utils/dq-element.js
WilcoFiers: Babel puts us in strict mode everywhere. We don't need to do this anymore. I've been gradually removing them from functions I'm working on.

lib/core/utils/get-selector.js
WilcoFiers: I **think** this will yield better results. I was thinking of things like the "col-md-12" stuff from bootstrap, or "sr-only". Those might be used in different places on the page. It seemed to me that having that info is better than having the nodeName. It massively depends ofcourse.

lib/core/utils/get-selector.js
WilcoFiers: Bit of experimentation, bit of guess work based on experience.

lib/core/utils/get-selector.js
WilcoFiers: Yeah, it kinda is. I put it in because YouTube videos are very common, and our video rules will always put them incomplete. If we don't do this, reviewing video's becomes kind of pointless because we can't identify that something is the same video. I'm not in favor of adding as many of these cases as we can find, but this one seemed worth it to me.

lib/core/utils/get-selector.js
WilcoFiers: There are filters which look for classes that may indicate state. I'm not sure how effective this is in practice. Classes can be pretty descriptive, so I'd like to see if we can get this to work. Probably the best test for this is how well this will perform when we record scripts. If we find that this isn't working well enough we can bump it down.

dylanb:
merging this now because the shadow DOM code needs this

PR #346, opened 2017-06-07
Jest example

WilcoFiers:
Small example of how to use aXe on a React component, using Jest, Enzyme and JSDOM.

### DIFF
--- a/doc/examples/jest+react/link.test.js
+++ b/doc/examples/jest+react/link.test.js
@@ -1,0 +1,18 @@
+import React from 'react';
+import axe from 'axe-core';
+import { mountToDoc } from './test-helpers';
+
+import Link from './Link';
+
+test('Link has no aXe violations', (done) => {
+  const linkComponent = mountToDoc(
+    <Link page="http://www.axe-core.org">aXe website</Link>
+  );
+  const linkNode = linkComponent.getDOMNode();
+
+  axe.run(linkNode, (err, { violations }) => {
+    expect(err).toBe(null);
+    expect(violations).toHaveLength(0);
+    done();
+  });
+});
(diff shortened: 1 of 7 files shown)


doc/examples/jest+react/link.test.js
```diff
@@ -0,0 +1,18 @@
...
+
+import Link from './Link';
+
+test('Link has no aXe violations', (done) => {
+  const linkComponent = mountToDoc(
+    <Link page="http://www.axe-core.org">aXe website</Link>
+  );
+  const linkNode = linkComponent.getDOMNode();
+
+  axe.run(linkNode, (err, { violations }) => {
```
marcysutton: Shouldn't we disable color-contrast since it doesn't work on JSDOM?

marcysutton:
Some test failures on this one that we should fix before committing.

doc/examples/jest+react/link.test.js
WilcoFiers: axe drops those into incomplete I believe. But we can make it explicit if you think that's better?

marcysutton: It used to just error out because of the lack of support for `document.createRange` and `elementsFromPoint`. If that got fixed, it's probably fine.

WilcoFiers: We fixed that in 2.2. That's one of the reasons we build `incomplete`, so we don't have to throw abort when there's an error. We can put it in incomplete and move on.

marcysutton:
I added this to develop too since it defaults to that branch on Github. We should base PRs on develop first to prevent having to cherry-pick or merge the wrong direction.

Issue #349, opened 2017-06-12
Allow axe.source to run in strict mode


WilcoFiers:
When running webdriver selenium in Firefox, `executeScript()` will run in script mode, thus `this` no longer refers to the window object. Because of that, the following won't work:

```js
axe.source = '(' + axeFunction.toString() + ')(this, this.document);';
```

dylanb:
@WilcoFiers What version of Firefox does this break in?

marcysutton:
Fixed in #355.

PR #355, opened 2017-06-14
fix: Update axe.source to work with Firefox webdriver

WilcoFiers:
axe-webdriverjs wasn’t running properly because axe.source wasn’t getting the window object passed in correctly.

### DIFF
--- a/lib/core/index.js
+++ b/lib/core/index.js
@@ -11,7 +11,7 @@ if (typeof define === 'function' && define.amd) {
 	});
 }
 if (typeof module === 'object' && module.exports && typeof axeFunction.toString === 'function') {
-    axe.source = '(' + axeFunction.toString() + ')(this, this.document);';
+    axe.source = '(' + axeFunction.toString() + ')(typeof window === "object" ? window : this);';
     module.exports = axe;
 }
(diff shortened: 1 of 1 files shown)


WilcoFiers:
This PR resolves https://github.com/dequelabs/axe-core/issues/349

marcysutton:
I copied this change into axe-webdriverjs and ran the tests, and it successfully gets rid of an error about `window.axe` not being defined. It took updating to Firefox 54 and geckodriver v0.17.0 to get it to work properly.

lib/core/index.js
```diff
@@ -11,7 +11,7 @@ if (typeof define === 'function' && define.amd) {
 	});
 }
 if (typeof module === 'object' && module.exports && typeof axeFunction.toString === 'function') {
-    axe.source = '(' + axeFunction.toString() + ')(this, this.document);';
+    axe.source = '(' + axeFunction.toString() + ')(typeof window === "object" ? window : this);';
```
dylanb: hey, aren't we supposed to be passing in two parameters here?

WilcoFiers:
We did, but that looked like a mistake to me, as the function it is calling
only takes one argument. It takes the document from the window object.

Issue #348, opened 2017-06-12
deduplicate language options


WilcoFiers:
In axe.js, the same lengthy array with all possible languages is included three times, needlessly adding 140k to axe-core.

PR #357, opened 2017-06-14
feat: moves html lang options from check to commons

downzer0:
The aXe rule/check for valid HTML `lang` attributes had an options array that was being included three times not only making the `axe.js` file three times larger than it needed to be, but also made other apps that consume aXe at least that much larger as well.

I've moved the options array from the check to an `axe.commons.utils` file that returns the same and updated the check JavaScript itself to use `options` if present, otherwise fallback to the returned array of every possible language code.

Closes https://github.com/dequelabs/axe-core/issues/348.


downzer0:
@dylanb @iandotkelly @marcysutton @WilcoFiers I'm working on tests, atm, but heads-up!

downzer0:
@dylanb @iandotkelly @marcysutton @WilcoFiers Ready for your review.

lib/commons/utils/valid-langs.js
```diff
@@ -0,0 +1,8123 @@
+/*global axe */
+
+axe.utils.validLangs = function () {
+	'use strict';
```
marcysutton: I don't think we need `use strict` as Babel adds it for us...as an FYI

test/checks/language/valid-lang.js
```diff
@@ -43,13 +43,12 @@ describe('valid-lang', function () {
 			assert.deepEqual(checkContext._data, ['lang="en-FOO"']);
 		});
 
-		it('should return true (and not throw) when given no options', function () {
+		it('should return false (and not throw) when given no options', function () {
```
marcysutton: How come the return values changed here?

downzer0: It's related to the `none` in the check. It's a bit backwards. Previously, there were `options` or `[]`, but now we've got `(options ? options : axe.commons.utils.validLangs)` - so either `null` or some, but never empty.

This is my best explanation. It's kind of hard to wrap my head around the tests that are true, that return false :)

lib/commons/utils/valid-langs.js
downzer0: Some of the other utils have it, but happy to remove if you advise.

test/checks/language/valid-lang.js
dylanb: Add a test that tests to make sure it returns true if passed an empty options array

dylanb:
@downzer0 - as mentioned in the call - the list of languages is built by the `grunt langs` command, so we need to make sure that build task still works with this new implementation

test/checks/language/valid-lang.js
downzer0: @dylanb doesn't the test on line 37 do that?

downzer0:
@dylanb Everything seems to build OK, and when I manually bring in the compiled `axe.js` file into our rulebuilder, things work there too, in addition to the filesize being 2/3 smaller.

Should we remove the Grunt task?

dylanb:
No, the grunt task does not get called automatically, it is there to build the list should the languages change - now admittedly, this happens very infrequently, but calling it periodically will automatically update for changes whereas maintaining that manually is almost impossible due to the size.

So the Grunt task should be changed to generate the code in the commons from a template (or something equivalent)

downzer0:
@dylanb I've made an update which alters the Gruntfile location to the new util file.

I originally had this generate a JSON file and then tried `require`-ing it in the check, but require is undefined. I opted to just re-build the entire check file with the languages array. I had to disable the JSHint double/single quote check for this file because JSON.stringify makes valid JSON (which has double-quotes) but our rules want single.

I've tested the build process and tested the output axe.js in our rulebuilder.

marcysutton:
So this task for running langs...we should document it better. There is a mention of `valid-lang` in [doc/developer-guide.md](https://github.com/dequelabs/axe-core/blob/develop/doc/developer-guide.md) we need to update. 

I'm also wondering if we should register a task at the bottom of the Gruntfile to make this more obvious.

downzer0:
@marcysutton Good catch, with the documentation - I'll get it updated. My only hesitation with the registered task is that, based on @dylanb's comment, it doesn't sound like this is used much. Also, you _can_ run it with `grunt langs`, it just isn't exposed as easily. I am happy to expose it, but will defer to you and Dylan on this.

marcysutton:
I didn't even know it was there, since there is another variable called `langs`. It is super confusing, so we should do something to document it at least.

downzer0:
Just looked - the docs don't need an update. I will get the Grunt task documented though.

downzer0:
We will document the Grunt tasks in a separate PR. Marcy is creating an issue for it.

marcysutton:
Looks good! Circle just needed a rebuild.

PR #359, opened 2017-06-15
Hidden content rule fixes

marcysutton:
There were a couple of issues with the new hidden content rules when I pulled them into the extensions.

### DIFF
--- a/lib/checks/visibility/hidden-content.js
+++ b/lib/checks/visibility/hidden-content.js
@@ -1,0 +1,15 @@
+let styles = window.getComputedStyle(node);
+
+if (axe.commons.dom.hasContent(node)) {
+  if (styles.getPropertyValue('display') === 'none') {
+    return undefined;
+  } else if (styles.getPropertyValue('visibility') === 'hidden') {
+    if (node.parentNode) {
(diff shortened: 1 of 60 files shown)


WilcoFiers:
Good point, although, who puts aria-hidden on a script tag.

lib/checks/visibility/hidden-content.js
```diff
@@ -1,6 +1,7 @@
 let styles = window.getComputedStyle(node);
 
-if (axe.commons.dom.hasContent(node)) {
+let whitelist = ['SCRIPT', 'HEAD', 'TITLE', 'NOSCRIPT', 'STYLE'];
+if (!whitelist.includes(node.tagName) && axe.commons.dom.hasContent(node)) {
```
WilcoFiers: tagName should be cast to uppercase to have it work with xhtml docs.

downzer0:
What if you change the selector in the rule to `body *`?

lib/checks/visibility/hidden-content.js
dylanb: Add template to this list

marcysutton:
I wanted to include HTML in case somebody did something really bad.

downzer0:
👍

PR #362, opened 2017-06-15
feat: updated rule and check for hidden content to omit head, templat…

downzer0:
Updates a previously added and merged rule that checks for hidden content, by omitting `head`, `template`, and `script`. Additional tests were added for this update.

### DIFF
--- a/lib/checks/visibility/hidden-content.js
+++ b/lib/checks/visibility/hidden-content.js
@@ -1,6 +1,8 @@
 let styles = window.getComputedStyle(node);
+let nodeName = node.nodeName.toUpperCase();
+let skipOver = ['HEAD', 'TEMPLATE', 'SCRIPT'];
 
-if (axe.commons.dom.hasContent(node)) {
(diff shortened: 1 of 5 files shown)


downzer0:
@marcysutton has this covered in https://github.com/dequelabs/axe-core/pull/359. Closing.

lib/checks/visibility/hidden-content.js
```diff
@@ -1,6 +1,8 @@
 let styles = window.getComputedStyle(node);
+let nodeName = node.nodeName.toUpperCase();
+let skipOver = ['HEAD', 'TEMPLATE', 'SCRIPT'];
```
dylanb: add noscript to this list

WilcoFiers: I don't think noscript should be there. That's one of those elements that could actually become visible under certain conditions.

`style` on the other hand should be on this list.

marcysutton: This PR wasn't the one that made it in. Style is in the list, but noscript is. We can remove it later.

PR #370, opened 2017-06-19
feat: Add option restoreScroll

WilcoFiers:
With this setting, axe will remember the scroll position of all elements that can scroll, and restore them once analyse is done.

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -311,6 +311,7 @@ Additionally, there are a number or properties that allow configuration of diffe
 | `absolutePaths` | `false` | Use absolute paths when creating element selectors
 | `iframes`       | `true`  | Tell axe to run inside iframes
 | `elementRef`    | `false` | Return element references in addition to the target
+| `restoreScroll` | `false` | Scrolls elements back to before axe started
 
 
 ###### Options Parameter Examples
--- a/lib/core/public/run-rules.js
+++ b/lib/core/public/run-rules.js
@@ -29,7 +29,14 @@ function runRules(context, options, resolve, reject) {
 		});
 	}
 	q.defer(function (res, rej) {
-		audit.run(context, options, res, rej);
+		if (options.restoreScroll) {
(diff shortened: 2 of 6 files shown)


doc/API.md
```diff
@@ -311,6 +311,7 @@ Additionally, there are a number or properties that allow configuration of diffe
 | `absolutePaths` | `false` | Use absolute paths when creating element selectors
 | `iframes`       | `true`  | Tell axe to run inside iframes
 | `elementRef`    | `false` | Return element references in addition to the target
+| `restoreScroll` | `false` | Scrolls elements back to before axe started
```
marcysutton: Why not default to `true`? It would make implementing it in our extensions much easier.

dylanb: I agree

WilcoFiers: I want to avoid that we have to recommend that you set this to false if you're using axe in a test environment. I didn't do performance tests on it, so I'm not sure if the performance impact is significant enough that we'd have to do that. But if it isn't, if you can safely run this 1000 times in your application and not have it be noticeable, than we should just do this and not have it be optional at all.

Thoughts?

dylanb: Run some performance tests and then lets decide based on that

WilcoFiers: @dylanb It looks like it adds +/- 20ms to the execution when using it in dqpl-react. At most the impact is 15%, which is the case for very small components. For larger once this obviously goes down as test time goes up.

dylanb: Can you run some comparisons on large sites like amazon.com pages or ebay.com pages?

WilcoFiers: With axe-cli, up to 200ms or so, on a 20 - 25 second run. So yeah, this add something like 1% to 15% to execution time.

WilcoFiers: @dylanb @marcysutton Based on previous discussion, I think we concluded to leave it on false (as it is now). Please approve if you think this is ok.

PR #373, opened 2017-06-20
fix offScreen test failing in Chrome

WilcoFiers:
- Errors got swallowed by the browser testrunner. That’s fixed now
- Chrome was failing one of the tests due to not cleaning up scroll position

### DIFF
--- a/build/tasks/test-webdriver.js
+++ b/build/tasks/test-webdriver.js
@@ -35,12 +35,12 @@ module.exports = function (grunt) {
 	/**
 	 * Test each URL
 	 */
-	function runTestUrls(driver, urls) {
+	function runTestUrls(driver, urls, errors) {
 		var url = urls.shift();
-		var errors = [];
+		errors = errors || [];
 
 		// Give each page enough time
-		driver.manage().timeouts().setScriptTimeout(600);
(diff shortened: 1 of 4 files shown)


build/tasks/test-webdriver.js
```diff
@@ -71,6 +71,7 @@ module.exports = function (grunt) {
 			if (urls.length > 0) {
 				return runTestUrls(driver, urls, errors);
 			} else {
+				driver.quit();
 				return Promise.resolve(errors);
```
dylanb: shouldn't you call resolve inside a then after the quite?

```return driver.quit().then(function () {
  Promise.resolve(errors);
});
```

marcysutton:
I think these test timeouts could use your expertise @dylanb.

Also let's be sure to squash this PR when it gets merged since it has commits that don't adhere to our policy.

dylanb:
ok, I fixed the tests - took two changes

The approach I took to syncing the iframe loading could be generalized to all iframe tests by creating a utility function that we include in all iframes that allows us to tell when the iframe has loaded.

WilcoFiers:
Awesome!! Thanks Dylan. That's why you're the iframe king :P

PR #374, opened 2017-06-20
fix implicit attribute use in duplicate-id

WilcoFiers:
The id property can be overloaded, so we should use getAttribute()

### DIFF
--- a/lib/checks/aria/required-parent.js
+++ b/lib/checks/aria/required-parent.js
@@ -34,11 +34,12 @@ function getAriaOwners(element) {
 		o = null;
 
 	while (element) {
-		if (element.id) {
-			o = document.querySelector('[aria-owns~=' + axe.commons.utils.escapeSelector(element.id) + ']');
+		if (element.getAttribute('id')) {
+			const id = axe.commons.utils.escapeSelector(element.getAttribute('id'));
+			o = document.querySelector(`[aria-owns~=${id}]`);
 			if (o) { owners.push(o); }
 		}
-		element = element.parentNode;
+		element = element.parentElement;
 	}
 
 	return owners.length ? owners : null;
--- a/lib/checks/label/explicit.js
+++ b/lib/checks/label/explicit.js
@@ -1,5 +1,7 @@
-if (node.id) {
-	var label = document.querySelector('label[for="' + axe.commons.utils.escapeSelector(node.id) + '"]');
+if (node.getAttribute('id')) {
+  const id = axe.commons.utils.escapeSelector(node.getAttribute('id'));
+	const label = document.querySelector(`label[for="${id}"]`);
+
 	if (label) {
 		return !!axe.commons.text.accessibleText(label);
 	}
--- a/lib/checks/label/explicit.js
+++ b/lib/checks/label/explicit.js
@@ -1,5 +1,5 @@
 if (node.getAttribute('id')) {
-  const id = axe.commons.utils.escapeSelector(node.getAttribute('id'));
+	const id = axe.commons.utils.escapeSelector(node.getAttribute('id'));
 	const label = document.querySelector(`label[for="${id}"]`);
 
 	if (label) {
--- a/lib/checks/label/multiple-label.js
+++ b/lib/checks/label/multiple-label.js
@@ -1,6 +1,6 @@
-var labels = [].slice.call(document.querySelectorAll('label[for="' +
-	axe.commons.utils.escapeSelector(node.id) + '"]')),
-	parent = node.parentNode;
+const id = axe.commons.utils.escapeSelector(node.getAttribute('id'));
+let labels = Array.from(document.querySelectorAll(`label[for="${id}"]`));
+let parent = node.parentNode;
 
 if (labels.length) {
 	// filter out hidden labels because they're fine
--- a/lib/checks/tables/th-has-data-cells.js
+++ b/lib/checks/tables/th-has-data-cells.js
@@ -30,10 +30,8 @@ var tableGrid = tableUtils.toGrid(node);
 
 // Look for all the bad headers
 var out = headers.reduce(function (res, header) {
-	if (
-		header.getAttribute('id') &&
-		reffedHeaders.includes(header.getAttribute('id'))
-	) {
+	if (header.getAttribute('id') &&
+			reffedHeaders.includes(header.getAttribute('id'))) {
 		return (!res ? res : true);
 	}
 
--- a/lib/checks/tables/th-has-data-cells.js
+++ b/lib/checks/tables/th-has-data-cells.js
@@ -30,7 +30,10 @@ var tableGrid = tableUtils.toGrid(node);
 
 // Look for all the bad headers
 var out = headers.reduce(function (res, header) {
-	if (header.id && reffedHeaders.indexOf(header.id) !== -1) {
+	if (
+		header.getAttribute('id') &&
+		reffedHeaders.includes(header.getAttribute('id'))
+	) {
 		return (!res ? res : true);
 	}
 
--- a/lib/commons/table/is-header.js
+++ b/lib/commons/table/is-header.js
@@ -10,8 +10,9 @@ table.isHeader = function (cell) {
 		return true;
 	}
 
-	if (cell.id) {
-		return !!document.querySelector('[headers~="' + axe.utils.escapeSelector(cell.id) + '"]');
+	if (cell.getAttribute('id')) {
+    const id = axe.utils.escapeSelector(cell.getAttribute('id'));
+		return !!document.querySelector(`[headers~="${id}"]`);
 	}
 
 	return false;
--- a/lib/commons/table/is-header.js
+++ b/lib/commons/table/is-header.js
@@ -11,7 +11,7 @@ table.isHeader = function (cell) {
 	}
 
 	if (cell.getAttribute('id')) {
-    const id = axe.utils.escapeSelector(cell.getAttribute('id'));
+		const id = axe.utils.escapeSelector(cell.getAttribute('id'));
 		return !!document.querySelector(`[headers~="${id}"]`);
 	}
 
--- a/lib/commons/text/accessible-text.js
+++ b/lib/commons/text/accessible-text.js
@@ -21,8 +21,9 @@ var phrasingElements = ['A', 'EM', 'STRONG', 'SMALL', 'MARK', 'ABBR', 'DFN', 'I'
  */
 function findLabel(element) {
 	var ref = null;
-	if (element.id) {
-		ref = document.querySelector('label[for="' + axe.utils.escapeSelector(element.id) + '"]');
+	if (element.getAttribute('id')) {
+		const id = axe.utils.escapeSelector(element.getAttribute('id'));
+		ref = document.querySelector(`label[for="${id}"]`);
 		if (ref) {
 			return ref;
 		}
--- a/lib/core/utils/get-selector.js
+++ b/lib/core/utils/get-selector.js
@@ -49,10 +49,10 @@ function getNthChildString (elm, selector) {
 const createSelector = {
   // Get ID properties
   getElmId (elm) {
-  	if (!elm.id) {
+  	if (!elm.getAttribute('id')) {
   		return;
   	}
-  	const id = '#' + escapeSelector(elm.id || '');
+  	const id = '#' + escapeSelector(elm.getAttribute('id') || '');
     if (
     	// Don't include youtube's uid values, they change  on reload
     	!id.match(/player_uid_/) &&
@@ -88,7 +88,7 @@ const createSelector = {
   },
   // Has a name property, but no ID (Think input fields)
   getElmNameProp (elm) {
-    if (!elm.id && elm.name) {
+    if (!elm.hasAttribute('id') && elm.name) {
       return '[name="' + escapeSelector(elm.name) + '"]';
     }
   },
--- a/lib/core/utils/get-xpath.js
+++ b/lib/core/utils/get-xpath.js
@@ -50,7 +50,7 @@ function getXPathArray(node, path) {
 		var element = {};
 		element.str = node.nodeName.toLowerCase();
 		// add the id and the count so we can construct robust versions of the xpath
-		var id = node.getAttribute && axe.utils.escapeSelector(node.getAttribute('id'))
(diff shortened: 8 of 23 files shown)


dylanb:
@WilcoFiers have you tested which of these is actually used by `getElementById`?

test/checks/shared/duplicate-id.js
```diff
@@ -49,4 +49,11 @@ describe('duplicate-id', function () {
 		assert.isTrue(checks['duplicate-id'].evaluate.call(checkContext, node));
 	});
 
+	it('should allow overwrote ids', function () {
```
dylanb: Where is the overwritten ID?

WilcoFiers:
@dylanb I'm not sure what you're asking for... Do you mean to check if `getElementById` uses the `.id` property instead of looking at the `id=""` attribute?

dylanb:
@WilcoFiers no, what I mean is, when the ID gets overridden, what it returns is not a string, we could test to make sure that behavior does not change. In your test you could add one more assertion to test for that.

dylanb:
ok, I deleted the confusing comments as replies so I could clarify:

1) This only happens on form elements
2) When it happens, the id attribute changes from a string to a reference to an HTMLElement
3) There are 39 places in the code where we refer to the element.id - this changes only one of them

I am not sure whether this is documented browser behavior or a side-effect? Can you find a reference?

If it is documented, we might be better off detecting it and having a rule that raises this as a violation rather than trying to fix this everywhere.

lib/checks/aria/required-parent.js
```diff
@@ -34,11 +34,15 @@ function getAriaOwners(element) {
 		o = null;
 
 	while (element) {
-		if (element.id) {
-			o = document.querySelector('[aria-owns~=' + axe.commons.utils.escapeSelector(element.id) + ']');
+		if (element.getAttribute('id')) {
+			o = document.querySelector(
+				'[aria-owns~=' + 
+				axe.commons.utils.escapeSelector(element.getAttribute('id')) +
```
marcysutton: The line breaks don't make this much easier to read. Perhaps the escapeSelector line should be indented to indicate a hierarchy?

WilcoFiers:
@dylanb @marcysutton Any further comments?

marcysutton:
Looks good to me!

dylanb:
Don't merge

--Dylan

WilcoFiers:
@dylanb Don't merge, because ... ?

lib/checks/forms/labelledby.js
```diff
@@ -4,7 +4,11 @@ this.data({
 });
 
 var matchingNodes = document.querySelectorAll('input[type="' +
-	axe.commons.utils.escapeSelector(node.type) + '"][name="' + axe.commons.utils.escapeSelector(node.name) + '"]');
+	axe.commons.utils.escapeSelector(node.getAttribute('type')) +
```
dylanb: Why these changes?

lib/checks/navigation/unique-frame-title.js
```diff
@@ -1,3 +1,5 @@
-var title = axe.commons.text.sanitize(node.title).trim().toLowerCase();
+var title = axe.commons.text.sanitize(
```
dylanb: Why?

lib/commons/dom/is-in-text-block.js
```diff
@@ -75,7 +75,7 @@ dom.isInTextBlock = function isInTextBlock(node) {
 			return false;
 
 		// Don't walk links, we're only interested in what's not in them.
-		} else if ((nodeName === 'A' && currNode.href) ||
+		} else if ((nodeName === 'A' && currNode.hasAttribute('href')) ||
```
dylanb: This is not the same - don't change this

lib/commons/text/accessible-text.js
```diff
@@ -32,21 +33,23 @@ function findLabel(element) {
 }
 
 function isButton(element) {
-	return ['button', 'reset', 'submit'].indexOf(element.type) !== -1;
+	return ['button', 'reset', 'submit'].includes(
+		(element.getAttribute('type') || '').toLowerCase()
```
dylanb: Don't change this

lib/commons/text/accessible-text.js
```diff
@@ -32,21 +33,23 @@ function findLabel(element) {
...
+	);
 }
 
 function isInput(element) {
 	var nodeName = element.nodeName.toUpperCase();
 	return (nodeName === 'TEXTAREA' || nodeName === 'SELECT') ||
-		(nodeName === 'INPUT' && element.type.toLowerCase() !== 'hidden');
+		(nodeName === 'INPUT' && (element.getAttribute('type') || '').toLowerCase() !== 'hidden');
```
dylanb: don't change this

lib/commons/text/accessible-text.js
```diff
@@ -111,16 +114,16 @@ function isEmbeddedControl(e) {
...
 		default:
 			return false;
 	}
 }
 
 function shouldCheckAlt(element) {
 	var nodeName = element.nodeName.toUpperCase();
-	return (nodeName === 'INPUT' && element.type.toLowerCase() === 'image') ||
-		['IMG', 'APPLET', 'AREA'].indexOf(nodeName) !== -1;
+	return (nodeName === 'INPUT' && (element.getAttribute('type') || '').toLowerCase() === 'image') ||
```
dylanb: don't change this

lib/commons/text/accessible-text.js
```diff
@@ -199,7 +202,12 @@ text.accessibleText = function(element, inLabelledByContext) {
 
 		if (isInput(element) && !inControlContext) {
 			if (isButton(element)) {
-				return element.value || element.title || defaultButtonValues[element.type] || '';
+				return (
```
dylanb: don't change

lib/core/utils/get-selector.js
```diff
@@ -81,15 +81,15 @@ const createSelector = {
     	nodeName = escapeSelector(nodeName);
       // Add [type] if nodeName is an input element
       if (nodeName === 'input' && elm.hasAttribute('type')) {
-        nodeName += '[type="' + elm.type + '"]';
+        nodeName += `[type="${ elm.getAttribute('type') }"]`;
```
dylanb: don't change

lib/core/utils/get-selector.js
```diff
@@ -81,15 +81,15 @@ const createSelector = {
...
       }
       return nodeName;
     }
   },
   // Has a name property, but no ID (Think input fields)
   getElmNameProp (elm) {
-    if (!elm.id && elm.name) {
-      return '[name="' + escapeSelector(elm.name) + '"]';
+    if (!elm.getAttribute('id') && elm.getAttribute('name')) {
```
dylanb: don't change

lib/rules/color-contrast-matches.js
```diff
@@ -1,7 +1,7 @@
 /* global document */
 
 var nodeName = node.nodeName.toUpperCase(),
-	nodeType = node.type,
+	nodeType = node.getAttribute('type'),
```
dylanb: don't change

lib/rules/color-contrast-matches.js
```diff
@@ -13,7 +13,7 @@ if (nodeName === 'INPUT') {
 }
 
 if (nodeName === 'SELECT') {
-	return !!node.options.length && !node.disabled;
+	return !!node.options.length && !node.hasAttribute('disabled');
```
dylanb: don't change

lib/rules/color-contrast-matches.js
```diff
@@ -24,11 +24,11 @@ if (nodeName === 'OPTION') {
 	return false;
 }
 
-if (nodeName === 'BUTTON' && node.disabled || axe.commons.dom.findUp(node, 'button[disabled]')) {
+if (nodeName === 'BUTTON' && node.hasAttribute('disabled') || axe.commons.dom.findUp(node, 'button[disabled]')) {
```
dylanb: don't change

lib/rules/color-contrast-matches.js
```diff
@@ -24,11 +24,11 @@ if (nodeName === 'OPTION') {
...
 	return false;
 }
 
-if (nodeName === 'FIELDSET' && node.disabled || axe.commons.dom.findUp(node, 'fieldset[disabled]')) {
+if (nodeName === 'FIELDSET' && node.hasAttribute('disabled') || axe.commons.dom.findUp(node, 'fieldset[disabled]')) {
```
dylanb: don't change

lib/rules/color-contrast-matches.js
```diff
@@ -40,23 +40,25 @@ if (nodeName === 'LABEL' || nodeParentLabel) {
 		relevantNode = nodeParentLabel;
 	}
 	// explicit label of disabled input
-	var candidate = relevantNode.htmlFor && doc.getElementById(relevantNode.htmlFor);
-	if (candidate && candidate.disabled) {
+	var candidate = relevantNode.hasAttribute('for') && doc.getElementById(relevantNode.getAttribute('for'));
```
dylanb: don't change

dylanb:
@WilcoFiers because I had not had time to review it yet...now that I have, I am totally confused - I thought that this change was specifically about the `id` attribute getting overridden within a `<form>` element - why did you change a bunch of other attributes?

WilcoFiers:
@dylanb The problem isn't limited to IDs. From what I found, any property on html, body, and form elements can be replaced with an accessor to a different element. You can even overwrite things like `getAttribute` and `parentNode` if you want. This stuff is part of DOM 0, but all browsers I tested still support it: https://openhome.cc/eGossip/JavaScript/Level0DOM.html

It's fortunately limited to a few elements, but because we're often using role selectors, we can't ever be entirely sure that element properties aren't what we think. You could even replace things like `parentNode` if you wanted. Some of these are less likely to be used then others as element names. I think we should avoid accessing properties that are likely to be used as field names, things like `name`, `title`, `id`, `type` and `disabled` (as in having a disability).

Probably `.htmlFor` and `.href` are secure. I could remove those, but I would recommend we keep the rest in. I could do some performance tests if you want, I'm sure there's a cost to this.

dylanb:
@WilcoFiers have we had complaints from users that this is occurring? I found 39 uses of simply `.id` within the code base. The changes you have made so far address only a fraction of even these cases. So if we are going to make these changes, we need to make them everywhere.

That being said, I don't see the need to make these changes if we have not had complaints because that means that this bad use of HTML is not occurring frequently enough to warrant the required changes.

Also, not all attributes return the same value as the call to `getAttribute()` - `href` is one example. This means that we need to not simply replace the uses, we need to inspect the code for each use and make sure that the implementation is not relying on those differences.

Bottom line: I don't think this is worth it.

WilcoFiers:
@dylanb fair enough. I'll role some stuff back. As for replacing all instances of `.id`, most of them come from rules and checks, not element references. I went through all of them.

lib/checks/label/explicit.js
```diff
@@ -1,5 +1,7 @@
-if (node.id) {
-	var label = document.querySelector('label[for="' + axe.commons.utils.escapeSelector(node.id) + '"]');
+if (node.getAttribute('id')) {
+  const id = axe.commons.utils.escapeSelector(node.getAttribute('id'));
+	const label = document.querySelector(`label[for="${id}"]`);
```
marcysutton: Some mixed indentation here.

lib/checks/label/multiple-label.js
```diff
@@ -1,6 +1,6 @@
-var labels = [].slice.call(document.querySelectorAll('label[for="' +
-	axe.commons.utils.escapeSelector(node.id) + '"]')),
-	parent = node.parentNode;
+const id = axe.commons.utils.escapeSelector(node.getAttribute('id'));
+let labels = Array.from(document.querySelectorAll(`label[for="${id}"]`));
```
marcysutton: `Array.from` here is so slick. I love it!

lib/checks/tables/th-has-data-cells.js
```diff
@@ -30,7 +30,10 @@ var tableGrid = tableUtils.toGrid(node);
 
 // Look for all the bad headers
 var out = headers.reduce(function (res, header) {
-	if (header.id && reffedHeaders.indexOf(header.id) !== -1) {
+	if (
```
marcysutton: Does this code fit on one line under 80 characters? It would be more consistent with our existing style.

lib/commons/table/is-header.js
```diff
@@ -10,8 +10,9 @@ table.isHeader = function (cell) {
 		return true;
 	}
 
-	if (cell.id) {
-		return !!document.querySelector('[headers~="' + axe.utils.escapeSelector(cell.id) + '"]');
+	if (cell.getAttribute('id')) {
+    const id = axe.utils.escapeSelector(cell.getAttribute('id'));
```
marcysutton: Indentation is mixed up here due to spaces/tabs, I'm guessing.

lib/core/utils/get-xpath.js
```diff
@@ -50,9 +50,10 @@ function getXPathArray(node, path) {
 		var element = {};
 		element.str = node.nodeName.toLowerCase();
 		// add the id and the count so we can construct robust versions of the xpath
-		if(node.getAttribute && node.getAttribute('id') &&
-			node.ownerDocument.querySelectorAll('#' + axe.utils.escapeSelector(node.id)).length === 1) {
-
+		if(
```
marcysutton: Can we put the first part of the expression inline with the `if(` for consistency's sake?

test/checks/shared/duplicate-id.js
```diff
@@ -49,4 +49,11 @@ describe('duplicate-id', function () {
...
+		fixture.innerHTML = '<form data-testelm="1" id="target"><input name="id"></form>';
```
marcysutton: This might be overkill for test code...but as stewards of accessibility it would be appropriate to have a label in the fixture.

marcysutton:
I left a few style comments, but otherwise this PR looks good to me.

lib/checks/tables/th-has-data-cells.js
dylanb: I agree - lets keep our coding style consistent

dylanb:
@WilcoFiers only one remaining jshint problem, then we can merge

PR #376, opened 2017-06-22
fix: Properly output error stack

WilcoFiers:
Small fix to ensure the stack is always exposed when a check throws an exception. This should make it much easier for us to track down problems people find in axe using the extensions.

### DIFF
--- a/lib/core/base/audit.js
+++ b/lib/core/base/audit.js
@@ -165,7 +165,7 @@ Audit.prototype.run = function (context, options, resolve, reject) {
 							result: axe.constants.CANTTELL,
 							description: 'An error occured while running this rule',
 							message: err.message,
-							help: err.stack || err.message,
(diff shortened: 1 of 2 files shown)


lib/core/base/audit.js
```diff
@@ -165,7 +165,7 @@ Audit.prototype.run = function (context, options, resolve, reject) {
 							result: axe.constants.CANTTELL,
 							description: 'An error occured while running this rule',
 							message: err.message,
-							help: err.stack || err.message,
```
marcysutton: Does anything depend on `help` being a part of this object, or was that just for manual debugging info?

WilcoFiers: No. The reason I deleted this is because this actually didn't work. This `help` property is actually overwritten later, so this line did nothing.

PR #378, opened 2017-06-23
fix: corrects color-contrast tests with overlapping elements

downzer0:
I believe this corrects the color-contrast check that tests for overlapping and obscuring elements. The previous code was checking for elements only at a specific set of coordinates, not within the bounding box. So, for example, if there was nothing at `0,0` then it would pass, even if there was overlap elsewhere.

Supporting documentation: https://stackoverflow.com/a/12067046

One nit change: I updated `elm` to `node` to make it more clear what is coming in (the node being checked). Happy to revert this. No additional changes were done.

### DIFF
--- a/lib/commons/color/get-background-color.js
+++ b/lib/commons/color/get-background-color.js
@@ -6,18 +6,18 @@ const graphicNodes = [
 /**
  * Reports if an element has a background image or gradient
  * @private
- * @param {Element} elm
+ * @param {Element} node
  * @param {Object|null} style
  * @return {Boolean}
  */
-function elmHasImage(elm, style) {
-	var nodeName = elm.nodeName.toUpperCase();
+function elmHasImage(node, style) {
+	var nodeName = node.nodeName.toUpperCase();
 	if (graphicNodes.includes(nodeName)) {
 		axe.commons.color.incompleteData.set('bgColor', 'imgNode');
 		return true;
 	}
 
-	style = style || window.getComputedStyle(elm);
+	style = style || window.getComputedStyle(node);
 	var bgImageStyle = style.getPropertyValue('background-image');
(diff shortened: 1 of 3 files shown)


lib/commons/color/get-background-color.js
```diff
@@ -53,17 +53,17 @@ function getBgColor(elm, elmStyle) {
  * @param {Element} bgNode
  * @return {Boolean}
  */
-function contentOverlapping(targetElement, bgNode) {
+function contentOverlapping(targetElement, fgNode) {
 	// get content box of target element
-	// check to see if the current bgNode is overlapping
-	var targetRect = targetElement.getClientRects()[0];
```
marcysutton: FYI: I used `getClientRects` because it would target the diff in coverage for an inline label wrapping an input, as opposed to the outer bounding box of the label. I guess with this change, the `fgNode` is the input instead of the label in that case?

downzer0: The `fgNode` in this case would be the foreground "whatever". The function receives two element parameters - one that's the "background" and one that's potentially overlapping (generally a child element). The only way you'd have overlapping would be with absolute positioning, but you'd still need a parent/child combo, right? Are there any scenarios I'm not covering here?

marcysutton: When the input was under audit and its parent was an implicit, inline label, the label would be skipped and not considered a background unless it was added to the stack manually. (also the label itself as a node, which contains text content). I couldn't guarantee which element was in the front since it was evaluating both the child and the parent for overlapping text content. If you look at the examples in this MDN doc, you can see `getClientRects` can be used to check inline content overlap. https://developer.mozilla.org/en-US/docs/Web/API/Element/getClientRects

But if this works, awesome. I want to run it through the aXe extension to be sure.

Issue #306, opened 2017-04-17
Radio buttons in table flagged as not being part of group


gerardkcohen:
The following example is flagged with [Rule Id: Radiogroup](https://dequeuniversity.com/rules/axe/1.1/radiogroup)

Ex:

```
        <table>
            <caption>Select a Product</caption>
            <thead>
                <tr>
                    <th scope="col">Select</th>
                    <th scope="col">Product</th>
                    <th scope="col">Description</th>
                    <th scope="col">Cost</th>
                </tr>
            </thead>
            <tbody>
                <tr>
                    <th scope="row">Basic Web Server</th>
                    <td>
                        <label>Select Basic
                            <input type="radio" name="product" value="basic" />
                        </label>
                    </td>
                    <td>Basic server to host your basic site</td>
                    <td>$199/yr</td>
                </tr>
                <tr>
                    <th scope="row">Awesome Web Server</th>
                    <td>
                        <label>Select Awesome
                            <input type="radio" name="product" value="awesome" />
                        </label>
                    </td>
                    <td>Awesome server to host your awesome site</td>
                    <td>$299/yr</td>
                </tr>
                <tr>
                    <th scope="row">SSL Cert</th>
                    <td>
                        <label>Select SSL Cert
                            <input type="radio" name="product" value="ssl" />
                        </label>
                    </td>
                    <td>Make sure your site is secure</td>
                    <td>$99/yr</td>
                </tr>
            </tbody>
        </table>

```
This should be valid use case, even though not wrapped in an explicit grouping/ form role. If not, a suggestion on how to remediate would be good.

Thanks!

SiTaggart:
Ooooh yeah, I have exactly the same question / issue! +1 on this

marcysutton:
@WilcoFiers @dylanb did you have any ideas for how to solve this? Adding `aria-labelledby` to the INPUT elements removes the violation, but it also stomps over the label text already provided. Perhaps we could allow `aria-describedby` to make the association with a table header? The table caption is announced in Voiceover for the first input, but not the rest. It isn't announced in JAWS or NVDA, at least in the versions I tested.

WilcoFiers:
Different origin, but #384 should address this too.

dylanb:
@marcysutton if you look at the example that originated this ticket, the table has a caption that is the group label and the radio buttons are the only inputs in the table.

This would be a case we could definitely pass.

Other cases might be:
1. Common `<th>` with text and the correct `scope`
[...]

PR #384, opened 2017-06-29
Allow tables to group form fields

WilcoFiers:


lib/checks/forms/fieldset.js
```diff
@@ -50,28 +50,62 @@ function checkARIAGroup(group, name) {
 	return true;
 }
 
+function checkTable (group, nodes) {
```
dylanb: Shouldn't we also check that the table is a data table and has headers for the cells that contain the inputs?

Do we require a table marked up in this fashion to have a `<caption>`?

WilcoFiers: I thought about that, but decided against it. Our `isDataTable` method is fairly conservative, only flagging things as such if it's absolutely certain. If we fail this test every time that method isn't certain about it, we risk false positives. That's why instead of doing that, I just checked if `role=presentation` is on it.

As for headers, I think it's fairly typical for a column of checkboxes to not have a header. For example, the issues table in Comply only has an empty th above the column of checkboxes.

dylanb: Yeah those checkboxes don't qualify as a group. Only ones that have the same name attribute. If they do have the same name attribute they need a group label. That is what this rule is all about. So we should only pass checkbox groups that actually have some form of label. The question is how do we identify that within a table?

lib/checks/forms/fieldset.js
```diff
@@ -50,28 +50,62 @@ function checkARIAGroup(group, name) {
...
+		'[role="group"]' + (node.type === 'radio' ? ',[role="radiogroup"]' : '')
+	);
+	const table = axe.commons.dom.findUp(element, 'table');
+		
+	if (fieldset) {
+		return checkFieldset(fieldset, name);
+	} else if (group) {
+		return checkARIAGroup(group, name);
+	} else if (table) {
+		return checkTable(table, matchingNodes);
```
marcysutton: Should we rename this since it isn't strictly checking fieldsets anymore?

lib/checks/forms/fieldset.js
WilcoFiers: We could check if they share a table heading? That's what I was originally thinking of doing.

dylanb: @WilcoFiers take a look at my last comment on this ticket https://github.com/dequelabs/axe-core/issues/306

Issue #386, opened 2017-06-29
duplicate-img-label check - Shadow DOM support


marcysutton:
Failing tests for the [duplicate-img-label check](https://github.com/dequelabs/axe-core/blob/develop/lib/checks/label/duplicate-img-label.js):
```
    1) should return false if no text is present
    2) should return false if aria-label duplicates img alt
    3) should return false if img and text have different text
    4) should return true if img and text have the same text
    5) should return true if img has ARIA label with the same text
    6) should return false if img and text are both blank
    7) should return false if img and text have superset/subset text
```

Support can be added by going through each of these tests and incorporating Shadow DOM utils, as a subtask of #87.

Full stack trace:
```
1) duplicate-img-label should return false if no text is present:
     
  visible@http://localhost:9876/axe.js:9427:31
  evaluate@http://localhost:9876/tmp/rules.js:1303:39
  http://localhost:9876/test/checks/label/duplicate-img-label.js:19:56
  callFn@http://localhost:9876/node_modules/mocha/mocha.js:4447:25
  run@http://localhost:9876/node_modules/mocha/mocha.js:4439:13
  runTest@http://localhost:9876/node_modules/mocha/mocha.js:4936:13
  http://localhost:9876/node_modules/mocha/mocha.js:5042:19
  next@http://localhost:9876/node_modules/mocha/mocha.js:4853:16
  http://localhost:9876/node_modules/mocha/mocha.js:4863:11
  next@http://localhost:9876/node_modules/mocha/mocha.js:4787:16
  http://localhost:9876/node_modules/mocha/mocha.js:4831:9
  timeslice@http://localhost:9876/node_modules/mocha/mocha.js:82:27

  2) duplicate-img-label should return false if aria-label duplicates img alt:
     
  visible@http://localhost:9876/axe.js:9427:31
  evaluate@http://localhost:9876/tmp/rules.js:1303:39
  http://localhost:9876/test/checks/label/duplicate-img-label.js:25:56
  callFn@http://localhost:9876/node_modules/mocha/mocha.js:4447:25
  run@http://localhost:9876/node_modules/mocha/mocha.js:4439:13
  runTest@http://localhost:9876/node_modules/mocha/mocha.js:4936:13
  http://localhost:9876/node_modules/mocha/mocha.js:5042:19
  next@http://localhost:9876/node_modules/mocha/mocha.js:4853:16
  http://localhost:9876/node_modules/mocha/mocha.js:4863:11
[...]

PR #402, opened 2017-06-30
fix: use virtualNode in duplicate-img-label

marcysutton:
Closes https://github.com/dequelabs/axe-core/issues/386


marcysutton:
The tests for this check are now passing:
```
  duplicate-img-label
    ✓ should return false if no img is present
    ✓ should return false if no text is present
    ✓ should return false if aria-label duplicates img alt
    ✓ should return false if img and text have different text
    ✓ should return true if img and text have the same text
    ✓ should return true if img has ARIA label with the same text
    ✓ should return false if img and text are both blank
    ✓ should return false if img and text have superset/subset text
```

Issue #388, opened 2017-06-29
title-only check - Shadow DOM support


marcysutton:
Failing tests for the [title-only check](https://github.com/dequelabs/axe-core/blob/develop/lib/checks/label/title-only.js):
```
    12) should return true if an element only has a title
    13) should return true if an element only has aria-describedby
```

Sub-task of #87. 

Full stack trace: 
```
12) title-only should return true if an element only has a title:
     
  label@http://localhost:9876/axe.js:8120:28
  label@http://localhost:9876/axe.js:9397:31
  evaluate@http://localhost:9876/tmp/rules.js:1378:42
  http://localhost:9876/test/checks/label/title-only.js:17:46
  callFn@http://localhost:9876/node_modules/mocha/mocha.js:4447:25
  run@http://localhost:9876/node_modules/mocha/mocha.js:4439:13
  runTest@http://localhost:9876/node_modules/mocha/mocha.js:4936:13
  http://localhost:9876/node_modules/mocha/mocha.js:5042:19
  next@http://localhost:9876/node_modules/mocha/mocha.js:4853:16
  http://localhost:9876/node_modules/mocha/mocha.js:4863:11
  next@http://localhost:9876/node_modules/mocha/mocha.js:4787:16
  http://localhost:9876/node_modules/mocha/mocha.js:4831:9
  timeslice@http://localhost:9876/node_modules/mocha/mocha.js:82:27

  13) title-only should return true if an element only has aria-describedby:
     
  label@http://localhost:9876/axe.js:8120:28
  label@http://localhost:9876/axe.js:9397:31
  evaluate@http://localhost:9876/tmp/rules.js:1378:42
  http://localhost:9876/test/checks/label/title-only.js:33:46
  callFn@http://localhost:9876/node_modules/mocha/mocha.js:4447:25
  run@http://localhost:9876/node_modules/mocha/mocha.js:4439:13
  runTest@http://localhost:9876/node_modules/mocha/mocha.js:4936:13
  http://localhost:9876/node_modules/mocha/mocha.js:5042:19
  next@http://localhost:9876/node_modules/mocha/mocha.js:4853:16
  http://localhost:9876/node_modules/mocha/mocha.js:4863:11
  next@http://localhost:9876/node_modules/mocha/mocha.js:4787:16
  http://localhost:9876/node_modules/mocha/mocha.js:4831:9
  timeslice@http://localhost:9876/node_modules/mocha/mocha.js:82:27
```

PR #403, opened 2017-06-30
fix: use virtualNode in title-only check

marcysutton:
Closes https://github.com/dequelabs/axe-core/issues/388

### DIFF
--- a/lib/checks/label/title-only.js
+++ b/lib/checks/label/title-only.js
@@ -1,2 +1,2 @@
-var labelText = axe.commons.text.label(node);
+var labelText = axe.commons.text.label(virtualNode);
 return !labelText && !!(node.getAttribute('title') || node.getAttribute('aria-describedby'));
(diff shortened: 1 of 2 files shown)


marcysutton:
Passing tests for `title-only`:

```
title-only
    ✓ should return true if an element only has a title
    ✓ should return true if an element only has aria-describedby
```

WilcoFiers:
@marcysutton @dylanb Shouldn't we add shadow DOM tests for these rules?

dylanb:
@WilcoFiers I thought about that and I came to the conclusion that we do not need tests for every rule - however I may be wrong... here is my thinking...

If the utility functions work properly (i.e. they are tested properly) and if the rule has a good level of tests for all of its edge cases, then adding shadow DOM tests does not add more quality unless the rule has some specific functionality itself that pertains to shadow DOM.

This does not mean that we cannot add shadow DOM test, it just means we don't necessarily need them - thoughts?

Issue #392, opened 2017-06-29
hidden-content check - Shadow DOM support


marcysutton:
Failing tests for the [hidden-content check](https://github.com/dequelabs/axe-core/blob/develop/lib/checks/visibility/hidden-content.js):

```
    35) should return undefined with display:none and children
    36) should return undefined with visibility:hidden and children
    37) should return true with visibility:hidden and parent with visibility:hidden
    38) should return true with aria-hidden and no content
```

Subtask of #87.

Full stack trace:
```
35) hidden content should return undefined with display:none and children:
     
  hasContent@http://localhost:9876/axe.js:8651:27
  evaluate@http://localhost:9876/tmp/rules.js:2076:85
  http://localhost:9876/test/checks/visibility/hidden-content.js:21:60
  callFn@http://localhost:9876/node_modules/mocha/mocha.js:4447:25
  run@http://localhost:9876/node_modules/mocha/mocha.js:4439:13
  runTest@http://localhost:9876/node_modules/mocha/mocha.js:4936:13
  http://localhost:9876/node_modules/mocha/mocha.js:5042:19
  next@http://localhost:9876/node_modules/mocha/mocha.js:4853:16
  http://localhost:9876/node_modules/mocha/mocha.js:4863:11
  next@http://localhost:9876/node_modules/mocha/mocha.js:4787:16
  http://localhost:9876/node_modules/mocha/mocha.js:4831:9
  timeslice@http://localhost:9876/node_modules/mocha/mocha.js:82:27

  36) hidden content should return undefined with visibility:hidden and children:
     
  hasContent@http://localhost:9876/axe.js:8651:27
  evaluate@http://localhost:9876/tmp/rules.js:2076:85
  http://localhost:9876/test/checks/visibility/hidden-content.js:27:60
  callFn@http://localhost:9876/node_modules/mocha/mocha.js:4447:25
  run@http://localhost:9876/node_modules/mocha/mocha.js:4439:13
  runTest@http://localhost:9876/node_modules/mocha/mocha.js:4936:13
  http://localhost:9876/node_modules/mocha/mocha.js:5042:19
  next@http://localhost:9876/node_modules/mocha/mocha.js:4853:16
  http://localhost:9876/node_modules/mocha/mocha.js:4863:11
  next@http://localhost:9876/node_modules/mocha/mocha.js:4787:16
  http://localhost:9876/node_modules/mocha/mocha.js:4831:9
[...]

Issue #395, opened 2017-06-29
hidden-content integration tests - Shadow DOM support


marcysutton:
Failing tests for the [hidden-content rule integration](https://github.com/dequelabs/axe-core/blob/develop/test/integration/rules/hidden-content/hidden-content.html):

```
passes
        35) should find ["#fixture"]
        36) should find ["#pass1"]
        37) should find ["#pass2"]
        38) should find ["#pass3"]
        39) should find ["#pass4"]
        40) should find ["#pass5"]
        41) should not return other results
      incomplete
        42) should find ["#canttell1"]
        43) should find ["#canttell2"]
```

Subtask of #87.

Full test result:
```
35) hidden-content hidden-content test passes should find ["#fixture"]:
     there are no passes
  

  36) hidden-content hidden-content test passes should find ["#pass1"]:
     there are no passes
  

  37) hidden-content hidden-content test passes should find ["#pass2"]:
     there are no passes
  

  38) hidden-content hidden-content test passes should find ["#pass3"]:
     there are no passes
  

  39) hidden-content hidden-content test passes should find ["#pass4"]:
     there are no passes
  

  40) hidden-content hidden-content test passes should find ["#pass5"]:
     there are no passes
  

  41) hidden-content hidden-content test passes should not return other results:
     there are no passes
  

  42) hidden-content hidden-content test incomplete should find ["#canttell1"]:
     Element not found
  

  43) hidden-content hidden-content test incomplete should find ["#canttell2"]:
     Element not found
```

PR #404, opened 2017-07-03
Fix: Use virtualNode in hidden-content rule

WilcoFiers:


WilcoFiers:
Should resolve #392 and #395.

test/checks/visibility/hidden-content.js
```diff
@@ -10,37 +10,44 @@ describe('hidden content', function () {
 		}
 	};
 
+	function checkSetup (html, options, target) {
+		fixture.innerHTML = html;
+    axe._tree = axe.utils.getFlattenedTree(fixture);
```
dylanb: indendation

test/checks/visibility/hidden-content.js
```diff
@@ -10,37 +10,44 @@ describe('hidden content', function () {
...
 		fixture.innerHTML = '';
 		checkContext._data = null;
+		axe._tree = undefined;
 	});
 
 	it('should return undefined with display:none and children', function () {
-    fixture.innerHTML = '<div id="target" style="display: none;"><p>Some paragraph text.</p></div>';
-		var node = fixture.querySelector('#target');
-		assert.isUndefined(checks['hidden-content'].evaluate.call(checkContext, node));
+    var params = checkSetup('<div id="target" style="display: none;"><p>Some paragraph text.</p></div>');
```
dylanb: indentation

dylanb: There are a bunch more of these below, not going to comment all of them

dylanb:
This check needs some actual shadow DOM support built in. It checks the parentNode. This means it needs to do:

```
	parent = (element.assignedSlot) ? element.assignedSlot : element.parentNode;
	if (parent.nodeType === 11) {
		parent = parent.host;
	}
```

Which we may want to generalize into a commons function so we can use it elsewhere

You will also want to add shadow DOM tests for this check.

test/checks/visibility/hidden-content.js
WilcoFiers: Annoying... this file had space/tabs mixed up when I started, and the editor didn't help :/

WilcoFiers:
@marcysutton @dylanb I've made a bunch of changes. Please review once more.

lib/checks/visibility/hidden-content.js
```diff
@@ -1,13 +1,16 @@
-let styles = window.getComputedStyle(node);
+const whitelist = ['SCRIPT', 'HEAD', 'TITLE', 'NOSCRIPT', 'STYLE', 'TEMPLATE'];
+if (
```
dylanb: Why this weird style? No need to introduce new styles for if statements.

WilcoFiers: This is a weird style? I've been writing multi-line if's like this since forever. The line was getting a bit long so I wrapped it.

dylanb: I notice that this style is used elsewhere in the code - but not consistently. I would prefer we stick to the original coding style for if statements and only break lines if they go above 80 characters - and also keep braces and parentheses on the same lines

dylanb: There are currently only 5 files that use this coding style - lets change them all back

marcysutton: I agree that one line is more readable, as long as it doesn't run over 80-100 chars.

lib/commons/dom/get-composed-parent.js
```diff
@@ -0,0 +1,19 @@
...
+    return element.assignedSlot; // content of a shadow DOM slot
+  } else if (element.parentNode) {
+    var parentNode = element.parentNode;
+    if (parentNode.nodeType === 1) {
+      return parentNode; // Regular node
+    } else if (parentNode.host) {
+      return parentNode.host; // Shadow root
+    }
+  }
+  return null; // Root node
```
marcysutton: Thanks for all the comments! They're helpful. 👍

test/checks/visibility/hidden-content.js
```diff
@@ -1,46 +1,81 @@
...
+	var fixture = document.getElementById('fixture');
+	var shadowSupport = document.body && typeof document.body.attachShadow === 'function';
 	var checkContext = {
 		_data: null,
 		data: function (d) {
 			this._data = d;
 		}
 	};
 
+	function checkSetup (html, options, target) {
```
marcysutton: Should we abstract this into a reusable utility so it doesn't have to get repeated in every test file needing Shadow DOM support?

test/checks/visibility/hidden-content.js
```diff
@@ -1,46 +1,81 @@
...
 
 	it('should skip whitelisted elements', function () {
 		var node = document.querySelector('head');
-		assert.isTrue(checks['hidden-content'].evaluate.call(checkContext, node));
+		axe._tree = axe.utils.getFlattenedTree(document.documentElement);
+		var virtualNode = axe.utils.getNodeFromTree(axe._tree[0], node);
+		assert.isTrue(checks['hidden-content'].evaluate(node, undefined, virtualNode));
+	});
+
+	(shadowSupport ? it : xit)('works on elements in a shadow DOM', function () {
```
marcysutton: Crafty 👍

test/commons/dom/get-composed-parent.js
```diff
@@ -0,0 +1,72 @@
+/* global xit */
+describe('dom.getComposedParent', function () {
+  'use strict';
+  var getComposedParent = axe.commons.dom.getComposedParent;
+  var fixture = document.getElementById('fixture');
+  var shadowSupport = document.body && typeof document.body.attachShadow === 'function';
```
marcysutton: I know it's short, but this line seems like something we could abstract into a common test utility.

dylanb: This line is only shadow DOM v1, so it should only be used when we are sure that we don't need tests for shadow DOM v0

WilcoFiers: @dylanb I'm not sure what you're suggesting. With Shadow DOM v0 I take it you mean the `<content>` element and `createShadowRoot()` method? Those are deprecated now, right? I hadn't considered it. I can see that we might need to write specific tests for our utilities to cover those cases, but I'm not sure our checks need them. Are you suggesting we do?

test/checks/visibility/hidden-content.js
WilcoFiers: Not sure. I like having some test utilities to work with, but I don't much like just dropping a bunch of extra functions on the global object. I suppose we could create a `/test/utils` dir, and have a global `testUtils` object that has everything on it. Something like that?

lib/checks/visibility/hidden-content.js
WilcoFiers: I would like us to stick with an 80 char line length. My setup enlarged screen doesn't fit more than that on it. I also don't much like reading longer lines. They hare hard to scan. Anyway, I fixed the line, adding double indent.

We could consider something like Prettier, if we want to get a cleaner, more consistent format.

test/checks/visibility/hidden-content.js
marcysutton: Yes. Something we could import would be helpful–then if it needed additional code, we could change it in one place.

test/commons/dom/get-composed-parent.js
marcysutton: The fact that there are different APIs for v0 and v1 make SD pretty difficult to work with. I'd be curious to hear how much test coverage we need for both.

dylanb: I think the general approach should be that if we are relying on utils and commons that have been tested with both versions, then the tests for checks and rules themselves do not need both

PR #409, opened 2017-07-06
Align impact levels with Deque Way

WilcoFiers:
Thanks to the work of Matt Isner, I had a nice list to work from. I’ve double checked that for each impact I changed, it only changes the impact level of the rules I wanted to change. The changes are as follow:

| Rule			| Impact from		| Impact to
|-----------------------|-----------------------|---------------
| accesskeys		| critical		| serious
| blink			| minor			| serious
| marquee		| minor			| serious
| document-title	| moderate		| serious
| duplicate-id		| critical		| moderate
| heading-order		| minor			| moderate
| link-in-text-b	| critical		| serious
| listitem		| critical		| serious
| p-as-heading		| critical		| serious
| scope-attr-val	| critical, serious	| critical, moderate
| server-side-ima	| minor			| critical
| skip-link		| critical		| moderate
| table-duplicate-name	| moderate		| minor
| table-fake-caption	| critical		| serious
| th-has-data-cells	| critical		| serious
| video-description	| serious		| critical
| bypass		| critical		| serious
| empty-heading		| moderate		| minor
| frame-title		| critical		| serious
| link-name		| critical		| serious
| object-alt		| critical		| serious

### DIFF
--- a/lib/checks/color/link-in-text-block.json
+++ b/lib/checks/color/link-in-text-block.json
@@ -2,7 +2,7 @@
   "id": "link-in-text-block",
   "evaluate": "link-in-text-block.js",
   "metadata": {
-    "impact": "critical",
+    "impact": "serious",
     "messages": {
       "pass": "Links can be distinguished from surrounding text in a way that does not rely on color",
       "fail": "Links can not be distinguished from surrounding text in a way that does not rely on color",
--- a/lib/checks/keyboard/accesskeys.json
+++ b/lib/checks/keyboard/accesskeys.json
@@ -3,7 +3,7 @@
   "evaluate": "accesskeys.js",
   "after": "accesskeys-after.js",
   "metadata": {
-    "impact": "critical",
+    "impact": "serious",
     "messages": {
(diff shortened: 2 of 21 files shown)


WilcoFiers:
We might want to hold off on releasing this until aXe 3.0, as it might be a impactful change for some users.

marcysutton:
How did you double-check the impact for rules vs. checks (that are potentially reused in multiple rules)? I'm wondering if we should write any tests around that to make it less of a manual process.

WilcoFiers:
@marcysutton frustratingly enough, just a regular file search, look at all the rules that have that particular check, then look at the other checks they have and make sure one of those has the impact level we want. With 'any', it'll always take the highest impact level, so just one of them has to have 'critical' for example. for the rule to turn out with critical.

dylanb:
color-contrast went from serious to serious?

I don't see a huge reason to delay this release...perhaps users could comment here if they think it will be disruptive

WilcoFiers:
The color-contrast hasn't changed. That one had been updated a few weeks back. I think we can merge this.

Issue #390, opened 2017-06-29
td-has-header - Shadow DOM support


marcysutton:
Failing tests for the [td-has-header check](https://github.com/dequelabs/axe-core/blob/develop/lib/checks/tables/td-has-header.js):
```
    17) should not be fooled by rowspan and colspan
    18) should return true each non-empty cell has a row header
    19) should return true each non-empty cell has a column header
    20) should return true each non-empty cell has aria-label
    21) should return true each non-empty cell has aria-labelledby
    22) should return true each non-empty cell has a headers attribute
    23) should return true there is at least one non-empty header
    24) should return true if the only data cells are empty
    25) should return false if a cell has no headers
    26) should return false if a cell has no headers - complex table
    27) should return false if the headers element is empty
    28) should return false if the headers element refers to non-existing elements
    29) should return false if all headers are empty
```

Subtask of #87. 

Full stack trace:
```
17) td-has-header should not be fooled by rowspan and colspan:
     
  hasContent@http://localhost:9876/axe.js:8651:27
  http://localhost:9876/tmp/rules.js:1938:35
  forEach@[native code]
  evaluate@http://localhost:9876/tmp/rules.js:1936:17
  http://localhost:9876/test/checks/tables/td-has-header.js:48:53
  callFn@http://localhost:9876/node_modules/mocha/mocha.js:4447:25
  run@http://localhost:9876/node_modules/mocha/mocha.js:4439:13
  runTest@http://localhost:9876/node_modules/mocha/mocha.js:4936:13
  http://localhost:9876/node_modules/mocha/mocha.js:5042:19
  next@http://localhost:9876/node_modules/mocha/mocha.js:4853:16
  http://localhost:9876/node_modules/mocha/mocha.js:4863:11
  next@http://localhost:9876/node_modules/mocha/mocha.js:4787:16
  http://localhost:9876/node_modules/mocha/mocha.js:4831:9
  timeslice@http://localhost:9876/node_modules/mocha/mocha.js:82:27

  18) td-has-header should return true each non-empty cell has a row header:
     
  hasContent@http://localhost:9876/axe.js:8651:27
  http://localhost:9876/tmp/rules.js:1938:35
  forEach@[native code]
  evaluate@http://localhost:9876/tmp/rules.js:1936:17
  http://localhost:9876/test/checks/tables/td-has-header.js:61:54
  callFn@http://localhost:9876/node_modules/mocha/mocha.js:4447:25
[...]

Issue #391, opened 2017-06-29
th-has-data-cells - Shadow DOM support


marcysutton:
Failing tests for the [th-has-data-cells check](https://github.com/dequelabs/axe-core/blob/develop/lib/checks/tables/th-has-data-cells.js):
```
    30) should return true each row header has a non-empty cell
    31) should return true each non-empty column header has a cell
    ✓ should return true if referred to with headers attr
    ✓ should return true if referred to with aria-labelledby
    ✓ should return true if the th element is empty
    32) should return true when the td has a content element
    33) should return undefined if a th has no data cells
    34) should return undefined if all data cells are empty
```

Subtask of #87. 

Full stack trace:
```
30) th-has-data-cells should return true each row header has a non-empty cell:
     
  hasContent@http://localhost:9876/axe.js:8651:27
  http://localhost:9876/tmp/rules.js:2059:47
  reduce@[native code]
  http://localhost:9876/tmp/rules.js:2058:67
  reduce@[native code]
  evaluate@http://localhost:9876/tmp/rules.js:2041:28
  http://localhost:9876/test/checks/tables/th-has-data-cells.js:30:58
  callFn@http://localhost:9876/node_modules/mocha/mocha.js:4447:25
  run@http://localhost:9876/node_modules/mocha/mocha.js:4439:13
  runTest@http://localhost:9876/node_modules/mocha/mocha.js:4936:13
  http://localhost:9876/node_modules/mocha/mocha.js:5042:19
  next@http://localhost:9876/node_modules/mocha/mocha.js:4853:16
  http://localhost:9876/node_modules/mocha/mocha.js:4863:11
  next@http://localhost:9876/node_modules/mocha/mocha.js:4787:16
  http://localhost:9876/node_modules/mocha/mocha.js:4831:9
  timeslice@http://localhost:9876/node_modules/mocha/mocha.js:82:27

  31) th-has-data-cells should return true each non-empty column header has a cell:
     
  hasContent@http://localhost:9876/axe.js:8651:27
  http://localhost:9876/tmp/rules.js:2052:47
  reduce@[native code]
  http://localhost:9876/tmp/rules.js:2051:66
  reduce@[native code]
  evaluate@http://localhost:9876/tmp/rules.js:2041:28
  http://localhost:9876/test/checks/tables/th-has-data-cells.js:41:58
  callFn@http://localhost:9876/node_modules/mocha/mocha.js:4447:25
[...]

PR #411, opened 2017-07-07
shadowdom table rules

WilcoFiers:


WilcoFiers:
This PR should close #390 and #391.

Issue #393, opened 2017-06-29
color-contrast matches - Shadow DOM support


marcysutton:
Failing tests for the [color-contrast matches file](https://github.com/dequelabs/axe-core/blob/develop/lib/rules/color-contrast-matches.js): 
```
    1) should match when there is text
    2) should match when there is text that is in the scroll reach of container
    3) should match when there is text that is only partially out of the container
    4) should match <button>
```

Subtask of #87.

Full test result:
```
1) color-contrast-matches should match when there is text:
     expected false to be true
  

  2) color-contrast-matches should match when there is text that is in the scroll reach of container:
     expected false to be true
  

  3) color-contrast-matches should match when there is text that is only partially out of the container:
     expected false to be true
  

  4) color-contrast-matches should match <button>:
     expected false to be true
```

PR #413, opened 2017-07-07
Shadow DOM support for color contrast matches 

marcysutton:
This PR builds on #404 to pull in some of the commons already in use.

Closes https://github.com/dequelabs/axe-core/issues/393.

### DIFF
--- a/lib/rules/color-contrast-matches.js
+++ b/lib/rules/color-contrast-matches.js
@@ -61,7 +61,7 @@ if (node.id) {
 	}
 }
 
-if (axe.commons.text.visible(node, false, true) === '') {
+if (axe.commons.text.visible(virtualNode, false, true) === '') {
 	return false;
 }
 
--- a/lib/rules/color-contrast-matches.js
+++ b/lib/rules/color-contrast-matches.js
@@ -2,7 +2,7 @@
 
 var nodeName = node.nodeName.toUpperCase(),
 	nodeType = node.type,
-	doc = document;
+	doc = axe.commons.dom.getRootNode(node);
 
 if (node.getAttribute('aria-disabled') === 'true' || axe.commons.dom.findUp(node, '[aria-disabled="true"]')) {
 	return false;
@@ -45,9 +45,9 @@ if (nodeName === 'LABEL' || nodeParentLabel) {
 		return false;
 	}
 
-	var candidate = node.querySelector('input:not([type="hidden"]):not([type="image"])' +
+	var candidate = axe.utils.querySelectorAll(virtualNode, 'input:not([type="hidden"]):not([type="image"])' +
 		':not([type="button"]):not([type="submit"]):not([type="reset"]), select, textarea');
-	if (candidate && candidate.disabled) {
+	if (candidate.length && candidate[0].actualNode.disabled) {
 		return false;
 	}
 
@@ -66,15 +66,15 @@ if (axe.commons.text.visible(virtualNode, false, true) === '') {
 }
 
 var range = document.createRange(),
-	childNodes = node.childNodes,
+	childNodes = virtualNode.children,
 	length = childNodes.length,
 	child, index;
 
 for (index = 0; index < length; index++) {
 	child = childNodes[index];
 
-	if (child.nodeType === 3 && axe.commons.text.sanitize(child.nodeValue) !== '') {
-		range.selectNodeContents(child);
+	if (child.actualNode.nodeType === 3 && axe.commons.text.sanitize(child.actualNode.nodeValue) !== '') {
+		range.selectNodeContents(child.actualNode);
 	}
 }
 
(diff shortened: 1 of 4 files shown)


marcysutton:
@dylanb I've completed the changes and added some more tests. Ready for another review.

test/rule-matches/color-contrast-matches.js
```diff
@@ -22,216 +24,326 @@ describe('color-contrast-matches', function () {
...
+		});
+
+		it('should look at the correct root node when looking up implicit label and disabled input', function () {
+			fixture.innerHTML = '<div id="parent">'+
+				'<input id="input">' +
+			'</div>';
+
+			var shadowRoot = document.getElementById('parent').attachShadow({ mode: 'open' });
+			shadowRoot.innerHTML = '<div id="shadowParent">' +
+					'<label for="input" id="shadowLabel">Label' +
```
dylanb: This is another explicit label

test/rule-matches/color-contrast-matches.js
```diff
@@ -22,216 +24,326 @@ describe('color-contrast-matches', function () {
...
+					'<label id="shadowLabel">Label</label>' +
+					'<input aria-labelledby="shadowLabel" disabled>' +
+				'</div>';
+
+			var shadowLabel = fixture.firstChild.shadowRoot.querySelector('#shadowLabel');
+			var tree = axe._tree = axe.utils.getFlattenedTree(fixture);
+			assert.isFalse(rule.matches(shadowLabel, axe.utils.getNodeFromTree(tree[0], shadowLabel)));
+		});
+
+		it('should look at the children of a virtual node for overlap', function () {
```
dylanb: Shouldn't this test be one that overlaps and therefore returns false?

test/rule-matches/color-contrast-matches.js
```diff
@@ -22,216 +24,326 @@ describe('color-contrast-matches', function () {
...
+
+			var shadowRoot = document.getElementById('firstChild').attachShadow({ mode: 'open' });
+			shadowRoot.innerHTML = 'Some text' +
+					'<p style="color: #fff;" id="shadowTarget">Other text</p>';
+
+			var firstChild = fixture.querySelector('#firstChild');
+			var tree = axe._tree = axe.utils.getFlattenedTree(fixture);
+			assert.isTrue(rule.matches(firstChild, axe.utils.getNodeFromTree(tree[0], firstChild)));
+		});
+	}
```
dylanb: I think we should add a test where the implicit label is in the light DOM and the input in the shadow DOM

```
fixture.innerHTML = '<label>Text<div id="firstChild"></div></label>'
...
shadowRoot.innerHTML = '<input id="shadowTarget" type="text" />'
...
```

test/rule-matches/color-contrast-matches.js
marcysutton: Doesn't the default CSS overlap since the text node is a child?

lib/rules/color-contrast-matches.js
```diff
@@ -40,41 +39,43 @@ if (nodeName === 'LABEL' || nodeParentLabel) {
 		relevantNode = nodeParentLabel;
 	}
 	// explicit label of disabled input
+	var doc = axe.commons.dom.getRootNode(node);
```
dylanb: should be `var doc = axe.commons.dom.getRootNode(relevantNode);`

dylanb: Also, probably use `let` to block scope it

test/rule-matches/color-contrast-matches.js
marcysutton: Also that conditional returns true–I thought it was checking to see if there was overlapping content worth evaluating.

test/rule-matches/color-contrast-matches.js
marcysutton: Test added.

test/rule-matches/color-contrast-matches.js
dylanb: :+1:

test/rule-matches/color-contrast-matches.js
dylanb: actually, I was thinking of the disabled case where the input that is inside the shadow DOM is disabled and it gets excluded from the color contrast check - apologies for not being clear enough.

marcysutton: The approach I took with other tests was to put a second, enabled input in the light DOM with the same ID and an explicit pairing with the label–that way it would return `true` if the shadow boundary wasn't being handled properly. Otherwise it's difficult to test the `isFalse` case by itself since that would mean it didn't get matched (and could have occurred for any number of reasons)

marcysutton:
@dylanb I've incorporated those two test changes, it's ready for another review.

PR #418, opened 2017-07-11
New rule, main landmark at least one

ssanaul:
We would like to add rules related to the ARIA authoring practices guide related to the use of landmarks.   

https://www.w3.org/TR/wai-aria-practices-1.1/#aria_landmark 

The at least one landmark is the first rule we have developed and would like to get your feedback.

**More Main Landmark Rules**
We are working on other rules related to main landmark , including:
1. No more than one main landmark in a document
2. Main landmark is a top level landmark

**Tag for Landmark Rules**
We initially use the tag **cat.keyboard** to match the all content in a landmark rule, but since there will be a number of rules we would like to submit related to landmarks, we would like to create a new **cat.landmark** category for rules related to landmarks.

**Contact**
If you have any question or comments you can also contact Jon Gunderson, e-mail: jongund@illinois.edu


lib/checks/keyboard/has-at-least-one-main.js
```diff
@@ -0,0 +1,3 @@
+var main = document.querySelector('main,div[role="main"]');
```
marcysutton: We already have a check for this, see [lib/checks/navigation/landmark.js](https://github.com/dequelabs/axe-core/blob/develop/lib/checks/navigation/landmark.js).

doc/rule-descriptions.md
```diff
@@ -33,6 +33,7 @@
 | input-image-alt | Ensures &lt;input type=&quot;image&quot;&gt; elements have alternate text | cat.text-alternatives, wcag2a, wcag111, section508, section508.22.a | true |
 | label-title-only | Ensures that every form element is not solely labeled using the title or aria-describedby attributes | cat.forms, best-practice | false |
 | label | Ensures every form element has a label | cat.forms, wcag2a, wcag332, wcag131, section508, section508.22.n | true |
+| landmark-at-least-one-main | Each page must have at least one main landmark, used to identify the main content. | cat.keyboard | true |
```
marcysutton: This should get automatically generated by the build system.

lib/rules/landmark-at-least-one-main.json
```diff
@@ -0,0 +1,16 @@
+{
+  "id": "landmark-at-least-one-main",
+  "selector": "html",
+  "tags": [
+    "cat.keyboard"
```
marcysutton: This should be tagged as `best-practice`, and potentially `cat.navigation` instead of `cat.keyboard`.

test/integration/full/document-title/frames/level1.html
```diff
@@ -8,4 +8,4 @@
 	<iframe id="frame2" src="level2.html"></iframe>
 	<iframe id="frame3" src="level2-a.html"></iframe>
 </body>
-</html>
+</html>
```
marcysutton: Can you revert this file since it just removed a line break?

test/integration/full/document-title/frames/level2-a.html
```diff
@@ -8,4 +8,4 @@
 <body>
 	Hi
 </body>
-</html>
+</html>
```
marcysutton: Same as above and below–can you revert these integration test files to re-introduce the line ending

jongund:
Thank you for your comments.
We will review the comments and resubmit the pull request.

lib/rules/landmark-at-least-one-main.json
dylanb: I think cat.keyboard is ok

lib/checks/keyboard/has-at-least-one-main.js
dylanb: The existing check does not return the data and does not implement an after, so this check is better and more complete.

lib/rules/landmark-at-least-one-main.json
dylanb: I agree with the best-practice, but cat.keyboard is what the current landmark check (used by the bypass rule) is categorized under

test/checks/keyboard/has-at-least-one-main.js
```diff
@@ -0,0 +1,77 @@
...
+	};
+
+	afterEach(function () {
+		fixture.innerHTML = '';
+		checkContext._data = null;
+	});
+
+	
+	it('should return false if main does not exist', function () {
+		var node = fixture.querySelector('main,div[role="main"]');
```
dylanb: Isn't this supposed to pass the 'html' element into the evaluate function?

dylanb: this same comment applies to the rest of the tests

dylanb:
This needs tests for the `after` function

doc/rule-descriptions.md
dylanb: it does and it did, so this change being committed as part of the PR is what we want so when it merges, the documentation in GitHub is updated

lib/rules/landmark-at-least-one-main.json
```diff
@@ -2,7 +2,8 @@
   "id": "landmark-at-least-one-main",
   "selector": "html",
   "tags": [
-    "cat.keyboard"
+    "cat.keyboard",
+	"cat.best-practice"
```
dylanb: just `best-practice`

ssanaul:
Thank you for the feedback. I have made the changes that were suggested. Sorry about the earlier commit messages, I didn't see the guidelines until now. Also, concerning the edited document-title iframes, it doesn't seem like a line break was added, but rather that the end html tag was deleted and replaced. If there is anything else I can do, please let me know.

lib/rules/landmark-at-least-one-main.json
```diff
@@ -0,0 +1,16 @@
+{
+  "id": "landmark-at-least-one-main",
+  "selector": "html",
+  "tags": [
+	"cat.best-practice"
+  ],
+  "metadata": {
+    "description": "Each page must have at least one main landmark, used to identify the main content.",
+    "help": "A <main> element or an element with a role='main' attribute defines a main landmark."
```
marcysutton: I should have mentioned this earlier, but the `help` text is what shows up in the sidebar of the aXe extensions. So it should be short, and represent the violation a user will encounter when the rule fails. Something like "Page should contain at least one main landmark" ?

test/checks/keyboard/has-at-least-one-main.js
```diff
@@ -0,0 +1,97 @@
...
+	};
+
+	afterEach(function () {
+		fixture.innerHTML = '';
+		checkContext._data = null;
+	});
+
+	
+	it('should return false if main does not exist', function () {
+		var node = document.createElement('html');
```
marcysutton: Was there a reason to add an HTML element here? The fixture should already exist in the document body.

ssanaul: The new tests now test on the already existing HTML element.

lib/rules/landmark-at-least-one-main.json
ssanaul: Rule is now tagged as best-practice.

jongund:
@WilcoFiers 

I am concerned that combining the two rules would create potentially confusing feedback to the user on how to fix the problem and users thinking that the main document and each iframe in a page all need a main landmark.

For example if a page has no main landmarks and contains several iframes (e.g. document objects), since none of the document objects has a main landmark they would all look like they failed and the user might think they need to add a main landmark to the main document and each of the iframe documents.  Clearly not what we usually want, typically we would only wnat one of those documents to have a main landmark on the page.

Also if user had more than one main landmark in any of document objects (e.g. main document or iframe documents), you wouldn't necessarily want the message to say that they had to have one main landmark in each of these documents.

So my main question is how does messaging going to work in the case of a main document and multiple iframes.   I am concerned people getting confused that each iframe needs a main landmark in the case where a page doesn't have any main landmarks, since the messaging system I think will make it look like both the main document and each of the iframes "failed" the page test.

I am also concerned in the case where the main document and several iframe documents have main landmarks, but the main document or one of the iframe documents has more than one main landmark, how will the messaging indicate that that main document or iframe document with more than one main landmark should only have one or maybe no main landmarks.

Is there a "page" level rule with an after function we could look at as a model for how these two rules could be combined with clear messaging on what the problems are and how to fix them?

WilcoFiers:
@jongund If you keep the checks separate, this shouldn't be a problem. The check messages are used to indicate how to resolve the error. So the rule would be something like "Use a single main landmark in the page", and in the remediation section it would say either: "Document has more than one main landmark", or "Document has no main landmarks". If the concern is about confusing users to add main into iframes, we can improve on the error message itself, something like "The unframed page does not have a main landmark".

jongund:
We will create a new pull request that combines the "at least one" and "no more than one" rules into a single rule and create a NEW pull request for "only-one" main landmark

marcysutton:
I'm getting lost which one is which, is "only-one" related to #462? Because by name it sounds like the same thing as "no more than one".

jongund:
@marcysutton 
Wilco asked that we combine the rule at-least-one and no -more-than-one into a single rule.  So have done that and are working on a new pull request with the NEW rule and if that rule is accepted you can just reject the pull requests for at-least-one and no-more-than-one.

marcysutton:
Ok Awesome! I wasn't sure if there still would be two rules total. Thanks again for your work on this!

ssanaul:
The check from this PR has been integrated into the rule, landmark-one-main.

PR #420, opened 2017-07-12
shadowDOM accessible text calculation

### LINKED ISSUE #399: Implement shadow DOM in axe.commons.text.accessibleText
Needs to accept a virtualNode instead of a node and traverse the flattened tree instead of the DOM tree

WilcoFiers:
This PR aims to close #399.

### DIFF
--- a/lib/commons/dom/find-elms-in-context.js
+++ b/lib/commons/dom/find-elms-in-context.js
@@ -1,6 +1,6 @@
 /* global axe, dom */
 /**
- * Find a elements reference from a given context
+ * Find elements referenced from a given context
  * 
  * @param object {
  *    context:  Node | virtual node    Element in the same context
--- a/lib/commons/dom/find-elms-in-context.js
+++ b/lib/commons/dom/find-elms-in-context.js
@@ -1,0 +1,26 @@
+/* global axe, dom */
+/**
+ * Find a elements reference from a given context
+ * 
+ * @param object {
+ *    context:  Node | virtual node    Element in the same context
+ *    value:    String  attribute value to search for
+ *    attr:     String  attribute name to search for
+ *    elm:      String  ndoeName to search for (optional)
+ *  }
+ * @return Array[Node]  
+ */
+dom.findElmsInContext = function ({ context, value, attr, elm = '' }) {
+  let root;
+  context = context.actualNode || context;
+  const escapedValue = axe.utils.escapeSelector(value);
+
+  if (context.nodeType === 9 || context.nodeType === 11) { // It's already root
+    root = context;
+  } else {
+    root = dom.getRootNode(context);
+  }
+  return Array.from(
+    root.querySelectorAll(elm + '[' + attr + '=' + escapedValue + ']')
+  );
+};
--- a/lib/commons/text/accessible-text.js
+++ b/lib/commons/text/accessible-text.js
@@ -21,9 +21,10 @@ var phrasingElements = ['A', 'EM', 'STRONG', 'SMALL', 'MARK', 'ABBR', 'DFN', 'I'
  */
 function findLabel({ actualNode }) {
 	let label;
-	const id = axe.utils.escapeSelector(actualNode.id);
-	if (id) {
-		label = document.querySelector('label[for="' + id + '"]');
+	if (actualNode.id) {
+		label = dom.findElmsInContext({
+			elm: 'label', attr: 'for', value: actualNode.id, context: actualNode
+		})[0];
 	} else {
 		label = dom.findUp(actualNode, 'label');
 	}
@@ -214,24 +215,29 @@ text.accessibleText = function(element, inLabelledByContext) {
 	}
 
 	function checkARIA (element, inLabelledByContext, inControlContext) {
+		let returnText = '';
 		const { actualNode } = element;
 		if (!inLabelledByContext && actualNode.hasAttribute('aria-labelledby')) {
-			return text.sanitize(dom.idrefs(actualNode, 'aria-labelledby').map(label => {
-				if (actualNode === label) {
-					encounteredNodes.pop();
-				} //let element be encountered twice
-
-				const vLabel = axe.utils.getNodeFromTree(axe._tree[0], label);
-				return accessibleNameComputation(vLabel, true, actualNode !== label);
+			// Store the return text, if it's empty, fall back to aria-label
+			returnText = text.sanitize(dom.idrefs(actualNode, 'aria-labelledby').map(label => {
+				if (label !== null) {// handle unfound elements by dom.idref
+					if (actualNode === label) {
+						encounteredNodes.pop();
+					} //let element be encountered twice
+					const vLabel = axe.utils.getNodeFromTree(axe._tree[0], label);
+					return accessibleNameComputation(vLabel, true, actualNode !== label);
+				} else {
+					return '';
+				}
 			}).join(' '));
 		}
 
-		if (!(inControlContext && isEmbeddedControl(element)) && 
+		if (!returnText && !(inControlContext && isEmbeddedControl(element)) && 
 				actualNode.hasAttribute('aria-label')) {
 			return text.sanitize(actualNode.getAttribute('aria-label'));
 		}
 
-		return '';
+		return returnText;
 	}
 
 	/**
@@ -244,18 +250,14 @@ text.accessibleText = function(element, inLabelledByContext) {
 	 * @return {string}
 	 */
 	accessibleNameComputation = function (element, inLabelledByContext, inControlContext) {
-		// TODO: Make sure I don't have to do this
 		let returnText;
-		if (element.actualNode instanceof Node !== true) {
-			/* global console */
-			var e = new Error('Invalid argument. Virtual Node must be provided');
-			console.error(e);
-			throw e;
-		}
-
 		// If the node was already checked or is null, skip
 		if (!element || encounteredNodes.includes(element)) {
 			return '';
+
+		// if the node is invalid, throw
+		} else if (element !== null && element.actualNode instanceof Node !== true) {
+			throw new Error('Invalid argument. Virtual Node must be provided');
 
 		//Step 2a: Skip if the element is hidden, unless part of labelledby
 		} else if(!inLabelledByContext && !dom.isVisible(element.actualNode, true)) {
--- a/lib/commons/text/accessible-text.js
+++ b/lib/commons/text/accessible-text.js
@@ -1,5 +1,5 @@
 /*global text, dom, aria, axe */
-/*jshint maxstatements: 25, maxcomplexity: 19 */
+/*jshint maxstatements: 27, maxcomplexity: 19 */
 
 var defaultButtonValues = {
 	submit: 'Submit',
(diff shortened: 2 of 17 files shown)


lib/commons/dom/find-elms-in-context.js
```diff
@@ -0,0 +1,26 @@
+/* global axe, dom */
+/**
+ * Find a elements reference from a given context
```
dylanb: sp: 'referenced`

lib/commons/text/accessible-text.js
```diff
@@ -80,14 +79,14 @@ function formValueText(element) {
...
+	if (nodeName === 'TEXTAREA' && actualNode.value) {
+		return actualNode.value;
 	}
 	return '';
 }
 
-function checkDescendant(element, nodeName) {
-	var candidate = element.querySelector(nodeName.toLowerCase());
+function checkDescendant({ actualNode }, nodeName) {
+	var candidate = actualNode.querySelector(nodeName.toLowerCase());
```
dylanb: should we put a note in here to indicate that this function is intended to be used only on elements that cannot contain a shadow root?

lib/commons/text/accessible-text.js
```diff
@@ -137,34 +138,34 @@ function nonEmptyText(t) {
  * @return {string}
  */
 text.accessibleText = function(element, inLabelledByContext) {
-	//todo: implement shadowDOM
-	var accessibleNameComputation;
-	var encounteredNodes = [];
+	let accessibleNameComputation;
+	const encounteredNodes = [];
+	if (element instanceof Node) {
```
dylanb: This works differently from a lot of the other commons - which do not try to maintain backwards compatibility. Whether we want to support backwards compatibility or not and if we do, we need to change those other utilities.

lib/commons/text/accessible-text.js
```diff
@@ -137,34 +138,34 @@ function nonEmptyText(t) {
...
+		return element.children.reduce((returnText, child) => {
+			const { actualNode } = child;
+			if (actualNode.nodeType === 3) {
+				returnText += actualNode.nodeValue;
+			} else if (actualNode.nodeType === 1) {
+				if (!phrasingElements.includes(actualNode.nodeName.toUpperCase())) {
 					returnText += ' ';
 				}
-				returnText += accessibleNameComputation(nodes[i], inLabelledByContext, inControlContext);
+				returnText += accessibleNameComputation(
```
dylanb: This breaks our current style conventions

test/checks/forms/fieldset.js
```diff
@@ -226,29 +226,27 @@ describe('fieldset', function () {
...
-			fixture.innerHTML = '<div id="grouplabel">Label</div>' +
+			fixtureSetup('<div id="grouplabel">Label</div>' +
 				'<div role="radiogroup" aria-labelledby="grouplabel">' +
 				'<input type="' + type + '" id="target" name="s.%$#n">' +
-				'<input type="' + type + '" name="s.%$#n"></div>';
-
+				'<input type="' + type + '" name="s.%$#n"></div>');
 			var node = fixture.querySelector('#target');
 			assert.isFalse(checks.fieldset.evaluate.call(checkContext, node));
-
```
dylanb: We need to add tests that deal with this sort of thing

```
        <div>
            <form>
                <div class="fieldset">
                    <span slot="legend">Shadow DOM fieldset</span>
                    <label>Option 1<input type="radio" value="one" name="group" /></label>
                    <label>Option 2<input type="radio" value="two" name="group" /></label>
                </div>
            </form>
        </div>
        <script>
            function createContentSlotted() {
                var group = document.createElement('fieldset');
                group.innerHTML = '<legend><slot name="legend"></slot></legend><slot></slot>';
                return group;
            }

            function makeShadowTreeSlotted(node) {
                var root = node.attachShadow({mode: 'open'});
                root.appendChild(createContentSlotted());
            }

            document.addEventListener('DOMContentLoaded', function() {
                document.querySelectorAll('.fieldset').forEach(makeShadowTreeSlotted);
            });
        </script>
```

test/checks/keyboard/focusable-no-name.js
```diff
@@ -2,37 +2,38 @@ describe('focusable-no-name', function () {
...
 		var node = fixture.querySelector('span');
 		assert.isTrue(checks['focusable-no-name'].evaluate(node));
 	});
 
 	it('should pass if the element is tabbable but has an accessible name', function () {
-		fixture.innerHTML = '<a href="#" title="Hello"></a>';
+		fixtureSetup('<a href="#" title="Hello"></a>');
 		var node = fixture.querySelector('a');
 		assert.isFalse(checks['focusable-no-name'].evaluate(node));
 	});
```
dylanb: Add a test where when an anchor has content inside the shadow root, it returns false. Test this for slotted content as well as fallback content inside the default slot

test/commons/text/accessible-text.js
```diff
@@ -273,198 +298,278 @@ describe('text.accessibleText', function() {
...
 
 		it('should not visit innerText of figure', function() {
 			fixture.innerHTML = '<figure>Hello<figcaption></figcaption></figure>';
-			var target = fixture.querySelector('figure');
+			axe._tree = axe.utils.getFlattenedTree(fixture);
+
+			var target = axe.utils.querySelectorAll(axe._tree, 'figure')[0];
 			assert.equal(axe.commons.text.accessibleText(target), '');
 		});
+
```
dylanb: I think we should add a couple more tests here;

1. fallback content for v1
2. correct use of an attribute (e.g. title) within the slotted shadow DOM content,
3. correct use of an attribute (e.g. title) within the shadow DOM content

test/checks/label/implicit.js
```diff
@@ -2,33 +2,34 @@ describe('implicit-label', function () {
...
 	});
 
 	it('should return false if a label is not present', function () {
 		var node = document.createElement('input');
 		node.type = 'text';
-		fixture.appendChild(node);
+		fixtureSetup(node);
 
 		assert.isFalse(checks['implicit-label'].evaluate(node));
 	});
```
dylanb: This needs to support this example

```
        <form>
            <label>Option 1<div class="slotted"><input type="text" name="slottedLabel" /></div></label>
        </form>
        <script>
            function createContentSlotted() {
                var group = document.createElement('p');
                group.innerHTML = '<slot></slot>';
                return group;
            }

            function makeShadowTreeSlotted(node) {
                var root = node.attachShadow({mode: 'open'});
                root.appendChild(createContentSlotted());
            }

            document.addEventListener('DOMContentLoaded', function() {
                document.querySelectorAll('.slotted').forEach(makeShadowTreeSlotted);
            });
        </script>
```

Which works properly in both Safari and Chrome with VO

test/checks/forms/fieldset.js
```diff
@@ -43,8 +43,8 @@ describe('fieldset', function () {
 		});
 
 		it('should return false if the group has no legend element', function () {
-			fixture.innerHTML = '<fieldset><input type="' + type + '" id="target" name="uniqueyname">' +
-				'<input type="' + type + '" name="uniqueyname"></fieldset>';
+			fixtureSetup('<fieldset><input type="' + type + '" id="target" name="uniqueyname">' +
```
dylanb: this test here should work when the fieldset has been found by going up the slotted content - see the example below

test/checks/tables/same-caption-summary.js
```diff
@@ -2,37 +2,38 @@ describe('same-caption-summary', function () {
...
 		assert.isFalse(checks['same-caption-summary'].evaluate(node));
 
 	});
 
 	it('should return true if summary and caption are the same', function () {
-		fixture.innerHTML = '<table summary="Hi"><caption>Hi</caption><tr><td></td></tr></table>';
+		fixtureSetup('<table summary="Hi"><caption>Hi</caption><tr><td></td></tr></table>');
 		var node = fixture.querySelector('table');
 
 		assert.isTrue(checks['same-caption-summary'].evaluate(node));
```
dylanb: This should catch the following situation

```
        <div class="table">
            <span slot="caption">Caption</span>
            <span slot="one">Data element 1</span>
            <span slot="two">Data element 2</span>
        </div>
        <script>
            function createContentTable() {
                var group = document.createElement('table');
                group.innerHTML = '<caption><slot name="caption"></slot></caption>' +
                    '<tr><td><slot name="one"></slot></td><td><slot name="two"></slot></td></tr>';
                group.setAttribute('summary', 'Caption');
                return group;
            }

            function makeShadowTreeTable(node) {
                var root = node.attachShadow({mode: 'open'});
                root.appendChild(createContentTable());
            }

            document.addEventListener('DOMContentLoaded', function() {
                document.querySelectorAll('.table').forEach(makeShadowTreeTable);
            });
        </script>
```

test/commons/text/accessible-text.js
```diff
@@ -42,10 +46,11 @@ describe('text.accessibleText', function() {
...
 			'  </label>' +
 			'</fieldset>';
+		axe._tree = axe.utils.getFlattenedTree(fixture);
 
-		var rule2a = fixture.querySelector('#beep');
-		var rule2b = fixture.querySelector('#flash');
-		assert.equal(axe.commons.text.accessibleText(rule2a), 'Beep');
+		// var rule2a = axe.utils.querySelectorAll(axe._tree, '#beep')[0];
+		var rule2b = axe.utils.querySelectorAll(axe._tree, '#flash')[0];
+		// assert.equal(axe.commons.text.accessibleText(rule2a), 'Beep');
```
dylanb: why is this commented?

lib/commons/text/accessible-text.js
WilcoFiers: The reason I did this is because it would have taken a whole lot of additional changes to make sure every method that uses accessibleText has access to a virtual node. It wasn't even for backward compatibility, but that's a good additional reason. Further, I think these methods should be useful outside of a rule too, which they aren't currently, since they rely on setup that gets done in another place.

I think, instead of calling `element = axe.utils.getNodeFromTree(axe._tree[0], element);` we need a method that will take any DOM element and return it's virtual counterpart. Something like `virtualNode = axe.utils.getVirtualNode(node)`. It can look at `axe._tree[0]` if available, and if it isn't, it should create a new tree and pull the virtual node from that. We could also make that method store virtual nodes in a set rather than a tree, so that lookup is faster.

lib/commons/text/accessible-text.js
WilcoFiers: Ugh... It's really frustrating going between axe-core and axe-devtools. :/ sorry

test/checks/forms/fieldset.js
WilcoFiers: Should be covered in #425

test/commons/text/accessible-text.js
WilcoFiers: oops.

lib/commons/text/accessible-text.js
dylanb: Yeah, the biggest problem with not passing in the virtualNode is the potential performance impact. I am worried about memory leaks if we use these APIs (or make them usable) without understanding what they do. So I don't support building this cache if it doesn't exist.

dylanb: I am going to accept this pull request and have opened a separate issue to deal with this https://github.com/dequelabs/axe-core/issues/435

test/checks/forms/fieldset.js
WilcoFiers: Should be covered in #425

WilcoFiers:
@dylanb Thanks for all the excellent comments. I've updated the PR accordingly. Can you review again?

test/checks/keyboard/focusable-no-name.js
dylanb: Can you add one more

```
		var node = document.createElement('a');
		var shadow = node.attachShadow({ mode: 'open' });
		shadow.innerHTML = '<slot>Fallback content</slot>';
		fixtureSetup(node);

		assert.isFalse(checks['focusable-no-name'].evaluate(node));
```

test/checks/tables/same-caption-summary.js
```diff
@@ -37,8 +38,23 @@ describe('same-caption-summary', function () {
...
-
 	});
 
+	(shadowSupport.v1 ? it : xit)('should match slotted caption elements', function () {
+		var node = document.createElement('div');
+		node.innerHTML = '<span slot="caption">Caption</span>' +
+      '<span slot="one">Data element 1</span>' +
```
dylanb: @WilcoFiers indentation

dylanb:
@WilcoFiers one small indentation problem and then I suggested another test - then we should be good to go

test/checks/tables/same-caption-summary.js
WilcoFiers: it's an indentation feature ;)

Issue #421, opened 2017-07-12
"aria-required-children" check to use the composed tree


WilcoFiers:
"aria-required-children" check to use the composed tree

marcysutton:
closed with #436.

PR #436, opened 2017-07-13
feat: add shadow support to aria-required-children

marcysutton:
I didn't add any shadow DOM tests for `aria-owns` because ID attributes get reset inside of shadow boundaries, so that technique won't work. Are there any other Shadow DOM scenarios I should consider?

Closes https://github.com/dequelabs/axe-core/issues/421

### DIFF
--- a/lib/checks/aria/required-children.js
+++ b/lib/checks/aria/required-children.js
@@ -3,7 +3,7 @@ implicitNodes = axe.commons.aria.implicitNodes,
 matchesSelector = axe.commons.utils.matchesSelector,
 idrefs = axe.commons.dom.idrefs;
 
-function owns(node, role, ariaOwned) {
+function owns(node, virtualTree, role, ariaOwned) {
 	if (node === null) { return false; }
 	var implicit = implicitNodes(role),
 	selector = ['[role="' + role + '"]'];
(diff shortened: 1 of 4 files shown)


test/checks/aria/required-children.js
```diff
@@ -10,97 +11,141 @@ describe('aria-required-children', function () {
 		}
 	};
 
+	function checkSetup (html, options, target) {
```
dylanb: This will have to be refactored once @WilcoFiers pull request gets merged - perhaps you should copy that change so that it will merge and eliminate refactoring?

lib/checks/aria/required-children.js
```diff
@@ -13,17 +13,17 @@ function owns(node, role, ariaOwned) {
...
+		!!axe.utils.querySelectorAll(virtualTree, selector)[0];
 }
 
 function ariaOwns(nodes, role) {
 	var index, length;
 
 	for (index = 0, length = nodes.length; index < length; index++) {
 		if (nodes[index] === null) { continue; }
-		if (owns(nodes[index], role, true)) {
+		let virtualTree = axe.utils.getFlattenedTree(nodes[index]);
```
dylanb: I think you should use `axe.utils.getNodeFromTree()` instead of recreating the flattened tree

marcysutton: That's the whole problem though–I don't have access to the virtualNode at that time, and the arguments for `getNodeFromTree` include `vNode, node`. This method is called recursively on a collection of elements from `idRefs`, which don't return virtualNodes. There's a comment in that function about only working with the light DOM. How would I get the virtualNode in this `ariaOwns` function without getting the flattened tree first?

test/checks/aria/required-children.js
marcysutton: Ok, I've merged in that branch and used the checkSetup in testutils.

dylanb:
NOTE: we need to wait for the other PR to be merged before finalizing review on this PR

marcysutton:
@dylanb did you mean to wait for the accessible-text PR? I rebased to get rid of conflicts with that branch and it should be good to go now.

dylanb:
@WilcoFiers the downside to changing to ES6 notation is that we lose the history of why a change was made, so I am also completely fine with leaving this as-is

PR #437, opened 2017-07-14
shadowDOM for media checks

### LINKED ISSUE #423: "caption" and "description" checks to search for <track> in composed tree
"caption" and "description" checks to search for <track> in composed tree

WilcoFiers:
Should close #423.

### DIFF
--- a/lib/checks/media/description.js
+++ b/lib/checks/media/description.js
@@ -3,7 +3,7 @@ var tracks = axe.utils.querySelectorAll(virtualNode, 'track');
 if (tracks.length) {
   // return false if any track has kind === 'description'
   var out = !tracks.some(({ actualNode }) => (
-    actualNode.getAttribute('kind').toLowerCase() === 'descriptions'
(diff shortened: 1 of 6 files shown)


lib/checks/media/description.js
```diff
@@ -1,12 +1,12 @@
...
-		var kind = tracks[i].getAttribute('kind');
-		if (kind && kind === 'descriptions') {
-			// only return for matching track, in case there are multiple
-			return false;
-		}
-	}
-	return true;
+  // return false if any track has kind === 'description'
+  var out = !tracks.some(({ actualNode }) => (
+    actualNode.getAttribute('kind').toLowerCase() === 'descriptions'
```
dylanb: Could you add tests for both checks that ensures that th check returns without an exception when the markup looks like

```
<audio><track></track></audio>
```

i.e. no `kind` attribute at all

WilcoFiers: Right you are. .toLowerCase() wouldn't work. Will add/fix

PR #438, opened 2017-07-14
fix: getComposedParent should not return slot nodes

WilcoFiers:


lib/commons/dom/get-composed-parent.js
```diff
@@ -6,7 +6,10 @@
  */
 dom.getComposedParent = function getComposedParent (element)  {
   if (element.assignedSlot) {
-    return element.assignedSlot; // content of a shadow DOM slot
+    // NOTE: If the display of a slot element isn't 'contents',
+    // the slot shouldn't be ignored. Chrome does not support this (yet) so,
+    // we'll skip this part for now.
+    return getComposedParent(element.assignedSlot); // content of a shadow DOM slot
```
dylanb: The comment should read `parent of shadow DOM slot`

PR #447, opened 2017-07-18
fix: Let findUp work on shadow root children

### LINKED ISSUE #422: "aria-required-parent" check to use the composed tree in getAriaOwners
"aria-required-parent" check to use the composed tree in getAriaOwners

WilcoFiers:
Should help resolve #422.

### DIFF
--- a/lib/commons/dom/find-up.js
+++ b/lib/commons/dom/find-up.js
@@ -9,25 +9,16 @@
  * @return {HTMLElement|null} Either the matching HTMLElement or `null` if there was no match
  */
 dom.findUp = function (element, target) {
-	'use strict';
-	/*jslint browser:true*/
+	let doc, matches,
+		parent = element;
 
-	var parent,
-		doc = axe.commons.dom.getRootNode(element),
-		matches;
-
-	matches = doc.querySelectorAll(target);
-	matches = axe.utils.toArray(matches);
-	if (doc === document && !matches.length) {
-		return null;
-	}
-
-	// recursively walk up the DOM, checking each parent node
-	parent = dom.getComposedParent(element);
-	while (parent && matches.indexOf(parent) === -1) {
-		parent = (parent.assignedSlot) ? parent.assignedSlot : parent.parentNode;
+	do {// recursively walk up the DOM, checking each parent node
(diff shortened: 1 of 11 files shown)


lib/commons/dom/find-up.js
```diff
@@ -9,33 +9,24 @@
  * @return {HTMLElement|null} Either the matching HTMLElement or `null` if there was no match
  */
 dom.findUp = function (element, target) {
-	'use strict';
```
dylanb: These changes don't seem to actually change anything in terms of functionality - is this purely a performance fix?

dylanb: ok, so the difference here is that if the starting node is the highest node in the shadow DOM, then the old code did not find a matching host node - whereas this code does?

test/commons/dom/find-up.js
```diff
@@ -107,12 +117,22 @@ describe('dom.findUp', function () {
...
+
+		fixture.innerHTML = '<label><div></div></label>';
+		makeShadowTree(fixture.querySelector('div'));
+		var tree = axe.utils.getFlattenedTree(fixture.firstChild);
+		var el = axe.utils.querySelectorAll(tree, 'a')[0];
+		assert.equal(axe.commons.dom.findUp(el.actualNode, 'label'), fixture.firstChild);
+	});
+
+	(shadowSupport.v1 ? it : xit)('should work on shadow root children', function () {
+		fixture.innerHTML = '<div role="list"><div id="something"></div></div>';
```
dylanb: shouldn't this attach to the div with role="list"?

lib/commons/dom/find-up.js
WilcoFiers: The problem is that if the child is a directly on the shadow root, it would start looking from the host (outside the shadow), even though the matches array still only contained elements inside the shadow tree.

Followup thought, should this use element.matches() instead of doing querySelectAll() and see if the element is included? I imagine that'd be faster.

WilcoFiers: The problem is that if the child is a directly on the shadow root, it would start looking from the host (outside the shadow), even though the matches array still only contained elements inside the shadow tree.

Followup thought, should this use element.matches() instead of doing querySelectAll() and see if the element is included? I imagine that'd be faster.

Issue #425, opened 2017-07-12
"fieldset" check to use the composed tree


WilcoFiers:
"fieldset" check to use the composed tree

dylanb:
This case needs to be handled

```
 <div>
            <form>
                <div class="fieldset">
                    <span slot="legend">Shadow DOM fieldset</span>
                    <label>Option 1<input type="radio" value="one" name="group" /></label>
                    <label>Option 2<input type="radio" value="two" name="group" /></label>
                </div>
            </form>
        </div>
        <script>
            function createContentSlotted() {
                var group = document.createElement('fieldset');
                group.innerHTML = '<legend><slot name="legend"></slot></legend><slot></slot>';
                return group;
            }

            function makeShadowTreeSlotted(node) {
                var root = node.attachShadow({mode: 'open'});
                root.appendChild(createContentSlotted());
            }

            document.addEventListener('DOMContentLoaded', function() {
                document.querySelectorAll('.fieldset').forEach(makeShadowTreeSlotted);
            });
        </script>
```

PR #448, opened 2017-07-18
shadowDOM fieldset check

WilcoFiers:
This solves #425. It builds on #447 (which should be merged first).


test/checks/forms/fieldset.js
```diff
@@ -1,7 +1,9 @@
 describe('fieldset', function () {
 	'use strict';
 	var fixture = document.getElementById('fixture');
+	var shadowSupport = axe.testUtils.shadowSupport;
```
marcysutton: This variable needs `.v1` at the end of it, or the tests below need it. There are some test failures as a result

PR #452, opened 2017-07-19
feat(duplicate-id): Add shadow DOM support

### LINKED ISSUE #429: "duplicate-id" to allow repeating IDs in different contexts
"duplicate-id" to allow repeating IDs in different contexts

WilcoFiers:
Closes #429.

### DIFF
--- a/test/checks/shared/duplicate-id.js
+++ b/test/checks/shared/duplicate-id.js
@@ -94,7 +94,7 @@ describe('duplicate-id', function () {
 		assert.lengthOf(checkContext._relatedNodes, 0);
 	});
 
-	(shadowSupport.v1 ? it : xit)('should not ignore slotted elements', function () {
+	(shadowSupport.v1 ? it : xit)('should compare slotted content with the outer tree', function () {
 		var node = document.createElement('div');
 		node.id = 'target';
 		node.innerHTML = '<p id="target">text</p>';
--- a/test/checks/shared/duplicate-id.js
+++ b/test/checks/shared/duplicate-id.js
@@ -58,7 +58,7 @@ describe('duplicate-id', function () {
 		assert.isTrue(checks['duplicate-id'].evaluate.call(checkContext, node));
 	});
 
-	(shadowSupport.v1 ? it : xit)('should find duplicate IDs in shadow trees', function () {
+	(shadowSupport.v1 ? it : xit)('should find duplicate IDs in the same shadow DOM', function () {
 		var div = document.createElement('div');
 		div.id = 'target';
 		var shadow = div.attachShadow({ mode: 'open' });
@@ -71,7 +71,7 @@ describe('duplicate-id', function () {
 		assert.deepEqual(checkContext._relatedNodes, [shadow.querySelector('p')]);
 	});
 
-	(shadowSupport.v1 ? it : xit)('should ignore same IDs in shadow trees', function () {
+	(shadowSupport.v1 ? it : xit)('should ignore duplicate IDs if they are in different document roots', function () {
 		var node = document.createElement('div');
 		node.id = 'target';
 		var shadow = node.attachShadow({ mode: 'open' });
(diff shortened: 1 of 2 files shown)


test/checks/shared/duplicate-id.js
```diff
@@ -58,4 +58,52 @@ describe('duplicate-id', function () {
...
+		var node = document.createElement('div');
+		node.id = 'target';
+		node.innerHTML = '<p id="target">text</p>';
+		var shadow = node.attachShadow({ mode: 'open' });
+		shadow.innerHTML = '<span id="target"><slot></slot></span>';
+		fixture.appendChild(node);
+
+		assert.isFalse(checks['duplicate-id'].evaluate.call(checkContext, node));
+		assert.lengthOf(checkContext._relatedNodes, 1);
+		assert.deepEqual(checkContext._relatedNodes, [node.querySelector('p')]);
```
marcysutton: The result of this looks like it is being ignored–there are 3 ids of `target`, and the result doesn't seem to include anything about the one inside the Shadow DOM. Can you expand on how this should work?

WilcoFiers: IDs are scoped to their respective root. In this example, the `div` and `p` both have `document` as the root, and so `p` is the duplicate in this example. But because `span` is in the shadow tree, it can have an ID that already exists in the document (or even in another shadow tree).

marcysutton: Maybe the description of the test is just confusing? `it should not ignore slotted elements`

test/checks/shared/duplicate-id.js
```diff
@@ -58,4 +58,52 @@ describe('duplicate-id', function () {
 		assert.isTrue(checks['duplicate-id'].evaluate.call(checkContext, node));
 	});
 
+	(shadowSupport.v1 ? it : xit)('should find duplicate IDs in shadow trees', function () {
```
dylanb: A slightly more accurate test description would be

```
Should find duplicate IDs in the same shadow DOM
```

test/checks/shared/duplicate-id.js
```diff
@@ -58,4 +58,52 @@ describe('duplicate-id', function () {
...
+		shadow.innerHTML = '<span id="target"></span><p id="target">text</p>';
+		var node = shadow.querySelector('span');
+		fixture.appendChild(div);
+
+		assert.isFalse(checks['duplicate-id'].evaluate.call(checkContext, node));
+		assert.lengthOf(checkContext._relatedNodes, 1);
+		assert.deepEqual(checkContext._relatedNodes, [shadow.querySelector('p')]);
+	});
+
+	(shadowSupport.v1 ? it : xit)('should ignore same IDs in shadow trees', function () {
```
dylanb: change to

```
Should ignore duplicate IDs if they are in different document roots
```

test/checks/shared/duplicate-id.js
```diff
@@ -58,4 +58,52 @@ describe('duplicate-id', function () {
...
+		node.id = 'target';
+		var shadow = node.attachShadow({ mode: 'open' });
+		shadow.innerHTML = '<span id="target"></span>';
+		fixture.appendChild(node);
+
+		assert.isTrue(checks['duplicate-id'].evaluate.call(checkContext, node));
+		assert.lengthOf(checkContext._relatedNodes, 0);
+	});
+
+	(shadowSupport.v1 ? it : xit)('should ignore same IDs outside shadow trees', function () {
```
dylanb: This test is a duplicate of the one above

test/checks/shared/duplicate-id.js
dylanb: I like the description of the test because it describes what is actually happening i.e. that the slotted content (light DOM) is being compared with the light DOM and not the shadow DOM it is slotted into. You could slightly clarify it by saying

```
Compare slotted content with the light DOM
```

dylanb:
@WilcoFiers also, you will need to rebase this change onto the shadowDOM branch because it seems to regress some of the tests on the region shadow DOM implementation.

test/checks/shared/duplicate-id.js
WilcoFiers: It's not. The top one starts in the light tree, and this one starts in the shadow tree.

WilcoFiers:
Updated the text and rebased.

PR #455, opened 2017-07-19
Sd/commons revisited

### LINKED ISSUE #435: Consistent treatment of virtualNode
Some utils (shorthand for utils and commons) are using the virtualNode whereas some are using node and then looking up the virtualNode in the flattened tree cache.

This is inconsistent but also raises the following concerns:

1. Performance of lookups
1. Reliance on cache making utilities less usable for other purposes
1. Memory leaks if cache left uncleared
1. No transparency when DOM can change and cache is out of date
1. Backwards compatibility of existing rule sets or custom rules

I propose that we do the following:

1. Create new utils that have a unique name that identifies them as needing a virtualNode
1. Create simple utils with the current names that perform the virtualNode lookup and then call the new util (for backwards compatibility)

In the builtin rules that will ship with this version of axe-core:

1. Use the new utils that require the virtualNode in every context where the virtualNode is available (i.e. every check and matches function) and
1. Call the lookup util in all other cases

@marcysutton @WilcoFiers thoughts?

WilcoFiers:
Should close #435.


lib/checks/label/duplicate-img-label.js
```diff
@@ -1,4 +1,4 @@
-const text = axe.commons.text.visible(virtualNode, true).toLowerCase();
+const text = axe.commons.text.visibleVirtual(virtualNode, true).toLowerCase();
```
marcysutton: Didn't @dylanb suggest doing this the other way around? (i.e. calling `visibleVirtual` from `visible` so the checks stay the same)

lib/commons/aria/label.js
```diff
@@ -2,21 +2,18 @@
...
  */
 aria.label = function (node) {
-	var ref, candidate;
-	if (node.actualNode instanceof Node === false) {
-		node = axe.utils.getNodeFromTree(axe._tree[0], node);
-	}
+	let ref, candidate;
 
-	if (node.actualNode.getAttribute('aria-labelledby')) {
+	if (node.getAttribute('aria-labelledby')) {
```
marcysutton: Being able to use `node` instead of `node.actualNode` makes our API much easier to work with. 👍

lib/commons/text/label.js
```diff
@@ -8,7 +8,7 @@
 text.label = function (node) {
 	var ref, candidate, doc;
 
-	candidate = aria.label(node);
+	candidate = aria.label(node.actualNode);
```
marcysutton: This doesn't seem consistent with the commons file above: `lib/commons/aria/label.js`

lib/checks/visibility/hidden-content.js
```diff
@@ -1,6 +1,6 @@
 const whitelist = ['SCRIPT', 'HEAD', 'TITLE', 'NOSCRIPT', 'STYLE', 'TEMPLATE'];
 if (!whitelist.includes(node.tagName.toUpperCase()) &&
-    axe.commons.dom.hasContent(virtualNode)) {
+    axe.commons.dom.hasContentVirtual(virtualNode)) {
```
marcysutton: Same as above. Should `hasContent` call `hasContentVirtual`?

lib/rules/color-contrast-matches.js
```diff
@@ -63,7 +63,7 @@ if (node.getAttribute('id')) {
 	}
 }
 
-if (axe.commons.text.visible(virtualNode, false, true) === '') {
+if (axe.commons.text.visibleVirtual(virtualNode, false, true) === '') {
```
marcysutton: Shouldn't `visible` call `visibleVirtual`? It would help backwards compatibility.

test/commons/dom/has-content-virtual.js
```diff
@@ -1,7 +1,7 @@
 /* global xit */
-describe('dom.hasContent', function () {
+describe('dom.hasContentVirtual', function () {
   'use strict';
-  var hasContent = axe.commons.dom.hasContent;
+  var hasContentVirtual = axe.commons.dom.hasContentVirtual;
```
marcysutton: In the test file for `test/commons/aria/label.js`, we're testing `label` directly. But here, we're testing `has-content-virtual`. Should we test both, or make the tests more consistent?

lib/checks/label/duplicate-img-label.js
WilcoFiers: That's not how I understood it. This made more sense to me, as this way, you don't have to look up the virtual node again, which is what you'd need to do if you just call `.visible`.

lib/commons/text/label.js
WilcoFiers: Right you are... fixing it :)

lib/rules/color-contrast-matches.js
WilcoFiers: yes

lib/checks/label/duplicate-img-label.js
marcysutton: It doesn't seem consistent with the test of this PR though.

lib/commons/dom/has-content-virtual.js
```diff
@@ -19,20 +19,29 @@ function hasChildTextNodes (elm) {
...
 };
+
+/**
+ * Like hasContentVirtual, except with a DOM Node
+ * IMPORTANT: This method requires the composed tree at axe._tree
+ *
+ * @param {Object}  DOM node
+ * @return boolean
+ */
+dom.hasContent = function hasContent (elm, noRecursion) {
```
dylanb: I think all of them should be implemented like this.

lib/commons/dom/is-offscreen.js
```diff
@@ -1,27 +1,26 @@
...
+	return true;
+}
 
-	var leftBoundary,
-		docElement = document.documentElement,
-		styl = window.getComputedStyle(element),
-		dir = window.getComputedStyle(document.body || docElement)
-			.getPropertyValue('direction'),
-		coords = dom.getElementCoordinates(element);
+dom.isOffscreen = function (element) {
```
dylanb: The fact that these style changes are mixed up in this pull request makes it much more difficult than it should be to review.

lib/commons/text/label-virtual.js
```diff
@@ -18,19 +18,22 @@ text.label = function (node) {
...
-	ref = axe.utils.getNodeFromTree(axe._tree[0], ref);
 	candidate = ref && text.visible(ref, true);
 	if (candidate) {
 		return candidate;
 	}
 
 	return null;
 };
+
+text.label = function (node) {
```
dylanb: This is good

lib/commons/text/visible-virtual.js
```diff
@@ -1,31 +1,26 @@
...
-			result += text.visible(child, screenReader);
+			return text.visibleVirtual(child, screenReader);
 		}
-	}
-
+	}).join('');
 	return text.sanitize(result);
 };
+
+text.visible = function (element, screenReader, noRecursing) {
```
dylanb: This is good

lib/rules/color-contrast-matches.js
marcysutton: Ok, we should make that change then. It would definitely help back compat to have the older function name referenced in this rule.

WilcoFiers: I had already done so: https://github.com/dequelabs/axe-core/pull/455/files?diff=unified#diff-4845776a844bde133665b6892f6ac14bR23

WilcoFiers:
@dylanb I'm not sure why you want us to add `aria.labelVirtual`. I switched `aria.label` from using virtualNode to DOM Node because everything that function does, it can do with regular DOM nodes. As far as I can see, `aria.label` shouldn't have been changed to take virtual nodes. It doesn't need them.

dylanb:
@WilcoFiers because it will not always fail when not set up correctly and I think that should be a requirement of this backwards compatibility.

WilcoFiers:
@dylanb @marcysutton Done, please have another look.

marcysutton:
I think this is the last PR to be merged before we do an alpha release. @WilcoFiers can you rebase to clean up the conflicts?

lib/rules/color-contrast-matches.js
marcysutton: Ah! Code review state fail.

PR #456, opened 2017-07-19
Add new rule, landmark-no-more-than-one-main

ssanaul:
This rule ensures that the document has no more than one main landmark.
Please let me know if anything needs to be changed. Thank you.

### DIFF
--- a/lib/rules/landmark-no-more-than-one-main.json
+++ b/lib/rules/landmark-no-more-than-one-main.json
@@ -2,11 +2,11 @@
   "id": "landmark-no-more-than-one-main",
   "selector": "html",
   "tags": [
-	"cat.best-practice"
+	"best-practice"
   ],
   "metadata": {
     "description": "Each document in a page must have no more than one main landmark, used to identify the main content.",
-    "help": "A <main> element or an element with a role='main' attribute defines a main landmark."
+    "help": "Each document in a page should not contain more than one main landmark."
   },
   "all": [],
   "any": [
--- a/lib/rules/landmark-no-more-than-one-main.json
+++ b/lib/rules/landmark-no-more-than-one-main.json
@@ -1,0 +1,16 @@
+{
+  "id": "landmark-no-more-than-one-main",
+  "selector": "html",
+  "tags": [
+	"cat.best-practice"
+  ],
+  "metadata": {
+    "description": "Each document in a page must have no more than one main landmark, used to identify the main content.",
+    "help": "A <main> element or an element with a role='main' attribute defines a main landmark."
+  },
+  "all": [],
+  "any": [
+    "has-no-more-than-one-main"
+  ],
+  "none": []
+}
--- a/test/checks/keyboard/has-no-more-than-one-main.js
+++ b/test/checks/keyboard/has-no-more-than-one-main.js
@@ -1,0 +1,37 @@
+describe('has-no-more-than-one-main', function () {
+	'use strict';
+
+	var fixture = document.getElementById('fixture');
+
+	afterEach(function () {
+		fixture.innerHTML = '';
+	});
+
+	it('should return false if there is more than one element with role main', function () {
+		var mainDiv = document.createElement('div');
+		mainDiv.setAttribute('role','main');
+		var anotherMain = document.createElement('div');
+		anotherMain.setAttribute('role','main');
+		fixture.appendChild(mainDiv);
+		fixture.appendChild(anotherMain);
+	});
+	
+	it('should return false if there is more than one main element', function () {
+		var mainLandmark = document.createElement('main');
+		var anotherMain = document.createElement('main');
+		fixture.appendChild(mainLandmark);
+		fixture.appendChild(anotherMain);
+	});
+
+	it('should return true if there is only one element with role main', function(){
+		var mainDiv = document.createElement('div');
+		mainDiv.setAttribute('role','main');
+		fixture.appendChild(mainDiv);
+	});
+	
+	it('should return true if there is only one main element', function(){
+		var mainLandmark = document.createElement('main');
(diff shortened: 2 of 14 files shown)


lib/rules/landmark-no-more-than-one-main.json
```diff
@@ -0,0 +1,16 @@
+{
+  "id": "landmark-no-more-than-one-main",
+  "selector": "html",
+  "tags": [
+	"cat.best-practice"
```
marcysutton: This should be just `best-practice`.

test/checks/keyboard/has-no-more-than-one-main.js
```diff
@@ -0,0 +1,37 @@
...
+		fixture.innerHTML = '';
+	});
+
+	it('should return false if there is more than one element with role main', function () {
+		var mainDiv = document.createElement('div');
+		mainDiv.setAttribute('role','main');
+		var anotherMain = document.createElement('div');
+		anotherMain.setAttribute('role','main');
+		fixture.appendChild(mainDiv);
+		fixture.appendChild(anotherMain);
```
marcysutton: You need assertions in these tests for them to test anything.

test/integration/full/landmark-no-more-than-one-main/frames/level1-fail.html
```diff
@@ -0,0 +1,12 @@
+<!doctype html>
+<html id="pass2">
+<head>
+	<meta charset="utf8">
+	<script src="/axe.js"></script>
+</head>
+<body>
+	<p>No main content should still pass</p>
```
marcysutton: Should this say `No main content should fail`?

test/integration/full/landmark-no-more-than-one-main/frames/level1-fail.html
```diff
@@ -0,0 +1,12 @@
+<!doctype html>
+<html id="pass2">
```
marcysutton: This says `pass2,` but the filename has fail in it. A little confusing.

test/integration/full/landmark-no-more-than-one-main/frames/level2.html
```diff
@@ -0,0 +1,10 @@
+<!doctype html>
+<html id="pass4">
+<head>
+	<meta charset="utf8">
+	<script src="/axe.js"></script>
+</head>
+<body>
+	<p>This should pass, too</p>
```
marcysutton: Does this pass because the Level 1 document has a main?

test/integration/full/landmark-no-more-than-one-main/frames/level1.html
```diff
@@ -0,0 +1,12 @@
+<!doctype html>
+<html id="pass2">
+<head>
+	<meta charset="utf8">
+	<script src="/axe.js"></script>
+</head>
+<body>
+	<p>No main content should still pass</p>
```
marcysutton: Should this fail? I wouldn't think a `<main>` in an iframe would count towards the main document. But maybe @dylanb or @WilcoFiers have some input?

jongund: This rule is making sure that there is no more than one main landmark in any one document (e.g. the container document or any of its iframes).   In general the ARIA authoring practices recommends only one main landmark per page.  But there is an exception for web sites that use iframes, for example web applications like portals often have content coming from other servers and services and are included in a web page within an iframe.  In this case where there is a high probability of distributed authorship, the ARIA Authoring Practices allows the main document and each iframe to have a main landmark.   

Tis rule is just checking that the main document or any documents in iframes have no more than one main landmark.

Hope this makes sense,
Jon

test/checks/keyboard/has-no-more-than-one-main.js
ssanaul: Oops.

test/integration/full/landmark-no-more-than-one-main/frames/level1-fail.html
ssanaul: Bad wording. I'll change it to 'No main content; should still pass.'

test/integration/full/landmark-no-more-than-one-main/frames/level2.html
ssanaul: It passes because this particular document does not have a main landmark. A document will pass under this rule as long as it has either just one main landmark or none at all.

test/integration/full/landmark-no-more-than-one-main/frames/level1.html
WilcoFiers: Hi @jongund, I think the concept is fine, although I think, given that you're also proposing an "at least one main" rule, those two can be part of the same rule. Have these two checks, but make them part of the same rule. What do you think?

jongund:
Wilco thanks for the response,

The use of iframes requires the use of two rules.   If you review the ARIA Authoring practices, they state that each frame can have at least one MAIN landmark.   
https://www.w3.org/TR/wai-aria-practices-1.1/#aria_lh_main 

So while most pages will have only one MAIN landmark, some will have multiple MAIN landmarks, but we only want to have at most one MAIN landmark in each document.

So the first rule is basically just looking for at least one main in the page including iframes, and the other rule is looking for more than one main in any one document.

Hope this makes sense.

Jon

jongund:
We will create a new pull request that combines the "at least one" and "no more than one" rules into a single rule and create a NEW pull request for "only-one" main landmark

WilcoFiers:
@jongund From your message I understand you're not continuing this PR and PR #418? If so, please close these and open the new one when it is ready.

ssanaul:
The check from this PR has been integrated into the rule, landmark-one-main.

PR #457, opened 2017-07-20
fix: get shadow DOM tests all passing

dylanb:


dylanb:
This is the first totally clean build of the shadow DOM branch since we broke the rules :-)

lib/core/utils/flattened-tree.js
```diff
@@ -123,6 +123,9 @@ axe.utils.getFlattenedTree = function (node, shadowId) {
 axe.utils.getNodeFromTree = function (vNode, node) {
 	var found;
 
+	if (vNode.actualNode === node) {
```
marcysutton: Since I'm still trying to keep these straight in my head, can you elaborate on what problem this solves?

dylanb: makes the additional test pass - specifically when some of our utilities look for the 'html' node in the vrtual tree, it was not being found.

PR #462, opened 2017-07-26
Add rule, landmark-main-is-top-level

ssanaul:
This rule ensures that each main landmark in a document is not nested within another landmark element. Once again, please let me know if there is anything you would like to see differently.


lib/checks/keyboard/main-is-top-level.js
```diff
@@ -0,0 +1,13 @@
+var landmarks = axe.commons.aria.getRolesByType('landmark');
+var parent = node.parentElement;
+while (parent){
+	var role = parent.getAttribute('role');
+	if (!role){
+		role = axe.commons.aria.implicitRole(parent);
+	}
+	if (role && landmarks.indexOf(role) >= 0){
```
WilcoFiers: Please use `!landmarks.includes(role)`.

lib/rules/landmark-no-more-than-one-main.json
```diff
@@ -0,0 +1,16 @@
+{
```
WilcoFiers: Please remove this rule from the PR.

lib/rules/landmark-main-is-top-level.json
```diff
@@ -0,0 +1,16 @@
+{
+  "id": "landmark-main-is-top-level",
+  "selector": "main,[role=main]",
+  "tags": [
+	"best-practice"
+  ],
+  "metadata": {
+    "description": "Each main landmark in a document must not be nested in another landmark",
+    "help": ""
```
WilcoFiers: Make sure to fill this out too. Description says what the rule does, help gives a hint on how to fix the issue.

lib/checks/keyboard/main-is-top-level.json
```diff
@@ -0,0 +1,11 @@
+{
+  "id": "main-is-top-level",
+  "evaluate": "main-is-top-level.js",
+  "metadata": {
+    "impact": "moderate",
+    "messages": {
+      "pass": "Main landmark is top level",
```
WilcoFiers: Please use a full sentence for both texts.

lib/checks/keyboard/main-is-top-level.js
```diff
@@ -0,0 +1,13 @@
+var landmarks = axe.commons.aria.getRolesByType('landmark');
+var parent = node.parentElement;
+while (parent){
+	var role = parent.getAttribute('role');
+	if (!role){
+		role = axe.commons.aria.implicitRole(parent);
```
WilcoFiers: I think this is too strict. There are roles that can certainly be permitted, off the top of my head: `form`, `application`, `presentation`. Probably a bunch more. Can you go through the list and make sure those kind of things get ignored?

lib/rules/landmark-main-is-top-level.json
marcysutton: The help text also goes in the aXe extension sidebar where we list out the violations.

lib/checks/keyboard/main-is-top-level.js
jongund: WIlco,

1. In ARIA 1.1 application and presentation roles are not considered landmarks.
2. In the case of a form landmark:
2.a Form landmarks should be fairly rare, there is discussion in the ARIA and HTML5 working groups that form landmarks should/must have an accessible name to be considered a landmark or at least useful, since without knowing what the form is about the navigation feature is somewhat limited since getting to form controls is easy from screen readers and makes the landmark more of a distraction than a useful feature.
2.b Forms are typically just one part of a page so should be in another landmark, the main landmark is what people will be looking for, not a form landmark.  In general Form landmarks should only be used if there are multiple forms on a page.

WilcoFiers: @jongund Both of those suggestions make sense to me. Seeing as it's still an open discussion if forms should be a landmark, make an exception for this rule. That one is particularly relevant for the `region` rule. You can adjust the type of application and presentation in `lib/commons/aria/index`.

jongund: Wilco,

Do you want us to modify in the file lib/commons/aria/index for role application:
type: 'landmark' to type: 'structure'

role presentation is already has type: 'structure' so no change need for it.

lib/checks/keyboard/main-is-top-level.js
jongund: We will make this change

jongund: @WilcoFiers 
We have updated the pull request with requested changes

lib/checks/keyboard/main-is-top-level.json
```diff
@@ -0,0 +1,11 @@
+{
+  "id": "main-is-top-level",
+  "evaluate": "main-is-top-level.js",
+  "metadata": {
+    "impact": "moderate",
+    "messages": {
+      "pass": "The main landmark is top level.",
```
WilcoFiers: This isn't a correct sentence, please use "is as the top level."

lib/rules/landmark-main-is-top-level.json
```diff
@@ -0,0 +1,16 @@
+{
+  "id": "landmark-main-is-top-level",
+  "selector": "main,[role=main]",
+  "tags": [
+	"best-practice"
+  ],
+  "metadata": {
+    "description": "Each main landmark in a document must not be nested in another landmark",
+    "help": "The main landmark in this document should not be contained within another landmark"
```
WilcoFiers: This is too wordy and won't fit well into the extension UI. I'd suggestion: "The main landmark should not be in another landmark"

test/integration/full/landmark-main-is-top-level/frames/level2-a.html
```diff
@@ -0,0 +1,15 @@
+<!doctype html>
+<html id="violation4">
+<head>
+	<meta charset="utf8">
+	<script src="/axe.js"></script>
+</head>
+<body>
+	<p>This iframe is also a violation</p>
+	<div role = "form">
```
WilcoFiers: This shouldn't be a failure. Neither should role="application". Make sure there is a test for that too.

lib/checks/keyboard/main-is-top-level.js
WilcoFiers: Use `.includes()` instead.

lib/checks/keyboard/main-is-top-level.js
```diff
@@ -0,0 +1,13 @@
+var landmarks = axe.commons.aria.getRolesByType('landmark');
+var parent = node.parentElement;
```
WilcoFiers: This PR has been open for a while now. Since than, we've merged shadow DOM in. It has a different method of getting at parent nodes. Please use `axe.commons.dom.getComposedParent()`. That way this check will look passed shadow DOM boundaries. You'll have to rebase to get the latest on develop into this branch.

jongund:
Wilco,

I am meeting with the student later today and we will:


  1.  Make the make the changes outlined in this e-mail
  2.  Update the pull requests to remove the obsolete proposed rules

If there are still additional issues after the changes maybe we can schedule a call to work on the remaining issues.

Jon

jongund:
Wilco,

In our discussion:
jongund<https://github.com/jongund> 6 days ago <https://github.com/dequelabs/axe-core/pull/462#discussion_r133823029>
WIlco,

  1.  In ARIA 1.1 application and presentation roles are not considered landmarks.
  2.  In the case of a form landmark:
2.a Form landmarks should be fairly rare, there is discussion in the ARIA and HTML5 working groups that form landmarks should/must have an accessible name to be considered a landmark or at least useful, since without knowing what the form is about the navigation feature is somewhat limited since getting to form controls is easy from screen readers and makes the landmark more of a distraction than a useful feature.
2.b Forms are typically just one part of a page so should be in another landmark, the main landmark is what people will be looking for, not a form landmark. In general Form landmarks should only be used if there are multiple forms on a page.
Top of Form
Bottom of Form
[WilcoFiers]<https://github.com/WilcoFiers>
WilcoFiers<https://github.com/WilcoFiers> 5 days ago <https://github.com/dequelabs/axe-core/pull/462#discussion_r133966810>
Contributor
@jongund<https://github.com/jongund> Both of those suggestions make sense to me. Seeing as it's still an open discussion if forms should be a landmark, make an exception for this rule. That one is particularly relevant for the region rule. You can adjust the type of application and presentation in lib/commons/aria/index.


It sounded like you agreed to change “application” in /lib/commons/aria/index.js to not be a landmark and that is part of our pull request and that you seemed to agree that FORM landmark restriction was also OK, since FORM landmarks should be rare in practice and a MAIN landmark would probably be better for identifying a page that is mostly a form.

So the test for bening contained in a FORM landmark should fail and APPLICATION does need to be tested because it is not a landmark in ARIA 1.1.

Jon

________________________________


In test/integration/full/landmark-main-is-top-level/frames/level2-a.html<https://github.com/dequelabs/axe-core/pull/462#discussion_r134702158>:

> @@ -0,0 +1,15 @@

+<!doctype html>

+<html id="violation4">

+<head>

+       <meta charset="utf8">

+       <script src="/axe.js"></script>

+</head>

+<body>

+       <p>This iframe is also a violation</p>

+       <div role = "form">

This shouldn't be a failure. Neither should role="application". Make sure there is a test for that too.

jongund:
We have updated the pull request, including changing the messaging and using axe.commons.getComposedParent.

The only outstanding issue seems to be whether a MAIN landmark can be contained by a FORM landmark.   Based on FORM landmarks being rare and if a FORM is the primary content of a page, the MAIN landmark would probably be as good if not better , since users ae looking for MAIN landmarks.   So the tests still view FORM as a failure.   Do we want to discuss it in a teleconference?

lib/checks/keyboard/main-is-top-level.js
WilcoFiers: @jongund This line still uses `node.parentElement`, please fix this too.

jongund:
We updated the rule to use:
1. axe.commons.getComposedParent
2. Updated the rule now ignores the <form> element for the purpose of looking at parent nodes for FORM landmarks.   The only way a <form> element would be tested as a FORM landmark is when the <form> element has an explicit role=form attribute.

WilcoFiers:
@jongund The tests aren't passing yet. Be sure to run `grunt test-fast` before you commit code for review. The other PR had this problem too.

jongund:
Anymore questions or issues with this pull request?

What is the status of accepting this rule?

WilcoFiers:
@jongund We're a little busy with the 2.4 releases at the moment. We certainly haven't forgotten about these PRs and will get to reviewing them as soon as possible.

test/integration/full/landmark-main-is-top-level/landmark-main-is-top-level-fail.js
```diff
@@ -0,0 +1,38 @@
...
+				assert.isNull(err);
+				results = r;
+				done();
+			});
+		});
+	});
+
+	describe('violations', function () {
+		it('should find 1', function () {
+			assert.lengthOf(results.violations, 1);
```
WilcoFiers: Add a node count check too:

`assert.lengthOf(results.violations[0].nodes, 2);`

lib/commons/aria/index.js
```diff
@@ -189,7 +189,7 @@ lookupTables.role = {
 		context: null
 	},
 	'application': {
-		type: 'landmark',
```
WilcoFiers: This is causing the following test to fail, be sure to fix that:

`1) table.isDataTable should be true if the table has a landmark role application:`

ssanaul: @WilcoFiers 
In ARIA 1.1, 'application' is not a landmark role.
https://www.w3.org/TR/wai-aria-1.1/#application
The new superclass role is 'structure'. 
Can you update the 'develop' branch to meet the ARIA 1.1 specification?

ssanaul:
`  table-fake-caption
    table-fake-caption test
      passes
        √ should find ["#pass1"]
        √ should find ["#pass2"]
        √ should not return other results
      violations
        √ should find ["#fail1"]
        √ should find ["#fail2"]
        √ should not return other results

  td-has-header
    td-has-header test

Warning: PhantomJS timed out, possibly due to a missing Mocha run() call. Use --force to continue.

Aborted due to warnings.`

This is the message I'm receiving currently. Is this what you're seeing as well @WilcoFiers ?

marcysutton:
@ssanaul you can always fire up the tests in a browser using `grunt dev` to see if there are any errors in the JavaScript console. That's usually how I debug those failures

WilcoFiers:
@ssanaul @jongund What's the status of this PR?

jongund:
Wilco,

Sulaiman is trying to setup a meeting with Marcy Sutton to try to debug why the integration tests are failing.

It seems we need a little more help with configuring the test cases correctly.

Do you have any time to meet with us this week to review our code with us?

Jon

ssanaul:
Is there anything I've excluded from this PR? I thought I had addressed all review comments.

marcysutton:
I'm still seeing two pretty simple comments from Wilco that weren't addressed, otherwise looks good to me.

> test/integration/full/landmark-main-is-top-level/frames/level2-a.html
> test/integration/full/landmark-main-is-top-level/landmark-main-is-top-level-fail.js

ssanaul:
Ok, I think that last commit should have resolved everything. Although, I think the suggested node count check `assert.lengthOf(results.violations[0].nodes, 2);` should be `assert.lengthOf(results.violations[0].nodes, 4);`

lib/checks/keyboard/main-is-top-level.js
```diff
@@ -0,0 +1,13 @@
+var landmarks = axe.commons.aria.getRolesByType('landmark');
+var parent = axe.commons.dom.getComposedParent(node);
+while (parent){
+	var role = parent.getAttribute('role');
+	if (!role && (parent.tagName.toLowerCase() !== 'form')){
```
WilcoFiers: This needs to work with `role=form` too.

test/integration/full/landmark-main-is-top-level/landmark-main-is-top-level-fail.js
WilcoFiers: @ssanaul This issue isn't addressed yet.

ssanaul: I thought 3e5c6ad6d39e6b5de278a2497587225042442431 resolved it. It checks for 4 nodes because I believe there are 4 violations.

lib/checks/keyboard/main-is-top-level.js
ssanaul: Does line 5 not accomplish that?

dylanb:
@ssanaul thank you for your contribution!!!

PR #465, opened 2017-08-01
fix: incomplete results should have impact

WilcoFiers:
This solves a problem we had with the axe-devtools extension, where, if there wasn’t also a violation for that same rule, the impact wouldn’t be available.


marcysutton:
This looks good to me.

PR #466, opened 2017-08-01
fix: Exclude `any` checks from output if one passed

### LINKED ISSUE #446: Remove the remaining `any` checks from a result
When a rule such as labels has both `any` and `none` checks, if the `any` has a pass, but the rule fails because of the `none`, all check results from `any` must be removed, not just the one check that passed. We saw this problem in the following code:

```html
<input type="text" id="fff">
<label for="fff">a</label>
<label for="fff">b</label>
```

WilcoFiers:
This should resolve #446.


marcysutton:
Nice! This also fixes an internal issue we found on a client site. (WWD-820)

PR #468, opened 2017-08-02
feat(aria): Support progressive ARIA 1.1 attributes / roles

### LINKED ISSUE #350: Potential issue with Aria 1.1 role=feed
I'm not totally sure if this is an issue as https://github.com/dequelabs/axe-core/issues/91 states all ARIA 1.1 roles are supported.

However [`role=feed`](https://www.w3.org/TR/wai-aria-1.1/#feed) is marked as an error in Lighthouse. [I opened an issue there](https://github.com/GoogleChrome/lighthouse/issues/2468#issuecomment-307170188) and was directed here.

I looked in https://github.com/dequelabs/axe-core/blob/9ce24bc9babdf33fff10dfeb4ef1c6d16f57df08/lib/commons/aria/index.js and https://github.com/dequelabs/axe-core/blob/020fd150ed0bf0702a89d19712bd38ea0c3817d8/test/integration/rules/aria-roles/aria-roles.html which seem like good places to look, but I'm not 100% sure, and `feed` isn't there.

WilcoFiers:
Add support for role=feed, role=term, aria-placeholder, aria-modal,
aria-current and aria-keyshortcuts.

This PR closes #350.

### DIFF
--- a/lib/commons/aria/index.js
+++ b/lib/commons/aria/index.js
@@ -71,6 +71,9 @@ lookupTables.attributes = {
 		type: 'nmtoken',
 		values: ['true', 'false', 'spelling', 'grammar']
 	},
+	'aria-keyshortcuts': {
+		type: 'string'
+	},
 	'aria-label': {
 		type: 'string'
 	},
@@ -83,6 +86,10 @@ lookupTables.attributes = {
 	'aria-live': {
 		type: 'nmtoken',
 		values: ['off', 'polite', 'assertive']
+	},
+	'aria-modal': {
+		type: 'boolean',
+		values: ['true', 'false']
 	},
 	'aria-multiline': {
 		type: 'boolean',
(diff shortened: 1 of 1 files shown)


marcysutton:
How's the support for all of these new things?

WilcoFiers:
@marcysutton as far as I know none of them are supported currently. The reason we're adding them is because it's better to use them now, even if they aren't supported, since they will presumably be implemented at some point. Also, unlike some of the other features new to ARIA 1.1, these features aren't likely to cause accessibility problems if they're not supported. They simply add to the user experience when support is there.

cycomachead:
They're also used in some of the ARIA authoring practices guide. It's somewhat confusing to be following guidelines for (in my case Modals) and seeing my specs fail because I'm using a new attribute.

PR #469, opened 2017-08-03
docs: document help text and incomplete messages

marcysutton:
Now that people are submitting rules, I thought we could use more documentation about how help text is used and how to create incomplete messages.

### DIFF
--- a/doc/rules.md
+++ b/doc/rules.md
@@ -31,16 +31,62 @@ The actual testing of elements in axe-core is done by checks. A rule has one or 
 | enabled              | Does the rule run by default
 | tags                 | Grouping for the rule, such as wcag2a, best-practice
 | metadata.description | Description of what a rule does
-| metadata.help        | Description of how to resolve an issue
+| metadata.help        | Short description of a violation, used in the aXe extension sidebar
 
 ## Check Properties
 
-| Prop. Name      | Description
-|-----------------|-----------------
-| id              | Unique identifier for the check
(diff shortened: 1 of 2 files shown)


doc/rules.md
```diff
@@ -31,16 +31,62 @@ The actual testing of elements in axe-core is done by checks. A rule has one or
...
+| Prop. Name         | Description
+|--------------------|-----------------
+| id                 | Unique identifier for the check
+| evaluate           | Evaluating function, returning a boolean value
+| options            | Configurable value for the check
+| after              | Cleanup function, run after check is done
+| metadata impact    | "minor", "serious", "critical"
+| metadata pass      | Describes why the check passed
+| metadata fail      | Describes why the check failed
+| metadata incomplete| Describes why the check returned undefined
```
WilcoFiers: should this say, why the check didn't complete?

doc/rules.md
```diff
@@ -31,16 +31,62 @@ The actual testing of elements in axe-core is done by checks. A rule has one or
...
+
+Incomplete results occur when axe-core can’t produce a clear pass or fail result,
+giving users the opportunity to review it manually. Incomplete messages can take
+the form of a string, or an object with arbitrary keys matching the data returned
+from the check.
+
+Incomplete is optional for checks, while pass and fail are required.
```
WilcoFiers: Is that true? It shouldn't be. Pass is required, fail/incomplete should both be optional. IMO this should just say that these are required if this is a result the check outputs.

Issue #401, opened 2017-06-30
Update documentation for all commons, checks and rules for Shadow DOM


marcysutton:
Wherever there's documentation on writing rules or using commons, we need to make sure implementers know their responsibilities when it comes to Shadow DOM.

PR #471, opened 2017-08-03
Shadow DOM docs

marcysutton:
Should close https://github.com/dequelabs/axe-core/issues/401. Feedback welcome!


doc/developer-guide.md
```diff
@@ -131,11 +137,59 @@ A CheckResult has the following properties:
...
+including the rest of the commons code. The naming of this special function
+should contain the original commons function name with `Virtual` added to signify
+it expects to operate on a virtual DOM tree.
+
+Let’s look at an example:
+
+```javascript
+axe.commons.text.accessibleText = function (element, inLabelledbyContext) {
+  let virtualNode = axe.utils.getNodeFromTree(axe._tree[0]); // throws an exception on purpose if axe._tree not correct
+  return axe.commons.text.virtualAccessibleText(virtualNode, inLabelledbyContext);
```
WilcoFiers: This would be `text.accessibleTextVirtual`, although I just realized I messed up and didn't rename that one. I made a ticket for it.

doc/developer-guide.md
```diff
@@ -131,11 +137,59 @@ A CheckResult has the following properties:
...
+}
+
+axe.commons.text.virtualAccessibleText = function (element, inLabelledbyContext) {
+  // rest of the commons code minus the virtual tree lookup, since it’s passed in
+}
+```
+
+`virtualAccessibleText` would never be called directly; rather, it would be called
```
WilcoFiers: That's not true. We always call the virtual functions if we've got a virtual node we can use. Only if we don't have one do we use the lookup function.

marcysutton: Wouldn't that make our rules/checks less backwards-compatible?

WilcoFiers:
Nah, new rules should use them when they can, as that's more performant. We
have the old ones mostly to stay backward competable.

doc/developer-guide.md
```diff
@@ -131,11 +137,59 @@ A CheckResult has the following properties:
...
+```javascript
+{
+  actualNode: <HTMLElement>,
+  children: <Array>,
+  shadowId: <String>
+}
+```
+
+A virtualNode is an object containing an HTML DOM element (`actualNode`), an
+array of child virtualNodes, and, if provided when querying the virtual tree, a
```
dylanb: The shadowID indicates whether the node is in a shadow root and if it is, which one it is inside.

doc/API.md
```diff
@@ -608,6 +613,120 @@ In axe-core v1 the main method for axe was `axe.a11yCheck()`. This method was re
 - .a11yCheck requires a context object, and so will not fall back to the document root.
 - .a11yCheck does not return a Promise.
 
+### Virtual DOM Utilities
+
+#### API Name: axe.utils.getFlattenedTree
```
dylanb: However, if we do, then we should point ot the algorithm definition in the HTML spec (see the source code comments for the link)

doc/developer-guide.md
```diff
@@ -115,9 +121,9 @@ return results.filter(function (r) {
 });
 ```
 
-#### Pass and Fail Templates
+#### Pass, Fail and Incomplete Templates
 
-Occasionally, you may want to add additional information about why a Check passed or failed into its message.  For example, the [aria-valid-attr](../lib/checks/aria/valid-attr.json) will add information about any invalid ARIA attributes to its fail message.  The message uses the [doT.js](http://olado.github.io/doT/) and is compiled to a JavaScript function at build-time.  In the Check message, you have access to the `CheckResult` as `it`.
+Occasionally, you may want to add additional information about why a Check passed, failed or returned undefined into its message.  For example, the [aria-valid-attr](../lib/checks/aria/valid-attr.json) will add information about any invalid ARIA attributes to its fail message.  The message uses the [doT.js](http://olado.github.io/doT/) and is compiled to a JavaScript function at build-time.  In the Check message, you have access to the `CheckResult` as `it`.
```
marcysutton: There is one in my other PR: #469

doc/API.md
```diff
@@ -21,6 +21,11 @@
 	1. [API Name: axe.registerPlugin](#api-name-axeregisterplugin)
 	1. [API Name: axe.cleanup](#api-name-axecleanup)
 	1. [API Name: axe.a11yCheck](#api-name-axea11ycheck)
+	1. [Virtual DOM Utilities](#virtual-dom-utilities)
+		1. [API Name: axe.utils.getFlattenedTree](#api-name-axeutilsgetflattenedtree)
+		2. [API Name: axe.utils.getNodeFromTree](#api-name-axeutilsgetnodefromtree)
+		3. [API Name: axe.utils.querySelectorAll](#api-name-axeutilsqueryselectorall)
+		4. [API Name: axe._tree](#api-name-axe_tree)
```
marcysutton: I might just remove it for now so we don't hold up the release. Because I have questions about how that would work, since it gets set internally during the audit: 1. when would you set or clear it? 2. what should happen if you try to manipulate it at the wrong time?

doc/API.md
```diff
@@ -608,6 +613,120 @@ In axe-core v1 the main method for axe was `axe.a11yCheck()`. This method was re
...
+##### Synopsis
+
+```javascript
+var element = document.body;
+axe.utils.getFlattenedTree(element, shadowId)
+```
+
+##### Parameters
+ - `node` – node. The current HTML node for which you want a flattened DOM tree.
+ - `shadowId` – string(optional). ID of the shadow DOM that is the closest shadow ancestor of the node
```
marcysutton: I'm not quite following. Can't we set it as `undefined` if it doesn't get passed in? This may also be moot if we drop documentation of this function.

doc/API.md
```diff
@@ -608,6 +613,120 @@ In axe-core v1 the main method for axe was `axe.a11yCheck()`. This method was re
...
+  children: [virtualNodes],
+  shadowId: undefined
+}
+```
+
+#### API Name: axe.utils.querySelectorAll
+
+##### Description
+
+A querySelectorAll implementation that works on the virtual DOM and Shadow DOM by manually walking the tree instead of relying on DOM API methods which don't step into Shadow DOM.
```
marcysutton: > It walks the flattened tree

Do you mean your implementation? Or `document.querySelectorAll`? I assume you meant the former, but I wanted to make sure.

doc/API.md
marcysutton: We need something to pass to `axe.utils.getNodeFromTree`, so I'd vote for documenting it and dropping `axe._tree`. But we could also do as Dylan suggests and create a method to set and clear the cache. I had some questions about that in a separate comment.

marcysutton:
> There are a few more methods I think we should add. The test utilities are important, as well as dom.getComposedParent, dom.getRootNode and maybe also dom.findUp.

I didn't add those because they are commons functions and not utils, but it sounds like we had a request for documenting the rest of the commons. I won't go through all of those, but I can start with these few and the testUtils.

doc/API.md
dylanb: nbd - I was just suggesting adding a note to say that it is not intended to be directly used by an external caller

doc/API.md
dylanb: I mean this implementation walks the flattened tree. It is like `document|Element.querySelectorAll` except for the flattened tree rather than the DOM (light DOM).

doc/API.md
```diff
@@ -608,6 +618,263 @@ In axe-core v1 the main method for axe was `axe.a11yCheck()`. This method was re
...
 - .a11yCheck does not return a Promise.
 
-## Section 3: Example Reference
+### Virtual DOM Utilities
+
+#### API Name: axe.utils.getFlattenedTree
+
+##### Description
+
+Recursvely return an array containing the virtual DOM tree for the node specified, excluding comment nodes and shadow DOM nodes `<content>` and `<slot>`. This method will return a composed tree containing both light and shadow DOM, if applicable.
```
dylanb: `composed tree` is an outdated term. The new term is `flattened tree`.

marcysutton: What about `axe.commons.dom.getComposedParent`? Is that still good?

doc/developer-guide.md
marcysutton: I added an example in #469.

marcysutton:
That's fine by me, I'll move the test utilities over there too. Thanks!

marcysutton:
Ok, I moved them to the developer guide and added a table of contents to it.

PR #476, opened 2017-08-08
feat: Add sri-history file and update process

WilcoFiers:
Since axe-core is injected into many many web pages, it is very important that we’re able to check the integrity of the file we’re injecting. Keeping the SRI of all previously published versions around makes this much easier.

https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity

### DIFF
--- a/sri-history.json
+++ b/sri-history.json
@@ -1,0 +1,54 @@
+{
+	"1.0.1": {
+		"axe.js": "sha256-F14wfpbaL6+ZafS2ufmz74/R6CD1L777gCOJVuLn5Ao=",
+		"axe.min.js": "sha256-GfbQgd6aKJKQT21HaQfVUuGEbMUfSKhi7RsWrrih7SM="
+	},
+	"1.1.0": {
+		"axe.js": "sha256-nG9+LLksY/HS6yG4x0iS08U7RAmTucy50uXQS5ndQnM=",
+		"axe.min.js": "sha256-5Lfk3/s+ieHttrDTOCVuepGjDgH2D1neTDDtXnJdIT0="
+	},
+	"1.1.1": {
+		"axe.js": "sha256-K0MLUClls79s14CT01F82LPc+ZubTBwt6fZDdN5iyFw=",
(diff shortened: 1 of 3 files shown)


sri-history.json
```diff
@@ -0,0 +1,54 @@
...
+	},
+	"2.2.3": {
+		"axe.js": "sha256-dV8RkE0hyyzj7qgValVkoP7Rtu8789BmbyeHZmidvqA=",
+		"axe.min.js": "sha256-BrNWjOcDL265Me8eKQosGWerJ6ju2g+G9+RvMWiBGOA="
+	},
+	"2.3.0": {
+		"axe.js": "sha256-rkyHB2lHjs+tissQLBUxuxIvWlzRbS4f4cdaH+TjQvo=",
+		"axe.min.js": "sha256-MGWkallV18uw6bSq6w8cjbGsf9v4rJtXP+NDtMEbO14="
+	},
+	"2.3.1": {
```
marcysutton: When I ran it my hashes changed for aXe-2.3.1, do you know why that is?

WilcoFiers: Yeah, because this branched from dev, which has changes that aren't in 2.3.1. If you replace your axe.js file with the 2.3.1 release it'll match.

marcysutton:
We need docs for this, since it isn't immediately clear when we should be running this command. It already impacted our latest releases of aXe-core.

WilcoFiers:
@marcysutton Was it because we're not using `npm version` to bump the version? This PR should warn you before publishing with the wrong SRI. Did that not happen?

marcysutton:
`npm version` doesn't update all the necessary files, and I haven't been using it. So yeah, that's probably the problem. We need to document this as part of our Build/Release process guidelines.

WilcoFiers:
I can add a comment to the build file explaining it further. We shouldn't be doing this stuff manually anyway. We'll have Attest-master ready before the next release, so we'll automate this step. But yeah, not using `npm version` is probably where this went wrong than. Is that because of the changelog thing?

Issue #282, opened 2017-02-09
ARIA 1.1: Update allowed values of aria-haspopup


SiTaggart:
With ARIA 1.1 comes a new set of allowed values for `aria-haspopup`. `haspopup` is no longer a boolean value and allows the following values:

- true
- false
- menu
- listbox
- tree
- grid 
- dialog

This can be referenced from the [aria-haspopup specification](https://www.w3.org/TR/wai-aria-1.1/#aria-haspopup)

Requires a change in type and additional values https://github.com/dequelabs/axe-core/blob/0fc66099816786170a4ad9e246a6b21437e72f6c/lib/commons/aria/index.js#L60

marcysutton:
Thanks, we'll incorporate them soon!

marcysutton:
Fixed with #479.

Issue #276, opened 2017-01-05
Incorrect reporting that Listbox does not allow aria-orientation


SiTaggart:
When applying `aria-orientation` to a component with role `listbox`, aXe will report a critical violation: 

- `Ensures ARIA attributes are allowed for an element's role`
- Summary: `ARIA attribute is not allowed: aria-orientation="horizontal"`

Given that the Implicit value for the role has a default `aria-orientation` of `vertical` it would seem this is a bug. You should be able to at least provide another orientation value that is not the default, especially if your listbox goes from left to right.

http://w3c.github.io/aria/aria/aria.html#listbox

AutoSponge:
Same for `tablist`. I think the problem is the use of [WAI-ARIA States and Properties Taxonomy](http://www.w3.org/TR/wai-aria/states_and_properties#state_prop_taxonomy) and [WAI-ARIA Definition of Roles](http://www.w3.org/TR/wai-aria/roles#role_definitions) in the [rules](https://dequeuniversity.com/rules/axe/2.1/aria-allowed-attr?application=axeAPI) instead of [WAI-ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/#tablist). There seem to be [conflicts](https://github.com/w3c/wcag/issues/259).

marcysutton:
Fixed with #479.

Issue #275, opened 2016-12-31
Incorrect reporting on role="treegrid" and it's owned children


SiTaggart:
Currently aXe will report that a `treegrid` requires a `treeitem` child. This is incorrect.

According to the ARIA Spec, [treegrid](https://www.w3.org/TR/wai-aria-1.1/#treegrid), similarly to a `grid`, requires a `row` or `rowgroup`. This means that [line 853](https://github.com/dequelabs/axe-core/blob/master/lib/commons/aria/index.js#L853) which declares that `treeitem` is `owned` by `treegrid`, should be closer to the `grid` rules on [line 341](https://github.com/dequelabs/axe-core/blob/master/lib/commons/aria/index.js#L341), which declares the role owns one of `row` or `rowgroup`.

Similarly it means that the required context for `treeitem` _should not_ include `treegrid` on [line 865](https://github.com/dequelabs/axe-core/blob/master/lib/commons/aria/index.js#L865), as per the ARIA 1.1 spec on `treeitem` https://www.w3.org/TR/wai-aria-1.1/#treeitem.

I can PR this if I get a spare moment, in the meantime just documenting it here.

marcysutton:
Fixed with #479.

PR #479, opened 2017-08-09
Aria 1.1 fixes

WilcoFiers:
This closes a few issues: #275, #276, #282.

### DIFF
--- a/lib/commons/aria/index.js
+++ b/lib/commons/aria/index.js
@@ -261,7 +261,7 @@ lookupTables.role = {
 	'combobox': {
 		type: 'composite',
 		attributes:  {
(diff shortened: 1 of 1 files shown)


lib/commons/aria/index.js
```diff
@@ -897,7 +897,7 @@ lookupTables.role = {
 		},
 		owned: null,
 		nameFrom: ['author', 'contents'],
-		context: ['treegrid', 'tree']
```
marcysutton: `treegrid` isn't a thing anymore?

WilcoFiers: It is, but treeitem doesn't require it as context:
https://www.w3.org/TR/wai-aria-1.1/#treeitem

marcysutton:
There were no tests for these changes, either. I'll see if I can add some.

Issue #160, opened 2016-01-08
fix false positive related to role="combobox"


dylanb:
Has been reported in https://github.com/dequelabs/axe-firefox-devtools/issues/33

dylanb:
The solution should be the following:

If the element on which the role has been placed is a native input element, then make the required children optional

WilcoFiers:
I raised this issue on html-aria and Steve Faulkner got back on it. We have a different issue. Axe has two 'types' of 'owned' propertyes, 'all' and 'one'. All means every one of the roles is required, one means just one of them. ARIA doesn't actually ever require all children to be present. combobox is the only role in Axe which is set to require multiple children. Other uses of 'all' only has one role.

The fix is to change this to 'one'. Though I think it's probably better to also remove this difference all together as it adds unnecessary complexity.

dylanb:
If there is a combobox implementation that does not use a standard input element and also does not have a textbox child, then this will have severe compatibility issues when a screen reader is turned on. 

For that reason, I think we should still raise the issue if there is no native element (note: this would be a temporary solution until the issue gets resolved in ARIA).

johncaseysmith:
Hey @dylanb, any updates on this? Does the [previously mentioned solution](https://github.com/dequelabs/axe-firefox-devtools/issues/33#issuecomment-171918557) still apply?

adamliptrot-oc:
Hi, what's the current situation with this? I'm running into this issue with automated testing.

isner:
Proposed solution: https://github.com/dequelabs/axe-core/pull/481

marcysutton:
Fixed with #481.

PR #481, opened 2017-08-09
fix(checks/aria/required-children): add exception for native input combobox missing textbox

isner:
Remove textbox from the list of missing roles when checking a combobox node that is a native input element

https://github.com/dequelabs/axe-core/issues/160

### DIFF
--- a/test/checks/aria/required-children.js
+++ b/test/checks/aria/required-children.js
@@ -49,6 +49,12 @@ describe('aria-required-children', function () {
 		assert.isTrue(checks['aria-required-children'].evaluate.call(checkContext, node));
 	});
 
+	it('should pass a native input with role comboxbox when missing child is role textbox', function () {
+		fixture.innerHTML = '<input type="text" role="combobox" aria-owns="listbox" id="target"></div><p role="listbox" id="listbox">Nothing here.</p>';
+		var node = fixture.querySelector('#target');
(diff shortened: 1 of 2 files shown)


test/checks/aria/required-children.js
```diff
@@ -49,6 +49,12 @@ describe('aria-required-children', function () {
 		assert.isTrue(checks['aria-required-children'].evaluate.call(checkContext, node));
 	});
 
+	it('should pass a native input with role comboxbox when missing child is role textbox', function () {
+		fixture.innerHTML = '<input type="text" role="combobox" aria-owns="listbox" id="target"></div><p role="listbox" id="listbox">Nothing here.</p>';
```
marcysutton: A note that one thing I had to do when merging this to develop (which now includes Shadow DOM) was to use `checkSetup` and `evaluate.apply` instead of `call`.

```
var params = checkSetup('<input type="text" role="combobox" aria-owns="listbox" id="target"></div><p role="listbox" id="listbox">Nothing here.</p>');
assert.isTrue(checks['aria-required-children'].evaluate.apply(checkContext, params));
```

isner: Good to know! I assume that change had already been made to all other tests in your Shadow DOM branch. Thanks for bringing this test into the shadowy future.

marcysutton: It's cool! The old method was still needed for develop-2x. Hopefully we'll be able to kill off that branch before too long.

PR #486, opened 2017-08-15
Precommit hook to prevent bad commit messages

marcysutton:
This hook works on the command line only–for some reason GitX throws an error (a known issue according to the [angular-precommit source code](https://github.com/ajoslin/angular-precommit/blob/master/index.js#L25)). 

I've asked the axe-core team to respond with preferred methods of writing commits so I can test different applications. But this will install the hook automatically, preventing commits which don't adhere to our [commit policy](https://github.com/dequelabs/axe-core/blob/develop/doc/code-submission-guidelines.md#git-commits) (which follows the Angular style).

We can use this in our other projects, too, since it doesn't depend on any specific linter.

### DIFF
--- a/bin/postinstall.sh
+++ b/bin/postinstall.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-if [ -d ".git" ]; then
+if [ -d ".git/hooks" ]; then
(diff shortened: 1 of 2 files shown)


marcysutton:
Circle crashed earlier but it seems to be working now, yay!

PR #487, opened 2017-08-16
Add tool for automating version bumps and changelog generation

marcysutton:
#This tool seems exactly like what we need! https://github.com/conventional-changelog/standard-version

It bumps the version automatically, but still gives us the opportunity to review before publishing–which was nice, because it tried to bump it to 3.1.0. For specific versions, we can use `npm run version -- --release-as 3.0.0` or `npm run version -- prerelease alpha`. It will create a tag for the version which we can then push and publish.

I had to reformat our existing changelog but having it pull in commit messages automatically will be a lot less work overall. The changelog content for 3.0.0-alpha.1 is lengthier this time around (and it duplicates our last release a bit because it's the first time we're using the tool), but it's also more accurate as long as we document changes in our commits. One thing we need to do going forward is to add `BREAKING CHANGE: ` to the footer of our commit messages, otherwise they won't get documented as breaking changes in the changelog (same as doing it manually–I could have missed them).

An unresolved error I got was that package-lock.json is ignored in .gitignore. Should we modify that behavior and start checking it in now? 

Closes https://dequesrc.atlassian.net/browse/WWD-819


marcysutton:
Since I'm publishing again today I'm going to merge this PR as I need it. No more manual changelogs!

PR #490, opened 2017-08-17
fix(docs): add minor spelling fix

rrgayhart:
correct misspelling in CONTRIBUTING doc

I don't think I've ever spelled license correctly the first time 😊


WilcoFiers:
Thanks!

PR #492, opened 2017-08-18
fix: use frame query that supports shadow dom

isner:
Supports [WWD-845] https://dequesrc.atlassian.net/browse/WWD-845


lib/core/public/cleanup-plugins.js
```diff
@@ -23,7 +23,7 @@ function cleanupPlugins(resolve, reject) {
     });
   });
 
-  axe.utils.toArray(document.querySelectorAll('frame, iframe')).forEach(function (frame) {
+  axe.utils.getFrames().forEach(function (frame) {
```
WilcoFiers: You should be able to use `axe.utils.querySelectorAll(axe._tree, 'frame, iframe')` instead.

isner: Thanks. I'll give that a try. This PR probably won't be necessary at all.

isner:
This enables `axe.cleanup` to target frames embedded in shadow roots. This is necessary to support clearing highlights in such frames.

lib/core/public/cleanup-plugins.js
```diff
@@ -23,14 +23,18 @@ function cleanupPlugins(resolve, reject) {
     });
   });
 
-  axe.utils.toArray(document.querySelectorAll('frame, iframe')).forEach(function (frame) {
+  var flattenedTree = axe.utils.getFlattenedTree(document.body)
```
isner: I had to create a new flattened tree to support this functionality. The only other place in which a flattened tree is created is during the audit itself. Since `axe.cleanup` can be called at any arbitrary time, we can't reuse the same tree.

lib/core/public/cleanup-plugins.js
```diff
@@ -23,14 +23,18 @@ function cleanupPlugins(resolve, reject) {
...
+  axe.utils.querySelectorAll(flattenedTree, 'iframe, frame').forEach(function (node) {
     q.defer(function (res, rej) {
-      return axe.utils.sendCommandToFrame(frame, {
+      return axe.utils.sendCommandToFrame(node.actualNode, {
         command: 'cleanup-plugin'
       }, res, rej);
     });
   });
 
+  flattenedTree = undefined
```
WilcoFiers: We don't need this. Garbage collector will do this for you.

isner: Will do. I'm curious, though, why we need to manually free up the memory here: https://github.com/dequelabs/axe-core/blob/develop/lib/core/base/audit.js#L182

lib/core/public/cleanup-plugins.js
```diff
@@ -23,9 +23,11 @@ function cleanupPlugins(resolve, reject) {
     });
   });
 
-  axe.utils.toArray(document.querySelectorAll('frame, iframe')).forEach(function (frame) {
+  var flattenedTree = axe.utils.getFlattenedTree(document.body)
+
+  axe.utils.querySelectorAll(flattenedTree, 'iframe, frame').forEach(function (node) {
     q.defer(function (res, rej) {
-      return axe.utils.sendCommandToFrame(frame, {
+      return axe.utils.sendCommandToFrame(node.actualNode, {
```
marcysutton: This is super tweaky, but I wonder if it would be more intuitive to write `virtualNode.actualNode` or `vNode.actualNode` so it communicates what that top level object is. That would help in other places too (but changing it everywhere is obvs outside the scope of this PR)

lib/core/public/cleanup-plugins.js
WilcoFiers: Turns out there was a lint error in this line (missing ;). Very strange this didn't get picked up in circle.

PR #498, opened 2017-08-23
feat(landmark-one-main): add rule ensuring one main landmark in document

ssanaul:
This is the rule that incorporates the checks from the previous PR's as requested.


jongund:
This pull request combined the at least one main landmark rule and more than one main landmark rule and is ready for your review.

lib/checks/keyboard/has-at-least-one-main.json
```diff
@@ -0,0 +1,12 @@
+{
+	"id": "has-at-least-one-main",
+	"evaluate": "has-at-least-one-main.js",
+	"after": "has-at-least-one-main-after.js",
```
WilcoFiers: This file is missing, which is why the build isn't working.

test/integration/full/landmark-at-least-one-main/landmark-at-least-one-main-fail.html
```diff
@@ -0,0 +1,24 @@
+<!doctype html>
```
WilcoFiers: We need integration tests for when there are no main landmarks too.

lib/checks/keyboard/has-at-least-one-main.js
```diff
@@ -0,0 +1,3 @@
+var main = document.querySelector('main,[role=main]');
```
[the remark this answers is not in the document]
ssanaul: What purpose does adding 'virtualnode' serve? I am getting errors concerning an undefined object.

lib/checks/keyboard/has-at-least-one-main.js
```diff
@@ -0,0 +1,3 @@
...
+this.data(!!main);
```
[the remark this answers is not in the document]
ssanaul: This tells the after function whether or not the main exists; if there's a main in any one document, all of the documents will pass. This will probably make more sense when I add the file for the after function.

lib/checks/keyboard/has-at-least-one-main.js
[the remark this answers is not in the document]
jongund: The change to use "const main = axe.utils.querySelectorAll(virtualNode, 'main,[role=main')" seems to break all our test cases and we get the following error:

"TypeError: Cannot read property 'length' of undefined"

Is there something we need to change in the test cases?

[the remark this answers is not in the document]
WilcoFiers: @jongund When you call the check in a test, and that test uses virtualNode, you need to pass a virtualNode object in. We've got a test utility to help set that up on `axe.testUtils.checkSetup`. You can see an example here:

https://github.com/dequelabs/axe-core/blob/develop/test/checks/aria/required-children.js#L22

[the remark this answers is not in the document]
marcysutton: The test utilities are also documented here: https://github.com/dequelabs/axe-core/blob/master/doc/developer-guide.md#test-utilities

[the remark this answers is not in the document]
ssanaul: It doesn't appear to be possible to test the rule on an 'html' node using checkSetup. Do you have any suggestions on how to go about doing this?

jongund:
Any additional questions on this pull request?
Do you need anymore changes?

WilcoFiers:
@ssanaul @jongund Several of my comments haven't been addressed yet, including the one that's causing the tests not to run. There isn't much I can review if the tests aren't passing, let alone not running at all. Please resolve these issues.

jongund:
@ssanaul @WilcoFiers 

Wilco we have a question about how the "target" attribute is used in a unit test, since we are looking at page level rules.

Also we just wanted to verify that all of the files we need to run test are included.

lib/checks/keyboard/has-at-least-one-main.js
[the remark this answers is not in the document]
marcysutton: Forgive me for trying to catch up on this PR...but why do you need to test the HTML node for a `main` rule? Tests are scoped to the `#fixture` element, which is already present in the test environment, to avoid auditing the test harness UI itself.

marcysutton:
Upon preparing remarks for https://github.com/whatwg/html/issues/100, it occurred to me that we might need to account for inert subtrees. What about this scenario:

A view of a web application opens in a new modal layer, with the background rendered `inert` either using the `inert` attribute (and polyfill) or using `aria-hidden="true"` and `tabindex="-1"` on every focusable control. Deque best practices state that h1 elements can start over in this new layer, since the heading hierarchy is isolated to this modal layer. Shouldn't that also be the case with `<main>`? That would allow an AT user to easily jump to the "new" main content. As long as the two elements had different labels and visibility was managed so only one was visible at a time, I'd think that would still be navigable.

If that scenario was acceptable, we'd want to check to make sure a `<main>` element in the background layer wasn't counted against the total number of elements.

Just wanted to throw that out there for your input before I respond more fully to WHATWG.

lib/checks/keyboard/has-at-least-one-main.js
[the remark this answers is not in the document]
ssanaul: The selector of the rule itself is 'html' so that the rule runs even if there are no main landmarks. My thought was that it wouldn't really be a proper test if it didn't begin from the specified node.

WilcoFiers:
@ssanaul @jongund What's the status of this PR? Tests still aren't passing so I'm assuming you're still working on it?

ssanaul:
Even though the unit tests are passing, the integration tests for the rule are only returning results for has-at-least-one-main, which itself is only checking the main DOM, not any iframes.
Any thoughts on why this is the case?

test/integration/full/landmark-one-main/landmark-one-main-fail.js
```diff
@@ -0,0 +1,34 @@
...
+		axe.run({ runOnly: { type: 'rule', values: ['landmark-one-main'] } }, function (err, r) {
+			assert.isNull(err);
+			results = r;
+			done();
+		});
+	});
+
+	describe('violations', function () {
+		it('should find 1', function () {
+			assert.lengthOf(results.violations, 1);
```
WilcoFiers: Check the number of nodes. I think that should be 2. You're not testing the second node here either.

test/testutils.js
```diff
@@ -94,4 +94,54 @@ testUtils.checkSetup = function (content, options, target) {
...
+/**
+ * Create check arguments with Shadow DOM
+ *
+ * @param Node|String 	Stuff to go into the fixture (html or node)
+ * @param Node|String 	Stuff to go into the shadow boundary (html or node)
+ * @param Object  			Options argument for the check (optional, default: {})
+ * @param String  			Target for the check, CSS selector (default: '#target')
+ * @return Array
+ */
+testUtils.shadowCheckSetup = function (content, shadowContent, options, target) {
```
WilcoFiers: Why is this here?

lib/checks/keyboard/has-at-least-one-main-after.js
```diff
@@ -0,0 +1,8 @@
+var hasMain = false;
+for (var i = 0; i < results.length && !hasMain; i++) {
+	hasMain = results[i].data;
```
WilcoFiers: This doesn't look like it would work. Basically, you're taking the last item on `results` and assigning that as the result for everything. That doesn't seem right. What was this supposed to do? Please comment the code, as these double loops are confusing.

test/testutils.js
marcysutton: I created it for them in a separate branch, there was a lot of boilerplate for shadow DOM tests that I wanted to smooth out.

lib/checks/keyboard/has-at-least-one-main-after.js
ssanaul: @WilcoFiers @jongund 

```
var hasMain = false;

//iterate through results from each document
//stops if any document contains a main landmark
for (var i = 0; i < results.length && !hasMain; i++) {
	hasMain = results[i].data;
}

//if any document contains a main landmark, set all documents to pass the check
//otherwise, fail all documents
//since this is a page level rule, all documents either pass or fail the requirement
for (var i = 0; i < results.length; i++) {
	results[i].result = hasMain;
}

return results;
```

Do these comments clear up the behavior of the function?

Since this is a page level rule all documents either fail or pass this rule.

We only need one main in any of the documents (e.g. main document or an iframe) to pass this rule, 

If some documents fail this rule (e.g. iframe), people will think they need to have a main landmark in every iframe.

Does this make sense?

Are we getting the logic right here, do we need to update the individual document results?

test/integration/full/landmark-one-main/landmark-one-main-fail.js
ssanaul: @WilcoFiers @jongund 
It appears as if results.violations only has one node even though it should be observing two violations. We couldn't find any implementation of a test for more than one violation. 
However, the testing results.passes in landmark-one-main-pass.js does return 4 nodes.
Any thoughts on why violations only seems to contain one node, for our test as well as the other tests in the aXe library?

lib/checks/keyboard/has-at-least-one-main-after.js
```diff
@@ -0,0 +1,16 @@
+var hasMain = false;
+
+//iterate through results from each document
+//stops if any document contains a main landmark
+for (var i = 0; i < results.length && !hasMain; i++) {
```
WilcoFiers: Tip for future work. For-loops generally make code harder to read. JS has a bunch of very excellent array methods which let you write more expressive code. If you had written your code like this, you wouldn't have needed the comment.
```js
const hasMain = results.some(result => result.data === true);
```

test/integration/full/landmark-one-main/landmark-one-main-fail.js
WilcoFiers: @ssanaul Looks like another timing issue. You're calling axe.run before the frames are loaded. Have a look at how that was solved in landmark-main-is-top-level in the before.

ssanaul: @WilcoFiers @jongund 
Both nodes are tested now.

WilcoFiers: Excellent. I think that will do it!! Thanks @ssanaul!

lib/checks/keyboard/has-at-least-one-main-after.js
ssanaul: Great, I'll keep that in mind. Thanks for the tip!

PR #506, opened 2017-08-31
fix: Ignore shadowRoots on elements that don't allow them

### LINKED ISSUE #505: Ignore shadowRoots on elements that don't allow them
Came across an issue of axe 3.0 not being able to build its tree correctly. Turned out it was caused by Chrome, which has recently added a shadow root to every HTML element. This tripped `axe.utils.flattenTree` up massively.

The solution is to only build the shadow tree for elements allowed to have shadow trees:
https://dom.spec.whatwg.org/#dom-element-attachshadow

Only valid custom element names, "article", "aside", "blockquote", "body", "div", "footer", "h1", "h2", "h3", "h4", "h5", "h6", "header", "main" "nav", "p", "section", or "span",

WilcoFiers:
Closes #505


lib/core/utils/flattened-tree.js
```diff
@@ -51,6 +51,20 @@ function getSlotChildren(node) {
 	return retVal;
 }
 
+const possibleShadowRoots = ['article', 'aside', 'blockquote',
```
marcysutton: What about custom elements? Could we do a blacklist approach instead of a whitelist and exclude `html`?

WilcoFiers: Custom elements are allowed as long as they have a dash in the name. Chrome doesn't let you attach shadow trees to other stuff, so I think this works best.

dylanb:
@wilco what about v0?

marcysutton:
@dylanb the only API I'd wonder about is `.shadowRoot`, which I see referenced in this v0 gist from Eric Bidel: https://gist.github.com/ebidel/65b4565947c6399c82ac6f7d1a276442

I can't find a spec with a list of allowed elements, so with custom elements allowed I'd assume this is fine.

marcysutton:
I'm going to merge this, because we need to release an alpha.2 for aXe coconut.

lib/core/utils/flattened-tree.js
dylanb: @WilcoFiers could you add a test to make sure that that statement does not change?

WilcoFiers: You mean test that something in the codebase didn't modify this array somehow? I could freeze it if you want.

marcysutton: Maybe write a test for custom elements?

dylanb: @WilcoFiers No, what I was suggesting is that we write a test to make sure you cannot attach a shadow root to a custom element without a hyphen

PR #507, opened 2017-09-03
docs: Document how to propose axe-core rules

WilcoFiers:
This PR is designed to do a few things:

1. Streamline how rules get added. We shouldn't have people just creating PRs without discussing ahead of time the rules they are trying to add.
2. Create a standard that we can use to document the rules in. This will help with the transparency of axe-core.
3. Make it easier for us to take axe-core rules and propose them as WCAG ACT Rules.

Here are a few example rules the ACT Taskforce has developed for comparison: wilcofiers.github.io/act-rules/

### DIFF
--- a/doc/act-rules-format.md
+++ b/doc/act-rules-format.md
@@ -1,0 +1,35 @@
+# W3C Standardized Rules
+
+Deque Systems is one of leading organizations in the development of standardized accessibility conformance testing rules. The [axe-core rules proposal format](./rule-proposal.md) is an adaptation of the [Accessibility Conformance Testing Rules Format](https://www.w3.org/TR/act-rules-format/).
+
+There are two ways a rule written in the axe-core rule format can be transformed into the ACT Rules format:
+
+## Method 1: Create a single rule
+
+This method is useful for rules with a small number of checks. 
+
+1. Add the test input type to it: `rendered page`
+2. Add an `assumptions` section, add possible assumptions to it
+3. Add an `outcomes` section, describing the different possible outcomes of the rule
+4. Add a `Validation Tests` section, that links to the integration tests
+5. Update the check to return pass/fail/cantTell instead of true/false/undefined
+6. Add control flow to the checks:
+  - `any` checks should only return `fail` in the last step. All steps leading up to it either return `pass` or say `continue to the next step`.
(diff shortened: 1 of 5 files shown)


doc/rule-proposal.md
```diff
@@ -0,0 +1,84 @@
...
+
+**Example 2:**
+
+`###` keyboard/focusable-no-name (none)
+
+1. If the element is not focusable, return false
+2. If the element has an accessible name, return false
+3. Otherwise return true
+
+## Template
```
isner: The template is missing a section for specifying tags, which you ask the submitter to supply in the section above.

doc/act-rules-format.md
```diff
@@ -0,0 +1,35 @@
+# W3C Standardized Rules
+
+Deque Systems is one of leading organizations in the development of standardized accessibility conformance testing rules. The [axe-core rules proposal format](./rule-proposal.md) is an adaptation of the [Accessibility Conformance Testing Rules Format](https://www.w3.org/TR/act-rules-format/).
+
+There are two ways a rule written in the axe-core rule format can be transformed into the ACT Rules format:
+
+## Method 1: Create a single rule
```
dylanb: Best practices for rules should be added:

1) Rules should only have one `none` check so that the error message is specific
2) Rules should not combine `any` and `none`, these should be broken out into separate rules
3) Checks should each only test a single specific case (either a passing technique or a single failing test)

doc/rule-proposal.md
```diff
@@ -0,0 +1,84 @@
...
+
+Before you start coding a new rule for axe-core, you *must* create a Github issue to document the rule you want to create. There are many considerations to writing a good rule. It must have no false positives, which is difficult, because there are always many many edge cases to consider. Additionally, a known problem of accessibility testing, is that not all testers hold the same interpretation of the accessibility guidelines. All rules *must* be consistent with the interpretation of Deque Systems, the developer behind Axe-core. 
+
+In addition to giving the axe-core development team an opportunity to provide feedback on the proposed rule, the Github Issue will serve as documentation of that rule for the future.
+
+## Rules Format
+
+All Github issues that propose a rule must be tagged as *rule*, and must use the following format:
+
+### Intro
```
dylanb: Should we add a section for the message that would appear if the rule fails? This is what the user will see most of the time.

WilcoFiers:
Updated the PR based on @dylanb and @isner feedback.

PR #508, opened 2017-09-05
fix: RestoreScroll was running  out of sync

WilcoFiers:
Fixes an issue reported in JIRA WWD-1115.


marcysutton:
@WilcoFiers should this also go to develop-2x?

Issue #91, opened 2015-07-29
ARIA 1.1 States and Properties


dsturley:
We currently support all ARIA 1.1 roles; but we may have missed some of the new states and properties.

dsturley:
## New Attributes
aria-modal
http://www.w3.org/TR/wai-aria-1.1/#aria-modal

aria-placeholder
http://www.w3.org/TR/wai-aria-1.1/#aria-placeholder

aria-colcount
http://www.w3.org/TR/wai-aria-1.1/#aria-colcount

aria-colindex
http://www.w3.org/TR/wai-aria-1.1/#aria-colindex

aria-colspan
http://www.w3.org/TR/wai-aria-1.1/#aria-colspan

aria-rowcount
http://www.w3.org/TR/wai-aria-1.1/#aria-rowcount

aria-rowspan
http://www.w3.org/TR/wai-aria-1.1/#aria-rowspan

aria-describedat
http://www.w3.org/TR/wai-aria-1.1/#aria-describedat

aria-current
http://www.w3.org/TR/wai-aria-1.1/#aria-current

### Check the change log
http://www.w3.org/TR/wai-aria-1.1/#h-changelog

marcysutton:
The more current list was written out in https://github.com/dequelabs/axe-core/issues/350.

PR #509, opened 2017-09-05
Aria 1.1. tests

marcysutton:
Hopefully I got them all! I didn't see a way to test the owned/context changes for treegrid in https://github.com/dequelabs/axe-core/pull/479/files.

Also, we should really find a way to replace the hard-coded IDs in those integration test files. Couldn't we generate those with JS? Reordering them is such a huge pain.


nschonni:
Probably a dumb question, but what IDs these as 1.1 tests vs 1.2?

marcysutton:
@nschonni I assume you mean ARIA 1.1 vs. ARIA 1.2? We don't differentiate ARIA versions really. We could tag them potentially, but I don't know if that would be useful. We simply don't add anything to axe-core that isn't widely supported. See https://github.com/dequelabs/axe-core/issues/91 and https://github.com/dequelabs/axe-core/issues/350.

nschonni:
@marcysutton thanks for the explanation. I was thinking more back to the Quail days where rulesets were strongly typed

PR #513, opened 2017-09-07
fix(responsible): Restrict error construction to known errors

WilcoFiers:


marcysutton:
I tested it in an extension and it works fine! Thanks Wilco.

Issue #499, opened 2017-08-26
Add 'aria-errormessage' to definitions list


AutoSponge:
context: https://github.com/dequelabs/axe-core/blob/645d1fa52f1bee7e452a3d9b5927b32005fa5a93/lib/commons/aria/index.js#L51

spec: https://www.w3.org/TR/wai-aria-1.1/#aria-errormessage

Suggested fix: insert

```js
	'aria-errormessage': {
		type: 'idref'
	},
```

WilcoFiers:
I have some reservations about allowing this into axe-core just yet. Do you know if there are any assistive technologies that support this yet? We have a policy of only taking in accessibility feature that have sufficient support. The only exception to that being, those features that wouldn't get in the way if they weren't support. I don't think that's the case here, but certainly open to some counter-arguments.

AutoSponge:
@WilcoFiers, as far as I know, `aria-errormessage` is not supported by browser/AT yet. Currently, techniques that use `aria-describedby` to link the error message read out when the linked element is focused. The error message is also read out when using `role=alert` or `aria-live=assertive`.

I see three options here:

# Option 1
Replacing `aria-describedby` with `aria-errormessage` and continued use of `aria-live` (adoption now)

## PRO: 
- the error message will read out when the error is added
- the code will be spec compliant
- will have future functionality when `aria-errormessage` is widely supported
- sends message to AT community is ready to support the feature

## CON: 
- focus element with current error state will not read out

# Option 2
Use both `aria-describedby` and `aria-errormessage` and continued use of `aria-live`  (adoption now)

## PRO: 
- the error message will read out when the error is added
- will have future functionality when `aria-errormessage` is widely supported.
- focus element with current error state will read out
- sends message to AT community is ready to support the feature _BUT_ requires AT to plan for conflict (see below)

## CON: 
- when `aria-errormessage` becomes widely available, _IF_ AT chooses to read both `aria-describedby` and `aria-errormessage`, the error message will be read twice on focus. Otherwise, `aria-errormessage` must proceed `aria-describedby` when both exist.

# Option 3
No changes to current pattern (opposes adoption now)

## PRO:
- No need to change code/patterns

## CON:
- AT may never implement since no one codes for it or is ready to support it

WilcoFiers:
[...]

Issue #497, opened 2017-08-22
Ensure CircleCI is running on PRs coming from forks


WilcoFiers:
This PR never got tested. It seems to be because it's from a fork, but we need to figure it out.
https://github.com/dequelabs/axe-core/pull/492

marcysutton:
I found out a little more information: https://twitter.com/shawnbot/status/907302027692089345

> It's a security measure—If your CI configuration includes private or encrypted environment variables, they could be compromised by the fork.

Trying to determine if there's a safe workflow, given that we have no environment variables as far as I can tell.

marcysutton:
Found the setting, you have to enable it in CircleCI Settings.

WilcoFiers:
Nice!

WilcoFiers:
@marcysutton can you make the same changes to our other open source projects?

marcysutton:
I did. But I only had access to aXe-webdriverjs and aXe-matchers.

PR #517, opened 2017-09-09
fix(aria-errormessage): adds support for aria-errormessage

AutoSponge:
closes https://github.com/dequelabs/axe-core/issues/499
see https://www.w3.org/TR/wai-aria-1.1/#aria-errormessage


AutoSponge:
I noticed https://github.com/dequelabs/axe-core/issues/497#issuecomment-328610443; if you need me to retarget this PR to a feature branch (not develop), have a maintainer create the target branch and I will update. That PR can be accepted then create a new PR from feature TBD to develop.

test/checks/aria/valid-attr-value.js
```diff
@@ -60,6 +60,23 @@ describe('aria-valid-attr-value', function () {
 		assert.isNull(checkContext._data);
 	});
 
+	it.only('should return true if aria-errormessage id is alert or aria-describedby', function () {
```
marcysutton: Need to remove the `.only` here.

lib/commons/aria/attributes.js
```diff
@@ -37,6 +37,24 @@ aria.validateAttr = function (att) {
...
 
 /**
+ * Validate the value of an ARIA attribute with an idref
+ * @param  {HTMLElement} doc Root element
+ * @param  {HTMLElement} node The element to check
+ * @param  {String} attr The name of the attribute
+ * @param  {String} value The value of the attribute
+ * @return {Boolean}
+ */
+aria.validateIdrefType = function (doc, node, attr, value) {
```
WilcoFiers: The logic looks excellent to me. What I'm wondering is if there's a better way of communicating what's going on to users. The proposed implementation doesn't give any sort of feedback about why in one case `aria-errormessage` is passed and in another it is failed. It'd be great if we could expose that somehow. I'm not too sure how we'd go about doing that though... Perhaps we could have a separate check for `aria-errormessage`? It could be a `none` check, that raises an error if none of these three cases are met. What do you think?

test/checks/aria/valid-attr-value.js
AutoSponge: ah. development artifact.

dylanb:
@AutoSponge can you also remove the `wcag2.1` from the commit message when you make this change

lib/commons/aria/attributes.js
AutoSponge: So, if this fn throws and I include a test for `throws`...

```js
throw new Error('The value of aria-errormessage must be supported using an additional technique that announces the message (e.g., aria-live, aria-describedby, role=alert, etc.)')
```

Is that enough? Too much?

marcysutton:
I asked him to put a link to the 2.1 spec in the commit, I should have been more specific. The commit body has a lot more room for stuff like that. The commit message should be under 100 characters.

AutoSponge:
@marcysutton I was thinking of [aria1.1](https://www.w3.org/TR/wai-aria-1.1/#aria-errormessage) but typed wcag2.1. I'll fix it.

AutoSponge:
Circle CI isn't showing me the failure (test section has infinite spinner) and I don't have permissions to trigger a build.

dylanb:
@AutoSponge 

```
 1) aria-allowed-attr aria-allowed-attr passing tests passes should find ["#pass69"]:
     Element not found
  2) aria-allowed-attr aria-allowed-attr passing tests passes should find ["#pass70"]:
     Element not found
  3) aria-allowed-attr aria-allowed-attr passing tests passes should find ["#pass71"]:
     Element not found
```

lib/commons/aria/attributes.js
AutoSponge: The tests are passing. The only thing I haven't tried to do is:
> Perhaps we could have a separate check for aria-errormessage?

Given my lack of familiarity with the testing harness, I'm not sure how I would go about that. Is it possible to make that an additional feature?

WilcoFiers:
@AutoSponge Any updates on this?

AutoSponge:
@WilcoFiers Github hasn't helped with the order of comments. My last comment was ~3 weeks ago: 

> The tests are passing. The only thing I haven't tried to do is:
> > Perhaps we could have a separate check for aria-errormessage?

> Given my lack of familiarity with the testing harness, I'm not sure how I would go about that. Is it possible to make that an additional feature?

I was waiting on input from you guys but @dylanb said you were on vacation so I didn't bug you :)

marcysutton:
I do think we need a new `aria-errormessage` check to provide appropriate feedback to the user. 

I started to fix this but I got hung up on the integration tests. The two checks are interfering with each other, and I'm not quite sure why. @AutoSponge you can see what I did here: https://github.com/dequelabs/axe-core/commit/4ae3fc38150dce829bc505383e67ad22e1eaa7d7?diff=unified

I have to move on to other stuff, so if someone else can pick this up that would help a lot!

AutoSponge:
@marcysutton I commented on your commit but it doesn't show here. So, just commenting to make sure you see: https://github.com/dequelabs/axe-core/commit/4ae3fc38150dce829bc505383e67ad22e1eaa7d7?diff=unified. When I made those changes it worked locally but I can't push to your branch.

AutoSponge:
I realized I could merge your changes here :)

marcysutton:
Thanks for the contribution, @AutoSponge!

PR #518, opened 2017-09-10
fix: check for role changes in lists and list items

### LINKED ISSUE #463: well formed lists and ARIA
Question from a WorldSpace customer: why does aXe not find this?

```
<ul role=”menubar”>
<li>stuff</li>
</ul>
```

I would think this would violate either 1.3.1 Lists or 4.1.1 Well Formed, since the list item has been orphaned due to the changed nature of the `<ul>` element.

AlmeroSteyn:
Change function of `dlitem`, `listitem`, `only-dlitem`s and `only-listitems` checks to take into account role changes to either parent elements or child list item elements. Previously changing the role of either elements did not trigger the correct symantic markup errors.

Closes  #463 

The following additions have been made to the current checks:

1. Changing the role of the parent `<ul>` or `<ol>` will have the child `<li>` tags report that they are not part of a list.
2. Changing the role of the parent `<dl>` will have the child `<dd>` or `<dt>` tags report that they are not part of a list.
3. Changing the role of a `<li>` tag will mean that it will get ignored by the check if the list contains any other valid content.
4. Filling a list with *ONLY* `<li>` tags with a different role will trigger an invalid content error.

### DIFF
--- a/lib/checks/lists/dlitem.js
+++ b/lib/checks/lists/dlitem.js
@@ -1,2 +1,3 @@
 var parent = axe.commons.dom.getComposedParent(node);
-return parent.nodeName.toUpperCase() === 'DL';
+return parent.nodeName.toUpperCase() === 'DL' && !parent.getAttribute('role');
+
--- a/lib/checks/lists/listitem.js
+++ b/lib/checks/lists/listitem.js
@@ -1,4 +1,11 @@
 var parent = axe.commons.dom.getComposedParent(node);
-return (['UL', 'OL'].includes(parent.nodeName.toUpperCase()) ||
-    (parent.getAttribute('role') || '').toLowerCase() === 'list');
-  
+
+var parentRole = (parent.getAttribute('role') || '').toLowerCase();
+
+var isListRole = parentRole === 'list';
+
+return (
+	(['UL', 'OL'].includes(parent.nodeName.toUpperCase()) &&
+		(!parentRole || isListRole)) ||
+	isListRole
+);
--- a/lib/checks/lists/only-listitems.js
+++ b/lib/checks/lists/only-listitems.js
@@ -1,18 +1,41 @@
 var bad = [],
 	permitted = ['STYLE', 'META', 'LINK', 'MAP', 'AREA', 'SCRIPT', 'DATALIST', 'TEMPLATE'],
-	hasNonEmptyTextNode = false;
+	hasNonEmptyTextNode = false,
+	hasListItem = false,
+	liItemsWithRole = 0,
+	isEmpty = true;
 
 virtualNode.children.forEach(({ actualNode }) => {
 	var nodeName = actualNode.nodeName.toUpperCase();
-	if (actualNode.nodeType === 1 && nodeName !== 'LI' && permitted.indexOf(nodeName) === -1) {
-		bad.push(actualNode);
+	var isListItemRole = false;
+
+	if (actualNode.nodeType === 1 && permitted.indexOf(nodeName) === -1) {
+		var role = (actualNode.getAttribute('role') || '').toLowerCase();
+		isListItemRole = role === 'listitem' || (nodeName === 'LI' && !role);
+		hasListItem = hasListItem || (nodeName === 'LI' && isListItemRole) || isListItemRole;
+
+		if (isListItemRole) {
+			isEmpty = false;
+		}
+
+		if (nodeName === 'LI' && !isListItemRole) {
+			liItemsWithRole++;
+		}
+
+		if (nodeName !== 'LI' && !isListItemRole) {
+			bad.push(actualNode);
+		}
+
 	} else if (actualNode.nodeType === 3 && actualNode.nodeValue.trim() !== '') {
 		hasNonEmptyTextNode = true;
 	}
 });
 
+var allLiItemsHaveRole = liItemsWithRole > 0 && virtualNode.children.filter(({ actualNode }) =>
+	actualNode.nodeName === 'LI').length === liItemsWithRole;
+
 if (bad.length) {
 	this.relatedNodes(bad);
 }
 
-return !!bad.length || hasNonEmptyTextNode;
+return !(hasListItem || (isEmpty && !allLiItemsHaveRole)) || !!bad.length || hasNonEmptyTextNode;
--- a/test/checks/lists/dlitem.js
+++ b/test/checks/lists/dlitem.js
@@ -21,6 +21,12 @@ describe('dlitem', function () {
 		assert.isFalse(checks.dlitem.evaluate.apply(null, checkArgs));
 	});
 
+	it('should fail if the dlitem has a parent <dl> with a changed role', function(){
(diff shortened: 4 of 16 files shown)


lib/checks/lists/only-listitems.js
```diff
@@ -1,18 +1,41 @@
...
 virtualNode.children.forEach(({ actualNode }) => {
 	var nodeName = actualNode.nodeName.toUpperCase();
-	if (actualNode.nodeType === 1 && nodeName !== 'LI' && permitted.indexOf(nodeName) === -1) {
-		bad.push(actualNode);
+	var isListItemRole = false;
+
+	if (actualNode.nodeType === 1 && permitted.indexOf(nodeName) === -1) {
+		var role = (actualNode.getAttribute('role') || '').toLowerCase();
+		isListItemRole = role === 'listitem' || (nodeName === 'LI' && !role);
+		hasListItem = hasListItem || (nodeName === 'LI' && isListItemRole) || isListItemRole;
```
WilcoFiers: This needs either a comment, or it needs to be cleared up a bit, because it's not immediately obvious what the if statements in this block do.

test/checks/lists/dlitem.js
```diff
@@ -21,6 +21,12 @@ describe('dlitem', function () {
 		assert.isFalse(checks.dlitem.evaluate.apply(null, checkArgs));
 	});
 
+	it('should fail if the dlitem has a parent <dl> with a changed role', function(){
```
WilcoFiers: Should be "should fail if the **dt element** has a parent <dl> with a changed role"

test/checks/lists/only-dlitems.js
```diff
@@ -80,30 +92,60 @@ describe('only-dlitems', function () {
...
+		assert.isTrue(checks['only-dlitems'].evaluate.apply(checkContext, checkArgs));
+	});
+
 	it('should return false if <meta> is used along side dt', function () {
 		var checkArgs = checkSetup('<dl id="target"><meta name="description" content=""><dt>A list</dt></dl>');
 
 		assert.isFalse(checks['only-dlitems'].evaluate.apply(checkContext, checkArgs));
 	});
 
+	it('should return true if <meta> is used along side dt with its role changed', function () {
```
WilcoFiers: This test seems unnecessary, and so do the new tests below this one.

test/checks/lists/only-listitems.js
```diff
@@ -60,43 +66,97 @@ describe('only-listitems', function () {
...
 	});
 
+	it('should return true if the list has only li items with their roles changed', function () {
+		var checkArgs = checkSetup('<ol id="target"><li role="menuitem">Not a list item</li><li role="menuitem">Not a list item</li></ol>');
+
+		assert.isTrue(checks['only-listitems'].evaluate.apply(checkContext, checkArgs));
+	});
+
+	it('should return false if the list has at least one li while others have their roles changed', function () {
+		var checkArgs = checkSetup('<ol id="target"><li >A list item</li><li role="menuitem">Not a list item</li></ol>');
```
WilcoFiers: @dylanb I remember you suggesting this should be a violation? My initial thought about these was that there might be roles we should allow. separator for instance would be one I think is very reasonable. Menuitem itself would fail here because it isn't owned by a menu, so that's its own error.

test/checks/lists/only-listitems.js
```diff
@@ -60,43 +66,97 @@ describe('only-listitems', function () {
...
+	});
+
 	it('should return false if <meta> is used along side li', function () {
 		var checkArgs = checkSetup('<ol id="target"><meta name="description" content=""><li>A list</li></ol>');
 
 		assert.isFalse(checks['only-listitems'].evaluate.apply(checkContext, checkArgs));
 	});
 
+	it('should return true if <meta> is used along side only li items with their roles changed', function () {
+		var checkArgs = checkSetup('<ol id="target"><meta name="description" content=""><li role="menuitem">Not a list item</li></ol>');
```
WilcoFiers: This one doesn't seem useful, and neither do tests added below this one.

test/checks/lists/only-listitems.js
```diff
@@ -60,43 +66,97 @@ describe('only-listitems', function () {
 		assert.isFalse(checks['only-listitems'].evaluate.apply(checkContext, checkArgs));
 	});
 
+	it('should return false if the list has only an element with role listitem', function () {
+		var checkArgs = checkSetup('<ol id="target"><div role="listitem">A list</div></ol>');
```
WilcoFiers: We should add tests for multiple listitem elms, and for mixing li and listitem.

marcysutton:
Hi @AlmeroSteyn, just checking in to see if you plan to address Wilco's feedback on this PR? We'd love to get this updated and merged.

WilcoFiers:
@marcysutton I heard from Almero that he wasn't. We need to sort this one out ourselves.

lib/checks/lists/dlitem.js
```diff
@@ -1,2 +1,3 @@
 var parent = axe.commons.dom.getComposedParent(node);
-return parent.nodeName.toUpperCase() === 'DL';
+return parent.nodeName.toUpperCase() === 'DL' && !parent.getAttribute('role');
```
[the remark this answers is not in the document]
marcysutton: Which would say it's okay to overload a `<dl>` with an ARIA role. Is that something we want to encourage? Wouldn't that also depend on the child items having `role="listitem"`?

lib/checks/lists/listitem.js
```diff
@@ -1,4 +1,11 @@
...
-return (['UL', 'OL'].includes(parent.nodeName.toUpperCase()) ||
-    (parent.getAttribute('role') || '').toLowerCase() === 'list');
-  
\ No newline at end of file
+
+var parentRole = (parent.getAttribute('role') || '').toLowerCase();
+
+var isListRole = parentRole === 'list';
+
+return (
```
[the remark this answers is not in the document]
marcysutton: @WilcoFiers which spacing convention are you referring to?

lib/checks/lists/only-listitems.js
marcysutton: It also seems a little redundant: 
```hasListItem = hasListItem || (nodeName === 'LI' && isListItemRole) || isListItemRole;```
The code checks the nodeName for 'LI' and a role more than once, and then checks against itself again a few times. I wonder if we can simplify.

lib/checks/lists/dlitem.js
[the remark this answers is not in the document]
jkodu: Done

lib/checks/lists/listitem.js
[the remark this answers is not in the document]
jkodu: Done

jkodu:
@AlmeroSteyn - cheers for the work.
I have circled up all comments on here and re-worked it in to this PR - https://github.com/dequelabs/axe-core/pull/949

Hence closing it.

test/checks/lists/only-listitems.js
dylanb: this example does not represent a violation IMO

Issue #516, opened 2017-09-09
postinstall task fails when installing on windows


molant:
**How to repro**
Run `npm install axe-core` on a Windows machine in regular cmd (not Bash on Windows).

This is the output:
```
C:\Projects\tests>npm install axe-core

> axe-core@2.4.0 postinstall C:\Projects\tests\node_modules\axe-core
> ./bin/postinstall.sh

'.' is not recognized as an internal or external command,
operable program or batch file.
npm WARN enoent ENOENT: no such file or directory, open 'C:\Projects\tests\package.json'
npm WARN tests No description
npm WARN tests No repository field.
npm WARN tests No README data
npm WARN tests No license field.

npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! axe-core@2.4.0 postinstall: `./bin/postinstall.sh`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the axe-core@2.4.0 postinstall script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\antonmo\AppData\Roaming\npm-cache\_logs\2017-09-09T03_30_18_327Z-debug.log
```

This is the npm debug log:
[2017-09-09T03_30_18_327Z-debug.txt](https://github.com/dequelabs/axe-core/files/1289516/2017-09-09T03_30_18_327Z-debug.txt)

We are looking into updating to the latest version in https://github.com/sonarwhal/sonar, but the tool needs to run in any OS so we can't at the moment :(

marcysutton:
Dang, that's a bummer. It's true, the postinstall script would probably only work on Git Bash. Either we need to exclude that npm script and file if it isn't supported, or we need a way to differentiate between end-user installs and development installs. I'll have to look into this some more. What I was trying to avoid was having contributors miss the pre-commit hook in an opt-in npm script instead of running it automatically.

marcysutton:
@molant in the future can you keep an eye out for our alpha releases? We had two alpha versions of 2.4.0 before this was released–would have been a good opportunity to find out about the Windows problem. I've been posting about them on Twitter and tagging on https://github.com/dequelabs/axe-core/releases.

molant:
> @molant in the future can you keep an eye out for our alpha releases? 
Will do. We were relying on greenkeeper to get notifications of new versions but it hasn't notify us of anything for the last few weeks :/

[...]

PR #520, opened 2017-09-12
fix(postinstall): use node, more conditionals

marcysutton:
Closes https://github.com/dequelabs/axe-core/issues/516

### DIFF
--- a/bin/postinstall.sh
+++ b/bin/postinstall.sh
@@ -1,8 +1,0 @@
-#!/bin/sh
-
-if [ -d ".git/hooks" ]; then
-	if [ ! -f ".git/hooks/commit-msg" ]; then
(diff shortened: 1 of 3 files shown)


marcysutton:
One caveat is I had to install `fs-extra` as a dependency to be able to copy the file, as `fs.copyFile` is only available in Node 8.5.0. Streams weren't cutting it for a pre commit-hook. I could potentially change fs-extra to a devDependency and use both fs and fs-extra in the script, if we don't want the dependency in axe-core.

marcysutton:
Woo! Thanks @isner

ahmadnassri:
this change seems to have broken node4 installs: https://github.com/dequelabs/axe-core/issues/541

PR #521, opened 2017-09-13
fix(aria): adding support for aria-expanded in menuitem

devaradhanm:


marcysutton:
Thanks for this PR! I'm including the discussion that occurred outside of Github for additional context:

> Looks like it is coming in the ARIA 1.2 spec: https://github.com/w3c/aria/issues/454
 
> "Given there is consensus this should be changed in the spec for ARIA 1.2 and that it is already implemented in browsers on all major platforms, per discussion in last week's APG meeting, we plan to leave it in the APG examples with an appropriate note that will link here."

senorfrench61:
Hi, I see menuitem is supported now. How about menu? I am setting aria-expanded to true and false on a link that has aria-popup="true" in order to view submenu items but to also see 'expanded' and 'collapsed'. It works fine in IE and FF, but not my Chrome Version 78.0.3904.97 (Official Build) (64-bit).

PR #539, opened 2017-09-20
docs: Clarify how to use sri-history in sri-update.js

WilcoFiers:
Hope this helps :) We'll add it to our auto-build tool.

### DIFF
--- a/build/sri-update.js
+++ b/build/sri-update.js
@@ -1,3 +1,17 @@
+/**
+ * SRI Update will update the sri-history.json file
+ * In the project root. The sri-history.json file contains
+ * SRI hashes of each released version. This can be used
+ * to validate that something is a known axe-core source
+ * file.
+ *
+ * When running npm version, this script will execute and
+ * update sri-history.json with the SRIs of axe{.*}.js.
+ * In npm publish, this script will be executed again
+ * with --validate, which will make sure the SRIs still
+ * match. If not, this script should be run again,
+ * and its changes committed.
+ */
 var path = require('path');
 var fs = require('fs');
 var sriToolbox = require("sri-toolbox");
--- a/build/sri-update.js
+++ b/build/sri-update.js
@@ -5,12 +5,8 @@
  * to validate that something is a known axe-core source
  * file.
  *
- * When running npm version, this script will execute and
+ * When running `npm run release`, this script will execute and
  * update sri-history.json with the SRIs of axe{.*}.js.
- * In npm publish, this script will be executed again
- * with --validate, which will make sure the SRIs still
- * match. If not, this script should be run again,
- * and its changes committed.
  */
 var path = require('path');
 var fs = require('fs');
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "version": "echo \"use 'npm run release' to bump axe-core version\" && exit 1",
     "prepublishOnly": "grunt build",
     "postinstall": "node build/utils/postinstall.js",
-    "release": "standard-version && node build/sri-update"
+    "release": "standard-version"
   },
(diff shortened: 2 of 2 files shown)


marcysutton:
We need to add this to the `release` script in package.json so it will execute when I run our automated release process. Then it will be executed with `npm run release`.

```
scripts: {
    "release": "node build/sri-update && standard-version"
}

**Comment by WilcoFiers** [comment]
@marcysutton Good point, I created a fix.

File: package.json [hunk 1]
```diff
@@ -50,10 +50,10 @@
     "build": "grunt",
     "test": "grunt test",
     "test-fast": "grunt test-fast",
-    "version": "node build/sri-update",
+    "version": "echo \"use 'npm run release' to bump axe-core version\" && exit 1",
```
nschonni: Why not alias this to `release` or just remove it entirely?

WilcoFiers: Because both `npm version` and `npm run release` bump the version, so if I alliase it its bumped twice, and if I dont put in the error we might accidentally use the wrong command and skip generating the changelog.

marcysutton: I like this approach, it makes it clear that `npm version` isn't the way to go.

build/sri-update.js
```diff
@@ -1,3 +1,17 @@
+/**
+ * SRI Update will update the sri-history.json file
+ * In the project root. The sri-history.json file contains
+ * SRI hashes of each released version. This can be used
+ * to validate that something is a known axe-core source
+ * file.
+ *
+ * When running npm version, this script will execute and
```
marcysutton: We need to update this comment with `npm run version`

marcysutton:
So this approach isn't going to work–we build axe-core on prepublishOnly so that step isn't accidentally forgotten before running `npm publish` (happened to me!). That means the build of axe-core in that step won't match the hash, which has to be built before being validated to be useful. I don't see how the hash/validate workflow could ever work.

I'd like to propose that we skip validating the hash in prepublishOnly, since it's blocking publishing at the moment. Instead, we can generate the hash as part of `npm run release` and skip validating it inside of axe-core. That makes the most sense to me since I thought the intent of the sri-history hashing was for consumers of axe-core, like WorldSpace Comply.

marcysutton:
I made some updates to reflect what I mentioned in my last comment. However, this begs the question–will the hash committed with the changes from `npm run release` still work if the axe-core version built in `prepublishOnly` and deployed to npm still match when we go to consume it downstream?

WilcoFiers:
Why wouldn't we want to run the validation?

marcysutton:
@WilcoFiers each run of `grunt build` produces a different hash, in my experience. So when `grunt build` happens on prepublishOnly along with `--validate`, it produces a different hash and no longer matches the one already committed.

package.json
```diff
@@ -50,10 +50,10 @@
     "build": "grunt",
     "test": "grunt test",
     "test-fast": "grunt test-fast",
-    "version": "node build/sri-update",
-    "prepublishOnly": "grunt build && node build/sri-update --validate",
+    "version": "echo \"use 'npm run release' to bump axe-core version\" && exit 1",
+    "prepublishOnly": "grunt build",
     "postinstall": "node build/utils/postinstall.js",
-    "release": "standard-version"
+    "release": "standard-version && node build/sri-update"
```
WilcoFiers: @marcysutton Are you alright with removing this, now that Attest-master is set up to publish axe-core?

marcysutton:
Oh yes, I'd love to get a demo from you on using that tool!

package.json
marcysutton: Oh yes, I'd love to get a demo from you on using that tool!

Issue #541, opened 2017-09-20
postinstall script fails on node4


ahmadnassri:
```
Exit code: 1
Command: node build/utils/postinstall.js
Arguments: 
Directory: /home/ahmad/Projects/ahmadnassri/nightwatch-accessibility/node_modules/axe-core
Output:
/home/ahmad/Projects/ahmadnassri/nightwatch-accessibility/node_modules/axe-core/build/utils/postinstall.js:4
	let commitMsg = '.git/hooks/commit-msg'
	^^^

SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:373:25)
    at Object.Module._extensions..js (module.js:416:10)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
```

brokenmass:
Can you release this as a 2.x.x version asap please ?

brokenmass:
Bump

marcysutton:
this was released as part of axe-core@2.4.2

nemanjacosovic:
@marcysutton I'm having this failed on:
```
node v4.4.7
npm  v3.10.6

```
But its working on
```
node v6.10.0
npm v5.6.0
```

marcysutton:
axe-core requires node 6.

PR #542, opened 2017-09-20
fix node4 build issue

ahmadnassri:
fixes: #541

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,9 @@
   "description": "Accessibility engine for automated Web UI testing",
   "version": "3.0.0-alpha.4",
   "license": "MPL-2.0",
+  "engines": {
+    "node": ">=4"
+  },
   "contributors": [
     {
       "name": "David Sturley",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -3,6 +3,9 @@
   "description": "Accessibility engine for automated Web UI testing",
   "version": "3.0.0-alpha.4",
   "license": "MPL-2.0",
+  "engines": {
+    "node": ">=4"
```
marcysutton: Do we need to specify a node version if we're downgrading it to use `var`?

renatoi:
Maybe also add supported node version to test to catch these issues in the future:
https://github.com/dequelabs/axe-core/blob/develop/circle.yml

marcysutton:
@renatoi our tests running on Circle have different requirements than axe-core for end-users. We transpile lib JS using Babel for backwards-compatibility but this isn't required for our tests. The postinstall script wasn't caught by Babel unfortunately.

package.json
ahmadnassri: it just produces a warning for anybody running on older versions, so it's useful

PR #543, opened 2017-09-20
chore(sri-history): add sri entry for 2.4.1

isner:


marcysutton:
looks good, thanks Matt!

marcysutton:
We should either change the base to `develop-2x` so this gets merged into the 2x branch first, or be sure to cherry-pick this onto develop-2x (we'll have to cherry-pick either way).

Thanks Matt!

Issue #549, opened 2017-09-26
Allow using role 'combobox' without child role 'textbox' on a native input element with type="search"


iamrafan:
Issue is related to #160 . It was fixed for only type ="text", need to add support for type="search" too.

marcysutton:
That seems reasonable to me, as an input with `type="search"` will have the same role of `textbox`. The only difference is the keyboard that is brought up on mobile devices.

isner:
Fixed by https://github.com/dequelabs/axe-core/pull/559.

iamrafan:
@isner thanks for the fix.

Issue #548, opened 2017-09-26
False positive for aria-required-children rule on role 'combobox'


iamrafan:
Aria 1.1 standard states child role ‘listbox’ is only needed when aria-expanded is set to true

Looks like Axe is following the Aria1.0 standard for role combobox . It is expecting both ‘listbox’ and ‘textbox’ as child roles, irrespective of the state of aria-expanded’.

![image](https://user-images.githubusercontent.com/22068434/30885509-6c01ea4e-a2c8-11e7-8660-4c5195248265.png)

marcysutton:
We have to do some testing to find out the accessibility supported status of this, but we'll report back as soon as we have more information. Test cases welcome

marcysutton:
@iamrafan if you have a working demo we can check out, that will make this go much faster.

iamrafan:
@marcysutton : Please find the sample tests from below
```
case-1: Valid Case - aria-expanded is true and textbox, listbox are direct children of combobox
<div aria-label="Tag" role="combobox" aria-expanded="true" aria-haspopup="listbox">
    <input type="search" role="textbox" aria-label="input-1">
    <ul role="listbox">
        <li role="option">Zebra</li>
        <li role="option" id="selected_option">Zoom</li>
    </ul>
</div>

<hr> case-2: Valid case - aria-expanded is true, textbox is direct child of combobox and element with role listbox is owned by
combobox
<div aria-expanded="true" role="combobox" aria-owns="Basicdrop2-list">
    <span id="Basicdrop2-option" role="textbox" aria-readonly="true">
        <span>Select an Option</span>
    </span>
</div>

<ul role="listbox" id="Basicdrop2-list">
    <li role="option">Zebra</li>
    <li role="option" id="selected_option2">Zoom</li>
</ul>

<hr> case-3: False positive: aria-expanded is false and child textbox is present, no need of role listbox
<div aria-expanded="false" role="combobox">
    <input type="search" role="textbox" aria-label="input-3">
</div>

<br>
<hr> case-4: Valid: aria-expaned is true and role combobox is missing child role listbox
<div aria-expanded="true" role="combobox">
    <span id="Basicdrop3-option" role="textbox">
        <span>Select an Option</span>
    </span>
</div>
```

isner:
[...]

PR #559, opened 2017-10-04
fix(required-children): add combobox > listbox exception

isner:
No longer requires a collapsed combobox to own a child listbox.

The relevant ARIA 1.1 spec: https://www.w3.org/TR/wai-aria-1.1/#combobox

https://github.com/dequelabs/axe-core/issues/548

### DIFF
--- a/lib/checks/aria/required-children.js
+++ b/lib/checks/aria/required-children.js
@@ -30,8 +30,8 @@ function ariaOwns(nodes, role) {
 	return false;
 }
 
-function missingRequiredChildren(node, childRoles, all) {
-	//jshint maxstatements: 19
+function missingRequiredChildren(node, childRoles, all, role) {
+	//jshint maxstatements: 22, maxcomplexity: 13
 	var i,
 	l = childRoles.length,
 	missing = [],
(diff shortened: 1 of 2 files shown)


isner:
Please hold off on merging this for a short while. I want to add a commit that will close https://github.com/dequelabs/axe-core/issues/549.

lib/checks/aria/required-children.js
```diff
@@ -51,7 +51,7 @@ function missingRequiredChildren(node, childRoles, all, role) {
 
 		// remove 'textbox' from missing roles if combobox is a native input
 		var textboxIndex = missing.indexOf('textbox');
-		if (textboxIndex >= 0 && node.tagName === 'INPUT' && node.type === 'text') {
+		if (textboxIndex >= 0 && node.tagName === 'INPUT' && ['text', 'search'].indexOf(node.type) >= 0) {
```
WilcoFiers: I think there are probably other types that make sense too.

isner: Goo point. Which other types are you thinking of? To me, the only other types that seem reasonable to use as the basis of a combobox are "email" and "url". Beyond that, "tel" _might_ be worth supporting.

isner: I added "email", "url", and "tel" to the excepted input types.

PR #566, opened 2017-10-10
fix(getSelector): Ensure nodename is eescaped

### LINKED ISSUE #563: aXe fails on XHTML documents with prefixed element names
When ran against an XHTML document with namespaced elements, aXe crashes with a syntax error on the `matchesSelector` function.

This is due to the selector creation using the element’s `nodeName`, which results in an (invalid) selector of the form `prefix:name`.

Since the DOM Standard [doesn't plan to support namespaces with selectors](https://dom.spec.whatwg.org/#scope-match-a-selectors-string), the best approach would probably be to use the element’s local name as the selector (instead of the prefixed node name).

This can be reproduced with the following document:

```xml
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
<head>
  <meta charset="utf-8" />
  <title>Math Sample</title>
</head>
<body>
<h1>Math Sample</h1>
<m:math alttext="x" xmlns:m="http://www.w3.org/1998/Math/MathML">
  <m:semantics>
    <m:mi>x</m:mi>
    <m:annotation-xml encoding="MathML-Content">
      <m:ci>x</m:ci>
    </m:annotation-xml>
  </m:semantics>
</m:math>
</body>
</html>
```

WilcoFiers:
Closes #563

### DIFF
--- a/lib/core/utils/get-selector.js
+++ b/lib/core/utils/get-selector.js
@@ -79,7 +79,6 @@ const createSelector = {
   // Get uncommon node names
   getUncommonElm (elm, { isCommonElm, isCustomElm, nodeName })  {
     if (!isCommonElm && !isCustomElm) {
-      nodeName = escapeSelector(nodeName);
       // Add [type] if nodeName is an input element
       if (nodeName === 'input' && elm.hasAttribute('type')) {
         nodeName += '[type="' + elm.type + '"]';
@@ -130,7 +129,7 @@ const createSelector = {
  * recognize the element by (IDs, aria roles, custom element names, etc.)
  */
 function getElmFeatures (elm, featureCount) {
-  const nodeName = elm.nodeName.toLowerCase();
+  const nodeName = escapeSelector(elm.nodeName.toLowerCase());
   const classList = Array.from(elm.classList) || [];
   // Collect some props we need to build the selector
   const props = {
--- a/test/core/utils/get-selector.js
+++ b/test/core/utils/get-selector.js
@@ -194,6 +194,20 @@ describe('axe.utils.getSelector', function () {
 		assert.equal(result[0], node);
 	});
 
+	it('should work on complex namespaced elements', function () {
+		fixture.innerHTML = '<m:math xmlns:m="http://www.w3.org/1998/Math/MathML">' +
+		  '<m:mi>x</m:mi>' +
+		  '<m:annotation-xml encoding="MathML-Content">' +
+		    '<m:ci>x</m:ci>' +
+		  '</m:annotation-xml>' +
+		'</m:math>';
+		var node = fixture.querySelector('m\\:ci');
+		var sel = axe.utils.getSelector(node);
+		var result = document.querySelectorAll(sel);
+		assert.lengthOf(result, 1);
+		assert.equal(result[0], node);
+	});
+
 	it('shouldn\'t fail if the node\'s parentNode doesnt have children, somehow (Firefox bug)', function () {
 		var sel = axe.utils.getSelector({
 			nodeName: 'a',


WilcoFiers:
@rdeltour What's the the use case for that? When would the same system want to change the namespace, even though the element would be the same. I think it's a little clearer what element you're supposed to look at with the namespace specified, no?

rdeltour:
> What's the the use case for that?

The use case is to make selectors match the element in an XHTML context, such as in EPUB.

Let me try to clarify. This has to deal with whether the element has a namespace or not in the DOM, which depends on how the document was parsed (for instance, as XML or as HTML).

Consider this example HTML document:

```html
<!doctype html>
<title>Example</title>
<meta charset="utf-8" />
<h1>Hi</h1>
<math xlink:href="#foo"></math>
<div xlink:href="#foo"></div>
<foo:bar></foo:bar>
```
It's HTML, and parsed as such by a browser. On this document:
- `document.querySelectorAll('[xlink\\:href]')` will return the `div` element
- `document.querySelectorAll('[*|href]')` will return the `math` element
- `document.querySelectorAll('foo\\:bar')` will return the `foo:bar` element

It's all in the rules defined by the HTML parsing algorithm. For instance, when the parser encounters an `xlink:href` attribute on an SVG or MathML element, it [adjusts](https://html.spec.whatwg.org/multipage/parsing.html#adjust-foreign-attributes) it to put it in the XLink Namespace.

Now, consider this XHTML document:

```xml
<html xmlns="http://www.w3.org/1999/xhtml" xmlns:m="http://www.w3.org/1998/Math/MathML" xml:lang="en" lang="en">
<head>
  <meta charset="utf-8" />
  <title>Example</title>
</head>
<body>
<h1>Hi</h1>
<m:math></m:math>
</body>
</html>
```
If parsed as XHTML (for instance, put it in a file with the '.xhtml' extension and open it in a browser, or serve it with the `application/xhtml+xml` media type, then:

- `document.querySelectorAll('m\\:math`) is empty
- `document.querySelectorAll('math`) contains the `m:math` element
- `document.querySelectorAll('*|math`) contains the `m:math` element

Again, the reason lies in the parsing algorithm. The MathML namespace is known and defined, so the local name is just "math" and the namespace is "http://www.w3.org/1998/Math/MathML".

So, what should aXe do (ideally)?

The easy approach would be to create `*|name` selectors for prefixed element names. That would probably work in most of the cases, since prefixed foreign names are very rarely found in HTML, and it would work in an XHTML context.  
This is assuming that you're only creating type selectors (i.e. for elements). If you're also creating attribute selectors, you'd have to make cases depending on [foreign attribute adjustments](https://html.spec.whatwg.org/multipage/parsing.html#adjust-foreign-attributes).

A more correct approach would be to define the selector based on the value of `document.contentType`. If it is `text/html`, the selector would be `prefix\\:localname`; if it is `application/xhtml+xml`, the selector would be `*|localname`.

Voilà. I hope it makes sense :-)
I'm happy to help with the PR if needed (btw, is there a way to run a single test in axe without having to run everything with `grunt test`? I couldn't find it at first sight, and it's not in the docs either).

WilcoFiers:
Makes sense, and I learned something new today :D. Thanks Romain. @dylanb What do you think? 

Also, yes you can run specific tests by doing `--grep="test name"`.

dylanb:
I think we should implement option 2 because it will only change selectors for XHTML documents which is less disruptive to our users

WilcoFiers:
Agreed. Any idea how to detect what parser was used on a page? I want to avoid changing the selectors as much as possible. If we can make it so that localName is used in xml documents, and nodeName in all other cases, that'd be ideal. Do you want to take a stab at updating this PR @rdeltour ?

rdeltour:
> Agreed. Any idea how to detect what parser was used on a page?

I think we can look at `document.contentType`

> Do you want to take a stab at updating this PR @rdeltour ?

Sure, I can try in the coming days (end of next week at the latest).

marcysutton:
I saw a real-world case of invalid/namespaced HTML tags rendered from the server today, and they took down axe-core and our extensions. Here's an example:

```<search:facet-list></search:facet-list>```

I have a question out as to whether this was an XHTML or HTML5 doctype, but I'd be curious to see if namespaced HTML tags in a regular document like this cause any problems for assistive technologies. If they do cause problems, it might be a good thing for axe-core to flag.

rdeltour:
> I'd be curious to see if namespaced HTML tags in a regular document like this cause any problems for assistive technologies. If they do cause problems, it might be a good thing for axe-core to flag.

Good question. To be honest I have no idea, but I suspect most of them would just ignore it (as HTML parsers do).
If AT do have a problem to process them, yes it should certainly be flagged! (using a specific rule rather than the current syntax error of course).

WilcoFiers:
PR superseded by #582

PR #568, opened 2017-10-10
perf(reporter): add option to limit result types to be processed

### LINKED ISSUE #512: Axe reporter is very slow with large datasets
On extremely large pages, it is possible for the reporter to take several times longer than the actual audit to complete. The main problem appears to be in `helpers.processAggregate()`. Running it 30 000 times took several minutes to complete. Part of the issue here is that the extension is hugely inefficient here because it's processing passes, which won't actually get displayed. 

Part of the solution should be in preventing that. But there's certainly an extra problem. The method doesn't (or at least shouldn't) do anything that's particularly slow, so we need to figure out what's so slow and we should fix it.

This problem can be reproduced on the following page: 
https://dev.office.com/fabric#/components/detailslist

isner:
This allows us to specify a short list of result types that we want `helper.processAggregate` to bother processing. For some reporting scenarios, we don't want to bother processing 'passes', for instance. This causes a massive performance improvement for those scenarios.

We can leverage this option in the axe extension to speed up performance there as well.

Closes https://github.com/dequelabs/axe-core/issues/512

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -231,7 +231,7 @@ axe.run(context, options, callback);
 #### Parameters axe.run
 
 * [`context`](#context-parameter): (optional) Defines the scope of the analysis - the part of the DOM that you would like to analyze. This will typically be the `document` or a specific selector such as class name, ID, selector, etc.
-* [`options`](#options-parameter): (optional) Set of options passed into rules or checks, temporarily modifying them. This contrasts with `axe.configure`, which is more permanent. [See below for more information](#axerun-parameters)
+* [`options`](#options-parameter): (optional) Set of options passed into rules or checks, temporarily modifying them. This contrasts with `axe.configure`, which is more permanent.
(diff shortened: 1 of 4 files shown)


isner:
@WilcoFiers Tests added. Let me know if the documentation is insufficient.

doc/API.md
```diff
@@ -231,7 +231,7 @@ axe.run(context, options, callback);
 #### Parameters axe.run
 
 * [`context`](#context-parameter): (optional) Defines the scope of the analysis - the part of the DOM that you would like to analyze. This will typically be the `document` or a specific selector such as class name, ID, selector, etc.
-* [`options`](#options-parameter): (optional) Set of options passed into rules or checks, temporarily modifying them. This contrasts with `axe.configure`, which is more permanent. [See below for more information](#axerun-parameters)
```
isner: The documentation for `axe.configure` is _above_, not below, and the internal link is invalid. I figured we ought to simply remove this.

dylanb:
@WilcoFiers @isner did the CircleCI build run for this PR?

marcysutton:
@dylanb it says it did, you can find the integrations under "View details" on the merged commit above

PR #573, opened 2017-10-17
fix: Remove axe.a11yCheck()

WilcoFiers:
We’ve had axe.run() since 2.1, docs have noted a11yCheck() was going away, so as of 3.0 we shouldn’t have it around anymore.


dylanb:
IMO we should make sure that all of our projects that depend on this use the run function before removing this

WilcoFiers:
Why? We've already put breaking changes into 3.0. Quite a few of our products haven't been updated to deal with that yet. We'll have to do that in the coming month or two, but I don't see why we should keep this PR open for that long.

dylanb:
Seeing as this involves primarily our commercial products, lets take the discussion offline. I would like to leave the PR open for the moment to allow comment by other users as to the impact on their applications.

WilcoFiers:
@dylanb Can you approve the merge?

PR #582, opened 2017-10-24
Fix `getSelector` for namespaced elements + introduce XHTML fixtures

### LINKED ISSUE #563: aXe fails on XHTML documents with prefixed element names

rdeltour:
Replaces #566
Closes #563

### DIFF
--- a/lib/core/utils/get-selector.js
+++ b/lib/core/utils/get-selector.js
@@ -1,4 +1,5 @@
 const escapeSelector = axe.utils.escapeSelector;
+let isXHTML;
 
 function isUncommonClassName (className) {
   return ![
@@ -27,7 +28,7 @@ function getDistinctClassList (elm) {
 }
 
 const commonNodes = [
-  'div', 'span', 'p', 
+  'div', 'span', 'p',
   'b', 'i', 'u', 'strong', 'em',
   'h2', 'h3'
 ];
@@ -79,7 +80,6 @@ const createSelector = {
   // Get uncommon node names
   getUncommonElm (elm, { isCommonElm, isCustomElm, nodeName })  {
     if (!isCommonElm && !isCustomElm) {
-      nodeName = escapeSelector(nodeName);
       // Add [type] if nodeName is an input element
       if (nodeName === 'input' && elm.hasAttribute('type')) {
         nodeName += '[type="' + elm.type + '"]';
@@ -130,7 +130,12 @@ const createSelector = {
  * recognize the element by (IDs, aria roles, custom element names, etc.)
  */
 function getElmFeatures (elm, featureCount) {
-  const nodeName = elm.nodeName.toLowerCase();
+  if (typeof isXHTML === 'undefined') {
+    isXHTML = axe.utils.isXHTML(document);
+  }
+  const nodeName = escapeSelector(isXHTML?
+     elm.localName
+    :elm.nodeName.toLowerCase());
   const classList = Array.from(elm.classList) || [];
   // Collect some props we need to build the selector
   const props = {
@@ -174,7 +179,7 @@ function generateSelector (elm, options, doc) {
   let { isUnique = false } = options;
   const idSelector = createSelector.getElmId(elm);
   const {
-    featureCount = 2, 
+    featureCount = 2,
     minDepth = 1,
     toRoot = false,
     childSelectors = []
--- a/lib/core/utils/is-xhtml.js
+++ b/lib/core/utils/is-xhtml.js
@@ -1,0 +1,16 @@
+
+/**
+ * Determines if a document node is XHTML
+ * @method isXHTML
+ * @memberof axe.utils
+ * @instance
+ * @param {Node} doc a document node
+ * @return {Boolean}
+ */
+axe.utils.isXHTML = function (doc) {
+	'use strict';
+	if (!doc.createElement) {
+		return false;
+	}
+	return doc.createElement('A').localName === 'A';
+};
--- a/test/core/utils/is-xhtml.js
+++ b/test/core/utils/is-xhtml.js
@@ -1,0 +1,21 @@
+describe('axe.utils.isXHTML', function () {
+	'use strict';
+
+	it('should be a function', function () {
+		assert.isFunction(axe.utils.isXHTML);
+	});
+
+	it('should return true on any document that is XHTML', function () {
+		var doc = document.implementation.createDocument('http://www.w3.org/1999/xhtml', 'html', null);
+		assert.isTrue(axe.utils.isXHTML(doc));
+	});
+
+	it('should return false on any document that is HTML', function () {
+		var doc = document.implementation.createHTMLDocument('Monkeys');
+		assert.isFalse(axe.utils.isXHTML(doc));
+	});
+
+	it('should return false on any document that is HTML - fixture', function () {
+		assert.isFalse(axe.utils.isXHTML(document));
+	});
+});
--- a/test/core/utils/get-selector.js
+++ b/test/core/utils/get-selector.js
@@ -194,6 +194,20 @@ describe('axe.utils.getSelector', function () {
 		assert.equal(result[0], node);
 	});
 
+	it('should work on complex namespaced elements', function () {
+		fixture.innerHTML = '<m:math xmlns:m="http://www.w3.org/1998/Math/MathML">' +
+		  '<m:mi>x</m:mi>' +
+		  '<m:annotation-xml encoding="MathML-Content">' +
+		    '<m:ci>x</m:ci>' +
+		  '</m:annotation-xml>' +
+		'</m:math>';
+		var node = fixture.querySelector('m\\:ci');
+		var sel = axe.utils.getSelector(node);
+		var result = document.querySelectorAll(sel);
+		assert.lengthOf(result, 1);
+		assert.equal(result[0], node);
+	});
+
 	it('shouldn\'t fail if the node\'s parentNode doesnt have children, somehow (Firefox bug)', function () {
 		var sel = axe.utils.getSelector({
 			nodeName: 'a',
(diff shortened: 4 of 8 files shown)


rdeltour:
Not quite sure what should I do about the codacy checks, I tried to follow the code style from the other files. Please LMK if you want me to update the PR!

marcysutton:
Thanks @rdeltour! The codacy thing was just added but it conflicts with our jshint rules, so you can probably disregard for now.

WilcoFiers:
Apologies. I meant to delete the Codacy hook but I hadn't done so yet. Shouldn't turn up on new commits anymore.

test/core/utils/is-xhtml.js
```diff
@@ -0,0 +1,21 @@
+describe('axe.utils.isXHTML', function () {
+	'use strict';
+
+	it('should be a function', function () {
+		assert.isFunction(axe.utils.isXHTML);
+	});
+
+	it('should return true on any document that is XHTML', function () {
+		var doc = document.implementation.createDocument('http://www.w3.org/1999/xhtml', 'html', null);
```
marcysutton: I had no idea this interface existed. 

the-more-you-know

test/xhtml/core/utils/is-xhtml.js
```diff
@@ -0,0 +1,7 @@
+describe('axe.utils.isXHTML', function () {
```
marcysutton: I don't think you need this file, it's redundant with `test/core/utils/is-xhtml.js`.

test/xhtml/core/utils/get-selector.js
```diff
@@ -0,0 +1,24 @@
+
+describe('axe.utils.getSelector', function () {
```
marcysutton: It's a little strange to have a test for `get-selector` duplicated in another directory. Can you make it work in the existing core unit tests by creating a document like in `test/core/utils/is-xhtml.js`?

rdeltour:
Thank you @marcysutton for the review!

re:
> It's a little strange to have a test for get-selector duplicated in another directory. Can you make it work in the existing core unit tests by creating a document like in test/core/utils/is-xhtml.js?

and

> I don't think you need this file, it's redundant with test/core/utils/is-xhtml.js.

let me try to explain why I did this:

As far as I understand, all aXe’s tests are run in an HTML fixture, injected in the `runner.tmpl` template. It means that in those tests, the DOM is **always** an HTML document.

But for some tests, I needed to be in an XHTML context. So, I created another fixture. All the tests in the `xhtml` directory are intended to be run in an XHTML fixture. That's why I also created a new template in [`test/xhtml/runner.tmpl`](https://github.com/dequelabs/axe-core/pull/582/files#diff-bffe78d0b629b92f06c8a9ee4c60d83b), and [added it to the test config](https://github.com/dequelabs/axe-core/pull/582/files#diff-6ec6d1d21440d5c0adbd37c8adec4ce7R25).

The idea is that the `xhtml` directory only contains the test specific to XHTML particularities, and the idea was intentionally to duplicate the structure of other tests.

For instance, if `test/core/utils/is-xhtml.js` contains the usual tests, run in an HTML fixture, and `test/xhtml/core/utils/is-xhtml.js` contains only the tests for `is-xhtml` that depend on an XHTML environment.

If the XHTML-specific tests were in the same directories as their HTML counterparts, they'd need an option to be ignored in the regular HTML-fixture-based tests, and enabled in the XHTML-fixture-based tests. I thought for now it would be clearer to just have to separate directory structures.

I hope it's clear 😅

Alternatively, another option would actually make aXe run **all** the existing tests both in HTML and XHTML environments. That would provide better coverage but would basically duplicate the testing time, so I thought that would be a bit aggressive.
This option would likely require to update `build/task/fixture.js` to generate both an HTML and XHTML fixtures from templates, and `build/test/config.js` to run them. (As far as I can see, I didn't dig very deep).

test/xhtml/core/utils/get-selector.js
rdeltour: see explanation in the general comment below 👇

test/xhtml/core/utils/is-xhtml.js
rdeltour: see explanation in the general comment below 👇

rdeltour: specifically, this will test that the `document` **from the current fixture** is XHTML. I copied the pattern from aXe's pre-existing [`is-html5.js` tests](https://github.com/dequelabs/axe-core/blob/develop/test/commons/dom/is-html5.js#L14)

lib/core/utils/get-selector.js
```diff
@@ -130,7 +129,9 @@ const createSelector = {
  * recognize the element by (IDs, aria roles, custom element names, etc.)
  */
 function getElmFeatures (elm, featureCount) {
-  const nodeName = elm.nodeName.toLowerCase();
+  const nodeName = escapeSelector(axe.utils.isXHTML(document)?
```
WilcoFiers: We could probably just check this once at the top of the file and keep the boolean around instead of having to call this every time this method runs.

lib/core/utils/is-xhtml.js
```diff
@@ -0,0 +1,16 @@
...
+ * @instance
+ * @param {Node} doc a document node
+ * @return {Boolean}
+ */
+axe.utils.isXHTML = function (doc) {
+	'use strict';
+	if (!doc.createElement) {
+		return false;
+	}
+	return doc.createElement('A').localName === 'A';
```
WilcoFiers: What does this do on JSDOM? We should make sure we don't accidentally treat JSDOM as XHTML.

test/core/utils/is-xhtml.js
```diff
@@ -0,0 +1,21 @@
+describe('axe.utils.isXHTML', function () {
```
WilcoFiers: @rdeltour Have you tested this with Edge and IE? This seems like one of those things that might not work across the board :)

test/xhtml/core/utils/get-selector.js
WilcoFiers: This makes sense, although I'd suggest we move it to `tests/integrations` instead.

Gruntfile.js
```diff
@@ -281,6 +281,21 @@ module.exports = function (grunt) {
 					}
 				}
 			},
+			xhtml: {
```
WilcoFiers: I think it would be easier to move this xhtml test into `tests/integrations`. That way we don't have to do all this scaffolding work. WDYT?

rdeltour: these are not strictly speaking integration tests, but if you feel that will fit better there (and disturb less the code organization), sure! let me know.

test/core/utils/is-xhtml.js
rdeltour: Confirmed to work on Edge and IE too.
The logic relies on the fact that HTML parsers lower-case the tag names when parsing. It's been suggested as a possible reliable option by [Anne van Kesteren on Twitter](https://twitter.com/annevk/status/922805253585690624).

lib/core/utils/is-xhtml.js
rdeltour: Yes, confirmed to work on JSDOM.

lib/core/utils/get-selector.js
rdeltour: > We could probably just check this once at the top of the file and keep the boolean around
where would be the best place to memoïze the boolean value? Using a `const` at the top of the file won't work.
Calling the function shouldn't be too costly in any case, so maybe it's too early an optimization?

Gruntfile.js
WilcoFiers: Yep, I think so.

lib/core/utils/get-selector.js
WilcoFiers: This could potentially create many thousands new elements, so I do think this is relevant. Why wouldn't a const up top work? Alternative the result can be stored in utils.isXHTML itself.

Gruntfile.js
rdeltour: ok, I'll try to update the PR accordingly (I'm currently on vacation)

lib/core/utils/get-selector.js
rdeltour: > Why wouldn't a const up top work?

I tried and got an error when running tests :/ But I haven't tried hard nor dug deep into the issue.

WilcoFiers: @rdeltour Can you update this according what we discussed at TPAC?

rdeltour:
> @rdeltour Any updates on this PR?

I was just working on it, in fact 😀. If time allows I will update it this afternoon.

WilcoFiers:
@rdeltour That's good to know. We are planning to release 2.6 today and I'd like to have this in if we could. Otherwise it'll have to wait for 3.0.

rdeltour:
@WilcoFiers ready to be reviewed!

PR #583, opened 2017-10-25
aria-errormessage for the 2x branch

marcysutton:
This PR adds #517 to the 2x branch. Rather than hard-coding a checkContext in the tests, I introduced a basic testutils file to ease the transition when we're finally ready to kill off the 2x branch.

### DIFF
--- a/lib/checks/aria/errormessage.js
+++ b/lib/checks/aria/errormessage.js
@@ -1,0 +1,25 @@
+options = Array.isArray(options) ? options : [];
+
+var attr = node.getAttribute('aria-errormessage'),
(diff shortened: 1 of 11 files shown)


lib/checks/aria/errormessage.js
```diff
@@ -0,0 +1,25 @@
+options = Array.isArray(options) ? options : [];
+
+var attr = node.getAttribute('aria-errormessage'),
+	hasAttr = node.hasAttribute('aria-errormessage');
+
+var doc = document;
```
marcysutton: This line is the only change I had to make to the code added in #517.

Issue #585, opened 2017-10-25
th-has-data-cells confusing error message


mattgarrish:
The error reported in axe is:

> All th element and elements with role=columnheader/rowheader must data cells which it describes

But the linked documentation page has the clearer title:

> All <th> elements and elements with role=columnheader/rowheader must have data cells they describe

Would be nice if the former matched the latter.

marcysutton:
This would be a great thing to fix in a PR, if you're able

marcysutton:
Closed with #586. Thanks again @mattgarrish !

mattgarrish:
NP!

PR #586, opened 2017-10-25
update th-has-data-cells help message

mattgarrish:
Fixes the message to match the prose used in the documentation, as noted in issue #585


marcysutton:
Thanks for the contribution, @mattgarrish!

PR #604, opened 2017-11-06
feat(reporter): return one result of each type instead of zero when r…

### LINKED ISSUE #603: Return 1 result of each type when resultTypes is used
This will allow a caller to differentiate the case of no results from the case where the results were truncated while still maintaining almost all of the performance benefits

dylanb:
…esultTypes is used

Fixes #603

### DIFF
--- a/test/core/reporters/helpers/process-aggregate.js
+++ b/test/core/reporters/helpers/process-aggregate.js
@@ -25,6 +25,25 @@ describe('helpers.processAggregate', function () {
         }],
         all: [],
         none: []
+      }, {
+        result: 'passed',
+        node: {
+          element: document.createElement('div'),
+          selector: 'main > .thing',
+          source: '<div class=\"thing\">Thing</div>',
+          xpath: '/main/div[@class="thing"]'
+        },
+        any: [{
+          id: 'passed-rule',
+          relatedNodes: [{
+            element: document.createElement('div'),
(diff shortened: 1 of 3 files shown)


test/core/reporters/helpers/process-aggregate.js
```diff
@@ -96,12 +135,17 @@ describe('helpers.processAggregate', function () {
...
-        var resultObject = helpers.processAggregate(results, { resultTypes: ['passes', 'violations'] });
+      it('should reduce the unwanted result types to 1 in the `resultObject`', function () {
+        var resultObject = helpers.processAggregate(results, { resultTypes: ['violations'] });
         assert.isDefined(resultObject.passes);
+        assert.equal(resultObject.passes[0].nodes.length, 1);
         assert.isDefined(resultObject.violations);
-        assert.isUndefined(resultObject.incomplete);
-        assert.isUndefined(resultObject.inapplicable);
+        assert.equal(resultObject.violations[0].nodes.length, 2);
+        resultObject = helpers.processAggregate(results, { resultTypes: ['passes'] });
```
marcysutton: Why the second call to `processAggregate` here? Is that what happens when it gets called as a result of an analysis?

WilcoFiers: It's just a second test to make sure it's not just 'violations', but also that it works for 'passes'.

Issue #607, opened 2017-11-06
Review item for contrast of inline element only occurs in some browsers


marcysutton:
This was originally filed as part of the QA pass for our Firefox extension, but it's a problem in axe-core. A review item for an inline link that spans two lines is sometimes found for the ABCD Computech site: http://abcdcomputech.dequecloud.com/

In a handful of browsers, there is a contrast review item for this node: 

```
<a href="contact.php">this page</a>
```

It falls through to the generic "Unable to determine contrast ratio" message. I think it's a problem with the code checking overlap, for which I've seen issues on inline elements. The link in this case wraps to two lines, sounding eerily similar to https://github.com/dequelabs/axe-core/issues/556.

## Browser Results

- aXe Chrome on Mac: review item found
- aXe Firefox on Mac: review item found
- aXe Chrome on Windows: review item found
- aXe Firefox on Windows: review item NOT found

Firefox on Windows is the only one that gets this right (there should be no individual review item). The surrounding body copy is a color violation (4.48:1), but the link itself is fine (9.40:1).

marcysutton:
I've narrowed this to `document.elementsFromPoint`, with which only Firefox on Windows includes the link spanning a line break. It's so odd that there's a difference cross-platform.

Issue #596, opened 2017-11-01
WCAG 2.0 1.4.3 (Incidental)


MrGrigri:
I have come across an error involving WCAG 2.0 1.4.3 (Incidental). I have an input checkbox element that is disabled. I am styling the text next to it utilizing the disabled pseudo-class and adjacent sibling CSS selectors. From what the regulation states for 'Incidental' or disabled text, these are an exemption to the rule. I have a Codepen if you would like to see it in action. I know that it meets the requirement, however, I thought that y'all could update the formula to see if the selector has 'disabled' in it. 

[here](https://codepen.io/MrGrigri/pen/zPOBzp) is a CodePen example.

Issue #556, opened 2017-10-02
False Positive: aXe unable to determine color contrast within nested tags when font-weight is being applied in CSS.


GPFAFF:
Working with the aXe chrome browser extension I noticed some inconsistent behavior.  It is only present when we are applying `font-weight: bold` to a `<b>` or `<span>` inside any parent node.    When the node is on one line presentationally within the DOM, no color contrast flag, but when its container forces a line break it flags it as unable to determine color contrast. 

```
<p class="content">
   Lorem Ipsum <b>Bold Text</b> and a <b>lot more bold text</b>
</p>
```

Attached are pictures of the false positives. 

### offending markup

![screen shot 2017-10-02 at 3 23 54 pm](https://user-images.githubusercontent.com/1102168/31098087-1f279f5c-a787-11e7-86ae-7941966da571.png)

### associated false positive

![screen shot 2017-10-02 at 3 39 58 pm](https://user-images.githubusercontent.com/1102168/31098389-fb88c9f8-a787-11e7-92eb-7a9001fe25eb.png)

marcysutton:
This is a known issue, but thanks for filing it! We have it filed internally but not on Github. And the details are very helpful!

PR #610, opened 2017-11-08
Color fixes: #607, #556, #596

marcysutton:
I fixed two separate color contrast bugs: 

- The sibling of a disabled input inside of a label should be exempt
- Inline elements spanning multiple lines shouldn't be incomplete

I decided against rewriting the color contrast algorithm entirely at this time, since the fix for #607 was as simple as adding a target to the background stack.

### DIFF
--- a/lib/commons/color/get-background-color.js
+++ b/lib/commons/color/get-background-color.js
@@ -121,6 +121,9 @@ function elmPartiallyObscured(elm, bgElm, bgColor) {
  * @param {Element} elm
  */
 function includeMissingElements(elmStack, elm) {
+	if (!elmStack.includes(elm)) {
+		elmStack.unshift(elm);
+	}
 	const elementMap = {'TD': ['TR', 'TBODY'], 'TH': ['TR', 'THEAD'], 'INPUT': ['LABEL']};
 	const tagArray = elmStack.map((elm) => {
 		return elm.tagName;
--- a/lib/rules/color-contrast-matches.js
+++ b/lib/rules/color-contrast-matches.js
@@ -35,8 +35,12 @@ if (nodeName === 'FIELDSET' && node.disabled || axe.commons.dom.findUp(node, 'fi
 var nodeParentLabel = axe.commons.dom.findUp(node, 'label');
 if (nodeName === 'LABEL' || nodeParentLabel) {
 	var relevantNode = node;
+	var relevantVirtualNode = virtualNode;
+
 	if (nodeParentLabel) {
 		relevantNode = nodeParentLabel;
+		// we need an input candidate from a parent to account for label children
+		relevantVirtualNode = axe.utils.getNodeFromTree(axe._tree[0], nodeParentLabel);
 	}
 	// explicit label of disabled input
 	let doc = axe.commons.dom.getRootNode(relevantNode);
@@ -45,7 +49,7 @@ if (nodeName === 'LABEL' || nodeParentLabel) {
 		return false;
 	}
 
-	var candidate = axe.utils.querySelectorAll(virtualNode, 'input:not([type="hidden"]):not([type="image"])' +
+	var candidate = axe.utils.querySelectorAll(relevantVirtualNode, 'input:not([type="hidden"]):not([type="image"])' +
 		':not([type="button"]):not([type="submit"]):not([type="reset"]), select, textarea');
 	if (candidate.length && candidate[0].actualNode.disabled) {
 		return false;
--- a/test/checks/color/color-contrast.js
+++ b/test/checks/color/color-contrast.js
@@ -95,6 +95,21 @@ describe('color-contrast', function () {
 		assert.deepEqual(checkContext._relatedNodes, []);
 	});
 
+	it('should return true for inline elements with sufficient contrast spanning multiple lines', function () {
+		fixture.innerHTML = '<p>Text oh heyyyy <a href="#" id="target">and here\'s <br>a link</a></p>';
+		var target = fixture.querySelector('#target');
+		assert.isTrue(checks['color-contrast'].evaluate.call(checkContext, target));
+		assert.deepEqual(checkContext._relatedNodes, []);
+	});
+
+	it('should return true for inline elements with sufficient contrast', function () {
+		fixture.innerHTML = '<p>Text oh heyyyy <b id="target">and here\'s bold text</b></p>';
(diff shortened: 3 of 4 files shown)


lib/commons/color/get-background-color.js
```diff
@@ -121,6 +121,9 @@ function elmPartiallyObscured(elm, bgElm, bgColor) {
  * @param {Element} elm
  */
 function includeMissingElements(elmStack, elm) {
+	if (!elmStack.includes(elm)) {
+		elmStack.unshift(elm);
```
marcysutton: Can you foresee any problems with putting the target at the front of the element stack? Do we need to check for z-index? This is shimming a failure by `document.elementsFromPoint` to pick up inline elements spanning line breaks.

isner: Since z-index could be manipulated by the page's styles the target element might actually be behind other elements in the stack. That seems problematic to me instinctually, but I am probably missing something.

marcysutton: I'll add another test for that case and see if it causes problems.

lib/commons/color/get-background-color.js
```diff
@@ -188,7 +188,9 @@ function sortPageBackground(elmStack) {
  * @return {Array}
  */
 color.getBackgroundStack = function(elm) {
-	let rect = elm.getBoundingClientRect();
+	// allows inline elements spanning multiple lines to be evaluated
+	let multipleRects = elm.getClientRects();
+	let rect = multipleRects.length > 1 ? multipleRects[0] : elm.getBoundingClientRect();
```
marcysutton: I found that if I used `getClientRects()` to gather an x, y to pass to document.elementsFromPoint, we could leverage that method to find elements on top. The only downside I could see is if a point in rects[1+] didn't overlap, the background stack could be different for those nodes.  A scenario could be if an inline element spanning multiple lines is only partially covered–we could potentially put it into incomplete if there's more than one clientRect but we'd have to check if something with color overlaps partially for that result to be helpful, and that seems like super overkill at this time.

WilcoFiers: I think there's a possible false positive we're ignoring if we're not checking each of those clientRects. I don't think we should put them all in incomplete. Instead I'd suggest we do elementsFromPoint on each of them, and only run it into incomplete if the resulting arrays don't match. WDYT?

marcysutton: @WilcoFiers I found one wrinkle in this plan: if the inline element has a line break in it (a la `<br>`), getClientRects() includes an extra rect for it. Perhaps I need to exclude that case, since it doesn't add anything relevant in terms of text content impacted by color? It would be a shame for axe-core to get hung up on a line break.

marcysutton: I also need a method for comparing arrays of DOM objects, if you have any performant suggestions.

WilcoFiers: No preference. Shallow compare is pretty easy though:

```js
arr1.length == arr2.length && arr1.every((val, i) => val === arr2[i])
```

I wouldn't try to optimise this. This isn't nearly as costly as many other things the color-contrast rule is doing.

marcysutton: Ha, that's really similar to what I ended up doing. But I still have to do a lot of finagling with the data structure. I'm getting close though.

marcysutton:
This PR is so close, but the trickiest of the tests are failing in PhantomJS and not in the browser (of course). This code is purely concerned with collecting the background stack, not about checking computed style. Here's the gist of the changes I had to make to incorporate Wilco's feedback:

- If an inline link spans more than one line, it has multiple clientRects. Check the background stack returned from each DOMRect's `document.elementsFromPoint`. If the elements in the arrays match, use the first array.
- If they don't match, pass a `null` result to trigger a review item and set `elmPartiallyObscuring` for remediation tips.

In writing the logic and tests for this, I found use two cases in conflict with each other:

- multiline inline links that aren't fully covered by a background (these should be undefined/incomplete)
- default implicit label wrapping an input (this should be true)

client rects outlined in element fragments

The default implicit label has multiple clientRects in Chrome/Safari, with the `input` being the only difference. In contrast, the multiline inline link gets a slightly different background stack for each link DOMRect (the elements each fragment covers).

I saw multiple possibilities for discerning between the two cases, but the easiest was to check whether the element under test is in the bounding client rect stack. In every real browser I checked, the multiline link wasn't in the stack returned from `elm.getBoundingClientRect()`, so it was a differentiator from the inline label. I added another condition in the array matching bit to say if the element under test is in the stack from getBoundingClientRect, pass it.

But in Phantom this doesn't work, and debugging it is a huge PITA. I have to put this down again to get ready for dotJS, but this is as far as I got.

Here's a codepen showing the essential problem: https://codepen.io/marcysutton/pen/pdjNqK?editors=0100

test/commons/color/get-background-color.js
```diff
@@ -192,6 +192,39 @@ describe('color.getBackgroundColor', function () {
...
+			'<p style="position: relative;z-index:1;">Text oh heyyyy <a href="#" id="target">and here\'s <br>a link</a></p>' +
+		'</div>';
+		var actual = axe.commons.color.getBackgroundColor(document.getElementById('target'), []);
+		assert.isNotNull(actual);
+		assert.equal(Math.round(actual.blue), 0);
+		assert.equal(Math.round(actual.red), 0);
+		assert.equal(Math.round(actual.green), 0);
+	});
+
+	it('should return null for a multiline block element not fully covering the background', function () {
```
marcysutton: This test fails in Firefox AND Phantom, which is super irritating. Essentially, this test returns an actual color even though half of the element isn't covering its parent. Rather than spending more time trying to fix this case, I'm going to capture the problem in #621 and remove the test from this PR.

lib/commons/color/get-background-color.js
```diff
@@ -203,7 +201,95 @@ color.getBackgroundStack = function(elm) {
...
+ * @param {Element} elm
+ * @return {Array}
+ */
+color.filteredRectStack = function(elm) {
+	let rectStack = color.getRectStack(elm);
+	if (rectStack && rectStack.length === 1) {
+		// default case, elm.getBoundingClientRect()
+		return rectStack[0];
+	}
+	else if (rectStack && rectStack.length > 1) {
```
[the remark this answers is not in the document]
marcysutton: Fixed.

PR #612, opened 2017-11-13
Enable best practices

### LINKED ISSUE #515: Consistency when enabling/disabling best practice rules
The "tabindex greater than 0" rule is a best-practice and so it shouldn't run in the axe-extension default. Reported by Dave Musser.

WilcoFiers:
Closes #515. This should not be put in develop-2x

BREAKING CHANGE: Skip-link rule no longer requires skip links with a focusable target.


dylanb:
@WilcoFiers what was the rationale behind not requiring a focusable target?

WilcoFiers:
@dylanb Chrome used to have the issue that if you had a skiplink to a non-focusable element, it would just blur in place, and the next item in the focus order would just be whatever comes after the skiplink. Other browsers blurred on the target of the skiplink, so that the next focusable element would be the thing after the *target* of the skiplink. After some testing it looks like this issue has been resolved, so skiplinks now work everywhere, regardless of if the target is focusable or not.

joelhsmith:
Excellent! Thank you all!

marcysutton:
I merged this into `develop-2x`.

PR #614, opened 2017-11-14
feat: header-one best practice rule added

### LINKED ISSUE #380: Remove race conditions from iframe integration tests
Most of our iframe tests currently rely on timing (or timeouts) to attempt to ensure that the iframe content is loaded when the tests start.

The result is frustrating failing tests and random setTimeouts strewn throughout the code.

We could utilize the approach that axe-core uses to determine which frames should return results, to also synchronize our integration tests.

[This change](https://github.com/dequelabs/axe-core/pull/373/files#diff-014284edec95675490b57945fb449ce7R6) shows how this can be done for frames that already should have axe.js in them - but this will not ensure correct loading of frames without axe.js and also will not cascade to child iframes.

So what we should do is write a utility function that will cascade a ping (that is repeated on an interval) to all the child frames and when these return, respond to pings from its parent. Tests would then make a (repeated) ping call to all their child iframes in the `before` to ensure that they (and their child iframes) are loaded.

WilcoFiers:
Also closes #380.


WilcoFiers:
@dylanb I considered that. The reason I didn't do that was because, even though that's certainly *less bad* if you have at least one frame with an `h1`, I still think we can say that it's better if they all had `h1`. I couldn't come up with a scenario where, for visible frames, that wouldn't be the case. If you've got a valid case then that would be a false positive.

dylanb:
A case where this is possible is when including something like Bazaarvoice as an iframe - the main content for the page could all be in the Bazaarvoice iframe and this would generate a false positive for the containing document

WilcoFiers:
Good point. I'll work on a fix.

marcysutton:
@wilcofiers are you still working on a fix for this PR?

WilcoFiers:
@marcysutton Yeah, from time to time. It's still on my radar.

WilcoFiers:
This rule needs a completely different approach. I'm closing this PR and starting with a clean slate.

PR #624, opened 2017-11-25
fix: don't put offscreen file inputs into review

### LINKED ISSUE #622: Color contrast: visually hidden file input should not go into needs review
In [Sarah Drasner's alt text generator Codepen](https://s.codepen.io/sdras/debug/9b9d4ab09d8a79bf295b07044b268f39), the visually hidden file input is going into needs review with the generic message `Unable to determine contrast ratio`.

visually hidden input open in Chrome devtools

The label shows a focus state when the user tabs onto the input, which is hidden from view with this CSS:
```
.inputfile {
    width: 0.1px;
    height: 0.1px;
    opacity: 0;
    overflow: hidden;
    position: absolute;
    z-index: -1;
}
.inputfile:focus + label {
    outline: 1px dotted #000;
    outline: -webkit-focus-ring-color auto 5px;
}
```
The important bit in there for contrast purposes is `opacity:0`. If an element is invisible, we shouldn't put it into needs review for color contrast. We can expand `axe.commons.dom.isVisible` to check for opacity:0, and then exclude it from the color-contrast-matches. Would that cause any issues? I've already got a fix for it, going to submit it as a separate PR.

https://s.codepen.io/sdras/debug/9b9d4ab09d8a79bf295b07044b268f39

marcysutton:
Closes #622.

I used `opacity: 0` and `position: absolute` to determine visibility. In my mind absolute positioning makes it more clearly an offscreen/visibility technique. But perhaps it should simply check opacity, since that's all that really matters for color contrast? I'm open to changes.

### DIFF
--- a/lib/commons/dom/is-visible.js
+++ b/lib/commons/dom/is-visible.js
@@ -20,6 +20,17 @@ function isClipped(clip) {
 	return false;
 
 }
+/**
+ * Determines if an element is hidden with opacity and position
+ * @method invisibleInPlace
+ * @memberof axe.commons.dom
+ * @private
+ * @param  {Object}  style Computed style value of node
+ * @return {Boolean}
+ */
+function invisibleInPlace(style) {
+	return style.getPropertyValue('opacity') === '0' && style.getPropertyValue('position') === 'absolute';
+}
 
 /**
  * Determine whether an element is visible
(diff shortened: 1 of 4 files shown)


rajrajiv6:
Improve

dylanb:
Hey - does this still need review?

marcysutton:
@dylanb I closed it based on comments from Wilco, which I added to the issue: #622

PR #626, opened 2017-11-29
fix(check): fix bug with async checks

justinhorvitz:
The correct field to set on CheckResult is 'result' not 'value'

### DIFF
--- a/lib/core/utils/check-helper.js
+++ b/lib/core/utils/check-helper.js
@@ -13,7 +13,7 @@ axe.utils.checkHelper = function checkHelper(checkResult, options, resolve, reje
 			this.isAsync = true;
 			return function (result) {
 				if (result instanceof Error === false) {
-					checkResult.value = result;
+					checkResult.result = result;
 					resolve(checkResult);
 				} else {
 					reject(result);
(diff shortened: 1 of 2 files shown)


marcysutton:
Was this not working before? A related issue would help.

WilcoFiers:
I don't think it was. I had tried to write the frames-untested rule and got stuck on the async stuff too. I never had a chance to look into it. But I'll have a look when checking this PR. Thanks for contributing @justinhorvitz.

justinhorvitz:
Since checkResult.result is undefined, all async checks will currently be considered incomplete.

dylanb:
@justinhorvitz This problem underscores a gap in our integration tests - can you add an integration test as part of this change?

justinhorvitz:
Unfortunately I don't believe I'll have the time to figure out how to add an integration test.

WilcoFiers:
I'm happy to work on that @justinhorvitz. Thanks for the contribution.

dylanb:
@WilcoFiers Do you know why the Circle tests are failing on this PR? I pulled the PR down onto my local machine and ran all tests and they passed.

WilcoFiers:
@dylanb No clue, I a rerun fixed it though. We do get the occasional PhantomJS timeout. Nothing to worry about I think.

WilcoFiers:
Closing the PR and continue the work in #630.

Issue #540, opened 2017-09-20
Incorrect aXe / Assure violation: Elements must only use allowed ARIA attributes: for aria-required on radio button


spanchang:
Example: `<input aria-required="true" name="contactme" value="email" type="radio">`

The ARIA specs note that 
i. aria-required is related to the HTML 'required' attribute
ii. Unless an exactly equivalent native attribute is available, host languages SHOULD allow authors to use the aria-required attribute on host language form elements that require input or selection by the user.

The HTML5 specs say:
For radio buttons, the required attribute is satisfied if any of the radio buttons in the group is selected.

To avoid confusion as to whether a  radio button group  is required or not, authors are encouraged to specify the attribute on all the radio buttons in a group. 
Indeed, in general, authors are encouraged to avoid having radio button groups that do not have any initially checked controls in the first place, as this is a state that the user cannot return to, and is therefore generally considered a poor user interface.

===

Also IMO: It cannot be a best practice recommendation. Not having the attribute on  radio button(s) does not enhance accessibility nor is it counter to the specs.

WilcoFiers:
Yep, I think you're right. This looks fine to me. I did a quick test with VO, which doesn't seem to mind. I don't expect other AT will mind either but we'll need to be sure. Have you tested this with NVDA and JAWS to see if there are any support issues that come up?

spanchang:
I do not encounter any problems with JAWS / NVDA.
Thanks,
Sailesh

On 9/20/17, Wilco Fiers <notifications@github.com> wrote:
> Yep, I think you're right. This looks fine to me. I did a quick test with
> VO, which doesn't seem to mind. I don't expect other AT will mind either but
> we'll need to be sure. Have you tested this with NVDA and JAWS to see if
> there are any support issues that come up?
>
> --
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/dequelabs/axe-core/issues/540#issuecomment-330886190


-- 
Sailesh Panchang
Principal Accessibility Consultant
Deque Systems Inc
Phone 703-225-0380 ext 105
Mobile: 571-344-1765

marcysutton:
We should check Mobile VO and TalkBack, as well.

spanchang:
How At/ browser combos behave with standard code should not dictate
whether something is a failure or not. If the AT/browser cannot handle
the required attribute or aria-required, it is a limitation of that
combo / platform and not a WCAG2 failure.
[...]

PR #629, opened 2017-12-01
fix(commons/aria): allow aria-required on radio and checkbox roles

HugoGiraudel:
This pull-request closes https://github.com/dequelabs/axe-core/issues/540.

### DIFF
--- a/lib/commons/aria/index.js
+++ b/lib/commons/aria/index.js
@@ -617,7 +617,7 @@ lookupTables.role = {
 	'radio': {
 		type: 'widget',
 		attributes:  {
-			allowed: ['aria-checked', 'aria-selected', 'aria-posinset', 'aria-setsize'],
+			allowed: ['aria-checked', 'aria-selected', 'aria-posinset', 'aria-setsize', 'aria-required'],
 		},
 		owned: null,
 		nameFrom: ['author', 'contents'],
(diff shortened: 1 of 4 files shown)


WilcoFiers:
I think we should have this for `checkbox` as well. Can you also add one or two tests?

HugoGiraudel:
I wasn‘t sure if you wanted it for checkbox. Happy to add it. Not sure how to add the tests though. Care to put me on the right track?

WilcoFiers:
Have a look at some of these:
- https://github.com/dequelabs/axe-core/blob/develop/test/checks/aria/required-attr.js
- https://github.com/dequelabs/axe-core/blob/develop/test/checks/aria/valid-attr-value.js
- https://github.com/dequelabs/axe-core/blob/develop/test/checks/aria/errormessage.js

HugoGiraudel:
I run through these files. I’m not sure what kind of test you would like for allowing `aria-required` attribute on `radio` and `checkbox` roles. I didn’t find a suite I could extend with that kind of check.

Do you test all allowed attributes somewhere maybe?

WilcoFiers:
We're not testing every attribute, no. But we are adding tests for every PR so that we can prevent regression on new issues. On second thought, maybe adding it as an integration tests is better:

https://github.com/dequelabs/axe-core/blob/develop/test/integration/rules/aria-allowed-attr/passes.html
https://github.com/dequelabs/axe-core/blob/develop/test/integration/rules/aria-allowed-attr/passes.json

Does that help?

HugoGiraudel:
I just committed something that I thought could be useful. Is it enough or would you like me to check integration tests as well?

PR #635, opened 2017-12-09
Get develop-2x up to date with support docs

### LINKED ISSUE #616: Expose lookupTables through public API
As part of our ongoing Support agreement, we had a request to expose the ARIA lookupTables through the public axe-core API. These are pretty stable, so I don't anticipate much work in the way of maintaining backwards compatibility.

I believe the goal is to expose the roles, attributes, and global attributes inside of rules and checks.

https://github.com/dequelabs/axe-core/blob/develop/lib/commons/aria/index.js

### LINKED ISSUE #617: Expose select commons/utils functions via public API
As part of our ongoing Support agreement, we need to expose select commons and util functions through the public API. For now we just need a few to start, but we may determine it makes more sense to expose an entire set at once.

- [aria.implicitRole](https://github.com/dequelabs/axe-core/blob/develop-2x/lib/commons/aria/roles.js#L105)
- [aria.label](https://github.com/dequelabs/axe-core/blob/develop-2x/lib/commons/aria/label.js)
- [dom.isVisible](https://github.com/dequelabs/axe-core/blob/develop-2x/lib/commons/dom/is-visible.js#L28)

marcysutton:
In adding docs for #616 and #617, I wanted to start with Tony's great work on jsdoc. It was pretty intertwined with Shadow DOM, so I spent quite a bit of time merging. Fortunately, this process pulled in a few PRs that we'd missed on the 2x branch, most notably some ARIA 1.1 and color contrast message stuff. I was careful not to include anything related to Shadow DOM, but this would benefit from a thorough review just to make sure.

I think commons functions are best documented in the developer guide, so I added them there. We can a) add more details and b) include more commons functions, if desired.  I think the jsdoc stuff Tony added helps a lot to clarify the source code.

Closes #616 
Closes #617


doc/developer-guide.md
```diff
@@ -115,9 +126,20 @@ return results.filter(function (r) {
...
-#### Pass and Fail Templates
+#### Pass, Fail and Incomplete Templates
+
+Occasionally, you may want to add additional information about why a Check passed, failed or returned undefined into its message.  For example, the [aria-valid-attr](../lib/checks/aria/valid-attr.json) will add information about any invalid ARIA attributes to its fail message.  The message uses the [doT.js](http://olado.github.io/doT/) and is compiled to a JavaScript function at build-time.  In the Check message, you have access to the `CheckResult` as `it`.
+
+```javascript
+// aria-valid-attr check
+"messages": {
+  "pass": "ARIA attributes are used correctly for the defined role",
+  "fail": "ARIA attribute{{=it.data && it.data.length > 1 ? 's are' : ' is'}} not allowed:{{~it.data:value}} {{=value}}{{~}}"
```
WilcoFiers: Should this have an `incomplete` example?

lib/commons/aria/attributes.js
```diff
@@ -1,43 +1,55 @@
-/*global aria, axe, lookupTables */
+/*global aria, axe, lookupTable */
```
WilcoFiers: We should pull these from `aria.lookupTable` instead of relying on an implicit global.

lib/commons/aria/roles.js
```diff
@@ -1,61 +1,76 @@
-/*global aria, lookupTables, axe */
+/* global aria, lookupTable, axe */
```
WilcoFiers: Same here, let's take this from `aria.lookupTable` instead of using a global.

doc/developer-guide.md
```diff
@@ -115,9 +126,20 @@ return results.filter(function (r) {
...
+}
+```
 
-Occasionally, you may want to add additional information about why a Check passed or failed into its message.  For example, the [aria-valid-attr](../lib/checks/aria/valid-attr.json) will add information about any invalid ARIA attributes to its fail message.  The message uses the [doT.js](http://olado.github.io/doT/) and is compiled to a JavaScript function at build-time.  In the Check message, you have access to the `CheckResult` as `it`.
+See [Developing Axe-core Rules](./rule-development.md) for more information
```
marcysutton: This section made it in twice somehow. Will remove.

doc/developer-guide.md
marcysutton: Probably, although I just took what was on jsdoc/develop.

iamrafan:
@marcysutton Can you please add #555 also to the 2x branch.

marcysutton:
@iamrafan the changes are in there, but the history got lost. I'm trying to see if I can recreate the commit history without having to do the whole merge by hand again.

doc/developer-guide.md
marcysutton: @WilcoFiers the previous sentence mentions the rule-development doc, which has thorough instructions for incomplete messages. I'd prefer not to have it in multiple places.

marcysutton:
I'm going to close this in favor of #636, which includes the commit history.

PR #636, opened 2017-12-11
Get develop-2x up to date with support docs (with commit history)

### LINKED ISSUE #616: Expose lookupTables through public API

### LINKED ISSUE #617: Expose select commons/utils functions via public API

marcysutton:
I redid this merge by cherry-picking commits so they would be included in the commit history, instead of a new commit from merging the jsdoc branch. It should be more helpful for our changelog as compared to the other one.. Hopefully I didn't miss anything.

@WilcoFiers to answer your question from the previous PR, we can cherry-pick the lookupTable commit into 3x.

Closes #616
Closes #617


iamrafan:
@marcysutton hey, I think #559 is still missing, can you please add it

marcysutton:
Hi @iamrafan! I assumed that was covered by https://github.com/dequelabs/axe-core/commit/86b4a0e3b1e6951221723698a722dfbfd31e1eff. It looked like the work had already been done in the 2x branch. @isner, can you confirm?

iamrafan:
@marcysutton Yes, my bad. It has already been added to 2x branch. I missed out as I could not find it in the release notes and the commit message being different here.

marcysutton:
It is in the changelog for 2.5.0, but the commit message could be a bit more descriptive with "Closes #559".

iamrafan:
Sure, thanks.

PR #638, opened 2017-12-12
Refactor dom.isFocusable to expose dom.isNativelyFocusable for convenience

0ddfell0w:
Refactor dom.isFocusable to expose dom.isNativelyFocusable. This allows us to distinguish between elements that are in the focus order because of tabindex and elements that are focusable by default.

This allows us to identify elements that are manually inserted into the tab order (i.e. isFocusable && !isNativelyFocusable), or to identify elements with redundant tabindex (e.g. input[tabindex=0])

There is no change in the logic of isFocusable, but it's also true that isNativelyFocusable is not **yet** being used by checks in axe-core, but it's convenient for tools wrapping/extending axe and for the authoring of future rules.

### DIFF
--- a/lib/commons/dom/is-focusable.js
+++ b/lib/commons/dom/is-focusable.js
@@ -12,6 +12,30 @@
 dom.isFocusable = function (el) {
 	'use strict';
 
+	if (dom.isNativelyFocusable(el)) {
+		return true;
+	}
+	// check if the tabindex is specified and a parseable number
+	var tabindex = el.getAttribute('tabindex');
+	if (tabindex && !isNaN(parseInt(tabindex, 10))) {
+		return true;
+	}
+
+	return false;
+};
+
+/**
+ * Determines if an element is focusable without considering its tabindex
+ * @method isNativelyFocusable
+ * @memberof axe.commons.dom
+ * @instance
+ * @param {HTMLElement} el The HTMLElement
+ * @return {Boolean} True if the element is in the focus order but wouldn't be
+ * if its tabindex were removed. Else, false.
+ */
+dom.isNativelyFocusable = function(el) {
+	'use strict';
+
 	if (!el ||
 		el.disabled ||
 		(!dom.isVisible(el) && el.nodeName.toUpperCase() !== 'AREA')) {
@@ -33,12 +57,5 @@ dom.isFocusable = function (el) {
 		case 'BUTTON':
 			return true;
 	}
-
-	// check if the tabindex is specified and a parseable number
-	var tabindex = el.getAttribute('tabindex');
-	if (tabindex && !isNaN(parseInt(tabindex, 10))) {
-		return true;
-	}
-
 	return false;
 };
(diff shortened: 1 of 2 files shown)


marcysutton:
This needs some tests to go along with the change.

Also, shouldn't we check for `tabindex="-1"` when evaluating natively focusable elements, since that changes the tab order?

marcysutton:
The problem I have with this API is it returns a boolean for any natively focusable element regardless of tabindex. So in that case, `<button tabindex="-1">` returns `true` from `isFocusable` by way of `isNativelyFocusable`, even though it isn't in the tab order. That makes `isFocusable` a misleading API method. 

As I mentioned, this distinction predates this PR, which I can tell by looking at the [tests](https://github.com/dequelabs/axe-core/blob/develop/test/commons/dom/is-focusable.js)–there are no `tabindex="-1"` cases on native elements.

I'm saying the native check inside of `isFocusable` _should_ take `tabindex="-1"` into account, instead of returning immediately with a boolean. You could still call `isNativelyFocusable` separately. 

We need tests for these cases–I can help you write some of them, if needed.

0ddfell0w:
isFocusable is supposed to return true for <button tabindex="-1"> because focus can be applied with a call to button.focus(). As far as I can tell, that's what "focusable" means in this case, which is why we see [code that checks isFocusable() && tabIndex > 1](https://github.com/dequelabs/axe-core/blob/develop/lib/checks/keyboard/focusable-no-name.js#L2) to see if an element is in the focus order.

The only way to make a button return false for isFocusable is to hide/disable it, which is the correct behavior. A visible, enabled button with tabindex=-1 is BOTH natively focusable and focusable because of tabindex being present. Thus, I don't think it's misleading for isNativelyFocusable to return true in this case.

0ddfell0w:
Note that 'isFocusable' is a misleading variable name in [the code sample above](https://github.com/dequelabs/axe-core/blob/develop/lib/checks/keyboard/focusable-no-name.js#L2) because it really represents whether or not the element is in the focus order. Should probably be renamed inFocusOrder

marcysutton:
Ok, fair enough. It is definitely a confusing API name. But at this point, we have to weigh renaming with backwards compatibility. @WilcoFiers any thoughts on this?

This PR still needs tests to be accepted.

WilcoFiers:
I'm fine with this change, as long as you add tests for this new method.

@0ddfell0w We should also note that `axe.commons` is not part of the official axe-core API. Which means no promise of backward compatibility, and no semantic versioning for those functions. We can change them or remove them in minor version of axe-core. We try not to, we know people use them, but we could if we had to. Just so you're aware that this is what you're building on top of. If you want to be safe, lock into a specific version of axe-core and upgrade with caution.

0ddfell0w:
Added tests, thanks!

0ddfell0w:
I would also like to flag some issues with isNativelyFocusable, namely that the check about visibility is too vague.

Elements hidden offscreen with absolute positioning can be forced into the focus order (we come across this more than we'd like), but elements with display:none or visibility:hidden are unfocusable. I can follow up with fixes after this PR is merged and the method is exposed? Demo: https://codepen.io/0ddfell0w/full/KZpZVZ

WilcoFiers:
I've opened a separate issue for the off screen elements.

PR #642, opened 2017-12-13
fix(rules/region): Treat <section> as a landmark if it has an accessible name #640

### LINKED ISSUE #640: Rule: 'region' - <section> element with non-empty 'aria-label' or 'aria-labelledby' or 'title' attributes is not being detected as a landmark
```
<section aria-label="name1">
        <h1>Section-1: With aria-label</h1>
        <p>Content 1: Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et
            dolore magna aliqua</p>
    </section>

    <section aria-labelledby="name2">
        <h1 id="name2">Section-2: With aria-labelledby</h1>
        <p>Content 2: Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat</p>
    </section>

    <section title="name3">
        <h1>Section-3: With title</h1>
        <p>Content 3: Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur</p>
    </section>
```

WilcoFiers:
This closes #640.


lib/checks/navigation/region.js
```diff
@@ -26,7 +26,9 @@ function isLandmark (node) {
 	if (node.hasAttribute('role')) {
 		return landmarkRoles.includes(node.getAttribute('role').toLowerCase());
 	} else {
-		return implicitLandmarks.includes(node.nodeName.toUpperCase());
+		return implicitLandmarks.some((implicitSelector) => {
```
marcysutton: That looks like magic to me. Can you add a comment about what's happening here?

isner: Looks like he's electing to use the `matchesSelector` utility to determine whether the node has an implicit landmark role, instead of simply checking whether the array of of implicit landmark roles includes the nodeName. I'm guessing this wasn't working for all nodes, since the nodeName of an element with an implicit landmark role is often different than the landmark role. E.g., given a `<header>` element, `#nodeName === 'HEADER'`, but its implicit role is 'banner'.

isner: This also allows the values in the role's "implicit" entry in the lookup table to be CSS selectors more complex than just the nodeName. E.g., `input[type="button"]`

marcysutton: Thanks! I'd still love to see a comment in the code for future reading.

WilcoFiers:
Also merged into 2x

iamrafan:
@WilcoFiers Thank you

PR #646, opened 2017-12-14
feat(rule): add check for applet-alt

### LINKED ISSUE #536: New rule: H35 - applet-alt
#  H35 - applet-alt

This test checks if a text alternative is provided for an applet by using the `alt` attribute to label an applet and provide the text alternative in the body of the `applet` element. 

**Tags:** wcag2a, wcag111, cat.text-alternatives

## Selector

Select each `applet` element.

### text-alternatives/applet-alt-alt 
1.    Check that the `applet` element contains an `alt` attribute with a text alternative for the `applet`: use `has-alt` rule.
### text-alternatives/applet-alt-text 
1.    Check that the `applet` element contains a text alternative for the `applet` in the body of the `applet` element.
### text-alternatives/applet-alt-aria 
1.    Check whether the text alternative is set with use of ARIA. Look for aria-label, aria-labelledby attributes.

Danidude:
Check for alternative text on applet tags

Close #536

### DIFF
--- a/lib/checks/label/applet-alt-alt.js
+++ b/lib/checks/label/applet-alt-alt.js
@@ -1,0 +1,4 @@
+let label = node.getAttribute('alt');
+let text = node.textContent;
+
+return (!!label && axe.commons.text.sanitize(label).trim() !== '' && axe.commons.text.sanitize(text).trim() !== '');
--- a/lib/checks/label/applet-alt-alt.json
+++ b/lib/checks/label/applet-alt-alt.json
@@ -1,0 +1,11 @@
+{
+  "id":"applet-alt-alt",
+  "evaluate": "applet-alt-alt.js",
+  "metadata": {
+    "impact": "serious",
+    "messages": {
+      "pass": "applet-alt got text",
+      "fail": "applet-alt is not there"
+    }
+  }
+}
--- a/lib/checks/shared/none-empty-text.js
+++ b/lib/checks/shared/none-empty-text.js
@@ -1,0 +1,2 @@
+var text = node.textContent;
+return !(!text || axe.commons.text.sanitize(text).trim() === '');
--- a/lib/rules/applet-alt.json
+++ b/lib/rules/applet-alt.json
@@ -1,0 +1,22 @@
+{
+  "id": "applet-alt",
+  "selector": "applet",
+  "matches": "",
+  "tags": [
+    "cat.text-alternatives",
+    "wcag2a",
+    "wcag111"
+  ],
+  "metadata": {
+    "description": "Checks if a text alternative is provided for an applet",
+    "help": "Label an applet by using the alt/labelledby/label text alternative and provide a text alternative in the body of the applet element "
+  },
+  "all": ["has-visible-text"],
+  "any": [
+        "aria-labelledby",
+        "applet-alt-alt",
+        "aria-label",
+        "none-empty-text"
+  ],
+  "none": []
+}
--- a/lib/checks/shared/none-empty-text.json
+++ b/lib/checks/shared/none-empty-text.json
@@ -1,0 +1,11 @@
+{
+  "id":"none-empty-text",
+  "evaluate": "none-empty-text.js",
+  "metadata": {
+    "impact": "serious",
+    "messages": {
+      "pass": "element got text",
+      "fail": "element got no text"
+    }
+  }
+}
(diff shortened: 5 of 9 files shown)


lib/checks/label/applet-alt-alt.json
```diff
@@ -0,0 +1,11 @@
+{
+  "id":"applet-alt-alt",
+  "evaluate": "applet-alt-alt.js",
+  "metadata": {
+    "impact": "serious",
+    "messages": {
+      "pass": "applet-alt got text",
```
marcysutton: These messages should have full sentences.

lib/rules/applet-alt.json
```diff
@@ -0,0 +1,22 @@
...
+  "id": "applet-alt",
+  "selector": "applet",
+  "matches": "",
+  "tags": [
+    "cat.text-alternatives",
+    "wcag2a",
+    "wcag111"
+  ],
+  "metadata": {
+    "description": "Checks if a text alternative is provided for an applet",
```
marcysutton: Description should be in the active tense, like "Applet must have a text alternative"

lib/rules/applet-alt.json
```diff
@@ -0,0 +1,22 @@
...
+    "help": "Label an applet by using the alt/labelledby/label text alternative and provide a text alternative in the body of the applet element "
```
marcysutton: The help text goes in the sidebar of the aXe extensions, so this needs to be shorter. You could swap description and help text to accomplish this. See my note about the active tense.

lib/rules/applet-alt.json
```diff
@@ -0,0 +1,22 @@
+{
+  "id": "applet-alt",
+  "selector": "applet",
+  "matches": "",
```
WilcoFiers: Remove this property if you're not using it.

lib/checks/label/applet-alt-alt.json
```diff
@@ -0,0 +1,11 @@
+{
+  "id":"applet-alt-alt",
```
WilcoFiers: Why is this called `applet-alt-alt`? Isn't `applet-alt` sufficient?

lib/checks/label/applet-alt-alt.js
```diff
@@ -0,0 +1,4 @@
+let label = node.getAttribute('alt');
```
WilcoFiers: This check shouldn't look at the content, you're already doing this with `none-empty-text`. I also think this check should be renamed to `none-empty-alt` or something like it. There is nothing specific to applets in this check to warrant putting `applet` in the check ID.

lib/checks/shared/none-empty-text.js
```diff
@@ -0,0 +1,2 @@
+var text = node.textContent;
```
WilcoFiers: Should this check consider other content types such as images / video / frames, etc? If so, you may want to use `has-visible-text` instead.

lib/checks/label/applet-alt-alt.js
Danidude: The reason I am looking at content in this context is because as you have stated, VoiceOver do not detect the alt on the applet tag. And as mentioned in the docs for H35(https://www.w3.org/TR/WCAG20-TECHS/H35.html), the support of the alt tag is varying in browsers. So there needs to be text inside the applet tag to have visible text. With area label or labeledby this is not needed.

And if you have any suggestions to check this differently I am all ears. I did forget to test with an applet loaded into the applet tag, and will do this as soon as possible :)

marcysutton:
@danidude were you able to do more AT testing, and what was the result?

Danidude:
After some more testing on applet tags with newer version of safari, I can not get the applet tag to work. With some reading on applet tags, it looks like most browsers now have set it as obsolete in favour for the object tag.

I have had different experience on different browsers with applet. In the safari I do not get anything from applet tags(Using tablet, I can see that the source code have the tags). In both Chrome and Firefox on Linux I do see the applet tag and its contents, however applet inside these do not work.

With some further reading it do looks like Mozilla Firefox removed support for Java plug-ins last summer (https://java.com/en/download/help/firefox_java.xml)

With this little support for this tag, and with this little accessibility. what are the thoughts on the applet tag? We could remove this test and ignore applet tags, or an other idea is to test for applet tags and give an warring to use object instead?

WilcoFiers:
I don't think it's up to axe-core to encourage people to use `object` over `applet`. That's not an accessibility question. If the technique for giving it an accessible name works in all the browsers that still support this, this rule makes sense. If you feel it's not worth the effort you certainly can withdraw it. I don't expect we'd ever build this in ourselves.

PR #652, opened 2017-12-15
Fix the `dlitem` rule for XHTML documents

### LINKED ISSUE #581: inconsistent dlitem result depending on file extension
DT and DD elements are always flagged as not being contained in a DL element when they're inside a document with an xhtml extension, but not when in one with an html extension.

For example, these are identical documents, but only the latter generates an error:
http://matt.garrish.ca/res/pr01.html
http://matt.garrish.ca/res/pr01.xhtml

I believe this is a similar problem with nodeName as reported in https://github.com/dequelabs/axe-core/issues/563

rdeltour:
In XHTML, `element.tagName` preserves the case.

In the `dlitem` check, the tag name is now upper-cased before
being tested as '`DL`', to make the rules works on XHTML documents.

Fixes #581

### DIFF
--- a/lib/checks/lists/dlitem.js
+++ b/lib/checks/lists/dlitem.js
@@ -1,2 +1,2 @@
-return node.parentNode.tagName === 'DL';
+return node.parentNode.tagName.toUpperCase() === 'DL';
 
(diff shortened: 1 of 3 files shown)


rdeltour:
@WilcoFiers ready to be reviewed. This is a very simple fix, so maybe it can be added to 2.6? (sorry for the late submission).

Note that this only affects the 2.x branch.

The CI failures seem to be unrelated to this PR, the very same tests fail when I'm on the current head of `develop-2x`.

PR #654, opened 2017-12-15
fix(commons/dom): change isFocusable to use unfocusableBecauseHidden

### LINKED ISSUE #647: Bug: dom.isFocusable mistakenly concludes that invisible elements are unfocusable

0ddfell0w:
change isFocusable to use unfocusableBecauseHidden function instead of isVisible

Adds functionality for detecting when hidden elements are still
focusable or in focus order.

fixes #647

Demo showing that off-screen and clipped elements remain in the focus order, where display:none and visibility:hidden elements do not (they are totally unfocusable)

### DIFF
--- a/lib/commons/dom/is-focusable.js
+++ b/lib/commons/dom/is-focusable.js
@@ -1,5 +1,19 @@
 /* global dom */
 /* jshint maxcomplexity: 20 */
+
+/**
+ * Determines if an element is hidden in such a way that makes it unfocusable
+ * @private
+ * @param {Element} el
+ * @return {Boolean} Whether the element is unfocusable due to being hidden
+ */
+function unfocusableBecauseHidden (el) {
+	var style = window.getComputedStyle(el);
+	return (
+		style === null || style.getPropertyValue('visibility') === 'hidden' ||
+		style.getPropertyValue('display') === 'none');
+}
+
 /**
  * Determines if an element is focusable
  * @method isFocusable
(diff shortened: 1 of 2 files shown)


dylanb:
Ok, this commit is totally confusing

`isVisible` has two modes:

1. screen reader mode - where content that is clipped, offscreen or otherwise accessible to the screen reader (or a keyboard) will return `true` if "visible" or `false` othwreise, and
1. default mode, where it will return `true` if the content is visible to sighted users AND screen readers

It seems like this same change could be achieved by calling `isVisible` in screen reader mode.

0ddfell0w:
Sure, that's fine too

dylanb:
@0ddfell0w @WilcoFiers I do wonder whether this new method might introduce some performance improvements though - thoughts on this versus "duplicate code"?

0ddfell0w:
https://github.com/dequelabs/axe-core/pull/655 for your suggestion

lib/commons/dom/is-focusable.js
```diff
@@ -1,5 +1,19 @@
 /* global dom */
 /* jshint maxcomplexity: 20 */
+
+/**
+ * Determines if an element is hidden in such a way that makes it unfocusable
+ * @private
+ * @param {Element} el
+ * @return {Boolean} Whether the element is unfocusable due to being hidden
+ */
+function unfocusableBecauseHidden (el) {
```
dylanb: If we keep this for performance - what about parent-child relationships? This function is a bit naive - look at isVisible for all the conditions that must be handled. Also add tests to handle all those conditions.

dylanb: Also, is it possible to do a faster implementation like https://github.com/jquery/jquery/blob/2d4f53416e5f74fa98e0c1d66b6f3c285a12f0ce/src/css/hiddenVisibleSelectors.js#L8

PR #655, opened 2017-12-15
fix(commons/dom): change isFocusable to check visibility in screenreader mode

### LINKED ISSUE #647: Bug: dom.isFocusable mistakenly concludes that invisible elements are unfocusable

0ddfell0w:
Lets isFocusable tell the difference between elements that are still focusable despite being invisible

fixes #647

### DIFF
--- a/lib/commons/dom/is-focusable.js
+++ b/lib/commons/dom/is-focusable.js
@@ -38,7 +38,7 @@ dom.isNativelyFocusable = function(el) {
 
 	if (!el ||
 		el.disabled ||
-		(!dom.isVisible(el) && el.nodeName.toUpperCase() !== 'AREA')) {
+		(!dom.isVisible(el, true) && el.nodeName.toUpperCase() !== 'AREA')) {
 		return false;
 	}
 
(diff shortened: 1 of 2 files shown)


0ddfell0w:
@dylanb done

dylanb:
Leaving for @WilcoFiers to determine when to merge

0ddfell0w:
@dylanb i actually think this might not work, isFocusable() will return True for the **unfocusable** `<button style="visibility:hidden" tabindex="0">Hidden button forced into the tab order</button>`

isFocusable will have the following logic:
is it natively focusable? No, because it's hidden
is it a button: Yes, therefor it's focusable.

0ddfell0w:
I'll add tests to isFocusable for the disabled/screenreader-invisible cases as part of resolving this issue

0ddfell0w:
@dylanb @WilcoFiers didn't want to force push over approved changes, so here's another PR with the changes I think are correct: https://github.com/dequelabs/axe-core/pull/658

WilcoFiers:
@0ddfell0w Does that mean we can close this? If so, please go ahead.

PR #658, opened 2017-12-18
fix(commons/dom): fix isFocusable functions by checking screenreader visibility and enabledness

### LINKED ISSUE #647: Bug: dom.isFocusable mistakenly concludes that invisible elements are unfocusable

0ddfell0w:
Adds tests for disabled/AT-hidden elements with tabindex

fixes #647


0ddfell0w:
@dylanb nooo problem. @WilcoFiers tagging you here in case you're deciding when to merge, closed the other PR

Issue #660, opened 2017-12-19
make the frame timeout limit configurable


WilcoFiers:
We had a client (issue WH-4054 on JIRA) that kept getting timeout errors while running axe on a particular page. A short term solution could be to extend this time limit. Longer term we should probably just add a feature that lets the frame report its still in progress if it's near the timeout limit. That way we can shorten the timeout too.

dylanb:
How about allowing a configurable timeout?

WilcoFiers:
Matt is working on that :)

isner:
Feature added by https://github.com/dequelabs/axe-core/pull/661

PR #661, opened 2017-12-19
feat(collect-results-from-frames): add frameWaitTime option

isner:
Closes https://github.com/dequelabs/axe-core/issues/660

### DIFF
--- a/test/core/utils/collect-results-from-frames.js
+++ b/test/core/utils/collect-results-from-frames.js
@@ -9,10 +9,10 @@ describe('axe.utils.collectResultsFromFrames', function () {
 		fixture.innerHTML = '';
 	});
 
-  it('should timeout after 30s', function (done) {
+  it('should timeout after 60s', function (done) {
     var orig = window.setTimeout;
     window.setTimeout = function (fn, to) {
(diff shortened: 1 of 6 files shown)


test/core/utils/collect-results-from-frames.js
```diff
@@ -37,6 +37,37 @@ describe('axe.utils.collectResultsFromFrames', function () {
...
+        return orig(fn, to);
+      }
+      return 'cats';
+    };
+
+    var frame = document.createElement('iframe');
+    frame.addEventListener('load', function () {
+      var context = new Context(document);
+      var params = { frameWaitTime: 90000 };
+      axe.utils.collectResultsFromFrames(context, params, 'stuff', 'morestuff', noop,
```
marcysutton: Is there a way to also cover this in an integration test, so we can make sure it works from axe.run?

dylanb: specifically, across iframes

isner: Good idea. I've added an integration test.

marcysutton:
Also merged to develop-2x. Thanks Matt!

PR #666, opened 2017-12-21
fix(core/utils/querySelectorAll): Ensure that  results do not contain duplicate nodes

WilcoFiers:
New issue I found when I had a site with `<main role=“main”>` reported as having two main landmarks.


marcysutton:
How about an integration test with the example included in your issue description?

lib/core/utils/qsa.js
```diff
@@ -44,10 +44,10 @@ function matchSelector (targets, exp, recurse) {
 	targets = Array.isArray(targets) ? targets : [targets];
 	targets.forEach((target) => {
 		if (matchesTag(target.actualNode, exp) &&
-			matchesClasses(target.actualNode, exp) &&
-			matchesAttributes(target.actualNode, exp) &&
-			matchesId(target.actualNode, exp) &&
-			matchesPseudos(target, exp)) {
+				matchesClasses(target.actualNode, exp) &&
```
dylanb: These are refactors that should not be included in a fix commit

lib/core/utils/qsa.js
```diff
@@ -199,14 +199,21 @@ matchExpressions = function  (domTree, expressions, recurse) {
 		var candidates = domTree;
 		exprArr.forEach((exp, index) => {
 			recurse = exp.combinator === '>' ? false : recurse;
-			if ([' ', '>'].indexOf(exp.combinator) === -1) {
+			if ([' ', '>'].includes(exp.combinator) === false) {
```
dylanb: This is a refactor - take this out into its own commit

lib/core/utils/qsa.js
WilcoFiers: Can you explain what the value is in doing things like that? We need a stricter commit policy if that's what you're looking for in it. I was going to squash this PR into a single commit. I've done that for most PRs, except if there were clearly multiple features / fixes in them. I was under the impression that that was the preferred approach.

WilcoFiers:
@dylanb @marcysutton Issues have been resolved. Please update your reviews.

dylanb:
I think we should rather use `refactor:` in the commit message for changes like the change to use `.includes` and `style:` when making changes like the whitespace change as documented here:

https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit

PR #667, opened 2017-12-22
feat(rule): Flag div/p/spans/headings in focus order

### LINKED ISSUE #632: New rule: Flag elements with tabindex >= 0 without semantics that signal interactivity (Best Practice)
# Flag elements inserted into focus order without semantics that signal interactivity.

Checks that all elements without an aria role have a tabindex value of -1. Elements without an interactive role (implicit or explicit) are unexpected in the tab order.

**Tags:** cat.keyboard
**Impact:** minor

## Selector

div, span, p, and heading elements inserted into the focus order

****
### keyboard/focus-order-semantics

Passes if
* element has role of type widget or composite OR
* element has a tag name that implies it might scroll<sup>1</sup> OR
* element has a role that implies it might scroll<sup>2</sup>

Other, generate a violation

<sup>1</sup> Tags for scrollable elements: article, section, nav, aside
<sup>2</sup> Roles for scrollable elements: article, complementary, contentinfo, directory, document, form, main, navigation, region

0ddfell0w:
Flag elements inserted into focus order without semantics that signal
interactivity.

fixes #632


0ddfell0w:
@marcysutton is it worrisome to you that this branch passes tests and is merge-able without the real tests running in CircleCI?

The existence of lib/rules/focus-order-semantics.json seems to prevent the tests from really running

lib/commons/dom/is-focusable.js
```diff
@@ -70,3 +70,19 @@ dom.isNativelyFocusable = function(el) {
...
+ * tabindex were unspecified.
+ * tabindex
+ * @method insertedIntoFocusOrder
+ * @memberof axe.commons.dom
+ * @instance
+ * @param {HTMLElement} el The HTMLElement
+ * @return {Boolean} True if the element is in the focus order but wouldn't be
+ * if its tabindex were removed. Else, false.
+ */
+dom.insertedIntoFocusOrder = function(el) {
```
WilcoFiers: Either this should be in its own file, or the test file you created for it should be merged with the existing tests for is-focusable.js. It's best to have the test files match the name of the files they're testing.

lib/rules/focus-order-semantics.json
```diff
@@ -0,0 +1,13 @@
+{
+  "id": "focus-order-semantics",
```
WilcoFiers: You'll need to add integration tests for this rule. Each rule should have integration tests.

lib/rules/focus-order-semantics.json
```diff
@@ -0,0 +1,13 @@
...
+  "selector": "div, h1, h2, h3, h4, h5, h6, [role=heading], p, span",
+  "matches": "inserted-into-focus-order-matches.js",
+  "tags": ["cat.name-role-value", "cat.keyboard"],
+  "metadata": {
+    "description":
```
WilcoFiers: These should be on a single line.

lib/rules/focus-order-semantics.json
```diff
@@ -0,0 +1,13 @@
+{
+  "id": "focus-order-semantics",
+  "selector": "div, h1, h2, h3, h4, h5, h6, [role=heading], p, span",
+  "matches": "inserted-into-focus-order-matches.js",
+  "tags": ["cat.name-role-value", "cat.keyboard"],
```
WilcoFiers: This should only have one category. I suggest `cat.keyboard`. Additionally, please add `best-practice` and `experimental` as well.

lib/rules/focus-order-semantics.json
```diff
@@ -0,0 +1,13 @@
...
+  "metadata": {
+    "description":
+      "Ensures elements placed in the focus order have an appropriate aria role for interactive content",
+    "help":
+      "Elements placed in the focus order must have an aria role appropriate for interactive content."
+  },
+  "any": ["valid-focus-order-widget", "valid-scrollable-semantics"]
```
WilcoFiers: `all` and `none` are required properties.  Please add them as empty arrays.

lib/checks/keyboard/inserted-into-focus-order.js
```diff
@@ -0,0 +1,2 @@
+return (axe.commons.dom.isFocusable(node) && node.tabIndex > -1 &&
```
WilcoFiers: Why does this file exist? It's not a check you're using anywhere. I'm guessing you started here and then moved it into a commons?

lib/checks/aria/valid-focus-order-widget.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * @param {HTMLElement} node
+ * @return {Boolean} Whether the element's role is appropriate for an element
+ *     in the focus order that isn't a scrollable region.
+ */
+function validFocusOrderWidget(node) {
```
WilcoFiers: You don't need a separate function for this. The content of this file is wrapped in a function that will execute each iteration. This results in this function getting redeclared over and over.  You can just use the function body as the check code.

lib/checks/aria/valid-focus-order-widget.js
```diff
@@ -0,0 +1,15 @@
+/**
```
WilcoFiers: We'll need some tests for this check.

lib/checks/aria/valid-focus-order-widget.json
```diff
@@ -0,0 +1,12 @@
+{
+  "id": "valid-focus-order-widget",
```
WilcoFiers: This check doesn't do anything with focus order. I think this should be renamed to has-widget-role or something like that. The pass / fail suggestions would also make more sense if we just talk about the element having / not having a widget role.

lib/checks/aria/valid-scrollable-semantics.json
```diff
@@ -0,0 +1,12 @@
+{
+  "id": "valid-scrollable-semantics",
```
WilcoFiers: I don't understand why this is a separate check. It seems to me this is overcomplicating things.

lib/rules/focus-order-semantics.json
```diff
@@ -0,0 +1,13 @@
+{
+  "id": "focus-order-semantics",
+  "selector": "div, h1, h2, h3, h4, h5, h6, [role=heading], p, span",
```
WilcoFiers: I don't quite understand what `[role=heading]` is doing in here. What error are you expecting to catch?

lib/checks/aria/valid-scrollable-semantics.js
```diff
@@ -0,0 +1,57 @@
...
+  region: true,
+  search: false
+};
+
+/**
+ * @param {HTMLElement} node
+ * @return {Boolean} Whether the element has a tag appropriate for a scrollable
+ *     region.
+ */
+function validScrollableTagName(node) {
```
WilcoFiers: I don't understand the purpose of this. The only part of the selector that lets these elements in is if they match `[role=heading]`. I don't think `<nav role="heading" tabindex="1">` should pass this rule, and I don't see any other path leading to this function passing anything either.

lib/checks/keyboard/inserted-into-focus-order.js
0ddfell0w: Yeah I was honestly moving all kinds of things around trying to figure out why tests wouldn't run, turns out it was because I had left the "all" and "none" fields out of the rule definition -- thanks so much for catching that -- i'll clean this up and resolve your comments and get it back to you

lib/rules/focus-order-semantics.json
0ddfell0w: 🙌thank you for catching this, I couldn't figure out why tests wouldn't run

lib/checks/aria/valid-scrollable-semantics.js
0ddfell0w: The example nav[role=heading][tabindex=1] should presumably fail a different test, since navs should not be allowed to have role=heading (or any other ARIA role).

My intuition is to make this check available for use in future rules, rules which may not have the same selector as our focus-order-semantics rule.

I'd argue that this check should stand on its own feet, and should PASS elements that have one of these tags as plausibly being scrollable regions. This seems more maintainable and safer with respect to false positives, but I happily defer to your experience with axe. What do you think?

lib/commons/dom/is-focusable.js
0ddfell0w: Sure, the same logic should apply to [dom.isNativelyFocusable](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/dom/is-focusable.js#L48), right?

lib/rules/focus-order-semantics.json
0ddfell0w: e.g. em[role=heading][tabindex=1]

lib/checks/aria/valid-scrollable-semantics.js
WilcoFiers: I'm fine with that. Just making sure I understand what's going on. Can you add a comment in to explain this?

lib/checks/aria/valid-scrollable-semantics.json
0ddfell0w: i may not understand the suggestion.

i'd make the same argument with respect to checks being available for use in future rules. It seems nice to have has-widget-role as a check, and this is the other case for which we would PASS elements as having valid semantics for being in the focus order.

How else would we want to express that, other than having this check?

lib/rules/focus-order-semantics.json
0ddfell0w: Done.

lib/rules/focus-order-semantics.json
0ddfell0w: Done.

lib/checks/aria/valid-scrollable-semantics.js
0ddfell0w: Done.

lib/checks/aria/valid-focus-order-widget.json
0ddfell0w: Done.

lib/checks/aria/valid-focus-order-widget.js
0ddfell0w: Done.

test/checks/aria/has-widget-role.js
```diff
@@ -0,0 +1,460 @@
+describe('has-widget-role', function () {
```
0ddfell0w: These roles are marked up as widgets in lib/commons/aria/index.js, but not in the aria spec

The ARIA spec seems to differ with axe on these:

## tooltip
https://github.com/dequelabs/axe-core/blame/develop/lib/commons/aria/index.js#L874
https://www.w3.org/TR/wai-aria-1.1/#tooltip

# Window roles
## dialog
https://github.com/dequelabs/axe-core/blame/develop/lib/commons/aria/index.js#L319
https://www.w3.org/TR/wai-aria-1.1/#dialog

## alertdialog
https://github.com/dequelabs/axe-core/blame/develop/lib/commons/aria/index.js#L191
https://www.w3.org/TR/wai-aria-1.1/#alertdialog

# Live region roles
## timer
https://github.com/dequelabs/axe-core/blame/develop/lib/commons/aria/index.js#L855
https://www.w3.org/TR/wai-aria-1.1/#timer

## status
https://github.com/dequelabs/axe-core/blame/develop/lib/commons/aria/index.js#L764
https://www.w3.org/TR/wai-aria-1.1/#status

## marquee
https://github.com/dequelabs/axe-core/blame/develop/lib/commons/aria/index.js#L494
https://www.w3.org/TR/wai-aria-1.1/#marquee

## log
https://github.com/dequelabs/axe-core/blame/develop/lib/commons/aria/index.js#L475
https://www.w3.org/TR/wai-aria-1.1/#log

## alert
https://github.com/dequelabs/axe-core/blame/develop/lib/commons/aria/index.js#L182
https://www.w3.org/TR/wai-aria-1.1/#alert

lib/checks/aria/valid-scrollable-semantics.js
0ddfell0w: let me know if the comment I left wasn't what you were suggesting

0ddfell0w:
@WilcoFiers should be ready for you to review again

lib/commons/dom/is-focusable.js
WilcoFiers: Yep

test/commons/dom/is-focusable.js
```diff
@@ -1,146 +1,657 @@
...
+		it('should return true for clipped span with tabindex 0', function () {
+			fixtureSetup('<span id="clippedSpanTabindex0" tabindex="0"></span>');
+			var node = fixture.querySelector('#clippedSpanTabindex0');
+			hideByClipping(node);
+
+			assert.isTrue(axe.commons.dom.insertedIntoFocusOrder(node));
+
+		});
+
+		it('should return true for clipped div with tabindex 0', function () {
```
WilcoFiers: You've added quite a few redundant tests in here. There's no obvious reason why this method would behave different for a `div` or `h4` than it does for a `span`. If you want to keep them, I suggest you combine them. You can have multiple asserts in the same statement. Leaving these tests here implies complexity that isn't there, which can confuse future devs.

test/checks/aria/has-widget-role.js
WilcoFiers: Can you create a new ticket for this?

test/checks/aria/has-widget-role.js
```diff
@@ -0,0 +1,452 @@
...
+	var node;
+	var checkContext = axe.testUtils.MockCheckContext();
+
+
+	afterEach(function () {
+		node.innerHTML = '';
+		checkContext._data = null;
+	});
+
+	it('should return false for elements with no role', function() {
```
WilcoFiers: For future reference, it's better if you use unit tests to check if the logic of your code. Most of these tests just check that the ARIA lookup table is complete. It's easier and better to do that in the implementation tests. You can leave it as is, either way it's tested, but as a rule of thumb it's a good one to keep in mind.

lib/commons/dom/is-focusable.js
```diff
@@ -70,3 +70,18 @@ dom.isNativelyFocusable = function(el) {
...
+ * tabindex were unspecified.
+ * @method insertedIntoFocusOrder
+ * @memberof axe.commons.dom
+ * @instance
+ * @param {HTMLElement} el The HTMLElement
+ * @return {Boolean} True if the element is in the focus order but wouldn't be
+ * if its tabindex were removed. Else, false.
+ */
+dom.insertedIntoFocusOrder = function(el) {
+	return (dom.isFocusable(el) && el.tabIndex > -1 &&
```
WilcoFiers: Please put the `el.tabIndex > -1` in the first part of the expression. The matcher will run on every node on the page, so we should put the fastest part first.

lib/checks/aria/valid-scrollable-semantics.json
```diff
@@ -0,0 +1,12 @@
...
+  "evaluate": "valid-scrollable-semantics.js",
+  "options": [],
+  "metadata": {
+    "impact": "minor",
+    "messages": {
```
WilcoFiers: I think it might be confusing to make this check about scrollable regions. The element it pulls up may not be scrollable at all. I'm thinking we do something like "Element does not have a landmark role appropriate for a focusable element". What do you think?

test/commons/dom/is-focusable.js
0ddfell0w: Removed all but the span tests

lib/checks/aria/valid-scrollable-semantics.json
0ddfell0w: I agree the existing message isn't great, but saying "does not have a landmark role..." doesn't capture the tag names also being OK, and we would probably want to say "in the focus order" or "in the tab order" instead of "focusable" -- since "focusable" elements with tabindex=-1 aren't a concern.

Happy to give you final say on the message, leaving for now but let me know if you have a replacement in mind

lib/commons/dom/is-focusable.js
0ddfell0w: Done.

lib/rules/focus-order-semantics.json
0ddfell0w: Done.

test/checks/aria/has-widget-role.js
0ddfell0w: i agree. these tests were useful while adding this feature, don't think they need to stay. leaving for now pending the discussion of https://github.com/dequelabs/axe-core/issues/672

0ddfell0w:
@WilcoFiers this is ready at your convenience

lib/rules/focus-order-semantics.json
```diff
@@ -0,0 +1,13 @@
...
+  "matches": "inserted-into-focus-order-matches.js",
+  "tags": ["cat.keyboard", "best-practice", "experimental"],
+  "metadata": {
+    "description": "Ensures elements placed in the focus order have an appropriate aria role for interactive content",
+    "help": "Elements placed in the focus order must have an aria role appropriate for interactive content."
```
WilcoFiers: We need to shorten this so that it fits in the Axe extension properly. I suggest: "Focusable elements have an appropriate role" (no "." at the end either please).

lib/checks/aria/has-widget-role.js
```diff
@@ -0,0 +1,6 @@
+var role = node.getAttribute('role');
```
WilcoFiers: This check should have tests. Nothing too complicated, just check that it works.

0ddfell0w: there are tests in this PR at test/checks/aria/has-widget-role.js

are there other things you're looking to test?

lib/rules/focus-order-semantics.json
0ddfell0w: Shortened, though I'm still trying to avoid equating focusable with in-focus-order so I didn't change it quite as you recommend. It's shorter than some other descriptions at least (e.g [link-in-text-block](https://github.com/dequelabs/axe-core/blob/develop/lib/rules/link-in-text-block.json#L141))

0ddfell0w:
@WilcoFiers ready again, though I might me misunderstanding what you're asking for with respect to tests

lib/checks/aria/has-widget-role.js
WilcoFiers: Wow, I'm not sure how I missed that. My bad!

lib/checks/aria/valid-scrollable-semantics.json
isner: I think you're both onto something here. I vote for a compromise that refers to the element's focusability instead of scrollability (I think this will be more easily apprehended by future readers), but does not mention role to the exclusion of tagname. How about, "Element has valid semantics for a focusable element"? 

This suggestion extends to the names of the map constants (`VALID_{TAG_NAMES|ROLES}_FOR_SCROLLABLE_REGIONS`) and the name of the check itself. Using language that refers to focus instead of scrolling will resonate more with the accessibility community, in my opinion, and will make things easier for everyone to follow.

isner: @0ddfell0w You make a good point about the term "focusable". I don't think it's a given that readers would fail to see the implicit distinction between _naturally_ and _programmatically_ focusable elements, but they might.

Maybe "Element has valid semantics for an element in the focus order"?

WilcoFiers: I agree. Let's go with that. I think that's the last change we need for this PR.

0ddfell0w: @WilcoFiers @isner done

WilcoFiers:
Whooho! Congratulations!

Issue #644, opened 2017-12-13
Ensure rule/check options can be configured to override default selectors


marcysutton:
We've made a change to `axe.commons.aria.lookupTable` to make the API appear less private, as compared to `axe.commons.aria._lut`. However, rather than creating additional APIs for configuring rules, we should better utilize a tool we already have: specifying options.

axe-core definitely allows you to specify options in checks. Options are also used in the color contrast rule, but we should do some investigating to see if this is actually working. This issue will be satisfied when options can be used to add additional attributes/selectors for existing rules.

The lookupTable could still be used to retrieve defaults, if needed, but in terms of configuring rules, options are the way to go. We should make sure this is easy to do for existing rules with `axe.configure`, not just custom ones.

marcysutton:
Closed with #673.

marcysutton:
An implementation note: this fix does not extend to all rules/checks, but the ones specifically requested (`aria-allowed-attr`, `aria-required-attr`). A broader, long-term solution would be to implement an Accessibility Supported API, but that's a huge undertaking and isn't currently prioritized.

PR #673, opened 2018-01-05
Support options in aria-allowed-attr and aria-required-attr

marcysutton:
This continues the work to avoid having to manipulate the lookupTable directly. `valid-attr` and `valid-attr-value` already used options; this PR adds options to `aria-allowed-attr` and `aria-required-attr`. The usage for these checks is now to plumb in an options object with a role and an array of accepted attribute values:

```
axe.configure({
    checks: [{
        id: 'aria-allowed-attr',
        options: {
            separator: ['aria-valuenow', 'aria-valuemin', 'aria-valuemax']
        }
    }]
});
axe.run(document,  {}, function(error, results) {
    assert.equal(results.violations.length, 0);
    done();
});
```

Closes https://github.com/dequelabs/axe-core/issues/644

cc @waabid @iamrafan


lib/checks/aria/allowed-attr.js
```diff
@@ -12,7 +14,7 @@ if (role && allowed) {
 	for (var i = 0, l = attrs.length; i < l; i++) {
 		attr = attrs[i];
 		attrName = attr.name;
-		if (axe.commons.aria.validateAttr(attrName) && allowed.indexOf(attrName) === -1) {
+		if (options.indexOf(attrName) === -1 && axe.commons.aria.validateAttr(attrName) && allowed.indexOf(attrName) === -1) {
```
WilcoFiers: You should wrap this line. You may also want to use `.includes`.

marcysutton:
I incorporated the small piece of feedback and also added options to `required-attr`, as well as some integration tests for axe.configure. There are other checks that (indirectly) reference the lookupTable, but they're quite complex (like `required-children`) so unless there's a specific need for those, I'll keep the scope of this PR small.

test/checks/aria/required-attr.js
```diff
@@ -57,4 +57,12 @@ describe('aria-required-attr', function () {
 		axe.commons.aria.requiredAttr = orig;
 	});
 
+	describe('options', function () {
+		it('should require provided attribute names', function () {
+			fixture.innerHTML = '<div role="slider" id="target"></div>';
```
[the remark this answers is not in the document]
marcysutton: Love it! will do.

lib/checks/aria/required-attr.js
```diff
@@ -1,10 +1,21 @@
+options = Array.isArray(options) ? options : [];
```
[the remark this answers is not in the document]
marcysutton: Sounds good, thanks for the tip!

marcysutton:
> The one thing I'm hesitant about with this is that this is starting to look a lot like accessibility supported functionality. We haven't got an API design for this yet, and I'm hoping we won't have to break this once settle on a way to do this.

We can make changes when we're ready, but these options will help support our users now. Options are already used in the codebase in a few places so there's precedent. But perhaps we can hold off from a big documentation push until we decide on an API.

marcysutton:
I've updated the PR so the checks now take an object as options:
```
{ 'separator': ['aria-valuemin', 'aria-valuemax', 'aria-valuenow']}
```

lib/checks/aria/allowed-attr.js
```diff
@@ -7,12 +9,22 @@ var attr, attrName, allowed,
 if (!role) {
 	role = axe.commons.aria.implicitRole(node);
 }
+
 allowed = axe.commons.aria.allowedAttr(role);
+
+if (Object.keys(options).length) {
```
WilcoFiers: You can simplify this.
```js
if (Array.isArray(options[role])) {
  allowed = axe.utils.uniqueArray(options[role].concat(allowed));
}
```

marcysutton: Done. I also removed a duplicate `to-array` file in axe.commons. It was almost exactly the same code duplicated, we don't want to maintain both if we don't have to.

marcysutton: Oops, I was confused by this comment...I've made the change and added it to this PR. Thanks!

test/integration/full/configure-options/configure-options.js
```diff
@@ -0,0 +1,50 @@
+describe('Check Configure Options', function() {
+	'use strict';
+
+	var target = document.querySelector('#target');
+
+	describe('aria-allowed-attr', function() {
+		it('should allow an attribute supplied in options', function(done) {
+			target.setAttribute('role', 'separator');
+			target.setAttribute('aria-valuenow', '0');
```
isner: In the event that 'aria-valuenow' becomes an allowed attribute in our lookup table, this test won't be testing anything anymore. Just to make the test more future-proof, I suggest using a random value like 'peanutbutter' instead of a real attribute so we can guarantee the test will always be testing the override functionality with a truly unexpected value.

lib/core/utils/to-array.js
```diff
@@ -7,4 +7,17 @@
...
+};
+
+
+/**
+ * Creates an array without duplicate values from 2 array inputs
+ * @param  {Array} arr1 First array
+ * @param  {Array} arr2 Second array
+ * @return {Array}
+ */
+axe.utils.uniqueArray = (arr1, arr2) => {
```
isner: This is super minor, but it'd be nice if this iterated over an arbitrary number of array inputs.

PR #675, opened 2018-01-06
docs: Add activity and usage badges

dylanb:
Also removed some badges that do not add a lot of value - probably most controversial is the CircleCI build status badge. Reason I don't think this adds value is that we never merge a change that does not pass CircleCI, so the most important build statuses are the build status of the individual pull requests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,11 @@
 # axe-core
 
-[![Join the chat at https://gitter.im/dequelabs/axe-core](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/dequelabs/axe-core?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
-[![Version](https://img.shields.io/npm/v/axe-core.svg)](https://www.npmjs.com/package/axe-core)
-[![Bower](https://img.shields.io/bower/v/axe-core.svg)](http://bower.io/search)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,11 +1,11 @@
 # axe-core
 
-[![Join the chat at https://gitter.im/dequelabs/axe-core](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/dequelabs/axe-core?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
-[![Version](https://img.shields.io/npm/v/axe-core.svg)](https://www.npmjs.com/package/axe-core)
-[![Bower](https://img.shields.io/bower/v/axe-core.svg)](http://bower.io/search)
 [![License](https://img.shields.io/npm/l/axe-core.svg)](LICENSE)
-[![CircleCI Build](https://circleci.com/gh/dequelabs/axe-core/tree/develop.svg?style=svg)](https://circleci.com/gh/dequelabs/axe-core/tree/develop)
```
marcysutton: No circle build anymore?

WilcoFiers: Maybe because no code was change? Just guessing here.

dylanb: @marcysutton see my comment above - if you think we should keep the build status, what is the rationale?

marcysutton:
We need to merge this to develop-2x and master so it will show up on npm before we release axe-core 3.0.

Issue #20, opened 2015-06-16
Add support for SVG images


dylanb:
We do not handle SVG images correctly at the moment, need to be able to calculate the accessible name correctly for inline SVG and report when not present.

dsturley:
I remember looking into this a while ago; there did not seem to be a consensus as to the proper or recommended way to add alt-text to inline SVG images.

dsturley:
@dylanb Can you define some requirements for this?  Browser support is flaky and I didn't see a definite answer from the specifications.

dylanb:
deferring until better browser support

dylanb:
Looking for help to document what portions of the accessible name calculation work properly for svg images (both inline and referenced via a `src` attribute on an `<img>` element.

Platforms that need to be supported are:

1. NVDA and FF on Windows (7 or 10)
1. JAWS and IE on Windows (7 or 10)
1. Safari VO on OS X latest
1. Safari and VO on iOS latest
1. Firefox and Talkback on Android
1. Chrome and Talkback on Android

StommePoes:
http://www.stommepoes.nl/work/firefox_svg.html This is a bit special because it's using a sprite and they're not nested. Would be great if mobile users could test these (I don't own an Android nor an iThing).

dylanb:
@StommePoes could you add a failing example to that page? I.e. an SVG image without any alt.

StommePoes:
Yes.

dylanb:
This page has interesting data about support http://simplyaccessible.com/article/7-solutions-svgs/

StommePoes:
Re that link ^ for Windows High Contrast, we've been avoiding inline fill colours and instead have set it that the SVGs fill in CSS equals "currentColor" and that it must have a parent wrapped around it with a set color: someAccessibleColor. When our text colours change on WHC settings, the SVG should follow suit.

dylanb:
@StommePoes example pretty please...

StommePoes:
Of using currentColor? Sure.

goodwitch:
Southwest Airlines uses svg for images sometimes.  See example in footer of the Southwest Cargo logo/link on www.southwest.com of /assets/images/globalnav/logos/swa_footer_logo_swa_cargo.svg

dylanb:
@goodwitch That example clearly shows how SVG is currently not supported.

[...]

PR #683, opened 2018-01-17
fix(rule): Remove coverage of `svg[role="img"]` for image-alt rule

codingmatty:
We have been advised that best practice is to use the following for inline SVGs:
```
<svg role="img">
  <title>SVG Title</title>
</svg>
```
which is echoed in [this article](https://css-tricks.com/accessible-svgs/#article-header-id-6).

I see that there is conversation in this issue https://github.com/dequelabs/axe-core/issues/20 about how to best go about it, but for now the test for `svg[role="img"]` needs to be removed since it isn't really valid.

### DIFF
--- a/test/integration/rules/image-alt/image-alt.html
+++ b/test/integration/rules/image-alt/image-alt.html
@@ -16,3 +16,4 @@
 <img src="img.jpg" id="pass8" aria-labelledby="ninjamonkeys">
 <div role="img" alt="blah" id="violation6"></div>
 <div role="img" aria-label="blah" id="pass9"></div>
+<svg role="img" id="ignore1"><title>SVG Title</title></svg>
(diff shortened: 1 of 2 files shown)


test/integration/rules/image-alt/image-alt.html
```diff
@@ -16,3 +16,4 @@
 <img src="img.jpg" id="pass8" aria-labelledby="ninjamonkeys">
 <div role="img" alt="blah" id="violation6"></div>
 <div role="img" aria-label="blah" id="pass9"></div>
+<svg rol="img" id="ignore1"><title>SVG Title</title></svg>
```
dylanb: There is a typo in the attribute

WilcoFiers: Please fix the typo. I think you're right about excluding svg from this rule. We should have a separate SVG rule instead.

Issue #681, opened 2018-01-14
Extend 'landmark-main-is-top-level' rule to fail if banner, complementary or contentinfo are not top-level landmarks


iamrafan:
Ensures that main, banner, complementary and contentinfo are top-level landmarks.

WilcoFiers:
I think those should be separate rules.

iamrafan:
Since all these landmarks should be top level landmarks, I think they can be part of the same rule.

WilcoFiers:
They could have been, but if we do we'd have to rename the rule. I'd rather not. Additionally, we can't do the same with the landmark-one-main rule, so combining them in one and not in the other would introduce an inconsistency.

iamrafan:
Sure

WilcoFiers:
This rule is being worked on in https://github.com/dequelabs/axe-core/pull/685

iamrafan:
@WilcoFiers thanks for the update. 
The only check it is missing is making sure ```complementary``` landmarks are top level.

jongund:
We can make a rule for complementary being top level,

WilcoFiers:
This issue is resolved as part of #761

iamrafan:
Thanks @jongund and @WilcoFiers

iamrafan:
@WilcoFiers #761 does not contain the new rule for ensuring complementary landmarks are top level, do you want me to create a new issue for this?

marcysutton:
@iamrafan a new issue sounds good. Thanks!

PR #685, opened 2018-01-17
Rules ensuring banner and contentinfo landmarks are top level

ssanaul:


dylanb:
@ssanaul thank you for the contribution - can you describe the logic you are attempting to implement and can you provide arguments or evidence supporting the assertion that this is a best practice?

lib/checks/keyboard/banner-is-top-level.js
```diff
@@ -0,0 +1,19 @@
+const landmarks = axe.commons.aria.getRolesByType('landmark');
```
WilcoFiers: This is basically a repeat of `main-is-top-level.js`. I suggest having these 3 checks point to the same evaluate file. You can change what element they test through the rule selector as well as the check.json options property. So you would have `main-is-top-level.json`, `banner-is-top-level.json` and `contentinfo-is-top-level.json` each point to `landmark-is-top-level.js`, and pass in a different `options` object which tells it what element you're trying to test.

jongund:
Dylan,

This rule is implementing requirements from the ARIA Authoring Practices 1.1:
- https://www.w3.org/TR/wai-aria-practices-1.1/#aria_lh_banner
- https://www.w3.org/TR/wai-aria-practices-1.1/#aria_lh_contentinfo

lib/checks/keyboard/banner-is-top-level.js
jongund: Not all landmarks need to be top level, only main, banner and contentinfo should be top level.  In many cases like region landmarks, they will identify subsections of "named" landmarks.  Some assistive technologies support navigation of the parent child relationships of landmarks.

See ARIA Authoring Practices 1.1 for landmarks:
https://www.w3.org/TR/wai-aria-practices-1.1/#aria_landmark

jongund: But what you seem to be saying is just have one rule, that says banner, contentinfo and main landmarks must be top level, so we will look at combining the three rules into one rule.

jongund: My main concern is that the messaging might not be as clear.  Since people may think they can only have one MAIN, CONTENTINFO or BANNER, when they can have one of each or they can all be top level

jongund: My main concern is messaging in a more general rule will not be as clear, since the message will talk about all three landmarks, and some pages may one have a MAIN landmark

marcysutton: I want to make sure you've seen #681 as part of this conversation

WilcoFiers: @jongund No no, please don't merge the rules. I think three rules makes the most sense. What I'm saying is that these three checks should have a shared `evaluate` function. Have a look here:
https://github.com/dequelabs/axe-core/tree/develop/lib/checks/mobile

Here are two checks, `meta-viewport` and `meta-viewport-large` that are very similar. Instead of each of these having it's own `.js` file, they both use `meta-viewport-scale.js`. This way we don't have two versions of this `.js` file that do essentially the same thing.

jongund: @WilcoFiers OK  We can do that, I don't think we need the options property set, since we can use the tag information from the rules json file to determine which landmark is being tested

WilcoFiers: As far as I know the check doesn't have access to what rule is running. But if you can find a good way to do it without options that's fine by me. :)

lib/checks/keyboard/landmark-is-top-level.js
```diff
@@ -0,0 +1,20 @@
...
+const nodeIsHeader = node.tagName.toLowerCase() === 'header' && node.getAttribute('role') !== 'banner';
+const nodeIsFooter = node.tagName.toLowerCase() === 'footer' && node.getAttribute('role') !== 'contentinfo';
+var parent = axe.commons.dom.getComposedParent(node);
+
+while (parent){
+	var role = parent.getAttribute('role');
+	if (!role && (parent.tagName.toLowerCase() !== 'form')){
+		role = axe.commons.aria.implicitRole(parent);
+	}
+  if (role && (nodeIsHeader || nodeIsFooter) && sectioning.includes(role)){
```
marcysutton: The indentation here is a little funky. This is also super nitpicky, but we use a space before the opening curly brace in function declarations and conditional statements: `function() {` and `if (conditional) {`

lib/checks/keyboard/landmark-is-top-level.js
```diff
@@ -0,0 +1,20 @@
+const landmarks = axe.commons.aria.getRolesByType('landmark');
+const sectioning = ['article', 'aside', 'main', 'navigation', 'section'];
+const nodeIsHeader = node.tagName.toLowerCase() === 'header' && node.getAttribute('role') !== 'banner';
+const nodeIsFooter = node.tagName.toLowerCase() === 'footer' && node.getAttribute('role') !== 'contentinfo';
+var parent = axe.commons.dom.getComposedParent(node);
+
+while (parent){
+	var role = parent.getAttribute('role');
```
marcysutton: Naming this `parentRole` would make it more clear what's being evaluated, since we just looked at the node's role above.

test/checks/keyboard/banner-is-top-level.js
```diff
@@ -0,0 +1,89 @@
...
+	it('should return false if banner landmark is in main element', function() {
+		var main = document.createElement('main');
+		var banner = document.createElement('div');
+		banner.setAttribute('role','banner');
+		main.appendChild(banner);
+		fixture.appendChild(main);
+		assert.isFalse(checks['banner-is-top-level'].evaluate(banner));
+	});
+
+	it('should return false if banner landmark is in main element', function () {
```
marcysutton: This test is an exact duplicate of the one above it. (I've done that before when copy/pasting tests to create new ones. oops)

test/checks/keyboard/banner-is-top-level.js
```diff
@@ -0,0 +1,89 @@
...
+		var main = document.createElement('div');
+		main.setAttribute('role','main');
+		var banner = document.createElement('div');
+		banner.setAttribute('role','banner');
+		main.appendChild(banner);
+		fixture.appendChild(main);
+		assert.isFalse(checks['banner-is-top-level'].evaluate(banner));
+	});
+
+	it('should return false if header is not sectioning element and in div with role search', function () {
```
marcysutton: This test name is confusing. How about something like "it should return false if header is inside of div with role search"? It doesn't seem important that the header isn't a sectioning element, more that the header isn't *inside* of a sectioning element so it's not a banner anymore (and it should be top level). But we don't have to go into that level of detail in a test name.

test/checks/keyboard/banner-is-top-level.js
```diff
@@ -0,0 +1,89 @@
...
+		fixture.appendChild(article);
+		assert.isTrue(checks['banner-is-top-level'].evaluate(header));
+	});
+
+	(shadowSupported ? it : xit)('should test if banner in shadow DOM is top level', function () {
+		var div = document.createElement('div');
+		var shadow = div.attachShadow({ mode: 'open' });
+		shadow.innerHTML = '<div role="banner">Banner landmark</div>';
+		var checkArgs = checkSetup(shadow.querySelector('[role=banner], header'));
+		assert.isTrue(checks['banner-is-top-level'].evaluate.apply(null, checkArgs));
```
marcysutton: My hunch is that the failing case might be a better test of Shadow DOM, to ensure it's actually testing across the boundary. You could put `role="banner"` inside of Shadow DOM just like you have it, but add a violating landmark in the light DOM and check that it `isFalse`.

test/checks/keyboard/complementary-is-top-level.js
```diff
@@ -0,0 +1,59 @@
...
+		fixture.appendChild(complementary);
+		assert.isTrue(checks['complementary-is-top-level'].evaluate(complementary));
+	});
+
+	(shadowSupported ? it : xit)('should test if complementary in shadow DOM is top level', function () {
+		var div = document.createElement('div');
+		var shadow = div.attachShadow({ mode: 'open' });
+		shadow.innerHTML = '<div role="complementary">complementary content</div>';
+		var checkArgs = checkSetup(shadow.querySelector('[role=complementary]'));
+		assert.isTrue(checks['complementary-is-top-level'].evaluate.apply(null, checkArgs));
```
marcysutton: Same thing as in test/checks/keyboard/banner-is-top-level.js: it would be good to test the failing case to make sure we're evaluating nodes across the computed tree.

test/integration/full/landmark-banner-is-top-level/frames/level1.html
```diff
@@ -0,0 +1,27 @@
+<!doctype html>
+<html id="pass2">
+<head>
+	<meta charset="utf8">
+	<script src="/axe.js"></script>
+</head>
+<body>
+	<p>This iframe should pass, too</p>
+
+	<div role="navigation">
```
marcysutton: This markup is repeated more than once. Is there any way to reuse some of these integration test HTML files? Maybe with conditional JS?

test/integration/full/landmark-complementary-is-top-level/landmark-complementary-is-top-level-pass.js
```diff
@@ -0,0 +1,35 @@
+
+describe('landmark-complementary-is-top-level test pass', function () {
+	'use strict';
+	var results;
+	before(function (done) {
+		window.addEventListener('load', function () {
+			axe.run({ runOnly: { type: 'rule', values: ['landmark-complementary-is-top-level'] } }, function (err, r) {
```
marcysutton: I bet we could cut down on the number of test files for these rules by making them more dynamic. Could you iterate over a list of relevant landmark rules with JavaScript and run the tests that way? It would be less boilerplate and less code to maintain. There's some precedent in the rule runner, although it's more complicated than you would probably need: https://github.com/dequelabs/axe-core/blob/develop/test/integration/rules/runner.js#L101

In the HTML fixtures, if you needed different markup regions you could potentially just put them into different named DIVs and use include/exclude in the axe.run config.

lib/rules/landmark-banner-is-top-level.json
```diff
@@ -0,0 +1,16 @@
+{
+  "id": "landmark-banner-is-top-level",
+  "selector": "[role=banner], header",
+  "tags": [
+    "best-practice"
+  ],
+  "metadata": {
+    "description": "A banner landmark identifies site-oriented content at the beginning of each page within a website",
```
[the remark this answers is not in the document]
jongund: We have updated the message to:

"A header element (i.e. context body element) or [role=banner] identifies site-oriented content at the beginning of each page within a website"

lib/rules/landmark-contentinfo-is-top-level.json
```diff
@@ -0,0 +1,16 @@
+{
+  "id": "landmark-contentinfo-is-top-level",
+  "selector": "[role=contentinfo], footer",
+  "tags": [
+    "best-practice"
+  ],
+  "metadata": {
+    "description": "A contentinfo landmark is a way to identify common information at the bottom of each page within a website",
```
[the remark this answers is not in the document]
jongund: Updated to use the same pattern as header message

lib/checks/keyboard/landmark-is-top-level.js
jongund: Update the code to remove extra spacing

lib/checks/keyboard/landmark-is-top-level.js
jongund: Updated the code to use parentNode

test/checks/keyboard/banner-is-top-level.js
jongund: Updated the test name

test/checks/keyboard/banner-is-top-level.js
jongund: Tried creating a shadow DOM test case, but the test case fails the test for a banner landmark in a shadow DOM node that is a descendant of a search landmark in the light DOM.

Can you look at our test case code?

ssanaul:
```
	(shadowSupported ? it : xit)('should test if banner in shadow DOM is top level', function () {
		var div = document.createElement('div');
		div.setAttribute('role','search');
		var shadow = div.attachShadow({ mode: 'open' });
		shadow.innerHTML = '<div role="banner">Banner landmark</div>';
		var checkArgs = checkSetup(shadow.querySelector('[role=banner], header'));
		assert.isFalse(checks['banner-is-top-level'].evaluate.apply(null, checkArgs));
	});
```

test/integration/full/landmark-complementary-is-top-level/landmark-complementary-is-top-level-pass.js
jongund: I would rather work on some more rules at this point, it seems that we could come back to this issue  when we have more time

test/checks/keyboard/banner-is-top-level.js
marcysutton: I've made some changes to the testUtils for Shadow DOM, I'll push my code today so you can take a look at how that's working now.

jongund: We tried the test again and the above test case is still is not failing.   Is there any changes we need to make in the test code above?

marcysutton: You need to use the Shadow DOM-specific testUtil `shadowCheckSetup` instead of `checkSetup`. It allows you to focus on the test markup instead of the boilerplate setup like `axe._tree`, which is missing from these tests. If you rebase this branch based on `origin/develop`, you'll have the latest changes I made.

```
var shadowCheckSetup = axe.testUtils.shadowCheckSetup;

// and in your test:
var checkArgs = shadowCheckSetup(
    '<div role="search"></div>',
    '<div role="banner" id="target">Banner landmark</div>'
);
```
In the newest testUtil API, you can pass HTML strings or nodes for the light DOM and Shadow DOM content, with `id="target"` on either of them depending on what you want to test. I find passing strings to be the easiest because you don't have to do any DOM manipulation yourself.

If you were needing to assert the data returned, you'd also need to use `var checkContext = axe.testUtils.MockCheckContext()`. But the test works with the above shadow test setup method.

ssanaul: Yep, that worked!

WilcoFiers:
Closing this PR, since it is superseded by #761

PR #701, opened 2018-01-29
perf(perf-timer): add instrumentation for Rule.runChecks

paulirish:
When running things with perfTimer, I noticed sometimes there was significant cost within the runChecks methods, however that wasn't always obvious from the existing performanceTimer measures.

This adds instrumentation around the `runChecks` methods.  

One note is that the intrumentation is synchronous, however checks finish asynchronously. But in my experience, it appears the majority of the cost happens in the first bit of synchronous runChecks work.  So this seems OK.

Screenshot including the nontrivial cost of `runchecks-duplicate-id`:

image

~[This patch doesn't pass JSHint.](https://user-images.githubusercontent.com/39191/35527867-f7fd761a-04e0-11e8-82ce-54a5b4935d24.png) I'm interested in your suggestions on handling that.~

### DIFF
--- a/lib/core/base/rule.js
+++ b/lib/core/base/rule.js
@@ -127,8 +127,11 @@ Rule.prototype.runChecks = function (type, node, options, resolve, reject) {
  * @param  {Function} callback Function to call when evaluate is complete; receives a RuleResult instance
  */
 Rule.prototype.run = function (context, options, resolve, reject) {
+	//jshint maxstatements: 17
 	const q = axe.utils.queue();
 	const ruleResult = new RuleResult(this);
+	const markStart = 'mark_runchecks_start_' + this.id;
+	const markEnd = 'mark_runchecks_end_' + this.id;
 	let nodes;
 
 	try {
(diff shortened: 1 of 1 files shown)


dylanb:
@paulirish you can get rid of the JSHint errors by hiking the var definitions up to the top of the function.

paulirish:
> you can get rid of the JSHint errors by hiking the var definitions up to the top of the function.

thanks, done.

dylanb:
As an FYI - the async code only runs when you are analyzing pages with iframes, so for the purposes of most perf testing, this code should be sufficient.

Thanks for the contribution!!

dylanb:
@WilcoFiers @marcysutton do we care about the commit log subject being "perf:"? We haven't used this before but I am ok with us adopting this going forward - WDYT?

marcysutton:
`perf:` sounds appropriate to me. I'm not 100% sure if [conventional-changelog](https://github.com/conventional-changelog/conventional-changelog) will include those commits in the automated changelog (they are allowed). We could always test it to find out.

WilcoFiers:
`perf:` doesn't show in the changelog. I think a 25% performance boost should be in `feat`.

dylanb:
@WilcoFiers this just adds the rule information to the performance timer functionality

paulirish:
retitled as "chore(performance): ". sgty?

Issue #706, opened 2018-02-01
run() function optional parameters create typescript error


plumpNation:
the axe.d.ts file has three optional parameters, however, the run function handles internally providing the callback on it's own.

`axe.run(callback);`

The recommended approaches I have found say use `undefined` for the optionals, however this causes issues with `normalizeRunParams()` that determines which parameter you are passing in.

I suggest adding overloads to the axe.d.ts file to solve this.

PR #707, opened 2018-02-01
chore(type-checking): Cannot use only callback in run() without ts error

plumpNation:
Addresses #706


WilcoFiers:
Cool! Can you also add the scenario where there's no callback? If you don't use a callback, aXe will return a thenable.

plumpNation:
Ok no worries. I'm going to remove the optional params and lock it down with explicit overloads instead.

I will also add some types so I can make the overloads more readable.

WilcoFiers:
Awesome. Thanks for the contribution @plumpNation Just a heads up, it'll be a few weeks before we can release this, since we're trying to put everything in its place to release aXe 3.0.

Issue #705, opened 2018-01-31
cleanupPlugins() will throw TypeError if its arguments are undefined.


lsiden:
cleanupPlugins() takes two args: resolve and reject.  Both must be defined by the caller or it will throw a TypeError (along with a different type of error).

Each of these arguments should have default values.

WilcoFiers:
Sounds reasonable. We'd welcome a pull request if you've got the time for it.

PR #709, opened 2018-02-01
Provide default arguments to axe.cleanup()

lsiden:
axe.cleanup() requires two functional arguments, 'resolve' and 'reject' that are not mentioned in the documentation (I will create a separate pull-request for that).  

This change will define default arguments with reasonable behavior:  default resolve() does nothing and default reject() logs an error message to the console.

This is related to Issue #705 .

### DIFF
--- a/lib/core/public/cleanup-plugins.js
+++ b/lib/core/public/cleanup-plugins.js
@@ -1,6 +1,9 @@
 
 function cleanupPlugins(resolve, reject) {
   'use strict';
+  resolve = resolve || function() {};
+  reject = reject || axe.log;
+
   if (!axe._audit) {
     throw new Error('No audit configured');
   }
--- a/doc/API.md
+++ b/doc/API.md
@@ -617,7 +617,18 @@ Register a plugin with the aXe plugin system. See [implementing a plugin](plugin
 
 ### API Name: axe.cleanup
 
-Call the plugin system's cleanup function. See [implementing a plugin](plugins.md).
+Call each plugin's cleanup function. See [implementing a plugin](plugins.md).
+
+The signature is:
+
+```
+    axe.cleanup(resolve, reject)
+```
+
+`resolve` and `reject` are functions that will be invoked on success or failure respectively.
+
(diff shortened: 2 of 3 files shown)


lib/core/public/cleanup-plugins.js
```diff
@@ -42,4 +42,8 @@ function cleanupPlugins(resolve, reject) {
   })
   .catch(reject);
 }
+
+function cleanupPlugins(resolve=(() => {}), reject=axe.log) {
```
WilcoFiers: You don't need to define a new function for this. Please just put these default params into the original function.

marcysutton:
Let's be sure to squash the commits when we merge this to a) get a commit message that adheres to our [commit policy](https://github.com/dequelabs/axe-core/blob/develop/doc/code-submission-guidelines.md), and b) get the merge commit out of there.

WilcoFiers:
@lsiden You'll have to agree to the license before we can merge in.

lib/core/public/cleanup-plugins.js
lsiden: I had to do this because of:

```
Warning: lib/core/public/cleanup-plugins.js: Non-simple parameter in strict mode (2:24) Use --force to continue.
```

In order to put default params in the original function signature, I would have to disable strict mode, which I don't think you want to do.

lsiden:
@WilcoFiers I already agreed to the License Agreement as "lsiden".  I think "Lawrence Siden" is from a the Github account "lsiden-deque@gmail.com" which I just deleted since I haven't used it.  I don't know why it still show up as a "contributor" to this PR.

lsiden:
Will do @marcysutton, once it's ready.  Thank you for reminding me.

lib/core/public/cleanup-plugins.js
WilcoFiers: This seems to be a JSHint bug. Just use fallback params instead:
```js
resolve = resolve || () => {};
reject = reject || axe.log;
```

Issue #687, opened 2018-01-19
Shadow DOM issue for color contrast rule


marcysutton:
axe-core is having trouble calculating color contrast inside of Shadow DOM, and it puts items into Needs Review for some reason. I've seen it occur for implicit labels with visually hidden text as well as buttons with visible text. I tested this with the latest dev build of aXe Coconut.

I believe @robdodson saw this issue a while back, too.

Example markup to populate within a shadow boundary (from my [a11y-demo-app](https://github.com/marcysutton/a11y-demo-app), uses [React-Shadow](https://github.com/Wildhoney/ReactShadow)):
```html
<trip-planner>
        <ShadowDOM include={[
          'css/normalize.css',
          '/fonts/font-awesome-4.2.0/css/font-awesome.min.css',
          '/css/trip-planner.css']
        }>
          <div>
            <form>
              <div>
                <div className="button-wrap left">
                  <button type="button" title="My trips">
                    <span class="visuallyhidden">My trips</span>
                    <i className="fa fa-star"></i>
                  </button>
                </div>
                <label className="fit">
                  <span class="visuallyhidden">Name of trip</span>
                  <input type="text" placeholder="Name of trip" />
                </label>
                <label className="fit">
                  <span class="visuallyhidden">Name of ski area</span>
                  <input type="text" placeholder="Name of ski area" />
                </label>
                <div className="button-wrap">
                  <button type="button" className="expand" title="Expand">
                    <span class="visuallyhidden">Expand widget</span>
                    <i className="fa fa-caret-down"></i>
                  </button>
                </div>
              </div>
              <div>
                <div className="inputGroup left">
                  <button type="button" aria-label="Select a previous date">
                    <i className="fa fa-caret-left"></i>
                  </button>
                  <label>
                    <span class="visuallyhidden">Date of trip</span>
                    <input type="text" placeholder="16.1.18" />
                  </label>
                  <button type="button" aria-label="Select a later date">
                    <i className="fa fa-caret-right"></i>
                  </button>
                </div>
                <div className="button-wrap right">
                  <button type="button" aria-label="Date picker">
                    <i className="fa fa-calendar"></i>
                  </button>
                </div>
                <div className="select right">
                  <label>
                    <span class="visuallyhidden">type of trip</span>
                    <select>
                      <option>Type of trip</option>
                      <option>Alpine ski resort</option>
                      <option>Backcountry</option>
                      <option>Nordic</option>
                    </select>
                  </label>
                </div>
[...]

PR #711, opened 2018-02-01
Shadow support for color contrast

marcysutton:
This PR adds color contrast support for pages with nested Shadow DOM. 

One TODO I still have is to support multiline elements inside of shadow boundaries (inline elements with line breaks, originally added in https://github.com/dequelabs/axe-core/pull/610). The actual background color for those is wrong in this PR because the tree walking across multiple roots doesn't really factor in client rects that partially overlap. I've struggled with it for the past week and decided to punt on it.

Part of the complication is that a shadow host is always returned from `document.elementsFromPoint` regardless of whether its content overlaps (i.e. absolute positioning, etc.). For block elements that don't visually overlap their host element, we have to filter out the host element from the stack or it ends up in Needs Review for `elmPartiallyObscured`. But then we're in a conundrum when multiline elements only partially overlap: our utilities for overlap checking only look at the boundingClientRect, and it would take some refactoring to check individual rects against a shadow host. 

I sorta wonder if the `contentOverlapping` method in `axe.commons.color.getBackgroundColor` should just go away....it seems like we're checking the same thing more than once, and it's way more complicated now (checking overlap of individual rects against multiple bgNodes, across multiple roots). Some of the work has been moved to the stack gathering steps, where we did more evaluation after the fact before. There's probably some redundancy that could get cleaned up as a next step.

Closes https://github.com/dequelabs/axe-core/issues/687

### DIFF
--- a/lib/commons/color/get-background-color.js
+++ b/lib/commons/color/get-background-color.js
@@ -57,8 +57,7 @@ function contentOverlapping(targetElement, bgNode) {
 	// get content box of target element
 	// check to see if the current bgNode is overlapping
 	var targetRect = targetElement.getClientRects()[0];
-	var doc = axe.commons.dom.getRootNode(targetElement);
-	var obscuringElements = doc.elementsFromPoint(targetRect.left, targetRect.top);
+	var obscuringElements = dom.shadowElementsFromPoint(targetRect.left, targetRect.top);
 	if (obscuringElements) {
 		for(var i = 0; i < obscuringElements.length; i++) {
 			if (obscuringElements[i] !== targetElement && obscuringElements[i] === bgNode) {
@@ -205,32 +204,6 @@ color.getCoords = function(rect) {
 	return {x, y};
 };
 /**
- * Get elements from point across shadow dom boundaries
- * @method getElementsFromPoint
- * @memberof axe.commons.color
- * @instance
- * @param {Object} nodeCoords X and Y coordinates of point
- * @param {Object} [root] Shadow root or document root
- * @return {Array}
- */
-color.getElementsFromPoint = function(nodeCoords, root = document) {
-  return root.elementsFromPoint(nodeCoords.x, nodeCoords.y)
-    .reduce((stack, elm) => {
-      if (axe.utils.isShadowRoot(elm)) {
-        const shadowStack = color.getElementsFromPoint(nodeCoords, elm.shadowRoot);
-        stack = stack.concat(shadowStack);
-        // filter host nodes which get included regardless of overlap
-        // TODO: refactor multiline overlap checking inside shadow dom
-        if (stack.length && axe.commons.dom.visuallyContains(stack[0], elm)) {
-        	stack.push(elm);
-        }
-      } else {
-      	stack.push(elm);
-      }
-      return stack;
-    }, []);
-};
-/**
  * Get relevant stacks of block and inline elements, excluding line breaks
  * @method getRectStack
  * @memberof axe.commons.color
@@ -241,7 +214,7 @@ color.getElementsFromPoint = function(nodeCoords, root = document) {
 color.getRectStack = function(elm) {
 	let boundingCoords = color.getCoords(elm.getBoundingClientRect());
 	if (boundingCoords) {
-		let boundingStack = color.getElementsFromPoint(boundingCoords);
+		let boundingStack = dom.shadowElementsFromPoint(boundingCoords.x, boundingCoords.y);
 		// allows inline elements spanning multiple lines to be evaluated
 		let rects = Array.from(elm.getClientRects());
 		if (rects && rects.length > 1) {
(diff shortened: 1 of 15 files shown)


test/commons/color/get-background-color.js
```diff
@@ -626,4 +648,146 @@ describe('color.getBackgroundColor', function () {
...
+		var actual = axe.commons.color.getBackgroundColor(target, [], false);
+
+		assert.closeTo(actual.red, 0, 0);
+		assert.closeTo(actual.green, 0, 0);
+		assert.closeTo(actual.blue, 0, 0);
+		assert.closeTo(actual.alpha, 1, 0);
+	});
+
+	(shadowSupported ? it : xit)
+	('should count an implicit label as a background element inside shadow dom', function () {
```
WilcoFiers: You should use a different colour so you can distinguish it from the background.

test/commons/color/get-background-color.js
```diff
@@ -626,4 +648,146 @@ describe('color.getBackgroundColor', function () {
...
+		assert.closeTo(actual.blue, 255, 0);
+		assert.closeTo(actual.alpha, 1, 0);
+	});
+
+	(shadowSupported ? it : xit)
+	('finds a color for absolutely positioned content when background is in shadow dom', function () {
+		fixture.innerHTML = '<div id="elm1" style="width:10em; height:0; position:absolute;"></div>' +
+			'<div id="elm2" style="color:green; position:absolute;">Text</div>';
+
+	  var elm1 = document.querySelector('#elm1');
```
WilcoFiers: Mixing tabs and spaces.

test/commons/color/get-background-color.js
```diff
@@ -626,4 +648,146 @@ describe('color.getBackgroundColor', function () {
...
+	(shadowSupported ? it : xit)
+	('finds colors for absolutely positioned elements across shadow boundaries', function () {
+		fixture.innerHTML = '<div id="container" style="background-color:black; height:20px; position:relative;"></div>';
+		var container = fixture.querySelector('#container');
+		var shadow = container.attachShadow({ mode: 'open' });
+		shadow.innerHTML = '<div id="shadowTarget" style="color:#333; height:20px; position:absolute; top:20px;">Text</div>';
+
+		var target = shadow.querySelector('#shadowTarget');
+		var actual = axe.commons.color.getBackgroundColor(target, []);
+		assert.closeTo(actual.red, 255, 0);
```
WilcoFiers: Why did you use `assert.closeTo()` here? If your margin is 0 you can use `assert.equal`, right? Same question for some of the other tests below.

test/commons/color/get-background-color.js
```diff
@@ -626,4 +648,146 @@ describe('color.getBackgroundColor', function () {
...
+		fixture.innerHTML = '<div id="container" style="background-color:black; height:20px;"></div>';
+		var container = fixture.querySelector('#container');
+		var shadow = container.attachShadow({ mode: 'open' });
+		shadow.innerHTML = '<div id="shadowTarget" style="color:#333;">Text<br>More text</div>';
+
+		var target = shadow.querySelector('#shadowTarget');
+		var actual = axe.commons.color.getBackgroundColor(target, []);
+		assert.isNull(actual);
+	});
 });
```
WilcoFiers: We should test slotted text with a background inside a shadow tree.

test/commons/color/get-background-color.js
marcysutton: I'd copied it from the existing tests, I can change it!

test/commons/color/get-background-color.js
marcysutton: Thanks Sublime 😡

lib/commons/color/get-background-color.js
```diff
@@ -204,7 +205,33 @@ color.getCoords = function(rect) {
...
- * Get elements from point for block and inline elements, excluding line breaks
+ * Get elements from point across shadow dom boundaries
+ * @method getElementsFromPoint
+ * @memberof axe.commons.color
+ * @instance
+ * @param {Object} nodeCoords X and Y coordinates of point
+ * @param {Object} [root] Shadow root or document root
+ * @return {Array}
+ */
+color.getElementsFromPoint = function(nodeCoords, root = document) {
```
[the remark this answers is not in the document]
marcysutton: Done.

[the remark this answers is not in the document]
marcysutton: However, in doing so, I added some tests that are apparently pretty brittle on Circle. they all pass locally. I don't think we should sink a bunch of time into this if the API is tested in a bunch of other places (getBackgroundColor, contrast check tests).

marcysutton:
I can't get that chrome:mobile test to fail locally even in the Responsive Design viewer at that viewport, so it's a Circle-only error. I'm also unable to debug Circle with Xquartz any longer...something is borked in that process for me. When I try to run Chrome or `xclock` I get `"Can't open display: localhost:10.0"` with very little permissions to fix it. Searching didn't bring up any promising debugging tips. https://github.com/dequelabs/axe-core/blob/develop/CONTRIBUTING.md#debugging-tests-that-only-fail-on-circleci

PR #712, opened 2018-02-03
fix(perf): improve select performance fixes #702

### LINKED ISSUE #702: Performance of axe.utils.select('*') is slow with lots of nodes
Example trace: https://chromedevtools.github.io/timeline-viewer/?loadTimelineFromURL=drive://1qJeP82OwoRZ0g4PS-bvJuf_FKI0UfIEm

The above trace is of the mlb homepage with a dev build of axe (including both #699 and #701)

You can see these pretty big chunks of array sorting:
image

Looking into it, it appears to be when `axe.utils.select` is called with a `selector` of `*` and then does the sort() to get document order.  Do we need these results in an order other than what qSA returns?

dylanb:
Improve the performance of `axe.utils.select`

This gets rid of the need for a sort in all instances except when there is an overlapping include context
Memoizes select so that subsequent calls to `axe.utils.select` from within `Audit.run` will return the cached results

### DIFF
--- a/lib/core/utils/qsa.js
+++ b/lib/core/utils/qsa.js
@@ -29,7 +29,7 @@ function matchesPseudos (target, exp) {
 
 	if (!exp.pseudos || exp.pseudos.reduce((result, pseudo) => {
 		if (pseudo.name === 'not') {
-			return result && !matchExpressions([target], pseudo.expressions, false, target.shadowId).length;
+			return result && !matchExpressions([target], pseudo.expressions, false).length;
 		}
 		throw new Error('the pseudo selector ' + pseudo.name + ' has not yet been implemented');
 	}, true)) {
@@ -193,12 +193,12 @@ function matchesSelector (node, exp) {
 		);
 }
 
-matchExpressions = function  (domTree, expressions, recurse, parentShadowId, filter) {
+matchExpressions = function  (domTree, expressions, recurse, filter) {
 	//jshint maxstatements:34
 	//jshint maxcomplexity:15
 	let stack = [];
 	let nodes = Array.isArray(domTree) ? domTree : [domTree];
-	let currentLevel = createLocalVariables(nodes, expressions, [], parentShadowId);
+	let currentLevel = createLocalVariables(nodes, expressions, [], domTree[0].shadowId);
 	let result = [];
 
 	while (currentLevel.nodes.length) {
@@ -266,13 +266,13 @@ axe.utils.querySelectorAll = function (domTree, selector) {
  * querySelectorAllFilter implements querySelectorAll on the virtual DOM with
  * ability to filter the returned nodes using an optional supplied filter function
  *
- * @method querySelectorAll
+ * @method querySelectorAllFilter
  * @memberof axe.utils
  * @instance
  * @param  {NodeList} domTree flattened tree collection to search
  * @param  {String} selector String containing one or more CSS selectors separated by commas
  * @param  {Function} filter function (optional)
- * @return {NodeList} Elements matched by any of the selectors and filtered by the filter function
+ * @return {Array} Elements matched by any of the selectors and filtered by the filter function
  */
 
 axe.utils.querySelectorAllFilter = function (domTree, selector, filter) {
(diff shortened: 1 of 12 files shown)


lib/core/utils/select.js
```diff
@@ -73,24 +67,68 @@ function pushNode(result, nodes, context) {
...
 axe.utils.select = function select(selector, context) {
+	//jshint maxstatements:20
 	'use strict';
 
 	var result = [], candidate;
-	for (var i = 0, l = context.include.length; i < l; i++) {
+	if (!Array.isArray(context.include)) {
+		context.include = Array.from(context.include);
+	}
+	context.include.sort(axe.utils.nodeSorter); // ensure that the order of the include nodes is document order
```
WilcoFiers: We should enforce array type and sort the include as part of `normalizeContext` instead. That way this function doesn't have any side effects that can cause bugs later, and we don't have to sort more than once.
https://github.com/dequelabs/axe-core/blob/develop/lib/core/base/context.js#L72

lib/core/utils/qsa.js
```diff
@@ -227,9 +259,26 @@ matchExpressions = function  (domTree, expressions, recurse) {
...
  */
 axe.utils.querySelectorAll = function (domTree, selector) {
+	return axe.utils.querySelectorAllFilter(domTree, selector);
+};
+
+/**
+ * querySelectorAllFilter implements querySelectorAll on the virtual DOM with
+ * ability to filter the returned nodes using an optional supplied filter function
+ *
+ * @method querySelectorAll
```
WilcoFiers: Should be `querySelectorAllFilter`.

lib/core/utils/qsa.js
```diff
@@ -227,9 +259,26 @@ matchExpressions = function  (domTree, expressions, recurse) {
...
+ * @memberof axe.utils
+ * @instance
+ * @param  {NodeList} domTree flattened tree collection to search
+ * @param  {String} selector String containing one or more CSS selectors separated by commas
+ * @param  {Function} filter function (optional)
+ * @return {NodeList} Elements matched by any of the selectors and filtered by the filter function
```
WilcoFiers: This returns an array, not a node list.

lib/core/utils/qsa.js
```diff
@@ -227,9 +259,26 @@ matchExpressions = function  (domTree, expressions, recurse) {
...
+ */
+
+axe.utils.querySelectorAllFilter = function (domTree, selector, filter) {
```
WilcoFiers: You should have a few tests for this. Particularly to test the filter option.

lib/core/utils/qsa.js
```diff
@@ -194,27 +173,80 @@ convertExpressions = function (expressions) {
...
+function matchesSelector (node, exp) {
+	return (matchesTag(node.actualNode, exp[0]) &&
+			matchesClasses(node.actualNode, exp[0]) &&
+			matchesAttributes(node.actualNode, exp[0]) &&
+			matchesId(node.actualNode, exp[0]) &&
+			matchesPseudos(node, exp[0])
+		);
+}
+
+matchExpressions = function  (domTree, expressions, recurse, parentShadowId, filter) {
```
WilcoFiers: Why did you need to pass the `parentShadowId` in separately? Looks to me like you could've just take that from `domTree`.

lib/core/utils/select.js
dylanb: We would have to do it in the Context constructor because include is modified after `normalizeContext`. I can make that change.

lib/core/utils/select.js
```diff
@@ -73,24 +67,68 @@ function pushNode(result, nodes, context) {
...
+			axe.utils.matchesSelector(candidate.actualNode, selector) &&
+			curried(candidate)) {
+			result = pushNode(result, [candidate]);
 		}
-		result = pushNode(result, axe.utils.querySelectorAll(candidate, selector), context);
+		result = pushNode(result, axe.utils.querySelectorAllFilter(candidate, selector, curried));
 	}
-
-	return result.sort(axe.utils.nodeSorter);
+	if (context.include.length > 1 && hasOverlappingIncludes(context.include)) {
```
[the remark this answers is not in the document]
dylanb: good idea

dylanb:
@WilcoFiers updated

PR #714, opened 2018-02-06
fix(perf): Improve the speed of the getSelector code

dylanb:
**DO NOT MERGE**

This code completely re-implements the getSelector code basing it around statistics generated from the target page rather than hard-coded values.

### DIFF
--- a/test/core/utils/get-selector.js
+++ b/test/core/utils/get-selector.js
@@ -22,6 +22,8 @@ describe('axe.utils.getSelector', function () {
 
 	afterEach(function () {
 		fixture.innerHTML = '';
+		axe._tree = undefined;
+		axe._selectorData = undefined;
 	});
 
 	it('should be a function', function () {
@@ -31,6 +33,7 @@ describe('axe.utils.getSelector', function () {
 	it('should generate a unique CSS selector', function () {
 		var node = document.createElement('div');
 		fixture.appendChild(node);
+		axe._tree = axe.utils.getFlattenedTree(document.documentElement);
 
 		var sel = axe.utils.getSelector(node);
 
@@ -45,6 +48,7 @@ describe('axe.utils.getSelector', function () {
 		var node = document.createElement('div');
 		node.className = '    ';
 		fixture.appendChild(node);
+		axe._tree = axe.utils.getFlattenedTree(document.documentElement);
 
 		var sel = axe.utils.getSelector(node);
 
@@ -59,6 +63,7 @@ describe('axe.utils.getSelector', function () {
 		var node = document.createElement('div');
 		node.id = 'monkeys#are.animals\\ok';
 		fixture.appendChild(node);
+		axe._tree = axe.utils.getFlattenedTree(document.documentElement);
 
 		var result = document.querySelectorAll(axe.utils.getSelector(node));
 		assert.lengthOf(result, 1);
@@ -69,6 +74,7 @@ describe('axe.utils.getSelector', function () {
 		var node = document.createElement('div');
 		node.className = '.  bb-required';
 		fixture.appendChild(node);
+		axe._tree = axe.utils.getFlattenedTree(document.documentElement);
 
 		var result = document.querySelectorAll(axe.utils.getSelector(node));
 		assert.lengthOf(result, 1);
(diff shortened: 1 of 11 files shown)


dylanb:
@WilcoFiers A careful review of the code to make sure that I have not missed any of the previous functionality that is important.

I did remove some tests because they were very dependent on the previous implementation.

dylanb:
I actually do like the idea of having two different implementations - one for scripting (which has different requirements) and one for analysis - which has to find a unique selector on the page in question and do it in high volumes, quickly.

The biggest problem with waiting until after 3.0 is that this does change the "API" in that it changes the selectors and possibly adds a new function call and a major version change is a better time to do that.

test/core/utils/get-selector.js
```diff
@@ -131,10 +121,11 @@ describe('axe.utils.getSelector', function () {
 		node = document.createElement('div');
 		node.className = 'dogs cats';
 		fixture.appendChild(node);
+		axe._tree = axe.utils.getFlattenedTree(document.documentElement);
 
 		var sel = axe.utils.getSelector(node);
 
-		assert.equal(sel, '#fixture > div.dogs.cats');
+		assert.equal(sel, '#fixture > div.cats');
```
marcysutton: Does this API only support one CSS class now?

WilcoFiers:
I've spent a few hours running this implementation on a dozen or so page, collecting all of those selectors and working out where this is breaking down. I've got the following suggestions:

- Use an attributes / class name blacklist to ignore state indicators:
  hidden, checked, selected, focused, all of their aria equivalents, etc.

- Ignore extremely long attributes+value pairs, class names and even IDs.
  I've found many very long `data-*`, `aria-*` attributes, as well as a couple of IDs and classes.
  There are also `item*` attributes custom attributes like `ng-*` that can be problematic. You may also want to completely exclude `on*` attributes.
  - If they are unique, still ignore them if they are 50+ chars long
  - If they aren't unique, ignore them if they are 25+ chars long

- Use more 'properties' for the target element, and fewer for its ancestors. Preferably the ancestor
  segment should only have one property, possibly the rarest of node name, attribute and class name.
  The target element should have two of these, except if it's a unique ID.

- `div` and `span` should not be part of the selector, unless it's the only thing available for it

- If an element has a role that's different from its native role, this role should be used
  in place of the node name. So: `<button role="slider">` is `[role^="slider"]`, but 
  `<main role="main">` is still `main`. We should also limit this to the first role
  in the role attribute I think.

- Use `nth-child` instead of `nth-of-type`. It's shorter and consistent with the current one so 
  there's less chance of problems with Comply.

- Use the existing href / src selector method, selecting only the end of the URL. Inclusing
  the whole URL is often unnecessary and makes the selector harder to read.

- Use `:root` to select the document element.

- I think you're over-escaping content inside of attribute values. I saw this for example:
  `div[tabindex="-\31 "]`, `a[href="\/style\/article\/david-hockney-at-80-interview-los-angeles\/index\.html"]`

dylanb:
Have done more research on CSS selector performance. Here is a good resource http://johnotander.com/css/2015/06/10/css-performance/

 I also did some performance testing in Chrome

Some good practices to consider:
1) CSS selectors are evaluated right-to-left, so selectors should be more specific farther right
2) Avoid universal selectors E.g. `.some-class *` will force the evaluation of all the nodes in the document. So will `.some-class [some-attribute=some-value]` because that is equivalent to `.some-class *[some-attribute=some-value]`
3) :nth-child is exepensive
4) dependent selectors are exepensive
5) classes and ids are cheap
6) :nth-of-type and attribute selectors are almost equivalent (and are cheap) when qualified by a class
7) Speed of [att$=""] seems to be related to the number of matches (not as sure about this)

Given that, I think we should do the following:

1) blacklist some state indicators `tabindex`, `hidden`, `checked`, `selected`, `focused`, `disabled` - mostly because these can also change
2) Try to use uncommon classes first before trying attributes (i.e. try all classes of equal or lower commonality before attributes)
3) Do not use the tag unless there is nothing else
4) Try to exhaust all uncommon classes and attributes before using the parent
5) truncate attribute values to max 15 characters both during stats generation and during selector generation
6) Do not ignore, or restrict class names and ids that are unique and long (find a compromise between better readability and better performance)
7) blacklist on* attributes

WilcoFiers:
Sounds good. I would add that you try to split attributes between 12 / 20 chars, instead of just cutting. Classes should probably split before a `-`, that seems pretty common, URLs after a `/`. Not sure about others, but I think that may help readability a bit.

dylanb:
Closing pull request, will resubmit another if the research works out

PR #720, opened 2018-02-07
axe.configure disableOtherRules option

### LINKED ISSUE #688: Add a disableOtherRules option to axe.configure
This will help get rid of some confusion about why rules are running after you called axe.configure with a new set of rules.

WilcoFiers:
Closes #688

### DIFF
--- a/test/integration/full/configure-options/configure-options.js
+++ b/test/integration/full/configure-options/configure-options.js
@@ -1,48 +1,80 @@
-describe('Check Configure Options', function() {
+describe('Configure Options', function() {
 	'use strict';
 
-	var target = document.querySelector('#target');
+	afterEach(function () {
+		axe.reset();
+	});
 
-	describe('aria-allowed-attr', function() {
-		it('should allow an attribute supplied in options', function(done) {
-			target.setAttribute('role', 'separator');
-			target.setAttribute('aria-valuenow', '0');
+	describe('Check', function() {
 
-			axe.configure({
-				checks: [{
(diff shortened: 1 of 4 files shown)


WilcoFiers:
@marcysutton I don't think integration tests are good for that. They are very slow because they need to load in a separate page. What exactly do you think isn't covered in the tests that you'd want to see covered?

marcysutton:
I mentioned integration tests because the core unit test relies on non-public API methods and properties, like `axe._load` and `axe._audit`. I always like to know that the whole system is working, not just the internal stuff.

WilcoFiers:
@marcysutton Updated, please review again.

test/integration/full/configure-options/configure-options.js
```diff
@@ -1,50 +1,82 @@
-describe('Check Configure Options', function() {
+describe('Configure Options', function() {
```
marcysutton: I named it `Check configure options` to distinguish between the other `options-parameter` tests in the same folder. Perhaps these new tests should be combined with those instead? Or all together?

WilcoFiers: oh! I see, you're right. I put them in the wrong place. My bad.

WilcoFiers: Actually, no, this is correct. The `options parameter` is the `axe.run` options. This is for `axe.configure`. Your existing test was for `axe.configure({ check })`, this new one is `axe.configure({ disableOtherRules })`.

PR #721, opened 2018-02-07
Fix audio-caption SC tag

### LINKED ISSUE #686: audio elements must have a track for captions - does this rule apply to prerecorded audio-only media
Does the rule "audio elements must have a track for captions" apply to prerecorded audio-only content: (https://dequeuniversity.com/rules/axe/2.6/audio-caption)? 

If yes: this is not a WCAG requirement. Per SC 1.2.1, audio-only content is required to have an "alternative for time-based media" where "alternative for time-based media" is a "document including correctly sequenced text descriptions of time-based visual and auditory information and providing a means for achieving the outcomes of any time-based interaction". So a transcript satisfies SC 1.2.1. SC 1.2.2 (referenced in https://dequeuniversity.com/rules/axe/2.6/audio-caption) is not applicable to prerecorded audio-only.

WilcoFiers:
Closes #686

### DIFF
--- a/doc/rule-descriptions.md
+++ b/doc/rule-descriptions.md
@@ -10,7 +10,7 @@
 | aria-roles | Ensures all elements with a role attribute use a valid value | cat.aria, wcag2a, wcag131, wcag411, wcag412 | true |
 | aria-valid-attr-value | Ensures all ARIA attributes have valid values | cat.aria, wcag2a, wcag131, wcag411, wcag412 | true |
 | aria-valid-attr | Ensures attributes that begin with aria- are valid ARIA attributes | cat.aria, wcag2a, wcag411 | true |
-| audio-caption | Ensures &lt;audio&gt; elements have captions | cat.time-and-media, wcag2a, wcag122, section508, section508.22.a | true |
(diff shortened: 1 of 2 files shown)


chaals:
Don't the tests **also** apply to 1.2.2, 1.2.4, and 1.2.9?

(I am not sure if I misunderstood the tag system though)

WilcoFiers:
@chaals I don't think it does. Those are video SCs, this is audio only.

chaals:
But they require, for video which includes audio (or for live-streamed audio in the case of 1.2.9 I think), audio captions - i.e. the same requirement.

(if they do apply, am I making the right assumption that multiple checkpoint tags should be applied?)

WilcoFiers:
The audio-caption rule only checks `<audio>` elements. I don't see how 1.2.2 says:

> all prerecorded audio content in **synchronized media**

There's a slight assumption we are making that the `<audio>` element isn't synchronised with something else on the page, but I think that's generally not the case. So 99% of the time I would say this only applies to 1.2.1. Also as a rule of thumb, we use one SC, not all that may be applicable.

PR #722, opened 2018-02-07
headings rules: Skip headings with a role change

### LINKED ISSUE #645: 'empty-heading' rule fails if role is set on <h1>-<h6> tags
Ex: 
```
<h1 role="banner" id="heading-1">Headings test</h1>
```
'empty-heading' rule failure might not be appropriate in such cases

WilcoFiers:
Closes #645


lib/rules/heading-matches.js
```diff
@@ -0,0 +1,13 @@
+// Get all valid roles
+let explicitRoles;
+if (node.hasAttribute('role')) {
+  explicitRoles = node.getAttribute('role').split(/\s+/i)
+    .filter(axe.commons.aria.isValidRole);
```
marcysutton: The splitting and filtering here seems like a utility we could reuse.

test/rule-matches/heading-matches.js
```diff
@@ -0,0 +1,55 @@
...
+    assert.isTrue(rule.matches(div));
+
+    div.setAttribute('role', 'slider heading');
+    assert.isTrue(rule.matches(div));
+  });
+
+  it('should return true on regular headings without roles', function () {
+    var h1 = document.createElement('h1');
+    var h2 = document.createElement('h2');
+    var h3 = document.createElement('h3');
```
marcysutton: Do these in-memory elements get cleaned up automatically by the test runner?

WilcoFiers: No, but they do by the garbage collector.

lib/rules/heading-matches.js
WilcoFiers: Seems pretty trivial. I would like us to at some point move towards a stronger getRoles() method, but I think that needs a little more thought, and we should probably have a better idea of how we do accessibility support before we can.

PR #725, opened 2018-02-09
Add github templates for issues and PRs

marcysutton:
We have to repeat ourselves a lot in issues and PRs, so it makes sense to add Github templates that point out common things (commit policy, accessibility supported, axe-core/integration versions, etc.). I started with the Angular templates so we may want to make additional changes. I'm open to suggestions!

### DIFF
--- a/doc/.github/ISSUE_TEMPLATE.md
+++ b/doc/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,26 @@
+<!-- Thanks for reporting an issue to axe-core. Please provide all necessary info to reproduce the issue. Without adequate details, your issue may be closed without investigation. -->
+
+## Issue Description
+<!-- Basic description of the problem. Is it a false positive? Feature request? -->
+
+### Current Behavior
+<!-- Describe how the issue manifests. -->
+
+### Expected Behavior
+<!-- Describe what the desired behavior would be. -->
+
+### Steps to reproduce the behavior
+<!--
+Please provide the *STEPS TO REPRODUCE* and if possible, a *MINIMAL DEMO* of the problem via Codepen, jsbin, Plunkr or similar.
+-->
+
+<pre><code>
+axe-core version: X.Y.Z
+axe-webdriver, extension or other integration version: X.Y.Z
+
+Browser and Assistive Technology versions
+
+For Tooling issues:
+- Node version: XX  <!-- run `node --version` -->
+- Platform:  <!-- Mac, Linux, Windows -->
+</code></pre>
--- a/doc/.github/ISSUE_TEMPLATE.md
+++ b/doc/.github/ISSUE_TEMPLATE.md
@@ -7,7 +7,7 @@
 <!-- Describe how the issue manifests. -->
 
 ### Expected Behavior
-<!-- Describe what the desired behavior would be. -->
+<!-- Describe what the desired behavior would be, and how it applies to "Accessibility Supported" https://github.com/dequelabs/axe-core/blob/develop/doc/accessibility-supported.md -->
 
 ### Steps to reproduce the behavior
 <!--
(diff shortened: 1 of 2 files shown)


doc/.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,26 @@
+<!-- Thanks for reporting an issue to axe-core. Please provide all necessary info to reproduce the issue. Without adequate details, your issue may be closed without investigation. -->
```
nschonni: cool, I never thought to do the instructions this way

doc/.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Issue Description
```
WilcoFiers: I'm not too keen on templates like this. I find them a bit off putting when I see them in other projects. Also, we already have a template for new rules. I think this will confuse things. I'd rather we just give some general instructions. I would suggest we add a basic comment at the top, something like:

> If you're reporting a bug, include a description of the issue and a page or code snippet where it can be reproduced. Please make sure you have tested this with the latest version of axe-core. When proposing a new rule, please use our [rule template](#).

doc/.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,26 @@
+## PR Checklist
+Please check if your PR fulfills the following requirements:
+
+- [ ] The commit message(s) follow our guidelines: https://github.com/dequelabs/axe-core/blob/develop/doc/code-submission-guidelines.md#git-commits
+- [ ] Changes to rules and checks appropriately support [Shadow DOM](https://github.com/dequelabs/axe-core/blob/develop/doc/developer-guide.md)
+- [ ] Tests for the changes have been added (for bug fixes / features)
+- [ ] Docs have been added / updated (for bug fixes / features)
+
+
+## What is the current behavior?
```
WilcoFiers: We don't need these if we require every PR to have an issue associated with it. You can find the description there.

doc/.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,26 @@
+## PR Checklist
+Please check if your PR fulfills the following requirements:
```
WilcoFiers: I like these. Should we add something for testing with AT?

doc/.github/ISSUE_TEMPLATE.md
marcysutton: I do think we should mention accessibility supported in the issue template though–more often than not that's where we have to bring it up. Also, reminding them to include a demo.

doc/.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,19 @@
+<!-- Thanks for reporting an issue to axe-core. Please provide all necessary info to reproduce the issue. Without adequate details, your issue may be closed without investigation. 
+
+If you’re reporting a bug, include a description of the issue and a page or code snippet where it can be reproduced. Please make sure you have tested this with the latest version of axe-core. When proposing a new rule, please use our [rule template](https://github.com/dequelabs/axe-core/blob/develop/doc/rule-proposal.md).
```
WilcoFiers: The markdown link won't render in a comment :)

doc/.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,19 @@
...
+
+Describe what the desired behavior would be, and how it applies to "Accessibility Supported" https://github.com/dequelabs/axe-core/blob/develop/doc/accessibility-supported.md
```
WilcoFiers: Not sure what you're asking for with "accessibility supported"?

marcysutton: I'm asking that people read it and apply it to their change. We have to mention it all the time.

PR #729, opened 2018-02-10
test(testUtils): Add awaitNestedLoad utility function

### LINKED ISSUE #380: Remove race conditions from iframe integration tests

WilcoFiers:
Closes #380

### DIFF
--- a/test/integration/full/context/context.html
+++ b/test/integration/full/context/context.html
@@ -7,6 +7,7 @@
 	<script src="/node_modules/mocha/mocha.js"></script>
 	<script src="/node_modules/chai/chai.js"></script>
 	<script src="/axe.js"></script>
+	<script src="/test/testutils.js"></script>
 	<script>
 		mocha.setup({
 			timeout: 10000,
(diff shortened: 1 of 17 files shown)


WilcoFiers:
@marcysutton I'm having a difficult time integrating it into those. Any objection to doing this another time? That wasn't what this PR was designed to solve anyway.

marcysutton:
@WilcoFiers sounds fine to me, I'd rather keep stuff moving.

PR #732, opened 2018-02-12
Banner contentinfo/no more than one

jongund:
These are two new tests for testing to make sure there is only one banner and contentinfo landmark on a page.   These rules are similar to no more than on main landmark on a page.

We tried to use options parameter to use one evaluate file, but ran into problems with the testing harness, where some of the unit tests failed, but the integration tests always passed.


WilcoFiers:
I think this PR should have refactored the main-is-top-level check into a generic landmark-is-top-level check, that can be reused by all three rules. That saves a whole lot of code. It's pretty easy to do, and it prevents a lot of changes.

```json
// main-is-top-level.json -> landmark-is-top-level.json
{
  "id": "landmark-is-top-level",
  "evaluate": "landmark-is-top-level.js",
  "metadata": {
    "impact": "moderate",
    "messages": {
      "pass": "The {{=it.data}} landmark is at the top level.",
      "fail": "The {{=it.data}} landmark is contained in another landmark."
    }
  }
}
```

```js
// main-is-top-level.js -> landmark-is-top-level.js
const role = node.getAttribute('node') || axe.commons.aria.implicitRole(node);
this.data(role);
const landmarks = axe.commons.aria.getRolesByType('landmark');
let parent = axe.commons.dom.getComposedParent(node);

while (parent) {
  const role = parent.getAttribute('role');
  if (!role && (parent.tagName.toLowerCase() !== 'form')){
    role = axe.commons.aria.implicitRole(parent);
  }
  if (role && landmarks.includes(role)){
    return false;
  }
  parent = axe.commons.dom.getComposedParent(parent);
}
return true;
```

You can use this check in all three rules. You'd have to change the selector to exclude elements with their default role changes, so for main it would be: `main:not([role]), [role="main"]`.

lib/checks/keyboard/has-no-more-than-one-banner.js
```diff
@@ -0,0 +1,29 @@
+const banners = axe.utils.querySelectorAll(virtualNode, 'header, [role=banner]');
+const sectioning = ['article','aside','main','nav','section'];
+var count = 0;
+
+function isBanner(node){
+  var parent = axe.commons.dom.getComposedParent(node);
+  while (parent){
+    if (sectioning.includes(parent.tagName.toLowerCase())){
```
WilcoFiers: I don't quite understand what this is supposed to do. So if a `<header>` is contained in a `<main>`, it's ignored? Why is that?

WilcoFiers:
This PR is superseded with #760.

Issue #690, opened 2018-01-22
Ensure all rules and checks have Shadow DOM coverage


marcysutton:
To prevent issues like https://github.com/dequelabs/axe-core/issues/687, we should go through every rule and check to make sure they have Shadow DOM tests. For example, color contrast was updated to work with Shadow DOM, but I missed adding Shadow DOM-specific unit or integration tests. 

We want to ensure quality before we can release aXe 3.0, so this will be an important step toward that milestone.

PR #733, opened 2018-02-13
Update remaining checks to support Shadow DOM

marcysutton:
I went through each of the checks to see if there was Shadow DOM coverage, and there were a couple that needed updating:

- header-present
- internal-link-present

There were a few others where I simply added unit tests:
- aria-errormessage
- p-as-heading

And a few I updated to use accessibleTextVirtual:
- focusable-no-name
- implicit
- button-has-visible-text
- has-visible-text

I also did a bunch of test cleanup and fixed up the shadowCheckSetup utility to support a wider range of scenarios.

Closes https://github.com/dequelabs/axe-core/issues/690

### DIFF
--- a/test/checks/forms/fieldset.js
+++ b/test/checks/forms/fieldset.js
@@ -4,19 +4,11 @@ describe('fieldset', function () {
 	var shadowSupport = axe.testUtils.shadowSupport.v1;
 	var fixtureSetup = axe.testUtils.fixtureSetup;
 
-	var checkContext = {
-		_data: null,
-		data: function (d) {
-			this._data = d;
-		},
-		relatedNodes: function (nodes) {
-			this._relatedNodes = Array.isArray(nodes) ? nodes : [nodes];
-		}
-	};
+	var checkContext = axe.testUtils.MockCheckContext();
 
 	afterEach(function () {
 		fixture.innerHTML = '';
-		checkContext._data = null;
+		checkContext.reset();
 	});
 
 	function tests(type) {
--- a/test/checks/keyboard/has-at-least-one-main.js
+++ b/test/checks/keyboard/has-at-least-one-main.js
@@ -39,7 +39,11 @@ describe('has-at-least-one-main', function () {
 	});
 	
 	it('should return true if one div has role equal to main', function() {
-		var params = checkSetup('<div id = "target" role = "main">Div with role main</div>');
+		var params = checkSetup('<div id="target" role = "main">Div with role main</div>');
+		var mainIsFound = checks['has-at-least-one-main'].evaluate.apply(checkContext, params);
+		assert.isTrue(mainIsFound);
+		assert.equal(checkContext._data, mainIsFound);
+	});
 		var mainIsFound = checks['has-at-least-one-main'].evaluate.apply(checkContext, params);
 		assert.isTrue(mainIsFound);
 		assert.equal(checkContext._data, mainIsFound);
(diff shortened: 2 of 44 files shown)


lib/checks/navigation/p-as-heading.js
```diff
@@ -1,4 +1,5 @@
-let siblings = Array.from(node.parentNode.children);
+let parent = node.parentNode;
```
dylanb: This change does not seem necessary

test/checks/forms/fieldset.js
```diff
@@ -4,19 +4,11 @@ describe('fieldset', function () {
 	var shadowSupport = axe.testUtils.shadowSupport.v1;
 	var fixtureSetup = axe.testUtils.fixtureSetup;
 
-	var checkContext = {
-		_data: null,
-		data: function (d) {
-			this._data = d;
-		},
-		relatedNodes: function (nodes) {
-			this._relatedNodes = Array.isArray(nodes) ? nodes : [nodes];
```
dylanb: Should we make `MockCheckContext` do the same conditional assignment as this is doing?

test/checks/keyboard/has-at-least-one-main.js
```diff
@@ -4,42 +4,53 @@ describe('has-at-least-one-main', function () {
...
 
 	afterEach(function () {
 		fixture.innerHTML = '';
 		checkContext.reset();
+		axe._tree = undefined;
 	});
 
 	it('should return false if no div has role property', function() {
-		var params = checkSetup('<div id = "target">No role</div>');
+		var params = checkSetup('<div id="target">No role</div>');
```
dylanb: this is a `style` change and should not be mixed up with other changes

test/checks/keyboard/tabindex.js
```diff
@@ -13,17 +13,14 @@ describe('tabindex', function () {
 		fixture.appendChild(node);
 
 		assert.isFalse(checks.tabindex.evaluate(node));
-
-
```
dylanb: style changes

test/testutils.js
```diff
@@ -100,27 +100,34 @@ testUtils.checkSetup = function (content, options, target) {
...
- * @param Node|String 	Stuff to go into the fixture (html or node)
+ * @param Node|String 	Stuff to go into the fixture (html string or DOM Node)
  * @param Node|String 	Stuff to go into the shadow boundary (html or node)
- * @param String  			Target selector for the check, can be inside or outside of Shadow DOM (default: '#target')
  * @param Object  			Options argument for the check (optional, default: {})
+ * @param String  			Target selector for the check, can be inside or outside of Shadow DOM (optional, default: '#target')
  * @return Array
  */
-testUtils.shadowCheckSetup = function (content, shadowContent, targetSelector, options) {
+testUtils.shadowCheckSetup = function (content, shadowContent, options, targetSelector) {
```
dylanb: Why the switch?

lib/checks/navigation/p-as-heading.js
marcysutton: ah, a leftover thing from some experimentation I did. you're right–I thought I restored it, but I guess not.

test/checks/forms/fieldset.js
marcysutton: It does do it–that's how I got the tests to pass

test/testutils.js
marcysutton: To make it more consistent with `checkSetup`. There are no instances of `targetSelector`, yet there are options passed in. I wanted to preserve the functionality but make the two methods more consistent

marcysutton:
I've updated the PR with a style-specific commit and addressed the other feedback.

PR #739, opened 2018-02-21
Example updates

WilcoFiers:


doc/examples/jest+react/package.json
```diff
@@ -8,21 +8,21 @@
...
-  "dependencies": {
-  },
+  "dependencies": {},
   "scripts": {
     "test": "jest"
   },
   "devDependencies": {
-    "axe-core": "^2.2.3",
-    "babel-jest": "^20.0.3",
+    "axe-core": "^2.6.1",
```
marcysutton: Should we update these to axe 3x since they're going onto develop?

doc/examples/selenium/package.json
```diff
@@ -1,20 +0,0 @@
-{
-  "name": "axe-webdriverjs-example",
```
marcysutton: No more Webdriver example? This still seems useful, people might not know about axe-webdriverjs.

doc/examples/jest+react/package.json
WilcoFiers: I didn't want to point to a beta or unreleased version unless I had to. The example doesn't have any shadow DOM. We can update this when we release 3.0.

doc/examples/selenium/package.json
WilcoFiers: People not being able to find their way to axe-webdriverjs is a different problem. We shouldn't teach people how to build their own, which is what this example was doing. Just use what's already out there.

PR #748, opened 2018-02-28
docs: Correct minor API.md example

WilcoFiers:
Noticed this problem in one of the docs.


marcysutton:
I looked through the source, and I see both `value` and `values`. We should do a little more work to make sure this is completely accurate before making the change.

WilcoFiers:
Sorted out the TypeScript, and raised a ticket #752 for the options validation.

PR #756, opened 2018-03-02
fix(region): Ignore forms without accessible name as landmarks

### LINKED ISSUE #723: Rule: 'region' - Do not treat <form> without accessible labels as form landmarks
Treat only form elements with non-empty ```aria-labelledby```, ```aria-label``` and ```title``` as form landmarks

WilcoFiers:
Closes #723

### DIFF
--- a/lib/checks/navigation/region.js
+++ b/lib/checks/navigation/region.js
@@ -17,18 +17,28 @@ const implicitLandmarks = landmarkRoles
 	.filter(r => r !== null);
 
 // Check if the current element is the skiplink
-function isSkipLink (node) {
-	return skipLink && skipLink === node;
+function isSkipLink (vNode) {
+	return skipLink && skipLink === vNode.actualNode;
 }
 
 // Check if the current element is a landmark
-function isLandmark (node) {
-	if (node.hasAttribute('role')) {
-		return landmarkRoles.includes(node.getAttribute('role').toLowerCase());
+function isLandmark (virtualNode) {
+	var node = virtualNode.actualNode;
+	var explictRole = (node.getAttribute('role') || '').trim().toLowerCase();
+
+	if (explictRole) {
+		if (explictRole === 'form') {
+			return !!aria.labelVirtual(virtualNode);
+		}
+		return landmarkRoles.includes(explictRole);
 	} else {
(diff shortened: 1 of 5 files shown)


test/checks/navigation/region.js
```diff
@@ -102,6 +102,56 @@ describe('region', function () {
...
+	});
+
+  it('treats forms with aria label as landmarks', function () {
+    var checkArgs = checkSetup('<form id="target" aria-label="foo"><p>This is random content.</p></form>');
+    assert.isTrue(checks.region.evaluate.apply(checkContext, checkArgs));
+  });
+  it('treats forms with aria label as landmarks', function () {
+    var checkArgs = checkSetup('<div role="form" id="target" aria-label="foo"><p>This is random content.</p></form>');
+    assert.isTrue(checks.region.evaluate.apply(checkContext, checkArgs));
+  });
```
isner: Maybe these two tests should be combined so the output (same name for two consecutive tests) doesn't smell of accidental copy-pasta.

WilcoFiers: Whoops! Yep, sounds good. Not sure how that snuck in.

isner: Same for "treats forms with an empty aria label as not a landmarks" below.

lib/checks/navigation/region.js
```diff
@@ -17,18 +17,30 @@ const implicitLandmarks = landmarkRoles
...
-	if (node.hasAttribute('role')) {
-		return landmarkRoles.includes(node.getAttribute('role').toLowerCase());
+function isLandmark (virtualNode) {
+	var node = virtualNode.actualNode;
+	var explictRole = (node.getAttribute('role') || '').trim().toLowerCase();
+
+	if (explictRole) {
+		if (explictRole === 'form') {
+			return !!aria.labelVirtual(virtualNode);
+		} else {
```
isner: Small point, but I'd like to see these "else" blocks removed where they simply return the "happy path":
```js
if (explicitRole === 'form') {
  // edge case
  return !!aria.labelVirtual(virtualNode);
}
// common case
return landmarkRoles.includes(explicitRole);
```

WilcoFiers:
@isner This one needs a thumbs up before I can merge, since it's been updated since your review.

PR #759, opened 2018-03-02
 feat(rules): add new rule aria-dpub-role-fallback 

rdeltour:
# aria-dpub-role-fallback

Ensures that DPUB ARIA roles that are unsupported and may have a negative
impact are only used on elements which have a compatible implicit ARIA
role fallback.

**Tags**: cat.aria, wcag2a, wcag131

## Selector

CSS selector `[role]`
+ JS matcher that matches the elements with the following roles: 'doc-backlink', 'doc-biblioentry', 'doc-biblioref', 'doc-cover', 'doc-endnote', 'doc-glossref', 'doc-noteref'

###  implicit-role-fallback (all)

1. gets the element’s role
2. if there’s no role, or the role is invalid, ignore (return `true`)
3. return `true` if and only if the role’s type is equal to the element’s implicit role (i.e. if the implicit role is a compatible fallback for the explicit role).

This PR implements the rule to complement #584

### DIFF
--- a/lib/rules/aria-dpub-role-fallback-matches.js
+++ b/lib/rules/aria-dpub-role-fallback-matches.js
@@ -1,0 +1,10 @@
+var role = node.getAttribute('role');
+return [
+	'doc-backlink',
+	'doc-biblioentry',
+	'doc-biblioref',
+	'doc-cover',
+	'doc-endnote',
+	'doc-glossref',
+	'doc-noteref'
+].includes(role);
--- a/test/rule-matches/aria-dpub-role-fallback-matches.js
+++ b/test/rule-matches/aria-dpub-role-fallback-matches.js
@@ -1,0 +1,55 @@
+describe('aria-dpub-role-fallback-matches', function () {
+	'use strict';
+
+	var fixture = document.getElementById('fixture');
+	var rule;
+
+	beforeEach(function () {
+		rule = axe._audit.rules.find(function (rule) {
+			return rule.id === 'aria-dpub-role-fallback';
+		});
+	});
+
+	afterEach(function () {
+		fixture.innerHTML = '';
+	});
+
+	it('is a function', function () {
+		assert.isFunction(rule.matches);
+	});
+
+	it('should not match elements with no role', function () {
+		fixture.innerHTML = '<div></div>';
+		var target = fixture.children[0];
+
+		assert.isFalse(rule.matches(target));
+	});
+
+	it('should not match elements with a nonsensical role', function () {
+		fixture.innerHTML = '<div role="yay"></div>';
+		var target = fixture.children[0];
+
+		assert.isFalse(rule.matches(target));
+	});
+
+	it('should not match elements with a non-DPUB role', function () {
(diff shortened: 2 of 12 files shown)


WilcoFiers:
@rdeltour Please rebase with develop. It has some fixes so that it passes the security test. You may also find that Circle is a bit flakey running Chrome. We're aware of that and can just rerun it if we need to.

lib/rules/aria-dpub-role-fallback-matches.js
```diff
@@ -0,0 +1,10 @@
+var role = node.getAttribute('role');
+return [
+		'doc-backlink',
+		'doc-biblioentry',
+		'doc-biblioref',
+		'doc-cover',
+		'doc-endnote',
+		'doc-glossref',
+		'doc-noteref'
+	].some(r => r === role);
```
WilcoFiers: You could do a `.includes` instead. Saved you the callback.

test/rule-matches/aria-dpub-role-fallback-matches.js
```diff
@@ -0,0 +1,97 @@
+describe('aria-dpub-role-fallback-matches', function () {
```
WilcoFiers: Your "role" tests are probably a bit too specific here. This should test that the code works. What this exact list of roles is shouldn't matter as much. The integration tests should solve that for us.

lib/rules/aria-dpub-role-fallback-matches.js
rdeltour: doh, right :-)

test/rule-matches/aria-dpub-role-fallback-matches.js
rdeltour: Mmm ok. Do you want me to simplify the test, or keep it as-is since it’s there already? If the former, do I just pick one role to test a positive match?

rdeltour:
> Please rebase with develop. It has some fixes so that it passes the security test.

I did (locally), but the security test still fails:

```
>> lodash 3.10.1 has known vulnerabilities:  severity: low; summary: Prototype pollution attack; https://hackerone.com/reports/310443
axe-core 3.0.0-beta.2
↳ grunt 1.0.2
 ↳ grunt-legacy-log 1.0.0
  ↳ lodash 3.10.1
>> lodash 3.10.1 has known vulnerabilities:  severity: low; summary: Prototype pollution attack; https://hackerone.com/reports/310443
axe-core 3.0.0-beta.2
↳ grunt-legacy-log 1.0.0
 ↳ lodash 3.10.1
Warning: Task "retire:node" failed. Use --force to continue.
```

test/rule-matches/aria-dpub-role-fallback-matches.js
WilcoFiers: Just pick one of the roles to test it with. In the future I'm hoping we can add rule options so you could configure it that way.

rdeltour: Done!

PR #760, opened 2018-03-03
New rules: No duplicate header/footer

### LINKED ISSUE #682: Extend 'landmark-one-main' rule to fail if a page has more than one banner, or contentinfo landmark
Ensures that a page contain not more than one main, banner and contentinfo landmarks unless in nested document or application role.

If extending this rule does not seem appropriate, it is a good case for a new best practice rule

WilcoFiers:
This PR replaces #732. Additionally, it closes #682.


bitbon57:
682

doc/rule-descriptions.md
```diff
@@ -34,6 +34,8 @@
 | label-title-only | Ensures that every form element is not solely labeled using the title or aria-describedby attributes | cat.forms, best-practice | true |
 | label | Ensures every form element has a label | cat.forms, wcag2a, wcag332, wcag131, section508, section508.22.n | true |
 | landmark-main-is-top-level | The main landmark should not be contained in another landmark | best-practice | true |
+| landmark-no-duplicate-banner | A banner landmark identifies site-oriented content at the beginning of each page within a website | best-practice | true |
```
isner: It looks like most of the rule description read as descriptions of what the rule checks for, not as descriptions of the benefit of a given accessibility feature. We should be consistent. This would become something like, "Ensures the document has no more than one banner landmark."

lib/checks/keyboard/page-no-duplicate.js
```diff
@@ -0,0 +1,18 @@
+if (!options || !options.selector || typeof options.selector !== 'string') {
+	throw new TypeError('visible-in-page requires options.selector to be a string');
+}
+
+let elms = axe.utils.querySelectorAll(virtualNode, options.selector);
+
+// Filter elements that, within certain contexts, don't map there role.
```
isner: ...map _their_ role.

WilcoFiers:
Updated

PR #761, opened 2018-03-03
New rules: contentinfo / banner-is-top-level

WilcoFiers:
These rules are similar to landmark-main-is-top-level, with the small difference that we skip the element if they have no role and they are nested in a article, aside, main, nav or section element.


doc/rule-descriptions.md
```diff
@@ -33,6 +33,8 @@
 | input-image-alt | Ensures &lt;input type=&quot;image&quot;&gt; elements have alternate text | cat.text-alternatives, wcag2a, wcag111, section508, section508.22.a | true |
 | label-title-only | Ensures that every form element is not solely labeled using the title or aria-describedby attributes | cat.forms, best-practice | true |
 | label | Ensures every form element has a label | cat.forms, wcag2a, wcag332, wcag131, section508, section508.22.n | true |
+| landmark-banner-is-top-level | A banner landmark identifies site-oriented content at the beginning of each page within a website | best-practice | true |
```
isner: It looks like most of the rule description read as descriptions of what the rule checks for, not as descriptions of the benefit of a given accessibility feature. We should be consistent. This would become something like, "Ensures a banner landmark is present at the top level of the document."

lib/checks/keyboard/landmark-is-top-level.json
```diff
@@ -0,0 +1,11 @@
+{
+  "id": "landmark-is-top-level",
+  "evaluate": "landmark-is-top-level.js",
+  "metadata": {
+    "impact": "moderate",
```
isner: By making this check more generic, is this impact level still accurate?

lib/rules/landmark-has-body-context.js
```diff
@@ -0,0 +1,6 @@
+const nativeScopeFilter = 'article, aside, main, nav, section';
+
+// Filter elements that, within certain contexts, don't map there role.
```
isner: `...map _their_ role.`

test/checks/keyboard/landmark-is-top-level.js
```diff
@@ -45,26 +52,27 @@ describe('main-is-top-level', function () {
...
-		assert.isTrue(checks['main-is-top-level'].evaluate(mainDiv));
+		assert.isTrue(check.evaluate.call(checkContext, mainDiv));
+		assert.deepEqual(checkContext._data, { role: 'main' });
 	});
 	
 	it('should return true if main is in form landmark', function () {
-		var mainDiv = document.createElement('div');
-		mainDiv.setAttribute('role','main');
+		var bannerDiv = document.createElement('div');
+		bannerDiv.setAttribute('role','banner');
```
isner: It feels like we're asking these existing tests to do double duty by adding in assertions that aren't exactly described by the test case. It might be better to break these out.

lib/checks/keyboard/landmark-is-top-level.json
WilcoFiers: Yes, this should be fine.

doc/rule-descriptions.md
WilcoFiers: Sorted. Weirdly enough the messages of these were quite different from the landmark-main-is-top-level. That one actually had a very nice description, so I just reused that.

WilcoFiers:
Updated, please have another look

test/checks/keyboard/landmark-is-top-level.js
```diff
@@ -45,26 +52,27 @@ describe('main-is-top-level', function () {
 		navDiv.setAttribute('role','navigation');
 		fixture.appendChild(navDiv);
 		fixture.appendChild(mainDiv);
-		assert.isTrue(checks['main-is-top-level'].evaluate(mainDiv));
+		assert.isTrue(check.evaluate.call(checkContext, mainDiv));
+		assert.deepEqual(checkContext._data, { role: 'main' });
 	});
 	
 	it('should return true if main is in form landmark', function () {
```
isner: Some of these test names are now outdated given the nature of the changes therein.

PR #763, opened 2018-03-05
New rule: page-has-heading-one

WilcoFiers:


doc/rule-descriptions.md
```diff
@@ -46,6 +46,7 @@
 | meta-viewport | Ensures &lt;meta name=&quot;viewport&quot;&gt; does not disable text scaling and zooming | cat.sensory-and-visual-cues, wcag2aa, wcag144 | true |
 | object-alt | Ensures &lt;object&gt; elements have alternate text | cat.text-alternatives, wcag2a, wcag111, section508, section508.22.a | true |
 | p-as-heading | Ensure p elements are not used to style headings | cat.semantics, wcag2a, wcag131, experimental | true |
+| page-has-heading-one | Ensure that the page, or at least one of its frames contains an level one heading | best-practice | true |
```
isner: "...contains _a level-one_ heading"

Note both the "an" to "a" change _and_ the added hyphen in "level-one" (it's acting as an adjective here).

lib/checks/keyboard/page-has-elm-after.js
```diff
@@ -0,0 +1,9 @@
+const elmUsedAnywhere = results.some(frameResult => frameResult.result === true);
+
+// If the element exists in any frame, set them all to true
+if (elmUsedAnywhere) {
+	results.forEach(result => {
+		result.result = elmUsedAnywhere;
```
isner: You're setting each frame's result to the value of `elmUsedAnywhere` when in truth, you only want to use this value when it's `true`. I think `result.result = true` would be more declarative.

WilcoFiers:
Updated, please have a second look @isner

PR #764, opened 2018-03-06
New rule: Test that frames have axe in them

### LINKED ISSUE #307: Prompt users to check iframes when they are disabled in API (new rule)
One we've merged #304, we need a way to float up iframes that were skipped to give a more complete picture of a11y on a given page. We could write a rule that puts these items into Needs Review.

From @dylanb:
> Create a rule that puts every iframe into a "needs review" result if there are iframes and `options.iframes === false`

Pros: every user would be able to know/display-to-end-users the fact that there are parts of the page that are not being analyzed

Cons: We would have to pass this option into every check so that it could operate appropriately dependent on the options - this is not a huge con IMO because it would allow all checks to modify their behavior based on options

WilcoFiers:
Closes #307

### DIFF
--- a/lib/checks/media/frame-tested.json
+++ b/lib/checks/media/frame-tested.json
@@ -1,0 +1,15 @@
+{
+  "id": "frame-tested",
+  "evaluate": "frame-tested.js",
+  "options": {
+    "isViolation": false
+  },
+  "metadata": {
+    "impact": "critical",
+    "messages": {
+      "pass": "The iframe was tested with axe-core",
+      "fail": "The iframe could not be tested with axe-core",
+      "incomplete": "The iframe still has to be tested with axe-core"
+    }
+  }
+}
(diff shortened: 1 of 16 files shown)


lib/checks/media/frame-tested.json
```diff
@@ -0,0 +1,15 @@
+{
+  "id": "frame-tested",
+  "evaluate": "frame-tested.js",
+  "options": {
+    "isViolation": false
+  },
+  "metadata": {
+    "impact": "critical",
+    "messages": {
+      "pass": "The iframe can be tested with axe-core",
```
isner: This should probably be "The iframe _was_ tested with axe-core" so as to position it opposite the failure message. As-is, it suggests that it's now safe to test.

test/testutils.js
```diff
@@ -11,6 +11,14 @@ testUtils.MockCheckContext = function () {
 	return {
 		_relatedNodes: [],
 		_data: null,
+		_onAsync: null,
+		async: function () {
```
isner: We should document the purpose of `async` and `_onAsync` props.

PR #768, opened 2018-03-08
Shadow DOM exclude

### LINKED ISSUE #744: API not properly excluding Shadow DOM content
I uncovered a bug when upgrading axe-matchers to axe-core@3x, where it would fail on Shadow DOM content that was supposed to be excluded through the API. I would expect that by excluding a shadow host element, it wouldn't return issues from nodes inside of it. I was able to reproduce it in axe-core, here's the basic fixture:

```html
<div id="container">
    <p id="test-passes">Passing text</p>
    <div id="shadow-host"></div>
</div>
<script>
    var host = document.getElementById('shadow-host');
    var shadowRoot = host.attachShadow({ mode: 'open' });
    shadowRoot.innerHTML =  '<div style="background-color: #333">' +
        '<p>Failing text</p>' +
        '<ul><p>Failing list</p></ul>' +
    '</div>';
</script>
```
When I call axe.run on it with `include: [['#container']], exclude: [['#shadow-host']]`, I get color contrast and list violations for the shadow content. It should exclude those, but it's including shadow content for both rules with a stated target of `#shadow-host`. 

Here's a failing test: https://github.com/dequelabs/axe-core/commit/0ef5bb980d648758ed8f2914b807a2bba33ad5e6

```javascript
(shadowSupported ? it : xit)
('should find no nodes in Shadow DOM', function (done) {
    var config = { runOnly: { type: 'rule', values: ['list', 'color-contrast'] } };
    axe.run({ include: [['#container']], exclude: [['#shadow-host']] }, config, function (err, results) {
        assert.isNull(err);
        assert.lengthOf(results.violations, 0, 'violations');
        assert.lengthOf(results.passes, 2, 'passes');
        done();
    });
});
```

WilcoFiers:
Closes #744

### DIFF
--- a/lib/core/base/audit.js
+++ b/lib/core/base/audit.js
@@ -144,7 +144,6 @@ Audit.prototype.run = function (context, options, resolve, reject) {
 	'use strict';
 	this.validateOptions(options);
 
-	axe._tree = axe.utils.getFlattenedTree(document.documentElement); //cache the flattened tree
 	axe._selectCache = [];
 	var q = axe.utils.queue();
 	this.rules.forEach(function (rule) {
@@ -197,6 +196,10 @@ Audit.prototype.after = function (results, options) {
 
 	return results.map(function (ruleResult) {
 		var rule = axe.utils.findBy(rules, 'id', ruleResult.id);
+		if (!rule) {
+			// If you see this, you're probably running the Mocha tests with the aXe extension installed
+			throw new Error('Result for unknown rule. You may be running mismatch aXe-core versions');
+		}
 
 		return rule.after(ruleResult, options);
 	});
--- a/lib/core/base/context.js
+++ b/lib/core/base/context.js
@@ -75,7 +75,6 @@ function normalizeContext(context) {
 
 	// typeof NodeList.length in PhantomJS === function
 	if (context && typeof context === 'object' || context instanceof NodeList) {
-
 		if (context instanceof Node) {
 			return {
 				include: [context],
@@ -130,7 +129,7 @@ function parseSelectorArray(context, type) {
 			nodeList = Array.from(document.querySelectorAll(item));
 			//eslint no-loop-func:0
 			result = result.concat(nodeList.map((node) => {
-				return axe.utils.getFlattenedTree(node)[0];
+				return axe.utils.getNodeFromTree(axe._tree[0], node);
 			}));
 			break;
 		} else if (item && item.length && !(item instanceof Node)) {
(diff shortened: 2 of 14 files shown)


lib/core/base/audit.js
```diff
@@ -144,7 +144,9 @@ Audit.prototype.run = function (context, options, resolve, reject) {
 	'use strict';
 	this.validateOptions(options);
 
-	axe._tree = axe.utils.getFlattenedTree(document.documentElement); //cache the flattened tree
+	if (!axe._tree) {
```
dylanb: Why conditional? We always want to do this to ensure that the state of the flattened tree is accurate

lib/core/base/context.js
```diff
@@ -201,9 +203,11 @@ function validateContext(context) {
  * @param {Object} spec Configuration or "specification" object
  */
 function Context(spec) {
-	/* eslint max-statements:["error",19], no-unused-vars:0 */
+	/* eslint max-statements:["error",22], no-unused-vars:0 */
 	'use strict';
-	var self = this;
+	if (!axe._tree) {
```
dylanb: Why do you need this here? I'd rather add it to the tests that are failing directly than have potential side effects that are difficult to track down.

lib/core/base/context.js
```diff
@@ -216,9 +220,9 @@ function Context(spec) {
 	this.include = parseSelectorArray(this, 'include');
 	this.exclude = parseSelectorArray(this, 'exclude');
 
-	axe.utils.select('frame, iframe', this).forEach(function (frame) {
-		if (isNodeInContext(frame, self)) {
-			pushUniqueFrame(self.frames, frame.actualNode);
+	axe.utils.select('frame, iframe', this).forEach((frame) => {
```
dylanb: These changes are a refactor and should not be included in this commit

lib/core/base/audit.js
WilcoFiers: Well, because I didn't want to overhaul this thing so close to the release... The right way to do this would've been to have Context keep the tree and not store it on axe._tree. It than could've made sure it was based on elements that have the same ownerDocument, and flatten it instead of whatever is on the global document.... That's just a lot of work. :/ I'll see what I can do.

WilcoFiers:
@dylanb I've updated so that the context is now solely responsible for creating axe._tree. I didn't do the "refactor" commit. I don't believe small changes like that need to be in their own commit. That would really clutter up the commit history IMO. If that was the intent we need to update the commit history policy first - and we need to discuss it because I'd be concerned about the impact such a thing has on productivity.

WilcoFiers:
OK, sun runRules is now responsible for setting and cleaning up of `axe._tree` and `axe._selectorData`, but since Context is the place where we figure out the root node, that's where we build the tree.

dylanb:
@WilcoFiers commit policy discussion should be taken offline...approved

PR #769, opened 2018-03-08
fix: make getSelector work with URIs that cannot be shortened

dylanb:


test/core/utils/get-selector.js
```diff
@@ -551,6 +551,16 @@ describe('axe.utils.getSelector', function () {
 		);
 	});
 
+	it('should work correctly when a URL attribute cannot be shortened', function () {
+		var node = document.createElement('a');
+		node.setAttribute('href', 'mars2.html?a=be_bold');
```
isner: It'd be nice to cache "mars2.html?a=be_bold" so it doesn't have to be repeated throughout the test. It would be less prone to typos.

test/core/utils/get-selector.js
```diff
@@ -551,6 +551,16 @@ describe('axe.utils.getSelector', function () {
...
+		fixture.appendChild(node);
+		axe._tree = axe.utils.getFlattenedTree(document.documentElement);
+
+		assert.isTrue(axe.utils.getSelector(node).indexOf('mars2.html?a=be_bold') !== -1);
```
isner: Suggestion: `assert.notInclude(axe.utils.getSelector(node), 'mars2.html?a=be_bold');`

lib/core/utils/get-selector.js
```diff
@@ -18,20 +18,25 @@ const MAXATTRIBUTELENGTH = 31;
  * @return {String}
```
isner: Whitespace changes in this file are inconsistent with the rest of the file.

test/core/utils/get-selector.js
dylanb: you mean `assert.include`

lib/core/utils/get-selector.js
dylanb: don't we have an eslint rule for that?

test/core/utils/get-selector.js
isner: Oops. Yes!

PR #774, opened 2018-03-12
feat: Make options.runOnly more forgiving about plurality

### LINKED ISSUE #752: fix option validation
The runOnly validation looks at a `value` array, not a `values` array. We should change that.

WilcoFiers:
BREAKING CHANGE: Incorrect use of runOnly now throws errors

Closes #752

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -376,6 +376,14 @@ Additionally, there are a number or properties that allow configuration of diffe
 	}
 	```
 
+  Alternatively, runOnly can be passed an array of tags:
+
+  ```javascript
+  {
+    runOnly: ["wcag2a", "wcag2aa"]
+  }
+  ```
+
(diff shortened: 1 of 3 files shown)


doc/API.md
```diff
@@ -376,6 +376,14 @@ Additionally, there are a number or properties that allow configuration of diffe
 	}
 	```
 
+  Alternatively, runOnly can be passed an array of tags:
```
marcysutton: The indentation here doesn't match the rest of the runOnly documentation.

marcysutton:
This looks good to me, it makes the API less error-prone. Thanks for adding tests!

WilcoFiers:
That API.md file is all over the place in terms of spacing. I've moved everything to tabs now. Hopefully sublime will be less confused about it next time. Please have another look Marcy.

PR #776, opened 2018-03-12
fix(rule): skip-link rule doesn't decode URI encoded href's

skydreamerr:
If href contains url encoded fragment like "<a href="#%3Ctarget%3E">Click Here</a>" it leads to false positive violation due to no target


WilcoFiers:
Looks great. Much appreciated!

PR #781, opened 2018-03-15
doc: Update projects list to be more current

marcysutton:
I moved all of the Deque projects to the top, as well as some of the more notable ones. I also added some that were missing.


WilcoFiers:
@marcysutton Should we add headings to this list to separate supported and unsupported project?

marcysutton:
I've updated it to list "Deque Projects" and "Community Projects". Does that sound ok?

PR #786, opened 2018-03-17
fix: Avoid timing issue with axe cleanup method

WilcoFiers:
Closes https://github.com/dequelabs/react-axe/issues/28

### DIFF
--- a/lib/core/public/load.js
+++ b/lib/core/public/load.js
@@ -18,7 +18,11 @@ function runCommand(data, keepalive, callback) {
 
 	switch (data.command) {
 		case 'rules':
-			return runRules(context, options, resolve, reject);
+			return runRules(context, options, function (results, cleanup) {
+				resolve(results);
+				// Cleanup AFTER resolve so that selectors can be generated
+				cleanup();
+			}, reject);
 		case 'cleanup-plugin':
 			return cleanupPlugins(resolve, reject);
 		default:
--- a/lib/core/public/run.js
+++ b/lib/core/public/run.js
@@ -112,33 +112,35 @@ axe.run = function (context, options, callback) {
 		});
 	}
 
-	axe._runRules(context, options, function (rawResults) {
-		let respond = function (results) {
-			try {
-				callback(null, results);
-			} catch(e) {
-				axe.log(e);
-			}
-			resolve(results);
-		};
-		if (options.performanceTimer) {
-			axe.utils.performanceTimer.end();
-		}
-
-		try {
-			let reporter = axe.getReporter(options.reporter);
-			let results = reporter(rawResults, options, respond);
-			if (results !== undefined) {
-				respond(results);
-			}
-		} catch (err) {
-			callback(err);
-			reject(err);
-		}
-	}, function (err) {
-		callback(err);
-		reject(err);
-	});
-
-	return p;
+	axe._runRules(context, options, function (rawResults, cleanup) {
+ 		let respond = function (results) {
+			cleanup();
+ 			try {
+ 				callback(null, results);
+ 			} catch(e) {
+ 				axe.log(e);
+ 			}
+ 			resolve(results);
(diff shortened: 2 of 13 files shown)


lib/core/public/load.js
```diff
@@ -18,7 +18,10 @@ function runCommand(data, keepalive, callback) {
 
 	switch (data.command) {
 		case 'rules':
-			return runRules(context, options, resolve, reject);
+			return runRules(context, options, function (results, cleanup) {
+				resolve(results);
+				cleanup();
```
stephenmathieson: Should `cleanup` be called before we `resolve()` here? I don't know if `resolve` is synchronous or not.

WilcoFiers: It's synchronous, and it generates selectors so it needs to go before we clean up the selectorData.

stephenmathieson: Cool, OK. Thank you for clarifying.

test/core/public/run.js
```diff
@@ -211,6 +226,24 @@ describe('axe.run', function () {
 				done();
 			});
 		});
+
+		it('is called after cleanup',
+		(!window.Promise) ? undefined :  function (done) {
```
stephenmathieson: This condition is rather difficult to follow. It may be nicer to write this as:

```js
const testWhenPromiseIsSupported = window.Promise ? it : it.skip
testWhenPromiseIsSupported('is called after cleanup', () => {
  // ...
})
```

stephenmathieson: Alternately, we could just add a Promise polyfill :wink:

isner:
It might be out-of-scope, but the top docs for the `runRules` function are out of date - they still refer to a callback instead of resolve/reject.

lib/core/public/run.js
```diff
@@ -112,8 +112,9 @@ axe.run = function (context, options, callback) {
 		});
 	}
 
-	axe._runRules(context, options, function (rawResults) {
+	axe._runRules(context, options, function (rawResults, cleanup) {
 		let respond = function (results) {
+			cleanup();
```
marcysutton: Can you could add a comment about why you had to run `cleanup` before `resolve` here since it differs from above?

test/core/public/run.js
```diff
@@ -155,6 +155,21 @@ describe('axe.run', function () {
 				throw new Error('err');
 			});
 		});
+
+		it('is called after cleanup', function (done) {
```
marcysutton: What is this test doing, exactly?

test/integration/full/rerun/rerun.html
```diff
@@ -0,0 +1,26 @@
+<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML+ARIA 1.0//EN" "http://www.w3.org/WAI/ARIA/schemata/html4-aria-1.dtd">
```
marcysutton: Dit you mean to use a different doctype than the frame doc that gets pulled in?

lib/core/public/run.js
```diff
@@ -112,28 +112,32 @@ axe.run = function (context, options, callback) {
...
 			}
-			resolve(results);
 		};
 		if (options.performanceTimer) {
 			axe.utils.performanceTimer.end();
 		}
 
-		try {
+		let results;
+		// try {
```
stephenmathieson: Why are you commenting this out? Maybe we should remove it if it's not used/needed.

test/core/public/run.js
stephenmathieson: It looks like it's monkey-patching `axe._runRules` and verifying the `cleanup` callback is invoked. Perhaps a more descriptive test name would help?

lib/core/public/run.js
WilcoFiers: My bad, committed something I shouldn't have :/

PR #789, opened 2018-03-20
fix: Overwrite incorrect sri-history entries

isner:
I validated each of the sri-history entries manually and discovered that the following 3 versions have inaccurate entries:

- 2.4.2
- 2.6.1
- 3.0.0-beta.1

I am personally responsible for adding the 2.4.2 and 2.6.1 entries. I'm fairly certain that I used the build/sri-update tool to generate those entries back when I added them, and today I generated the entries by `npm i`ing axe-core and generating the hash using the sri-toolbox module. I don't see why those two methods would produce different results, so I'm baffled. I'm guessing human error.

### DIFF
--- a/sri-history.json
+++ b/sri-history.json
@@ -56,8 +56,8 @@
 		"axe.min.js": "sha256-xVpSddrzxbQVFp90jaOUOQGp3fwCwK4sYxYmkIU3N2c="
 	},
 	"2.4.2": {
-		"axe.js": "sha256-N21EZIlTis66XyG9F0DUSG+makyS31j4/yIBtvLKjHc=",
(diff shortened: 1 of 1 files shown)


isner:
@WilcoFiers Right you are. I've fixed that one too.

marcysutton:
I wonder if it's missing an automated build step so `npm build` will run before `npm run release` gets fully executed. `npm build` gets triggered for `npm publish` because of a similar manual building requirement that tripped us up when releasing.

WilcoFiers:
@marcysutton They added a bunch of lifecycle hooks to standard-version as of 4.1.0. That should help with this. `node build/sri-update` should run in the postbump hook.

Issue #771, opened 2018-03-09
Table caption and summary equivalence should be case insensitive


dylanb:
This table should fail

```
<!doctype html>
<html lang="en">
	<head>
		<title>Table Summary and Caption</title>
	</head>
	<body>
		<table summary="My Table">
			<caption> my table </caption>
			<thead>
				<tr><th scope="col">Head</th></tr>
			</thead>
			<tbody>
				<tr><td>Data</td></tr>
			</tbody>
		</table>
	</body>
</html>
```

PR #791, opened 2018-03-20
fix: allow mixed casing of caption/summary

marcysutton:
Closes https://github.com/dequelabs/axe-core/issues/771

### DIFF
--- a/lib/checks/tables/same-caption-summary.js
+++ b/lib/checks/tables/same-caption-summary.js
@@ -1,2 +1,4 @@
-// passing node.caption to accessibleText instead of using the logic in accessibleTextVirtual on virtualNode
-return !!(node.summary && node.caption) && node.summary === axe.commons.text.accessibleText(node.caption);
+// passing node.caption to accessibleText instead of using
+// the logic in accessibleTextVirtual on virtualNode
+return !!(node.summary && node.caption) &&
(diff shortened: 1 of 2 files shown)


lib/checks/tables/same-caption-summary.js
```diff
@@ -1,2 +1,4 @@
-// passing node.caption to accessibleText instead of using the logic in accessibleTextVirtual on virtualNode
-return !!(node.summary && node.caption) && node.summary === axe.commons.text.accessibleText(node.caption);
+// passing node.caption to accessibleText instead of using
+// the logic in accessibleTextVirtual on virtualNode
+return !!(node.summary && node.caption) &&
```
isner: This goes beyond the fix your adding, but wanted to put it out there for consideration because I see this short-circuit syntax used in many of the checks. I'd suggest refactoring the short-circuit syntax for a few reasons:

- It more difficult to read
- It doesn't provide any performance benefit
- It is more likely cause unnoticed TypeError conditions in the future as we modify it

```
if (!(node.summary && node.caption)) {
  return false;
}
return node.summary.toLowerCase() === axe.commons.text.accessibleText(node.caption).toLowerCase();
```
It's not a big deal in this particular check because the statement is fairly straight-forward, but in others where the conditions are more complex, the pitfalls are exacerbated.

Issue #242, opened 2016-10-20
Hidden form label: failure message not clearly identifying the problem


ithinkihaveacat:
axe incorrectly reports that an element within the following HTML does not have a label (when embedded within a larger HTML file):

``` html
        <form class="search" action="/" data-gp-element="search">
          <label class="gp-offline-disable" for="s">Search</label>
          <input class="gp-offline-disable" type="search" name="s" id="s" placeholder="Search" value="">
          <button type="submit">Go</button>
        </form>
```

The input is correctly detected if the entire HTML file is as above; it's only a problem if embedded within a larger file.

To reproduce, first load https://cdn.rawgit.com/ithinkihaveacat/8577d6d00eb08f48b4a6e3fd015c379d/raw/f943193a01bff03ea7df37b9bf4ef206d9f2486c/index.html, then paste the following in the JS console to run the accessibility tests:

``` js
function loadScript(src) {
  return new Promise(resolve => {
    const e = document.createElement("script");
    e.src = src;
    e.onload = resolve; 
    document.head.appendChild(e);
  });
}

loadScript("https://unpkg.com/axe-core/axe.min.js").then(() => {;
  axe.a11yCheck(document, r => console.log(r.violations[2]));
});
```

Actual output:

A `Form elements must have labels` error on the element `<input class="gp-offline-disable" type="search" name="s" id="s" placeholder="Search" value="">`.

Expected output:

No error report (on this element; there are other accessibility errors in the HTML).

paulirish:
Aye. this was discovered downstream via Lighthouse. 

Looking at the payload, it's not clear which of the label checks this is failing with:
![image](https://cloud.githubusercontent.com/assets/39191/19567034/35076dea-96a1-11e6-88be-5ef813d182de.png)

I'm seeing empty `relatedNodes` arrays for all 5 of the label checks.

trav-is:
aXe is correctly calling this out due to display: none;

ithinkihaveacat:
Interesting, yes I can confirm setting display: block eliminates the warning, thanks!

```js
function loadScript(src) {
  return new Promise(resolve => {
    const e = document.createElement("script");
    e.src = src;
    e.onload = resolve; 
    document.head.appendChild(e);
  });
}

document.querySelector("*[for='s']").style.display = "block";

loadScript("https://unpkg.com/axe-core/axe.min.js").then(() => {;
  axe.a11yCheck(document, console.log.bind(console));
});
```

[...]

PR #792, opened 2018-03-20
fix(rule): improve messaging for hidden labels

marcysutton:
As suggested by Dylan in #242, I made a new check for hidden explicit labels to provide more clear messaging about what to fix. Let me know if there are ways to improve it.

Closes https://github.com/dequelabs/axe-core/issues/242


marcysutton:
It seems like there could be more than one templating opportunity here–we should discuss all the options, as it would be require work than just this one check.

marcysutton:
@WilcoFiers you never answered my question yesterday–should we merge this and change it later to fix the issue for Lighthouse, or wait until we address templating? There are multiple checks in this rule, so templating would require handling of more than just this one.

Issue #526, opened 2017-09-15
axe-core / lib / rules / checkboxgroup.json


devarshipant:
Remove duplicate "that" in the description:
"description": "Ensures related <input type=\"checkbox\"> elements have a group and **that** that group designation is consistent"

marcysutton:
We'd happily welcome a PR :)

devarshipant:
Sure! :)

Issue #561, opened 2017-10-06
TH in layout table when table has role=presentation


spanchang:
Gokhan Boybek reported on  10/06/17 that
Assure/Fireeyes is reporting a failure when a layout table contains `<th>` elements. Ok this is clearly a failure of 1.3.1 (https://www.w3.org/TR/WCAG20-TECHS/F46.html)
But it is called out as a failure even when the table contains role="presentation". Isn't that a false positive? Because the aria specs (see at http://www.w3.org/TR/wai-aria-practices-1.1/#presentation_role) clearly indicates that if role="presentation" is applied to the <table> tag, the descendants (including `<th>`) inherit this role as well.
So the client is questioning why Fireeyes is reporting this as a defect when the table tag had role="presentation", which is inherited by the th tag.

Sailesh: I agree, this is not a failure and should not be reported.

WilcoFiers:
That would make sense. Can you test with VoiceOver, NVDA and JAWS to make sure tables with `role=presentation` or `role=none` get ignored even if they have `th` / `caption` / `scope=""` on them?

spanchang:
Those roles have been working fine for a while now to hide the table
semantics even of a  table that is structured like a data table, say 2
col x7 rows.
The caption and TH cells are read normally as one arrows through but
one cannot navigate  the content using table nav keys. The TH cell
content still appears bold.
But this is not the content author's problems and should not be
flagged if those roles are present.
The browser has to be fixed so that TH cell is not rendered bold for
layout tables / table with those roles.
Best regards,


On 10/8/17, Wilco Fiers <notifications@github.com> wrote:
> That would make sense. Can you test with VoiceOver, NVDA and JAWS to make
> sure tables with `role=presentation` or `role=none` get ignored even if they
> have `th` / `caption` / `scope=""` on them?
>
> --
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/dequelabs/axe-core/issues/561#issuecomment-335008407


-- 
Sailesh Panchang
Principal Accessibility Consultant
Deque Systems Inc
Phone 703-225-0380 ext 105
Mobile: 571-344-1765

marcysutton:
Just to confirm the rule that failed, was it `th-has-data-cells`?

marcysutton:
It was `layout-table` that failed. That scenario was already fixed with https://github.com/dequelabs/axe-core/pull/828, so I'm going to close this.

PR #793, opened 2018-03-20
Check typo & data-table matches tests

marcysutton:
There were two issues addressed here, but one was already fixed (#561). The other is a single word change in a check description.

Closes https://github.com/dequelabs/axe-core/issues/526

### DIFF
--- a/doc/rule-descriptions.md
+++ b/doc/rule-descriptions.md
@@ -15,7 +15,7 @@
 | blink | Ensures &lt;blink&gt; elements are not used | cat.time-and-media, wcag2a, wcag222, section508, section508.22.j | true |
 | button-name | Ensures buttons have discernible text | cat.name-role-value, wcag2a, wcag412, section508, section508.22.a | true |
 | bypass | Ensures each page has at least one mechanism for a user to bypass navigation and jump straight to the content | cat.keyboard, wcag2a, wcag241, section508, section508.22.o | true |
-| checkboxgroup | Ensures related &lt;input type=&quot;checkbox&quot;&gt; elements have a group and that that group designation is consistent | cat.forms, best-practice | true |
+| checkboxgroup | Ensures related &lt;input type=&quot;checkbox&quot;&gt; elements have a group and that the group designation is consistent | cat.forms, best-practice | true |
 | color-contrast | Ensures the contrast between foreground and background colors meets WCAG 2 AA contrast ratio thresholds | cat.color, wcag2aa, wcag143 | true |
 | definition-list | Ensures &lt;dl&gt; elements are structured correctly | cat.structure, wcag2a, wcag131 | true |
 | dlitem | Ensures &lt;dt&gt; and &lt;dd&gt; elements are contained by a &lt;dl&gt; | cat.structure, wcag2a, wcag131 | true |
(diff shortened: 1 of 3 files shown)


lib/checks/tables/th-has-data-cells.js
```diff
@@ -2,6 +2,11 @@ var tableUtils = axe.commons.table;
 var cells = tableUtils.getAllCells(node);
 var checkResult = this;
 
+var exemptRoles = ['presentation', 'none'];
+if (node.hasAttribute('role') && exemptRoles.includes(node.getAttribute('role'))) {
+	return true;
```
WilcoFiers: These tables should be ignored, not passed. I suggest putting this in the selector or matcher instead.

doc/rule-descriptions.md
```diff
@@ -15,7 +15,7 @@
 | blink | Ensures &lt;blink&gt; elements are not used | cat.time-and-media, wcag2a, wcag222, section508, section508.22.j | true |
 | button-name | Ensures buttons have discernible text | cat.name-role-value, wcag2a, wcag412, section508, section508.22.a | true |
 | bypass | Ensures each page has at least one mechanism for a user to bypass navigation and jump straight to the content | cat.keyboard, wcag2a, wcag241, section508, section508.22.o | true |
-| checkboxgroup | Ensures related &lt;input type=&quot;checkbox&quot;&gt; elements have a group and that that group designation is consistent | cat.forms, best-practice | true |
+| checkboxgroup | Ensures related &lt;input type=&quot;checkbox&quot;&gt; elements have a group and that group designation is consistent | cat.forms, best-practice | true |
```
dylanb: Actually, this is not a typo, perhaps you could change it to

```
Ensures related &lt;input type=&quot;checkbox&quot;&gt; elements have a group and that the group designation is consistent
```

dylanb: Clarification: there are two sentences here, individually, they could read

Ensures related input elements have a group and
Ensures that the group designation is consistent

lib/checks/tables/th-has-data-cells.js
dylanb: agree

marcysutton: Ok, I've updated it. Thanks!

marcysutton:
This PR has been updated with your feedback and is ready for another review.

test/rule-matches/data-table-matches.js
```diff
@@ -0,0 +1,38 @@
+describe('data-table-matches', function () {
```
WilcoFiers: This test should have a "should return false" case as well.

marcysutton: Added, and I fixed the typos in the first two test names.

lib/rules/data-table-matches.js
```diff
@@ -1 +1,5 @@
+var exemptRoles = ['presentation', 'none'];
+if (node.hasAttribute('role') && exemptRoles.includes(node.getAttribute('role'))) {
+	return false;
+}
 return axe.commons.table.isDataTable(node);
```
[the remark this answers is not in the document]
marcysutton: That commons function also checks if the table is focusable, does that still apply?
```js
// is-data-table.js
// The element is not focusable and has role=presentation
if ((role === 'presentation' || role === 'none') && !dom.isFocusable(node)) {
    return false;
}
```

marcysutton:
I kept the data-table-matches tests but the underlying issue in #561 was already fixed. This is now a very minor PR and is ready to be merged.

Issue #765, opened 2018-03-06
Update fix message for rule "button-name" on <button> elements


iamrafan:
button-name rule passes for ```<button>``` element having only non-empty title attribute as an accessible label

Add the following message to ```Fix at least one (1) of these issues:``` list 
> Element has a title attribute and the title attribute is empty

Add the following message in ```How to fix the problem``` list  for <button> in the help page (https://dequeuniversity.com/rules/axe/2.6/button-name)
> Non-empty title attribute

WilcoFiers:
I'm not sure I fully understand. Are you asking to allow titles as a way to pass button-name?

iamrafan:
'button-name' rule already considers non-empty title attribute as an acc name for <button> elements. 
But when a <button> fails button-name rule, the fix messages shown in the screenshot (below)_s do not include the ones I mentioned above
![image](https://user-images.githubusercontent.com/22068434/37115971-9c10a190-2201-11e8-9845-7faa7039d258.png)

![image](https://user-images.githubusercontent.com/22068434/37115988-ae1d8dda-2201-11e8-92f5-9a3bf66d1022.png)

WilcoFiers:
Gotcha, ok so this is related to #747 and #767. The `has-visible-text` check doesn't do what its name implies it does, e.g. check that an element contains text. It computes the accessible name instead, which is why we're not testing for title separately.

iamrafan:
I am saying the current tests are just fine but just that users don't know one of the possible fixes for button-name failure is to use non-empty title. If we update the fix messages in 'how to fix' and the help page, users will know this fix.

Bdw, it is not related to both #747 and #767 

this is related to the ```<button>```  elements, the following example
```
<button class="toggle-button" title="Collapse" aria-expanded="true" style="left: 182px;"><span class="bowtie-icon bowtie-chevron-left"></span></button>
```

Issue #747, opened 2018-02-28
false positive: input type="button" with @title only


jasonkiss:
`<input type="button" title="Search">` currently fails the "Buttons must have discernible text" test, but the `title` attribute is a permitted source for programmatically calculating the element's accessible name, as per the [HTML-AAM](https://w3c.github.io/html-aam/#input-type-button-input-type-submit-and-input-type-reset).

marcysutton:
I wonder if it was written that way for historical reasons, since `title` wasn't always exposed. In any case, we have a similar issue open for messaging (#765) and will look into a fix.

PR #794, opened 2018-03-20
fix: allow title on button-name

marcysutton:
This change makes it more clear that `title` is allowed for accessible naming of button elements and button inputs.

Closes https://github.com/dequelabs/axe-core/issues/747
Closes https://github.com/dequelabs/axe-core/issues/765

### DIFF
--- a/lib/rules/button-name.json
+++ b/lib/rules/button-name.json
@@ -20,7 +20,8 @@
     "aria-label",
     "aria-labelledby",
     "role-presentation",
-    "role-none"
+    "role-none",
+    "non-empty-title"
   ],
   "none": [
     "focusable-no-name"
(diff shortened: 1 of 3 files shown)


iamrafan:
@marcysutton ```title``` is already being allowed as an accessible name for ```<button>``` elements. #765 is more about the fix messages conveying that only title can be used as an accessible label when ```button-name``` rule fails on ```<button>``` elements for not having any accessible label

marcysutton:
The change in this PR explicitly adds the `non-empty-title` check so it will expose `title` as a method to fix. Here is the failureSummary with this change, you can see the new bit at the end:

```"Fix all of the following:
  Element is in tab order and does not have accessible text

Fix any of the following:
  Element has a value attribute and the value attribute is empty
  Element has no value attribute or the value attribute is empty
  Element does not have inner text that is visible to screen readers
  aria-label attribute does not exist or is empty
  aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty or not visible
  Element's default semantics were not overridden with role="presentation"
  Element's default semantics were not overridden with role="none"
  Element has no title attribute or the title attribute is empty"```

iamrafan:
sure, thanks for the clarification.

Issue #754, opened 2018-03-01
False negative: 'label' rule should fail if a <select> element has an empty implicit label


iamrafan:
Example:
```
<label>
    <span class="label"></span>			
    <select>
        <option value="1" selected="selected">Please choose a region</option>
        <option value="2">Coastal</option>
        <option value="3">Forest</option>
        <option value="4">Grasslands</option>
        <option value="5">Mountains</option>
    </select>
</label>
```

dylanb:
@iamrafan This is definitely an issue. Is it ok if we fix this in the release after 3.0.0?

iamrafan:
@dylanb Sure

marcysutton:
The accessibleTextVirtual function is counting the select's text as the label–"Please choose a region" is announced in Voiceover, and likely other ATs. But that doesn't help if the first option is "Desert" instead. Working on a fix.

PR #820, opened 2018-04-03
fix(rule): disallow select options for accessible naming

marcysutton:
This change impacts our interpretation of the accessible name computation algorithm, specifically disallowing `select` labels from steps 2E and 2F. Select option values are still used when referenced by `aria-labelledby`.

Closes https://github.com/dequelabs/axe-core/issues/754


marcysutton:
Here's a question: should select values be used for `aria-labelledby`? Because that's the case that's failing in this PR, which disallowed select content from being used as an implicit label. It seems like the same logic could extend to `aria-labelledby`.

From `test/commons/text/accessible-text.js`:
```
it('should use handle nested selects properly in labelledby context', function() {
		fixture.innerHTML = '<div id="t2label">This is <select multiple ' +
			'aria-labelledby="t1label" aria-label="ARIA Label" id="t1">' +
			'<option selected>first</option><option>second</option><option selected>third</option>' +
			'</select> of <i>everything</i></div>' +
			'<div id="t1label">This is a <b>label</b></div>' +
			'<label for="t1">HTML Label</label>' +
			'<input type="text" id="t2" aria-labelledby="t2label">';
		axe._tree = axe.utils.getFlattenedTree(fixture);

		var target = axe.utils.querySelectorAll(axe._tree, '#t2')[0];
		assert.equal(axe.commons.text.accessibleTextVirtual(target), 'This is first third of everything');
	});
```

WilcoFiers:
I'm not sure I understand your question here Marcy. Can you elaborate?

dylanb:
@marcysutton no - `aria-labelledby` can point to anything as long as that thing has a text value

marcysutton:
Ok, this change ended up being pretty simple. I added a few more test cases including the markup Dylan provided yesterday. It should be good for another review.

test/integration/rules/label/label.html
```diff
@@ -34,4 +35,16 @@
 
 	<label><label><input type="text" id="fail22"></label></label>
 	<label for="fail23">Hi</label><label for="fail23">Foo</label><input type="text" id="fail23">
+
+	<div>
+		<label for="fail24">My Select</label>
+		<label for="pass16">
+			<select id="fail24">
```
dylanb: This needs to pass, not fail - the test here is to ensure that the actual calculated label is correct and equals "My Select" and does not include "Chosen"

test/checks/label/implicit.js
```diff
@@ -39,4 +41,17 @@ describe('implicit-label', function () {
...
+	        '<option value="1" selected="selected">Please choose a region</option>' +
+	        '<option value="2">Coastal</option>' +
+	        '<option value="3">Forest</option>' +
+	        '<option value="4">Grasslands</option>' +
+	        '<option value="5">Mountains</option>' +
+	    '</select>' +
+		'</label>');
+		assert.isFalse(checks['implicit-label'].evaluate.apply(checkContext, params));
+	})
 });
```
dylanb: I think you should add the other cases from my markup as unit tests. I think you will find that some of them will fail (see comment below)

test/integration/rules/label/label.html
marcysutton: The IDs might be confusing things, but I'll take another look

test/checks/label/implicit.js
marcysutton: There are already tests that assert nested cases and references...but I'll look again.

marcysutton: I added the two unique test cases–they still pass.

test/integration/rules/label/label.html
marcysutton: @dylanb are you saying that an input explicitly labeled by an empty label that wraps a select, which in turn is explicitly labeled by a text string should pass? I'd be really surprised if that worked everywhere.
- explicitly labeled input pointing to empty label
- empty label wraps select with options
- select is explicitly labeled with "My select"
- input is supposed to pick up "My select" ?

The other cases work fine.

dylanb: It works fine on VO on OS X and I have no reason to doubt that it works elsewhere but you are welcome to test. VO does slightly botch it by reading too much, but it does read the bit that it is supposed to read - which is the key and is specified in the ARIA accessible name calculation algorithm.

marcysutton: Ok, I think my confusion was whether native labels should have the same behavior as aria-labelledby (use select values as a label by for/id or aria-labelledby reference). I'll do some testing to find out.

marcysutton: Note: it does not work the same for native labels everywhere as it does for aria-labelledby. See this demo, which I tested in VO on Safari and Chrome, NVDA and Firefox, NVDA and IE, and JAWS / IE. The native version was not labeled in VO or in a version of NVDA prior to the latest (I updated NVDA today and checked before and after). In the name of "accessibility supported", I don't think we should count on it being always available for users even if the spec says it should work that way. https://codepen.io/marcysutton/pen/MrapyQ

marcysutton:
I've added support for `aria-labelledby` pointing to select options while still failing empty implicit labels for selects (cases `#pass16` and `fail24` below). 

```
<label>
    <select id="fail24">
        <option selected="selected">Chosen</option>
        <option>Not Selected</option>
    </select>
</label>
<input aria-labelledby="fail24" type="text" id="pass16" />
```

However, the below series of intertwined combinations has varied results depending on the browser and AT combo.  I'm not quite sure how to move forward with this, since we're aiming to recommend techniques that work everywhere. One thing I've observed is that wrapping a select with an empty label messes with the explicit pairing for the select (`<label for="pass17">My select</label>` and `<label for="fail27" />`). Maybe we should skip scenarios where multiple labels are defined, since each browser seems to treat them differently? (Also, see my above comment about native input label by reference vs. aria-labelledby, which also has varied support. You can see this happening below in the AT results)
```
    <label for="pass17">My Select</label>
    <label for="fail27">
        <select id="pass17">
            <option selected="selected">Chosen</option>
            <option>Not Selected</option>
        </select>
    </label>
    <input aria-labelledby="pass17" type="text" id="pass18" />
    <input type="text" id="fail27" />
```
## Voiceover/Safari
- `select#pass17`: announced as "Chosen, My select, pop up button"
- `input[aria-labelledby]#pass18`: announced as "Chosen, my select, edit text"
- `input#fail17`: announced as "Chosen, edit text"

## Voiceover/Chrome
- `select#pass17`: announced as "Chosen, menu item"
- `input[aria-labelledby]#pass18`: announced as "Chosen, edit text"
- `input#fail17`: announced as "Chosen, my select, edit text"

## iOS Voiceover/Safari
- `select#pass17`: announced as "My select, chosen, pop up button"
- `input[aria-labelledby]#pass18`: announced as "Text field"
- `input#fail17`: announced as "Chosen, text field"

## JAWS 17 / IE11
- `select#pass17`: announced as "My select, combobox, chosen"
- `input[aria-labelledby]#pass18`: announced as "Chosen, not selected, edit"
- `input#fail17`: announced as "Chosen, not selected, edit"

## NVDA 2018 / Firefox 59
- `select#pass17`: announced as "My select, combobox, chosen collapsed"
- `input[aria-labelledby]#pass18`: announced as "Chosen, edit has autocomplete"
- `input#fail17`: announced as "Chosen, edit has autocomplete, blank"

The only things I didn't check were JAWS 18 and Android Talkback. But I think based on these results you can see that it's really not consistent at all. Is it possible to isolate this PR on the original case and not get too much further into the weeds here?

marcysutton:
The tests appear to be failing on this, which should be resolved with Wilco's PR:
```
Warning: TypeError: null is not a function (evaluating 'self._onAsync(result, self)')
    at undefined (http://localhost:9876/test/testutils.js:21)
    at undefined (http://localhost:9876/tmp/rules.js:1919) Use --force to continue.
```

marcysutton:
@WilcoFiers I moved the tests into accessibleTextVirtual as we discussed.

dylanb:
Your `#fail27` case is supposed to pass as it is being read correctly by all AT AND it is marked up correctly

Your `#pass18` case is not working on iOS VO but it is marked up correctly - hence we should report this as a bug to Apple and if they accept it, we should let it pass - otherwise we should fail this case.

Your `#pass17` case is failing on Chrome Voiceover and it is marked up correctly - their is an accessibility supported combination that works for OS X, so we should report the bug to Google and let this case pass.

marcysutton:
@dylanb where did you see `#fail27` in the code?  I can file bugs for the other two.

marcysutton:
That example didn't get implemented, because it was super complicated and not supported in most places I checked.

marcysutton:
I noticed one error in my results above–the items marked with `fail17` were actually referencing `fail27`. I made a change in the above comment to reflect that. 

I'm pretty sure the results in Voiceover and Safari changed when I updated OSX...so support for this is really all over the place. I'm finding different results in Safari than what I captured above 20 days ago.

Moving forward, we need a more standard way of testing this stuff. If QA could help establish a standard process including Talkback and multiple desktop screen reader versions, that would be great. We have to make some tough decisions to make about whether to let this kind of stuff pass when it only works in the latest ATs...I saw changes when compiling these results using both NVDA 2017 and 2018, and Voiceover after an update to OSX. I'm still not sure how to move forward.

WilcoFiers:
I've done a little more testing. using for/id to label an input element with a selected option doesn't work. I think we should fail this.

dylanb:
@WilcoFiers @marcysutton can you create a new set of examples showing how you think the code should behave so I can do a review efficiently?

marcysutton:
There are examples in this Codepen: https://codepen.io/marcysutton/pen/jxwOaB?editors=1000

I still feel the code in this PR is sufficient. In general the aria-labelledby cases work okay, but native input labeling by reference to a select does not. This PR already covers those two cases.

dylanb:
Tough one - well done!

Issue #698, opened 2018-01-25
New rule: Flag if landmarks on a page do not have a unique role or role+label combination


iamrafan:
# landmark-unique

**Description:** This rule checks to make sure landmarks reused on a page have unique labels

### Rule Help
Ensures that landmarks have a unique role or role+label combination

### Tags 
wcag131, wcag 241

## Selector

```<header>, <aside>, <footer>, <form>, <main>, <nav>, <section>, [role="banner"], [role="complementary"], [role="contentinfo"], [role="form"], [role="main"], [role="navigation"], [role="region"], [role="search"]```

### ( none )

1. Check that none of the landmarks have same role + label combination

WilcoFiers:
Excellent idea, although I do not think this is a WCAG requirement, so I suggest you use `best-practice` as the only tag.

We should probably exclude main, banner and contentinfo from this list, since we already have / will soon have rules that test they are unique. The remediation suggestion for those should only ever be that they remove the extra ones, adding an aria-label to those (which is what this rule will be suggesting) doesn't actually fix the problem.

iamrafan:
Sure and yes, it makes sense to exclude ```main```, ```banner``` and ```contentinfo``` from this rule

WilcoFiers:
Another one that needs special attention is forms. Forms aren't considered landmarks unless they have an accessible name, so you could have multiple `<form>`, but not multiple `<form aria-label="Contact">`.

iamrafan:
Sure, the same holds true when using ```<section>``` element without having an accessible label.

iamrafan:
#866

WilcoFiers:
Closing as a duplicate.

PR #821, opened 2018-04-03
feat(landmark-accessible-name-nav): add rule ensuring unique nav labels

ssanaul:
Add rule ensuring unique nav labels


WilcoFiers:
@ssanaul Can you put together an issue following the [rule proposal format](https://github.com/dequelabs/axe-core/blob/develop/doc/rule-proposal.md)? I want to discuss this in a little more detail before we review this. I think the scope of this should be broader - maybe have this be true for all landmarks?

marcysutton:
This sounds a lot like https://github.com/dequelabs/axe-core/issues/698

WilcoFiers:
I'm going to close this while we discuss the proposal for #866. We can create a new PR once we figure out what we want from this rule.

PR #822, opened 2018-04-03
Typings: Fix ElementContext type

### LINKED ISSUE #819: Typings: ElementContext seems to have a wrong type
From the documentation, it seems that axe.run function can take a Node, selector or an include-exclude object https://github.com/dequelabs/axe-core/blob/develop/doc/API.md#context-parameter

However, typings expect an object containing a property node, or property selector or an include-exclude object:
```ts
interface ElementContext {
	node?: Object,
	selector?: string,
	include?: any[],
	exclude?: any[]
}
```
I think the correct type of ElementContext should be like that:
```ts
type ElementContext = Node | string | { include?: any[], exclude?: any[]}
```

Going to create a PR with this fix

dannsam:
fixes #819

### DIFF
--- a/axe.d.ts
+++ b/axe.d.ts
@@ -13,8 +13,8 @@ declare module axe {
 	type RunOnlyType = "rule" | "rules" | "tag" | "tags";
 
 	type RunOnlyObject = {
-		include?: string[],
-		exclude?: string[]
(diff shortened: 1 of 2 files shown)


WilcoFiers:
Awesome, thanks for the contribution @dannsam

Issue #818, opened 2018-04-01
Conflict between lighthouse and aXe


dylanb:
Seems as though the attempt to filter out messages from different applications is not working correctly

https://github.com/GoogleChrome/lighthouse/issues/4857

dylanb:
The offending line of code https://github.com/dequelabs/axe-core/blob/develop/lib/core/utils/respondable.js#L15

It should read

```
if (typeof axe !== 'undefined' && axe._audit && axe._audit.application) {
```

PR #825, opened 2018-04-04
Use application name in respondable

isner:
The most important change here is respondable.js [line 15](https://github.com/dequelabs/axe-core/pull/825/files#diff-addc6dbd1628709793c7399662a26bfeR15), in response to https://github.com/dequelabs/axe-core/issues/818. We want the source comparison string to use the application name when it exists.

Since the Audit's `application` property is being hard-coded [here](https://github.com/dequelabs/axe-core/blob/develop/lib/core/base/audit.js#L42) as "axeAPI", this also changes the fallback to match so we're using the same application name everywhere. In practice, even though we had set `application` to "axeAPI", almost all message responses were identified as coming from "axe" because the fallback was being applied.

### DIFF
--- a/lib/core/utils/respondable.js
+++ b/lib/core/utils/respondable.js
@@ -12,7 +12,7 @@
 	 */
 	function _getSource() {
 		var application = 'axe', version = '', src;
-		if (typeof axe !== 'undefined' && axe._audit && !axe._audit.application) {
+		if (typeof axe !== 'undefined' && axe._audit && axe._audit.application) {
 			application = axe._audit.application;
 		}
 		if (typeof axe !== 'undefined') {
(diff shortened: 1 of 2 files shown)


marcysutton:
@isner what was the reason for changing to `axeAPI`? Was it a conflict with the `axe` global?

isner:
@marcysutton There is no functional reason - it's purely for consistency's sake. Recall that we explicitly set `axe._audit.application` to "axeAPI" [here](https://github.com/dequelabs/axe-core/blob/develop/lib/core/base/audit.js#L42). Assume a scenario in which an audit has not yet been created and respondable messaging occurs. Today, `_getSource` will identify these as coming from "axe" until an audit is created, at which time the same messaging system will identify future messages as coming from "axeAPI". My feeling is that there's no reason this shift in application name should occur, or really _any_ shift in application name, until an API user explicitly sets branding/application with an `axe.configure` call.

Issue #812, opened 2018-03-30
False positive: 'layout-table' rule failure on tables with role=presentation


iamrafan:
'layout-table' rule fails for the following cases
```
    <table id="table-1" role="presentation">
        <thead>
            <tr>
                <th>
                    Name2
                </th>
                <th>
                    Type2
                </th>
            </tr>
        </thead>
    </table>

    <table id="table-2" role="presentation">
        <caption>Monthly savings caption2</caption>
        <tr>
            <td>
                <a href="#">t2 s1</a>
            </td>
            <td>
                <a href="#">t2 s2</a>
            </td>
        </tr>
        <tr>
            <td>January</td>
            <td>$100</td>
        </tr>
    </table>

    <table id="table-3" role="presentation" summary="SECOND- Monthly savings">
        <tr>
            <td>
                <a href="#">t2 s1</a>
            </td>
            <td>
                <a href="#">t2 s2</a>
            </td>
        </tr>
        <tr>
            <td>J1</td>
            <td>k1</td>
        </tr>
    </table


```

c.c @waabid

WilcoFiers:
This is a duplicate of https://github.com/dequelabs/axe-core/issues/561. It needs to be tested to make sure that this doesn't actually trip up any AT in the baseline.

iamrafan:
I tested these cases with Voiceover/OS X, JAWS/IE, NVDA/FF and Narrator/Edge. None of the combinations announce header, caption or summary used here

WilcoFiers:
Cool, in that case lets change axe to allow it :)

dylanb:
Suggest changing this to a best practice

WilcoFiers:
I don't agree with that. The layout-table rule has an unfortunate name, but it genuinely tests for WCAG violations, e.g. not using `<th>` / `<caption>` on layout tables. Using `role="presentation"` makes those elements inert, but without it, that would be a violation.

dylanb:
I am suggesting making it a best practice violation if there is a summary for example with role="presentation"

WilcoFiers:
Ahh, gotcha. Maybe. I'd say that is less important though.

jkodu:
@dylanb @WilcoFiers 
I will be picking this up.

jkodu:
Microsoft is going to work/ resolve this issue.

marcysutton:
This should be fixed with #828.

PR #828, opened 2018-04-04
fix(layout-table): check for presentation, none roles and isfocusable for matches

waabid:
#812


WilcoFiers:
Should we have this use the `role-presentation` and `role-none` checks instead?

waabid:
@WilcoFiers I believe we do not want to run the rule on them in the case they have those roles; using those checks would still end up resulting in evaluating the current logic of the rule which is not applicable in these cases.

@marcysutton I've made the changes for more truthy tests and focusable elements: let me know what you think.

WilcoFiers:
@waabid Fair enough :)

Issue #372, opened 2017-06-19
accessible name calculation does not correctly handle empty aria-labelledby


dylanb:
As described here

https://www.w3.org/TR/accname-aam-1.1/#mapping_additional_nd_te

An empty aria-labelledby attribute should be ignored, yet this code

`<button aria-labelledby="">Button</button>`

is raising a violation

PR #829, opened 2018-04-05
fix(rule): allow empty aria-labelledby values

marcysutton:
This change allows empty aria-labelledby attributes in the valid-attr-value rule. 

Closes https://github.com/dequelabs/axe-core/issues/372

### DIFF
--- a/lib/commons/aria/attributes.js
+++ b/lib/commons/aria/attributes.js
@@ -55,7 +55,7 @@ aria.validateAttr = function (att) {
  * @return {Boolean}
  */
 aria.validateAttrValue = function (node, attr) {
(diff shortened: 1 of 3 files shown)


lib/commons/aria/attributes.js
```diff
@@ -84,6 +84,10 @@ aria.validateAttrValue = function (node, attr) {
 		return !!(value && doc.getElementById(value));
 
 	case 'idrefs':
+		// exempt attributes that allow empty strings
+		if ((attrInfo.values && attrInfo.values.indexOf('') !== -1) && value.trim().length === 0) {
```
dylanb: nit pick: `attrInfo.values.includes`

PR #842, opened 2018-04-12
chore: move github templates so they are read

marcysutton:
Pretty sure our issue and PR templates aren't getting read because they were in `doc` instead of `docs` or at the root: https://help.github.com/articles/creating-an-issue-template-for-your-repository/


marcysutton:
I'm going to merge this, since it's low risk but impacts our Github users. We can always move it again if need be.

PR #844, opened 2018-04-16
chore: add grunt build step to npm run release

marcysutton:
This should hopefully get axe-core into the proper state before creating a new entry for sri-history. My best guess as to why sri-history was changing was because it had been updated without running `grunt build` first, so the source was actually different. This makes it an automatic build step instead of relying on the publisher to remember it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,6 +49,11 @@
   ],
   "main": "axe.js",
   "typings": "axe.d.ts",
+  "standard-version": {
+    "scripts": {
+      "postbump": "npm run sri-update"
+    }
+  },
   "scripts": {
     "build": "grunt",
     "test": "grunt test",
@@ -56,7 +61,8 @@
     "version": "echo \"use 'npm run release' to bump axe-core version\" && exit 1",
     "prepublishOnly": "grunt build",
     "postinstall": "node build/utils/postinstall.js",
-    "release": "standard-version && node build/sri-update"
+    "release": "standard-version -a",
+    "sri-update": "grunt build && node build/sri-update && git add sri-history.json"
   },
   "devDependencies": {
     "angular-precommit": "^1.0.3",


package.json
```diff
@@ -56,7 +56,7 @@
     "version": "echo \"use 'npm run release' to bump axe-core version\" && exit 1",
     "prepublishOnly": "grunt build",
     "postinstall": "node build/utils/postinstall.js",
-    "release": "standard-version && node build/sri-update"
```
stephenmathieson: Maybe make `grunt build` a prerelease step instead?

marcysutton: There isn't an npm `prerelease` script, so we'd have to make one...but it would only have one thing in it. We do also call `grunt build` on `prepublishOnly` so we don't publish without building first. But honestly I don't see what a custom script would gain us since this is such a simple change.

https://docs.npmjs.com/misc/scripts

stephenmathieson: If you add a `pre` before a custom script, npm will respect it.

With a package.json that looks like:

```json
  "scripts": {
    "prerelease": "echo prerelease",
    "release": "echo release"
  },
```

Running `npm run release` will output both prerelease and release:

```
∴ npm run release

> foo@1.0.0 prerelease /Users/stephen/foo
> echo prerelease

prerelease

> foo@1.0.0 release /Users/stephen/foo
> echo release

release
```

marcysutton: I'm still not really seeing what this will gain us. It also doesn't address `prepublishOnly`, which isn't the same context as `release`

package.json
```diff
@@ -56,7 +56,7 @@
...
+    "release": "grunt build && standard-version && node build/sri-update"
```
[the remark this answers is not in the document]
marcysutton: Ok, I'll look into that. Thanks for the info

marcysutton:
Ok it took some wrangling, but I think I got it! SRI history is now being updated and committed as part of the `npm run release` command.

PR #861, opened 2018-04-24
fix: Prevent color rules from crashing Chrome 66 #856

### LINKED ISSUE #856: Infinite loop in dom.shadowElementsFromPoint
The code that steps into shadow boundaries to test color contrast is throwing an infinite loop in Chrome 66. If you run the color-contrast check tests in Chrome 65, they work fine. In Chrome 66, they throw an error: `Maximum call stack size exceeded` 
Here is the relevant code: https://github.com/dequelabs/axe-core/blob/develop/lib/commons/dom/shadow-elements-from-point.js

Perhaps something changed in Chrome since that code was shipped? I'll have to download an older version to know for sure (though I did also [ask Rob Dodson](https://twitter.com/marcysutton/status/987476758843740160) if he knows anything). When the code was written, `document.elementsFromPoint` behaved like this (in a simplified way):

- `document.elementsFromPoint(x, y)` returned nodes from the light dom only, e.g. `['#container', 'body', 'html']`
- `root.elementsFromPoint(x, y)` would return nodes from the current shadow root only, e.g. `['#target']`
- I combined the two arrays to get a complete stack of nodes to test further.

Now an infinite loop is throwing, something I would have caught in development. I've since tried preventing it by adding various breakers, but I'm still not quite getting it. 

I did notice a difference with `elementsFromPoint` once I started doing [more experiments](https://codepen.io/marcysutton/pen/VQZmYY?editors=1010):

- `document.elementsFromPoint(x, y)` still returns nodes from the light DOM, e.g. `['#container', 'body', 'html']`
- `root.elementsFromPoint(x, y)` now returns shadow *and* light DOM nodes, e.g. `['#target', '#container', 'body', 'html']`
- We only need the second array now, not a concatenated version.

It would be good to confirm whether that behavior really did change. Would we have to support both versions? 

Furthermore, how exactly is this infinite loop occurring? We still need to run `document.elementsFromPoint` to get the outer stack, step into any shadowRoots and recurse to get the whole stack of elements. I tried storing a top-level variable to filter out multiple calls to the same element or root node, but it still recursed to infinity. I have to put this task down now, but that's what I uncovered so far.

WilcoFiers:
Closes #856.

### DIFF
--- a/lib/commons/dom/shadow-elements-from-point.js
+++ b/lib/commons/dom/shadow-elements-from-point.js
@@ -9,11 +9,17 @@
  * @param {Object} [root] Shadow root or document root
  * @return {Array}
  */
-dom.shadowElementsFromPoint = function(nodeX, nodeY, root = document) {
-	return root.elementsFromPoint(nodeX, nodeY)
+dom.shadowElementsFromPoint = function(nodeX, nodeY, root = document, i=0) {
+	if (i > 999) {
+		throw new Error('Infinite loop detected');
+	}
+	return Array.from(root.elementsFromPoint(nodeX, nodeY))
+		// As of Chrome 66, elementFromPoint will return elements from parent trees.
+		// We only want to touch each tree once, so we're filtering out nodes on other trees.
+		.filter(nodes => dom.getRootNode(nodes) === root)
 		.reduce((stack, elm) => {
 			if (axe.utils.isShadowRoot(elm)) {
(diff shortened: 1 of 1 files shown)


stephenmathieson:
I strongly recommend writing a test for this change. This is a pretty bad bug that we should prevent from happening again.

dylanb:
@marcysutton any word from the Chrome team on whether the API changed and if that was intentional?

PR #898, opened 2018-05-16
feat: generated impacts as a part of rule descriptions

jkodu:
## PR Checklist

- [x] Docs have been added / updated (for bug fixes / features)

## Description of the changes

- As a part of the documentation for rule descriptions, an additional column describing the impact/ severity of each rule has been added.

- [ ] Yes
- [x] No

## Other information

@WilcoFiers - can you point me to the issue for the same?


jkodu:
@WilcoFiers - review please. Amendment as requested in place.

build/configure.js
```diff
@@ -145,19 +141,77 @@ function buildRules(grunt, options, commons, callback) {
 
 				return c.options === undefined ? id : c;
 			});
-
 		}
 
+		function parseImpactForRule(rule) {
+			
+
+			function capitalize(s) {
```
WilcoFiers: I'm finding it difficult to review code that has nested functions like this. Github won't let me collapse this so I can't simply skip over them when reviewing the outer-most function. The point of abstracting is to let me skip over some of the details, but this way I have to consider multiple scopes when reading through this. I find this is harder to read than it even would have been if the abstractions weren't put in here at all. I'd much prefer that any function that can be moved to the top of the file be moved there.

@isner @stephendonner WDYT? Is that just me?

build/configure.js
```diff
@@ -145,19 +141,77 @@ function buildRules(grunt, options, commons, callback) {
...
+				} else{
+					return [];
+				}
+			}
+
+			var cumulativeScores = getUniqueArr(
+				getScore(rule.any, true) 							// Any: get highest impact
+					.concat(getScore(rule.all, false)) 	// All: get all unique
+					.concat(getScore(rule.none, false)) // None: get all unique
+			).sort().reverse(); //order highest to lowest
```
WilcoFiers: I would think lowest to highest. Not a big deal though.

build/configure.js
stephenmathieson: While I agree, this seems to be the existing code-style. We tend to nest function declarations a lot.

Don't think this should necessarily be a blocker, but maybe others feel differently. ¯\_(ツ)_/¯

build/configure.js
```diff
@@ -145,19 +141,77 @@ function buildRules(grunt, options, commons, callback) {
...
+					}
+					return out;
+				}, []);
+			}
+
+			function getScore(checkCollection, onlyHighestScore) {
+				var scores = getImpactScores(checkCollection);
+				if(scores && scores.length) {
+					return onlyHighestScore
+						? [Math.max.apply(null, scores)]
```
stephenmathieson: Might read better as:

```js
[Math.max(...scores)]
```

jkodu: This file is not es6 friendly, leaving it as is...

build/configure.js
jkodu: Sorted - lowest to highest...

build/configure.js
jkodu: Ignoring above for now, until we reach a consensus...

build/configure.js
```diff
@@ -145,19 +141,77 @@ function buildRules(grunt, options, commons, callback) {
...
+					return self.indexOf(value) === index;
+				})
+			}
+
+			function getImpactScores(checkCollection) {
+				return checkCollection.reduce(function (out, check) {
+					var id = typeof check === 'string' ? check : check.id;
+					var definition = clone(findCheck(checks, id));
+					if (!definition) {
+						grunt.log.error('check ' + id + ' not found');
```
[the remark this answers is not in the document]
jkodu: Unsure, not a blocker so leaving as is, until I read up grunt docs.

build/configure.js
```diff
@@ -145,19 +141,77 @@ function buildRules(grunt, options, commons, callback) {
...
+					return onlyHighestScore
+						? [Math.max.apply(null, scores)]
+						: getUniqueArr(scores);
+				} else{
+					return [];
+				}
+			}
+
+			var cumulativeScores = getUniqueArr(
+				getScore(rule.any, true) 							// Any: get highest impact
```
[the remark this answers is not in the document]
jkodu: refactored so no more beautiful :shipit:

PR #927, opened 2018-05-28
fix: Make empty role=lisbox elements as incomplete

### LINKED ISSUE #874: "aria-required-children" failure on role "listbox" for missing child nodes in the DOM
```aria-required-children``` fails for a ```listbox``` not having a child role ```option```. 
In a few cases authors tend to create the child ```option```elements on the fly. 

So, on roles like ```listbox``` should we tweak ```aria-required-children``` rule to:
- Needs Review: If a ```listbox``` has no child elements that are exposed to the ATs
- Fail: If a ```listbox``` has child elements exposed to ATs and is missing role ```option```

Link to live demo: https://www.bing.com/

<pre><code>axe-core version: 3.0.0-beta.3
axe-Coconut version: 3.3.0-beta.0
</code></pre>

### PR DESCRIPTION
**Pull request by WilcoFiers**
## PR Checklist

- [ ] Docs have been added / updated (for bug fixes / features)

## Description of the changes

- Github issue: Closes #874

- [ ] Yes
- [x] No

### DIFF
--- a/lib/checks/aria/required-children.js
+++ b/lib/checks/aria/required-children.js
@@ -1,7 +1,8 @@
-var requiredOwned = axe.commons.aria.requiredOwned,
-	implicitNodes = axe.commons.aria.implicitNodes,
-	matchesSelector = axe.commons.utils.matchesSelector,
-	idrefs = axe.commons.dom.idrefs;
+const requiredOwned = axe.commons.aria.requiredOwned;
+const implicitNodes = axe.commons.aria.implicitNodes;
+const matchesSelector = axe.commons.utils.matchesSelector;
+const idrefs = axe.commons.dom.idrefs;
+const reviewEmpty = options && Array.isArray(options.reviewEmpty) ? options.reviewEmpty : [];
 
 function owns(node, virtualTree, role, ariaOwned) {
 	if (node === null) { return false; }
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: lib/checks/aria/required-children.js [hunk 1]
```diff
@@ -86,4 +87,10 @@ var missing = missingRequiredChildren(node, childRoles, all, role);
 if (!missing) { return true; }
 
 this.data(missing);
-return false;
+
+console.log(role, reviewEmpty.includes(role), missing);
```
jkodu: left intentionally?

WilcoFiers: Nope. Why did you approve?

jkodu: No harm, hence...

dylanb:
@WilcoFiers shouldn't we just pass these, rather than making them a review item? An empty list is a valid list after all.

dylanb:
I think it is actually an oversight on the part of ARIA. Even their own examples have this issue.

lib/checks/aria/required-children.json
```diff
@@ -1,11 +1,15 @@
 {
   "id": "aria-required-children",
   "evaluate": "required-children.js",
+  "options": {
+    "reviewEmpty": ["listbox"]
```
BrendanMcK: Should allow 'list' also?

Issue #97, opened 2015-07-31
New Rule: Disallowed ARIA role for given element


dsturley:
Some elements have a specific set of roles they can be given:

http://www.w3.org/html/wg/drafts/html/CR/dom.html#sec-strong-native-semantics

marcysutton:
The link in that comment throws a 404. But I'm pretty sure we covered this with implicit roles.

dylanb:
@marcysutton Here is the new link https://www.w3.org/TR/html5/dom.html#sec-strong-native-semantics you should check that we have it covered

lemnis:
I was writing a function that calculates the role that should be used by the assistive technology. It contains the check mentioned above.

My code can be found at https://gist.github.com/lemnis/5f753515a34a8fc513bca6865c122330. It has some extra functionality but could be used as a starting point.  
It's based on a newer version than the one mentioned above, https://www.w3.org/TR/2017/WD-html-aria-20171013/#docconformance.

I would like to write the code for this rule, but I do not have any experience with the code base yet.

----

**Tags:** cat.aria, wcag411, wcag412, wcag2a

## Selector

`[role]`

## Checks

### elementHasAllowedRole ( any )

1. Get the role
2. Lookup if the given role is allowed with the element
3. Return false if the element doesn't support the given role
4. Otherwise, return true

dylanb:
@lemnis thank you for your contribution but we cannot use that code unless you submit it as a pull request and accept the CLA

WilcoFiers:
@lemnis Thanks for contributing! I think there are two things to talk about before reviewing the actual code you wrote. Firstly the rule ID. I think `aria-role-allowed` would fit better with what we've got currently. What do you think?

Secondly, have you done any testing with assistive technologies to see what they do with elements that have illegal roles on them? Are the roles ignored, or are they used? I didn't think browsers ever ignore the role attribute when building the accessibility tree. If that's the case, we should tag this as a `best practice` only.

lemnis:
@WilcoFiers I think `aria-allowed-role` would be better, as there is already `aria-allowed-attr`. So it would keep the word order better in sync with existing roles.

I did a small test with Chrome and Safari in combination with force displaying a style and script tag with an added `role=button`. Both were happy to tell me it's a button. Nor did I hear that roles are ever ignored. Adding it as `best-practice` would be the best choice.

WilcoFiers:
Works for me. Thanks Lisa!

[...]

PR #945, opened 2018-06-07
feat(rule): aria-allowed-role

jkodu:
This rule is a taken over PR from the community contribution (PR: https://github.com/dequelabs/axe-core/pull/623, @lemnis - thanks for the initial work).

The rule computes if a given role on a node is an allowed aria role.

Closes issue:
- https://github.com/dequelabs/axe-core/issues/97#issuecomment-346859268

## Reviewer checks


jkodu:
Test/CI failure seems to relate to grunt-mocha issue - https://github.com/disqus/grunt-mocha/pull/7
Investigating.

lib/checks/aria/aria-allowed-role.js
```diff
@@ -0,0 +1,126 @@
...
+ * Helper object with functions to computed is role allowed for a given tag
+ */
+const elementCheck = {
+	a: (node, role) => {
+		if (node.namespaceURI === 'http://www.w3.org/2000/svg') {
+			return true;
+		}
+		if (node.href) {
+			return axe.commons.aria.isAllowedSubsetRole('aWithHref', role);
+		} else {
```
stephenmathieson: This is an unnecessary block. If we just `return` on the next line, we can keep the happy path closer to the left.

lib/checks/aria/aria-allowed-role.js
```diff
@@ -0,0 +1,126 @@
...
+			return role === 'menuitem';
+		}
+		return axe.commons.aria.isAllowedSubsetRole('button', role);
+	},
+	img: (node, role) => {
+		const hasAllowedEmptyAltRole = axe.commons.aria.isAllowedSubsetRole('imgWithEmptyAlt', role);
+		if (node.alt) {
+			// any role except the roles used by empty alt values
+			return !hasAllowedEmptyAltRole;
+		} else {
```
stephenmathieson: Can remove this `else {` block.

lib/checks/aria/aria-allowed-role.js
```diff
@@ -0,0 +1,126 @@
...
+			return hasAllowedEmptyAltRole;
+		}
+	},
+	input: (node, role) => {
+		switch (node.type) {
+		case 'button':
+			return axe.commons.aria.isAllowedSubsetRole('inputTypeButton', role);
+		case 'checkbox':
+			if (role === 'button' && node.hasAttribute('aria-pressed')) {
```
stephenmathieson: Can just `return role === 'button' && node.hasAttribute('aria-pressed')`

lib/checks/aria/aria-allowed-role.js
```diff
@@ -0,0 +1,126 @@
...
+			return role === 'menuitemradio';
+		default:
+			return false;
+		}
+	},
+	li: (node, role) => {
+		const hasImplicitListitemRole = node.matches('ol li, ul li');
+		if (hasImplicitListitemRole) {
+			return axe.commons.aria.isAllowedSubsetRole('li', role);
+		} else {
```
stephenmathieson: Can remove this `else` block.

lib/commons/aria/roles.js
```diff
@@ -123,9 +122,8 @@ aria.implicitNodes = function (role) {
  * @memberof axe.commons.aria
  * @instance
  * @param  {HTMLElement} node The node to test
- * @return {Mixed}      Either the role or `null` if there is none
+ * @return {Mixed}      			Either the role or `null` if there is none
```
stephenmathieson: Is all of this newly added whitespace intentional?

lib/commons/aria/roles.js
```diff
@@ -189,9 +187,66 @@ aria.implicitRole = function (node) {
...
+	return true;
+};
+
+/**
+ * Verify if a given role is an allowed ARIA subset role.
+ * @method isAllowedSubsetRole
+ * @memberof axe.commons.aria
+ * @instance
+ * @param 	{String} 	tagName 	tag name of a node
+ * @param 	{String} 	role 			aria role to check
```
stephenmathieson: More strange whitespace.

stephenmathieson:
Several small this left inline. This is looking good tho!

lib/checks/aria/aria-allowed-role.js
```diff
@@ -0,0 +1,126 @@
...
+// check if the element should be ignored, by an user setting
+if (ignoredTags && ignoredTags.indexOf(tagName) > -1) {
+	return true;
+}
+
+const roleSegments = axe.commons.aria.getRoleSegments(node);
+const implicitRole = axe.commons.aria.implicitRole(node);
+
+/**
+ * Helper object with functions to computed is role allowed for a given tag
```
isner: I understand what is meant here, but the language is quite unclear. Maybe the result of surgical edits? 😄

lib/checks/aria/aria-allowed-role.js
jkodu: Prefer to keep this one!

lib/commons/aria/roles.js
jkodu: got a bit creative with tabbing, so things are nicely tabulated/ lined up on an IDE. Nevertheless doesn't appeal on here.

jkodu:
@stephenmathieson @isner 
Updated. Thanks for the review.

lib/commons/aria/roles.js
```diff
@@ -189,9 +187,151 @@ aria.implicitRole = function (node) {
...
+		svg: (node, role) => {
+			// if in svg context it all roles may be used
+			if (node.parentNode && node.parentNode.namespaceURI === 'http://www.w3.org/2000/svg') {
+				return true;
+			}
+			return isAllowedSubsetRole('svg', role);
+		}
+	};
+
+	const element = axe.commons.aria.lookupTable.elementAriaAllowedRolesMap[tagName];
```
WilcoFiers: if instead of looking that the node name matches, we include something to match on specific attributes, things will probably be significantly easier, and require far fewer special case functions like we've got in this `elementCheck` object.

lib/rules/aria-allowed-role.json
```diff
@@ -0,0 +1,18 @@
+{
+	"id": "aria-allowed-role",
```
WilcoFiers: Thie rule has a lot of overlap with the aria-dpub-role-fallback rule. We may have to exclude the dpub roles from this rule. I'd like to know how the two relate. This rule might actually cover everything the DPUB rule does - if that's the case, we should consider deprecating the DPUB rule.

stephenmathieson:
Just a quick tip - to resolve conflicts, take "yours" then commit. Prettier will automatically format the files and the style conflicts will be resolved. The same goes for the rest of your PRs... sorry if the Prettier changes stepped on your toes at all.

lib/commons/aria/roles.js
```diff
@@ -189,9 +187,151 @@ aria.implicitRole = function (node) {
...
+
+/**
+ * Get roles applied to a given node
+ * @method getRoleSegments
+ * @memberof axe.commons.aria
+ * @instance
+ * @param {HTMLElement} node HTMLElement
+ * @return {Array<String>} return an array of roles applied to the node, if no roles, return an empty array.
+ */
+aria.getRoleSegments = function (node) {
```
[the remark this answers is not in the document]
jkodu: These are in the `aria/roles.js`, which in my reasoning seems to be right place for these utility functions related to aria roles.

lib/rules/aria-allowed-role.json
jkodu: I changed the check in the `aria-dpub-role-fallback` rule. The check for this rule is `implicit-role-fallback` which seems to do certain checks covered in `aria-allowed-role` rule (this one). I swapped out the check on `aria-dpub-role-fallback` to use `aria-allowed-role` and the tests seems to pass. So I think we can deprecate the old one.

lib/rules/aria-allowed-role.json
```diff
@@ -0,0 +1,18 @@
...
+	"excludeHidden": false,
+	"selector": "[role]",
```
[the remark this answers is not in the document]
jkodu: @WilcoFiers - I can look into MathML roles etc, guess have to build an Object Map of tagNames vs roles for the same.

Do you see this as something this rule should cover?

lib/commons/aria/roles.js
jkodu: I do not want to pollute/ hoist the lookupTable in aria with this hash of functions just yet. Leaving as is.

lib/commons/aria/roles.js
```diff
@@ -189,9 +187,151 @@ aria.implicitRole = function (node) {
...
+ * @param {String} tagName tag name of a node
+ * @param {String} role aria role to check
+ * @return {Boolean} retruns true/false
+ */
+aria.isAllowedRole = function (node, tagName, role) {
+	/**
+ 	* Collection of functions, which evaluate if a given ROLE is allowed on a specified type of ELEMENT.
+ 	* @type {Object}
+ 	*/
+	const elementCheck = {
```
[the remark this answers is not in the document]
jkodu: I do not want to pollute/ hoist the lookupTable in aria with this hash of functions just yet. Leaving as is.

lib/commons/aria/index.js
```diff
@@ -193,15 +193,17 @@ lookupTable.globalAttributes = [
 	'aria-relevant'
 ];
 
-lookupTable.role = {
+const role = {
```
stephenmathieson: This is a very large/hard to review file. Might be nice to split this up somehow, or even pull this stuff from a dependency.

No need to change now tho.

jkodu:
@WilcoFiers - review please

test/sanbox-rule-aria-allowed-role.html
```diff
@@ -0,0 +1,31 @@
+<!doctype html>
```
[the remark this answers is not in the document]
WilcoFiers: @jkodu

lib/commons/aria/roles.js
```diff
@@ -204,3 +202,248 @@ aria.implicitRole = function(node) {
...
+ * @param {HTMLElement} node HTMLElement
+ * @return {Array<String>} return an array of roles applicable to the node, if no roles, return an empty array.
+ * @private
+ */
+function getApplicableRolesForNode(node) {
+	const tagName = node.tagName.toUpperCase();
+	const nodeAttrs = Array.from(node.attributes).map(a => a.name.toUpperCase());
+
+	const rolesTable = axe.commons.aria.lookupTable.role;
+	return Object.keys(axe.commons.aria.lookupTable.role).reduce((out, role) => {
```
WilcoFiers: This reduce seems unnecessary. You're only ever doing `.includes()` with the result. You already know what role you are testing - all you have to do is pass the role in as an argument and do `lookupTable[role].allowedElements`.

lib/commons/aria/roles.js
```diff
@@ -204,3 +202,248 @@ aria.implicitRole = function(node) {
...
+		if (
+			roleValue.allowedElements &&
+			Array.isArray(roleValue.allowedElements) &&
+			roleValue.allowedElements.length
+		) {
+			// enumerate allowed elements and extract matching tagName
+			roleValue.allowedElements.forEach(el => {
+				// string
+				if (typeof el === 'string') {
+					if (el.toUpperCase() === tagName) {
```
WilcoFiers: We could just use uppercase in the lookup table.

lib/commons/aria/roles.js
```diff
@@ -204,3 +202,248 @@ aria.implicitRole = function(node) {
...
+				if (el.tagName.toUpperCase() === tagName) {
+					if (!el.attributes) {
+						out.push(role);
+						return;
+					}
+					el.attributes.forEach(attr => {
+						// check if attribute exsits
+						if (nodeAttrs.includes(attr.name.toUpperCase())) {
+							// check if value matches
+							if (attr.value === '') {
```
WilcoFiers: I suggest leaving attr.value undefined if you're not interested, and have empty string actually mean that the thing is an empty string. That's something actually need to match for ``

lib/commons/aria/roles.js
```diff
@@ -204,3 +202,248 @@ aria.implicitRole = function(node) {
...
+								out.push(role);
+								return;
+							}
+							const nodeAttrValue = node.getAttribute(attr).toUpperCase();
```
WilcoFiers: Should we trim this? (I'm not sure, please dig into the spec a bit to see if that matters.)

lib/commons/aria/index.js
```diff
@@ -1446,3 +1910,99 @@ lookupTable.role = {
 		type: 'abstract'
 	}
 };
+// extend lookupTable.role with a non-enumerable custom property to hold element/ tag names that can have any role
+Object.defineProperty(role, 'NO_ROLE', {
```
WilcoFiers: Maybe make these separate properties on the lookup table - I don't think hiding them with `enumerable: false ` is the right way to go about it.

lib/commons/aria/roles.js
```diff
@@ -204,3 +202,248 @@ aria.implicitRole = function(node) {
...
+ * @param {String} role aria role to check
+ * @return {Boolean} retruns true/false
+ * @private
+ */
+function validateAllowedAriaRole(node, tagName, role) {
+	/**
+	 * Collection of functions, which evaluate if a given ROLE is allowed on a specified type of ELEMENT.
+	 * @type {Object}
+	 */
+	const elementCheck = {
```
WilcoFiers: It's hard to tell what these things do. Some of them look like they could have been handled by the allowedElements property - can you add some tests that show how each of these is supposed to work? You may want to hoist this out of the function too.

lib/rules/aria-allowed-role.json
WilcoFiers: I'll reach out to @rdeltour to talk about how we want to handle DPUB roles in this rule.

lib/rules/aria-allowed-role.json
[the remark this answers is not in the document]
jkodu: Agreement that, MathML is de-scoped for this rule.

lib/checks/aria/aria-allowed-role.json
```diff
@@ -0,0 +1,15 @@
...
+	"evaluate": "aria-allowed-role.js",
+	"options": {
+		"allowImplicit": true,
+		"ignoredTags": []
+	},
+	"metadata": {
+		"impact": "minor",
+		"messages": {
+			"pass": "ARIA role(s) is allowed for given element",
+			"fail": "Role(s) must be one of the allowed ARIA roles"
```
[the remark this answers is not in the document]
jkodu: @WilcoFiers - no clue how to achieve this, any ideas welcome.

lib/commons/aria/roles.js
jkodu: Not going to move this out of a function, as I do not see any re-use for this just yet. Have added tests.

jkodu:
@WilcoFiers @stephenmathieson @dylanb @isner @marcysutton - review again.

lib/rules/aria-allowed-role.json
rdeltour: I haven't reviewed the PR details, but what would be the rationale for keeping the rule separate? If `aria-allowed-role` covers everything that `aria-dpub-role-fallback` does, then I would be in favor of deprecating the latter.

If you explicitly want the defaults to _not_ include DPUB roles, then yes sharing the underlying checks sounds like a good simplification :-)

lib/rules/aria-allowed-role.json
```diff
@@ -0,0 +1,18 @@
...
+	"tags": [
+		"cat.aria",
+		"best-practice"
+	],
+	"metadata": {
+		"description": "Ensure role attribute has a value appropriate for the element",
```
marcysutton: Let's change this to "Ensures role attribute has an appropriate value for the element" to be consistent with #1003.

lib/commons/aria/roles.js
```diff
@@ -204,3 +202,252 @@ aria.implicitRole = function(node) {
...
+			return true;
+		},
+		OPTION: ({ node }) => {
+			const withinOptionList = node.matches(
+				'select > option, datalist > option, optgroup > option'
+			);
+			return !withinOptionList;
+		},
+		SELECT: ({ node, role }) =>
+			!node.multiple && node.size <= 1 && role === 'menu',
```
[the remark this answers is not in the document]
jkodu: Yes, here - https://github.com/dequelabs/axe-core/pull/945/files#diff-de25aedf1004f73f868f51628270f232R375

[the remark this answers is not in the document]
jkodu: Unsure how to simplify further, if role on (non multiple) select is menu, whose size is <= 1, return true

lib/commons/aria/roles.js
```diff
@@ -204,3 +202,252 @@ aria.implicitRole = function(node) {
...
+					}
+					return isApplicable;
+				case 'radio':
+					return role === 'menuitemradio';
+				default:
+					return false;
+			}
+		},
+		LI: ({ node, isApplicable }) => {
+			const hasImplicitListitemRole = node.matches('ol li, ul li');
```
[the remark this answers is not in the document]
jkodu: Yes, here - https://github.com/dequelabs/axe-core/pull/945/files#diff-de25aedf1004f73f868f51628270f232R375

lib/commons/aria/roles.js
```diff
@@ -204,3 +202,252 @@ aria.implicitRole = function(node) {
...
+			}
+			if (typeof attr.value === 'undefined') {
+				isApplicable = true;
+				return;
+			}
+			if (tagName === 'IMG' && attr.value === '') {
+				isApplicable = true;
+				return;
+			}
+			if (!node.getAttribute(attr.name)) {
```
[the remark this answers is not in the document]
jkodu: Very different, in L246, we check against nodeAttrs constructed in L217, where as here we are reading an attribute for the given node.

doc/rule-descriptions.md
```diff
@@ -3,6 +3,7 @@
 | accesskeys | Ensures every accesskey attribute value is unique | Serious | best-practice, cat.keyboard | true |
 | area-alt | Ensures &lt;area&gt; elements of image maps have alternate text | Critical | cat.text-alternatives, wcag2a, wcag111, section508, section508.22.a | true |
 | aria-allowed-attr | Ensures ARIA attributes are allowed for an element&apos;s role | Critical | cat.aria, wcag2a, wcag412 | true |
+| aria-allowed-role | Ensure role attribute has a value appropriate for the element | Minor | cat.aria, best-practice | true |
```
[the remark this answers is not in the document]
jkodu: Any suggestions what this could be @marcysutton

lib/commons/aria/roles.js
```diff
@@ -204,3 +202,252 @@ aria.implicitRole = function(node) {
...
+				node.parentNode.namespaceURI === 'http://www.w3.org/2000/svg'
+			) {
+				return true;
+			}
+			return isApplicable;
+		}
+	};
+
+	const lookupTableRole = axe.commons.aria.lookupTable.role;
+	const canHaveNoRole = lookupTableRole.NO_ROLE.allowedElements.includes(
```
WilcoFiers: I don't know where my previous comment went about not extending role with non-enumberable properties... but here it is again. I don't think this is a good idea. This quite unintuitive. We know there are users that manipulate the lookup table. This can lead to unpredictable and difficult to debug issues. Please put the `lookupTable.role.NO_ROLE.alloweDlement` array on `lookupTable.elmsAllowedNoRole`, and a similar thing for `elmsAllowedAnyRole`.

lib/commons/aria/roles.js
```diff
@@ -204,3 +202,252 @@ aria.implicitRole = function(node) {
...
+		if (!node) {
+			return roles;
+		}
+		if (node.hasAttribute('role')) {
+			const nodeRoles = axe.utils.tokenList(
+				node.getAttribute('role').toLowerCase()
+			);
+			roles = roles.concat(nodeRoles);
+		}
+		if (node.hasAttributeNS('http://www.idpf.org/2007/ops', 'type')) {
```
rdeltour: `epub:type` shouldn't be looked up in this function, as it has absolutely no accessibility support and most probably never will. I'd suggest removing this whole `if` block.

WilcoFiers: That's very good to know. Thanks Romain.

lib/rules/aria-allowed-role.json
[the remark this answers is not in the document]
jkodu: Note to self: It should not fail for custom elements

lib/rules/aria-allowed-role.json
jkodu: I did a switcheroo of the checks for these rules. and a few integration tests break and need some re-work. I have created an issue for this, so this deprecation can be handled in as a separate piece of work, than colliding with that this PR is meant to be. 

Issue: https://github.com/dequelabs/axe-core/issues/1033

lib/checks/aria/aria-allowed-role.js
```diff
@@ -0,0 +1,23 @@
...
+ */
+const { allowImplicit = true, ignoredTags = [] } = options || {};
+const tagName = node.tagName.toUpperCase();
+
+// check if the element should be ignored, by an user setting
+if (ignoredTags.map(t => t.toUpperCase()).includes(tagName)) {
+	return true;
+}
+
+const unallowedRoles = axe.commons.aria.getAriaUnAllowedRoles(
```
WilcoFiers: Can you rename this to `getElmUnallowedRoles`.

lib/checks/aria/aria-allowed-role.json
```diff
@@ -0,0 +1,15 @@
...
+	"evaluate": "aria-allowed-role.js",
+	"options": {
+		"allowImplicit": true,
+		"ignoredTags": []
+	},
+	"metadata": {
+		"impact": "minor",
+		"messages": {
+			"pass": "ARIA role is allowed for given element",
+			"fail": "{{=it.data}} role{{=it.data && it.data.length > 1 ? 's are' : ' is'}} not allowed for given element"
```
WilcoFiers: I think you'd have to do `it.data.join(', ')` so that there's a space between the roles. I would also do "Roles X, Y, Z are not allowed for given element".

lib/commons/aria/roles.js
```diff
@@ -204,3 +202,248 @@ aria.implicitRole = function(node) {
 		ariaAttributes
 	).shift();
 };
+
+/**
```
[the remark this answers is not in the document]
WilcoFiers: Still not done ;)

lib/commons/utils/valid-html-tags.js
```diff
@@ -0,0 +1,134 @@
...
+ * Verifies that if a given html tag is valid
+ * @method isValidHtmlTag
+ * @memberof axe.commons.utils
+ * @instance
+ * @param htmlTag htmlTag to check if valid
+ * @return {Boolean} true/ false
+ */
+axe.utils.isValidHtmlTag = function isValidHtmlTag(htmlTag) {
+	'use strict';
+	return htmlTags.includes(htmlTag.toLowerCase());
```
WilcoFiers: This will match elements of other markup languages too. For example, SVG has A, SCRIPT and STYLE elements. That could lead to false positives.

lib/commons/aria/get-element-unallowed-roles.js
```diff
@@ -0,0 +1,85 @@
...
+	if (
+		document.createElement(tagName).toString() === '[object HTMLUnknownElement]'
+	) {
+		return [];
+	}
+
+	const roleSegments = getRoleSegments(node);
+	const implicitRole = axe.commons.aria.implicitRole(node);
+	// stores all roles that are not allowed for a specific element most often an element only has one explicit role
+	return roleSegments.filter(role => {
```
WilcoFiers: Why aren't you returning booleans in this?

lib/commons/aria/get-element-unallowed-roles.js
```diff
@@ -0,0 +1,85 @@
...
+	const tagName = node.nodeName.toUpperCase();
+
+	// by pass custom elements
+	if (!axe.utils.isHtmlElement(node)) {
+		return [];
+	}
+
+	// by pass other elements that may not be a known html element
+	if (
+		document.createElement(tagName).toString() === '[object HTMLUnknownElement]'
```
WilcoFiers: Unnecessary due to line 47. Please remove.

lib/commons/aria/get-element-unallowed-roles.js
jkodu: We have discussed this in the past, the result of this filter is an array of unallowed roles for a given element. 

This array is returned to `aria-allowed-role` check, which needs to populate the data for doT templating with an listing of roles for fail message. 

Hence we need to return role, than a boolean.

lib/commons/aria/get-element-unallowed-roles.js
jkodu: Custom components that are written to override generic html tags still do not change the type of tag. Tested here - https://jsfiddle.net/Jey/bprv8fqe/7/

Hence removing this check.

lib/commons/aria/get-element-unallowed-roles.js
WilcoFiers: Sorry Jey. I meant inside the callback. You're doing `return;` and `return role;` instead of `return false` / `return true`. It's nitpicky, I know. But returning undefined rather than false, and returning a string rather than true makes it look like this code is supposed to do something that it doesn't actually do.

dylanb:
Does it make sense to start this pull request over with a single commit?

jkodu:
@dylanb - Probably worth taking into a new PR, or squash and merge. I will speak to @WilcoFiers to make a decision. Thanks.

test/integration/rules/aria-allowed-role/aria-allowed-role.html
```diff
@@ -0,0 +1,28 @@
+<div id='pass-div-has-any-role' role='divAnyRoleEvenInvalid'></div>
```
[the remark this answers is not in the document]
jkodu: Sorted!

stephenmathieson:
> Does it make sense to start this pull request over with a single commit?

Since this PR only adds a single feature, I think it makes much more sense to just squash it rather than doing any "extra" work.

dylanb:
I agree with @stephenmathieson : squash

jkodu:
@dylanb - yup, will be squashing it. Cheers.

jkodu:
@WilcoFiers - Please review again. Have made changes based on our previous chat. Have enhanced no role and any role as per https://www.w3.org/TR/html-aria/ and enhanced tests.

@dylanb @marcysutton - reviews welcome

lib/commons/aria/validate-node-and-attributes.js
```diff
@@ -0,0 +1,89 @@
+/* global aria */
+/**
+ * @description Method that validates a given node against a list of constraints.
+ * @param {HTMLElement} node node to verify attributes against constraints
+ * @param {Array<String|Object>} constraintsArray an array containing TAGNAME or an OBJECT abstraction with conditions and attributes
+ * @return {Boolean} true/ false based on if node passes the constraints expected
+ */
+aria.validateNodeAndAttributes = function validateNodeAndAttributes(
```
WilcoFiers: I think this one would probably be a little easier if instead of passing in an array of constraints, you had passed it one constraint. It simplifies the function quite a bit and it makes it more generic. I think something like a "nodeMatchesDefinition" where we can pass it different things could be useful in other places as well.

I don't think this needs to change now, but it's something to think about.

jkodu:
@WilcoFiers 
- Conflicts resolve. Review/ Approve.

PR #949, opened 2018-06-08
fix(rule): update for lists, list-items to handle role changes.

jkodu:
This rule updates list(items) to cater to role changes to either parent or child list elements.

This rule is a taken over PR from the community contribution (PR: https://github.com/dequelabs/axe-core/pull/518 @AlmeroSteyn - thanks for the initial work).

Closes issue:
- https://github.com/dequelabs/axe-core/issues/463

## Reviewer checks


lib/checks/lists/dlitem.js
```diff
@@ -1,2 +1,21 @@
...
+const ALLOWED_ROLES = ['list'];
+const parentTagName = parent.nodeName.toLowerCase();
+const parentRole = (parent.getAttribute('role') || '').toLowerCase();
+
+if (parentTagName === 'dl') {
+	if (parentRole) {
+		if (ALLOWED_ROLES.includes(parentRole)) {
+			return true;
+		}
+		return false;
```
dylanb: Should check to see whether the other role is a valid ARIA role or otherwise it will not actually override the semantic

lib/checks/lists/listitem.js
```diff
@@ -1,4 +1,14 @@
...
+
+if (!parent) {
+	return false;
+}
+
+const ALLOWED_TAGS = ['ul', 'ol'];
+const ALLOWED_ROLES = ['list'];
+
+const parentTagName = parent.nodeName.toLowerCase();
+const parentRole = (parent.getAttribute('role') || '').toLowerCase();
```
dylanb: filter roles to only accept valid ARIA roles

lib/checks/lists/only-dlitems.js
```diff
@@ -1,19 +1,41 @@
-var bad = [],
```
dylanb: What is this set of changes doing?

lib/checks/lists/only-listitems.js
```diff
@@ -1,18 +1,84 @@
-var bad = [],
```
dylanb: Same question for this - what are the changes doing differently?

test/checks/lists/only-listitems.js
```diff
@@ -18,12 +18,30 @@ describe('only-listitems', function () {
...
+	});
+
 	it('should return false if the list has whitespace', function () {
 		var checkArgs = checkSetup('<ol id="target"><li>Item</li>    </ol>');
 
 		assert.isFalse(checks['only-listitems'].evaluate.apply(checkContext, checkArgs));
 	});
 
+	it('should return false if the list has only an element with role listitem', function () {
+		var checkArgs = checkSetup('<ol id="target"><div role="listitem">A list</div></ol>');
```
dylanb: The problem with this one is it is an HTML validation problem because `<ol>` is only supposed to have `<li>` and `script supporting` children https://www.w3.org/TR/html52/grouping-content.html#the-ol-element

test/checks/lists/only-listitems.js
```diff
@@ -18,12 +18,30 @@ describe('only-listitems', function () {
...
+
+		assert.isFalse(checks['only-listitems'].evaluate.apply(checkContext, checkArgs));
+	});
+
+	it('should return false if the list has only multiple mixed elements with role listitem', function () {
+		var checkArgs = checkSetup('<ol id="target"><div role="listitem">list</div><li role="listitem">list</li><div role="listitem">list</div></ol>');
```
dylanb: Same HTML validation problem

test/checks/lists/only-listitems.js
```diff
@@ -62,6 +80,23 @@ describe('only-listitems', function () {
...
+		assert.isFalse(checks['only-listitems'].evaluate.apply(checkContext, checkArgs));
+	});
+
+	it('should return true if the list has only li items with their roles changed', function () {
+		var checkArgs = checkSetup('<ol id="target"><li role="menuitem">Not a list item</li><li role="menuitem">Not a list item</li></ol>');
+		assert.isTrue(checks['only-listitems'].evaluate.apply(checkContext, checkArgs));
+	});
+
+	it('should return true if <link> is used along side only li items with their roles changed', function () {
+		var checkArgs = checkSetup('<ol id="target"><link rel="stylesheet" href="theme.css"><li role="menuitem">Not a list item</li></ol>');
```
dylanb: What is the rationale for this?

jkodu:
@dylanb 
Cheers for the above, will address.

Could you expand on this comment as well (from the old PR)?
https://github.com/dequelabs/axe-core/pull/518#discussion_r145373271

lib/checks/lists/only-dlitems.js
jkodu: @dylanb - these are just changes removed from old pr - https://github.com/dequelabs/axe-core/pull/518

lib/checks/lists/only-listitems.js
```diff
@@ -1,18 +1,84 @@
...
+	'area',
+	'script',
+	'datalist',
+	'template'
+];
+
+const getIsListItemRole = (r, t) => {
+	let out = false;
+	out = (r === 'listitem' || (t === 'li' && !r));
+	return out;
```
WilcoFiers: What?

lib/checks/lists/only-listitems.js
```diff
@@ -1,18 +1,84 @@
...
+	'map',
+	'area',
+	'script',
+	'datalist',
+	'template'
+];
+
+const getIsListItemRole = (r, t) => {
+	let out = false;
+	out = (r === 'listitem' || (t === 'li' && !r));
```
WilcoFiers: `!r` should test that this is a valid non-abstract role, not just that it's null. This entire function could do with some cleanup. Don't use single letter variables, and just return on this line rather than assign it to a var and immediately return that.

lib/checks/lists/only-listitems.js
```diff
@@ -1,18 +1,84 @@
...
+let out = virtualNode.children
+	.reduce((out, { actualNode }) => {
+		/*eslint 
+		complexity: ["error", 11] 
+		*/
+		const tagName = actualNode.nodeName.toLowerCase();
+
+		switch (actualNode.nodeType) {
+		case 1:
+			if (!ALLOWED_TAGS.includes(tagName)) {
```
WilcoFiers: The logic of this block is pretty difficult to follow, far more than it needs to be.

lib/checks/lists/only-listitems.js
```diff
@@ -1,18 +1,84 @@
...
+
+const allLiItemsHaveRole = out.liItemsWithRole > 0 &&
+	virtualNodeChildrenOfTypeLi.length === out.liItemsWithRole;
+
+if (out.badNodes.length) {
+	this.relatedNodes(out.badNodes);
 }
 
-return !!bad.length || hasNonEmptyTextNode;
+return (
```
WilcoFiers: It's really difficult to see if this code actually does what it should. Please break this up.

lib/checks/lists/dlitem.js
jkodu: Note: role='fake-role' fallback to semantic.

jkodu:
@dylanb @WilcoFiers - review again.

test/checks/lists/only-listitems.js
WilcoFiers: @dylanb I think that's on me - I replied to it saying that Axe shouldn't be a validator. As long as it works in AT, it shouldn't matter if it's valid HTML or not.

test/checks/lists/only-listitems.js
WilcoFiers: See my previous comment.

test/checks/lists/only-listitems.js
WilcoFiers: @dylanb This isn't new - we've allowed pretty much any non-content elements inside of lists from way back. If you think we should change that I propose we do it in a separate PR.

jkodu:
@WilcoFiers - based on our chat, this can be approved I believe.

jkodu:
@dylanb @marcysutton - review plz

lib/checks/lists/dlitem.js
```diff
@@ -1,2 +1,26 @@
-var parent = axe.commons.dom.getComposedParent(node);
-return parent.nodeName.toUpperCase() === 'DL';
+const parent = axe.commons.dom.getComposedParent(node);
+if (!parent) {
+	return false;
```
marcysutton: What are the cases where there isn't a composed parent?

lib/checks/lists/dlitem.js
```diff
@@ -1,2 +1,26 @@
...
+}
+
+const parentTagName = parent.nodeName.toUpperCase();
+if (parentTagName !== 'DL') {
+	return false;
+}
+
+const parentRole = (parent.getAttribute('role') || '').toLowerCase();
+
+if (!parentRole) {
```
marcysutton: With this conditional and the one above, am I reading correctly that this will only match DL elements with a role bolted on? What about other elements?

jkodu: It does tackle them, because we return true. Refer integration tests as well.

lib/checks/lists/listitem.js
```diff
@@ -1,5 +1,18 @@
...
+if (parentRole && !axe.commons.aria.isValidRole(parentRole)) {
+	return false;
+}
+
+const isListRole = parentRole === 'list';
 return (
-	['UL', 'OL'].includes(parent.nodeName.toUpperCase()) ||
-	(parent.getAttribute('role') || '').toLowerCase() === 'list'
+	(ALLOWED_TAGS.includes(parentTagName) && (!parentRole || isListRole)) ||
+	isListRole
```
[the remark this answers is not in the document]
jkodu: leaving as is, abstraction reads better here, as otherwise it feels repeated.

PR #951, opened 2018-06-10
fix(accessible-name): Allow fallback labels when input has id

WilcoFiers:
## PR Checklist

- [ ] Changes have been tested in [major browsers and Assistive Technologies](https://github.com/dequelabs/axe-core/blob/develop/doc/accessibility-supported.md#accessibility-supported)
- [ ] Docs have been added / updated (for bug fixes / features)

## Description of the changes

I was poking around in the accessibleName method when I found/ noticed this little bug. It turns out that if you have an input element with an ID, Axe doesn't bother looking for an implicit label. That should obviously have been based on if there was actually an explicit label, not if the input has an ID.

- [ ] Yes
- [x] No

### DIFF
--- a/test/commons/text/accessible-text.js
+++ b/test/commons/text/accessible-text.js
@@ -778,6 +778,18 @@ describe('text.accessibleTextVirtual', function() {
 			});
 		});
 
+		it('should find implicit labels with id that does not match to a label', function () {
+			types.forEach(function(type) {
+				var t = type ? ' type="' + type + '"' : '';
+				fixture.innerHTML = '<label for="t1">Hello World' +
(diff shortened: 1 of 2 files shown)


test/commons/text/accessible-text.js
```diff
@@ -778,6 +778,18 @@ describe('text.accessibleTextVirtual', function() {
 			});
 		});
 
+		it('should find implicit labels with id', function () {
```
marcysutton: Isn't the point of this test that the ID doesn't match the label? Might be good to add "invalid ID" in the test name

marcysutton: That would work!

Issue #950, opened 2018-06-09
As of 3.0.3 the comments seem to get stripped out of axe.js


WilcoFiers:
I had a quick look at recent PRs and the changelog, and I'm not seeing any reason why this may have changed. Needs further investigating.

marcysutton:
I went digging and found it occurred when we updated dependencies with Greenkeeper. I've created a PR that fixes it: https://github.com/dequelabs/axe-core/pull/952

PR #952, opened 2018-06-13
chore: restore comments in axe.js file

marcysutton:
When we updated grunt-contrib-uglify through Greenkeeper, it didn't throw an error on the deprecated `preserveComments` option. It needed to be updated to `output.comments` instead.

Closes https://github.com/dequelabs/axe-core/issues/950

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -217,7 +217,9 @@ module.exports = function (grunt) {
 						braces: true,
 						quote_style: 1
 					},
-					preserveComments: /^!/
+					output: {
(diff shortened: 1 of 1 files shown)


WilcoFiers:
Yikes :/ OK, we have to look at how we protect ourselves from this stuff in the future. Should we add a test that checks the comment block is at the top of the axe.js file? Not sure how we'd do that :-/

marcysutton:
Seems like we could add an integration test that checks the structure of the file. That shouldn't be too difficult...

Issue #906, opened 2018-05-21
Proposal: Need for loading external assets in axe-core


jkodu:
# Load external assets in axe-core:

## Requirement:
In order to support both existing rules and upcoming new WCAG2.1 rules (Eg: issue https://github.com/dequelabs/axe-core/issues/851), there is a need to load external assets, in this case external stylesheets. 

Having tried a few options as detailed in the issue to get hold of the assets, believe the best way to go ahead would be to introduce ability to load external assets in axe-core,.

**Note**: The assets can be loaded in parallel when axe is run against rules that are not dependent on external assets.

## Considerations:
At this juncture although the need is only for the construction of CSSOM, there may be a need to support pre-fetching of external assets as support for newer rules are incorporated (Eg: AOM).

## Suggestions:
- To define a list/ enumeration of possible assets that needs to be pre-fetched.
```js
const preLoadEnumeration = Object.freeze([ 'CSSOM', 'AOM' ]);
```
- To introduce with in each check, an additional attribute `preload: ['CSSOM']`, which defines dependencies of external assets.
- To introduce a top level global configuration object on `age.run`, like so:
```js
preload: {
  assets: [ 'CSSOM' ],
  timeout: 5000 // (ms) if not passed, defaults to an internal constant.
}
```
This will provide end-users of the library to easily control loading external assets. 

## Known Catches:
- Load time for assets?
- Order of loading of assets and the effect this may have on results/ false positives?

## Some Preliminary Work:
- Refer commit - https://github.com/dequelabs/axe-core/commit/7e953e16e9777b1abe41b61ea91a9a7a2c2e63f8

## Next steps:
- Thoughts & Ideas welcome.

WilcoFiers:
- We shouldn't "wait" for these assets to load. We can load them, while running rules that don't require these assets. That might have been your intention, but I think this should be explicit.

- I think `{preload: false}` should be an option of `axe.run`, not of `axe.configure`.

- Symbol() isn't supported in IE9, so we can't really use that.

- We need a way to set the preload timeout time. We could maybe reuse the preload option, if its an int, that will be the timeout, if falsey, it's off

For clarification, there are several rules that we need this for, not just WCAG 2.1 rules. An important feature request was to have Axe test the hover / focus state of rules. We can only do that with access to the CSSOM. This is expected to cut manual audit time quite substantially.

jkodu:
@WilcoFiers - updated proposal based on your comments.

[...]

PR #958, opened 2018-06-20
feat(preload): cssom assets

jkodu:
- This PR tackles the functionality of preloading assets, currently only `cssom`.
- Only the boilerplate code to `preload` is built, this has not been injected into current `run` methods or `checks`.
- The philosophy is such that, once this configuration is set, the `audit.run` calls a `axe.utils.preload` queue to retrieve any external assets, and parses the results and passes it back to `audit.run`.

Closes issue:
- https://github.com/dequelabs/axe-core/issues/906
- https://github.com/dequelabs/axe-core/issues/1031

## Reviewer checks

- [x] Code is reviewed for security by: <Stephen


WilcoFiers:
Can you break this PR up? You're trying to do too many thing in one PR.

jkodu:
@WilcoFiers - Sorry, I cannot split this PR further. This is one complete functionality, they rely on each other. I have explained the philosophy in the description. Happy to discuss.

lib/core/base/audit.js
```diff
@@ -142,48 +160,63 @@ Audit.prototype.addCheck = function (spec) {
...
-						res(errResult);
 
-					} else {
-						rej(err);
+	// preload assets if necessary
+	axe.utils.preload(options)
+		.then(([preloadedAssets]) => { // q - retuns array -> destructure/ pluck the first item, is an key value map of preloadedAssets
+			// mutate options with preloadedAssets
+			options = {
+				...options,
```
stephenmathieson: Why spread this? It's much faster (and clearer) to just do:

```js
options.preloadedAssets = preloadedAssets
```

lib/core/base/audit.js
```diff
@@ -142,48 +160,63 @@ Audit.prototype.addCheck = function (spec) {
...
+			this.rules
+				.reduce((out, rule) => {
+					if (axe.utils.ruleShouldRun(rule, context, options)) {
+						const marker = performanceMarker(rule.id, options.performanceTimer);
+						out.defer((resolve, reject) => {
+							rule.run(
+								context,
+								options,
+								(result) => {
+									marker();
```
stephenmathieson: This is really just a style nit, but this code is very difficult to read due to the "pyramid of doom" here. It might make this easier on the eyes/brain if we pulled some of this logic out of here and into smaller functions.

For example,

```js
// construct a queue of rules to execute
this.rules.reduce(getDeferredRule, axe.utils.queue())
```

stephenmathieson: Nice article describing why it's "easier" to read/write/maintain when code isn't indented this way: https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88.

test/core/utils/preload-config.js
```diff
@@ -0,0 +1,80 @@
+describe('axe.utils.preloadConfig', function () {
+	'use strict';
+
+	it('should be a function', function () {
+		assert.isFunction(axe.utils.preloadConfig);
+	});
+
+	it('should return default preload configuration if no preload options', function () {
+		var actual = axe.utils.preloadConfig({});
+		var expected = {
```
WilcoFiers: This seems fragile. Better to just pull in the default and test that 1) it has the properties you need from it 2) it matches what's returned.

lib/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,104 @@
...
+			ids.push(node.shadowId)
+			return true;
+		})
+		.map((node) => {
+			return node.actualNode.ownerDocument;
+		});
+	ids = undefined;
+
+	return documents
+		.reduce((out, ownerDocument) => {
```
WilcoFiers: If you name "out" its easier to read.

lib/core/utils/preload.js
```diff
@@ -0,0 +1,37 @@
...
+	if (preloadConfig.preload) {
+		preloadConfig.assets.forEach((asset) => {
+			q.defer((res, rej) => {
+				preloadFunctionsMap[asset]({ asset, timeout: preloadConfig.timeout })
+					.then((result) => {
+						res(
+							result.reduce((out, asset) => {
+								return {
+									...out,
+									...asset
```
WilcoFiers: Don't be a JS ninja. If your preload function just did `return { assetType, result }` you wouldn't have to do this weird reduce thing.

test/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,120 @@
...
+	it('should return a queue', function () {
+		fixture.innerHTML = pageTpl;
+		var tree = axe._tree = axe.utils.getFlattenedTree(fixture);
+		var args = {
+			asset: ['cssom'],
+			timeout: 30000,
+			treeRoot: tree
+		};
+		var actual = axe.utils.preloadCssom(args);
+		assert.isObject(actual);
```
WilcoFiers: maybe test properties?

test/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,120 @@
...
+
+	it('should be a function', function () {
+		assert.isFunction(axe.utils.preloadCssom);
+	});
+
+	it('should return a queue', function () {
+		fixture.innerHTML = pageTpl;
+		var tree = axe._tree = axe.utils.getFlattenedTree(fixture);
+		var args = {
+			asset: ['cssom'],
```
WilcoFiers: Why do we need to tell a CSSOM loader that it's loading CSSOM? That seems unnecessary.

test/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,120 @@
+describe('axe.utils.preloadCssom', function () {
```
WilcoFiers: This is missing a bunch of tests:
- Does it actually return the right style sheets
- Does it make requests and than return stylesheets that were fetched
- Does it properly fetch all shadow DOM stylesheets as well
- Does it ignore disabled style sheets

test/core/base/audit.js
```diff
@@ -535,16 +535,19 @@ describe('Audit', function () {
 			var targetRule = mockRules[mockRules.length - 1],
```
WilcoFiers: You're not testing that assets are preloaded

lib/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,104 @@
+/**
+ * Returns a CSSStyleSheet object for a given style sheet content/ text
+ * @param {string} cssText content for the stylesheet
+ * @return {CSSStyleSheet}
+ * @private
+ */
+const getCssSheet = (cssText) => {
+	let htmlHead = document.implementation.createHTMLDocument().head;
```
WilcoFiers: We should check the performance cost of creating a new document for every sheet. Perhaps make this so we do this only once per run?

lib/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,104 @@
...
+	style.type = 'text/css';
+	style.appendChild(document.createTextNode(cssText));
+
+	// added style to temporary document
+	htmlHead.appendChild(style);
+
+	// cleanup style and temporary document after return
+	setTimeout(() => {
+		htmlHead.removeChild(style);
+		htmlHead = undefined;
```
WilcoFiers: Leave it for the garbage collector.

test/core/utils/preload.js
```diff
@@ -0,0 +1,45 @@
+
+
+describe('axe.utils.preload', function () {
```
WilcoFiers: You're only testing that a queue is returned, not that it actually loaded anything.

test/core/utils/xhr-q.js
```diff
@@ -0,0 +1,108 @@
...
+
+		axe.utils.xhrQ(config)
+			.then(function (results) {
+				var response = results[0];
+				assert.equal(response.status, 200);
+				done();
+			})
+			.catch(function () {
+				assert.fail('should not have rejected the queue.')
+				done();
```
WilcoFiers: This never gets called. `assert.fail` always throws an error, so at that point `done()` won't be executed. I think the way to do this is to pass an error into done - you may want to verify that that's the case for this particular library.

test/integration/rules/runner.tmpl
```diff
@@ -7,12 +7,14 @@
...
 		<script src="../../../node_modules/mocha/mocha.js"></script>
 		<script src="../../../node_modules/chai/chai.js"></script>
+		<script src="/node_modules/sinon/pkg/sinon.js"></script>
 		<script>
 			mocha.setup({
 				timeout: 10000,
 				ui: 'bdd'
 			});
 			var assert = chai.assert;
+			var sinon = sinon;
```
WilcoFiers: I don't think that means what you think it means ;-)

test/sandbox-preload.html
```diff
@@ -0,0 +1,52 @@
+<!DOCTYPE html>
```
WilcoFiers: Why are these files here?

lib/checks/language/valid-lang.js
```diff
@@ -1,11 +1,23 @@
...
+ * the below is a hack as options is an array in certain checks, and an object is other checks.
+ * this was to get some integration/ full tests passing.
+ * TODO:
+ * need to come up with a more robust typing of args to prevent complex object mutation
+ */
+langs = (
+	options && Array.isArray(options)
+		? options
+		: axe.commons.utils.validLangs()
+).map(getBaseLang);
```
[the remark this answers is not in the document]
WilcoFiers: Also, the fact that you needed to do this means you've introduced a breaking change. We can't do a 4.0 release at the time so we'll have to work around it.

lib/core/base/audit.js
```diff
@@ -133,6 +133,24 @@ Audit.prototype.addCheck = function (spec) {
 	}
 };
 
+
+const performanceMarker = (id, shouldMark) => {
```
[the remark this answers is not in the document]
stephenmathieson: Babel handles this. It'll be transpiled to a named function.

lib/checks/language/valid-lang.js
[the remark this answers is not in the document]
stephenmathieson: I don't think this is a breaking change. I believe it's still providing an array sometimes, and an object sometimes.

lib/core/constants.js
```diff
@@ -27,7 +27,9 @@
 		results: [],
 		resultGroups: [],
 		resultGroupMap: {},
-		impact: Object.freeze(['minor', 'moderate', 'serious', 'critical'])
+		impact: Object.freeze(['minor', 'moderate', 'serious', 'critical']),
+		preloadAssets: Object.freeze(['cssom']),
+		preloadAssetsTimeout: 30000
```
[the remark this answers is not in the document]
stephenmathieson: What is the difference between a default that never changes and a constant?

lib/core/base/audit.js
```diff
@@ -142,48 +160,63 @@ Audit.prototype.addCheck = function (spec) {
...
-							message: err.message,
-							stack: err.stack,
-							error: err
-						});
-						res(errResult);
 
-					} else {
-						rej(err);
+	// preload assets if necessary
+	axe.utils.preload(options)
```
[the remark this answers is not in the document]
jkodu: I will be extracting all of the preload logic that needs to be in run into a separate PR. Will tackle this comment there.

lib/core/constants.js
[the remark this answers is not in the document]
jkodu: Changing default to 10000ms

test/core/base/audit.js
jkodu: I will be extracting all of the preload logic that needs to be in run into a separate PR. Will tackle this comment there.

test/core/base/audit.js
```diff
@@ -535,16 +535,19 @@ describe('Audit', function () {
...
 				passed = false,
-				orig, options;
+				orig,
+				options;
 
 			fixture.innerHTML = '<a href="#">link</a>';
 			orig = rule.run;
 			rule.run = function (node, o, callback) {
-				assert.deepEqual(o, options);
+				assert.property(o, 'rules');
```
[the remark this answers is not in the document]
jkodu: I will be extracting all of the preload logic that needs to be in run into a separate PR. Will tackle this comment there.

lib/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,120 @@
...
+						timeout
+					})
+					.then(xhrResponse => {
+						if (
+							xhrResponse &&
+							Array.isArray(xhrResponse) &&
+							xhrResponse.length
+						) {
+							xhrResponse.forEach(r => {
+								const text = r.responseText ? r.responseText : r.response;
```
stephenmathieson: In what case(s) would `.responseText` not be truthy (where `.response` is)? Should we document that here?

lib/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,120 @@
...
+ */
+function preloadCssom({ asset, timeout, treeRoot = axe._tree[0] }) {
+	const documents = getDocumentsFromTreeRoot(treeRoot);
+	const q = axe.utils.queue();
+
+	if (!documents.length) {
+		return q;
+	}
+
+	const getSheetFromTextFn = (function() {
```
stephenmathieson: What is the purpose of this IIFE? It looks like we are trying to do some scoping tricks with `htmlHead`, but I don't understand _why_.

lib/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,120 @@
...
+		let htmlHead = document.implementation.createHTMLDocument().head;
```
stephenmathieson: `const`

lib/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,120 @@
...
+		return (cssText, href) => {
+			// create style node with css text
+			let style = document.createElement('style');
```
stephenmathieson: `const`

lib/core/utils/preload.js
```diff
@@ -0,0 +1,136 @@
...
+axe.utils.shouldPreload = shouldPreload;
+
+/**
+ * Constructs a configuration object representing the preload requested assets & timeout
+ * @param {Object} options run configuration options (or defaults) passed via axe.run
+ * @return {Object}
+ */
+function getPreloadConfig(options) {
+	// default fallback configuration
+	let out = {
```
stephenmathieson: Doesn't look like `out` is overwritten. Should use `const` here.

lib/core/utils/preload.js
```diff
@@ -0,0 +1,136 @@
...
+		assets: axe.constants.preloadAssets,
+		timeout: axe.constants.preloadAssetsTimeout
+	};
+
+	// if type is boolean
+	if (typeof options.preload === typeof true) {
```
stephenmathieson: This is a very strange condition. Why not just do:

```
if (typeof options.preload === 'boolean') {
```

lib/core/utils/preload.js
```diff
@@ -0,0 +1,136 @@
...
+		return out;
+	}
+
+	// if type is object - ensure an array of assets to load is specified
+	if (isPreloadValidObject(options.preload)) {
+		const requestedAssets = [];
+		options.preload.assets.forEach(asset => {
```
stephenmathieson: Looks like we're constructing a new array of assets in this loop. Maybe we should `.map()` and `.filter()` instead of this `.forEach()`?

IMO the usage of `.map()` makes the intent of the code a little easier to pickup.

lib/core/utils/preload.js
```diff
@@ -0,0 +1,136 @@
...
+				// unique assets to load, in case user had requested same asset type many times.
+				if (!requestedAssets.includes(a)) {
+					requestedAssets.push(a);
+				}
+				return a;
+			} else {
+				const e =
+					`Requested asset: ${a}, not supported by aXe.` +
+					`Supported assets are: ${axe.constants.preloadAssets
+						.map(_ => _)
```
stephenmathieson: What is the purpose of this `.map()`?

lib/core/utils/preload.js
```diff
@@ -0,0 +1,136 @@
...
+		out.assets = requestedAssets;
+
+		if (options.preload.timeout) {
+			if (
+				typeof options.preload.timeout === 'number' &&
+				!Number.isNaN(options.preload.timeout)
+			) {
+				out.timeout = options.preload.timeout;
+			} else {
+				throw new Error(`preload timeout specified is not of type number`);
```
stephenmathieson: Should this sort of validation be done sooner? It seems like we've already performed a lot of instructions so far, and now we're just going to throw and halt the program.

lib/core/utils/preload.js
```diff
@@ -0,0 +1,136 @@
...
+			}
+		}
+
+		return out;
+	} else {
```
stephenmathieson: This block is unnecessary, since we `return` on L92.

stephenmathieson: Also, I suggest doing the inverse of this and de-denting the happy path above. For example:

```
if (!isPreloadValidObject(options.preload)) {
  throw new Error('No assets....')
}
```

lib/core/utils/preload.js
```diff
@@ -0,0 +1,136 @@
...
+
+	const shouldPreload = axe.utils.shouldPreload(options);
+	if (!shouldPreload) {
+		return q;
+	}
+
+	const preloadConfig = axe.utils.getPreloadConfig(options);
+
+	preloadConfig.assets.forEach(asset => {
+		q.defer((resolve, reject) => {
```
stephenmathieson: What is the purpose of this top-level `.defer()`? It looks like this could be rewritten as:

```
prelaodedConfig.assets.forEach(asset => {
  preloadFunctionsMap[asset]({ asset, timeout: preloadConfig.timeout })
    .then(results => results[0])
})
```

lib/core/utils/xhr-q.js
```diff
@@ -0,0 +1,73 @@
+/**
+ * Returns a then(able) queue of XHR's
+ * @param {Object} config configuration for XMLHttpRequest
+ * @return {Object}
+ */
+axe.utils.xhrQ = (config) => {
```
stephenmathieson: Writing your own XHR utility scares me. HTTP requests are complicated things and servers do not always behave in reasonable ways. I know there's not currently a way to `require()` modules in this code base, but we may want to consider putting a temporary solution to that problem in place so that we don't have to handle all of the weird edge cases that mature libraries like [Axios](https://github.com/axios/axios) and [Superagent](https://github.com/visionmedia/superagent) have already dealt with.

Maybe we could do something like this:

```
$ npm install axios
echo 'axe.utils.axios = (function () {' > lib/core/utils/axios.js
cat node_modules/axios/dist/axios.js >> lib/core/utils/axios.js
echo '  var axios = window.axios' >> lib/core/utils/axios.js
echo '  delete axios' >> lib/core/utils/axios.js
echo '  return axios' >> lib/core/utils/axios.js
echo '})()' >> lib/core/utils/axios.js
```

There's possibly even a Grunt plugin or something that could do this for us.

test/core/utils/xhr-q.js
```diff
@@ -0,0 +1,108 @@
+/*global sinon */
+
+describe('axe.utils.xhrQ', function() {
```
stephenmathieson: I think we need a few more tests here if we're going to bring in our own HTTP request thing. Consider how many tests are in https://github.com/axios/axios/tree/master/test vs here.

lib/core/utils/preload-cssom.js
dylanb: This is broken. It is using a different document object to create the elements that it is then attempting to insert in the htmlHead's document. You need to store and use the document that gets created on current line 93

lib/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,120 @@
...
+			style.type = 'text/css';
+			style.href = href;
+			style.appendChild(document.createTextNode(cssText));
+			// added style to temporary document
+			htmlHead.appendChild(style);
```
dylanb: Is this guaranteed synchronous?

lib/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,120 @@
...
+	Array.from(ownerDocument.styleSheets).forEach(sheet => {
+		if (sheet.disabled) {
+			return;
+		}
+
+		try {
+			// attempt to resolve if sheet is relative/ same domain
+			sheet.cssRules && q.defer(resolve => resolve(sheet));
+		} catch (e) {
+			const deferredSheet = (resolve, reject) => {
```
dylanb: Why do we need the xhr call at all? Won't creating the `<style>` element in the head of the new document cause the file to be loaded? Won't the file already be in the cache because the page loaded? You will need to explain the logic of how this all works to me.

lib/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,120 @@
...
+	const q = axe.utils.queue();
+
+	Array.from(ownerDocument.styleSheets).forEach(sheet => {
+		if (sheet.disabled) {
+			return;
+		}
+
+		try {
+			// attempt to resolve if sheet is relative/ same domain
+			sheet.cssRules && q.defer(resolve => resolve(sheet));
```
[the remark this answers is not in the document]
jkodu: Cannot use `if` here as the `catch` block won't be triggered (if passes for both relative and external sheets).

lib/core/utils/preload-cssom.js
jkodu: Ignore the above - likely will change, once axios is in place - https://github.com/dequelabs/axe-core/pull/990

lib/core/utils/preload-cssom.js
jkodu: Aim is to re-use the same `htmlHead` with out having to create a new one over and over again, hence the IIFE to create the instance, and from then on, it is used to create and append stylesheets and read rules out of it.

lib/core/utils/preload-cssom.js
jkodu: Good spot!

lib/core/utils/preload.js
jkodu: Came up with a different workflow, so can throw earlier, than in an iteration, also keeps the line of sight in the code to the left.

lib/core/utils/preload.js
jkodu: Given this a thought, should not throw here, fallback to defaults instead.

lib/core/utils/preload.js
jkodu: prefer to keep the top level `q`, then all the preloaded asset types lives under one `q`, when is then(ed) in the `run` setup.

lib/core/utils/xhr-q.js
jkodu: See - #990

test/core/utils/xhr-q.js
jkodu: See - #990, likely we will get rid of `xhrQ`

jkodu:
@stephenmathieson @dylanb @WilcoFiers - review again please

jkodu:
@WilcoFiers @stephenmathieson @dylanb 
Update:
- Have embraced `axios` over `xhrQ`.
- @dylanb - has confirmed he is happy with the changes on a private thread

Please review again.

lib/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,110 @@
...
+function loadCssom(ownerDocument, timeout, getSheetFromTextFn) {
+	const q = axe.utils.queue();
+
+	Array.from(ownerDocument.styleSheets).forEach(sheet => {
+		if (sheet.disabled) {
+			return;
+		}
+
+		try {
+			// cannot use if/else here as the catch block won't be triggered (if passes for both relative and external sheets)
```
stephenmathieson: I do not understand this. Why would accessing the `.cssRules` property cause an exception to be thrown? Unless `sheet` is `null` or `undefined`, the JavaScript runtime will allow you to read properties from it:

```js
sheet = 'sheet'
// => "sheet"
sheet.cssRules
// => undefined

sheet = true
// => true
sheet.cssRules
// => undefined

sheet = {}
// => {}
sheet.cssRules
// => undefined

sheet = null
// => null
sheet.cssRules
// => VM352:1 Uncaught TypeError: Cannot read property 'cssRules' of null
    at <anonymous>:1:7
(anonymous) @ VM352:1
```

We're also sure that `sheet` is non-null, since we access `.href` from it on L25.

lib/core/utils/preload-cssom.js
stephenmathieson: I think we can avoid the IIFE tho:

```js
const dynamicDoc = document.implementation.createHTMLDocument()
const getSheetFromTextFn = cssText => {
  // ...
}
```

lib/core/utils/preload.js
```diff
@@ -0,0 +1,140 @@
...
+function shouldPreload(options) {
+	if (!options) {
+		return false;
+	}
+
+	if (!options.preload) {
+		return false;
+	}
+
+	if (typeof options.preload === typeof true) {
```
stephenmathieson: Please use `"boolean"` here.

lib/core/utils/preload.js
```diff
@@ -0,0 +1,140 @@
...
+		preload.assets.length
+	);
+}
+
+/**
+ * Returns a boolean which decides if preload is configured
+ * @param {Object} options run configuration options (or defaults) passed via axe.run
+ * @return {boolean}
+ */
+function shouldPreload(options) {
```
stephenmathieson: It looks like this entire function could be simplified into a few LOC:

```js
const shouldPreload = options => {
  return options && isValidPreloadObject(options.preload)
}
```

lib/core/constants.js
```diff
@@ -31,7 +31,9 @@
 		results: [],
 		resultGroups: [],
 		resultGroupMap: {},
-		impact: Object.freeze(['minor', 'moderate', 'serious', 'critical'])
+		impact: Object.freeze(['minor', 'moderate', 'serious', 'critical']),
+		preloadAssets: Object.freeze(['cssom']),
```
stephenmathieson: What is this for? Why do we need to `.freeze()` it?

Maybe add a comment here explaining "why" we need this (frozen) array?

lib/core/utils/preload.js
```diff
@@ -0,0 +1,140 @@
...
+
+	// if type is object - ensure an array of assets to load is specified
+	if (!isPreloadValidObject(options.preload)) {
+		throw new Error(
+			'No assets configured for preload in aXe run configuration'
+		);
+	}
+
+	// check if assets are valid items
+	const areRequestedAssetsValid = options.preload.assets.reduce(
```
stephenmathieson: It looks like this should be an `#every()`, not a `#reduce()`.

lib/core/utils/preload.js
```diff
@@ -0,0 +1,140 @@
...
+	);
+
+	if (!areRequestedAssetsValid) {
+		throw new Error(
+			`Requested assets, not supported by aXe.` +
+				`Supported assets are: ${axe.constants.preloadAssets.join(', ')}.`
+		);
+	}
+
+	out.assets = options.preload.assets.reduce((out, asset) => {
```
stephenmathieson: This could be a `#filter()` instead.

lib/core/utils/preload.js
jkodu: Nah, keeping this function as preload can be configured in various ways & this function keeps it clear to check. No harm.

Eg:
preload: true, // true | false (default) | { assets: ['cssom'], timeout: 30000 (optional) }

lib/core/utils/preload-cssom.js
jkodu: @stephenmathieson - try it.

Spin up a html page with one relative stylesheet, one cross-domain stylesheet and check `document.styleSheets[n].cssRules` on the external stylesheet.

Happy to find an alternative.

jkodu:
@WilcoFiers @stephenmathieson 
review/ approve

lib/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,112 @@
...
+			return;
+		}
+
+		try {
+			if (sheet.cssRules) {
+				// accessing .cssRules throws for external (cross-domain) sheets, which is handled in the catch
+				q.defer(resolve => resolve(sheet));
+			}
+		} catch (e) {
+			const deferredSheet = (resolve, reject) => {
```
WilcoFiers: Probably just inline this `q.defer` call.

lib/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,112 @@
...
+ * Returns an array of documents with a given root node/ tree
+ * @param {Object} treeRoot - the DOM tree to be inspected
+ * @return {Array} documents
+ * @private
+ */
+function getDocumentsFromTreeRoot(treeRoot) {
+	let ids = [];
+	const documents = axe.utils
+		.querySelectorAllFilter(treeRoot, '*', node => {
+			if (ids.includes(node.shadowId)) {
```
WilcoFiers: since shadowId can be undefined, might be safer if we test that explicitly, so `node.shadowId && ids.includes(...`. We've been burned by bad polyfills before.

lib/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,112 @@
...
+	return q;
+}
+
+/**
+ * Returns an array of documents with a given root node/ tree
+ * @param {Object} treeRoot - the DOM tree to be inspected
+ * @return {Array} documents
+ * @private
+ */
+function getDocumentsFromTreeRoot(treeRoot) {
```
WilcoFiers: Should we make this available on `axe.utils`?

lib/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,112 @@
...
+	const documents = getDocumentsFromTreeRoot(treeRoot);
+	const q = axe.utils.queue();
+
+	if (!documents.length) {
+		return q;
+	}
+
+	const getSheetFromTextFn = (function() {
+		const dynamicDoc = document.implementation.createHTMLDocument();
+		return cssText => {
```
WilcoFiers: Use a named function, helps with debugging.

lib/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,112 @@
...
+			return style.sheet;
+		};
+	})(); // invoke immediately
+
+	documents.forEach(doc => {
+		q.defer((resolve, reject) => {
+			loadCssom(doc, timeout, getSheetFromTextFn)
+				.then(sheets =>
+					resolve({
+						[asset]: sheets
```
WilcoFiers: This should just be `.then(resolve)`.

lib/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,112 @@
...
+ * @memberof axe.utils
+ * @instance
+ * @param {Object} object argument which is a composite object, with attributes asset, timeout, treeRoot(optional), resolve & reject
+ * asset - type of asset being loaded, in this case cssom
+ * timeout - timeout for any network calls made
+ * treeRoot - the DOM tree to be inspected
+ * resolve/ reject - promise chainable methods
+ * @return {Object}
+ */
+function preloadCssom({ asset, timeout, treeRoot = axe._tree[0] }) {
```
WilcoFiers: `asset` seems unnecessary here. If all we need if a property name for the return object, we should solve that outside of this function.

lib/core/utils/preload.js
```diff
@@ -0,0 +1,130 @@
...
+axe.utils.shouldPreload = shouldPreload;
+
+/**
+ * Constructs a configuration object representing the preload requested assets & timeout
+ * @param {Object} options run configuration options (or defaults) passed via axe.run
+ * @return {Object}
+ */
+function getPreloadConfig(options) {
+	// default fallback configuration
+	const out = {
```
WilcoFiers: Rename this to `preloadConfig` or something.

lib/core/utils/preload.js
WilcoFiers: The queue isn't chainable, so that wouldn't work. It only gives you one "than".

test/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,169 @@
...
+	afterEach(function() {
+		fixture.innerHTML = '';
+	});
+
+	it('should be a function', function() {
+		assert.isFunction(axe.utils.preloadCssom);
+	});
+
+	it('should return a queue', function() {
+		fixture.innerHTML = pageTpl;
```
WilcoFiers: The fixture is a `div` in the test page. That's not something you can just innerHTML a page into and expect it to work.

test/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,169 @@
...
+			shadowRoot.innerHTML =
+				'<link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.1.0/css/bootstrap.min.css" />';
+
+			var tree = (axe._tree = axe.utils.getFlattenedTree(fixture));
+			args.treeRoot = tree;
+			var actual = axe.utils.preloadCssom(args);
+			actual
+				.then(function(results) {
+					var sheets = results[0].cssom;
+					assert.lengthOf(sheets, 1);
```
WilcoFiers: Are you sure this isn't the testrunner stylesheet? Can you change the test so we can be sure you got the right sheet?

test/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,169 @@
...
+		var page =
+			'<!DOCTYPE html><html><head> <link disabled rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.1.0/css/bootstrap.min.css" /> </head><body></body></html>';
+		fixture.innerHTML = page;
+		var tree = (axe._tree = axe.utils.getFlattenedTree(fixture));
+		args.treeRoot = tree;
+		var actual = axe.utils.preloadCssom(args);
+		actual
+			.then(function(results) {
+				var sheets = results[0].cssom;
+				assert.lengthOf(sheets, 1); // the 1 stylesheet fetched is of the parent test runner page
```
WilcoFiers: Please test that this is indeed the test runner stylesheet.

test/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,169 @@
...
+			'<!DOCTYPE html><html><head> <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.1.0/css/bootstrap.min.css" /> </head><body></body></html>';
+		fixture.innerHTML = page;
+		var tree = (axe._tree = axe.utils.getFlattenedTree(fixture));
+		args.treeRoot = tree;
+		var actual = axe.utils.preloadCssom(args);
+		actual
+			.then(function(results) {
+				var sheets = results[0].cssom;
+				assert.lengthOf(sheets, 1);
+				sheets.forEach(function(s) {
```
WilcoFiers: why do you loop over an array with 1 item?

test/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,169 @@
...
+					assert.lengthOf(sheets, 1);
+					done();
+				})
+				.catch(function(error) {
+					done(error);
+				});
+		}
+	);
+
+	it('should make xhr for external stylesheet', function(done) {
```
WilcoFiers: That's not what you're testing. You may want to just mock axios instead to be sure this works. Best if we don't need to go to the network for our tests.

test/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,169 @@
...
+		var page =
+			'<!DOCTYPE html><html><head> <style>html {width: 100vh;}</style> <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.1.0/css/bootstrap.min.css" /> </head><body></body></html>';
+		fixture.innerHTML = page;
+		var tree = (axe._tree = axe.utils.getFlattenedTree(fixture));
+		args.treeRoot = tree;
+		var actual = axe.utils.preloadCssom(args);
+		actual
+			.then(function(results) {
+				var sheets = results[0].cssom;
+				assert.lengthOf(sheets, 2);
```
WilcoFiers: Test that you got the sheets you're expecting.

test/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,169 @@
...
+
+	var shadowSupported = axe.testUtils.shadowSupport.v1;
+	(shadowSupported ? it : xit)(
+		'should fetch all shadow DOM stylesheets',
+		function(done) {
+			fixture.innerHTML = '<div id="target" role="list"></div>';
+			var target = document.querySelector('#target');
+			var shadowRoot = target.attachShadow({ mode: 'open' });
+			shadowRoot.innerHTML =
+				'<link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.1.0/css/bootstrap.min.css" />';
```
WilcoFiers: Might be better to mock axios so we don't need to go to the network.

lib/core/utils/preload.js
stephenmathieson: Oh, weird. I figured it worked the same way as a native Promise.

lib/core/utils/preload-cssom.js
jkodu: Leaving it here for now, can migrate if other cssom rules need this, I don't think so, because preload should be the only place to construct this data.

lib/core/utils/preload-cssom.js
jkodu: I understand the thinking here, but node.shadowId seems to be undefined 13 times in my example from the qsa function. A bit confusing, can we chat on this?

lib/core/utils/preload.js
jkodu: Still keeping this one, as is.

lib/core/utils/preload.js
```diff
@@ -0,0 +1,136 @@
...
+
+/**
+ * Returns a then(able) queue with results of all requested preload(able) assets. Eg: ['cssom'].
+ * If preload is set to false, returns an empty queue.
+ * @param {Object} options run configuration options (or defaults) passed via axe.run
+ * @return {Object} queue
+ */
+function preload(options) {
+	const preloadFunctionsMap = {
+		cssom: axe.utils.preloadCssom
```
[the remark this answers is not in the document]
jkodu: This is just an object map of various preload options/ types and the respective functions to execute. 

The tests for cssom, already test the execution of the respective function & any further additions to this hash, will have their own tests. So do not see the need to abstract any more.

test/core/utils/preload-cssom.js
jkodu: Decided to write integration(full) tests for testing network requests

test/core/utils/preload-cssom.js
jkodu: It is not one item, there could be multiple sheets for the document & in this case, there is the `mocha.css` & anything added via before or to the fixture.

jkodu:
@WilcoFiers @stephenmathieson @marcysutton @dylanb @isner - review plz

test/core/utils/preload.js
```diff
@@ -0,0 +1,140 @@
...
+				.then(function(results) {
+					assert.isDefined(results);
+					done();
+				})
+				.catch(function(error) {
+					done(error);
+				});
+		});
+
+		it('should return an object with property cssom', function(done) {
```
WilcoFiers: This should test that the CSS preload function is called correctly, and that the returned value is the one that the CSS preload returned.

test/integration/full/preload-cssom/preload-cssom.js
```diff
@@ -0,0 +1,130 @@
+/* global axe, sinon, Promise */
+describe('preload cssom integration test pass', function() {
+	'use strict';
+
+	before(function(done) {
+		function start() {
+			// Stop messing with my tests Mocha!
+			document.querySelector('#mocha h1').outerHTML =
+				'<h2>preload cssom integration test</h2>';
```
WilcoFiers: You probably don't need that :)

test/integration/full/preload-cssom/preload-cssom.js
```diff
@@ -0,0 +1,130 @@
...
+
+	afterEach(function() {
+		stub.restore();
+	});
+
+	it('should ignore disabled stylesheets on top level document', function(done) {
+		var actual = axe.utils.preloadCssom(args);
+		actual.then(function(results) {
+			if (stub.callCount > 0) {
+				// This is a hack to ignore assertion on webdriver phantomjs mode, as sion stub is not hit
```
WilcoFiers: Please include some code to make sure this is only ignored in PhantomJS. We've used `if (window.PHANTOMJS) {` in the past. Be sure to do this for others too.

test/integration/full/preload-cssom/preload-cssom.js
```diff
@@ -0,0 +1,130 @@
...
+		var actual = axe.utils.preloadCssom(args);
+		actual.then(function(results) {
+			if (stub.callCount > 0) {
+				// This is a hack to ignore assertion on webdriver phantomjs mode, as sinon stub is not hit
+				var sheets = results[0];
+				var externalSheet = sheets.filter(function(s) {
+					return s.isExternal;
+				})[0];
+				assert.isDefined(externalSheet);
+				assert.property(externalSheet, 'cssRules');
```
WilcoFiers: This should contain the `body { overflow: auto; }` rule, no? Let's make sure it does.

lib/core/utils/preload.js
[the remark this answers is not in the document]
WilcoFiers: See my comment about testing that preloadCssom is called and returned correctly. We'll need to do something to prove that this works.

test/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,90 @@
...
+			.then(function(results) {
+				assert.lengthOf(results[0], 2); // returned from queue, hence the index look up
+				done();
+			})
+			.catch(function(error) {
+				done(error);
+			});
+	});
+
+	it('should fail if number of sheets returned does not match stylesheets defined in document', function(done) {
```
WilcoFiers: This isn't what's being tested. It's not true either, this should return styles in shadow DOM too.

test/integration/full/preload-cssom/preload-cssom.js
```diff
@@ -0,0 +1,130 @@
...
+			if (stub.callCount > 0) {
+				// This is a hack to ignore assertion on webdriver phantomjs mode, as sion stub is not hit
+				var sheets = results[0];
+				assert.lengthOf(sheets, 2);
+			}
+			done();
+		});
+	});
+
+	it('should ensure external stylesheet is fetched on top level document', function(done) {
```
WilcoFiers: For both the top window and for the frames (including nested frames) we should test that the following is being returned:

lib/core/utils/preload-cssom.js
jkodu: That is cool by me 🍺

test/integration/full/preload-cssom/preload-cssom.js
jkodu: No need of this anymore. I decided to stub manually instead of using a framework.

test/integration/full/preload-cssom/preload-cssom.js
jkodu: I do, otherwise there is no title generated, this seems be a problem in all tests under integration/full

test/integration/full/preload-cssom/preload-cssom.js
jkodu: Done, ignoring shadowDOM in this PR.

test/core/utils/preload-cssom.js
jkodu: Done, ignoring shadowDOM in this PR.

lib/core/utils/preload.js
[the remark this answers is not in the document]
jkodu: Ignored as per discussion.

jkodu:
@WilcoFiers - This is ready for review, albeit shadow DOM implementation.

test/core/utils/preload.js
```diff
@@ -0,0 +1,140 @@
+describe('axe.utils.preload unit tests', function() {
+	'use strict';
+
+	describe('axe.utils.preload', function() {
+		it('should be a function', function() {
```
stephenmathieson: This doesn't seem like a very important/useful test 😆

test/core/utils/preload.js
```diff
@@ -0,0 +1,140 @@
...
+			assert.isFunction(axe.utils.preload);
+		});
+
+		it('should return a queue', function() {
+			var options = {
+				preload: true
+			};
+			var actual = axe.utils.preload(options);
+			assert.isObject(actual);
```
stephenmathieson: This doesn't ensure the returned value is a "queue", but an object instead.

test/integration/full/preload-cssom/preload-cssom.js
```diff
@@ -0,0 +1,202 @@
...
+			);
+		}
+		if (document.readyState !== 'complete') {
+			window.addEventListener('load', start);
+		} else {
+			start();
+		}
+	});
+
+	function createStub(shouldReject) {
```
stephenmathieson: What happened to `sinon`? It's **much** more feature complete than this.

test/integration/full/preload-cssom/preload-cssom.js
```diff
@@ -0,0 +1,202 @@
...
+					done();
+				}
+				doneCalled = true;
+			});
+		});
+	}
+
+	beforeEach(function() {
+		if (window.PHANTOMJS) {
+			assert.ok('PhantomJS is a liar');
```
stephenmathieson: Haha, yes it is, but this assertion should probably just be a comment instead.

test/integration/full/preload-cssom/preload-cssom.js
```diff
@@ -0,0 +1,202 @@
...
+			this.currentTest.fn = function() {
```
stephenmathieson: What is this doing? It looks like it's skipping the test using undocumented/private Mocha methods/properties. This seems dangerous.

Maybe do this instead:

```js
var f = window.PHANTOMJS ? it.skip : it
f('Some Test', function (done) {
  // This is skipped in PhantomJS!
})
```

test/integration/full/preload-cssom/preload-cssom.js
```diff
@@ -0,0 +1,202 @@
...
+		createStub();
+	});
+
+	afterEach(function() {
+		restoreStub();
+	});
+
+	describe('tests for current top level document', function() {
+		it('should return inline stylesheets defined using <style> tag', function(done) {
+			getPreload().then(function(results) {
```
stephenmathieson: Do we need a `.catch()` block here?

test/integration/full/preload-cssom/preload-cssom.js
jkodu: Was not embracing any of the features of sinon :(. Also sinon stub did not work in selenium webdriver. Needed a million hacks to get a stub in place. Until we change to a more modern test suite, this should do the trick.

test/integration/full/preload-cssom/preload-cssom.js
jkodu: Nah, keeping this. I see this as a part of some old code :)

test/integration/full/preload-cssom/preload-cssom.js
jkodu: Fair, tried that, but could not get it to work. Do not see this as a major problem. Moving on.

test/integration/full/preload-cssom/preload-cssom.js
jkodu: Well, the tests are engineered to pass/ fail. I am testing for catch later too. I expect this test to pass and not error/ throw.

test/integration/full/preload-cssom/preload-cssom.js
jkodu: Oh well. No harm

jkodu:
PR updated. Review again.

test/integration/full/preload-cssom/preload-cssom.js
stephenmathieson: Fair enough. Might be useful to document that somewhere for the "next guy".

test/integration/full/preload-cssom/preload-cssom.js
stephenmathieson: I do see this a blocker. It's never a good idea to rely on private APIs.

Mocha could remove this functionality in a patch release and our tests would start failing in super cryptic ways. axe-core tests are confusing enough as-is, so I think we need to all do our part in making them as easy-to-understand/debug as possible.

The code snippet I provided works, so I don't see why we shouldn't just use it.

test/integration/full/preload-cssom/preload-cssom.js
stephenmathieson: So, if I'm understanding this test correctly, if an engineer edited some part of the API this test uses and caused `getPreload()` to throw an error, we'd rather have an unhandled rejection than a "pretty" test failure?

This seems wrong and I suspect will cause someone headaches eventually.

We could just add `.catch(done)` to these Promises and prevent crashes/unhandled rejections.

test/integration/full/preload-cssom/preload-cssom.js
marcysutton: Once we remove Phantom those can go. But until then it was a workaround for features lacking in Phantom or requiring painful workarounds

test/integration/full/preload-cssom/preload-cssom.js
jkodu: I think you have contradicted yourself here, give me a ring to discuss (easier).

lib/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,108 @@
+/**
+ * Returns a then(able) queue of CSSStyleSheet(s)
+ * @param {Object} ownerDocument document object to be inspected for stylesheets
+ * @param {number} timeout on network request for stylesheet that need to be externally fetched
+ * @param {Function} getSheetFromTextFn a utility function to generate a style sheet from text
+ * @return {Object} queue
+ * @private
+ */
+function loadCssom(ownerDocument, timeout, getSheetFromTextFn) {
```
WilcoFiers: You should just declare `getSheetFromText` in the file scope so you don't need to pass it around. This adds unnecessary complexity.

lib/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,108 @@
...
+			// external sheet -> make an xhr and q the response
+			q.defer((resolve, reject) => {
+				axe.imports
+					.axios({
+						method: 'get',
+						url: sheet.href,
+						timeout
+					})
+					.then(({ data }) => {
+						const sheet = getSheetFromTextFn(data, true); //second argument acts as > isExternal - true
```
WilcoFiers: That second argument is useful. It's always external. More importantly, `isExternal` shouldn't be set in `getSheetFromTextFn` at all. It belongs in the same function that you set isExternal = false (this one).

lib/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,108 @@
...
+ * @param {number} timeout on network request for stylesheet that need to be externally fetched
+ * @param {Function} getSheetFromTextFn a utility function to generate a style sheet from text
+ * @return {Object} queue
+ * @private
+ */
+function loadCssom(ownerDocument, timeout, getSheetFromTextFn) {
+	const q = axe.utils.queue();
+
+	Array.from(ownerDocument.styleSheets).forEach(sheet => {
+		if (sheet.disabled) {
```
WilcoFiers: You should add `&& sheet.cssRules.length <= 0`, rather than add an if statement in the try/catch block. Currently the code looks like it's missing an "else".

lib/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,108 @@
...
+ * @instance
+ * @param {Object} object argument which is a composite object, with attributes asset, timeout, treeRoot(optional), resolve & reject
+ * asset - type of asset being loaded, in this case cssom
+ * timeout - timeout for any network calls made
+ * treeRoot - the DOM tree to be inspected
+ * resolve/ reject - promise chainable methods
+ * @return {Object}
+ */
+function preloadCssom({ timeout, treeRoot = axe._tree[0] }) {
+	const documents = getDocumentsFromTreeRoot(treeRoot);
```
WilcoFiers: If we're not doing shadow DOM in this PR, than we need to replace this expression. Otherwise we're just committing buggy shadow DOM code to develop, rather than not having it supported.

lib/core/utils/preload.js
```diff
@@ -0,0 +1,116 @@
+/**
+ * Validated the preload object
+ * @param {Object | boolean} preload configuration object or boolean passed via the options parameter to axe.run
+ * @return {boolean}
+ * @private
+ */
+function isValidPreloadObject(preload) {
+	if (typeof preload !== 'object') {
+		return preload;
```
WilcoFiers: preload=false shouldn't return false in `isValidPreloadObject`. This is misleading.

test/integration/full/preload-cssom/preload-cssom.js
```diff
@@ -0,0 +1,239 @@
...
+	function getPreload(root) {
+		var config = {
+			asset: 'cssom',
+			timeout: 10000,
+			treeRoot: axe.utils.getFlattenedTree(root ? root : document)
+		};
+		return axe.utils.preloadCssom(config);
+	}
+
+	function commonTestsForRootAndFrame(root) {
```
WilcoFiers: One more thing we forgot to test is if cross domain sheets loaded through @import can be accessed. I suspect they might not, so there's still some work in there I think.

lib/core/utils/preload-cssom.js
jkodu: Solved!

lib/core/utils/preload-cssom.js
jkodu: I have just solved for shadowDOM, and it would make sense to fold all of that into the same PR. So yes, shadowDOM is in the same PR.

lib/core/utils/preload-cssom.js
jkodu: we cannot move .cssRules to here, as we need the catch to trigger for external stylesheets. Trying to read a .cssRules on external resource throws a SecurityError, which flows into the catch block. This is documented below.

jkodu: Although I have refactored slightly

test/integration/full/preload-cssom/preload-cssom.js
jkodu: Yup, that is right. This hack is spread across the code base. I have only borrowed it.

test/integration/full/preload-cssom/preload-cssom.js
jkodu: Indeed, you are right. I have come up with a solution for the same. Which is why shadowDOM work is going into this PR as against a new one.

test/integration/full/preload-cssom/preload-cssom.js
WilcoFiers: It isn't. I put this in the page-has-heading-one tests to make sure I can test that page without mocha getting in the way. That's the only place its needed.

test/integration/full/preload-cssom/preload-cssom.js
```diff
@@ -0,0 +1,280 @@
...
+							return s.isExternal;
+						})[0].sheet;
+						assertStylesheet(externalSheet, 'body', 'body{overflow:auto;}');
+						done();
+					})
+					.catch(done);
+			}
+		);
+
+		shouldIt('should reject external stylesheets', function(done) {
```
WilcoFiers: Do you mean, "should reject if external sheet fails to load"?

lib/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,231 @@
...
+			const externalRules = rules.filter(r => r.href);
+			// if no external rules then return an expanded version of
+			if (!externalRules.length) {
+				q.defer(resolve => resolve(getCssomSheet(sheet, false, shadowId)));
+				return;
+			}
+
+			// there are external rules, best to filter non-external rules and create a new stylesheet to avoid duplication
+			const nonExternalRules = rules.filter(rule => !rule.href);
+			// concat all cssText into a string for non external rules
```
WilcoFiers: I'm having a very hard time understanding what's happening here. As best I can tell you're trying to turn same origin sheets and style elements with import statements into one long stylesheet... that doesn't seem like a good idea to me. This means we lose all context of how why and where a stylesheet got included. It also ignores that `@import` can have its own media value, and that they can be recursive. We also shouldn't ignore `@import` on cross-origin sheets.

Please be sure to include tests that covers all of these things.

lib/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,231 @@
+/**
+ * Construct an extended object of cssom stylesheet with added attribtues for isExternal and shadowId
+ * @method getCssomSheet
+ * @private
+ * @param {CSSStyleSheet} sheet stylesheet
+ * @param {Boolean} isExternal flag to specify if the stylesheet was fetched via xhr
+ * @param {String} shadowId (optional) string representing shadowId if style/ sheet was constructed from shadowDOM assets
+ */
+function getCssomSheet(sheet, isExternal, shadowId) {
```
stephenmathieson: This could be a 1-liner: `const getCssomSheet = (sheet, isExternal, shadowId) => ({ sheet, isExternal, shadowId })`.

Also, this function seems unnecessary.

lib/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,231 @@
...
+	 * so taking a copy of the function here for now
+	 * TODO:
+	 * cleanup later, once build is re-engineered
+	 * @method getRootNode
+	 * @param {Element} node
+	 * @returns {DocumentFragment|Document}
+	 * @private
+	 */
+	function getRootNode(node) {
+		var doc = (node.getRootNode && node.getRootNode()) || document; // this is for backwards compatibility
```
stephenmathieson: Backwards compatibility with what?

lib/core/utils/preload.js
```diff
@@ -0,0 +1,119 @@
+/**
+ * Validated the preload object
+ * @param {Object | boolean} preload configuration object or boolean passed via the options parameter to axe.run
+ * @return {boolean}
+ * @private
+ */
+function isValidPreloadObject(preload) {
+	return (
+		typeof preload === 'object' &&
+		preload.hasOwnProperty('assets') &&
```
stephenmathieson: Can remove the `.hasOwnProperty` check, as `Array.isArray(undefined)` returns `false`.

lib/core/utils/preload.js
```diff
@@ -0,0 +1,119 @@
...
+		Array.isArray(preload.assets)
+	);
+}
+
+/**
+ * Returns a boolean which decides if preload is configured
+ * @param {Object} options run configuration options (or defaults) passed via axe.run
+ * @return {boolean}
+ */
+axe.utils.shouldPreload = function shouldPreload(options) {
```
stephenmathieson: Could simplify this:

```js
axe.utils.shouldPreload = options => options && isValidPreloadObject(options.preload)
```

lib/core/utils/preload.js
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	// check if requested assets to preload are valid items
+	const areRequestedAssetsValid = options.preload.assets.every(a =>
+		axe.constants.preloadAssets.includes(a.toLowerCase())
+	);
+
+	if (!areRequestedAssetsValid) {
+		throw new Error(
+			`Requested assets, not supported by aXe.` +
```
stephenmathieson: IIRC we're not capitalizing the "x" anymore.

lib/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,231 @@
...
+/**
+ * Returns an array of objects with root(document)
+ * @param {Object} treeRoot the DOM tree to be inspected
+ * @return {Array<Object>} array of objects, which each object containing a root (document) and an optional shadowId
+ * @private
+ */
+function getAllRootsInTree(tree) {
+	/**
+	 * Clone of axe.commons.dom.getRootNode
+	 * Note:
```
stephenmathieson: This seems like something that needs to be fixed first. Duplicating these functions will lead to maintenance nightmares.

stephenmathieson: Side note - I'm looking into what is required to get `axe.commons` included in the tests.

lib/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,231 @@
...
+					loadCssom(root, timeout, convertTextToStylesheet)
+						.then(resolve)
+						.catch(reject);
+				});
+				return out;
+			}, axe.utils.queue())
+			// await loading all such documents assets, and concat results into one object
+			.then(assets => {
+				resolve(
+					assets.reduce((out, cssomSheets) => {
```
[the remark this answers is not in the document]
jkodu: True, I was going to a check on forEach, will refactor.

lib/core/utils/preload-cssom.js
jkodu: Cloned/ Copy pasted method. 
Going to try and fix the runner template to have axe.commons instead.

lib/core/utils/preload-cssom.js
jkodu: Note: Paired on the above, refactored as discussed.

lib/core/utils/preload.js
```diff
@@ -0,0 +1,119 @@
...
+ * @return {boolean}
+ */
+axe.utils.shouldPreload = function shouldPreload(options) {
+	if (!options || !options.preload) {
+		return false;
+	}
+	if (typeof options.preload === 'boolean') {
+		return options.preload;
+	}
+	return isValidPreloadObject(options.preload);
```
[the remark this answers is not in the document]
jkodu: That is what we are doing.

	const shouldPreload = axe.utils.shouldPreload(options);
	if (!shouldPreload) {
		return q;
	}

	const preloadConfig = axe.utils.getPreloadConfig(options);

lib/core/utils/preload.js
jkodu: to avoid `i18n` concerns removed aXe.

test/core/utils/preload.js
jkodu: Sorted!

test/core/utils/preload-cssom.js
jkodu: Sorted!

test/integration/full/preload-cssom/preload-cssom.js
jkodu: Sorted!

test/integration/full/preload-cssom/preload-cssom.js
jkodu: Note: 
Basic one-level deep nested @import is tackled here.
Have created an issue to tackle further enhancements - https://github.com/dequelabs/axe-core/issues/1054

lib/commons/dom/get-root-node.js
```diff
@@ -7,12 +7,6 @@
...
  */
-dom.getRootNode = function(node) {
-	var doc = (node.getRootNode && node.getRootNode()) || document; // this is for backwards compatibility
-	if (doc === node) {
-		// disconnected node
-		doc = document;
-	}
-	return doc;
-};
+dom.getRootNode = axe.utils.getRootNode;
```
stephenmathieson: Should we add a warning here when this method is called? Just aliasing it doesn't seem to give users any reason _not_ to use it.

Something like:

```js
let didWarn = false
dom.getRootNode = (...args) => {
  if (!didWarn) {
    console.warn('axe-core: dom.getRootNode has been deprecated; use utils.getRootNode instead.')
    didWarn = true
  }
  axe.utils.getRootNode(...args)
}
```

lib/core/utils/preload-cssom.js
```diff
@@ -0,0 +1,205 @@
...
+					getExternalStylesheet({ resolve, reject, url: rule.href });
+				});
+			});
+
+			// in the same sheet -  get inline rules in <style> tag or in a CSSStyleSheet excluding @import or nested link
+			const inlineRules = rules.filter(rule => !rule.href);
+
+			// concat all cssText into a string for inline rules
+			const inlineRulesCssText = inlineRules
+				.reduce((out, rule) => {
```
stephenmathieson: This `.reduce()` seems to be performing the function of `.map()`. Should we refactor to use the intended native method here?

lib/commons/dom/get-root-node.js
jkodu: I like this.

Reason I did not do this is right away is, we are bypassing the function to `axe.utils` for the time being. 

The `jsdoc` `@deprecated` notice is a flag for the dev's (us) to take some action on a later date.

There are multiple places with in the code base still referring to `axe.commons.dom.getRootNode`, until we change all these references internally first,  to `axe.utils.getRootNode` believe we should not warn on the console.

And before all the internal changes come be made, this PR needs to be merged.

Also reckon a deprecation notice should dictate when an API is going to be deprecated, so we need to agree on that, something like `this function has been deprecated from v3.2.0, use newFunction`.

Happy to add this, once we reach that consensus.

@marcysutton @WilcoFiers - thoughts?

Thanks @stephenmathieson

lib/core/utils/preload-cssom.js
jkodu: I am taking each `CSSRule` and appending its `cssText` in to an array - `out`. I intend to keep this as `reduce`, as is, because I believe there will be logic going into this iteration as I tackle for `@import` and `media` rules.

Appreciate the comment.

lib/core/utils/preload.js
stephenmathieson: Why was my comment marked as resolved? You didn't change anything or respond.....?

jkodu: Simply because, I do not want to check if `preload` is a valid `object` until I am sure that it is an `object`. Hence a `boolean` check and an `undefined` check earlier.

Preload can take 3 values: `true`, `false` or `{ assets: ['assetKey' }` as defined in the [docs](https://github.com/dequelabs/axe-core/pull/958/files/027b3a183847bfcf1c68e69ea3d453c191126675#diff-8be2f642524623d5012c7b2cf4fa4eb9R466).

Doing this `options && isValidPreloadObject(options.preload)` ignores and overrides the boolean values due to `&&`.

Scenarios (with your suggestion):

| Preload Value | options | isValidPreloadObject | output | expected |
| ------------- |---------| -------------------- | ------ | -------- | 
| `options: undefined or null` | false | false | false | false |
| `options: { preload : true }`| true  | false | false | true (BOOM!) |
| `options: { preload : false }`| true  | false | false | false |
| `options: { preload : { assets: ['cssom'] } }`| true  | true | true | true |

I am sure there is a clever way to shorten things further, but this keeps it easy to follow.
Prefer to keep this for that reason. 

Appreciate the comments.

Issue #972, opened 2018-06-26
Incorrect typings for `axe.RunOnly`


stephenmathieson:
The TS definitions for `axe.RunOnly` are incorrect. As suggested by [axe-core.org](https://www.axe-core.org/docs/#options-parameter), the `runOnly` option should have a property `values` which is an Array of rule IDs. However, in the TS definition, [`RunOnly`](https://github.com/dequelabs/axe-core/blob/v3.0.3/axe.d.ts#L28) is either an Array of `TagValue`s or a `RunOnlyObject` (which has properties `include` and `exclude`, making it look like it belongs to `ElementContext`, not `RunOnly`).

It looks like these definitions are built by hand, so if you're all on-board, I'd  be happy to fix it.

dylanb:
yep, it should always be an `Array` of `String`s which are either tags or rule IDs

jkodu:
PR - https://github.com/dequelabs/axe-core/pull/973

marcysutton:
Shouldn't the issue stay open until the PR is merged?

PR #973, opened 2018-06-26
fix: typescript interface updates

jkodu:
## PR Checklist

- [x] Docs have been added / updated (for bug fixes / features)

## Description of the changes

- Github issue: https://github.com/dequelabs/axe-core/issues/972

- [ ] Yes
- [x] No

### DIFF
--- a/axe.d.ts
+++ b/axe.d.ts
@@ -37,13 +37,13 @@ declare module axe {
 		exclude?: string[] | string[][]
 	}
 
-	type RunCallback = (error: Error, results:AxeResults) => void;
+	type RunCallback = (error: Error, results: AxeResults) => void;
 
 	type ElementContext = Node | string | RunOnlyObject;
 
 	interface RunOnly {
 		type: RunOnlyType,
-		values?: TagValue[] | RunOnlyObject
+		values?: TagValue[] | string[]
 	}
 	interface RunOptions {
 		runOnly?: RunOnly,
@@ -121,12 +121,12 @@ declare module axe {
 	}
 	interface AxePlugin {
 		id: string,
-		run(...args:any[]): any,
+		run(...args: any[]): any,
 		commands: {
 			id: string,
-			callback(...args:any[]): void
+			callback(...args: any[]): void
 		}[],
-		cleanup?(callback:Function): void
+		cleanup?(callback: Function): void
 	}
 
 	let plugins: any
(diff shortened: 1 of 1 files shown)


axe.d.ts
```diff
@@ -4,36 +4,54 @@
...
-	type RunOnlyType = "rule" | "rules" | "tag" | "tags";
+	enum RunOnlyType {
+		rule = "rule",
+		rules = "rules",
+		tag = "tag",
+		tags = "tags"
+	};
 
-	type resultGroups = "inapplicable" | "passes" | "incomplete" | "violations";
+	type ResultGroups = "inapplicable" | "passes" | "incomplete" | "violations";
```
stephenmathieson: Should this be an `enum`? That way we get better intellisense/etc

marcysutton: Possibly, when this definition was first written support for enums wasn't good.

axe.d.ts
```diff
@@ -4,36 +4,54 @@
...
 
 	type RunOnlyObject = {
```
marcysutton: This is a confusing name if it isn't being used in runOnly (only in ElementContext)

stephenmathieson: It'd be a breaking change to rename it tho, right?

stephenmathieson:
Is anything preventing this from being merged/released? This is preventing me from using `axe-core` in a TypeScript project.

axe.d.ts
```diff
@@ -4,36 +4,54 @@
...
+	enum ImpactValue {
+		minor = "minor",
+		moderate = "moderate",
+		serious = "serious",
+		critical = "critical"
+	}
 
-	type TagValue = "wcag2a" | "wcag2aa" | "section508" | "best-practice";
+	enum TagValue {
+		wcag2a = "wcag2a",
```
stephenmathieson: This will break users' code if they were previously using `TagValue`s.

For example, we're no longer able to call [`axewebdriverjs#withTags(['wcag2a'])`](https://github.com/dequelabs/axe-webdriverjs/blob/ce8d9acc83f0b4d3c52f5df557c3561259e00469/lib/index.js#L72), but instead, will need to do `axewebdriverjs#withTags([ axe.TagValue.wcag2a ])`.

Are we comfortable making a breaking change here?

axe.d.ts
jkodu: You are right @marcysutton enums got introduced in ts 2.4 I believe

jkodu:
@stephenmathieson 
You are right, we cannot do the enum change as much as we would like it, it is a breaking change. 

But in my experience, that is the idea of typings. 
With every `major` release interfaces/ object structure does change and the availability of typings only make it easy for the end user to adapt to the newer change.

I have rolled back to `type` and just do the fix you are after.

PR #990, opened 2018-07-03
chore: ability to add external libs (axios)

jkodu:
- The aim of this PR is to allow for adding of external libraries that are readily available and well tested, to avoid overhead of creating self contained utilities within axe (not re-invent the wheel). 
- The implementation shows adding external library `axios` (which allows handling network requests), with some generic code written for allowing easier addition of newer libraries in the future.
- It is achieved by modification to the grunt build process. A new task `generate-imports` is added as a sub-task with in build step. This prepares the external assets by mounting them on `axe.imports` object as is (if not UMD wrapped), or after unwrapping UMD resolver and getting the factory to be mounted.
- `axios` was chosen over `superagent` based on this perf test - https://jsperf.com/axios-vs-superagent/

## Reviewer checks


WilcoFiers:
None of these are things that impact the users, so "feat" and "fix" are inappropriate for this PR.

WilcoFiers:
Also please don't fill out the checkboxes they are "to be filled out by PR reviewer(s)"

jkodu:
@WilcoFiers - In it's true sense this is a new feature, albeit unseen by users. We are creating a feature of being able to add external dependencies. What prefix/ commit standard works better? chore?

jkodu:
@WilcoFiers - I have added the test for axios. Please squash & merge, so the commit can be prefixed as `chore`. Thank you.

WilcoFiers:
Recording what we discussed earlier today:
- Axios will have to be put on axe.import or something so that it can't be overwritten
- Axios will have to be exposed in axe.source as well so it's still around when we inject axe.

jkodu:
Have hit a roadblock with this one now.
- Attempting to inject an external library into `axe.imports` namespace, creates side effects like overriding define[] & module.exports from axe. This is seen only when using `axe` via `require` as against script inlining which works perfectly.

Mounting the library on `axe.imports` was done by reading the lib & writing to file, with in `axe/core/imports` directory, which was later picked up by the build step to concat.
```js
return `
/* global axe, axios */
axe.imports["${library}"] = (function() {
let externalExistence = window["${library}"];
${grunt.file.read(source)} 

if(window["${library}"]) {
	delete window["${library}"];
}
if (externalExistence) {
	window["${library}"] = externalExistence
}
return ${library};
})();
`;
```

At this juncture, believe there are only a few ways of moving forward with this.

1) Use the current implementation in this PR, wherein we expose dependencies as globals.
2) Embrace `requirejs` for managing dependency and bundling (build step changes).
3) Introduce progressive typescript, the tsc - compiler as a pre-step to build, which then allows doing import statements in ts files.

Alternative suggestions welcome.

stephenmathieson:
@jkodu I think I have a pretty simple solution to the `define()` and `module.exports` checks. We can just create a shadowing (scoped) variables tricking Axios' UMD wrapper. For example:

```js
axe.imports.axios = (function () {
  // Clobber these globals, tricking UMD
  var define = undefined
  var module = undefined
  var exports = undefined

  // Keep a copy of the previous Axios global (if set)
  var _axios = window.axios

  ${axios_source_code}

  var axios = window.axios

  // Reset window.axios (if necessary)
  delete window.axios
  if (_axios) {
    window.axios = _axios
  }

  // Return "our" copy of Axios
  return axios
})()
```

jkodu:
@WilcoFiers - review again

build/tasks/generate-imports.js
```diff
@@ -0,0 +1,88 @@
...
+					writeLibrary(libName, sourceCode);
+				}
+			};
+
+			// Iterate through each library to import and process the code
+			Object.keys(LIBS_TO_IMPORT).forEach(key => {
+				processImport(key, LIBS_TO_IMPORT[key]);
+			});
+
+			done();
```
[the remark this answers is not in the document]
jkodu: You are right, initially I was using `fs` with promises for file read write, but got rid of it. This can go.

build/tasks/generate-imports.js
```diff
@@ -0,0 +1,88 @@
+/*eslint-env node */
+const UglifyJS = require('uglify-js');
+
+module.exports = grunt => {
```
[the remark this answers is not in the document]
jkodu: There are tests validating axios in umd integration tests. As we add multiple libraries, we can enhance the tests around each library.

jkodu:
@WilcoFiers 
- updated umd tests
- good catch on //.test missing, added
- migrated library(s) to import to come from Gruntfile.
- tested generated axe on node, browser, on axe extensions (Worldspace) with custom rule to verify axe.imports.

Please review again.

build/tasks/generate-imports.js
```diff
@@ -0,0 +1,83 @@
...
+	grunt.registerMultiTask(
+		'generate-imports',
+		'Task for generating an axe.imports module with external dependencies.',
+		function() {
+			// Convenience method that utilises uglifyjs tree-transformer to unwrap umd module resolver
+			const removeUMD = new UglifyJS.TreeTransformer(node => {
```
stephenmathieson: We should consider moving this to something like Babel (in the long-term). No need to change now.

build/tasks/generate-imports.js
```diff
@@ -0,0 +1,83 @@
...
+					const unWrappedCode = UglifyJS.parse(sourceCode)
+						.transform(removeUMD)
+						.print_to_string({ comments: true });
+					cb(null, unWrappedCode);
+				} catch (e) {
+					cb(e, null);
+				}
+			};
+
+			const hasUmdWrapper = sourceCode => {
```
stephenmathieson: This is a little scary and seems brittle. I think it's fine for now, since we *know* Axios has a UMD wrapper, but before we start adding more "imports", we may want to revisit this.

build/tasks/generate-imports.js
```diff
@@ -0,0 +1,83 @@
...
+				return (
+					/typeof exports/.test(sourceCode) &&
+					/typeof define/.test(sourceCode) &&
+					/typeof module/.test(sourceCode)
+				);
+			};
+
+			const writeLibrary = (libName, factory) => {
+				const lib = `/* global axe */ axe.imports["${libName}"] = ${factory}`;
```
stephenmathieson: No need for this ESLint directive, since we're ignoring this file anyway 😉

lib/core/imports/index.js
```diff
@@ -0,0 +1,10 @@
+/*eslint no-unused-vars: 0*/
+/* exported utils */
```
stephenmathieson: What is this directive for?

lib/core/imports/index.js
```diff
@@ -0,0 +1,10 @@
...
+
+/**
+ * Namespace for imports which holds globals of external dependencies.
+ * @namespace imports
+ * @memberof axe
+ */
+
+var imports = (axe.imports = {});
```
stephenmathieson: I don't think we need `var imports` here. Can probably just have `axe.imports = {}`.

PR #993, opened 2018-07-05
feat: Add WCAG 2.1 autocomplete-valid rule

WilcoFiers:
New rule for WCAG 2.1, based on my design here:
https://github.com/auto-wcag/auto-wcag/pull/170/files

## Reviewer checks

### DIFF
--- a/lib/checks/forms/autocomplete-appropriate.js
+++ b/lib/checks/forms/autocomplete-appropriate.js
@@ -1,0 +1,48 @@
+// Select and textarea is always allowed
+if (node.nodeName.toUpperCase() !== 'INPUT') {
+	return true;
+}
+
+const number = ['text', 'search', 'number'];
+const url = ['text', 'search', 'url'];
+const allowedTypesMap = {
+	bday: ['text', 'search', 'date'],
+	email: ['text', 'search', 'email'],
+	'cc-exp': ['text', 'search', 'month'],
+	'street-address': [], // Not even the default
+	tel: ['text', 'search', 'tel'],
+	'cc-exp-month': number,
+	'cc-exp-year': number,
+	'transaction-amount': number,
+	'bday-day': number,
+	'bday-month': number,
+	'bday-year': number,
+	'new-password': ['text', 'search', 'password'],
+	'current-password': ['text', 'search', 'password'],
+	url: url,
+	photo: url,
+	impp: url
+};
+
+if (typeof options === 'object') {
+	// Merge in options
+	Object.keys(options).forEach(key => {
+		if (!allowedTypesMap[key]) {
+			allowedTypesMap[key] = [];
+		}
+		allowedTypesMap[key] = allowedTypesMap[key].concat(options[key]);
+	});
+}
+
+const autocomplete = node.getAttribute('autocomplete');
+const autocompleteTerms = autocomplete
+	.split(/\s+/g)
+	.map(term => term.toLowerCase());
+const purposeTerm = autocompleteTerms[autocompleteTerms.length - 1];
+const allowedTypes = allowedTypesMap[purposeTerm];
+
+if (typeof allowedTypes === 'undefined') {
+	return node.type === 'text';
+}
+
+return allowedTypes.includes(node.type);
--- a/lib/checks/forms/autocomplete-valid.js
+++ b/lib/checks/forms/autocomplete-valid.js
@@ -1,0 +1,102 @@
+let {
+	standaloneTerms = [],
+	qualifiedTerms = [],
+	qualifiers = [],
+	locations = [],
+	looseTyped = false
+} =
+	options || {};
+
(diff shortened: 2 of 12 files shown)


lib/checks/forms/autocomplete-appropriate.js
```diff
@@ -0,0 +1,48 @@
...
+	impp: url
+};
+
+if (typeof options === 'object') {
+	// Merge in options
+	Object.keys(options).forEach(key => {
+		if (!allowedTypesMap[key]) {
+			allowedTypesMap[key] = [];
+		}
+		allowedTypesMap[key] = allowedTypesMap[key].concat(options[key]);
```
jkodu: an assumption is made that options.key is either an array of strings or string? do we need to be preventive and check for type and throw?

WilcoFiers: It should throw if options is incorrect. I can put in a separate statement to enforce it rather than get a "property of undefined" error.

lib/checks/forms/autocomplete-appropriate.js
```diff
@@ -0,0 +1,48 @@
...
+	});
+}
+
+const autocomplete = node.getAttribute('autocomplete');
+const autocompleteTerms = autocomplete
+	.split(/\s+/g)
+	.map(term => term.toLowerCase());
+const purposeTerm = autocompleteTerms[autocompleteTerms.length - 1];
```
jkodu: if autocomplete is empty, purposeTerm will hit an out of bounds exception as it will aim to retrieve item from array index -1

lib/checks/forms/autocomplete-valid.js
```diff
@@ -0,0 +1,102 @@
...
+	'tel-local',
+	'tel-local-prefix',
+	'tel-local-suffix',
+	'tel-extension',
+	'email',
+	'impp'
+]);
+
+const autocomplete = node.getAttribute('autocomplete');
+const autocompleteTerms = autocomplete.split(/\s+/g).map(term => {
```
jkodu: same comment as above, any index lookup will result in out of bounds if autocomplete is undefined.

lib/checks/forms/autocomplete-appropriate.js
dylanb: what @jkodu said

lib/rules/autocomplete-matches.js
```diff
@@ -0,0 +1,45 @@
+const { text, aria, dom } = axe.commons;
+
+const autocomplete = node.getAttribute('autocomplete');
+if (!autocomplete || text.sanitize(autocomplete) === '') {
+	return false;
```
dylanb: I think this line of code negates the comment asking you to add check for the case where `autocomplete` is empty or missing.

test/checks/forms/autocomplete-valid.js
```diff
@@ -0,0 +1,104 @@
...
+	});
+
+	it('returns false if `shipping` or `billing` is used before section-*', function() {
+		var params = autocompleteCheckParams(
+			'shipping section-foo standalone-term'
+		);
+		assert.isFalse(evaluate.apply(checkContext, params));
+	});
+
+	it('returns true if "home", "work", "mobile", "fax" or "pager" is used before aqualifier', function() {
```
dylanb: nit: missing space `aqualifier`

test/integration/rules/autocomplete-valid/autocomplete-valid.html
```diff
@@ -0,0 +1,57 @@
...
+<input type="hidden" autocomplete="username" id="inapplicable6" />
+
+<!-- native disabled -->
+<input autocomplete="username" disabled id="inapplicable7" />
+
+<!-- ARIA disabled -->
+<input autocomplete="username" aria-disabled="true" id="inapplicable8" />
+
+<!-- non-widget element -->
+<input type="button" role="none" tabindex="-1" autocomplete="username" id="inapplicable9" />
```
dylanb: I think we should have at least one example of every single valid value (or valid value combo) to ensure that there are no typos

lib/checks/forms/autocomplete-appropriate.js
WilcoFiers: I'll add something to catch empty string explicitly. It shouldn't matter though. The only way `split()` can return an empty array is if you do `''.split('')`.

lib/rules/autocomplete-matches.js
WilcoFiers: It should, but in the case of a custom rule someone might have messed with this. The check should handle it too.

test/integration/rules/autocomplete-valid/autocomplete-valid.html
WilcoFiers: sure

WilcoFiers:
@dylanb @jkodu PR has been updated.

Issue #558, opened 2017-10-04
New rule: xml-lang-mismatch


jmucha:
# Update:
A detailed rule description with test cases can be found here: https://auto-wcag.github.io/auto-wcag/rules/SC3-1-1-html-xml-lang-match.html

# Language of parts: xml-lang-mismatch

This test checks the value of the `xml:lang` attribute and the presence of the `lang` attribute.

**Tags:** wcag2a, wcag311, cat.language

## Selector

descendent-or-self::body[@xml:lang][@lang]

### language/xml-lang
1.    Check that the value of the `xml:lang` attribute is equal to the `lang` attribute. If true, pass; else 2.

### language/xml-lang-dialects
1.   Check that the language can match different dialects, e.g.

- `<html lang="en" xml:lang="en-US">`
- `<html xml:lang="nl" lang="nl-BE">`

If true, pass; else fail.

WilcoFiers:
Proposed name: xml-lang-mismatch

jkodu:
Proposed name has been updated to `html-xml-lang-mismatch`.
PR - https://github.com/dequelabs/axe-core/pull/999

WilcoFiers:
Rule was completed. Closing the issue.

PR #999, opened 2018-07-10
feat(rule): xml-lang-mismatch

jkodu:
The rule checks that for the html element, there is no mismatch between the primary language in non-empty lang and xml:lang attributes, if both are used.

See also - https://auto-wcag.github.io/auto-wcag/rules/SC3-1-1-html-xml-lang-match.html for further details of the rule.

Closes issue:
- https://github.com/dequelabs/axe-core/issues/558

## Reviewer checks


jkodu:
Tidying up. Closing PR, will re-open when work on this rule is started again.

jkodu:
Updated PR. Reviews welcome.

lib/checks/language/xml-lang-mismatch.json
```diff
@@ -0,0 +1,11 @@
+{
+  "id": "xml-lang-mismatch",
+  "evaluate": "xml-lang-mismatch.js",
+  "metadata": {
+    "impact": "moderate",
+    "messages": {
+      "pass": "Value of lang or xml:lang attribute match if both are specified and included in the list of valid languages",
+      "fail": "Value of lang or xml:lang attribute either does not match or not included in the list of valid languages"
```
WilcoFiers: We should simplify, something like "Lang and xml:lang attributes do not have the same base language"

lib/checks/language/xml-lang-mismatch.js
```diff
@@ -0,0 +1,34 @@
...
+// eg: lang='en-GB' and xml:lang='fr-FR' will yeild langSubTagsToValidate = ['en', 'fr']
+if (langSubTagsToValidate.length > 1) {
+	return false;
+}
+
+// ensure that the lang subtag value is valid
+const isValid = axe.utils.validLangs().includes(langSubTagsToValidate[0]);
+
+// return
+return isValid;
```
WilcoFiers: This shouldn't look at valid langs. That's covered by another rule (see my comment on matches). Instead, all this should do is 

```js
const { getBaseLang } = axe.commons.text;
return getBaseLang(node.getAttribute('lang')) === getBaseLang(node.getAttribute('xml:lang'));
```

lib/rules/xml-lang-mismatch.json
```diff
@@ -0,0 +1,18 @@
...
+	"id": "xml-lang-mismatch",
+	"matches": "xml-lang-mismatch-matches.js",
+	"tags": [
+		"cat.language",
+		"wcag2a",
+		"wcag311"
+	],
+	"metadata": {
+		"description": "Ensures that for the html element, there is no mismatch between the primary language in non-empty lang and xml:lang attributes, if both are used",
+		"help": "The value of the primary language subtag for the lang and xml:lang attributes should be the same for the html element"
```
WilcoFiers: I don't expect most users will know what the "primar language subtag" means. I'd suggest we say:

```json
{
  "description": "Ensure that HTML elements with both valid lang and xml:lang attributes agree on the base language of the page",
  "help": "HTML elements with lang and xml:lang must have the same base language"
}
```

test/checks/language/xml-lang-mismatch.js
```diff
@@ -0,0 +1,71 @@
...
+		assert.isTrue(
+			checks['xml-lang-mismatch'].evaluate.call(checkContext, node)
+		);
+	});
+
+	it('should return false if lang and xml:lang are not identical', function() {
+		node.setAttribute('lang', 'en');
+		node.setAttribute('xml:lang', 'fr-FR');
+		fixture.appendChild(node);
+		assert.isTrue(
```
WilcoFiers: Shouldn't this be `assert.isFalse`?

lib/checks/language/xml-lang-mismatch.js
```diff
@@ -0,0 +1,34 @@
...
+const xmlLangValue = node.getAttribute('xml:lang');
+// if no xml:lang is specified, can't compute mismatch - return false
+if (!xmlLangValue) {
+	return false;
+}
+langSubTagsToValidate.push(axe.utils.getBaseLang(langValue));
+
+// unique the produced array, if identical primary subtag values were supplied, this would result in an array of length 1.
+// eg: if lang='en-GB' and xml:lang='en-US', the below will produce langSubTagsToValidate = ['en']
+langSubTagsToValidate = axe.utils.uniqueArray(langSubTagsToValidate, []);
```
WilcoFiers: Deduplicating an array is a very round-about way to check that two values are the same. Please just use two variables and do an equality comparison.

test/core/utils/get-base-lang.js
```diff
@@ -0,0 +1,18 @@
...
+	'use strict';
+
+	it('returns base lang as peanut for argument peanut-BUTTER', function() {
+		var actual = axe.utils.getBaseLang('peanut-BUTTER');
+		assert.equal(actual, 'peanut');
+	});
+
+	it('returns false when expecting fr instead of en', function() {
+		var actual = axe.utils.getBaseLang('en');
+		assert.notEqual(actual, 'fr');
```
WilcoFiers: This seems unnecessary. What you should check is case sensitivity though.

test/core/utils/get-base-lang.js
```diff
@@ -0,0 +1,18 @@
+describe('axe.utils.getBaseLang', function() {
```
WilcoFiers: Also check things with multiple dashes. What should this do 'foo-bar-baz'? It's clearly invalid, so maybe this needs to return empty or null?

test/checks/language/xml-lang-mismatch.js
```diff
@@ -0,0 +1,71 @@
+describe('xml-lang-mismatch', function() {
+	'use strict';
+
+	var node;
+	var fixture = document.getElementById('fixture');
+	var checkContext = axe.testUtils.MockCheckContext();
+
+	beforeEach(function() {
+		node = document.createElement('html');
```
WilcoFiers: You could use a div instead, so you don't get into browser support trouble. The check is agnostic to what node its testing.

lib/checks/language/xml-lang-mismatch.json
```diff
@@ -0,0 +1,11 @@
+{
+  "id": "xml-lang-mismatch",
```
WilcoFiers: Should we call this html-xml-lang-mismatch, to distinguish it from the one where we look at all other elements that have lang and xml:lang?

jkodu: Not right now. When we refactor other lang rules, can do that as we will have better context of all rules involved. I will leave the rule name and file names as is, in this PR.

I have created an issue for this - https://github.com/dequelabs/axe-core/issues/1066

jkodu: Sorted!

lib/checks/language/xml-lang-mismatch.js
```diff
@@ -0,0 +1,5 @@
+const { getBaseLang } = axe.commons.utils;
+const primaryLangValue = getBaseLang(node.getAttribute('lang') || '');
+const primaryXmlLangValue = getBaseLang(node.getAttribute('xml:lang') || '');
+
+return primaryLangValue === primaryXmlLangValue;
```
WilcoFiers: This could be the cleanest looking check we've got. Great job Jey.

lib/commons/utils/get-base-lang.js
```diff
@@ -0,0 +1,15 @@
+/* global axe */
+/**
+ * Convenience function to extract primary language subtag from a given value
+ * @method getBaseLang
+ * @memberof axe.commons.utils
+ * @instance
+ * @param {String} value value specified as lang or xml:lang attribute
+ * @return {String}
+ */
+axe.utils.getBaseLang = function getBaseLang(lang) {
```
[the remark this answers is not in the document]
jkodu: Sorted!

jkodu:
@WilcoFiers 
All minor points, attended to. Cheers.

PR #1002, opened 2018-07-11
Region: allow forms with non-empty title, plain div[role=form]

### LINKED ISSUE #962: False positive: 'region' rule fails for <form> elements with non-empty "title" attribute
This issue is related to #723 

```<form>``` elements with non empty title should be considered as ```form``` landmarks

- Demo:
```
<form id="1" title="name 3">A form element with a non-empty title</form>
```
<pre><code>
axe-core version: 3.0.3
axe-Coconut version: 3.4.0-alpha.2

</code></pre>

### PR DESCRIPTION
**Pull request by marcysutton**
Form elements with a title attribute are now counted as landmarks, whereas forms without a label or title are not. ARIA forms are counted as landmarks regardless of label or title.

Closes  #962 

## Reviewer checks

### DIFF
--- a/lib/checks/navigation/region.js
+++ b/lib/checks/navigation/region.js
@@ -30,16 +30,14 @@ function isLandmark(virtualNode) {
 	var explictRole = (node.getAttribute('role') || '').trim().toLowerCase();
 
 	if (explictRole) {
-		if (explictRole === 'form') {
-			return !!aria.labelVirtual(virtualNode);
-		}
 		return landmarkRoles.includes(explictRole);
 	} else {
 		// Check if the node matches any of the CSS selectors of implicit landmarks
 		return implicitLandmarks.some(implicitSelector => {
 			let matches = axe.utils.matchesSelector(node, implicitSelector);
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: lib/checks/navigation/region.js [hunk 1]
```diff
@@ -30,16 +30,14 @@ function isLandmark(virtualNode) {
...
-		}
 		return landmarkRoles.includes(explictRole);
 	} else {
 		// Check if the node matches any of the CSS selectors of implicit landmarks
 		return implicitLandmarks.some(implicitSelector => {
 			let matches = axe.utils.matchesSelector(node, implicitSelector);
 			if (node.tagName.toLowerCase() === 'form') {
-				return matches && !!aria.labelVirtual(virtualNode);
+				let title = node.getAttribute('title') || '';
+				return matches && (!!aria.labelVirtual(virtualNode) || !!axe.commons.text.sanitize(title) );
```
jkodu: minor detail, why aim to sanitize title if tile is empty (in some cases)? worth adding an && to check for that.

marcysutton: Ok, I pushed another commit to not sanitize empty strings.

PR #1007, opened 2018-07-13
fix: Allow aria idref(s) to be empty

### LINKED ISSUE #179: aria-activedescendent="" - what problem does this cause?
In a few of our frameworks aria-activedescendent is set to "" until the component takes focus. Once it gets focused then it gets set correctly to a valid idref.
I'm not sure from the spec whether an empty string is valid or not for aria-activedescendent, but I really don't think this causes an accessibility issue.
Any chance this error can be suppressed for aria-activedescendent unless the component actually has focus - when it of course needs to be valid.

### LINKED ISSUE #298: consider allowing null values for aria attributes
Some frameworks such as SalesForce seem to place null values for aria-describedby etc... and then let content authors populate them later. 

We should weigh the value of flagging those as failures (current condition) vs. assuming they are not going to damage the AT so just leave them. On one hand, they "could" be a bad reference to something they should have been referencing. On the other hand they might be there for no reason as part of the framework that can be added later, or optionally. I don't presume to know the best way to go, but it is a consideration, because I currently am showing a failure of WCAG in AXE, for something that isn't a failure. Perhaps sniff to see if there are lots of those, and if so ignore???

Invalid ARIA attribute values: aria-describedby="" aria-activedescendant=""

### LINKED ISSUE #994: Allow empty string values for 'aria-current' attribute
ARIA 1.1 spec for aria-current states
> If the attribute is not present or its value is an empty string or undefined, the default value of false applies and the aria-current state MUST NOT be exposed by user agents or assistive technologies.

Consider allowing empty strings for ```aria-current```

WilcoFiers:
We have a few issues open related to allowing empty idref(s) for ARIA:
- Closes #994
- Closes #298
- Closes #179

## Reviewer checks


doc/rule-descriptions.md
```diff
@@ -36,12 +36,12 @@
...
-| landmark-banner-is-top-level | The banner landmark should not be contained in another landmark | Moderate | cat.semantics, best-practice | true |
-| landmark-contentinfo-is-top-level | The contentinfo landmark should not be contained in another landmark | Moderate | cat.semantics, best-practice | true |
-| landmark-main-is-top-level | The main landmark should not be contained in another landmark | Moderate | cat.semantics, best-practice | true |
-| landmark-no-duplicate-banner | Ensures the document has no more than one banner landmark | Moderate | cat.semantics, best-practice | true |
-| landmark-no-duplicate-contentinfo | Ensures the document has no more than one contentinfo landmark | Moderate | cat.semantics, best-practice | true |
-| landmark-one-main | Ensures a navigation point to the primary content of the page. If the page contains iframes, each iframe should contain either no main landmarks or just one | Moderate | cat.semantics, best-practice | true |
+| landmark-banner-is-top-level | Ensures the banner landmark is at top level | Moderate | cat.semantics, best-practice | true |
```
jkodu: Multiple PR's where these changes for rule descriptions are shown :( revert here, See: https://github.com/dequelabs/axe-core/pull/1017/files#r205952776

marcysutton:
We're saying there's no accessibility impact for any ARIA attribute when an idref value is missing?Looks simple enough. However, one concern I have is what if an idref value is missing due to an error, and developers won't realize something is broken by passing them all?

PR #1015, opened 2018-07-18
fix: Ignore abstracts in determining element roles

### LINKED ISSUE #1014: aria.isValidRole should take abstract roles into consideration
Most of the time when we call aria.isValidRole, we're trying to test if an element's role was changed. Since abstract roles are ignored by AT (there is nothing they map to in the accessibility API, so I'm making the assumption that they do), setting an abstract role should be treated the same way as setting an invalid role. There are a number of places where we're not doing that.

I'm working on a fix.

WilcoFiers:
Added `allowAbstract` to `aria.isValidRole` and put it to use in a few places. Couple of things to note:

- There are other places where `aria.isValidRole` is used, I looked at those and I didn't think they needed this. But you might want to check it out for yourself.
- This PR fixes a bug that was introduced in a PR that landed earlier today. The PR caused invalid roles to fail the list rules, rather than having it ignore them. There's no ticket for it.

Closes #1014 

## Reviewer checks


lib/checks/lists/listitem.js
```diff
@@ -1,18 +1,21 @@
 const parent = axe.commons.dom.getComposedParent(node);
 if (!parent) {
-	return false;
+	// Can only happen with detached DOM nodes and roots:
```
jkodu: should we do similar check for `dlitem` ?

lib/checks/lists/listitem.js
```diff
@@ -1,18 +1,21 @@
...
+) {
 	return false;
 }
 
-const isListRole = parentRole === 'list';
-return (
-	(ALLOWED_TAGS.includes(parentTagName) && (!parentRole || isListRole)) ||
-	isListRole
-);
+return ['UL', 'OL'].includes(parentTagName);
```
jkodu: ;)

lib/rules/heading-matches.js
```diff
@@ -1,11 +1,13 @@
...
-		.getAttribute('role')
-		.split(/\s+/i)
-		.filter(axe.commons.aria.isValidRole);
+	explicitRoles = axe.utils
+		.tokenList(node.getAttribute('role'))
+		.filter(role =>
+			axe.commons.aria.isValidRole(role, { disallowAbstract: true })
+		);
 }
+console.log(explicitRoles);
```
jkodu: intentional?

lib/checks/lists/listitem.js
WilcoFiers: readable code FTW

lib/checks/lists/listitem.js
WilcoFiers: Could have. If I wasn't already in the process of changing these functions I would.

lib/rules/heading-matches.js
WilcoFiers: sorted

lib/checks/lists/listitem.js
```diff
@@ -1,18 +1,18 @@
...
 const parentTagName = parent.nodeName.toUpperCase();
-
 const parentRole = (parent.getAttribute('role') || '').toLowerCase();
-if (parentRole && !axe.commons.aria.isValidRole(parentRole)) {
+
+if (parentRole === 'list') {
+	return true;
+}
+
+if (parentRole && axe.commons.aria.isValidRole(parentRole)) {
```
marcysutton: Shouldn't this check for ARIA roles that aren't `list` to return false, like the code did before? 
```
if (parentRole && !axe.commons.aria.isValidRole(parentRole)) {
```

lib/commons/aria/roles.js
```diff
@@ -6,15 +6,15 @@
...
  * @instance
  * @param  {String}  role The role to check
+ * @param  {Object}  options Use `allowAbstract` if you want abstracts
  * @return {Boolean}
  */
-aria.isValidRole = function(role) {
-	'use strict';
-	if (aria.lookupTable.role[role]) {
-		return true;
+aria.isValidRole = function(role, { allowAbstract } = {}) {
```
marcysutton: The description of this issue says `disallowAbstract`, while this says `allowAbstract`. Shouldn't they be the same?

doc/rule-descriptions.md
```diff
@@ -36,12 +36,12 @@
...
-| landmark-banner-is-top-level | The banner landmark should not be contained in another landmark | Moderate | cat.semantics, best-practice | true |
-| landmark-contentinfo-is-top-level | The contentinfo landmark should not be contained in another landmark | Moderate | cat.semantics, best-practice | true |
-| landmark-main-is-top-level | The main landmark should not be contained in another landmark | Moderate | cat.semantics, best-practice | true |
-| landmark-no-duplicate-banner | Ensures the document has no more than one banner landmark | Moderate | cat.semantics, best-practice | true |
-| landmark-no-duplicate-contentinfo | Ensures the document has no more than one contentinfo landmark | Moderate | cat.semantics, best-practice | true |
-| landmark-one-main | Ensures a navigation point to the primary content of the page. If the page contains iframes, each iframe should contain either no main landmarks or just one | Moderate | cat.semantics, best-practice | true |
+| landmark-banner-is-top-level | Ensures the banner landmark is at top level | Moderate | cat.semantics, best-practice | true |
```
jkodu: Minor: move these documentation updates to separate PR

lib/checks/aria/invalidrole.js
```diff
@@ -1 +1,3 @@
-return !axe.commons.aria.isValidRole(node.getAttribute('role'));
+return !axe.commons.aria.isValidRole(node.getAttribute('role'), {
+	allowAbstract: true
```
marcysutton: The word `allow` is throwing me off here, like abstract roles are being "permitted" rather than "included for evaluation". Maybe that's just me though?

Issue #1008, opened 2018-07-16
axe-puppeteer


anoop-gupt:
I would like to contribute in axe-puppeteer repo

anoop-gupt:
sure

stephenmathieson:
Yep! I'll push something shortly 😄

jkodu:
Closing issue as PR has been merged.

mohanraj-r:
@stephenmathieson @WilcoFiers  
The doc at [axe-core/doc/examples/puppeteer at master · dequelabs/axe-core](https://github.com/dequelabs/axe-core/tree/master/doc/examples/puppeteer) has been very useful. Thanks! 

Btw, How do I pass in the [options parameter](https://github.com/dequelabs/axe-core/blob/develop/doc/API.md#options-parameter) for `axe.Run()` inside `page.evaluateHandle()`?

stephenmathieson:
Hey @mohanraj-r! The short/sweet example added in #1018 doesn't allow you to provide custom options to `axe.run()`. To get this to work, you'll have to slightly modify it:

```js
// Whatever options you need go here:
const options = {
  runOnly: {
    type: 'tag',
    values: ['wcag2a']
  }
}

const handle = await page.evaluateHandle(`
  ${axeCore.source}

  // Run axe with your options.
  axe.run(${JSON.stringify(options)})
`)
```

I do want to warn you that the Puppeteer example is missing many useful features that `axe-webdriverjs` supports. It's not quite "production ready" - we've just got it there as a reference for developers looking to build their own library.

mohanraj-r:
Thank you @stephenmathieson 

1. I tried your suggestion, but couldn't get it to work. E.g. I am trying to restrict result types to just violations and incomplete. But when I run it, I am getting `inapplicable` in output as well 
```js
const axeConfig = {
  runOnly: {
    type: "tag",
    values: ["wcag2a", "wcag2aa"]
  },
  resultTypes: ["violations", "incomplete"]
};

 const axeRun = await page.evaluateHandle(`
    ${axeCore.source} // Inject axe source code
    axe.run(${JSON.stringify(axeConfig)}) // Run axe
    `);
```

Output: 
```
...
 incomplete:
   [ { id: 'color-contrast',
       impact: 'serious',
       tags: [Array],
       description:
        'Ensures the contrast between foreground and background colors meets WCAG 2 AA contrast ratio thresholds',
       help: 'Elements must have sufficient color contrast',
       helpUrl:
[...]

PR #1018, opened 2018-07-23
doc(examples): add simple Puppeteer example

stephenmathieson:
As requested, this patch adds an example of using `axe-core` with [Puppeteer](https://github.com/GoogleChrome/puppeteer). The example is very minimal and does not include all of `axe-webdriverjs`'s features.

Related to #1008.

## Reviewer checks

### DIFF
--- a/doc/examples/puppeteer/README.md
+++ b/doc/examples/puppeteer/README.md
@@ -1,0 +1,10 @@
+# axe-puppeteer-example
+
+This example demonstrates how to use `axe-core` with [Puppeteer](https://github.com/GoogleChrome/puppeteer).
+
+## To run the example
+
+* Ensure Node v8+ is installed and on `PATH`
+* Move to the `doc/examples/puppeteer` directory
+* Run `npm install`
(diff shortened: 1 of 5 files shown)


doc/examples/puppeteer/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "axe-puppeteer",
```
WilcoFiers: Should we add a node version here?

doc/examples/puppeteer/README.md
```diff
@@ -0,0 +1,10 @@
+# axe-puppeteer-example
+
+This example demonstrates how to use `axe-core` with [Puppeteer](https://github.com/GoogleChrome/puppeteer).
```
WilcoFiers: I think it's worth a note that this example does not run inside of iframes.

doc/examples/puppeteer/package.json
stephenmathieson: No, I don't think so. It's specified in the README which is what people are going to see on GitHub.

Issue #1021, opened 2018-07-25
only-listitems and only-dlitems pick up hidden <dom-repeat> as disallowed child


erikkroes:
As discussed on slack with @marcysutton and @alice. 

When creating a `<ul>`, `<ol>` or `<dl>` using `<dom-repeat>`, you get a `<dom-repeat style="display:none">` inside your list, alongside the listitems. Both `only-listitems` and `only-dlitems` report this as an issue, as the `<dom-repeat>` is not one of the `ALLOWED_TAGS` in the rule, and it ignores the fact the element is hidden with `display:none`.
`display:none` should be taken in consideration as it hides the element from the accessibility tree.

[Example](https://next.plnkr.co/edit/3AD8dwuF30hoRXjI)

Also, the pass and fail messages are incorrect for both rules. For example, the pass for `only-listitems` states `List element only has direct children that are allowed inside <li> elements`. The rule does not check for what is inside `<li>` elements, but what's inside `<ul>` or `<ol>` elements.

dylanb:
a very simple fix for this problem would be to ignore `display:none` items

PR #1022, opened 2018-07-28
fix: display:none should be considered by only-listitems and only-dlitems

erikkroes:
- Added a condition to only-listitems and only-dlitems to consider if an element is set to display:none.

Closes issue:

- https://github.com/dequelabs/axe-core/issues/1021

## Reviewer checks


lib/checks/lists/only-listitems.js
```diff
@@ -17,6 +17,13 @@ const getHasListItem = (hasListItem, tagName, isListItemRole) => {
 	return hasListItem || (tagName === 'LI' && isListItemRole) || isListItemRole;
 };
 
+const getIsHidden = actualNode => {
+	return (
+		window.getComputedStyle(actualNode, null).getPropertyValue('display') ===
```
jkodu: DRY, worth extracting this method to axe.utils

lib/checks/lists/only-dlitems.js
```diff
@@ -11,19 +11,32 @@ const ALLOWED_TAGS = [
...
 	const { actualNode } = childNode;
 	const tagName = actualNode.nodeName.toUpperCase();
 
 	if (actualNode.nodeType === 1 && !ALLOWED_TAGS.includes(tagName)) {
 		const role = (actualNode.getAttribute('role') || '').toLowerCase();
+		const isHidden = getIsHidden(actualNode);
+
 		if ((tagName !== 'DT' && tagName !== 'DD') || role) {
-			if (!ALLOWED_ROLES.includes(role)) {
+			if (!ALLOWED_ROLES.includes(role) && !isHidden) {
```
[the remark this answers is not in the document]
erikkroes: I followed the style of only-listitems.js. I found it to be more readable, but can change it if you like. The two files are already quite different from each other. More than I expected them to be.

lib/checks/lists/only-listitems.js
erikkroes: I wasn't aware of the utils, and I see it actually has an is-hidden.js already. Guess I can rewrite using that.

lib/checks/lists/only-dlitems.js
```diff
@@ -11,19 +11,32 @@ const ALLOWED_TAGS = [
 
 const ALLOWED_ROLES = ['definition', 'term', 'list'];
 
+const getIsHidden = actualNode => {
+	return (
+		window.getComputedStyle(actualNode, null).getPropertyValue('display') ===
```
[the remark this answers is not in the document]
jkodu: @WilcoFiers - your view on above is much appreciated.

lib/checks/lists/only-listitems.js
jkodu: Because of the above disparity, I have put in this PR - https://github.com/dequelabs/axe-core/pull/1023 which will give an idea of API methods available.

WilcoFiers: I don't think this should use the existing isHidden method, as that looks up the tree. We don't need that here. Just check for `display:none;` and `visibility:hidden;`.

erikkroes: Still worth extracting then?

lib/checks/lists/only-dlitems.js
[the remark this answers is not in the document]
marcysutton: The only styles that should apply to this are ones that impact the accessibility tree (I use the [Chrome A11y Devtools inspector](https://bit.ly/chrome-a11y) for this). Erik is right that opacity:0 is still exposed, so without some other CSS style affecting its render state it shouldn't fail by itself. 

Visibility: visible on a child is interesting, I didn't realize that could be overridden. `aria-hidden="true"` doesn't work the same way–it applies to whole HTML subtrees and can't be overridden. `aria-hidden="true"` would be appropriate to cover here.

lib/checks/lists/only-listitems.js
marcysutton: I think so, yes. See my comment above about `aria-hidden="true"`

erikkroes:
The 1001 file changes come from some file permissions that I changed, and came along. Fixed that locally already. Will pick up the rest of the comments as well.

WilcoFiers:
Superseded by #1022. Want to get it into the release of Monday, so I took it over as there didn't seem to be any progress on it anymore.

team-a11y:
Ah sorry for keeping you waiting. Other things got in the way. Great to see it got picked up!

erikkroes:
Ah sorry for keeping you waiting. Other things got in the way. Great to see it got picked up!

PR #1036, opened 2018-07-31
feat: add runtime localization support

stephenmathieson:
This patch adds support for runtime localization. A user may now provide a `locale` object (in the same shape as the locales in `/locale/*.json`) to `axe.configure()`.

For example:

```js
axe.configure({
  locale: {
    rules: {
      'some-rule': {
        help: 'the help message',
        description: 'the description'
      }
    },
    checks: {
      [...]
    }
  }
})
```

These locale strings are run thru `doT`, enabling template support. For example:

```js
axe.configure({
  locale: {
    rules: {
      foo: {
        help: 'some help',
        description: 'foo: {{~it.data:value}} {{=value}}{{~}}.'
      }
    },
    checks: {
      bar: {
        pass: 'the pass message',
        fail: 'something useful: {{~it.data:value}} {{=value}}{{~}}.'
      }
    }
  }
})
```

Upon calling `axe.reset()`, any custom locale will be reset to the "default" axe-core shipped with.

## Reviewer checks

- [x] Code is reviewed for security by: @isner, @jkodu


axe.d.ts
```diff
@@ -71,6 +71,24 @@ declare namespace axe {
...
+		};
+	}
+	interface CheckLocale {
+		[key: string]: {
+			pass: string;
+			fail: string;
+			incomplete: string | { [key: string]: string };
+		};
+	}
+	interface Locale {
```
jkodu: need to add `ts` tests for these changes by perhaps invoking Spec with these argument changes.

doc/API.md
```diff
@@ -183,6 +185,7 @@ axe.configure({
       * `tags` - array(optional, default `[]`). A list if the tags that "classify" the rule. In practice, you must supply some valid tags or the default evaluation will not invoke the rule. The convention is to include the standard (WCAG 2 and/or section 508), the WCAG 2 level, Section 508 paragraph, and the WCAG 2 success criteria. Tags are constructed by converting all letters to lower case, removing spaces and periods and concatinating the result. E.g. WCAG 2 A success criteria 1.1.1 would become ["wcag2a", "wcag111"]
       * `matches` - string(optional, default `*`). A filtering CSS selector that will exclude elements that do not match the CSS selector.
   * `disableOtherRules` - Disables all rules not included in the `rules` property.
+  * `locale` - A locale object to apply (at runtime) to all rules and checks, in the same shape as `/locales/*.json`.
```
isner: Is this enough for users to understand how to utilize the templating feature?

stephenmathieson: Probably not. I'm hoping to get some feedback here that helps me create the missing portion(s) of documentation.

marcysutton: You could add a note to the README: https://github.com/dequelabs/axe-core#localization

lib/core/base/audit.js
```diff
@@ -40,8 +40,191 @@ function Audit(audit) {
...
+ * to the default ("first") configuration.
+ */
+
+Audit.prototype._setPreviousLocale = function() {
+	if (this._previousLocale) {
+		return;
+	}
+
+	const locale = {
+		checks: {},
```
marcysutton: Should we keep track of a `lang` key here for any reason? The existing locale files include that alongside rules and checks.

lib/core/base/audit.js
```diff
@@ -40,8 +40,191 @@ function Audit(audit) {
...
+	// If the message(s) are Strings, they have not yet been run
+	// thru doT (which will return a Function).
+	if (typeof pass === 'string') {
+		pass = axe.imports.doT.compile(pass);
+	}
+	if (typeof fail === 'string') {
+		fail = axe.imports.doT.compile(fail);
+	}
+	return {
+		impact: a.impact,
```
marcysutton: What if they want to override the impact?

lib/core/base/audit.js
stephenmathieson: It's not used by anything, so saving it didn't seem super important. What do you think we need it for?

lib/core/base/audit.js
stephenmathieson: This is not currently supported with the existing localization setup. Adding it now falls out of scope of the task I was given.

If it's something we know we want, we can add support after this PR lands.

doc/API.md
stephenmathieson: @WilcoFiers and I have plans to work on documentation for this today.

lib/core/base/audit.js
```diff
@@ -40,8 +40,191 @@ function Audit(audit) {
...
+ *
+ * Both locale `a` and the returned shape resemble:
+ *
+ *    {
+ *      impact: string,
+ *      messages: {
+ *        pass: string | function,
+ *        fail: string | function,
+ *        incomplete: string | {
+ *          [key: string]: string | function
```
WilcoFiers: I don't think the value can be a function.

lib/core/base/audit.js
```diff
@@ -40,8 +40,191 @@ function Audit(audit) {
...
+	this._previousLocale = null;
 }
 
+/**
+ * Build and set the previous locale. Will noop if a previous
+ * locale was already set, as we want the ability to "reset"
+ * to the default ("first") configuration.
+ */
+
+Audit.prototype._setPreviousLocale = function() {
```
WilcoFiers: Not a huge deal, but I'd prefer we use named functions where we can in axe-core. It tends to make debugging a lot easier.

lib/core/base/audit.js
WilcoFiers: +1 to Stephen's comment. You can change impact by doing this:

```js
axe.configure({
  checks: [{
    id:'foo',
    metadata: { impact: 'minor' }
  }]
});
```

I think that instead of copying over `impact`, we should just spread the rest of the props in (`...a`). That way we can't accidentally forget to copy stuff over if we add stuff to metadata in the future.

lib/core/base/audit.js
```diff
@@ -40,8 +40,191 @@ function Audit(audit) {
...
+	// thru doT (which will return a Function).
+	if (typeof help === 'string') {
+		help = axe.imports.doT.compile(help);
+	}
+	if (typeof description === 'string') {
+		description = axe.imports.doT.compile(description);
+	}
+	return {
+		help: help || a.help,
+		description: description || a.description
```
WilcoFiers: This is missing the `helpUrl` property. I think we should spread a (`...a`) so we just copy over the remainder.

lib/core/base/audit.js
```diff
@@ -40,8 +40,191 @@ function Audit(audit) {
...
 }
 
+/**
+ * Build and set the previous locale. Will noop if a previous
+ * locale was already set, as we want the ability to "reset"
+ * to the default ("first") configuration.
+ */
+
+Audit.prototype._setPreviousLocale = function() {
+	if (this._previousLocale) {
```
WilcoFiers: I think instead of calling this previousLocale, we should call this defaultLocale or originalLocale. Same for the function names. We're not going back to the previous one. This is going back to the original locale (which is what reset should do). You may want to put a test in place for that as well.

lib/core/base/audit.js
WilcoFiers: It's a good point. I think we should include a lang key on the results object. I think we should do that separately from this PR though.

test/core/public/configure.js
```diff
@@ -239,4 +239,230 @@ describe('axe.configure', function() {
 		assert.equal(axe._audit.rules[3].id, 'black-panther');
 		assert.equal(axe._audit.rules[3].enabled, true);
 	});
+
+	describe('given a locale object', function() {
```
WilcoFiers: I think we need tests for:

1. An axe.configure call with both metadata and a locale, make sure that locale is always used over metadata (should it throw if we're trying to set both?)
2. Test that if you pass an object into the message function, the template ran correctly.

axe.d.ts
```diff
@@ -71,6 +71,24 @@ declare namespace axe {
 		target: string[];
 		html: string;
 	}
+	interface RuleLocale {
+		[key: string]: {
```
jkodu: Why are these optional attributes?

lib/core/base/audit.js
jkodu: +1

lib/core/base/audit.js
stephenmathieson: I think you're mistaking this method for a function expression. Because this is a method, we'll see its implicit name in stack traces.

Given the code:

```js
function F () {}

F.prototype.foo = function () {
  throw new Error('boom')
}

const f = new F
f.foo()
```

We'll clearly see "foo" in the stack trace:

```
node foo.js
/Users/stephen/dev/src/github.com/dequelabs/axe-core/foo.js:4
  throw new Error('boom')
  ^

Error: boom
    at F.foo (/Users/stephen/dev/src/github.com/dequelabs/axe-core/foo.js:4:9)
    at Object.<anonymous> (/Users/stephen/dev/src/github.com/dequelabs/axe-core/foo.js:8:3)
    at Module._compile (internal/modules/cjs/loader.js:702:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:713:10)
    at Module.load (internal/modules/cjs/loader.js:612:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:551:12)
    at Function.Module._load (internal/modules/cjs/loader.js:543:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:744:10)
    at startup (internal/bootstrap/node.js:238:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:572:3)
```

lib/core/base/audit.js
stephenmathieson: It can, and will be. The result of calling `dot.compile()` on the string is a function.

lib/core/base/audit.js
stephenmathieson: 👍 I'll spread `a` here.

test/core/public/configure.js
stephenmathieson: Tests have been added.

PR #1037, opened 2018-08-01
test: throw a helpful error when running alongside the extension

stephenmathieson:
Currently running the `axe-core` test suite in a browser that has either the axe or attest extension enabled will yield unpredictable and confusing results. This patch causes the test suite to throw an error instructing people to disable their extension(s) before running their tests.

I ran into this while working on a feature and wasted ~1 day trying to debug. :'(

Side note - I'm opening a seperate PR into the extensions repository which exposes the `__AXE_EXTENSION__` global.`

## Reviewer checks

- [x] Code is reviewed for security by: @isner

### DIFF
--- a/test/testutils.js
+++ b/test/testutils.js
@@ -1,3 +1,10 @@
+// Let the user know they need to disable their axe/attest extension before running the tests.
+if (window.__AXE_EXTENSION__) {
+	throw new Error(
+		'You must disable your axe/attest browser extension in order to run the test suite.'
+	);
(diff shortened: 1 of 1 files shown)


test/testutils.js
```diff
@@ -1,3 +1,10 @@
+// Let the user know they need to disable their axe/attest extension before running the tests.
```
isner: NBD, but the error is self-documenting.

PR #1047, opened 2018-08-05
Use jscodeshift to replace var with const/let

WilcoFiers:
Removed all vars from lib (except for imports). Mostly this was just to get a feeling of how to use jscodeshift. I think this can be a useful tool for modernising our codebase.

**Reviewer tip**: If you don't want to review 1000 file changes, look at the commits separately. I did a couple by hand that the codemod couldn't handle. I found that we were referencing some stuff in lib in our tests, which caused problems in Phantom. I fixed that by referencing tmp instead.

## Reviewer checks


lib/core/utils/collect-results-from-frames.js
```diff
@@ -16,15 +16,15 @@ function err(message, node) {
 axe.utils.sendCommandToFrame = function(node, parameters, resolve, reject) {
 	'use strict';
 
-	var win = node.contentWindow;
+	let win = node.contentWindow;
```
jkodu: trivial: this can be const, can be changed later.

WilcoFiers:
Updated, please review again.

stephenmathieson:
I'm -1 on this change.

We are unable to use `const` and `let` within the `./test` directory. Requiring users to use `const` and `let` within `./lib` provides additional mental overhead and will make contributing to this project even harder.

Once we're able to use `const` and `let` everywhere, I'm 100% on-board here. Until then, I don't think this is a good move.

dylanb:
@stephenmathieson as a first step in the process of getting to const and let everywhere, this is worthwhile. I vote for accepting this and then moving quickly to do the same for `./test`

WilcoFiers:
@stephendonner I've opened an issue #1059 for this. We're already in the situation that you're supposed to write ES6 in lib and ES3 in ./test. This doesn't change that. I agree with Dylan that we should get Babel running on ./test soon too.

stephenmathieson:
I think putting barriers in place and _forcing_ (rather than subtlety hinting) that people should treat the two areas of the codebase differently is going to just complicate things further. 

I don't understand the rush to get this in. This PR was mostly automated, and could easily be re-opened once `./test` supports modern JS. IMO its more of a problem that developers cannot `require()` modules or use their normal language syntax than a few `var` statements could ever be.

I won't be upset if you decide to pull rank here. I just want it to be heard that I think we're investing time solving the wrong problem right now.

marcysutton:
I agree with Stephen here. Not being able to `require` is a bigger problem than enforcing `let` and `const`...especially since those get transpiled to `var` anyway.

jkodu:
I see both @stephenmathieson and @marcysutton point here.
But I am +1 for slow progressive enhancement.

WilcoFiers:
OK, closing the PR. I'll rerun it when we've solved the ./test babel thing. A PR like this can get stale quickly. I did it mostly to get some practice with codeshift, which is something we'll probably need to modernise our code base.

WilcoFiers:
**Please leave the branch**

PR #1058, opened 2018-08-07
fix: improve test runner for logging and assertion on incomplete errors

jkodu:
This is an improvement to current test runner, by enabling `debug: true` in `axe.run` and further more assertions on the results. These changes help with catching errors with better logging especially with tests run in webdriver.

Closes issue: NA

## Reviewer checks


test/integration/rules/runner.js
```diff
@@ -102,12 +102,27 @@
...
 									assert.isNull(err);
+									// assert that result is defined
+									assert.isDefined(r);
+									// assert that result has certain keys
+									assert.hasAnyKeys(r, ['incomplete', 'violations', 'passes']);
+									// flatten results
 									results = flattenResult(r);
+									// assert incomplete in results does not have error
+									if (results.incomplete) {
+										assert.isUndefined(results.incomplete.error);
```
WilcoFiers: results.incomplete should be an array here, no? So shouldn't this be:

```js
results.incomplete.forEach(function (result) {
  assert.isUndefined(results.error)
});
```

test/integration/rules/runner.js
```diff
@@ -102,12 +102,27 @@
 							axe.run(
 								fixture,
 								{
+									/**
+									 * The debug flag helps log errors in a fairly detailed fashion,
+									 * when tests fail in webdriver
+									 */
+									debug: true,
```
WilcoFiers: This prevents errors from getting passed to `results.incomplete` and for axe.run to throw. We can do one or the other solution (whichever gives the clearer error), but not both.

jkodu: That is not true.

If any of the assertions fail (with or without this flag), the current set-up does not give a detailed logging on which part of the code failed. Enabling `debug`, prints the line number in question for the failure with respect to the library, and not just the chai assertion.

Also, if an assertion fails I would expect to throw (for example like below)
image

I have tested this with both enabling/ disabling the debug flag and is working as expected.

test/integration/rules/runner.js
jkodu: Well, as a part of flattenResult in L121, we pluck the first item in the incomplete array.

If we were to move the assertion before flattening the results, we have to `forEach`. No harm in either way. Shall refactor.

jkodu:
@WilcoFiers - please review again.

WilcoFiers:
@jkodu The commit messages on this PR are incorrect. Please keep in mind, anything done to repo infrastructure is "chore". If it doesn't impact the end user, it's not a fix.

stephenmathieson:
I think the "test" scope would make more sense, or even "build". I'm of the opinion we're using "chore" far too frequently

stephenmathieson:
Also, we can easily correct the message when squashing... :wink:

Issue #1045, opened 2018-08-03
Failure message for meta-viewport rule is unclear


liloda:
One of our users found this failure message confusing:
Fix the following: \<meta> tag disables zooming on mobile devices

Suggest mentioning the attribute that actually causes the problem (user-scalable=no), something like this:
Fix the following:  user-scalable="no" on \<meta> tag disables zooming

marcysutton:
I submitted a PR to fix this: #1061

PR #1061, opened 2018-08-07
fix(meta-viewport): show attribute in message

marcysutton:
A user reported the meta-viewport check had a confusing failure message, so I changed it to pass along the failing attribute:

"Fix the following: user-scalable="no" on <meta> tag disables zooming"

Closes https://github.com/dequelabs/axe-core/issues/1045

## Reviewer checks

### DIFF
--- a/lib/checks/mobile/meta-viewport-scale.js
+++ b/lib/checks/mobile/meta-viewport-scale.js
@@ -26,10 +26,12 @@ if (
 }
 
 if (!lowerBound && result['user-scalable'] === 'no') {
+	this.data('user-scalable=no');
 	return false;
 }
 
 if (result['maximum-scale'] && parseFloat(result['maximum-scale']) < minimum) {
+	this.data('maximum-scale');
 	return false;
 }
 
--- a/lib/checks/mobile/meta-viewport.json
+++ b/lib/checks/mobile/meta-viewport.json
@@ -8,7 +8,7 @@
     "impact": "critical",
     "messages": {
       "pass": "<meta> tag does not disable zooming on mobile devices",
-      "fail": "<meta> tag disables zooming on mobile devices"
+      "fail": "{{=it.data}} on <meta> tag disables zooming on mobile devices"
     }
   }
 }
--- a/test/checks/mobile/meta-viewport-scale.js
+++ b/test/checks/mobile/meta-viewport-scale.js
@@ -2,9 +2,11 @@ describe('meta-viewport', function() {
 	'use strict';
 
 	var fixture = document.getElementById('fixture');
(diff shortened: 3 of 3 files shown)


lib/checks/mobile/meta-viewport-scale.js
```diff
@@ -26,10 +26,12 @@ if (
 }
 
 if (!lowerBound && result['user-scalable'] === 'no') {
+	this.data(['user-scalable=no']);
```
jkodu: Why is this an array? Is so should we do a join in the doT templates?

lib/checks/mobile/meta-viewport-scale.js
```diff
@@ -26,10 +26,12 @@ if (
...
 	return false;
 }
 
 if (result['maximum-scale'] && parseFloat(result['maximum-scale']) < minimum) {
+	this.data(['maximum-scale']);
```
jkodu: Are we saying if `user-scalable=no` is already set on data, `maximum-scale` will then override the data? Should we not concat?

lib/checks/mobile/meta-viewport.json
```diff
@@ -8,7 +8,7 @@
     "impact": "critical",
     "messages": {
       "pass": "<meta> tag does not disable zooming on mobile devices",
-      "fail": "<meta> tag disables zooming on mobile devices"
+      "fail": "{{~it.data:value}} {{=value}}{{~}} on <meta> tag disables zooming on mobile devices"
```
jkodu: data seems to be an array, values should be joined here to construct the string...

lib/checks/mobile/meta-viewport-scale.js
marcysutton: It returns false. There can only be one result.

lib/checks/mobile/meta-viewport-scale.js
marcysutton: I'd followed another example, there can only be one result but that looked like what doT was expecting.

lib/checks/mobile/meta-viewport.json
marcysutton: Changed it to a string.

WilcoFiers: We need to do `{{=it.data}}` otherwise you get this:

screenshot 2018-08-09 11 18 22

marcysutton:
@WilcoFiers good catch, I've updated the PR and rebased from develop.

dylanb:
@marcysutton seems like we should add some tests to ensure that the appropriate data is coming through?

marcysutton:
@dylanb not a bad idea, do you know of any existing places in the codebase where we test the messages?

Issue #918, opened 2018-05-25
Add ATSupported checks to ARIA rules


WilcoFiers:
One of the difficulties with the ARIA rules is that it is unclear if an ARIA attribute / role wasn't supported, or if they made had a typo. To solve this we should include all available ARIA attributes and add checks that flag unsupported properties as violations. This has the added benefit of making it easy to ignore unsupported properties by simply disabling these checks.

WilcoFiers:
On second thought, perhaps a separate rule instead... Thoughts?

marcysutton:
This would help to differentiate between these two scenarios, which are currently pretty cryptic:
- attribute refers to an invalid or non-existing ID attribute
- the attribute is completely invalid

marcysutton:
> include all available ARIA attributes 
Like adding everything from the ARIA spec to the lookupTable? We can't rely on aria-query completely due to its reliance on Sets, IIRC?

> add checks that flag unsupported properties as violations
It seems like the actual marking of support should live in the lookupTable, and the checks simply look up those flags and report them to the user. Disabling the new checks would put axe back to how it operates now.

Is there anything else you want to add @WilcoFiers?

WilcoFiers:
I don't have it all worked out... I do think we should use aria-query. We can't use it directly, but there's probably a way that we can use it to build an object structure that we can safely include into axe.

As for where the support data should be. Probably somewhere as a data structure in `commons.aria`. I think it should probably be its own structure, rather than weave it into lookuptable, so that it's easier to configure.

If you're not sure how to proceed, maybe a quick design doc would help.

WilcoFiers:
Some more thoughts. I think we should turn whatever we do into one, or possibly multiple JSON files, rather than have them live in JS code. All this is static data, lets store it that way.

WilcoFiers:
I'm liking this plan a lot :-). One tweak to make to it. for `axe.commons.support()`, use it as a setter. It could work something like this:

```js
axe.commons.support({
  roles: { figure: false }
})
```

marcysutton:
Here's something I found that could be related: the [`aria-allowed-attr-matches`](https://github.com/dequelabs/axe-core/blob/develop/lib/rules/aria-allowed-attr-matches.js) file requires some kind of role to find out whether an attribute is allowed, which is probably too granular for the new `unsupported-attr` check. I think it would be good to look at all ARIA attributes present on the node for support regardless of role, because people might put them wherever.

[...]

PR #1064, opened 2018-08-10
feat: flag unsupported roles

marcysutton:
Adds `unsupported` property to the `lookupTable.role` dictionary. New option for `aria.isValidRole` to `flagUnsupported` roles. Defaults to `false` since this commons function is reused for multiple checks.

Part one of https://github.com/dequelabs/axe-core/issues/918.

TODO: 
- Add `unsupported` to lookupTable.attributes
- Figure out what to do with globalAttributes, which are currently a flat array
- Add `axe.commons.support` API and enable control of it through `axe.configure`

## Reviewer checks

### DIFF
--- a/build/tasks/aria-supported.js
+++ b/build/tasks/aria-supported.js
@@ -46,18 +46,30 @@ module.exports = function(grunt) {
 					.reduce((out, [key] = item) => {
 						switch (listType) {
 							case 'supported':
-								if (subject.includes(key)) {
+								if (
+									subject.hasOwnProperty(key) &&
+									subject[key].unsupported === false
+								) {
 									out.push([`${key}`, 'Yes']);
 								}
 								break;
 							case 'unsupported':
-								if (!subject.includes(key)) {
+								if (
+									(subject[key] && subject[key].unsupported === true) ||
+									!subject.hasOwnProperty(key)
+								) {
 									out.push([`${key}`, 'No']);
 								}
 								break;
 							case 'all':
 							default:
-								out.push([`${key}`, subject.includes(key) ? 'Yes' : 'No']);
+								out.push([
+									`${key}`,
+									subject.hasOwnProperty(key) &&
+									subject[key].unsupported === false
+										? 'Yes'
+										: 'No'
+								]);
 								break;
 						}
 						return out;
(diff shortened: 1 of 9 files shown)


build/tasks/aria-supported.js
```diff
@@ -46,18 +46,30 @@ module.exports = function(grunt) {
...
+									!subject.hasOwnProperty(key)
+								) {
 									out.push([`${key}`, 'No']);
 								}
 								break;
 							case 'all':
 							default:
-								out.push([`${key}`, subject.includes(key) ? 'Yes' : 'No']);
+								out.push([
+									`${key}`,
```
stephenmathieson: What's this template string for? This could just be `out.push(key, ...)`.

lib/checks/aria/unsupportedrole.json
```diff
@@ -0,0 +1,11 @@
+{
+  "id": "unsupportedrole",
```
stephenmathieson: Should this be hyphenated? `unsupported-role` is more ascetically pleasing and easier to read (IMO).

stephenmathieson: Ah, makes sense. I'm +1 for consistency, so no need to change

build/tasks/aria-supported.js
jkodu: @stephenmathieson @dylanb 
`out` is what eventually becomes the array to pass to mdTable like so - https://github.com/dequelabs/axe-core/blob/develop/build/tasks/aria-supported.js#L78

And the generated output is here - https://github.com/dequelabs/axe-core/blob/develop/doc/aria-supported.md

stephenmathieson: Right, but why are we wrapping a variable in backticks here? Consider the following:

```js
var a = `${out}`
var b = out
```

lib/checks/aria/unsupportedrole.json
```diff
@@ -0,0 +1,11 @@
...
+  "evaluate": "unsupportedrole.js",
+  "metadata": {
+    "impact": "critical",
+    "messages": {
+      "pass": "ARIA role is supported",
+      "fail": "The role used is not widely supported in assistive technologies"
```
WilcoFiers: Should this say "in screen readers", or "in screen readers and other assistive technologies"?

lib/commons/aria/index.js
```diff
@@ -771,7 +828,12 @@ lookupTable.role = {
 			one: ['article']
 		},
 		nameFrom: ['author'],
-		context: null
+		context: null,
+		unsupported: false
+	},
+	figure: {
```
WilcoFiers: I think we should add the other properties here too.

jkodu:
@WilcoFiers 
Could we merge this after - https://github.com/dequelabs/axe-core/pull/945 as otherwise the roles lookupTable merge conflicts is going to be fun.

Issue #1062, opened 2018-08-08
Create a way to give deprecation warnings


WilcoFiers:
Currently the only option is to do axe.log() and give deprecation warnings that way. The problem with that is that those messages may not be exposed to the users in all situations, or worse, they may become part of the output stream and thus making it unappeasable. We need a standard solution that different integrations can opt into. Either make it part of the results object somehow, or have it be a callback you can pass into axe? I'm not sure about the solution yet, open to suggestions.

WilcoFiers:
We've largely deprecated use of custom evaluate / matches functions so this isn't as meaningful as it used to be. Tools have also gotten better about this, so IDEs can give these warnings for us. I don't think this ticket is important anymore.

PR #1071, opened 2018-08-15
feat: Deprecate audio-caption rule

### LINKED ISSUE #997: aXe violation for `<audio>` element without caption should be a "need inspection"
aXe flags an `<audio>` element without an associated `<track>` element as a violation. 
This is not always the case.

An `<audio>` element requires a text transcript, not synchronized captions.
That transcript can be provided in a myriad of ways other than using the `<track>` element, e.g. by displaying it in HTML on the same page or even as a download link adjacent to the `<audio`> element.

There are also situations where the `<audio>` element itself does not need a transcript at all such as when it is used as an accessible aalternative to a visual cue.

We use an `<audio>` element to play a popping sound as an incoming chat message notification for a screen reader user. That does not need a caption.

I propose to revise this requirement to "needs verification".

WilcoFiers:
Disabling / deprecating this rule. Audio captions do not work (https://www.iandevlin.com/blog/2015/12/html5/webvtt-and-audio/), and there are other ways to make audio accessible.

Closes #997.

## Reviewer checks


jkodu:
The problem with `axe-codeowners` is that, when someone outside of the team approves you do not pass the `waiting on code review` check. The status still remains as `review required`.

WilcoFiers:
@dylanb There's a ticket open #1062 to figure out how to notify users of deprecated stuff in axe-core.

PR #1075, opened 2018-08-15
fix: Do not require media captions / descriptions

### LINKED ISSUE #816: Caption files embedded inside the media wrapper should not fail
Currently it appears that aXe-core is only looking for the presence of the <track> element to deliver caption files, but there is a second method of doing so: place the h.265 encoded video and the AAC encoded audio AND the Vtt/TTML file all inside of the media wrapper (traditionally either .mp4 or .mkv files). 

There is an existing API that "looks inside" the wrapper when being rendered onscreen, which looks for those files to deliver the captioning. Ideally, aXe-core would use a similar method to detect whether the time-stamped file is bundled inside of the media wrapper, as a second test (after checking for the presence of <track> as it does today)

WilcoFiers:
Do not require media captions / descriptions
Closes #816 

## Reviewer checks


build/tasks/validate.js
```diff
@@ -27,6 +27,27 @@ function hasUniqueId() {
 	};
 }
 
+function hasTwoOutcomes(messages) {
```
jkodu: Looks like this is normalising multiple outcomes, should change function name to something generic like `normaliseOutcomes` or `hasMultipleOutcomes` rather than exactly `Two`?

build/tasks/validate.js
```diff
@@ -84,16 +105,20 @@ function createSchemas() {
...
-							},
-							pass: {
-								required: true,
-								type: 'string'
-							}
+						conform: hasTwoOutcomes,
+						messages: {
+							conform: 'Must have at least two valid messages'
 						}
+						// propferties: {
```
jkodu: minor: 🤣 `propferties`, know it is commented.... delete instead?

dylanb: yeah, delete this comment

lib/checks/media/caption.json
```diff
@@ -5,7 +5,6 @@
     "impact": "critical",
     "messages": {
       "pass": "The multimedia element has a captions track",
-      "fail": "The multimedia element does not have a captions track",
       "incomplete": "A captions track for this element could not be found"
```
dylanb: Could make this message more explicit about what to do to check whether an alternative exists

WilcoFiers:
@dylanb @jkodu Updated the PR.

PR #1076, opened 2018-08-15
fix: Allow divs as groups in dl

### LINKED ISSUE #262: Support expanded definition list rule in the WHATWG HTML spec which allows for divs as parents of dt/dds
I was wondering if the definition list rule would or should ever be expanded to the WHATWG HTML spec which seems to allow for divs as parents of dt/dds (https://html.spec.whatwg.org/multipage/semantics.html#the-dl-element) if screen readers/browsers support this (I don't know if they do)? I imagine the WCAG is strict to the W3C's HTML5 spec, but don't know for sure. 

I've had this come up and trying to figure out how to avoid disabling this rule or find some other semantic markup for the relationships given the divs can't be changed.

WilcoFiers:
Allow div elements to group content inside of dl elements.

Closes #262

## Reviewer checks

- [x] Code is reviewed for security by: Jey


test/checks/lists/only-dlitems.js
```diff
@@ -176,6 +176,24 @@ describe('only-dlitems', function() {
...
+	it('should return false if the list has dt and dd inside a div group', function() {
+		var checkArgs = checkSetup(
+			'<dl id="target"><div><dt>An item</dt><dd>A list</dd></div></dl>'
+		);
+		assert.isFalse(
+			checks['only-dlitems'].evaluate.apply(checkContext, checkArgs)
+		);
+	});
+
+	it('should return true if the list has dt and dd inside a div group with a role', function() {
```
dylanb: What about a test with other content inside the div?

WilcoFiers: Resolving in a new PR

PR #1087, opened 2018-08-21
ci: automate publishing releases to npm

stephenmathieson:
This patch adds configuration for automatically publishing to npm, pending a manual hold. To accept the hold, a user with write access must login to CircleCI and click the "Approve" button.

All (green) commits to the `develop` branch will result in a "next" version being published to npm. The published version will resemble `X.Y.Z-next.COMMITSH`, where `X.Y.Z` are the existing SEMVER numbers, and `COMMITSIH` is the Git SHA1 for the commit. The version is published with the [dist tag](https://docs.npmjs.com/cli/dist-tag) "next". This enables users to `npm install axe-core@next` to get the latest version on GitHub.

All (green) commits to the `master` branch will result in a production version being published to npm. It is assumed that `package.json`'s version field will be updated as part of each commit to `master` (CircleCI does not do this for us).

For a further explanation, this process is documented in [Google Drive](https://docs.google.com/document/d/14S2ktHpBVNPAR7JdkptOOgmufbZIV0m7orz6N95tPzk) (private whitepaper, sorry OSS folks).

## Reviewer checks

### DIFF
--- a/build/canary.js
+++ b/build/canary.js
@@ -1,0 +1,23 @@
+#!/usr/bin/env node
+
+const fs = require('fs');
+const path = require('path');
+const assert = require('assert');
+
+const pkgFile = path.resolve(__dirname, '..', 'package.json');
+const pkg = JSON.parse(fs.readFileSync(pkgFile));
+
+const { CIRCLE_SHA1 } = process.env;
+assert(CIRCLE_SHA1, 'CIRCLE_SHA1 enviornment variable not set');
+
+// Shorten the SHA
+const GIT_SHA = CIRCLE_SHA1.substr(0, 7);
+
+// Strip the "dist tag" from the version (if it exists)
+const version = pkg.version.replace(/-\w+\.\d$/, '');
+const nextVersion = `${version}-canary.${GIT_SHA}`;
+
+pkg.version = nextVersion;
+fs.writeFileSync(pkgFile, JSON.stringify(pkg, null, 2));
+
+console.log(`Updated version to ${nextVersion}`);
--- a/circle.yml
+++ b/circle.yml
@@ -36,15 +36,15 @@ jobs:
       - <<: *restore_dependency_cache
       - run: npm run test
 
-  # Release a "canary" version
-  canary_release:
+  # Release a "next" version
+  next_release:
     <<: *defaults
     steps:
       - checkout
       - <<: *set_npm_auth
       - <<: *restore_dependency_cache
-      - run: build/canary.js
-      - run: npm publish --tag=canary
+      - run: build/next-version.js
+      - run: npm publish --tag=next
 
   # Release a "production" version
   production_release:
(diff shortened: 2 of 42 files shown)


circle.yml
```diff
@@ -1,17 +1,84 @@
...
+
+  # Release a "canary" version
+  canary_release:
+    <<: *defaults
+    steps:
+      - checkout
+      - <<: *set_npm_auth
+      - <<: *restore_dependency_cache
+      - run: build/canary.js
+      - run: npm publish --tag=canary
```
WilcoFiers: I think this needs to be `--tag=next`. That's certainly how we've done it so far, and it's what axe Coconut uses.

build/canary.js
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env node
+
+const fs = require('fs');
+const path = require('path');
+const assert = require('assert');
+
+const pkgFile = path.resolve(__dirname, '..', 'package.json');
+const pkg = JSON.parse(fs.readFileSync(pkgFile));
```
WilcoFiers: Should we run standard-version instead? It updates a bunch more stuff, including package-lock, bower.json and the changelog.

build/canary.js
```diff
@@ -0,0 +1,23 @@
...
+
+const { CIRCLE_SHA1 } = process.env;
+assert(CIRCLE_SHA1, 'CIRCLE_SHA1 enviornment variable not set');
+
+// Shorten the SHA
+const GIT_SHA = CIRCLE_SHA1.substr(0, 7);
```
jkodu: Only thing that would be nice, is to auto generate a readme file with a link back to that commit in develop.

And if like @WilcoFiers suggests if we are able to get the change log, add that to the readme as well, as otherwise the numerous packages on npm will soon go out of context and someone has to look through commit history to figure out what each version has...

circle.yml
stephenmathieson: Sure, will update. It'd be nice if we caught these things in the planning document rather than during implementation, but oh well 🙃

build/canary.js
stephenmathieson: This repository doesn't use `package-lock.json`, and I wasn't aware Bower was still a thing, haha. 😆 

What would this look like with `standard-version`? I've never used the tool before.

Also, this too would have been nice to discuss in planning rather than during implementation.

build/canary.js
stephenmathieson: I don't know why we'd add something to a README. We don't do that for "regular" versions, and it's not something I've seen other packages doing. I understand the desire to update the CHANGELOG, but think that doing so makes the README update unnecessary.

This said, if we need to do this, I'm happy to make the change if it's decided that we need it.

jkodu: Ha ha, I will leave it with @WilcoFiers to decide if we need a README or not. 

We can always get this out and enhance as needed. Not a blocker in anyway.

PR #1097, opened 2018-08-24
feat: Break up duplicate-id rule for ARIA+labels and for active elements

### LINKED ISSUE #163: More impactful duplicate ID check option
Duplicate IDs are only an accessibility issue under two conditions:
1. The item is an interactive element - in this case, the duplicate ID could cause assistive technology (such as Dragon) to target the incorrect item, resulting in potentially irreversible adverse actions, or
2. The item is referenced through one of the ID reference techniques. In this case, the wrong element could be referenced

Change the duplicate ID test, to have an option that will cause it to only report on interactive elements (testing the entire DOM to see whether there are references to an element is too slow)

WilcoFiers:
Break the duplicate-id rule up into three rules:

- duplicate-id-aria: Test the IDs referenced with aria and with label[for] are unique
- duplicate-id-active: Test that focusable elements have a unique ID
- duplicate-id: Check that all other IDs are unique

Closes #163 

## Reviewer checks


test/integration/rules/duplicate-id-active/duplicate-id-active.html
```diff
@@ -0,0 +1,19 @@
+<p id="ignore1">This is my first paragraph with this ID.</p>
+<div style="display:none">
+  <p id="ignore1">This is my second paragraph with this ID.</p>
+</div>
+
+<button id="fail1" class="fail1"></button>
+<button id="fail1"></button>
```
dylanb: Should we have more examples of failures? For example a `<div tabindex="-1" id="fail2">blah</div><input id="fail2" />` etc.?

paulirish:
It took us a while to find it, but thanks for doing this. 

We really like this approach and agree it's got much higher signal:noise. :)

PR #1098, opened 2018-08-24
fix: Allow only-dlitem / only-listitem to have any hidden content

### LINKED ISSUE #1021: only-listitems and only-dlitems pick up hidden <dom-repeat> as disallowed child

WilcoFiers:
This change allows any hidden element in ul, ol and dl elements.

Closes #1021 

## Reviewer checks


test/checks/lists/only-dlitems.js
```diff
@@ -194,6 +194,42 @@ describe('only-dlitems', function() {
...
+	it('returns false if there are visibility:hidden elements that normally would not be allowed', function() {
+		var checkArgs = checkSetup(
+			'<dl id="target"> <dt>An item</dt> <dd>A list</dd> <h1 style="visibility:hidden">heading</h1> </dl>'
+		);
+		assert.isFalse(
+			checks['only-dlitems'].evaluate.apply(checkContext, checkArgs)
+		);
+	});
+
+	it('returns false if there are aria-hidden=true elements that normally would not be allowed', function() {
```
dylanb: hmmm, I don't agree with this case

test/checks/lists/only-listitems.js
```diff
@@ -189,6 +189,42 @@ describe('only-listitems', function() {
...
+	it('returns false if there are visibility:hidden elements that normally would not be allowed', function() {
+		var checkArgs = checkSetup(
+			'<ul id="target"> <li>An item</li> <h1 style="visibility:hidden">heading</h1> </ul>'
+		);
+		assert.isFalse(
+			checks['only-listitems'].evaluate.apply(checkContext, checkArgs)
+		);
+	});
+
+	it('returns false if there are aria-hidden=true elements that normally would not be allowed', function() {
```
dylanb: I disagree with this

dylanb:
ok, I was mistaken in thinking this was a 4.1.1 rule and not a 1.3.1 rule

Issue #1108, opened 2018-08-28
No banner comments in `axe.min.js` (PR #1112)


jkodu:
Some time between releases `3.0.2` and `3.1.1` the banner comments on `axe.js` and `axe.min.js` have been lost.

```js
/*! aXe v3.0.2
 * Copyright (c) 2018 Deque Systems, Inc.
 *
 * Your use of this Source Code Form is subject to the terms of the Mozilla Public
 * License, v. 2.0. If a copy of the MPL was not distributed with this
 * file, You can obtain one at http://mozilla.org/MPL/2.0/.
 *
 * This entire copyright notice must appear in every copy of this file you
 * distribute or in any file that contains substantial portions of this source
 * code.
 */
```

This caused `rspec` and `cucumber` in `axe-matchers` and `attest-ruby` tests which rely on `jslib` injection to fail. 
```ruby
expect(page).to have_received(:execute_script).with(a_string_starting_with ("/*! aXe"))

its(:jslib) { is_expected.to start_with("/*! aXe") }
```

Analysis:
v3.0.2 - Has banner comments.
v3.0.3 - Neither files have banner comments.
v3.1.1 - `axe.min.js` is missing the banner comments.

jkodu:
I have worked around the test cases to verify other strings in those respective ruby test cases, so no immediate work is necessary for this.

PR #1112, opened 2018-08-29
fix: banner comment in generated axe files

jkodu:
- This PR ensures, generation of banner comments on `axe.js` and `axe.min.js` files as a part of `grunt build` step.
- `grunt-contrib-uglify` > 2.x has some API changes which replaced `preserveComments` with `options.comments` which is now adapted here.

Closes issue:
- https://github.com/dequelabs/axe-core/issues/1108

## Reviewer checks

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -231,7 +231,7 @@ module.exports = function(grunt) {
 						quote_style: 1
 					},
 					output: {
-						comments: /^!/
+						comments: /^\/*! aXe/
 					}
 				}
 			},
@@ -243,9 +243,8 @@ module.exports = function(grunt) {
 					};
 				}),
 				options: {
-					preserveComments: function(node, comment) {
-						// preserve comments that start with a bang
-						return /^!/.test(comment.value);
+					output: {
+						comments: /^\/*! aXe/
 					},
 					mangle: {
 						reserved: ['commons', 'utils', 'axe', 'window', 'document']


dylanb:
@jkodu Any way we can add an integration test for this? Seems like this keeps happening and it would be good to catch it in the build pre-publish.

stephenmathieson:
An integration test seems like overkill to me, but this should work:

```bash
#!/bin/bash

e="/*! aXe"
for file in axe.js axe.min.js; do
  src=$(cat $file)
  if [[ $src != $e* ]]; then
    echo "$file does not contain required comment" >&2 
    exit 1
  fi
done 
```

jkodu:
@dylanb @stephenmathieson 

Sure thing, can be done, but I will tackle that one (writing tests) separately as I need to write some tests to ensure these files are generated in the first place, before verifying the contents for comments etc. 

This was the agreed solution we came up in the retrospective to have tests around generated files and their timestamps, to avoid the issue we had with releasing axe earlier this week without generated files. @WilcoFiers and I believe this needs to be across all the projects, so will get around to that later.

WilcoFiers:
@jkodu In thinking some more about build processes. I don't think we need prepublish tests for all projects, just the ones with complicated build steps. For some it might just be enough to check that the `dist` directory is there. Either way, please create a ticket to add a prepublish test to axe-core.

jkodu:
Issue created to ensure build generated files exist - https://github.com/dequelabs/axe-core/issues/1117

Issue #1082, opened 2018-08-21
CSSOM: Safari not returning shadowRoot stylesheets (PR #1113)


jkodu:
In Safari, shadowRoot node does not support `.styleSheets` property.
A work around needs to be put in place.

jkodu:
PR #1113

Issue #1078, opened 2018-08-17
CSSOM: Handle Priority of styles applied (PR #1113)


jkodu:
Based on this comment.
https://github.com/dequelabs/axe-core/pull/971/files#r210937594

Need to work on a solution to figure out priority of styles loaded, to check for negations of style overrides if any...

jkodu:
PR - https://github.com/dequelabs/axe-core/pull/1113

PR #1113, opened 2018-08-29
fix: CSSOM generation for shadowRoot in Safari

jkodu:
This PR handles the edge case where for a given shadowRoot, the `root.styleSheets` lookup is returned as `undefined | null`.

This meant evaluation of styles with in shadowDOM failed.

This fix, tackles the edge case by parsing the content of the shadowRoot, to reverse engineer `styleSheets` from `style` tag if any specified with in the shadowRoot.

Upon reverse engineering the sheets, it is lead to flow into the generic `loadCssom` function to do the necessary generation of CSSOM as intended.

The concept of priority in stylesheets loaded is also introduced. The priority computation will be enhanced when handling multiple nested `@import` stylesheets in a separate PR after this lands.

Closes issue:
- https://github.com/dequelabs/axe-core/issues/1082
- https://github.com/dequelabs/axe-core/issues/1078

## Reviewer checks


jkodu:
@WilcoFiers 
Updated with handling `link` as requested as well.
Have also introduced the concept of `priority`, but needs more work to tackle other scenarios and enhancing `css-orientation-lock` tests.

Please review this again.

jkodu:
@dylanb - review appreciated.

lib/core/utils/preload-cssom.js
```diff
@@ -75,14 +149,24 @@ function loadCssom({ root, shadowId }, timeout, convertTextToStylesheetFn) {
 	});
 
 	// iterate to decipher multi-level nested sheets if any (this is essential to retrieve styles from shadowDOM)
-	sheets.forEach(sheet => {
+	sheets.forEach((sheet, sheetIndex) => {
+		/* eslint max-statements: ["error", 20] */
+
+		// basic priority
+		const sheetLevelPriority = `${rootIndex}${sheetIndex}`;
```
stephenmathieson: I don't think I understand what's happening here. It looks like we're concatenating (not adding) numbers together in a string to be used as a numerical value later on. Can you explain how this works?

I'm stuck on the following:

A "priority" string is created with a `rootIndex` of `1` and a `sheetIndex` of `3000`. It is then coerced to the integer `13000`.

Later on, another string is created with `rootIndex` of `2` and `sheetIndex` of `1`. It's then coerced to `21`.

`13000` is larger than `21`, but its "root" was higher. Is this OK?

lib/core/utils/preload-cssom.js
```diff
@@ -38,11 +38,38 @@ function loadCssom({ root, shadowId }, timeout, convertTextToStylesheetFn) {
...
-	// handle .styleSheets non existent on certain shadowDOM root
-	const rootStyleSheets = root.styleSheets
-		? Array.from(root.styleSheets)
-		: null;
-	if (!rootStyleSheets) {
+	let styleSheets = null;
+
+	if (!root.styleSheets) {
+		/**
+		 * In Safari: (Issue: https://github.com/dequelabs/axe-core/issues/1082)
```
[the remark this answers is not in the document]
jkodu: Reported, but do not have an issue/ bug reported link from them yet.

lib/core/utils/preload-cssom.js
```diff
@@ -1,48 +1,122 @@
...
-					data,
-					isExternal: true,
-					shadowId,
-					root
+	if (
+		rootDocument.nodeName.toUpperCase() === '#DOCUMENT-FRAGMENT' &&
+		shadowId
+	) {
+		// retrieve shadowRoot style sheets as []
+		styleSheets = Array.from(rootDocument.children).reduce((out, node) => {
```
WilcoFiers: Why are you only looking at child nodes, instead of doing getElementsByTagName for `style` and `link` elements? It might make sense to break up the function into different load methods depending on the elements. Reduces the complexity of this function.

lib/core/utils/preload-cssom.js
```diff
@@ -75,14 +149,24 @@ function loadCssom({ root, shadowId }, timeout, convertTextToStylesheetFn) {
 	});
 
 	// iterate to decipher multi-level nested sheets if any (this is essential to retrieve styles from shadowDOM)
-	sheets.forEach(sheet => {
+	sheets.forEach((sheet, sheetIndex) => {
+		/* eslint max-statements: ["error", 20] */
+
+		// padding sheet index with root index to create priority
+		const sheetLevelPriority = Number(`${rootIndex}.${sheetIndex}`);
```
WilcoFiers: Doesn't this break if we've got more than 10 sheets? Is this priority something you came up with or is it part of a spec some place? This seems fragile.

lib/core/utils/preload-cssom.js
```diff
@@ -158,7 +269,7 @@ function getAllRootsInTree(tree) {
 		.map(node => {
 			return {
 				shadowId: node.shadowId,
-				root: axe.utils.getRootNode(node.actualNode)
+				rootDocument: axe.utils.getRootNode(node.actualNode)
```
WilcoFiers: Maybe stick with rootNode? That's what we're calling it everywhere else.

lib/core/utils/preload-cssom.js
```diff
@@ -38,11 +38,38 @@ function loadCssom({ root, shadowId }, timeout, convertTextToStylesheetFn) {
...
+		 * - Hence lookup all `<style>` tag(s) with in the shadowRoot and convert to styleSheet dynamically
+		 * - Pass these converted styleSheets to parse and construct cssom of shadowRoot(s)
+		 */
+		styleSheets = Array.from(root.children)
+			.filter(node => {
+				return node.nodeName.toUpperCase() === 'STYLE';
+			})
+			.map(node => {
+				const sheet = convertTextToStylesheetFn({
+					data: node.textContent
```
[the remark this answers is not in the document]
jkodu: Works, just fine as shown here: https://github.com/dequelabs/axe-core/pull/1113/files#diff-b195c7e1a05c62e60347f579d8f7dddeR339

It was the `style` tag that needed special parsing as above.

lib/core/utils/preload-cssom.js
```diff
@@ -1,48 +1,113 @@
...
+ * @property {Object} arg.rootNode document or shadowDOM root document for which to process CSSOM
+ * @property {Number} arg.rootIndex a number representing the index of the document or shadowDOM, used for priority
+ * @property {String} arg.shadowId an id if undefined denotes that given root is a shadowRoot
+ * @property {Number} timeout abort duration for network request
+ * @param {Function} getStyleSheet a utility function to generate a style sheet for a given text content
+ * @return {Object} queue
+
  */
-function loadCssom({ root, shadowId }, timeout, convertTextToStylesheetFn) {
+function loadCssom({ rootNode, rootIndex, shadowId, timeout, getStyleSheet }) {
```
WilcoFiers: Can you please break this function up into more readable chunks? It's a 200 line function that does half a dozen different things.

jkodu:
@WilcoFiers 
Thanks for the refactor suggestions, embraced most of it. 

The suggestion as per this snippet https://github.com/dequelabs/axe-core/pull/1113#discussion_r222249129, to use `sheetLoader` which then is `deferred` into the `queue` is not as simple, so left that section after minimal refactor.

Reviews welcome.

lib/core/utils/preload-cssom.js
```diff
@@ -1,58 +1,82 @@
...
+		shadowId,
+		priority
+	} = options;
+	axe.imports
+		.axios({
+			method: 'get',
+			url,
+			timeout
+		})
+		.then(({ data }) => {
```
stephenmathieson: I don't recall off the top of my head. Does Axios treat non-2xx response codes as errors? If not, we might want to add something like:

```js
.then(({ data, status }) => {
  if ((status / 100 | 0) !== 2) {
    throw new Error('non-2xx response')
  }
  // ...
})
```

WilcoFiers: Docs say that it does.

Issue #1116, opened 2018-08-30
aria-allowed-role with custom autocomplete (combobox + listbox, aria 1.0 vs aria 1.1) (PR #1118)


georgiee:
An error newly introduced during our housekeeping. `axe-core` was upgraded from `3.0.3` to `3.1.1`.

The following markup got identified as invalid (shorten)
```
"<input aria-autocomplete="list" aria-label="some label" autocomplete="off"  role="combobox" type="text" aria-expanded="true" aria-owns="autocomplete-0">"

<!-- somewhere else-->
<div id='autocomplete-0'>[...]</div>
```

Report from CLI looks like this (shorten):
```
id:"aria-allowed-role",
impact: "minor",
helpUrl: "https://dequeuniversity.com/rules/axe/3.1/aria-allowed-role?application=axeAPI",
nodes: [{
  failureSummary: "Fix any of the following: role combobox  is not allowed for given element",
  html: '...',
  impact: "minor"
}]
```

The linked help resource by axe cli is the following:
https://dequeuniversity.com/rules/axe/3.1/aria-allowed-role?application=axeAPI

More helping is this URL:
https://www.w3.org/TR/wai-aria-1.1/#combobox

So the problem is the role `combobox` directly on the input. With aria 1.1 it must be on the parent while the text should get the role textbox or searchbox.

The linked w3.org document also says the following.
> The ARIA 1.0 specification describes a combobox pattern where a text input element has the combobox role and owns a listbox element. User agents, assistive technologies, and conformance checkers SHOULD continue to support the ARIA 1.0 pattern so that existing implementations of the ARIA 1.0 pattern remain functional.

Should Axe continue reporting it as `minor`, allow me to change the aria version with the configuration (couldn't find this option yet) or is it expected just to disable it like shown in the following workaround? 

Workaround:
```
axe.configure({
      rules: [
        {
          id: 'aria-allowed-role',
          enabled: false
        }
      ]
});
```

Thanks for reading.

WilcoFiers:
Thanks for reporting. This does seem to be incorrect. `Combobox` is an allowed role on `input[type=text]`. @jkodu can you go through this again and verify what we missed for this?

georgiee:
From https://www.w3.org/TR/wai-aria-1.1/#combobox
> combobox (role)
> Authors MUST ensure an element with role combobox contains or owns a text input element with role textbox or searchbox and that the text input has aria-multiline set to false.

[...]

Issue #1110, opened 2018-08-28
`aria-allowed-role` issues with axe-core>=3.1.0  (PR #1118)


majapw:
<pre><code>axe-core version: 3.1.0</code></pre>

Hi there! (Sorry for opening so many issues)

Since upgrading our internal version, I've also noticed some issues with other tests and the new `allowed-role` rule. 

Specifically we have an input with `type="text"` and `role="textbox"` failing the `aria-allowed-role` spec:
```
Message: ARIA role must be appropriate for the element
Read more: https://dequeuniversity.com/rules/axe/3.1/aria-allowed-role?application=axeAPI
Code: <input autocomplete="off" id="input_1" name="input_1" placeholder="This is placeholder text!" role="textbox" value="Re" type="text">
```

Reading through https://www.w3.org/TR/html5/dom.html#sec-strong-native-semantics, this seems like it should be allowed by this rule (although perhaps it is failing because the role is redundant---if that is the case, maybe the message should be different?).

jkodu:
@majapw 

Thanks again. No need to apologise for helping us with valid feedback. We appreciate it.

Once again, let us run a few tests on this and get back. I agree that, the message could be more helpful. Once we narrow down the reason, will circle back with information. The code snippet is very helpful. 

Keep it coming.

majapw:
Thanks @jkodu I did some more testing, and it does seem like this issue stems from the redundant role, rather than the role being inappropriate. It would be great if there was some differentiation between these two scenarios as the messaging is pretty ambiguous. 

Basically, it would be great to differentiate between the two cases:
```
<input type="text" role="foobar" /> // role is inappropriate for the element
```
and
```
<input type="text" role="textbox" /> // role is redundant for the element
```

I think the messaging is otherwise very confusing! :) (but also shoutout to `eslint-plugin-jsx-a11y` for helping me find the underlying cause)

WilcoFiers:
Neither of those cases should be failing. `role=foobar` is invalid and should be ignored by aria-allowed-role, and redundant roles should be permitted by default. There's an option in the `aria-allowed-role` check to not allow it, but it should be allowed by default.

jkodu:
PR - https://github.com/dequelabs/axe-core/pull/1118

majapw:
I am also seeing the same error on the following code:
```
Message: ARIA role must be appropriate for the element
--
[...]

PR #1118, opened 2018-08-31
fix: ignore invalid and allow redundant role in aria-allowed-role

jkodu:
This PR tackles fix for ignoring invalid roles and allowing redundant roles in aria-allowed role. It also handles backwards compatibility for ARIA 1 vs 1.1 for certain roles like `combobox`, `spinbutton` and `searchbox`.

Closes issue:
- https://github.com/dequelabs/axe-core/issues/1110
- https://github.com/dequelabs/axe-core/issues/1116

## Reviewer checks


test/checks/aria/aria-allowed-role.js
```diff
@@ -91,6 +91,36 @@ describe('aria-allowed-role', function() {
...
+		var node = document.createElement('input');
+		node.setAttribute('type', 'text');
+		node.setAttribute('role', 'searchbox');
+		fixture.appendChild(node);
+		assert.isTrue(
+			checks['aria-allowed-role'].evaluate.call(checkContext, node)
+		);
+	});
+
 	it('returns false when MENU has type context', function() {
```
isner: Looks like the test description needs to be inverted to match the new expected result.

lib/commons/aria/get-role.js
```diff
@@ -28,6 +29,11 @@ aria.getRole = function getRole(
 		}
 		return aria.isValidRole(role, { allowAbstract: abstracts });
 	});
+
+	if (segments) {
+		return validRoles;
+	}
```
WilcoFiers: It's not at all obvious that `segments` wouldn't return the implicit role. I'm also not a fan of having this turn into a thing that can return either strings or arrays. You're trying to do two things with the same function. I suggest you keep the `getRoleSegments` function around, and have `getRole` call it to get the validRoles value instead.

lib/commons/aria/index.js
```diff
@@ -1687,7 +1695,15 @@ lookupTable.role = {
 		nameFrom: ['author'],
 		context: null,
 		implicit: ['input[type="search"]'],
-		unsupported: false
+		unsupported: false,
+		allowedElements: [
```
WilcoFiers: General question: How did we miss this, and how do we know we're not missing more stuff this time around?

lib/rules/aria-allowed-role-matches.js
```diff
@@ -0,0 +1,18 @@
...
+
+// get role(s) as array(segments), with fallback and dpub
+const roleSegments = getRole(node, {
+	segments: true,
+	fallback: true,
+	dpub: true
+});
+
+// filter invalid roles
+const validRoles = roleSegments.filter(role => isValidRole(role));
```
WilcoFiers: getRole already filters out invalid roles, doesn't it? Couldn't you just do this:

```javascript
return getRole(node, { noImplicit: true, dpub: true }) !== null
```

test/integration/rules/aria-allowed-role/aria-allowed-role.html
```diff
@@ -7,7 +7,6 @@
 <section id="pass-section-role-doc-bib" role="doc-bibliography"></section>
 <ul><li id='pass-li-role-doc-biblioentry' role="doc-biblioentry"></li></ul>
 <aside id='pass-aside-doc-example' role='doc-example'></aside>
-<div id='pass-div-has-any-role' role='divAnyRoleEvenInvalid'></div>
```
WilcoFiers: Instead of removing, I suggest you put in a valid role.

lib/commons/aria/index.js
jkodu: This is the holy grail source of truth - https://www.w3.org/TR/html-aria/ to construct these mappings.

There are 3 caveats with this:
1) Misinterpretation.
2) Conformance between ARIA1.0 and ARIA1.1, that was the case with https://github.com/dequelabs/axe-core/issues/1116. There were multiple ways to use a role. 1.1 mandates usage of role on parent, but backwards compatibility with 1.0 means role can be set on said element itself.
3) When https://www.w3.org/TR/html-aria/ is updated, we need a handle to confirm that things have not been changed, and adapt to new changes. Perhaps can watch/ github hook for the same and streamline the mapping.

To answer your specific Q, we missed this case due to an overlap of 1 and 2 above.

I am going through the table once more to ensure the mappings are right.

test/integration/rules/aria-allowed-role/aria-allowed-role.html
jkodu: Done. Added it back again. This diff only watches this line, so comment is not going stale.

lib/commons/aria/get-role-segments.js
```diff
@@ -0,0 +1,44 @@
...
+ * Returns all roles applicable to element in a list
+ *
+ * @method getRoleSegments
+ * @memberof axe.commons.aria
+ * @instance
+ * @param {Element} node
+ * @returns {Array} Roles list or empty list
+ */
+
+aria.getRoleSegments = function getRoleSegments(node) {
```
WilcoFiers: Either add tests for this, or don't move this into a commons. My preference would be the latter. I think you've made this PR much larger than it needed to be for the fix.

lib/commons/aria/get-element-unallowed-roles.js
```diff
@@ -71,6 +37,12 @@ aria.getElementUnallowedRoles = function getElementUnallowedRoles(
 				return true;
 			}
 		}
+
+		// if role and implicit role are same, it is redundant - ignore
+		if (role === implicitRole) {
```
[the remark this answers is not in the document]
jkodu: You are right, that flow path, only handles the edge case for `table>tr[role='row']`. I shall refactor.

lib/commons/aria/get-role-segments.js
jkodu: There are tests (https://github.com/dequelabs/axe-core/pull/1118/files#diff-c01d2c9028bf4f6a16bace64d8ce59daR1) indeed . They were at the bottom of the PR.

I need to use `getRoleSegments` in 2 places, one within the `matches`, the other within the `check`, hence the abstraction to commons.

But believe I could use `getRole` in `matches` and I will move `getRoleSegments` out of commons.

jkodu:
@WilcoFiers - reviewer checks were not filled up. Doing it on your behalf.

PR #1119, opened 2018-09-04
fix(imports): try/catch with fallback to avoid CSP issues

stephenmathieson:
This patch updates the logic we use to get a reference to the global scope. Previously, if a page (or Chrome extension) had a strict CSP set ("unsafe-eval"), we caused an exception to be thrown by using the `Function` constructor. Now we `try/catch` the statement and fallback to using `window`.

## Reviewer checks

### DIFF
--- a/build/tasks/generate-imports.js
+++ b/build/tasks/generate-imports.js
@@ -91,9 +91,19 @@ module.exports = grunt => {
 					// envs, the UMD method should be used instead.
 					const wrappedLibrary = `
 						(function (module, exports, define, require, process) {
-							// Get a reference to the "true" global scope. This works in
-							// ES5's "strict mode", browsers, node.js and other environments.
-							var global = Function('return this')();
+							// Get a reference to the "true" global scope. This works in ES5's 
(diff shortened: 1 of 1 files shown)


stephenmathieson:
@jkodu while **we** can change **our** `manifest.json`, it's possible other users of axe-core cannot. This patch prevents them from having to change their manifest (which we cannot require users to do without releasing a new major version of axe-core).

stephenmathieson:
Realizing that there's much more to this. The following lines in `doT` are also going to break CSP:

- `eval`: https://github.com/olado/doT/blob/master/doT.js#L40
- `Function()`: https://github.com/olado/doT/blob/master/doT.js#L69
- `Function()`: https://github.com/olado/doT/blob/master/doT.js#L83
- `Function()`: https://github.com/olado/doT/blob/master/doT.js#L133

Let's just hope this doesn't break anything for customers... 😞

marcysutton:
Another use case for CSP is axe-cli, so if you could address it in axe-core that could help elsewhere maybe? https://github.com/dequelabs/axe-cli/issues/19

Issue #767, opened 2018-03-08
Link-name fails with role="presentation" content


WilcoFiers:
The following snippet should not fail the link-name rule:

```html
<a href="link">
  <span role="presentation">Link text</span>
</a>
```

Issue #835, opened 2018-04-06
figure element content should not be ignored for discernible text of links


dylanb:
The following markup works in VO on OS X, I believe it will work everywhere. The accessible name is "Leader" and this shows up as a false positive in the aXe Chrome extension

```
<!doctype html>
<html lang="en-us">
  <head>
    <title>title attribute</title>
  </head>
  <body>
    <a href="javascript:;">
      <figure>
        <img src="./thing.jpg" alt="Leader">
      </figure>
    </a>
  </body>
</html>
```

mfairchild365:
If I had to venture a guess, this is due to the [HTML-AAM standard](https://www.w3.org/TR/html-aam-1.0/). It [states that the alt attribute should be ignored when computing the accessible name for the `<figure>` element](https://www.w3.org/TR/html-aam-1.0/#figure-and-figcaption-elements). Instead, the accessible name should come from the `<figcaption>`.

For example, the same `<figure>` element does not have an accessible name in VoiceOver when it isn't wrapped by a link.

dylanb:
Yep...but the use of the `alt` seems to be accessibility supported, so we need to allow it

mfairchild365:
Fair enough. As AT aligns with the standards, would it be fair to assume that they would stop supporting the `alt` attribute for this case? If so, would be be good to surface a notice that suggests a better, more robust and future proof way of accomplishing this?

PR #1134, opened 2018-09-12
fix: link-name to be deciphered from children

jkodu:
This PR fixes the rule `link-name` to take into consideration, nested children text content for deciphering `accessibleTextVirtual`.

This allows this rule to pass in scenarios like:

```html
<a href="link"><span role="presentation">Link text</span></a>

<a id="pass11" href="link"><div>Some content</div><span role="presentation"></span></a>
```

Closes issue:
- https://github.com/dequelabs/axe-core/issues/767

## Reviewer checks

### DIFF
--- a/lib/commons/text/accessible-text-virtual.js
+++ b/lib/commons/text/accessible-text-virtual.js
@@ -455,6 +455,11 @@ text.accessibleTextVirtual = function accessibleTextVirtual(
 			return element.actualNode.getAttribute('title');
 		}
 
+		//handle children with role presentation
+		if (role === 'presentation') {
+			return getInnerText(element, false, false);
+		}
+
 		return '';
 	};
 
--- a/test/checks/shared/has-visible-text.js
+++ b/test/checks/shared/has-visible-text.js
@@ -34,4 +34,81 @@ describe('has-visible-text', function() {
 			checks['has-visible-text'].evaluate.apply(checkContext, params)
 		);
 	});
+
+	it('should return true if there is children with role presentation and text content', function() {
+		var params = checkSetup(
+			'<a id="target" href="link"><span role="presentation">Link text</span></a>'
+		);
+		var actual = checks['has-visible-text'].evaluate.apply(
+			checkContext,
+			params
+		);
+		assert.isTrue(actual);
+	});
+
+	it('should return false if there is children with role presentation and empty text content', function() {
+		var params = checkSetup(
+			'<a id="target" href="link"><span role="presentation"></span></a>'
+		);
+		var actual = checks['has-visible-text'].evaluate.apply(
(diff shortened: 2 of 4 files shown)


jkodu:
@WilcoFiers 

There seems to be a lot of context on some of these bugs, that is not detailed in the issue tickets. I came up with this fix, after looking stepping through the current code base. 

Can you link me to the issue for `figure`? In the interim I will go through the spec.

WilcoFiers:
This one: https://github.com/dequelabs/axe-core/issues/835. I've been thinking of rewriting this one ever since accname 1.1 hit candidate rec. I don't like the existing implementation. It's outdated, clunky and difficult to understand. Here's a draft of what I've been thinking of doing:

```js
text.computationSteps = Object.freeze([
  getAriaLabelledbyText, // Step 2B.1
  getAriaDescribedbyText, // Step 2B.2
  getAriaLabelText, // Step 2C
  getNativeTextAlternative, // Step 2D
  getControlInLabelText, // Step 2E
  getNameFromContent, // Step 2F
  getTextNodeContent, // Step 2G
  getDescendingContentText, // Step 2H
  getTooltipText // Step 2I
])

text.accessibleTextVirtual = function accessibleTextVirtual (node, context = {}) { 
  // Step 2 A, check visibility
  if (!aria.isHidden(node, context)) {
    return ''
  }

  // Return the value of the first step that returns with text
  return text.computationSteps.reduce((text, step) => {
    if (test !== null) {
      return text
    }
    return text.sanitize(step(node, context))
  }, null)
}

/// ------- These can go into their own files:

function getAriaLabelledbyText () {}

function getAriaDescribedbyText () {}

function getAriaLabelText ({ actualNode }) {
  return actualNode.getAttribute('aria-label')
}

function getNativeTextAlternative () {}

function getControlInLabelText () {}

function getNameFromContent () {}

function getTextNodeContent ({ actualNode }) {
  if (actualNode.nodeType !== 3) {
    return null;
  }
  return actualNode.textContent;
}

function getDescendingContentText () {}

function getTooltipText ({ actualNode }) {
  return actualNode.getAttribute('title')
}
```

WilcoFiers:
The bug seems to be a thing that was incorrect in an earlier version of the algorithm. The steps don't match anymore. Didn't dive into it too deeply, but it seems Step 2H was added to resolve the bug that we're seeing here. That step clearly wasn't part of the spec when we implemented it.

WilcoFiers:
Some more thoughts. Most of these methods in my draft rewrite up there already exist in one for or another in checks. With this approach we deduplicate much of what we're doing in acc name computation.

jkodu:
Decided to rewrite `accessible-name` rule. So closing PR

Issue #529, opened 2017-09-15
New rule: H32 - submit-button


jmucha:
#  H32 -  submit-button

The test checks, if data in a form can be submitted explicitly by using submit buttons (input type="submit", input type="image", or button type="submit").

**Tags:** wcag2a, wcag322, cat.forms, experimental

## Selector

Select each node that has an `form` element with `action` and `method` attributes

## forms/submit-button
- Check if the form contains a button [one of (input type="submit", input type="image", or button type="submit"]; Pass if true.
- Check if only `textarea` element(s) are present with no buttons; Forms don't have input:not(hidden) or select elements; Fail if true.
- Otherwise, cantTell

WilcoFiers:
I think in our discussions we decided on limiting this to `form` elements with `action` and `method` attributes.

@goodwitch can you help review this one? There was recently a discussion in JC about a related issue, right?

goodwitch:
@WilcoFiers the key words to focus on...are "when the form can be submitted implicitly".   See the discussion (from the Jedi Council) documented here https://bitbucket.org/dmusser/deque-way/pull-requests/110/consensus-jedi-decision-on-simple-search/diff#comment-44684588

WilcoFiers:
@goodwitch I'm not clear on what these conclusions mean. Is it correct to conclude that forms without a submit button should not be considered accessibility violations? Are there ever cases where that wouldn't be true?

goodwitch:
@WilcoFiers forms that can be submitted implicitly are not a failure of WCAG 2.0 A/AA.   Here is a description of when a form cannot be implicitly submitted (and when it can be implicitly submitted):

> If the form has no submit button, then the implicit submission mechanism **must do nothing** if the form has more than one field that blocks implicit submission, and must submit the form element from the form element itself otherwise.

jmucha:
I wonder what if scripting is off. The page should work without client-side scripting too.

@WilcoFiers good point. We talked about limiting it.

WilcoFiers:
So @goodwitch raises the concern that in certain cases forms with implicit submit mechanisms do meet the accessibility requirements. I think maybe we should do the following:

1. Raise a violation if the form contains only `textarea` elements, since pressing enter on those won't submit the form
2. Have the result be incomplete in all other cases

We should also exclude forms that don't contain any visible form controls.

jmucha:
I've tweaked it a bit. 
[...]

PR #1138, opened 2018-09-18
[WIP] feat(rule): add new rule submit-button

Danidude:
This will add the new rule: Submit-button

Closes issue: #529

## Reviewer checks


test/integration/rules/submit-button/submit-button.json
```diff
@@ -0,0 +1,7 @@
+{
+	"description": "submit-button test",
+	"rule": "submit-button",
+	"violations": [["#fail1"]],
+	"passes":  [["#pass1"], ["#pass2"], ["#pass3"]],
+	"incomplete": [["#undefiend1"]]
```
jkodu: Would like to see more test cases.

test/integration/rules/submit-button/submit-button.html
```diff
@@ -0,0 +1,24 @@
...
+  <input type="submit"></input>
+</form>
+
+<form id="pass2" action="#" method="post">
+  <input></input>
+  <img type="submit"></input>
+</form>
+
+<form id="pass3" action="#" method="post">
+  <input></input>
```
jkodu: As highlighted by GitHub, there is no need for a closing tag for `input`

test/integration/rules/submit-button/submit-button.json
Danidude: Any test cases that comes to mind? Will add test cases for disabled buttons.

lib/rules/submit-button.json
```diff
@@ -0,0 +1,17 @@
...
+  "id": "submit-button",
+  "selector": "form[action][method]",
+  "tags": [
+    "wcag2a",
+    "wcag322",
+    "cat.forms",
+    "experimental"
+  ],
+  "metadata": {
+    "description": "Check if an form can be submited with an element",
```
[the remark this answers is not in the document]
Danidude: Text updated.

lib/rules/submit-button.json
```diff
@@ -0,0 +1,17 @@
...
+    "help": "Forms with action and method must have an way to be submited."
```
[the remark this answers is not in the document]
Danidude: Removed redundant `.`

lib/checks/forms/submit-button.json
```diff
@@ -0,0 +1,11 @@
+{
+  "id": "submit-button",
+  "evaluate": "submit-button.js",
+  "metadata": {
+    "impact": "serious",
+    "messages": {
+      "pass": "Form has a way to submit content.",
```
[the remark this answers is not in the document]
Danidude: Removed `.`

lib/checks/forms/submit-button.js
```diff
@@ -0,0 +1,9 @@
+if (node.querySelector('input[type="submit"], img[type="submit"], button[type="submit"]')) {
```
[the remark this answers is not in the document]
Danidude: Added checking for disabled submit buttons, now if the button is disabled it will no longer pass.

test/integration/rules/submit-button/submit-button.html
Danidude: Edited the HTML code to remove the closing `input` tag.

lib/checks/forms/submit-button.js
```diff
@@ -0,0 +1,13 @@
+if (
+	node.querySelector(
```
WilcoFiers: I don't think this works across shadow DOM boundaries. If you've got a form element in the light DOM and a submit button in a shadow DOM node inside that form, this won't return it. You can use `axe.utils.querySelectorAll` instead. That will do it.

lib/checks/forms/submit-button.js
```diff
@@ -0,0 +1,13 @@
...
+		'input[type="submit"]:not([disabled]), img[type="submit"]:not([disabled]), button[type="submit"]:not([disabled])'
+	)
+) {
+	return true;
+}
+
+if (!node.querySelectorAll(':not(textarea)').length) {
+	return false;
```
WilcoFiers: I think Dylan has a valid point here:
https://github.com/dequelabs/axe-core/issues/529#issuecomment-424114684

There's always the possibility that someone scripted up a link as the submit button, or any number of other things. You can simply take out this if statement to solve for that.

lib/checks/forms/submit-button.json
```diff
@@ -0,0 +1,11 @@
+{
+  "id": "submit-button",
+  "evaluate": "submit-button.js",
+  "metadata": {
+    "impact": "serious",
+    "messages": {
+      "pass": "Form has a way to submit content",
+      "fail": "Form can not be submited by any indicated element"
```
WilcoFiers: We'll need an "incomplete" message here telling the user to check that this form doesn't get automatically submitted. You can remove the "fail" key as that wouldn't be used anymore.

lib/rules/submit-button.json
```diff
@@ -0,0 +1,17 @@
...
+  "selector": "form[action][method]",
+  "tags": [
+    "wcag2a",
+    "wcag322",
+    "cat.forms",
+    "experimental"
+  ],
+  "metadata": {
+    "description": "Ensure that forms can be submitted by using submit buttons",
+    "help": "Forms with action and method must have an way to be submited"
```
WilcoFiers: I think this is describing the wrong problem. The problem isn't that forms need a submit mechanism, the problem is that users lose their focus if forms automatically submit after user input.

Danidude:
Updated branch with changes, please look them over as see if any change is needed :) @jkodu @WilcoFiers

lib/checks/forms/submit-button.js
```diff
@@ -0,0 +1,24 @@
...
+  virtualNode,
+  'input[type="submit"], input[type="image"], button:not([type="reset"],[type="button"])'
+)
+
+let inputNodes = axe.utils.querySelectorAll(
+  virtualNode,
+  'input'
+)
+
+if (inputNodes.length === 1 && !submitAllButtons.length) {
```
WilcoFiers: I have a number or problems with this implementation:

1. This rule should consider the type of the input element. type=checkbox can't be submitted this way for example.
2. This rule should consider other form fields. If this form has select or textarea in it, this case shouldn't pass.
3. This rule should consider forms without form fields
4. This rule consider ARIA form fields in addition to native ones

test/checks/forms/submit-button.js
```diff
@@ -0,0 +1,117 @@
+describe('submit-button', function() {
```
WilcoFiers: These tests don't properly describe what's going on in the check. There are only a few things happening in this check that need to be tested:

- Form has a submit button that isn't disabled
- Form has no submit button, and no fields
- Form has no submit button, and a single input element
- Form has no submit button, all other scenarios

If you want to test the individual selectors, you can put them in an array and loop each of them into a describe group. Something like this:

```js
['<input type="submit" />', '<button></button>', ...].forEach(function (buttonText) {
	describe(buttonText, function () {
		...
	})
});
```

lib/rules/submit-button.json
```diff
@@ -0,0 +1,17 @@
+{
+  "id": "submit-button",
+  "selector": "form[action][method]",
+  "tags": [
+    "wcag2a",
+    "wcag322",
+    "cat.forms",
+    "experimental"
```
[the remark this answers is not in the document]
WilcoFiers: This rule doesn't turn up failures. It is something that should be tested for WCAG, so I'd say the tags are correct. Perhaps we should introduce a needs-review tag though, for rules that don't return violations?

Danidude:
Just so I understand, this rule was following: https://www.w3.org/TR/WCAG20-TECHS/H32.html
But have changed focus to: https://www.w3.org/TR/2008/REC-WCAG20-20081211/#consistent-behavior-unpredictable-change

jkodu:
Closing this PR for now. 

Issue #537, opened 2017-09-15
New rule: H33 - link-text


jmucha:
#  H33 - link-text

The objective of this test is to check whether an additional text (`title` element) is provided to describe the purpose of the link.

**Tags:** wcag2a, wcag244, cat.text-alternatives, best-practice

## Selector

Select all anchor `a` elements.

### text-alternatives/link-text (any)

1.    For each `a` element check if the link text is unique. If true, pass; else 2.
2.    Check if the links have different titles. If true, pass; else fail.

WilcoFiers:
I don't think this one works all that well. Links aren't required to have a title by WCAG 2.4.4. What I'd suggest doing with this one is to look for links with the same link text, and check if they have different titles. This would be a best-practice.

devarshipant:
If title is present on a link, at a minimum, the test should check and report if the title contains part of the link text. The tester should then manually check the context in which the title is used. Manual override is required to make a determination without calling it a violation.

If the title is exactly the same as the link (innerText), then JAWS 17 / 18 ignore the title and NVDA does not bother speaking the title, so that can be considered a pass.

jmucha:
Let's do it as you suggested @WilcoFiers

Danidude:
Looking over this rule, should the selector be a[title]? As if the link do not contain an title it is no point in checking if the text is different?

WilcoFiers:
It's not just about the title. It's about the link text + title. I would suggest using `a[href]`. We don't want to catch anchor elements without href in this.

Danidude:
Okay, will make this change then, then it also should be ´[role=link]´?

Danidude:
An few more questions are as follows.

 - This test should also compare the href value, to see if they are different. Example here on github the "Pricing" link is the same text as bottom and top, and same URL. But it should not fail an test as this. So link with same text and same href should also pass. 

 - Also what defines 'link text'? Is it only the text between the `<a>` tag, or is aria-label/labelledby also included?

 - This rule looks a lot similar to an other rule we are writing: #532 . Should one of these be dropped? Or re-written?

jkodu:
Hi Dani,

Answers to your questions below:

<blockquote>
Okay, will make this change then, then it also should be [role=link]?
</blockquote>
A: 
An anchor element with a `href` will have a semantic `role` of `link`. But `[role='link']` can perhaps be set on other element types like `button`. Believe this rule is catering only to anchor elements. 
[...]

PR #1143, opened 2018-09-18
feat(rule): add new link-text rule

Danidude:
This will add the new rule link-text that tests if other links have the same link-text, if so if they have different title text.

Closes issue:  #537

## Reviewer checks


lib/checks/text-alternatives/link-text.js
```diff
@@ -0,0 +1,29 @@
+function getText(elem) {
```
jkodu: Could we use `axe.commons.text.accessibleTextVirtual` here? @WilcoFiers, WDYT?

lib/checks/text-alternatives/link-text.js
```diff
@@ -0,0 +1,29 @@
...
+	let returnObj = {};
+
+	returnObj['link-text'] = elem.innerText;
+	returnObj['link-title'] = elem.title;
+	returnObj['link-href'] = elem.href;
+
+	return returnObj;
```
jkodu: Small thing, would read better with just `return { key: value, ... }`

lib/checks/text-alternatives/link-text.js
jkodu: @Danidude - Can you also double check the method, I have suggested above to see if it suffices what are you after?

lib/checks/text-alternatives/link-text.js
jkodu: Also, please `jsdoc` document methods, something like https://github.com/dequelabs/axe-core/blob/develop/lib/commons/aria/label-virtual.js#L4

lib/checks/text-alternatives/link-text.js
```diff
@@ -0,0 +1,29 @@
...
+}
+
+let nodeText = getText(node);
+
+var elems = document.body.querySelectorAll('a[href]');
+
+// Loop trough all the elements to check against
+for (var i = 0; i < elems.length; i++) {
```
jkodu: Using a `for` loop here, leads to redundant iterations.
Use `.some`, to exit iteration on first match.

Reference:
[MDN Array Some](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some)

lib/checks/text-alternatives/link-text.js
```diff
@@ -0,0 +1,29 @@
...
+	let elem = elems[i];
+	if (elem !== node) {
+		let newElemText = getText(elem);
+		if (
+			newElemText['link-text'] === nodeText['link-text'] &&
+			nodeText['link-title'] === newElemText['link-title'] &&
```
jkodu: Small thing, for easier readability, maintain the nodes of comparison left to right the same. Above, nodeText and newElemText are changing sides...

test/checks/text-alternatives/link-text.js
```diff
@@ -0,0 +1,85 @@
...
+		fixture.appendChild(node);
+		var node2 = document.createElement('a');
+		node2.setAttribute('title', 'woohoo');
+		node2.setAttribute('href', '#2');
+		node2.textContent = 'some text';
+		fixture.appendChild(node2);
+
+		assert.isFalse(checks['link-text'].evaluate(node));
+	});
+});
```
jkodu: It would be good to see tests with shadowDOM.

lib/checks/text-alternatives/link-text.js
Danidude: @jkodu with the help of the docs you showed me, I saw I could use an other method: `axe.commons.text.accessibleText`

WilcoFiers:
The Auto-WCAG community group is currently working on a similar group. We should make sure the rule we're implementing here is consistent with that rule. You may want to wait with updating this until that rule is finalised.

https://github.com/auto-wcag/auto-wcag/pull/220

jkodu:
@Danidude 

Asides the CI failing, could you address - https://github.com/dequelabs/axe-core/pull/1143#pullrequestreview-158279379

jkodu:
@Danidude has updated the PR. Reviews welcome.

lib/checks/text-alternatives/link-text.js
```diff
@@ -0,0 +1,12 @@
+let nodeText = axe.commons.text.accessibleText(node);
+let nodeDescripbedBy = axe.commons.aria.ariaDescribedby(node);
```
jkodu: nitpick, spelling mistake - `nodeDescibedBy`

lib/checks/text-alternatives/link-text.js
```diff
@@ -0,0 +1,12 @@
...
+
+let elems = axe.utils.querySelectorAll(axe._tree, 'a[href]');
+return !Array.prototype.slice.call(elems).some(function(element) {
```
jkodu: use `Array.from` instead of `.slice....`

lib/checks/text-alternatives/link-text.js
marcysutton: This still needs a tweak to be spelled correctly: `nodeDescribedBy` (missing the `r`)

lib/checks/text-alternatives/link-text.json
```diff
@@ -0,0 +1,11 @@
+{
+  "id": "link-text",
+  "evaluate": "link-text.js",
+  "metadata": {
+    "impact": "moderate",
+    "messages": {
+      "pass": "Link have an uniqe text and title",
```
marcysutton: This could use some spelling and grammar help:

`Link has unique text and title content`

lib/checks/text-alternatives/link-text.json
```diff
@@ -0,0 +1,11 @@
...
+      "fail": "Links text and title are the same"
```
marcysutton: Link should be singular here: `Link text and title content are the same`

lib/commons/aria/get-describedby.js
```diff
@@ -0,0 +1,47 @@
...
+  * @memberof axe.commons.aria
+  * @instance
+  * @param  {Object} actualNode The virtualNode to test
+  * @return {Mixed}  String of visible text, or `null` if no aria-describedby is found
+  */
+ aria.ariaDescribedbyVirtual = function({ actualNode }) {
+   let ref, candidate;
+
+   if (actualNode && actualNode.getAttribute('aria-describedby')) {
+     // aria-labelledby
```
marcysutton: Is this a stray comment (aria-labelledby isn't referenced)?

lib/checks/text-alternatives/link-text.json
```diff
@@ -0,0 +1,11 @@
+{
+  "id": "link-text",
```
marcysutton: This could use a more descriptive name. It's checking for unique link text, right? I would think something like `unique-link-text` would be more clear. WDYT @WilcoFiers ?

lib/rules/link-text.json
```diff
@@ -0,0 +1,20 @@
+{
+  "id": "link-text",
```
marcysutton: I think this would be more clear as `unique-link-text`. It's pretty vague as written.

lib/rules/link-text.json
```diff
@@ -0,0 +1,20 @@
...
+  "selector": "a[href]",
+  "tags": [
+    "wcag2a",
+    "wcag244",
+    "cat.text-alternatives",
+    "best-practice"
```
marcysutton: This should be tagged with `best-practice` only and not the WCAG stuff. `cat.text-alternatives` is ok.

jkodu:
Closing this PR for now. 

Issue #527, opened 2017-09-15
New rule: F55 - focus


jmucha:
#  F55 - focus

This test checks for javascript *this.blur()* function.

**Tags:** wcag2a, wcag211, cat.keyboard

## Selector

Select each node that has `onfocus` attribute.

### keyboard/focus ( none )

1.    Check if *this.blur* function is a value of the `onfocus` attribute (exact match). If true, fail; else 2;
2.    Check if the event handler contains `this.blur()` function. If true end with `incomplete` status; else pass.

WilcoFiers:
Should this be an exact match, or does simply containing `this.blur` trigger a fail?

jmucha:
I think it should be an exact match. However with this function we have an ambiguous case since the value of the `onfocus` attribute can be another JavaScript function that calls `this.blur()`.

WilcoFiers:
Agreed. Exact match is best. We could make it `incomplete` when `this.blur()` is contained within the handler, but it might not ever be executed so we can't be 100% sure.

jmucha:
That seems to be a good solution. Edited the rule.

Danidude:
Another approach to test elements with `onFocus`, is to give the element focus and wait for some time and see if the element still has focus. If not then `incomplete`, else pass.

WilcoFiers:
The problem with firing focus events is that it can cause the page to change in unpredictable ways. We can't have the page change while we're testing it, that would lead to inconsistent results. Triggering events could even trigger irreversible effects, such as submitting a form to delete something.

Danidude:
Ok, will not go that route then. An note I made about the check keyboard/focus ( none ), have to fail for the rule to pass. However the check says it will fail when it finds this.blur(), witch in return will make the rule pass. I am guessing the check should be all and not none.

AutoSponge:
There's a way to set pseudo state without triggering events (in Chrome). You have to use [`CSS.forcePseudoState`](https://chromedevtools.github.io/devtools-protocol/tot/CSS#method-forcePseudoState). I have proof-of-concept code for this if interested.

WilcoFiers:
Seeing as there has been no activity on this for nearly a year, and I don't think there's any interest from the axe-core team to develop this rule ourselves, I'm going to close this issue.

PR #1145, opened 2018-09-18
feat(rule): new focus rule that checks onFocus for this.blur

Danidude:
Add rule focus, that checks if onFocus contains this.blur, making keyboard navigation harder/impossible.

Closes issue: #527

## Reviewer checks


lib/checks/keyboard/focus.js
```diff
@@ -0,0 +1,7 @@
+if (node.getAttribute('onFocus') === 'this.blur()') {
+	return false;
+} else if (node.getAttribute('onFocus').indexOf('this.blur()') > -1) {
+	return undefined;
+} else {
+	return true;
+}
```
jkodu: I would lose the `if ... else if ... else` and `return` as is.
Each of the code block returns either way at the moment.

lib/checks/keyboard/focus.json
```diff
@@ -0,0 +1,11 @@
+{
+  "id": "focus",
+  "evaluate": "focus.js",
+  "metadata": {
+    "impact": "serious",
+    "messages": {
+      "pass": "Element will not lose focus when resiving focus",
```
jkodu: Editorial `receiving`

lib/checks/keyboard/focus.json
```diff
@@ -0,0 +1,11 @@
...
+      "fail": "Element loses focus when it resives focus"
```
jkodu: Same as above.

lib/rules/focus.json
```diff
@@ -0,0 +1,16 @@
+{
+  "id": "focus",
+  "selector": "[onFocus]",
+  "tags": [
+    "wcag2a",
+    "wcag211",
+    "cat.keyboard"
+  ],
+  "metadata": {
+    "description": "Ensure that executes code onFocus do not loose it when code is executed",
```
jkodu: This sentence is misleading.

Do you mean `Ensure that onFocus code upon execution does not lose focus`, or something similar?

lib/rules/focus.json
```diff
@@ -0,0 +1,16 @@
...
+    "help": "Elements should not loose focus when it just gained it"
+  },
```
jkodu: `lose`

test/integration/full/focus/focus.html
```diff
@@ -0,0 +1,31 @@
...
+		mocha.setup({
+			timeout: 10000,
+			ui: 'bdd'
+		});
+		var assert = chai.assert;
+	</script>
+</head>
+<body>
+  <input placeholder="test 1" onFocus="this.blur()"></input>
+  <input placeholder="test 2" onFocus="blurThis(event)"></input>
```
jkodu: `input` does not need a closing tag.

Danidude:
An example of this use is on this webpage: http://www.nmpd.gov.lv/nmpd/arhivs/?page=10&doc=195 @dylanb

lib/checks/keyboard/focus.js
```diff
@@ -0,0 +1,7 @@
+if (node.getAttribute('onFocus') === 'this.blur()') {
+	return false;
```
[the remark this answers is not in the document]
Danidude: yes, we are at the moment looking at exact match. 

That will also the same thing, and will trigger in the second test, when the onFocus contains `this.blur`. We can add even more special cases other then the current where this test should fail because of `this.blur` if that is wanted.

lib/checks/keyboard/focus.js
Danidude: Unsure how to write this into an single return statement, do you have an example of this?

jkodu:
@WilcoFiers - Do you have an opinion on the same?

lib/rules/focus.json
marcysutton: I'd rewrite this as "Elements should not blur on focus"

lib/checks/keyboard/focus.js
marcysutton: Something like this maybe? The two if blocks could be combined potentially, but since they return different results this probably works fine.

```
if (node.getAttribute('onFocus') === 'this.blur()') {
	return false;
} 
if (node.getAttribute('onFocus').indexOf('this.blur()') > -1) {
	return undefined;
}
return true;
```

marcysutton:
I think this rule has value because of the horrendous accessibility issue caused by it (blurring on focus for some reason, like removing the visible focus style). It wouldn't solve for cases where the focus/blur events were bound in memory or with an undetectable framework on render, but it seems like an easy win to me. @dylanb are there any false positives you're thinking of?

dylanb:
@marcysutton No, I was more concerned about adding rules that add time to the analysis that do not find issues. The selector itself is an attribute selector - this is the slowest selector because it has to look at every element on the page. Could we at least restrict this to naturally tab-focusable elements?

jkodu:
@Danidude - could you do the update as per @dylanb comments - `restrict this to naturally tab-focusable elements`

Danidude:
@jkodu Sure, I can look into that and see what can be done. But then we are ignoring elements that are not inert focusable with `type` and `tabIndex` attributes?

jkodu:
@Danidude 

The way forward with this is as below:

1) Narrow down the rule to well-known focusable elements as suggested above.
2) Tag this rule as `experimental`.

Could you do the above changes please.

jkodu:
@Danidude 

The below comments needs to be addressed:

- https://github.com/dequelabs/axe-core/pull/1145/files/31d62122b433edf1b2d46772a9d2e0adc3d99cf9#r227067048

- https://github.com/dequelabs/axe-core/pull/1145/files/31d62122b433edf1b2d46772a9d2e0adc3d99cf9#r219429627

- https://github.com/dequelabs/axe-core/pull/1145/files/31d62122b433edf1b2d46772a9d2e0adc3d99cf9#r227067048

- `Narrow down the rule to well-known focusable elements as suggested above.`

lib/rules/focus.json
```diff
@@ -0,0 +1,17 @@
+{
+  "id": "focus",
+  "selector": "a[href][onFocus], link[href][onFocus], button[onFocus], input[onFocus], select[onFocus], textarea[onFocus], audio[controls][onFocus], video[controls][onFocus]",
```
WilcoFiers: Minor point, but `link` isn't a clickable element. `area` and `summary` are though. Should switch that up.

lib/checks/keyboard/focus.js
```diff
@@ -0,0 +1,7 @@
+if (node.getAttribute('onFocus') === 'this.blur()') {
```
WilcoFiers: This should account for whitespace. Simply putting `.trim()` on the attribute value should do.

lib/rules/focus.json
```diff
@@ -0,0 +1,17 @@
+{
+  "id": "focus",
```
WilcoFiers: Not too keen on this rule / check ID. May I suggest "onfocus-no-blur"?

test/integration/full/focus/focus.js
```diff
@@ -0,0 +1,33 @@
+
+describe('focus test failure', function () {
```
WilcoFiers: I don't think this rule needs a "full" test. Those are just for rules that require things like iframes or that apply to the HTML element. That's not the case for this rule. Full tests are pretty slow so it's better not to use them when we don't need them.

lib/checks/keyboard/focus.js
```diff
@@ -0,0 +1,7 @@
...
+  return false;
+}
+if (node.getAttribute('onFocus').indexOf('this.blur()') > -1) {
```
WilcoFiers: Sorry, this is a little nit-picky, but HTML attributes are all lower case. We should stick to that convention. In this particular case it doesn't matter, but there are scenarios where upper case in attributes can cause problems, so as a general practice we should just make sure our use of things match what is defined in the spec.

lib/rules/focus.json
Danidude: For some reason, do the selector have an problem with `area` element. It works just fine in Chrome, however Firefox and circleci tests cant find the element, and reports an missing element in the tests. Have yet to figure out why these two browsers differs on this element.

lib/checks/keyboard/focus.json
```diff
@@ -0,0 +1,11 @@
+{
+  "id": "focus",
+  "evaluate": "focus.js",
+  "metadata": {
+    "impact": "serious",
+    "messages": {
+      "pass": "Element will not lose focus when receiving focus",
+      "fail": "Element loses focus when it is receiving focus"
```
marcysutton: These messages are worded confusingly. How about something like this:
```
"pass": "Element does not lose focus after receiving focus",
"fail": "Element is blurred after receiving focus"
```

jkodu:
Closing this PR for now. 

PR #1154, opened 2018-09-20
feat: Add allowEmpty option for aria-valid-attr

### LINKED ISSUE #994: Allow empty string values for 'aria-current' attribute

WilcoFiers:
Allow specifying per ARIA attribute which ones are allowed to be empty. Changes include:
- aria-haspopup: Now allowed to be empty
- aria-invalid: Now allowed to be empty
- aria-current: Now allowed to be empty
- aria-valuetext: No longer allowed to be empty

Closes #994

## Reviewer checks


lib/commons/aria/index.js
```diff
@@ -78,16 +85,20 @@ lookupTable.attributes = {
...
 	},
 	'aria-keyshortcuts': {
-		type: 'string'
+		type: 'string',
+		allowEmpty: true
 	},
 	'aria-label': {
-		type: 'string'
+		type: 'string',
+		allowEmpty: true
```
dylanb: Do we have a test to make sure that an empty aria-label will not pass an input with `aria-label=""`?

lib/commons/aria/index.js
```diff
@@ -55,21 +59,24 @@ lookupTable.attributes = {
 		values: ['copy', 'move', 'reference', 'execute', 'popup', 'none']
 	},
 	'aria-errormessage': {
-		type: 'idref'
+		type: 'idref',
+		allowEmpty: true
```
dylanb: Is this new? If so, where are the tests for it?

WilcoFiers: In order to make this configurable per attribute I added this. This property is used in `aria.validateAttrValue`, which is where the tests for it are. All changes (see PR description) are also tested in integration tests.

lib/commons/aria/index.js
WilcoFiers: `aria-label` has always been allowed empty. See: https://github.com/dequelabs/axe-core/pull/1154/files#diff-0dbb83fa6e11da52031f948e3f85a3ecL105

Rules that require elements to have an accessible name don't allow empty or whitespace values to pass.

lib/commons/aria/index.js
```diff
@@ -78,16 +85,20 @@ lookupTable.attributes = {
...
 	},
 	'aria-label': {
-		type: 'string'
+		type: 'string',
+		allowEmpty: true
 	},
 	'aria-labelledby': {
-		type: 'idrefs'
+		type: 'idrefs',
+		allowEmpty: true
```
dylanb: There is no current test to test the empty case for `aria-labelledby` - we should add one

lib/commons/aria/index.js
```diff
@@ -113,10 +124,12 @@ lookupTable.attributes = {
 		values: ['horizontal', 'vertical']
 	},
 	'aria-owns': {
-		type: 'idrefs'
+		type: 'idrefs',
+		allowEmpty: true
```
dylanb: There is no current test for `aria-owns` to test that it is allowed to be empty

lib/commons/aria/index.js
```diff
@@ -113,10 +124,12 @@ lookupTable.attributes = {
...
 	},
 	'aria-placeholder': {
-		type: 'string'
+		type: 'string',
+		allowEmpty: true
```
dylanb: There are no tests for `aria-placeholder`

lib/commons/aria/index.js
```diff
@@ -78,16 +85,20 @@ lookupTable.attributes = {
 	},
 	'aria-invalid': {
 		type: 'nmtoken',
+		allowEmpty: true,
 		values: ['true', 'false', 'spelling', 'grammar']
 	},
 	'aria-keyshortcuts': {
-		type: 'string'
+		type: 'string',
+		allowEmpty: true
```
dylanb: No tests for `aria-keyshortcuts`

lib/commons/aria/index.js
```diff
@@ -55,21 +59,24 @@ lookupTable.attributes = {
...
 	},
 	'aria-expanded': {
 		type: 'nmtoken',
 		values: ['true', 'false', 'undefined']
 	},
 	'aria-flowto': {
-		type: 'idrefs'
+		type: 'idrefs',
+		allowEmpty: true
```
dylanb: No current test for `aria-flowto` to test that it passes when empty

lib/commons/aria/index.js
dylanb: There is no current test to test that this passes when empty

lib/commons/aria/index.js
```diff
@@ -37,14 +38,17 @@ lookupTable.attributes = {
...
 	},
 	'aria-current': {
 		type: 'nmtoken',
+		allowEmpty: true,
 		values: ['page', 'step', 'location', 'date', 'time', 'true', 'false']
 	},
 	'aria-describedby': {
-		type: 'idrefs'
+		type: 'idrefs',
+		allowEmpty: true
```
dylanb: No current test fo `aria-describedby` to show that it passes when empty

lib/commons/aria/index.js
```diff
@@ -37,14 +38,17 @@ lookupTable.attributes = {
 		type: 'int'
 	},
 	'aria-controls': {
-		type: 'idrefs'
+		type: 'idrefs',
+		allowEmpty: true
```
dylanb: No current test to show that this passes when empty

WilcoFiers:
@dylanb you were totally right. Actually found a bug with aria-errormessage. This should solve it.

lib/commons/aria/attributes.js
```diff
@@ -45,70 +45,3 @@ aria.validateAttr = function(att) {
 	'use strict';
 	return !!aria.lookupTable.attributes[att];
 };
-
-/**
```
marcysutton: I wish this file was moved in a separate commit so I could see if there were any changes to it side by side.... Lots of style changes mixed in here.

marcysutton:
@WilcoFiers are you going to add any more unit tests to this like Dylan mentioned, or is it covered enough by the integration tests?

test/checks/aria/errormessage.js
```diff
@@ -5,8 +5,15 @@ describe('aria-errormessage', function() {
 	var shadowSupported = axe.testUtils.shadowSupport.v1;
 	var shadowCheckSetup = axe.testUtils.shadowCheckSetup;
 	var checkContext = axe.testUtils.MockCheckContext();
+	var attrData = Object.assign(
+		{},
+		axe.commons.aria.lookupTable.attributes['aria-errormessage']
+	);
 
 	afterEach(function() {
+		axe.commons.aria.lookupTable.attributes[
```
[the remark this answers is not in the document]
WilcoFiers: Not that I know of.

lib/commons/aria/attributes.js
WilcoFiers: I'm not sure how to do this.

marcysutton: I thought maybe you could make the changes in the existing file in one commit and move/rename it in a separate commit. I don't think it's worth holding up this PR, just something to keep in mind moving forward.

lib/checks/aria/errormessage.js
```diff
@@ -1,12 +1,16 @@
...
-	hasAttr = node.hasAttribute('aria-errormessage');
+const attr = node.getAttribute('aria-errormessage');
+const hasAttr = node.hasAttribute('aria-errormessage');
 
-var doc = axe.commons.dom.getRootNode(node);
+const doc = dom.getRootNode(node);
 
-function validateAttrValue() {
-	var idref = attr && doc.getElementById(attr);
+function validateAttrValue(attr) {
```
marcysutton: Why is there a method here called `validateAttrValue` when there is also one in `commons.aria`? Should they be the same method, or at least have different names?

WilcoFiers: I don't know. It was something you approved. Is it relevant to this PR?

marcysutton: That PR doesn't reference the commons method, only the inline function.

Issue #1149, opened 2018-09-18
Rule: SC2-5-3-label-content-name-mismatch (PR #1159)


jkodu:
Spec: https://auto-wcag.github.io/auto-wcag/rules/SC2-5-3-label-content-name-mismatch.html

dylanb:
@jkodu the link referenced in this issue is broken, can you find the new link and update this

PR #1159, opened 2018-09-26
[WIP] feat(rule): label content name mismatch

jkodu:
This PR implements a new WCAG2.1 rule - label content name mismatch

Closes issue:
- https://github.com/dequelabs/axe-core/issues/1149

## Reviewer checks


lib/checks/label/label-content-name-mismatch.js
```diff
@@ -0,0 +1,34 @@
+function getAriaText(elm) {
+	if (elm.getAttribute('aria-label')) {
+		return elm
+			.getAttribute('aria-label')
+			.toLowerCase()
+			.trim();
+	}
+	if (elm.getAttribute('aria-labelledby')) {
```
WilcoFiers: aria-labelledby is an element reference. You're just picking up the ID of the element, not the actual text. We have some code for this, but its not isolated properly. Let me complete my accessible name update before completing this. One of the things I'm doing there is to create a getAriaLabelledbyName function.

lib/checks/label/label-content-name-mismatch.js
```diff
@@ -0,0 +1,34 @@
...
+	return null;
+}
+
+const accessibleText = getAriaText(node);
+// if no accessible text - fail
+if (!accessibleText || !accessibleText.length) {
+	return false;
+}
+
+const contentText = node.textContent.toLowerCase().trim();
```
WilcoFiers: You should use `text.sanitize`. It takes things like double spaces into account as well. Related question, does textContent solve for HTML entities like `&nbsp;`? We should probably have a test that makes sure HTML Entities aren't seen as different characters.

lib/checks/label/label-content-name-mismatch.js
```diff
@@ -0,0 +1,34 @@
...
+// if no text content - fail
```
WilcoFiers: Why wouldn't we ignore these types of elements?

lib/rules/label-content-name-mismatch.json
```diff
@@ -0,0 +1,17 @@
+{
+  "id": "label-content-name-mismatch",
+  "matches": "label-content-name-mismatch-matches.js",
+  "tags": [
+    "wcag21a",
+    "wcag253"
```
WilcoFiers: We should make this rule experimental for the time being. I don't trust that this isn't going to give false positives.

lib/checks/label/label-content-name-mismatch.json
```diff
@@ -0,0 +1,11 @@
+{
+  "id": "label-content-name-mismatch",
+  "evaluate": "label-content-name-mismatch.js",
+  "metadata": {
+    "impact": "serious",
+    "messages": {
+      "pass": "Interactive element contains visible label as part of it's accessible name",
+      "fail": "Interactive element does not contain visible label as part of it's accessible name"
```
WilcoFiers: I suggest: The visible text of the element is different from its accessible name

lib/checks/label/label-content-name-mismatch.js
```diff
@@ -0,0 +1,34 @@
...
+if (!contentText || !contentText.length) {
+	return false;
+}
+
+// if text content is not part of accessible text - fail
+if (!accessibleText.includes(contentText)) {
```
WilcoFiers: I don't know that we have a string polyfill of `.includes`. Make sure this works on IE.

lib/rules/label-content-name-mismatch-matches.js
```diff
@@ -0,0 +1,33 @@
+/**
+ * Applicability:
+ * This rule applies to any element that has:
+ * - a semantic role that is a widget that supports name from content, and
+ * - visible text content, and
+ * - an aria-label or aria-labelledby attribute.
+ */
+const { aria } = axe.commons;
+const role = aria.getRole(node, { noImplicit: false, fallback: true });
```
WilcoFiers: Because of our accessibility support baseline, axe-core shouldn't support fallback roles for the time being. Also `noImplicit: false` is the default, so you don't need to set this options object at all.

lib/rules/label-content-name-mismatch-matches.js
```diff
@@ -0,0 +1,33 @@
...
+
+const rolesWithNameFromContents = aria.getRolesWithNameFromContents();
+
+// if role is not one of roles with name from contents - exclude
+if (!rolesWithNameFromContents.includes(role)) {
+	return false;
+}
+
+// if no accessible name attributes - exclude
+if (!node.getAttribute('aria-labelledby') && !node.getAttribute('aria-label')) {
```
WilcoFiers: You should probably use `.trim()` on these to make sure they're actually empty.

lib/rules/label-content-name-mismatch-matches.js
```diff
@@ -0,0 +1,33 @@
...
+	return false;
+}
+
+// if no text content - exclude
+if (node.textContent.toLowerCase().trim().length <= 0) {
```
WilcoFiers: I don't think `.textContent is reliable enough for this. We don't want to check against hidden text. I think we have a visibleText method (not sure about the name) that gets you only text that's displayed.

lib/rules/label-content-name-mismatch.json
```diff
@@ -0,0 +1,17 @@
...
+  ],
+  "metadata": {
+    "description": "Ensures that interactive elements labelled through their content must have their visible label as part of their accessible name",
+    "help": "Interactive elements must have their visible label as part of their accessible name"
```
WilcoFiers: Name from content isn't just for interactive elements. I'm struggling to come up with a short but still descriptive text for this rule. Maybe talk to Jean or Carrie for suggestions?

test/checks/label/label-content-name-mismatch.js
```diff
@@ -0,0 +1,127 @@
...
+	});
+
+	it('returns false when element has an empty aria-label', function() {
+		fixture.innerHTML =
+			'<button id="target" name="link" aria-label="">The full label</button>';
+		var node = fixture.querySelector('#target');
+		var tree = (axe._tree = axe.utils.getFlattenedTree(fixture));
+		var vNode = axe.utils.getNodeFromTree(tree[0], node);
+		var actual = check.evaluate(node, options, vNode);
+		assert.isFalse(actual);
```
WilcoFiers: If aria-label is empty / whitespace only it ignored as the accessible name, and the content will be used as label instead. This case should pass.

test/checks/label/label-content-name-mismatch.js
```diff
@@ -0,0 +1,127 @@
...
+	});
+
+	it('returns false when element has an empty aria-labelledby', function() {
```
WilcoFiers: Same here, empty aria-labelledby will be ignored by the name computation. This shouldn't fail.

test/checks/label/label-content-name-mismatch.js
```diff
@@ -0,0 +1,127 @@
...
+			node,
+			options,
+			axe.utils.getNodeFromTree(tree[0], node)
+		);
+		assert.isFalse(actual);
+	});
+
+	it('returns false when element has no text content', function() {
+		fixture.innerHTML =
+			'<button id="target" name="link" aria-label="Ok"></button>';
```
WilcoFiers: This too, if there is no content, there is no mismatch and so this shouldn't fail.

test/checks/label/label-content-name-mismatch.js
```diff
@@ -0,0 +1,127 @@
...
+	it('returns false when not all of visible label is included in accessible name', function() {
+		fixture.innerHTML =
+			'<button id="target" name="link" aria-label="the full">The full label</button>';
+		var node = fixture.querySelector('#target');
+		var tree = (axe._tree = axe.utils.getFlattenedTree(fixture));
+		var vNode = axe.utils.getNodeFromTree(tree[0], node);
+		var actual = check.evaluate(node, options, vNode);
+		assert.isFalse(actual);
+	});
+});
```
WilcoFiers: You're missing quite a few tests here:
- aria-labelledby with reference to an element that matches / that doesn't match
- aria-labelledby with references to multiple elements
- aria-labelledby with references to non-existing elements
- Elements with hidden texts like: `<button aria-label="help">? <i class="sr-only">help</i></button>
- Elements with ASCII as non-text content `<button aria-label="close">X</button>`
- ...

test/rule-matches/label-content-name-mismatch-matches.js
```diff
@@ -0,0 +1,87 @@
+describe('label-content-name-mismatch-matches', function() {
```
WilcoFiers: I don't think the following test case is addressed:
https://auto-wcag.github.io/auto-wcag/_testcases-embeds/SC2-5-3-label-content-name-mismatch_inapplicable_example_5.html

test/rule-matches/label-content-name-mismatch-matches.js
```diff
@@ -0,0 +1,87 @@
...
+	});
+
+	it('returns false non text content', function() {
+		fixture.innerHTML =
+			'<button aria-label="close"><i class="fa fa-icon-close"></i></button>';
+		var target = fixture.querySelector('button');
+		var actual = rule.matches(target);
+		assert.isFalse(actual);
+	});
+});
```
WilcoFiers: Need tests for aria-labelledby.

jkodu:
Dependency on - https://github.com/dequelabs/axe-core/pull/1163
Closing until the above is merged.

lib/checks/label/label-content-name-mismatch.js
jkodu: We seem to have a polyfill for `String.prototype.includes` - https://github.com/dequelabs/axe-core/blob/develop/lib/core/utils/pollyfills.js#L288

lib/rules/label-content-name-mismatch-matches.js
jkodu: Do you mean `axe.commons.text.accessibleText`?

lib/rules/label-content-name-mismatch.json
jkodu: @jeankaplansky - can you help with a better help text here?

test/checks/label/label-content-name-mismatch.js
jkodu: `getAriaLabelledbyText` authored for accessible name computation, does not take that into account, and does not return the content as text :/.

test/checks/label/label-content-name-mismatch.js
jkodu: Same comment as above, repeating:

`getAriaLabelledbyText` authored for accessible name computation, does not take that into account, and does not return the content as text :/. 

Perhaps the accessible name computation should take this into consideration. Correct me if I am wrong.

lib/rules/label-content-name-mismatch.json
jeankaplansky: Yep. I'm on it.

jeankaplansky:
[Help/Description for Jey](https://trello.com/c/T5gw0bFn/52-help-description-for-jey)

jkodu:
Dependency on - https://github.com/dequelabs/axe-core/pull/1163
Closing until the above is resolved.

lib/rules/label-content-name-mismatch.json
jeankaplansky: @WilcoFiers , @jkodu : Is there a specific reason we cannot say "Ensures that ELEMENTS labeled through their content must have their visible label as part of their accessible name"? Ditto with the value for help: "ELEMENTS must have their visible label as part of their accessible name."

This approach accounts for all elements regardless of functional semantic intent (interactive or plain content).

PR #1163, opened 2018-09-26
fix: Solve several accessible-name issues

### LINKED ISSUE #767: Link-name fails with role="presentation" content

### LINKED ISSUE #835: figure element content should not be ignored for discernible text of links

### LINKED ISSUE #1162: Use aria-valuenow in accessible name
The accessible name computation does not take aria-valuenow and aria-valuetext into account when computing the accessible text. I noticed this while rewriting the accessible text computation.

### LINKED ISSUE #1337: False positive: label rule fails for input having explicit label with role presentation
- Link to live demo: https://codepen.io/iamrafan/pen/GzqOZp

-  AT test results:

  | AT/Browser | Announced as
-- | -- | --
1 | Edge/Narrator | Name-case-1   editing
2 | Edge/JAWS | Name-case-1 Edit
3 | Edge/NVDA | Name-case-1  edit
4 | Chrome/NVDA | Name-case-1  edit
5 | Chrome/JAWS | Name-case-1  edit
6 | IE/JAWS | Name-case-1  edit
7 | IE/NVDA | Name-case-1  edit
8 | Firefox/NVDA | edit  has auto complete

<pre><code>axe-core version: 3.1.2
axe-Coconut version: 3.7.0</code></pre>

WilcoFiers:
Accessible Name computation.

- Closes #1337
- Closes #1162 
- Closes #835 
- Closes #767

## Reviewer checks


WilcoFiers:
I came across this earlier today. The acceptance tests for AccName 1.1. I'm going to see if we can somehow use this.
https://www.w3.org/wiki/AccName_1.1_Testable_Statements#Name_test_case_661

lib/commons/text/form-control-value.js
```diff
@@ -0,0 +1,101 @@
...
+	) {
+		return actualNode.value;
+	}
+
+	if (nodeName === 'SELECT') {
+		return nativeSelectValue(virtualNode);
+	}
+
+	if (selectRoles.includes(role)) {
+		let listbox;
```
WilcoFiers: TODO: Should pull this code into `getAriaSelectValue`

lib/commons/text/form-control-value.js
```diff
@@ -0,0 +1,101 @@
...
+			if (opts[i].selected) {
+				returnText += ' ' + opts[i].text;
+			}
+		}
+		return text.sanitize(returnText);
+	}
+	return '';
+}
+
+function getAriaSelectValue(virtualNode, context) {
```
WilcoFiers: Should be called `ariaSelectValue`.

lib/commons/text/form-control-value.js
```diff
@@ -0,0 +1,101 @@
+/* global text, aria */
+// TODO: Should blacklist so things don't go stale as easilly
```
WilcoFiers: Shouldn't forget this.

lib/commons/text/name-from-content.js
```diff
@@ -0,0 +1,56 @@
...
+		? roleDef && roleDef.nameFrom.includes('contents')
+		: !roleDef || roleDef.nameFrom.includes('contents');
+
+	if (
+		isNamedFromContent ||
+		context.inLabelledByContext ||
+		actualNode.nodeName.toUpperCase() === 'LABEL'
+	) {
+		// 2F.i
+		return text.nativeTextMethods.subtreeText(virtualNode, context);
```
WilcoFiers: Maybe subtreeText should be in its own file. It's important enough.

lib/commons/text/accessible-text-virtual.js
```diff
@@ -216,247 +10,73 @@ function nonEmptyText(t) {
...
-	 * @param {VirtualNode} element The VirtualNode instance of the HTMLElement
-	 * @param {Boolean} inLabelledByContext True when in the context of resolving a labelledBy
-	 * @param {Boolean} inControlContext True when in the context of textifying a widget
-	 * @return {string}
-	 */
-	accessibleNameComputation = function(
-		element,
-		inLabelledByContext,
-		inControlContext
+function getTooltipText({ actualNode }) {
```
WilcoFiers: This should go into its own file and be used in the title check.

lib/commons/text/accessible-text-virtual.js
```diff
@@ -216,247 +10,73 @@ function nonEmptyText(t) {
...
-
-	function getLayoutTableText(element) {
-		// // check if layout table only has one cell
-		if (
-			!axe.commons.table.isDataTable(element.actualNode) &&
-			axe.commons.table.getAllCells(element.actualNode).length === 1
-		) {
-			return getInnerText(element, false, false).trim();
-		}
+	const computationSteps = [
```
WilcoFiers: Should put this in its own function for the sake of testability

lib/commons/text/name-from-content.js
```diff
@@ -0,0 +1,57 @@
...
+	/**
+	 * Note: Unlike what ACCNAME-1.1 says, axe-core allows "name from content"
+	 * for elements without a role, as there MIGHT BE cases where accName
+	 * is called on elements without a role. This gives axe a way to work out
+	 * if these elements are still available to assistive technologies.
+	 *
+	 * For a correct implementation use context.strict = true
+	 */
+	let isNamedFromContent;
+	if (context.strict) {
```
WilcoFiers: TODO: Figure out if other than just our test cases, we actually need this.

lib/commons/aria/get-owned.js
```diff
@@ -0,0 +1,10 @@
+/* global aria, dom */
+aria.getOwnedVirtual = function getOwned({ actualNode, children }) {
+	return dom.idrefs(actualNode, 'aria-owns').reduce((ownedElm, element) => {
+		if (element) {
+			let virtualNode = axe.utils.getNodeFromTree(axe._tree[0], element);
```
WilcoFiers: const

marcysutton:
@WilcoFiers can you describe the fixes this intends to solve, and the status of this PR?

lib/checks/forms/group-labelledby.js
```diff
@@ -1,34 +1,45 @@
...
-		.filter(function(n) {
-			var labelNode = doc.getElementById(n);
-			return labelNode && axe.commons.text.accessibleText(labelNode, true);
-		}).length !== 0
-);
+const sharedIds = matchingNodes
+	.map(m => {
+		return utils.tokenList(m.getAttribute('aria-labelledby') || '');
+	})
+	.reduce((prev, curr) => {
```
jkodu: Although `reduce` can be called with out an `initialValue | Optional`, in which case, the first value in the `array` is used as `initialValue`, this will cause an error/ break when the above `map` returns an empty array, as the below line `prev.filter` will fail as `prev` will be `undefined`.

lib/commons/aria/get-aria-labelledby-text.js
```diff
@@ -0,0 +1,45 @@
...
+	 */
+	if (
+		node.nodeType !== 1 ||
+		context.inLabelledByContext ||
+		context.inControlContext
+	) {
+		return '';
+	}
+
+	const refs = dom.idrefs(node, 'aria-labelledby').filter(elm => elm);
```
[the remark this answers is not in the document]
WilcoFiers: dom.idrefs does return undefined when an element can't be found. We should probably improve the API but I think that's out of scope for this PR.

lib/commons/aria/get-aria-labelledby-text.js
```diff
@@ -0,0 +1,45 @@
...
+	return refs.reduce((accessibleName, elm) => {
+		const accessibleNameAdd = text.accessibleText(elm, {
+			// Prevent the infinite reference loop:
+			inLabelledByContext: true,
+			// Allow hidden content if the elm is hidden
+			includeHidden: !dom.isVisible(elm, true),
```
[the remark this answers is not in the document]
WilcoFiers: I need to look if this is part of the accessibility tree, axe.utils.isHidden doesn't do that.

lib/commons/aria/get-aria-label-text.js
```diff
@@ -0,0 +1,8 @@
+/* global aria */
+aria.getAriaLabelText = function getAriaLabelText(node) {
+	node = node.actualNode || node;
+	if (node.nodeType !== 1) {
+		return '';
+	}
+	return node.getAttribute('aria-label') || '';
```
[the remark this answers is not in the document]
WilcoFiers: No, we can only sanitize after everything is concatenated.

lib/commons/aria/get-owned-virtual.js
```diff
@@ -0,0 +1,13 @@
+/* global aria, dom */
+aria.getOwnedVirtual = function getOwned({ actualNode, children }) {
+	if (!actualNode || !children) {
+		throw new Error('getOwnedVirtual requires a virtual node');
+	}
+	return dom.idrefs(actualNode, 'aria-owns').reduce((ownedElm, element) => {
```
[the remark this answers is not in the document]
WilcoFiers: Renamed. This should have been called `ownedElms`, since this is an array. I do prefer giving the name of the thing I'm building up in the reduce function.

lib/commons/aria/arialabelledby-text.js
```diff
@@ -0,0 +1,52 @@
...
+	node = node.actualNode || node;
+	/**
+	 * Note: The there are significant difference in how many "leads" browsers follow.
+	 * - Firefox stops after the first IDREF, so it
+	 * 		doesn't follow aria-labelledby after a for:>ID ref.
+	 * - Chrome seems to just keep iterating no matter how many levels deep.
+	 * - AccName-AAM 1.1 suggests going one level deep, but to treat
+	 * 		each ref type separately.
+	 *
+	 * Axe-core's implementation behaves most closely like Firefox as it seems
```
dylanb: nit: describe the way this works as "implementing the common denominator"

test/commons/text/accessible-text.js
```diff
@@ -61,6 +62,8 @@ describe('text.accessibleTextVirtual', function() {
 		var rule2a = axe.utils.querySelectorAll(axe._tree, '#beep')[0];
 		var rule2b = axe.utils.querySelectorAll(axe._tree, '#flash')[0];
 		assert.equal(axe.commons.text.accessibleTextVirtual(rule2a), 'Beep');
+		// Chrome: "Flash the screen Number of times to flash screen times"
+		// Firefox: "Flash the screen Number of times to flash screen 3 times"
```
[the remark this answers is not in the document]
WilcoFiers: Tested, fixing the comments separately.

Issue #1176, opened 2018-10-04
Browser extension: axe is ignoring hidden labels for input fields even when it should acknowledge them


chirag64:
There are common instances where we often require labels of input fields to be hidden (e.g. Search box because it already has a search icon in it).

As per [this W3C article](https://www.w3.org/WAI/tutorials/forms/labels/#note-on-hiding-elements), the recommended way to achieve this is to add the below CSS code to the label:
<pre>.visuallyhidden {
  border: 0;
  clip: rect(0 0 0 0);
  height: 1px;
  margin: -1px;
  overflow: hidden;
  padding: 0;
  position: absolute;
  width: 1px;
}</pre>

But even after implementing this, aXe still throws an error for the field saying `Form elements must have labels`.

Here is a working demo of the same: https://jsfiddle.net/xk2vaq6f/1/embedded/result

Volker-E:
Same issue at https://wikimediafoundation.org/ on element `input#wmf-subscribe-input-email`. First idea was that `clip: rect( 1px 1px 1px 1px )` would be needed to not count as 'hidden'. But changing didn't resolve error output.

chirag64:
> Same issue at https://wikimediafoundation.org/ on element `input#wmf-subscribe-input-email`. First idea was that `clip: rect( 1px 1px 1px 1px )` would be needed to not count as 'hidden'. But changing didn't resolve error output.

Right, clipping does seem to be the problem. Even if you use large values like `100px` with it, it does not resolve the issue, which means it does not bother checking for its values.

WilcoFiers:
Interestingly, that's the exact same bug I reported at almost the same time as you. #1777. I'll close that one as a duplicate.

chirag64:
Have submitted a PR for this. Please let me know if I've missed something as per project guidelines, I'll happily fix it

chirag64:
Any idea when this fix will be available in the stable versions of the Chrome & Firefox extensions? We've a defect open on our system that has a dependency on this. 😓

calder12:
This is still an issue, Using Lighthouse v4 shows elements using the CSS in the initial report as "Form element has explicit <label> that is hidden" Using Lighthouse v 3.2.0 (which uses axe-core 3.0.0 beta) correctly identifies the labels.

Any idea on a resolution time for this?

nolanlawson:
This still appears to be happening in axe Chrome extension v3.7.0. The `<label>` I'm using has the following styles:

```css
.sr-only {
[...]

PR #1182, opened 2018-10-09
fix(is-visible): Clipped labels are only marked as hidden if clip is 0px in all directions

chirag64:
fix(is-visible): Clipped labels are only marked as hidden if clip is 0px in all directions

    Instead of marking all clipped labels as hidden, we're now checking their values as well

    Closes issue: #1176

## Reviewer checks

### DIFF
--- a/lib/commons/dom/is-visible.js
+++ b/lib/commons/dom/is-visible.js
@@ -13,7 +13,7 @@ function isClipped(clip) {
 	'use strict';
 
 	var matches = clip.match(
-		/rect\s*\(([0-9]+)px,?\s*([0-9]+)px,?\s*([0-9]+)px,?\s*([0-9]+)px\s*\)/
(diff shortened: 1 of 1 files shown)


chirag64:
So I realize this is failing a few tests, but now I'm confused what should be the expected behavior? According to the W3C article in [the comment I mentioned earlier](https://github.com/dequelabs/axe-core/issues/1176#issue-366892942), `clip: rect(0 0 0 0);` is a valid way to hide labels and should not throw errors.

So I'm guessing instead of allowing clip values of rect greater than 0, we need to consider removing all checks of clip since it is a valid way to hide labels. Any thoughts around this?

WilcoFiers:
This solution isn't going to work. What you're trying to do is check if the height / width is 0. What the code you changed was doing is check that the bottom isn't greater than the top, and that right isn't greater than left. The problem you're trying to solve comes from a different bug, which is that this check shouldn't look at if something is visually hidden, but if it's exposed to screen readers. @stephenmathieson is working on a fix.

PR #1187, opened 2018-10-10
fix: consider element's accessible names when labels are hidden

### LINKED ISSUE #1176: Browser extension: axe is ignoring hidden labels for input fields even when it should acknowledge them

stephenmathieson:
Closes #1176

## Reviewer checks

- [X] Has documentation updated, a DU ticket, or requires no documentation change
- [X] Includes new tests, or was unnecessary
- [X] Code is reviewed for security by: Marcy Sutton


lib/checks/label/hidden-explicit-label.js
```diff
@@ -2,9 +2,12 @@ if (node.getAttribute('id')) {
 	const root = axe.commons.dom.getRootNode(node);
 	const id = axe.commons.utils.escapeSelector(node.getAttribute('id'));
 	const label = root.querySelector(`label[for="${id}"]`);
+	const isLabelVisible = label && axe.commons.dom.isVisible(label);
 
-	if (label && !axe.commons.dom.isVisible(label)) {
-		return true;
+	if (isLabelVisible === false) {
```
WilcoFiers: I think this now fails if there is no label. Should probably just put the if statement back to the way it was.

stephenmathieson: Good catch. Will update!

lib/checks/label/hidden-explicit-label.js
```diff
@@ -4,7 +4,9 @@ if (node.getAttribute('id')) {
 	const label = root.querySelector(`label[for="${id}"]`);
 
 	if (label && !axe.commons.dom.isVisible(label)) {
```
WilcoFiers: We should change this so it does `!axe.commons.dom.isVisible(label, true))`.

Here's a test case for it:

```js
it('should fail when the label has aria-hidden=true', function() {
	var html = '';
	html += '<div>';
	html += '  <label for="target" aria-hidden="true">';
	html += '    Hello world';
	html += '  </label>';
	html += '  <input id="target">';
	html += '</div>';
	var args = checkSetup(html, {}, '#target');
	assert.isTrue(check.evaluate.apply(check, args));
});
```

stephenmathieson: Good call! Updated.

PR #1188, opened 2018-10-10
fix: check for required functionality before running an analysis

### LINKED ISSUE #1136: TypeError: Cannot read property 'querySelectorAll' of undefined
Injecting `axe-core` and calling `axe.run()` on http://speedtest.att.com/speedtest/ produces an uncaught exception:

```
VM367:1 Uncaught (in promise) TypeError: Cannot read property 'querySelectorAll' of undefined
    at t (<anonymous>:1:62019)
    at Object.axe.utils.getSelectorData (<anonymous>:1:62440)
    at Object.x [as _runRules] (<anonymous>:1:35739)
    at Object.axe.run (<anonymous>:1:37950)
    at <anonymous>:1:5
```

This is an issue with 3.0.3 and 3.1.1 (and probably other versions, but I haven't tested them).

I have only just begun debugging this, and will update if/when I find any more information.

stephenmathieson:
This patch adds a few feature checks to the axe-core library. This ensures axe-core is able to rely on native methods that the host environment may not have (or may have overwritten with broken code).

If the host environment is missing functionality, a helpful "SupportError" will be raised rather than an unexpected crash.

I'm not sure if I've added these feature tests in the correct place. Happy to move them if need be :)

Closes #1136.

## Reviewer checks

### DIFF
--- a/lib/core/index.js
+++ b/lib/core/index.js
@@ -5,6 +5,72 @@
 var axe = axe || {};
 axe.version = '<%= pkg.version %>';
 
+// Feature-test a few native methods. This allows axe to exit early
+// with a helpful error message if the page is either missing
+// required JavaScript features, or has overwritten them with
+// broken code (see https://github.com/dequelabs/axe-core/issues/1136).
+
+let testResult;
+
+// Array.prototype.some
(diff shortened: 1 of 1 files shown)


stephenmathieson:
Adding polyfills seems like a much better solution. Will open a new PR.

Issue #1208, opened 2018-10-28
Update dom.isFocusable function (PR#1211)


jkodu:
- remove the usage of `isVisible`.
- write a new function `isCssVisible` and swap usage of `isVisible`.

isCssVisible should
- ignore `aria-hidden`.
- ignore off-screen elements.
- ensure parent visibility inheritance.

jkodu:
PR merged. Closing issue.

PR #1211, opened 2018-10-29
fix: introduce dom.isHiddenWithCSS for use in dom.isFocusable

jkodu:
This PR does the below:
- creates a new function `isCssVisible` to satisfy requirements from issue https://github.com/dequelabs/axe-core/issues/1208
- swaps the `dom.isVisible` usage in focus computation with `dom.isCssVisible`
- these changes are necessary for completing the rule `aria-hidden-focus` - https://github.com/dequelabs/axe-core/pull/1166, which exists as a WIP PR.

Closes issue:
- https://github.com/dequelabs/axe-core/issues/1208

## Reviewer checks

- [X] Has documentation updated, a DU ticket, or requires no documentation change
- [X] Includes new tests, or was unnecessary
- [X] Code is reviewed for security by: << Marcy Sutton >>


lib/commons/dom/is-css-visible.js
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	const nodeName = el.nodeName.toUpperCase();
+	if (['STYLE', 'SCRIPT'].includes(nodeName)) {
+		return false;
+	}
+
+	const style = window.getComputedStyle(el, null);
+	if (!style) {
+		return false;
```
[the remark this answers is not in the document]
jkodu: I think this may be a bit intrusive to `throw` for this reason. Happy to discuss.

jkodu:
@WilcoFiers - ready for review.

lib/commons/dom/is-hidden-with-css.js
```diff
@@ -0,0 +1,50 @@
...
+		el = el.host; // swap to host node
+	}
+
+	if (['STYLE', 'SCRIPT'].includes(el.nodeName.toUpperCase())) {
+		return false;
+	}
+
+	const style = window.getComputedStyle(el, null);
+	if (!style) {
+		throw new Error('Style does not exist for the given element.');
```
marcysutton: Where would this scenario occur?

lib/commons/dom/is-hidden-with-css.js
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Determine whether an element is hidden based on css
+ * @method isHiddenWithCSS
+ * @memberof axe.commons.dom
+ * @instance
+ * @param {HTMLElement} el The HTML Element
+ * @param {Boolean} descendentVisibilityValue (Optional) immediate descendant visibility value used for recursive computation
+ * @return {Boolean} the element's hidden status
+ */
+dom.isHiddenWithCSS = function isHiddenWithCSS(el, descendentVisibilityValue) {
```
marcysutton: I like the new name for this method. 👍 

Are you pulling any of the code out of is-visible, or is that mostly duplicate at this point?

test/commons/dom/is-hidden-with-css.js
```diff
@@ -0,0 +1,225 @@
...
+
+	it('should return false for off-screen fixed positioned element', function() {
+		fixture.innerHTML =
+			'<div id="target" style="position: fixed; top: -9999px">Hi</div>';
+		var el = document.getElementById('target');
+		var actual = isHiddenWithCSSFn(el);
+		assert.isFalse(actual);
+	});
+
+	it('should return false on detached elements', function() {
```
marcysutton: When would this case arise? Would these even be returned from querying the DOM?

lib/commons/dom/is-hidden-with-css.js
WilcoFiers: I think this happens with unmounted elements.

lib/commons/dom/is-hidden-with-css.js
```diff
@@ -0,0 +1,50 @@
...
+
+	if (style.getPropertyValue('display') === 'none') {
+		return true;
+	}
+
+	const visibilityValue = style.getPropertyValue('visibility');
+	if (
+		(visibilityValue === 'hidden' && !descendentVisibilityValue) ||
+		(visibilityValue === 'hidden' &&
+			(descendentVisibilityValue && descendentVisibilityValue !== 'visible'))
```
WilcoFiers: should be `=== 'hidden'` There are more values on CSS visibility than just hidden and visible.

lib/commons/dom/is-hidden-with-css.js
```diff
@@ -0,0 +1,50 @@
...
+	) {
+		return true;
+	}
+
+	const parent = dom.getComposedParent(el);
+	if (parent) {
+		return dom.isHiddenWithCSS(parent, visibilityValue);
```
WilcoFiers: Nice try on the recursion, but it doesn't work this way. This function shows `#foo` as hidden, which it isn't:

```html
<div style="visibility:hidden">
  <div style="visibility:hidden">
    <div style="visibility:visible" id="foo">
      Visible
    </div>
  </div>
</div>
```

lib/commons/dom/is-hidden-with-css.js
jkodu: This has some replication from `is-visible` and `is-hidden`. We may need to do some refactor afterwards.

test/commons/dom/is-hidden-with-css.js
jkodu: You are right, probably won't be returned when querying the DOM. No harm in this test case, as that is exactly what we are testing.

lib/commons/dom/is-hidden-with-css.js
jkodu: Have tackled for the above.

lib/commons/dom/is-hidden-with-css.js
marcysutton: I don't even know how that would come up in the context of an audit, but ok :)

Issue #942, opened 2018-06-06
Refactor: To accommodate namespace for tagName determination (PR#1219)


jkodu:
Based on discussion:
- https://github.com/dequelabs/axe-core/pull/623#discussion_r159419441

jkodu:
@WilcoFiers 
I have gone through this, and need your opinion before any work is done.

See comment put in description. It looks like in some places we use `node.tagName` and other `node.nodeName`. Further, in some places it is `toLowerCase()`, whilst others are `toUpperCase()`.

Question based on the discussion though, there seems to agreement on 2 things.

1) Implement a new `getTagName` function which returns preferably an `uppercased` string, also considering `namespace` eg: `HTML:ARTICLE` or `SVG:A` based on the usage of the node.

2) We simply just walk away from `node.tagName` usage to `node.nodeName.toUpperCase()`

Which route would you like to go with this?

jkodu:
@WilcoFiers - assigning this to you, to get above Q answered.

WilcoFiers:
We should always use `node.nodeName.toUpperCase()`.

jkodu:
PR Merged. Closing issue.

PR #1219, opened 2018-11-02
fix: Avoid IE problems by using nodeName instead of tagName

jkodu:
This is a maintenance PR, to update `node.tagName` usages to `node.nodeName`.

Did not update the usages in:
- `lib/core/utils/css-parser.js` - as this is a copy/pasted (imported) module.
- `lib/core/utils/qsa.js` - as this seems to break a lot of tests.

Closes issue:
- https://github.com/dequelabs/axe-core/issues/942

## Reviewer checks

### DIFF
--- a/lib/checks/navigation/heading-order.js
+++ b/lib/checks/navigation/heading-order.js
@@ -5,7 +5,7 @@ if (ariaHeadingLevel !== null) {
 	return true;
 }
 
-var headingLevel = node.tagName.match(/H(\d)/);
+var headingLevel = node.nodeName.match(/H(\d)/);
 
 if (headingLevel) {
 	this.data(parseInt(headingLevel[1], 10));
--- a/lib/checks/navigation/heading-order.js
+++ b/lib/checks/navigation/heading-order.js
@@ -5,7 +5,7 @@ if (ariaHeadingLevel !== null) {
 	return true;
 }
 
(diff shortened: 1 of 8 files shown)


lib/checks/navigation/heading-order.js
```diff
@@ -5,7 +5,7 @@ if (ariaHeadingLevel !== null) {
 	return true;
 }
 
-var headingLevel = node.tagName.match(/H(\d)/);
+var headingLevel = node.nodeName.match(/H(\d)/);
```
marcysutton: do you need `toUpperCase()` here?

marcysutton:
I'm seeing additional usage in `commons/color/get-background-color`, should that be updated as well?

lib/checks/navigation/heading-order.js
jkodu: I am going to leave this on as is (no need to uppercase).

jkodu:
@marcysutton - the usage in `commons/color/get-background-color`, seems to me like an object with property `tagName` rather than an `HTMLElement` itself. Hence did not change it (as doing so broke a lot of tests).

lib/checks/navigation/heading-order.js
WilcoFiers: I disagree. In xhtml nodeName is lower case, which means this won't match.

Issue #1194, opened 2018-10-17
aria-allowed-role should ignore hidden roles


WilcoFiers:
I just found this case in the wild:

```html
<div style="display:none">
  <button role="presentation">Some button</button>
</div>
```

Adding that role is completely pointless, but it isn't an accessibility issue. We should mark these things for review IMO.

marcysutton:
It would fall under a "ghost controls" rule as well, due to the aria-hidden and tabindex on it. Definitely an accessibility issue.

jkodu:
@WilcoFiers, @marcysutton 

Well, in today's implementation, we flag this as a violation with the message - `role presentation  is not allowed for given element`.

Should we instead mark it as `needs review`?

jkodu:
PR Merged. Closing issue.

PR #1225, opened 2018-11-07
fix: flag hidden elms with disallowed role(s) for review

jkodu:
This PR does the below for the rule `aria-allowed-role`

- Note: The rule spec does not exclude hidden elements `excludeHidden: false` (this is by choice).
- Change has been put in to flag elements as `needs review`, which are hidden (via CSS and to screen readers), and have redundant/ disallowed roles.

Closes issue:
- https://github.com/dequelabs/axe-core/issues/1194

## Reviewer checks

- [x] Code is reviewed for security by @WilcoFiers


lib/checks/aria/aria-allowed-role.js
```diff
@@ -18,6 +19,10 @@ const unallowedRoles = axe.commons.aria.getElementUnallowedRoles(
 
 if (unallowedRoles.length) {
 	this.data(unallowedRoles);
+	if (!dom.isVisible(node, true)) {
+		// flag hidden elements for review
+		return null;
```
WilcoFiers: This should return undefined. I know this also works, but we should do this consistently.

lib/checks/aria/aria-allowed-role.js
```diff
@@ -3,6 +3,7 @@
  * https://www.w3.org/TR/html-aria/#docconformance
  * https://www.w3.org/TR/SVG2/struct.html#implicit-aria-semantics
  */
+const { dom } = axe.commons;
```
WilcoFiers: Please add an "inapplicable" message to the check JSON so that we communicate with the user what's going on.

jkodu: @WilcoFiers 

I have gone with below. Let me know if you have suggestions.

`"incomplete": "Element is not visible, and ARIA role{{=it.data && it.data.length > 1 ? 's' : ''}} {{=it.data.join(', ')}} {{=it.data && it.data.length > 1 ? 'are' : ' is'}} not allowed for given element"`

jkodu: @marcysutton - any suggestions for the incomplete message - https://github.com/dequelabs/axe-core/pull/1225/files#diff-2f072a5102050cd26c55026d5d9f9c4fR13

test/checks/aria/aria-allowed-role.js
```diff
@@ -51,6 +51,32 @@ describe('aria-allowed-role', function() {
 		);
 	});
 
+	it('returns undefined (needs review) when element is hidden and has redundant role', function() {
```
WilcoFiers: This is not a "redundant role".

test/checks/aria/aria-allowed-role.js
```diff
@@ -51,6 +51,32 @@ describe('aria-allowed-role', function() {
...
+			'role="presentation"></button>';
+		var target = fixture.querySelector('#target');
+		var actual = checks['aria-allowed-role'].evaluate.call(
+			checkContext,
+			target
+		);
+		assert.isUndefined(actual);
+	});
+
+	it('returns undefined (needs review) when element is with in hidden parent and has redundant role', function() {
```
WilcoFiers: Same here, not a "redundant role".

PR #1238, opened 2018-11-14
docs: Add intro to aria-supported page

### LINKED ISSUE #1038: ARIA Supported doc needs more details
The recently-added [aria-supported.md](https://github.com/dequelabs/axe-core/blob/develop/doc/aria-supported.md) doc is pretty light on details. I know that it consists of a diff generated from aria-query, so our options are somewhat limited. But I don't think an issue was ever filed to make it more robust.

We had talked about this internally as part of Support:

> explain what we’re seeing here, what/when to expect next version
> - Add text indicating this is about ARIA 1.1
> - Indicate why each is not supported

We make no mention of ARIA 1.1, which was partially the reason this doc was created. There are also no reasons or explanations as to why things aren't supported.

Perhaps we could use some of the data from #918 to add more details to this doc? I suppose that would only work if the techniques were ARIA related. Otherwise they should probably go into the [accessibility-supported.md](https://github.com/dequelabs/axe-core/blob/develop/doc/accessibility-supported.md) doc.

WilcoFiers:
This adds some context to the aria-supported page.

Closes #1038

## Reviewer checks


build/tasks/aria-supported.js
```diff
@@ -84,7 +84,10 @@ module.exports = function(grunt) {
 				const content = getMdContent(
 					`# ARIA Roles and Attributes ${
 						listType === 'all' ? 'available' : listType
-					} in axe-core.`,
+					} in axe-core.\n\n` +
```
marcysutton: Can you limit the character width of this string literal to reduce horizontal scrolling? I've worked on teams with standards of 80 or 100 chars.

doc/aria-supported.md
```diff
@@ -1,5 +1,11 @@
 # ARIA Roles and Attributes unsupported in axe-core.
 
+It can be difficult to know which features of web technologies are accessible across different platforms, and with different screen readers and other assistive technologies. Axe-core does some of this work for you, by raising issues when accessibility features are used that are known to cause problems.
+
+This page contains a list of ARIA 1.1 features that axe-core raises as unsupported. To learn more about accessibility support in our article [We’ve got your back with “Accessibility Supported” in axe](https://www.deque.com/blog/weve-got-your-back-with-accessibility-supported-in-axe/).
```
marcysutton: This sentence isn't quite complete, but it could be easily fixed by removing the first word:
`Learn more about accessibility support in our article...`

WilcoFiers:
@marcysutton can you review again?

Issue #795, opened 2018-03-20
New rule: Ensure complementary landmarks are top level


iamrafan:
# landmark-complementary-is-top-level

**Description:** Ensure complementary landmarks are top level landmarks

### Tags 
best-practice

## Selector

```<aside>, [role="complementary"]```

### ( none )

1. Check that none of the complementary landmarks are nested inside any HTML5 or role landmarks

WilcoFiers:
Agree, this makes sense. We'll see about including this into our plans for 3.x

dylanb:
What is the basis for this? What is the impact on accessibility if these are not top level?

WilcoFiers:
Confusing page structure. It's pretty low impact I'd say - it is one of the ARIA best practices.

dylanb:
I'm worried about noise - experimental?

WilcoFiers:
Can you elaborate on "noise"?

dylanb:
Developers having to sift through false positives that have dubious impact on users

WilcoFiers:
I wouldn't say that any of the aria best practices are dubious. They are as close to a predefined best practices list as we have.

WilcoFiers:
@dylanb I want to move forward on this, any objection?

nickcolley:
Hello @WilcoFiers !

> Developers having to sift through false positives that have dubious impact on users

I run into this issue a lot. In our most recent case we're using `<aside>` inside the page to show an 'Recommended story'. Sort of an inline advert for another article.

This aligns well with the HTML definition of how to use `<aside>`.

It'd be helpful for us to have any research or evidence that led to introducing this rule, what impact does it have on real users of assistive technologies?

BurntCaramel:
I agree with @nickcolley — I found this rule surprising. Is the solution that I should try to move all my `<aside>` elements outside my `<main>` (I’m not really sure this is possible). Or should I change my `<aside>` to a `<section>` or `<div>`?

I followed the HTML spec’s suggested usage so I find this seems to not align with that? Should the HTML spec say something else?

dylanb:
@WilcoFiers I think we should post some information here on why this is better for users

kryp71c:
[...]

PR #1239, opened 2018-11-14
feat: new rule landmark-complementary-is-top-level

marcysutton:
Best practice requiring asides and complementary landmarks to be top level, in line with the ARIA Authoring Practices Guide.

Closes https://github.com/dequelabs/axe-core/issues/795

## Reviewer checks


lib/rules/landmark-has-body-context.js
```diff
@@ -1,7 +1,7 @@
 const nativeScopeFilter = 'article, aside, main, nav, section';
 
 // Filter elements that, within certain contexts, don't map their role.
-// e.g. a <footer> inside a <main> is not a banner, but in the <body> context it is
+// e.g. a <header> inside a <main> is not a banner, but in the <body> context it is
 return (
 	node.hasAttribute('role') ||
```
jkodu: Could this be extracted to `"selector": "[role]"`?

test/checks/keyboard/landmark-is-top-level.js
```diff
@@ -23,6 +23,16 @@ describe('landmark-is-top-level', function() {
 		assert.deepEqual(checkContext._data, { role: 'main' });
 	});
 
+	it('should return false if the complementary landmark is in another landmark', function() {
+		var mainLandmark = document.createElement('main');
+		var aside = document.createElement('div');
+		aside.setAttribute('role', 'complementary');
+		aside.appendChild(mainLandmark);
+		fixture.appendChild(aside);
```
jkodu: Nitpick: I know, there are multiple tests written with DOM api like above using `createElement`, `appendChild` etc. I found that using the `testutils.fixtureSetup()` function to be helpful for this, and saves us a few lines and also makes it easy to read.

test/checks/keyboard/landmark-is-top-level.js
```diff
@@ -23,6 +23,16 @@ describe('landmark-is-top-level', function() {
 		assert.deepEqual(checkContext._data, { role: 'main' });
 	});
 
+	it('should return false if the complementary landmark is in another landmark', function() {
```
jkodu: Also, just had a thought, what happens if the first `header` element after `body` element happens with in `shadowDOM`?

Question being, do we need to consider `shadowDOM` for this rule?

marcysutton: The check already has Shadow DOM coverage.

marcysutton:
@jkodu thanks for the tip. Does that require installing circleci globally on your machine?

lib/rules/landmark-has-body-context.js
marcysutton: I ended up not using this file, and the only thing I changed in it was a comment typo. Leaving as-is.

test/checks/keyboard/landmark-is-top-level.js
marcysutton: I've updated the tests, you're right that they were more difficult to read!

lib/rules/landmark-has-body-context.js
WilcoFiers: I had to look it up too, but yes aside always maps to role= complementary, unlike footer and header which only map to a role if they are scoped to the body.

PR #1240, opened 2018-11-15
chore: cli for axe rule generation

jkodu:
This PR does the below:
- Provides a CLI for axe RULE creation, by creating rules, associated checks and tests assets. 
- handles/ validates against empty rule or check names.
- handles/ validates against pre-existing rule or check files and does not allow for overrides.
- gives option to create matches file.
- gives option to recursively create multiple checks per rule.
- gives option to create unit and integration test files.
- generated files have helper notes and skeleton code.
- To test locally, clone branch and run `npm install` & `npm run rule-gen`.

image

Closes issue: NA

## Reviewer checks


bin/rule-generator.js
```diff
@@ -0,0 +1,448 @@
...
+const path = require('path');
+const chalk = require('chalk');
+const figlet = require('figlet');
+const inquirer = require('inquirer');
+const cmd = require('node-command-line');
+const axeFile = path.join(__dirname, '..', 'axe.js');
+const filehound = require('filehound');
+
+const DIRECTORIES = {
+	RULES: path.join(__dirname, '..', '/lib/rules'),
```
stephenmathieson: Should replace the path separators here and let the `path` module handle it for you. This helps support other platforms (Windows).

For example:

```js
{
  RULES: path.join(__dirname, '..', 'lib', 'rules'),
  // ...
}
```

bin/rule-generator.js
```diff
@@ -0,0 +1,448 @@
...
+	TEST_CHECKS_INTEGRATION: path.join(__dirname, '..', '/test/integration/rules')
+};
+
+const showBanner = () => {
+	const COLOR_DQ_HEX = '#0077c8';
+	console.log(chalk.hex(COLOR_DQ_HEX)(figlet.textSync('Axe Rule Generator')));
+};
+
+const QUESTIONS = {
+	GET_RULE_NAME: {
```
stephenmathieson: Not a big deal, but WHY IS NEARLY EVERY VARIABLE IN THIS FILE UPPER-CASED? 😄

bin/rule-generator.js
```diff
@@ -0,0 +1,448 @@
...
+
+	const ruleName = `${GET_RULE_NAME.toLowerCase()}`;
+
+	const ruleJson = {
+		name: `${ruleName}.json`,
+		content: JSON.stringify(
+			{
+				id: `${ruleName}`,
+				selector: '',
+				...(GET_IS_RULE_MATCHES && {
```
stephenmathieson: What is happening here? These statements are a bit difficult to follow.

bin/rule-generator.js
```diff
@@ -0,0 +1,448 @@
...
+			chalk.black.bgRed.bold(
+				`Axe does not exist.Triggering build using - 'npm run build'.Rule Generator will continue after build.`
+			)
+		);
+		await cmd.run('npm run build');
+	}
+
+	// compute essentials
+	// 1. Get rule SPEC json file names
+	const ruleJsonSpecFiles = await filehound
```
stephenmathieson: Consider replacing `filehound` with [`globby`](https://www.npmjs.com/package/globby). Globby has a much more familiar API and doesn't require you to read its documentation to figure out how to use it; just give it a glob pattern and you're good to go.

bin/rule-generator.js
```diff
@@ -0,0 +1,448 @@
...
+		.create()
+		.paths(DIRECTORIES.RULES)
+		.depth(0)
+		.ext('json')
+		.find();
+	axeRulesFileNames = ruleJsonSpecFiles.map(
+		f =>
```
stephenmathieson: Please add a comment here. What is this doing/why?

package.json
```diff
@@ -55,6 +55,7 @@
     }
   },
   "scripts": {
+    "axe-rule-gen": "node bin/rule-generator",
```
stephenmathieson: Not sure we need to prefix this with `axe-`. When you run it, you'll be doing so from within the axe-core repo, so it's probably a little redundant.

package.json
```diff
@@ -64,7 +65,10 @@
     "release": "standard-version -a",
     "next-release": "standard-version --scripts.prebump=./build/next-version.js --skip.commit=true --skip.tag=true",
     "sri-update": "grunt build && node build/sri-update && git add sri-history.json",
-    "fmt": "prettier --write  *.js, **/*.ts '{build,doc,lib,test}/**/*.js'"
+    "fmt": "prettier --write  *.js, **/*.ts '{bin,build,doc,lib,test}/**/*.js'"
+  },
+  "bin": {
```
stephenmathieson: Why are you adding this?

stephenmathieson: Since the binary is using development dependencies, this won't work.

Please remove `"bin": { ... }`.

package.json
```diff
@@ -95,17 +102,20 @@
...
     "html-entities": "^1.2.0",
     "husky": "^1.1.1",
+    "inquirer": "^6.2.0",
     "jquery": "^3.0.0",
     "jsdoc": "^3.5.5",
     "lint-staged": "^8.0.0",
     "markdown-table": "^1.1.2",
     "minami": "^1.2.3",
     "mocha": "^5.2.0",
+    "node-command-line": "^1.0.2",
```
stephenmathieson: Consider replacing this with [`execa`](https://www.npmjs.com/package/execa). It's *much* better supported/trusted, and handles tons of weird Windows edge cases.

package.json
```diff
@@ -95,17 +102,20 @@
...
     "prettier": "^1.14.0",
     "promise": "~8.0.1",
     "retire": "^2.0.1",
     "revalidator": "~0.3.1",
     "selenium-webdriver": "~3.6.0",
+    "shelljs": "^0.8.3",
```
stephenmathieson: Why was this added?

package.json
marcysutton: It does make sense from a global context...but do we really need that? I don't see much of a reason to have this exist globally since the feature only applies to this repo.

bin/rule-generator.js
```diff
@@ -0,0 +1,448 @@
...
+			// 1) check if valid rule name
+			if (
+				!(
+					ruleName &&
+					ruleName.length > 0 &&
+					ruleName.search(/^[a-zA-Z-]+$/) !== -1
+				)
+			) {
+				throw new Error(
+					'RULE name should not be empty and can contain alphabets and dashes.'
```
marcysutton: We should add the word "only" here, as in "can only contain alphabets and dashes."

bin/rule-generator.js
```diff
@@ -0,0 +1,448 @@
...
+			}
+			return true;
+		}
+	},
+	GET_CHECK_CATEGORY: {
+		/**
+		 * Choose category for the CHECK? :
+		 * 	(Eg: aria, color, forms, label, language, media...)
+		 */
+		name: 'GET_CHECK_CATEGORY',
```
marcysutton: What are these redundant name properties for?

bin/rule-generator.js
```diff
@@ -0,0 +1,448 @@
...
+		 * Create INTEGRATION test files? (Yes/ No)
+		 */
+		name: 'GET_IS_INTEGRATION_TEST_ASSETS',
+		type: 'confirm',
+		message: 'Would you like to create INTEGRATION test files?'
+	}
+};
+
+const questionAndGetFilesToCreate = async () => {
+	const getCheck = async (checks = []) => {
```
marcysutton: Should this be plural `getChecks` since it returns an array of checks?

bin/rule-generator.js
```diff
@@ -0,0 +1,448 @@
...
+	let filesToGenerate = [];
+
+	const ruleName = `${GET_RULE_NAME.toLowerCase()}`;
+
+	const ruleJson = {
+		name: `${ruleName}.json`,
+		content: JSON.stringify(
+			{
+				id: `${ruleName}`,
+				selector: '',
```
marcysutton: Should there be a prompt for the `selector` so new users don't forget to change it? I'm thinking of the Hackathon scenario. Same for tags and metadata. Maybe we could add a flag to skip certain fields in case you don't want to fill them all out, sort-of like the behavior of `npm init -y`.

bin/rule-generator.js
```diff
@@ -0,0 +1,448 @@
...
+let axeRulesIds;
+let axeChecksFileNames;
+let axeChecksIds;
+
+const run = async () => {
+	// ensure axe exists, if not build axe, then start the generator
+	const axeExists = fs.existsSync(axeFile);
+	if (!axeExists) {
+		console.log(
+			chalk.black.bgRed.bold(
```
marcysutton: Do these colors work on light and dark terminal themes?

bin/rule-generator.js
```diff
@@ -0,0 +1,448 @@
...
+				`Axe does not exist.Triggering build using - 'npm run build'.Rule Generator will continue after build.`
+			)
+		);
+		await cmd.run('npm run build');
```
marcysutton: Slick 👍

bin/rule-generator.js
```diff
@@ -0,0 +1,448 @@
...
+
+	// rule-generator banner
+	showBanner();
+
+	// questions
+	const files = await questionAndGetFilesToCreate();
+
+	// create the files
+	if (!files || !files.length) {
+		console.log(chalk.white.bgRed.bold(`No files to generate.`));
```
marcysutton: Same question as above (and below): do these colors work on a variety of themes?

bin/rule-generator.js
jkodu: Meh! I have to break away from this habit. I blame my c# past.

bin/rule-generator.js
jkodu: @marcysutton I do not follow, which of these is redundant?

bin/rule-generator.js
jkodu: @marcysutton 
Can we add these as new requirements please.
All of the above are possible.

Just create an issue and assign to me. I will prioritise when I have downtime.

bin/rule-generator.js
jkodu: Well, extending the object with another property based on a condition. Not too keen to change it.

bin/rule-generator.js
jkodu: Also @marcysutton - I have added some basic docs to the developer guide, would like your opinion on the same.

package.json
```diff
@@ -77,10 +78,15 @@
     "babelify": "^10.0.0",
     "blanket": "~1.2.3",
     "chai": "~4.1.2",
+    "chalk": "^2.4.1",
     "clone": "~2.1.1",
     "dot": "~1.1.2",
     "eslint-config-prettier": "^3.0.0",
+    "execa": "^1.0.0",
+    "figlet": "^1.2.1",
+    "filehound": "^1.16.4",
```
stephenmathieson: I believe this can now be removed

bin/rule-generator.js
marcysutton: The object and the `name` property. I couldn't find where `name` was used.

package.json
jkodu: Done.

bin/rule-generator.js
```diff
@@ -0,0 +1,455 @@
...
+		__dirname,
+		'..',
+		'test',
+		'integration',
+		'rules'
+	)
+};
+
+const showBanner = () => {
+	console.log(chalk.hex('#0077c8')(figlet.textSync('Axe Rule Generator')));
```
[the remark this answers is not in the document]
jkodu: Tested on both dark and white backgrounds in terms of contrast. See - https://github.com/dequelabs/axe-core/pull/1240#discussion_r234565788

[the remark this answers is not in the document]
marcysutton: Wilco makes a good point: I do love the ASCII treatment but it's inaccessible to screen reader users. We either need to repeat the "Axe Rule Generator" text in plain text or get rid of it. 

I've run into it with a slide deck before, and even though I had the text repeated below I still got a complaint about it from a blind person. https://github.com/marcysutton/empathy-driven-development

marcysutton:
The functionality works great. One thing I'm seeing is an extra / in some of the check file paths:
```
Successfully generated RULE and respective files:

/Users/marcysutton/Sites/Deque/axe-core/lib/rules/color-pseudo.json
/Users/marcysutton/Sites/Deque/axe-core/lib/rules/color-pseudo-matches.js
/Users/marcysutton/Sites/Deque/axe-core/test/rule-matches/color-pseudo-matches.js
/Users/marcysutton/Sites/Deque/axe-core/lib/checks/color//color-pseudo.json //<- this one
/Users/marcysutton/Sites/Deque/axe-core/lib/checks/color//color-pseudo.js //<- this one
/Users/marcysutton/Sites/Deque/axe-core/test/checks/color//color-pseudo.js //<- this one
/Users/marcysutton/Sites/Deque/axe-core/test/integration/rules/color-pseudo/color-pseudo.html
/Users/marcysutton/Sites/Deque/axe-core/test/integration/rules/color-pseudo/color-pseudo.json
```

marcysutton:
Also, be sure to update the PR description with the correct info.

build/rule-generator.js
```diff
@@ -0,0 +1,53 @@
...
+		console.log(
+			chalk.red.bold(
+				`Axe does not exist. Triggering build using - 'npm run build'. Rule generation will continue after build.`
+			)
+		);
+		await execa.shell('npm run build');
+	}
+
+	// rule-generator banner
+	console.log(chalk.hex('#0077c8')(figlet.textSync('Axe Rule Generator')));
```
WilcoFiers: This figlet should be removed. It's not accessible.

jkodu:
@marcysutton - thanks for the catch on the generated file path anamoly. Updated now.
@WilcoFiers - removed figlet.

Review welcome again.

jkodu:
@marcysutton - this needs an approval from you, when you have a moment.

build/rule-generator.js
```diff
@@ -0,0 +1,52 @@
...
+	const files = getFilesMetaData(answers);
+
+	// create the files
+	if (!files || !files.length) {
+		console.log(chalk.red.bold(`No files to generate.`));
+	}
+
+	try {
+		const result = await Promise.all(files.map(createFile));
+		console.log(
```
marcysutton: The only thing I'd add to this is an extra space before "Successfully generated" to give a little more breathing room between sections. Otherwise it works great!

PR #1245, opened 2018-11-19
chore: remove unused "has-listitem" check

### LINKED ISSUE #1127: Unnecessary/unused check "has-listitem"
Currently, axe-core (version 3.1.1) contains a check called [`has-listitem`](https://github.com/dequelabs/axe-core/blob/c8b1d6290401a5bd0150dd6bca9549916f9d9c5e/lib/checks/lists/has-listitem.json), but it is not used by any rules. Because it's not used by any rules, it never receives a `.data` property in the `axe._audit` object, which prevents runtime localization from working.

When we attempt to apply locale at runtime, this error is raised:

https://github.com/dequelabs/axe-core/blob/c8b1d6290401a5bd0150dd6bca9549916f9d9c5e/lib/core/base/audit.js#L188

This is due to the existing locale for the check:

https://github.com/dequelabs/axe-core/blob/c8b1d6290401a5bd0150dd6bca9549916f9d9c5e/locales/de.json#L432-L435

Link to live demo: https://codepen.io/anon/pen/xaYrYE

----

My suggestion is we remove the unused check, since it's currently only adding confusion and size to the built `axe.js` file.

Alternately, we can remove the runtime exception, but that will prevent users from seeing problems in their locale files.

stephenmathieson:
This patch removes the unused "has-listitem" check.

I've confirmed with both @akornmeier and @downzer0 that this check is not used in any custom rules, so this should not be considered a breaking change.

Closes #1127.

## Reviewer checks


marcysutton:
No knowledge of this check in custom rules from the internal team. I'd say this is safe to remove as it hasn't been referenced by a rule in over 3 years.

jkodu:
@marcysutton - please fill out the reviewer comments.

Issue #864, opened 2018-04-26
region rule should ignore content of noscript elements


nico3333fr:
Hi there, 

I've used it on https://van11y.net/ thanks to Sonarwhal https://sonarwhal.com/scanner/39c2fa0b-cfaa-49c9-bc5f-3564a2eaec21 

It reported me this error: "Content should be contained in a landmark region"

It seems (look at the thread on Twitter https://twitter.com/molant/status/989497275134001154) that it is due to:

```
<noscript><p class="invisible"><img src="/piwik/piwik.php?idsite=1" alt="" /></p></noscript>
```

This piece of code is for Matomo Analytics (ex-Piwik). It seems to be a false positive.

I hope this will help you to improve your tool. 

Cheers,
Nicolas

jkodu:
@WilcoFiers 
I was looking to pick this up and the first point of fix was to introduces a `region-matches.js` for region rule and ignore `noscript` and its children. 

But your comment reads, `axe-core` should ignore `noscript`. Am I right in thinking we need to make this a global filter of some sort, even before executing the matches? 

Thoughts?

jkodu:
Not able to reproduce the issue.

PR #1253, opened 2018-11-21
fix: exclude noscript and template from dom.isVisible

jkodu:
PR does the below:
- excludes `NOSCRIPT` and `TEMPLATE` when computing `isVisible`.

Closes issue: https://github.com/dequelabs/axe-core/issues/864

## Reviewer checks

### DIFF
--- a/lib/commons/dom/is-visible.js
+++ b/lib/commons/dom/is-visible.js
@@ -1,5 +1,5 @@
 /* global dom */
-/* eslint complexity: ["error", 18] */
+/* eslint complexity: ["error", 20] */
 
 /**
  * Determines if an element is hidden with the clip rect technique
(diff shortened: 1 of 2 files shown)


lib/commons/dom/is-visible.js
```diff
@@ -58,6 +58,8 @@ dom.isVisible = function(el, screenReader, recursed) {
 		style.getPropertyValue('display') === 'none' ||
 		nodeName.toUpperCase() === 'STYLE' ||
 		nodeName.toUpperCase() === 'SCRIPT' ||
+		nodeName.toUpperCase() === 'NOSCRIPT' ||
+		nodeName.toUpperCase() === 'TEMPLATE' ||
```
WilcoFiers: I'd suggest an `array.includes` solution instead of adding even more "or" cases.

PR #1254, opened 2018-11-21
feat: ARIA supported checks

marcysutton:
This branch adds a new check to report unsupported ARIA attributes to the user. We have to make a decision on how to modify the rule to support it, which you can find discussed in #918.

Closes issue: https://github.com/dequelabs/axe-core/issues/918

## Reviewer checks


lib/rules/aria-allowed-attr.json
```diff
@@ -7,6 +7,6 @@
 		"help": "Elements must only use allowed ARIA attributes"
 	},
 	"all": [],
-	"any": ["aria-allowed-attr"],
+	"any": ["aria-allowed-attr", "aria-unsupported-attr"],
```
WilcoFiers: What's the reasoning behind adding this to `aria-allowed-attr` instead of to `aria-valid-attr`? I haven't quite figured out which place I think is best, so I'd like to know your reasoning for it.

lib/checks/aria/unsupportedattr.js
```diff
@@ -0,0 +1,22 @@
...
+		})
+		.map(candidate => {
+			return candidate.name.toString();
+		});
+}
+if (unsupported.length) {
+	this.data(unsupported);
+	return true;
+}
+return false;
```
WilcoFiers: I'm not sure the logic works here. Shouldn't you return `true` when this passes? I would also think that this needs to be in `all` instead of `any` in the rule. That decision may be why you  used this strange logic.

lib/rules/aria-allowed-attr.json
marcysutton: It's a really fine line. I went for "allowed" because technically these attributes are valid, but we're recommending they not be used because of support purposes. Both `aria-allowed-attr` and `aria-valid-attr` use the same commons method to validate, except `aria-allowed-attr` has some more logic to check the role. The new unsupported check could go in either rule, really.

lib/commons/aria/index.js
```diff
@@ -10,177 +10,232 @@ var aria = (commons.aria = {}),
...
+		values: ['page', 'step', 'location', 'date', 'time', 'true', 'false'],
+		unsupported: false
 	},
 	'aria-describedby': {
 		type: 'idrefs',
-		allowEmpty: true
+		allowEmpty: true,
+		unsupported: false
+	},
+	'aria-describedat': {
```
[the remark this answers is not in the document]
marcysutton: I can leave it off, I only added it because it was mentioned in the aria-supported diff. But if we did include it, if someone tried to use it they'd see it was unsupported. I like the idea of an "unstandardized" flag in here.

lib/commons/aria/index.js
```diff
@@ -10,177 +10,232 @@ var aria = (commons.aria = {}),
 lookupTable.attributes = {
 	'aria-activedescendant': {
 		type: 'idref',
-		allowEmpty: true
+		allowEmpty: true,
+		unsupported: false
```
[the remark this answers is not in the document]
marcysutton: Leaving them off and using falsey values instead sounds fine to me!

lib/rules/aria-allowed-attr.json
marcysutton: I still believe that the correct rule for this behavior is `aria-allowed-attr` rather than `aria-valid-attr`, so we can get the correct messaging. I'm going to move forward with the change to the matches function so we can keep this rolling.

lib/checks/aria/unsupportedattr.js
marcysutton: @WilcoFiers the `aria-roles` unsupported stuff works on `none`. I would actually think that should be the case here as well.

marcysutton:
This is ready for another review.

test/checks/media/description.js
```diff
@@ -40,7 +40,6 @@ describe('description', function() {
 			shadow.innerHTML = '<video><slot></slot></video>';
 
 			var checkArgs = checkSetup(node, {}, 'video');
-			axe.log(checkArgs);
```
marcysutton: This commit should have `test` in the scope, and is unrelated to this PR. But it kept irritating me when running tests.

lib/rules/aria-allowed-attr-matches.js
```diff
@@ -1,16 +1,9 @@
...
-		var attrs = node.attributes;
-		for (var i = 0, l = attrs.length; i < l; i++) {
-			if (aria.test(attrs[i].name)) {
-				return true;
-			}
+const aria = /^aria-/;
+if (node.hasAttributes()) {
+	let attrs = node.attributes;
+	for (let i = 0, l = attrs.length; i < l; i++) {
+		if (aria.test(attrs[i].name)) {
```
WilcoFiers: Sorry, should have noticed this in my first review. It seems like you're not filtering out elements with only invalid ARIA attributes:

```
<input aria-invalidattr="foo" />
```

This element shouldn't be applicable to this rule. It's a pretty minor bug, so I'll leave it for you to decide if you want to bother with it. The fix is to use `axe.commons.aria.validateAttr` in the if statement here.

test/checks/aria/unsupportedattr.js
```diff
@@ -0,0 +1,52 @@
...
+		);
+		assert.isTrue(check.evaluate.apply(checkContext, params));
+	});
+
+	it('should return true with multiple unsupported and supported attributes', function() {
+		axe.commons.aria.lookupTable.attributes['aria-mccheddarton'] = {
+			unsupported: true
+		};
+		axe.commons.aria.lookupTable.attributes['aria-bagleypants'] = {
+			unsupported: true
```
WilcoFiers: Should one of these be `false`? That's what your test description implies.

Issue #1252, opened 2018-11-21
Move eslint into a separate CI task


WilcoFiers:
Linting is currently part of the test task. It can run in parallel if we move it to a separate task. It also creates better feedback in Github that way.

PR #1255, opened 2018-11-22
chore: exec lint as separate CI step

jkodu:
Execute `lint` as a separate step in CI.

Closes issue: https://github.com/dequelabs/axe-core/issues/1252

## Reviewer checks

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,6 +28,14 @@ jobs:
           paths:
             - node_modules
 
+  # Run ESLINT
+  lint:
+    <<: *defaults
+    steps:
+      - checkout
+      - <<: *restore_dependency_cache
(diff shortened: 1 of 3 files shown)


jkodu:
@WilcoFiers - we need to enable `lint` check as a required step above.

.circleci/config.yml
```diff
@@ -61,6 +69,10 @@ workflows:
   build:
     jobs:
       - dependencies
+      # Run linting
+      - lint:
+          requires:
```
dylanb: Shouldn't this go into the job definition for lint above?

WilcoFiers: I don't think so. Requires is part of the workflow:
https://circleci.com/docs/2.0/configuration-reference/#requires

PR #1256, opened 2018-11-22
chore: using browserify to import dependencies

jkodu:
This is a developer workflow enhancement PR.

Ability to `require` external modules has been difficult in Axe.
Previous attempts to `require` and inject modules into `axe` global have been not stable, due to having to adapt for various module bundling standards followed by different dependencies.

This PR, circumvents the above problems and introduces a fluent way to `require`, by embracing `browserify`.

- Removed old `generate-imports` grunt multi task, which was flaky.
- Introduce npm script - `npm run imports-gen` which passes dependencies to mount on `axe` as `arguments`. 
- `axe.imports = { ...myDependencies }` is constructed at build time.
- Deleted copy pasted library [css-selecter-parser](https://www.npmjs.com/package/css-selector-parser) and changed into as a `require` dependency. 

Closes issue: NA

## Reviewer checks


build/imports-generator.js
```diff
@@ -0,0 +1,109 @@
...
+
+/**
+ * Convert arguments to object map of require statement
+ * @param {Array<String>} args key value pair of arguments
+ * @returns {Object} Eg: { axios: require('axios'), doT: require('dot') }
+ */
+function parseArgsAsRequireMap(args) {
+	const argMap = {};
+
+	if (args.length % 2) {
```
stephenmathieson: Please add a comment explaining why this validates arguments.

build/imports-generator.js
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * Create file with contents at specified location
+ * @method createFile
+ * @param {String} path file path, inclusive of file name
+ * @param {String} content contents of the file
+ * @returns {Promise}
+ */
+function createFile(path, content) {
+	return new Promise((resolve, reject) => {
+		mkdirp(getDirName(path), err => {
```
stephenmathieson: Consider replacing this with [`make-dir`](https://github.com/sindresorhus/make-dir). `mkdirp` is unmaintained and  does not return a Promise.

stephenmathieson:
```js
const {dirname:getDirName} = require('path')
const makeDir = require('make-dir')
const {writeFile} = require('fs/promises')

const createFile = (path, content) => makeDir(getDirName(path)).then(() => writeFile(path, content))
```

package.json
```diff
@@ -56,6 +56,7 @@
 	},
 	"scripts": {
 		"api-docs": "jsdoc --configure .jsdoc.json",
+		"imports-gen": "node ./build/imports-generator --axios 'axios' --doT 'dot' --CssSelectorParser 'css-selector-parser'",
```
stephenmathieson: I'm concerned that if we keep this mapping as CLI flags, the list may become very difficult to maintain in the future.

Why not just add the mapping to `./build/imports-generator.js`?

```js
const mapping = [ ['axios','axios'], ['doT', 'dot'], ['CssSelectorParser', 'css-selector-parser'] ]
```

build/imports-generator.js
```diff
@@ -0,0 +1,109 @@
...
+		await createFile(importsFile, fileContent);
+	} catch (error) {
+		throw new Error(
+			`Cannot create file axe.imports with require modules`,
+			error
+		);
+	}
+
+	// browserify `lib/core/imports/index.js`
+	browserify(importsFile).bundle(async (err, result) => {
```
stephenmathieson: Are we sure this is going to work in all cases? What about node.js usage?

build/imports-generator.js
```diff
@@ -0,0 +1,109 @@
...
+		}
+	});
+}
+
+// exec
+run();
+
+/**
+ * Convert arguments to object map of require statement
+ * @param {Array<String>} args key value pair of arguments
```
marcysutton: Can you put an example here? "Key value pair" sounds like an object, but it's an array of strings.

build/imports-generator.js
```diff
@@ -0,0 +1,109 @@
...
+ * @returns {Object} Eg: { axios: require('axios'), doT: require('dot') }
+ */
+function parseArgsAsRequireMap(args) {
+	const argMap = {};
+
+	if (args.length % 2) {
+		throw new Error('Invalid number of arguments');
+	}
+
+	for (let i = 0; i < args.length; i += 2) {
```
marcysutton: Comments for this loop would be good too, since it has an atypical increment (`i += 2`) and some magic replacing going on below.

Gruntfile.js
```diff
@@ -340,6 +329,10 @@ module.exports = function(grunt) {
 			}
 		},
 		run: {
+			npm_run_imports: {
```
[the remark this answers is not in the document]
jkodu: I thought about this @stephenmathieson, but this required developer documentation update at various places. 

Our docs are tied to steps like `grunt dev`, `grunt build` and `grunt translate`. So the only way to trigger any current npm script is via `grunt-run`, like above.

You concern of what confusion happened in above #1252 etc., will not re-occur as I have already stripped out multiple references to same task and refactored it.

As you can see from most of the chore PR's, once we more to pure npm scripts, it will be easier to update documentation when we have change all of the grunt tasks. So the answer is NO for the time being.

As if I do this change above now, a developer running `grunt build` will yield different results to `npm run build`.

build/imports-generator.js
```diff
@@ -0,0 +1,109 @@
+/* global Promise */
```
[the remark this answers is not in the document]
jkodu: image

Unfortunately not, our eslint rules aint happy for that.

build/imports-generator.js
jkodu: Some of our tests ensure that, what gets injected into `axe.imports` is as expected. 

I have done some node testing locally. But we will ensure that it works before a release.

package.json
jkodu: On second thought moved it to mapping...

build/imports-generator.js
jkodu: I am having minimal luck with fs/promises for some reason. 
Cheers for the `make-dir` suggestion.

jkodu: Asides `fs/promises` API is have a stability level 1 & experimental flag.

build/imports-generator.js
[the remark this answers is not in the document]
stephenmathieson: Oh strange. PR incoming :)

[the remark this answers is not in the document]
stephenmathieson: #1263

build/imports-generator.js
stephenmathieson: How do you expect we'll "ensure that it works before a release"? Instead of waiting until release-time, can we perform those tests on your branch?

jkodu: Will discuss with @WilcoFiers.

build/imports-generator.js
```diff
@@ -0,0 +1,89 @@
...
+		throw new Error(
+			`Cannot create file axe.imports with require modules`,
+			error
+		);
+	}
+
+	// browserify `lib/core/imports/index.js`
+	browserify(importsFile).bundle(async (err, result) => {
+		if (err) {
+			throw new Error('Cannot browserify axe.imports', err);
```
stephenmathieson: I'm not sure it makes sense to use "browserify" as a verb 🤔

Gruntfile.js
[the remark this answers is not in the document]
marcysutton: That may be true about `grunt-run` and so on, so I can understand not wanting to take all of that on. As an aside, though, I wouldn't let developer docs stop you from making any changes; we can always update the docs. I'm happy to help with that!

dylanb:
Could @marcysutton and/or @stephenmathieson review to make sure their issues have been addressed and then let me know and I'll review?

stephenmathieson:
@dylanb I'm waiting on a response for https://github.com/dequelabs/axe-core/pull/1256#discussion_r236400526

build/imports-generator.js
```diff
@@ -0,0 +1,89 @@
...
+// Global Namespace : Library name
+const moduleMap = {
+	axios: 'axios',
+	doT: 'dot',
+	CssSelectorParser: 'css-selector-parser'
+};
+
+async function run() {
+	// generate `axe.imports` content
+	const fileContent = `
```
WilcoFiers: I'm not in favour of dynamically generating this file. Let's just put this file in source control and have this import-generator copy over into tmp.

Gruntfile.js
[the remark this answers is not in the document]
jkodu: Some of my `chore` work in axe-core is moving away from grunt to npm scripts. I will make the switch to above recommendation in a different PR when time is right :).

jkodu:
@dylanb - thanks for the comment. I believe @WilcoFiers has a handle on this for review and approvals.

WilcoFiers:
Looks fine to me. @jkodu can you run this in the Attest extension and in Attest node to make sure this doesn't accidentally break anything?

jkodu:
@WilcoFiers 

Test both extensions and node suite. Works.

jkodu:
@WilcoFiers 

Test both extensions and node suite. Works.

jkodu:
@WilcoFiers 

Tested both extensions and node suite. Works 👏

build/imports-generator.js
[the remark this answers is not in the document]
stephenmathieson: This directive can now be removed.

build/imports-generator.js
stephenmathieson: Any thoughts here?

build/imports-generator.js
```diff
@@ -0,0 +1,62 @@
...
+run();
+
+/**
+ * Create file with contents at specified location
+ * @method createFile
+ * @param {String} path file path, inclusive of file name
+ * @param {String} content contents of the file
+ * @returns {Promise}
+ */
+function createFile(path, content) {
```
stephenmathieson: Not a blocker, but I'd still like to see this simplified. Maybe:

```js
const makeDir = require('make-dir')
const fs = require('fs')
const {promisify} = require('util')

const writeFile = promisify(fs.writeFile)

const createFile = (path, content) => makeDir(getDirName(path)).then(() => writeFile(path, content))
```

jkodu: Between, this PR and the axe rule generator PR, there is more or less the same createFile fn. Once both are merged. I plan to refactor this into a shared utility fn.

build/imports-generator.js
jkodu: No change here. Leaving as is.

Issue #1174, opened 2018-10-03
False negative: 'image-alt' rule does not fail for alt=" "


iamrafan:
- Demo: ```<img id="test" src="" alt=" ">```

- AT Tests:

  | AT/Browser | alt=" " (space)
-- | -- | --
1 | Edge/Narrator | Image
2 | Edge/JAWS | Graphic
3 | Edge/NVDA | Graphic
4 | Chrome/NVDA | Graphic
5 | Chrome/JAWS | Ignored by AT
6 | IE/JAWS | Ignored by AT
7 | IE/NVDA | Graphic
8 | Safari/voiceover | Ignored by AT
9 | Firefox/NVDA | Graphic

<pre><code>axe-core version: 3.1.1
axe-Coconut version: 3.5.0</code></pre>

WilcoFiers:
This should probably be a separate check so that we can report the issue accurately, and so that its easy to turn off.

PR #1260, opened 2018-11-27
feat(image-alt): require alt text or empty strings

marcysutton:
This change to the image-alt rule accounts for space characters, and fails if they are present. Empty alt attributes are still fine, and text content is still fine. This fixes cases where ATs do not skip decorative images because of the space characters in the alt attribute.

I'm open to name changes for the check, so please review.

Closes https://github.com/dequelabs/axe-core/issues/1174

## Reviewer checks


lib/checks/label/alt-empty-space.js
```diff
@@ -0,0 +1,5 @@
+let nn = node.nodeName.toLowerCase();
+let validSetup =
+	node.hasAttribute('alt') && (nn === 'img' || nn === 'input' || nn === 'area');
```
dylanb: The rule selector will not include `<input>` and `<area>` elements. Why not make this a new, separate rule altogether?

marcysutton:
@dylanb check the issue, there is some support info: #1174

lib/checks/label/alt-empty-space.js
marcysutton: This check doesn't need to include `input` or `area`, so that part can be removed. This only applies to image alt, so in my opinion it should not be a separate rule. `input` and `area` both use the existing check `non-empty-alt`.

WilcoFiers: I agree with Marcy on this one. I think this one is better as a check that's part of image-alt, provided that we split off the `[role=img]`, which we have a separate support ticket for.

WilcoFiers:
Waiting for @marcysutton to update. Putting this into WIP.

marcysutton:
PR updated.

lib/checks/label/alt-space-value.json
```diff
@@ -0,0 +1,11 @@
+{
+	"id": "alt-space-value",
+	"evaluate": "alt-space-value.js",
+	"metadata": {
+		"impact": "critical",
+		"messages": {
+			"pass": "Element has a valid alt attribute value",
+			"fail": "Element has an alt attribute containing only a space character"
```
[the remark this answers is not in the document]
marcysutton: That's fine by me, I've updated the PR.

lib/checks/label/alt-empty-space.js
```diff
@@ -0,0 +1,5 @@
...
+let validAttrValue = /^\s+$/.test(node.getAttribute('alt'));
+return validSetup && validAttrValue;
```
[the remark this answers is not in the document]
marcysutton: Fixed.

marcysutton:
I've updated this PR, I don't understand what's happening with the rule-descriptions though. Git is acting really weird with that file, I tried adding/committing it and it keeps popping up again.

PR #1268, opened 2018-12-04
chore: enable geckodriver tests on CI

jkodu:
This PR does the below:
- Installs `geckodriver` on docker base image `node-10-browsers-legacy`.
- Installs latest firefox.

This **allows** us to run `firefox` tests, which used to be skipped in the webdriver tests.

Previously:
image

Now (runs tests on firefox):
image

Closes issue: NA

## Reviewer checks


.circleci/config.yml
```diff
@@ -42,6 +42,20 @@ jobs:
     steps:
       - checkout
       - <<: *restore_dependency_cache
+      - run:
```
stephenmathieson: It would be nice to cache this stuff so we're not making our builds even slower. This isn't a blocker (IMO), but something to think about instead.

.circleci/config.yml
```diff
@@ -42,6 +42,20 @@ jobs:
...
+          name: Install geckodriver
+          command: |
```
stephenmathieson: Is it worth putting all this stuff in a `./.circleci/install-geckodriver.sh` file?

stephenmathieson:
@WilcoFiers @jkodu something we may want to investigate is using the `node:10-browsers` image rather than `node:10-browsers-legacy`. The "browsers" image already has Firefox in it, but it's missing PhantomJS. To compensate for this, we could install PhantomJS from `npm` (https://www.npmjs.com/package/phantomjs-prebuilt) and add it to `$PATH`.

This would mean we no longer have to download Firefox and the PhantomJS stuff is cached, so we'd avoid the additional install time on CI 😄

marcysutton:
And/or we could get rid of Phantom....

stephenmathieson:
> And/or we could get rid of Phantom....

This would be great, but unfortunately, we're extremely tightly coupled with [`grunt-mocha`](http://npmjs.org/grunt-mocha), which relies on PhantomJS. It's a much larger project to drop Phantom than it seems, whereas `npm install --save-dev phantomjs-prebuilt` and `export PATH=./node_modules/.bin:$PATH` are pretty simple wins.

jkodu:
@WilcoFiers 

I have moved the install scripts as shell scripts (but unsure how to cache this).

I think the times these steps are taking are not a concern at the moment.
image

I did not change the docker image to `node-10:browsers` as against `node-10:browsers-legacy`, or attempt to move away from phantomjs, as the suggestions above are better of, tackled in a different PR.

jkodu:
@marcysutton 
I did explore moving away from `grunt-mocha`, this will rid us of PhantomJS dependency. Again not for this PR.

stephenmathieson:
@jkodu I disagree that the docker image change should be discussed elsewhere. Just swapping an image and adding a dev dep seems like a much less complicated (and more robust) way of accomplishing what you're trying to do here.

jkodu:
Moved to docker image `node-XXXXX-browsers` instead of `**-legacy`.

Install and cache `phantomjs` instead, which takes between 00:00 and 00:02 tops.

Caching for the first time:
image

Restoring cache:
image

And now parallel tests execute as follows:
image

PR #1282, opened 2018-12-17
refactor: createFile method moved as shared build utility

jkodu:
Whilst developing `rule generator cli` and `imports generator using browserify`, `createFile` function was duplicated in two PR's. Here it is cleaned up.

Closes issue: NA

## Reviewer checks


build/rule-generator/directories.js
```diff
@@ -0,0 +1,22 @@
...
+		__dirname,
+		'..',
+		'..',
+		'test',
+		'integration',
+		'rules'
+	)
+};
+
+module.exports = {
```
stephenmathieson: What else do you expect to have in this module? If all you're going to ever export is an object, it seems unnecessary to nest it within another.

Also, `const directories = require('./directories).directories` is pretty redundant.

I suggest changing this to `module.exports = directories` or exporting directories individually (`exports.testChecksIntegration = path.join(....)`).

build/shared/create-file.js
```diff
@@ -0,0 +1,19 @@
...
+ * Create file with given contents at specified location
+ * @method createFile
+ * @param {String} path file path, inclusive of file name
+ * @param {String} content contents of the file
+ * @returns {Promise}
+ */
+const createFile = (path, content) =>
+	makeDir(getDirName(path)).then(() => writeFile(path, content));
+
+module.exports = {
```
stephenmathieson: Just export `createFile` directly.

build/rule-generator.js
```diff
@@ -45,8 +46,14 @@ async function run() {
 	}
 
 	try {
-		const result = await Promise.all(files.map(createFile));
-
+		const result = await Promise.all(
+			files.map(async meta => {
```
stephenmathieson: Not a big deal, but this could be a 1-liner:

```js
files.map(meta => createFile(`${meta.dir}/${meta.name}`, meta.content))
```

stephenmathieson: Ah, makes sense. 👍

PR #1283, opened 2018-12-18
docs: Editorial changes for readme

WilcoFiers:
## Reviewer checks


README.md
```diff
@@ -10,34 +10,30 @@
 [![Join the chat at https://gitter.im/dequelabs/axe-core](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/dequelabs/axe-core?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 [![Package Quality](http://npm.packagequality.com/shield/axe-core.svg)](http://packagequality.com/#?package=axe-core)
 
-The Accessibility Engine for automated testing of HTML-based user interfaces. Drop the axe on your accessibility defects!
+![Axe Logo](doc/axelogo2018.png)
 
-Thanks for helping us to make axe-core the most widely used HTML accessibility testing library in the galaxy! Share your axe-core story using #axecoresquad or by submitting it [here](https://accessibility.deque.com/axe-core-stories) to get a sweet T-shirt.
+Axe is an accessibility testing engine for websites and other HTML-based user interfaces. It's fast, secure, lightweight, and was built to seamlessly integrate with any existing test environment so you can automate accessibility testing alongside your regular functional testing.
```
marcysutton: "HTML-based interfaces" will be outdated soon, should we make a note to update it later? cc @caitlinthefirst

README.md
```diff
@@ -66,7 +62,7 @@ axe.run(function (err, results) {
 
 ## Supported Browsers
 
-The [axe API](doc/API.md) fully supports the following browsers:
+The [aXe API](doc/API.md) fully supports the following browsers:
```
marcysutton: I thought we were getting rid of the capital X? cc @caitlinthefirst

README.md
caitlinthefirst: Yeah, I think that would be a good idea. We'll need to chat about what to put there instead.

README.md
caitlinthefirst: We are. I clearly missed some of them.

marcysutton: @caitlinthefirst whatever you prefer! You're a member of the project, so if you'e comfortable you should be able to check out the `marketing-readme` branch and add a new commit. Otherwise I'm happy to make a change for you!

WilcoFiers: I've fixed them.

README.md
WilcoFiers: Axe-core will always be "HTML-based", even if we include linting, it'll never be for languages that do not target HTML.

PR #1284, opened 2018-12-19
fix: Allow div groups for dlitem rule

### LINKED ISSUE #1265: False Positive: Ensures DT and DD elements are contained by a DL
Expectation: 
According to WHATWG wrapping child elements of a DL should be allowed without errors. 
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl#Wrapping_name-value_groups_in_%3Cdiv%3E_elements

Actual: 
When using a DL list and all of the necessary child elements axe chrome extension is throwing an error if you wrap the DT and DD with a DIV for styling purposes.  The error that it reports "Ensures DT and DD  elements are contained by a DL Impact: serious"

Motivation: 
The W3C validator and WHATWG allow for wrapping child elements of a DL with a DIV acceptable.

Code Tested: 
`<dl><div><dt>Name</dt><dd>Godzilla</dd></div><div><dt>Born</dt>  <dd>1952</dd></div><div><dt>Birthplace</dt><dd>Japan</dd></div><div>  <dt>Color</dt>  <dd>Green</dd></div></dl>`

<pre><code>axe-devtools 3.6.0
axe-core 3.1.1

For Tooling issues:
axe chrome extension
</code></pre>

WilcoFiers:
Allow grouping in divs for the dlitem rule.

Closes #1265

## Reviewer checks


lib/checks/lists/dlitem.js
```diff
@@ -1,17 +1,22 @@
...
+	parentRole = axe.commons.aria.getRole(parent, { noImplicit: true });
+}
 
 // Unlike with UL|OL+LI, DT|DD must be in a DL
 if (parentTagName !== 'DL') {
 	return false;
 }
 
-const parentRole = (parent.getAttribute('role') || '').toLowerCase();
-if (!parentRole || !axe.commons.aria.isValidRole(parentRole)) {
```
WilcoFiers: isValidRole is done by `aria.getRole()`, in case you're wondering.

marcysutton: I'm surprised it works with the `noImplicit: true`, can you explain that bit? This applies to native elements as well as bolt-on roles, right? The tests seem to pass, so 🤷‍♀️

WilcoFiers: The requirements here are: It needs to be in a `dl`, and if that `dl` has a role attribute, it can only be a `list`. So what's happening is first I check the parentTagName, and then I check if the explicit role (which is what we get from `noImplicit: true` is `null` or `'list'`.

test/checks/lists/dlitem.js
```diff
@@ -10,60 +10,116 @@ describe('dlitem', function() {
...
+					nodeName +
+					' id="target">My list item</' +
+					nodeName +
+					'></div></dl>'
+			);
+			assert.isTrue(checks.dlitem.evaluate.apply(null, checkArgs));
+		});
+	});
+
+	it('returns false if the dd/dt is in a div with a role with a dl as grandparent', function() {
```
marcysutton: I'd add `with a list role` to the test name because that's the important bit

test/checks/lists/dlitem.js
```diff
@@ -88,4 +144,17 @@ describe('dlitem', function() {
 			assert.isFalse(checks.dlitem.evaluate.apply(null, checkArgs));
 		}
 	);
+
+	(shadowSupport.v1 ? it : xit)(
+		'should return truewhen the item is grouped in dl > div',
```
marcysutton: Add a space to the test name: `true when`. It should also mention Shadow DOM

PR #1298, opened 2019-01-03
fix: respect preload set to false

jkodu:
This PR does the below:
- Noticed a bug with `shouldPreload`, when `preload: false` is passed in via options, `shouldPreload` returned true (handled by checking for undefined/ null).

- Say, in cases where `runOnly` is configured to any `cssom` related rule like `css-orientation-lock`, but `preload` is set to `false`, no proper `incomplete` message was provided.

Eg: 
```js
axe.run({ preload: false, runOnly: {type: 'rule', values: ['css-orientation-lock'] }})
```

The above snippet will return no `assets` such as cssom, as `preload: false` overrides the need to load assets. In this case, there was no helpful `incomplete` message.

Closes issue: NA.

## Reviewer checks


jkodu:
Added more test cases for `preload`, as for the `css-orientation-lock` there is already a test case for the incomplete scenario - https://github.com/dequelabs/axe-core/blob/develop/test/integration/full/css-orientation-lock/incomplete.js#L12

PR #1302, opened 2019-01-04
fix: speed up getNodeFromTree

paulirish:
`axe.utils.getNodeFromTree` currently doesn't return early if it has found a result. It continues iterating through the rest of the children. But the perf win is pretty huge: ~600ms on the page I'm testing.

To test I ran axe against [this URL](https://www.guidetoonlineschools.com/online-schools) like this:

```js
await window.axe.run(document, {performanceTimer:true, elementRef: true,runOnly: {  type: 'tag',  values: ['wcag2a', 'wcag2aa']},resultTypes: ['violations', 'inapplicable'],});
```

And then I looked at the `Measure runchecks_color-contrast took XXXX ms` line. (As that rule uses this method quite a lot).

### numbers:

| `develop` | this branch |
|-|-|
| 3126 | 2351 |
| 3004 | 2399 |
| 2846 | 2360 |

image

On average, a 620ms improvement in this one rule. 

## Reviewer checks


lib/core/utils/flattened-tree.js
```diff
@@ -142,15 +142,12 @@ axe.utils.getNodeFromTree = function(vNode, node) {
 		return vNode;
 	}
 	vNode.children.forEach(candidate => {
-		var retVal;
-
-		if (candidate.actualNode === node) {
+		if (found) {
+			return;
+		} else if (candidate.actualNode === node) {
```
stephenmathieson: nit: can remove this `else` since we return from the condition above.

WilcoFiers:
Fantastic! Thanks for the contribution @paulirish!

PR #1303, opened 2019-01-04
fix(perf): update end mark for rules

paulirish:
With performanceTimer on, we get some nice usertiming measures added. But in #701, some _jerk_ picked an endpoint for rules that I, personally, think can be improved. :)

This moves the end mark of each rule to when the checks have synchronously finished. This change means we don't include the asynchronous bit afterwards, but that async bit is the subject of #1172.

For accurate timing, it makes more sense to keep these measures synchronous. It also makes reading the flame chart more logical.

Screenshots of before and after:
image

## Reviewer checks


paulirish:
And FWIW this PR still makes sense, because of
https://github.com/dequelabs/axe-core/blob/80c1c740fbac42abcf23b39ed3b3832fbd80180f/lib/core/base/rule.js#L212

This setTimeout pushes the end mark to after the end of the big axe block of execution. Basically making it look like all these Rules take 1-10 seconds longer than they really do.

stephenmathieson:
What's the status of this PR? Is there anything that needs to happen before this gets reviewed?

/cc @dequelabs/axe-team

PR #1316, opened 2019-01-16
feat(rule): Require unique aria labels in checkboxgroup & radiogroup

### LINKED ISSUE #1111: False negative: 'group-labelledby' check in 'checkboxgroup' rule does not fail corner cases
Screen reader experience for the following corner cases is not as expected, hence the ```checkboxgroup``` rule should fail

- Case-1: Ungrouped checkboxes with same name and each having aria-labelledby pointing to a common label

- Case-2: Ungrouped checkboxes with explicit [label] and each having aria-labelledby pointing one of the checkbox labels

- Case-3: Ungrouped checkboxes with implicit [label] and each having aria-labelledby pointing one of the checkbox labels

**Link to live demo:** https://codepen.io/iamrafan/pen/MqjwOz

**AT test results:**
image

WilcoFiers:
Update the group-labelledby check to enforce that in addition to having a shared label, radiobuttons and checkboxes with the same name also have a unique label. This impacts the checkboxgroup and radiogroup rules.

I merged in a fix from #1314, which changes the gruntfile. We should probably more that PR first.

Closes #1111

## Reviewer checks

### DIFF
--- a/lib/checks/forms/group-labelledby.js
+++ b/lib/checks/forms/group-labelledby.js
@@ -1,34 +1,61 @@
-this.data({
-	name: node.getAttribute('name'),
-	type: node.getAttribute('type')
-});
+const { dom, text, utils } = axe.commons;
 
-var doc = axe.commons.dom.getRootNode(node);
-var matchingNodes = doc.querySelectorAll(
-	'input[type="' +
-		axe.commons.utils.escapeSelector(node.type) +
-		'"][name="' +
-		axe.commons.utils.escapeSelector(node.name) +
-		'"]'
+const type = utils.escapeSelector(node.type);
+const name = utils.escapeSelector(node.name);
+const doc = dom.getRootNode(node);
+const data = {
+	name: node.name,
+	type: node.type
+};
+
+const matchingNodes = Array.from(
+	doc.querySelectorAll(`input[type="${type}"][name="${name}"]`)
 );
+// There is only one node with this name & type, so there's no need for a group
 if (matchingNodes.length <= 1) {
+	this.data(data);
 	return true;
 }
 
-// Check to see if there's an aria-labelledby value that all nodes have in common
-return (
-	[].map
-		.call(matchingNodes, function(m) {
-			var l = m.getAttribute('aria-labelledby');
-			return l ? l.split(/\s+/) : [];
-		})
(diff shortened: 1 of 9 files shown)


lib/checks/forms/group-labelledby.js
```diff
@@ -1,34 +1,61 @@
...
+
+if (uniqueLabels.length > 0 && sharedLabels.length > 0) {
+	this.data(data);
+	return true;
+}
+
+if (uniqueLabels.length > 0 && sharedLabels.length === 0) {
+	data.failureCode = 'no-shared-label';
+} else if (uniqueLabels.length === 0 && sharedLabels.length > 0) {
+	data.failureCode = 'no-unique-label';
```
dylanb: Why not always fail this even when there is no shared label?

dylanb:
@WilcoFiers seems like the change to make @stephenmathieson a code owner broke our ability to approve changes in that code

jkodu:
@dylanb PR - https://github.com/dequelabs/axe-core/pull/1331, fixed the codeowners issue.

naltatis:
Feedback: Just updated axe an ran into this check. We finally could find the issue, we were having, but had to read through the tests and code modifications here to understand what was wrong.

This is the message we recieved
```
Ensures related <input type="radio"> elements have a group and that the group designation is consistent https://dequeuniversity.com/rules/axe/3.2/radiogroup?application=axeAPI

Fix any of the following:
  Elements with the name "segments-deliveryaddress-deliveryButtons" do not all have a unique label, referenced through aria-labelledby
  Element does not have a containing fieldset or ARIA group
  #billing-radiobutton
```

The actual problem was, that our radio buttons had a `aria-labelledby` that was pointing to the groups legend. Additionally the radios had an individual label (connected via for/id and label > input nesting), which we thought was sufficient. We were able to get rid of the error by expanding the aria annotiation by the radio inputs label-id `aria-labelledby="legend-id"` -> `aria-labelledby="legend-id label1-id"`.
The error message above and the linked documentation did not point to that.

Maybe you should consider improving the message.

dylanb:
@naltatis Do you think you could create a new issue in this project that references this and also maybe suggest a text that might have helped you figure out what was going on?

PR #1318, opened 2019-01-16
fix: autocomplete-appropriate node type resolution

jkodu:
`Firefox` returns `node.type` as `text` for certain scenarios like - `<input autocomplete="bday-month" type="month">`, hence using `getAttribute` to resolve the value of `type` solves for this edge case.

Due to this, the number of violations returned across different browsers were different as described in the bug below.

Closes issue:
- https://dequesrc.atlassian.net/browse/WWD-1957

## Reviewer checks

### DIFF
--- a/lib/checks/forms/autocomplete-appropriate.js
+++ b/lib/checks/forms/autocomplete-appropriate.js
@@ -48,12 +48,21 @@ const allowedTypes = allowedTypesMap[purposeTerm];
 
 /**
  * Note:
+ * Inconsistent response for `node.type` across browsers, hence resolving and sanitizing using getAttribute
+ * Example:
  * Firefox returns `node.type` as `text` for `type='month'`
- * Hence using `node.getAttribute('type')` over `node.type`
+ *
+ * Reference HTML Spec - https://html.spec.whatwg.org/multipage/input.html#the-input-element to filter allowed values for `type`
+ * and sanitize (https://html.spec.whatwg.org/multipage/input.html#value-sanitization-algorithm)
  */
-const nodeType = node.hasAttribute('type') ? node.getAttribute('type') : 'text';
+let type = node.hasAttribute('type')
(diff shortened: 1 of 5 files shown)


lib/checks/forms/autocomplete-appropriate.js
```diff
@@ -45,8 +45,15 @@ if (axe.commons.text.autocomplete.stateTerms.includes(purposeTerm)) {
 }
 
 const allowedTypes = allowedTypesMap[purposeTerm];
+
+/**
+ * Note:
+ * Firefox returns `node.type` as `text` for `type='month'`
+ * Hence using `node.getAttribute('type')` over `node.type`
+ */
+const nodeType = node.hasAttribute('type') ? node.getAttribute('type') : 'text';
```
WilcoFiers: I agree that we need to use the attribute here rather than the DOM property, but you'll have to normalise it. I don't know exactly what that involves, you can look that up in the spec. It's probably case insensitive. It also needs to default to "text" when there's an invalid value here, not just when type isn't set. Otherwise an element with `type="foobar"` wouldn't be allowed to have autocomplete allowed for text.

lib/commons/utils/valid-input-type.js
```diff
@@ -0,0 +1,38 @@
+/* global axe */
+
+/**
+ * Returns array of valid input type values
+ * @method validInputTypes
+ * @memberof axe.commons.utils
+ * @instance
+ * @return {Array<Sting>}
+ */
+axe.utils.validInputTypes = function validInputTypes() {
```
WilcoFiers: This should be assigned to axe.commons.utils, not to axe.utils. Also, I don't think this should be a function. It's inconsistent with how other thing in axe behave.

lib/commons/utils/valid-input-type.js
```diff
@@ -0,0 +1,38 @@
+/* global axe */
```
WilcoFiers: If you create a new commons, please provide tests.

PR #1321, opened 2019-01-16
feat: add a reference to the `node` a rule failed on

### LINKED ISSUE #1317: Expose the element that is tested when a rule throws
When an error occurs while running a rule, axe-core catches that error and puts it in incomplete. What it doesn't do is expose which element it was testing when that error happened. Exposing this will help with debugging.

stephenmathieson:
This patch adds a `.node` reference to an errored rule result. This can be used to determine which node the rule errored on and will be helpful for debugging purposes.

Closes #1317.

## Reviewer checks


stephenmathieson:
Seems this isn't quite as simple as I had hoped. Since the results object is serialized (`JSON.stringify()`), we can't just attach the `node` here (due to circular references). Perhaps a **unique** selector for the element would work instead?

Thoughts anyone?

straker:
Would getting the `outerHTML` be sufficient for debugging in these cases?

stephenmathieson:
I don't think `outerHTML` would be enough by itself, since it won't always point to a single node. For example, if axe-core errored on one of the `<span>`s in the below chunk of HTML, we wouldn't know which:

```html
<div>
  <span>hello</span>
  <span>hello</span>
  <span>hello</span>
</div>
```

This is why I was thinking a (unique) selector would work, since we'd get back something like: `div > span:nth-child(2)` which paints a clearer picture.

This said, perhaps an object which resembles the following would work?

```json
{
  "html": "<span>hello</span>",
  "selector": "div > span:nth-child(2)"
}
```

straker:
I guess since axe reports issues using a unique selector then it would work. Though not being able to click on the inspect node button right above it might still be problematic for DOM that repeats structure. But I can't think of a better alternative without having the actual node to reference.

jkodu:
@stephenmathieson, is it a `DqElement` at this point? If yes, there is a getter named `selector` which can return `a unique CSS selector for the element`

See - https://github.com/dequelabs/axe-core/blob/develop/lib/core/utils/dq-element.js#L56

stephenmathieson:
Looks like we can use the `toJSON` method which should provide all necessary information.

stephenmathieson:
Turns out, this isn't a `DqElement`. It's something else. I'm trying to figure out what's going on here.

Didn't think this was going to be nearly as complicated as seems to be 🤷‍♀️

jkodu:
Updated. Tests are fixed.

lib/core/base/check.js
```diff
@@ -58,6 +58,7 @@ Check.prototype.enabled = true;
  * @param  {Function} callback Function to fire when check is complete
  */
 Check.prototype.run = function(node, options, context, resolve, reject) {
+	/* eslint max-statements:0 */
```
WilcoFiers: This seems unnecessary.

stephenmathieson: The "max statements" rule seems unnecessary. We always increase the minimum rather than refactoring our code. IMO it should be disabled.

Issue #1324, opened 2019-01-17
unsupportedrole check fails all invalid roles


WilcoFiers:
The following code shouldn't fail the unsupportedrole check:

```html
<input role="foo">
<input role="input">
```

The problem this causes is that these code snippets are reported as:

> Fix all of the following
> - Abstract role cannot be directly used
> - The role used is not widely supported in assistive technologies

PR #1328, opened 2019-01-17
fix: don't flag invalid roles in unsupportedrole

straker:
The check for unsupportedrole also checked if the role was valid, causing invalid roles to be flagged as unsupported.

Closes issue: #1324

## Reviewer checks

### DIFF
--- a/lib/checks/aria/unsupportedrole.js
+++ b/lib/checks/aria/unsupportedrole.js
@@ -1,1 +1,1 @@
-return axe.commons.aria.isUnsupportedRole(node.getAttribute('role'));
+return axe.commons.aria.isUnsupportedRole(axe.commons.aria.getRole(node));
--- a/lib/checks/aria/unsupportedrole.js
+++ b/lib/checks/aria/unsupportedrole.js
@@ -1,3 +1,1 @@
-return !axe.commons.aria.isValidRole(node.getAttribute('role'), {
-	flagUnsupported: true
-});
+return axe.commons.aria.isUnsupportedRole(node.getAttribute('role'));
--- a/lib/commons/aria/roles.js
+++ b/lib/commons/aria/roles.js
@@ -22,6 +22,19 @@ aria.isValidRole = function(
 };
 
 /**
+ * Check if a given role is unsupported
+ * @method isUnsupportedRole
+ * @memberof axe.commons.aria
+ * @instance
+ * @param {String} role The role to check
+ * @return {Boolean}
+ */
+aria.isUnsupportedRole = function(role) {
+	const roleDefinition = aria.lookupTable.role[role];
+	return roleDefinition ? roleDefinition.unsupported : false;
+};
+
+/**
  * Get the roles that get name from the element's contents
  * @method getRolesWithNameFromContents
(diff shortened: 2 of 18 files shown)


test/checks/aria/unsupportedrole.js
```diff
@@ -12,7 +12,7 @@ describe('unsupportedrole', function() {
 			type: 'widget',
 			unsupported: true
 		};
-		fixture.innerHTML = '<div id="target" role="mccheddarton">Contents</div>';
+		fixture.innerHTML = '<div id="target" role="mcheddarton">Contents</div>';
```
stephenmathieson: Not a big deal at all, but what's a "mcheddarton"?

straker: I believe it's the name of Marcy's dog.

lib/checks/aria/unsupportedrole.js
```diff
@@ -1,3 +1 @@
-return !axe.commons.aria.isValidRole(node.getAttribute('role'), {
-	flagUnsupported: true
-});
+return axe.commons.aria.isUnsupportedRole(node.getAttribute('role'));
```
jkodu: I know the `aria.isValidRole` does not `sanitize` the `role`. It may be worth doing `trim` or `axe.commons.text.sanitize`, as well as `toLoweCase`, so it matches the values in the `lookupTable`

lib/commons/aria/roles.js
```diff
@@ -21,6 +21,19 @@ aria.isValidRole = function(
...
 
+/**
+ * Check if a given role is unsupported
+ * @method isUnsupportedRole
+ * @memberof axe.commons.aria
+ * @instance
+ * @param {String} role The role to check
+ * @return {Boolean}
+ */
+aria.isUnsupportedRole = function(role) {
```
WilcoFiers: I don't think we need a new method for this. I don't expect we'll have another use for it. Also, if we do add a method to this, it should go into its own file and have unit tests for that. Makes the repo easier to navigate that way.

lib/checks/aria/unsupportedrole.js
WilcoFiers: That's not the bug this is trying to fix, but fair point. Better to use `aria.getRole(node, { noImplicit: true })`, that does everything you need to do to get the role explicit role of an element. It also handles case sensitivity better, and will eventually be able to deal with fallback roles too.

test/checks/aria/unsupportedrole.js
stephenmathieson: Haha, ok

lib/commons/aria/roles.js
straker: My reasoning was that the fix to pass on invalid roles when checking for unsupported roles inside `isValidRole` created two separate code paths in the function. At that point I figured they should be 2 separate functions to keep them doing one thing each.

lib/checks/aria/unsupportedrole.js
straker: Why wouldn't we want the elements implicit role as well?

PR #1330, opened 2019-01-22
fix: deprecate axe.commons.utils namespace

jkodu:
This PR does the below:
- Migrated utility functions from `axe.commons.utils` namespace to `axe.utils`

Why?
- There were a plethora of namespace pollution - eg: `var utils = (commons.utils = axe.utils);`

Closes issue: NA

## Reviewer checks


lib/commons/utils/index.js
```diff
@@ -1,10 +0,0 @@
-/* exported utils */
-/* global axe */
-/*eslint no-unused-vars: 0*/
-/**
- * Namespace for general utilities.
- * @namespace utils
- * @memberof axe.commons
- */
-
-var utils = (commons.utils = axe.utils);
```
WilcoFiers: We need to keep this around. Removing this would be a breaking change.

test/commons/utils/index.js
```diff
@@ -1,21 +0,0 @@
...
-		assert.isFunction(axe.commons.utils.matchesSelector);
-	});
-});
-describe('utils.clone', function() {
-	'use strict';
-
-	it('should be a function', function() {
-		assert.isFunction(axe.commons.utils.clone);
-	});
-});
```
WilcoFiers: Same here, we don't want to make a breaking change. Just check that `assert.equal(axe.utils, axe.commons.utils)`.

lib/commons/utils/index.js
stephenmathieson: What would this break?

stephenmathieson:
@WilcoFiers I don't understand why removing the aliased namespace would be a breaking change. From what I can tell, it's not part of any public API. Can you please clarify?

lib/commons/utils/index.js
WilcoFiers: Commons are pseudo-public. They are used in custom rules which clients use. We have to be careful about changing commons.

PR #1335, opened 2019-01-28
feat(rule): Label and Name from Content mismatch WCAG21 (Issue #1149)

jkodu:
This PR does the below:
- Implements a new rule -> label and name from content mismatch.
- This PR has been branched from PR #1163 , to utilise the `a11y-name` computation, so reviews required only on commit - https://github.com/dequelabs/axe-core/pull/1335/commits/ef1c638731c9fec36ac3dcca9fb025d7f4b425b9, also ideally this should be merged after the earlier said PR is merged.

Closes issue:
- https://github.com/dequelabs/axe-core/issues/1149

## Reviewer checks


test/rule-matches/label-content-name-mismatch-matches.js
```diff
@@ -0,0 +1,170 @@
...
+	it('returns false for non-widget role that does support name from content', function() {
+		var target = fixtureSetup(
+			'<div role="doc-chapter" aria-label="OK">Next</div>',
+			'div'
+		);
+		var actual = rule.matches(target);
+		assert.isFalse(actual);
+	});
+
+	it('returns true for non-widget role that does support name from content', function() {
```
WilcoFiers: Doesn't the rule say that it should only test widgets? Why are we testing this node?

test/integration/rules/label-content-name-mismatch/label-content-name-mismatch.html
```diff
@@ -0,0 +1,20 @@
...
+<button id="pass7" name="link" aria-label="Next Page in the list">Next Page</button>
+
+<!-- Fail -->
+<div id="fail1" role="link" aria-label="OK">Next</div>
+<button id="fail2" name="link" aria-label="the full">The full label</button>
+<div id="fail3" role="link" aria-label="OK">Next</div>
+<button id="fail4" name="link" aria-label="the full">The full label</button>
+
+<div role="button" id="fail5" aria-labelledby="foo">some content</div>
+<div id="foo">123</div>
```
WilcoFiers: We should add the inapplicable test cases from Auto-WCAG in here.

lib/checks/label/label-content-name-mismatch.js
```diff
@@ -0,0 +1,25 @@
...
+ * if there is no `accessibleText` or `textContent`
+ */
+const accText = accessibleText(node).toLowerCase();
+const visibleTextContent = visible(node).toLowerCase();
+
+/**
+ * if `textContent` is not part of `accessibleText` -> fail
+ */
+
+if (!accText.includes(visibleTextContent)) {
```
WilcoFiers: How about `<button aria-label="next page">next page &gt;</button>`, or things similar. I think we need to come up with a way to handle things like punctuations and other things that people wouldn't "read out". Another scenario is `<button aria-label="page 1">page one</button>` or the inverse of it.

lib/checks/label/label-content-name-mismatch.json
```diff
@@ -0,0 +1,11 @@
+{
+	"id": "label-content-name-mismatch",
+	"evaluate": "label-content-name-mismatch.js",
+	"metadata": {
+		"impact": "serious",
+		"messages": {
+			"pass": "Element contains visible text as part of it's accessible name",
+			"fail": "The visible text of the element is different from it's accessible name"
```
WilcoFiers: How about "text inside the element is not included in the accessible name"?

lib/rules/label-content-name-mismatch-matches.js
```diff
@@ -0,0 +1,62 @@
...
+		arialabelText,
+		arialabelledbyText,
+		getRole,
+		getRolesWithNameFromContents
+	},
+	text: {
+		accessibleText,
+		visible: visibleText // -> get visible text
+	}
+} = axe.commons;
```
WilcoFiers: Why did you do this? I don't think this is good for readability.

lib/rules/label-content-name-mismatch-matches.js
```diff
@@ -0,0 +1,62 @@
+/**
+ * Applicability:
+ * Rule applies to any element that has
+ * a) a semantic role that is `widget` that supports name from content
```
WilcoFiers: You're not checking if the semantic role is widget. When you do, please note that the "type" property in the lookup table is not the same as it is in the ARIA spec. We'll need to figure something out about that. Maybe blacklist the ones that have type: widget that aren't widgets according to ARIA? I'm not sure if those things can be changed without breaking a bunch of stuff unfortunately.

lib/checks/label/label-content-name-mismatch.js
```diff
@@ -0,0 +1,25 @@
...
+	return false;
```
WilcoFiers: We need to do something about `<button aria-label="close">x</button>`. It might be sufficient to put elements with text content of a single char into needs review?

lib/rules/label-content-name-mismatch-matches.js
```diff
@@ -0,0 +1,62 @@
...
+
+const visibleTextContent = visibleText(node);
+/**
+ * if no `contentText` or contains `unicode` - ignore `node`
+ */
+if (!visibleTextContent || /[^\u0000-\u00ff]/.test(visibleTextContent)) {
+	return false;
+}
+
+const accText = accessibleText(node);
```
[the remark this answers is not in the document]
jkodu: This seems to be not true.
For example, for below snippet:
`<button id="target" name="link" aria-label="">Until the very end.</button>`

`isAriaLabel` - returns `""`
`accessibleText` - returns `"Until the very end."`

lib/commons/text/unicode.js
```diff
@@ -0,0 +1,111 @@
...
+			'\u2600-\u26FF' + // Misc Symbols
+			'\u2700-\u27BF' + // Dingbats
+				']'
+		),
+		/**
+		 * Regex for Emoji matching
+		 * es6(esque) regex using '/u'
+		 *
+		 * - https://github.com/tc39/proposal-regexp-unicode-sequence-properties#matching-all-emoji-including-emoji-sequences
+		 * - https://github.com/mathiasbynens/emoji-regex
```
WilcoFiers: Why didn't you include this as an import?

test/commons/text/is-human-interpretable.js
```diff
@@ -0,0 +1,55 @@
+describe('text.isHumanInterpretable', function() {
+	it('returns 0 when no given string is empty', function() {
```
WilcoFiers: "when no given string" typo

test/commons/text/is-human-interpretable.js
```diff
@@ -0,0 +1,55 @@
...
+		var actual = axe.commons.text.isHumanInterpretable('');
+		assert.equal(actual, 0);
+	});
+
+	it('returns 0 when given a single character that is blacklisted as icon', function() {
```
WilcoFiers: Either put the char in the test name, or do all of these assertions in the same test, but don't use the same test name for different tests.

test/commons/text/is-human-interpretable.js
```diff
@@ -0,0 +1,55 @@
...
+	});
+
+	it('returns 0 when given a single character that is a punctuation is used as icon', function() {
+		var actual = axe.commons.text.isHumanInterpretable('?');
+		assert.equal(actual, 0);
+	});
+
+	it('returns 0 when given a string has punctuations', function() {
+		var actual = axe.commons.text.isHumanInterpretable('I like 🏀');
+		assert.equal(actual, 0);
```
WilcoFiers: This shouldn't return 0. Clearly this is text. Also, an emoji isn't a punctuation.

test/commons/text/is-human-interpretable.js
```diff
@@ -0,0 +1,55 @@
...
+	});
+
+	it('returns 0 when given a string has emoji', function() {
+		var actual = axe.commons.text.isHumanInterpretable('I like 🏀');
+		assert.equal(actual, 0);
```
WilcoFiers: As above.

test/commons/text/is-human-interpretable.js
```diff
@@ -0,0 +1,55 @@
...
+	});
+
+	it('returns 0 when given a string has non BMP character (eg: windings font)', function() {
+		var actual = axe.commons.text.isHumanInterpretable('I ✂ my hair');
+		assert.equal(actual, 0);
```
WilcoFiers: As above, this is text.

lib/commons/text/is-human-interpretable.js
```diff
@@ -0,0 +1,57 @@
...
+	 * 					eg: '💪'
+	 * 					eg: I saw a shooting 💫
+	 */
+
+	if (!str.length) {
+		return 0;
+	}
+
+	// Step 1
+	if (str.length === 1) {
```
WilcoFiers: You don't need this if statement. It does nothing.

test/commons/text/accessible-text.js
```diff
@@ -113,7 +113,7 @@ describe('text.accessibleTextVirtual', function() {
 
 		var target = axe.utils.querySelectorAll(axe._tree, '#t1')[0];
 		assert.equal(
-			axe.commons.text.accessibleTextVirtual(target),
+			axe.commons.text.accessibleTextVirtual(target, { debug: true }),
```
WilcoFiers: These debuggers need to be removed.

test/commons/text/unicode.js
```diff
@@ -0,0 +1,104 @@
+describe('text.isNonBmpUnicode', function() {
```
WilcoFiers: You should add tests for the different unicode ranges.

test/integration/rules/label-content-name-mismatch/label-content-name-mismatch.json
```diff
@@ -0,0 +1,14 @@
+{
+	"description": "label-content-name-mismatch tests",
+	"rule": "label-content-name-mismatch",
+	"violations": [["#fail1"], ["#fail2"], ["#fail3"], ["#fail4"], ["#fail5"]],
```
WilcoFiers: Please add "incomplete" integration tests as well.

test/rule-matches/label-content-name-mismatch-matches.js
```diff
@@ -0,0 +1,144 @@
...
+	});
+
+	it('returns false if given element has no role', function() {
+		var vNode = queryFixture('<div id="target"></div>');
+		var actual = rule.matches(vNode.actualNode, vNode);
+		assert.isFalse(actual);
+	});
+
+	it('returns false if element role is not supported with name from contents', function() {
+		var vNode = queryFixture('<textarea id="target"></textarea>');
```
WilcoFiers: This elm could just as easily be false because it doesn't have an accessible name and/or text content. This test doesn't prove what you say it does.

test/rule-matches/label-content-name-mismatch-matches.js
```diff
@@ -0,0 +1,144 @@
...
+
+	it('returns false for `non-widget` role', function() {
+		var vNode = queryFixture(
+			'<a role="contentinfo" id="target" aria-label="some content">Content Information</a>'
+		);
+		var actual = rule.matches(vNode.actualNode, vNode);
+		assert.isFalse(actual);
+	});
+
+	it('returns true for non-widget role that does support name from content', function() {
```
WilcoFiers: I don't think you resolved my comment here.

test/rule-matches/label-content-name-mismatch-matches.js
```diff
@@ -0,0 +1,144 @@
...
+		var vNode = queryFixture(
+			'<button id="target" aria-label="close"><i class="fa fa-icon-close"></i></button>'
+		);
+		var actual = rule.matches(vNode.actualNode, vNode);
+		assert.isFalse(actual);
+	});
+
+	it('returns false for hidden (non visible) text content', function() {
+		var vNode = queryFixture(
+			'<button id="target" aria-label="close"><span style="display:none"></span></button>'
```
WilcoFiers: There is no hidden content in this test.

test/rule-matches/label-content-name-mismatch-matches.js
```diff
@@ -0,0 +1,144 @@
...
+	});
+
+	it('returns true when visible text is combination of alphanumeric and punctuation characters', function() {
+		var vNode = queryFixture(
+			'<button id="target" aria-label="next page">next page &gt;</button>'
+		);
+		var actual = rule.matches(vNode.actualNode, vNode);
+		assert.isTrue(actual);
+	});
+});
```
WilcoFiers: You should add a test for an elm with an implicit name-from-content role that is replaced with an explicit role that does not have name from content. For example `<button role="tooltip">`

package.json
```diff
@@ -86,6 +86,7 @@
 		"clone": "~2.1.1",
 		"css-selector-parser": "^1.3.0",
 		"dot": "~1.1.2",
+		"emoji-regex": "^7.0.3",
```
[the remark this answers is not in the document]
jkodu: Locked version to `7.0.3` and performed an audit.

image

test/rule-matches/label-content-name-mismatch-matches.js
jkodu: `tooltip` does support name from content, for example `status` does not, added a testcase.

lib/checks/label/label-content-name-mismatch.js
```diff
@@ -34,15 +34,17 @@ function isStringContained(compare, compareWith) {
...
  *
  * @param {String} str given text to curate
  * @returns {String}
  */
 function curateString(str) {
-	return text.sanitize(
-		text.replacePunctuation(
-			text.replaceNonBmpUnicode(text.replaceEmojiUnicode(str))
-		)
+	const noEmojiUnicodeStr = text.replaceUnicode(str, (type = 'emoji'));
```
WilcoFiers: Why not:

```js
text.replaceUnicode(str, {
  emoji: true,
  nonBmp: true,
  punctuations: true
})
```

jkodu: Firstly, unicode only comprises `emoji` and `nonBmp` at the moment. Punctuations are handled separately. I wouldn't want to mix and match everything into the same options object as it can get verbose. 

Hence went with the simpler `type=*` argument.

jkodu: After discussion, updated as agreed.

test/integration/rules/label-content-name-mismatch/label-content-name-mismatch.html
```diff
@@ -0,0 +1,35 @@
...
+<div id="fail3" role="link" aria-label="OK">Next</div>
+<button id="fail4" name="link" aria-label="the full">The full label</button>
+<div id="labelForFail5">123</div>
+<div role="button" id="fail5" aria-labelledby="labelForFail5">some content</div>
+
+<!-- incomplete -->
+<button id="incomplete1" aria-label="comet">☄️</button>
+<button id="incomplete2" aria-label="☄️">shooting star</button>
+<button id="incomplete3" aria-label="help">?</button>
+<button id="incomplete4" aria-label="&#x1F354">&#x1F354</button>
```
WilcoFiers: You didn't close this entity code with a `;`.

test/integration/rules/label-content-name-mismatch/label-content-name-mismatch.html
```diff
@@ -0,0 +1,35 @@
...
+<button id="incomplete5" aria-label="close">&#10060;</button>
+<button id="incomplete6" aria-label="wink">;)</button>
+<button id='incomplete7' aria-label="close">X</button>
+<button id="incomplete8" aria-label="comet">☄️</button>
+<button id="incomplete9" aria-label="help">?</button>
+<button id="incomplete10" aria-label="&#x1F354">&#x1F354</button>
```
WilcoFiers: Same here.

test/rule-matches/label-content-name-mismatch-matches.js
```diff
@@ -0,0 +1,154 @@
...
+		return rule.id === 'label-content-name-mismatch';
+	});
+
+	afterEach(function() {
+		fixture.innerHTML = '';
+		axe._tree = undefined;
+	});
+
+	it('returns false if given element has no role', function() {
+		var vNode = queryFixture('<div id="target"></div>');
```
WilcoFiers: This test doesn't have content or aria-label either. Your test doesn't prove what it presupposes.

test/rule-matches/label-content-name-mismatch-matches.js
```diff
@@ -0,0 +1,154 @@
...
+		var actual = rule.matches(vNode.actualNode, vNode);
+		assert.isFalse(actual);
+	});
+
+	it('returns false if element role is not supported with name from contents', function() {
+		var vNode = queryFixture(
+			'<div id="target" role="progressbar" aria-valuenow="20" aria-valuemin="0" aria-valuemax="100">20 %</div>'
```
WilcoFiers: Same here, aria-label(ledby) is missing.

test/rule-matches/label-content-name-mismatch-matches.js
```diff
@@ -0,0 +1,154 @@
...
+		);
+		var actual = rule.matches(vNode.actualNode, vNode);
+		assert.isFalse(actual);
+	});
+
+	it('returns false if implicit element role is overridden to a role that dos not support name from contents', function() {
```
WilcoFiers: "dos" -> "does"

test/rule-matches/label-content-name-mismatch-matches.js
```diff
@@ -0,0 +1,154 @@
...
+		var vNode = queryFixture(
+			'<button id="target" role="status">Your changes were automatically saved.</button>'
```
WilcoFiers: Same here, aria-label(ledby) is missing.

Issue #1054, opened 2018-08-06
CSSOM: Tackle nested @import stylesheets


jkodu:
As an enhancement to the work done on PR - https://github.com/dequelabs/axe-core/pull/958, which aims to fetch and resolve all `CSSStyleSheets` that are `inline, relative and are cross-domain`, there is a need to parse `CSSStyleRules` from each sheet to understand the usage of `@import` statements to refer to relative or cross-domain sheets, to ascertain if they are to be fetched.

These references can happen at various nested levels and a returned result/ stylesheet of an @import resource could lead to further @imports and so on, so care should be taken to handle infinite recursion.

The issues to tackle are:

- [ ] nested @import rules at multiple levels (eg: nested @import inside an @import(ed) resource).

- [ ] media/ conditionals to @import or link rules -https://developer.mozilla.org/en-US/docs/Web/CSS/@import

- [ ] @import inside of media queries

jkodu:
It is worth tackling this, only after https://github.com/dequelabs/axe-core/pull/1113 is merged.

PR #1339, opened 2019-01-29
feat(utils): Update CSSOM for nested @import computation

jkodu:
This PR does the below:
- handles cssom resolution from nested `@import`.
	- tackles multiple level nesting @import.
	- tackles cyclic nesting via @import.
	- tackles duplication of @imports across sheets.
	- tackles both cross origin and same origin @import(s).
- updated tests
- using `Promise` instead of `axe.utils.queue`

Closes issue:
- https://github.com/dequelabs/axe-core/issues/1054

## Reviewer checks


jkodu:
Refactoring of CSSOM will be done on the code in `develop` branch & then this PR will be re-opened.

lib/core/utils/parse-crossorigin-stylesheets.js
```diff
@@ -0,0 +1,56 @@
...
+			 */
+			return axe.utils.parseStylesheet(
+				result.sheet,
+				options,
+				priority,
+				importedUrls,
+				result.isCrossOrigin
+			);
+		})
+		.catch(e => Promise.reject(e));
```
stephenmathieson: This `.catch` block is unnecessary and can safely be removed.

lib/core/utils/parse-crossorigin-stylesheets.js
```diff
@@ -0,0 +1,56 @@
...
+				root: options.rootNode,
+				shadowId: options.shadowId
+			});
+
+			/**
+			 * Note:
+			 * Safety check to stop recursion, if there are numerous nested `@import` statements
+			 */
+			if (importedUrls.length > axe.constants.preload.maxImportUrls) {
+				return Promise.resolve(result);
```
stephenmathieson: Why wrap `result` in a Promise? You can just do `return result` instead.

lib/core/utils/parse-sameorigin-stylesheets.js
```diff
@@ -0,0 +1,124 @@
...
+			const newPriority = [...priority, cssRuleIndex];
+			const axiosOptions = {
+				method: 'get',
+				url: importUrl,
+				timeout: options.timeout
+			};
+			const isCrossOriginRequest = /^https?:\/\/|^\/\//i.test(importUrl);
+
+			importedUrls.push(importUrl);
+			return axe.imports
```
stephenmathieson: This looks like the same code from `parseCrossOriginStylesheets`. It even includes the same unnecessary Promise-wrapping.

Can we clean this up (and only worry about these LOCs once)?

lib/core/utils/parse-sameorigin-stylesheets.js
```diff
@@ -0,0 +1,124 @@
...
+	const nonImportCSSRules = rules.filter(r => r.type !== 3);
+
+	// no further rules to process in this sheet
+	if (!nonImportCSSRules.length) {
+		return Promise.all(promises);
+	}
+
+	// convert all `nonImportCSSRules` style rules into `text` and chain
+	promises.push(
+		new Promise(resolve =>
```
stephenmathieson: Just do:

```js
promises.push(Promise.resolve(options.convertDataToStylesheet({...}))
```

lib/core/utils/preload-cssom.js
```diff
@@ -27,25 +26,26 @@ axe.utils.preloadCssom = function preloadCssom({
...
 	if (!rootNodes.length) {
-		return q;
+		return Promise.all(promises);
 	}
 
 	const dynamicDoc = document.implementation.createHTMLDocument();
 	const convertDataToStylesheet = getStyleSheetFactory(dynamicDoc);
 
-	q.defer((resolve, reject) => {
+	const p = new Promise((resolve, reject) => {
```
stephenmathieson: Rather than wrapping this in a Promise, we should just use the Promise returned by `getCssomForAllRootNodes` directly.

lib/core/utils/preload.js
```diff
@@ -74,36 +74,37 @@ axe.utils.getPreloadConfig = function getPreloadConfig(options) {
...
 	if (!shouldPreload) {
-		return q;
+		return Promise.all(promises);
 	}
 
 	const preloadConfig = axe.utils.getPreloadConfig(options);
 
 	preloadConfig.assets.forEach(asset => {
-		q.defer((resolve, reject) => {
+		const p = new Promise((resolve, reject) => {
```
stephenmathieson: Why wrap this in a Promise?

lib/core/base/audit.js
```diff
@@ -436,21 +436,21 @@ Audit.prototype.run = function(context, options, resolve, reject) {
...
-					res(assets);
-				})
+				.then(assets => resolve(assets))
 				.catch(err => {
+					/**
+					 * Note:
+					 * we do not reject, to allow other (non-preload) rules to `run`
+					 * -> instead we resolve as `undefined`
+					 */
 					// resolve as undefined, to allow rule.run to continue
```
straker: duplicate comment?

lib/core/utils/parse-crossorigin-stylesheets.js
```diff
@@ -0,0 +1,54 @@
...
+ * @param {Array<String>} importedUrls urls of already imported stylesheets
+ * @param {Boolean} isCrossOrigin boolean denoting if a stylesheet is `cross-origin`
+ * @returns {Promise}
+ */
+axe.utils.parseCrossOriginStylesheets = function parseCrossOriginStylesheets(
+	url,
+	options,
+	priority,
+	importedUrls,
+	isCrossOrigin
```
straker: Is this param needed? I would assume it would always be `true` since the function is specifically used to load and parse cross origin stylesheets.

lib/core/utils/parse-crossorigin-stylesheets.js
```diff
@@ -0,0 +1,54 @@
...
+) {
+	const axiosOptions = {
+		method: 'get',
+		url,
+		timeout: options.timeout
+	};
+
+	importedUrls.push(url);
+	return axe.imports.axios(axiosOptions).then(({ data }) => {
```
straker: Do we need a `catch`? None of the calls to this function handle it.

lib/core/utils/parse-crossorigin-stylesheets.js
```diff
@@ -0,0 +1,54 @@
...
+			root: options.rootNode,
+			shadowId: options.shadowId
+		});
+
+		/**
+		 * Note:
+		 * Safety check to stop recursion,
+		 * if there are numerous nested `@import` statements
+		 */
+		if (importedUrls.length > axe.constants.preload.maxImportUrls) {
```
straker: Since `importedUrls` is passed around from the start of getting all stylesheets, won't this check only allow you to load 10 files max? So if there are 11 stylesheets, and the first 10 have no import but the 11th one does, we won't load any imports from the 11th file? Also, this prevents a single css file from having more than 10 imports (even if they are all leaf node files) since we won't load the 11th.

lib/core/utils/parse-sameorigin-stylesheets.js
```diff
@@ -0,0 +1,94 @@
...
+		// only href that are not already imported
+		.filter(url => !importedUrls.includes(url));
+
+	/**
+	 * iterate `@import` rules and fetch styles
+	 */
+	const promises = cssImportUrlsNotAlreadyImported.map(
+		(importUrl, cssRuleIndex) => {
+			const newPriority = [...priority, cssRuleIndex];
+			const isCrossOriginRequest = /^https?:\/\/|^\/\//i.test(importUrl);
```
straker: So even if the final url is the same origin as the current page, this check will mark it as cross origin.

So if the page is `http://example.com` and has a stylesheet url of `http://example.com/styles.css`, it will be treated as cross origin (even though it is same origin).

lib/core/utils/parse-sameorigin-stylesheets.js
```diff
@@ -0,0 +1,94 @@
...
+
+			return axe.utils.parseCrossOriginStylesheets(
```
straker: Why is a same origin stylesheet going through the cross origin parse function? If the function's main purpose is to load a cross origin stylesheet or a same origin or cross origin import, we should probably rename the rule to reflect that it loads stylesheet urls (same or cross origin) and doesn't parse them.

lib/core/utils/parse-sameorigin-stylesheets.js
jkodu: Nope. 

The lines here - https://github.com/dequelabs/axe-core/pull/1339/files/bb47753e63ef18f52d17f3efd4336635062509f7#diff-c5302a50f699518e36fca8d4a242798dR46, does the filtering to ensure only `url's` that are not from the same origin flow into this code path.

lib/core/utils/parse-crossorigin-stylesheets.js
jkodu: What do you mean by leaf node files?
Also `importedUrls` is only absolute `urls` and not relative ones.
The number `10` is up for debate. But yes, only first 10 will be loaded.

lib/core/utils/parse-crossorigin-stylesheets.js
jkodu: We fail safely if any preload fails, see - https://github.com/dequelabs/axe-core/pull/1339/files#diff-093a28bf6f7487126983e1ddcb9fc288R445, hence the top-level catch.

lib/core/utils/parse-sameorigin-stylesheets.js
jkodu: Given 

`@import "iam/a/relative/style.css"`
`@import "https://iam.io/external.css"`

The first one needs to be interpreted as `relative` with `isCrossOrigin` flag as `false`, and the other one as `true`. Have to use `axios` to fetch a relative or absolute resource & interpret it accordingly.

lib/core/utils/parse-crossorigin-stylesheets.js
jkodu: This is why we need it - https://github.com/dequelabs/axe-core/pull/1339#discussion_r278497604

lib/core/utils/parse-crossorigin-stylesheets.js
jkodu: Note: Try and use `preloadAssetsTimeout` to exit for recursion, as against using a random number `maxImportUrls`

test/core/utils/preload.js
```diff
@@ -1,81 +1,54 @@
...
-				assert.isDefined(results);
-				done();
-			})
-			.catch(function(error) {
-				done(error);
-			});
+	it('returns a Promise', function() {
+		var actual = axe.utils.preload({});
+		assert.isTrue(
+			Object.prototype.toString.call(actual) === '[object Promise]'
```
straker: This won't work if the Promise is polyfilled in the test environment (such as IE). Would be better to do `actual instanceof Promise`

stephenmathieson: The [algorithm used here](https://github.com/then/is-promise/blob/master/index.js) is downloaded 6.5 million times a week. I suggest we use it rather than `instanceof` or `toString()` checks.

jkodu: This is a simple assertion, so I would rather leave it as is.

straker:
@WilcoFiers needs to have final say. Changes look good to me with two caveats:

1. The test to check if a promise is returned will break the tests in any environment where Promise is polyfilled, such as IE 11. If we're OK with that then no change is needed.
2. The unit and integration tests for the main parse stylesheet function tests functionality as a whole set. If we're OK that the separate util functions are not tested individually then no change is needed.

Issue #947, opened 2018-06-07
Add environment details to output format


WilcoFiers:
In order to make better identify causes of issues it would be helpful for Axe-core to provide some information in the output format of the test environment. In addition to the URL, window width / height is often a cause of differences in results. User agents can also be responsible, so it'd be helpful to have that in there too.

WilcoFiers:
Another thought, we should maybe include the runOnly object if it was provided, since that too is a common cause for confusion. Possibly also include if custom rules were used or if it was running stock rules.

marcysutton:
Ooh, these are all great ideas!

WilcoFiers:
Proposed format:

```json
{
  "testEngine": {
    "name": "axe-core",
    "version": "3.1.2" // axe.version
  },
  "testTool": {
    "name": "AttestChrome", // axe.configure({ branding { application }})
    "version": "2.3.0" // Need to add to axe.configure({ branding })
  },
  "toolOptions": {}, // available in the reporter
  "testEnvironment": {
    "userAgent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/72.0.3626.81 Safari/537.36",
    "windowWidth": 1280, // window.innerWidth
    "windowHeight": 1024,
    "orientationAngle": 0, // If available, window.screen.orientation.angle
    "orientationType": "landscape-primary"
  }
```

WilcoFiers:
What'd be very cool for a future version is if it could also return the scope. This is available on the context object. We could take selectors from the node(s) in the include / exclude properties of those and put them in the output. That would require adding the context to the reporter, which will require an API change unfortunately.

PR #1353, opened 2019-02-06
feat: add environment details to results

straker:
Adds tool options, axe-core version, branding name, and environment details to the result object.

Closes issue: #947 

## Reviewer checks


axe.d.ts
```diff
@@ -35,6 +49,10 @@ declare namespace axe {
 		resultTypes?: resultGroups[];
 	}
 	interface AxeResults {
+		toolOptions: RunOptions;
```
stephenmathieson: Are all of these new properties _always_ in the results object? If not, we want to make them optional:

```
interface AxeResults {
  toolOptions?: RunOptions;
  testEngine?: TestEngine;
  // ...
}
```

lib/core/reporters/na.js
```diff
@@ -9,7 +9,25 @@ axe.addReporter('na', function(results, options, callback) {
 	}
 
 	var out = helpers.processAggregate(results, options);
+	var orientation =
```
stephenmathieson: We should probably use the standards-compliant object first:

```js
var orientation = screen.orientation || screen.msOrientation || screen.mozOrientation || {}
```

Just incase someone defined a non-standard property of the same name.

Not a big deal either way tho.

lib/core/reporters/no-passes.js
```diff
@@ -9,8 +9,25 @@ axe.addReporter('no-passes', function(results, options, callback) {
 	options.resultTypes = ['violations'];
 
 	var out = helpers.processAggregate(results, options);
+	var orientation =
```
stephenmathieson: Is it worth moving this into a `getScreenOrientation` function since it's duplicated across files?

Again, not a big deal IMO, just a thought.

lib/core/reporters/v2.js
```diff
@@ -9,7 +9,26 @@ axe.addReporter(
 			options = {};
 		}
 		var out = helpers.processAggregate(results, options);
+		var orientation =
+			screen.msOrientation ||
+			(screen.orientation || screen.mozOrientation || {});
+
 		callback({
+			toolOptions: options,
```
stephenmathieson: Since we're copy/pasting this frequently, does it belong in its own function?

Then we could do:

```
callback({
   ...getEnvironmentDetails(),
  violations: out.violations,
  passes: out.passes,
  // ...
})
```

I'm not sure whether or not it's "worth it", but I generally consider it time to refactor when the same code is copied into 3 different places.

test/core/reporters/na.js
```diff
@@ -245,4 +245,25 @@ describe('reporters - na', function() {
 			done();
 		});
 	});
+	it('should add version information', function(done) {
+		axe.run(naOption, function(err, results) {
+			try {
```
stephenmathieson: This `try/catch` seems very awkward. Can't we just do:

```js
axe.run(naOption, function (err, results) {
  if (err) {
    return done(err)
  }

  // Assertions go here
  done()
})
```

test/core/reporters/no-passes.js
```diff
@@ -209,4 +209,25 @@ describe('reporters - no-passes', function() {
 			done();
 		});
 	});
+	it('should add version information', function(done) {
+		axe.run(noPassOpt, function(err, results) {
+			try {
```
stephenmathieson: Same comment about the `try/catch`.

test/core/reporters/v1.js
```diff
@@ -283,4 +283,25 @@ describe('reporters - v1', function() {
 			done();
 		});
 	});
+	it('should add version information', function(done) {
+		axe.run(optionsV1, function(err, results) {
+			try {
```
stephenmathieson: Same comment about the `try/catch`.

test/core/reporters/v1.js
```diff
@@ -283,4 +283,25 @@ describe('reporters - v1', function() {
...
+				assert.isNull(err);
+				assert.isObject(results.testEngine, 'tesetEngine');
+				assert.equal(results.testEngine.name, 'axe-core');
```
stephenmathieson: Why are we testing the same thing 3 different times? Is there a case where 1 of these tests _could_ fail but the others wouldn't? If not, I think we only need to do this once.

test/core/reporters/v2.js
```diff
@@ -227,4 +227,25 @@ describe('reporters - v2', function() {
 			done();
 		});
 	});
+	it('should add version information', function(done) {
+		axe.run(optionsV2, function(err, results) {
+			try {
```
stephenmathieson: Same duplicate test and `try/catch` concern.

test/core/reporters/na.js
straker: The `try/catch` is for the assertions. A failed assertion throws an error but `done` is not called with it, so the unit test just hangs until timeout, then stops the test there without reporting what the error was.

lib/core/reporters/v2.js
straker: I thought about this, but everything was duplicated across the different reporters so I just went with it. If we did split this out, would it be its own `utils` function?

lib/core/reporters/na.js
straker: We could, I just copied the MDN code to grab it https://developer.mozilla.org/en-US/docs/Web/API/Screen/orientation

axe.d.ts
straker: All reporters now have it except the raw reporter, which I thought used the `Result` interface

stephenmathieson: OK, works for me. I just wanted to verify.

lib/core/reporters/na.js
stephenmathieson: 🤷‍♀️ OK, that's fine I suppose

lib/core/reporters/v2.js
stephenmathieson: I don't know where it belongs. I don't think it makes sense to put it on the public `axe.utils` namespace.

@WilcoFiers where do "internal" utilities go?

test/core/reporters/na.js
stephenmathieson: Ohh, wow, OK. That's something we should change. I wonder if updating `mocha` would solve the problem for us.

Obviously don't worry about this now.

test/core/reporters/v1.js
stephenmathieson: These are different reporters. Just answered my own question. Disregard my question.

lib/core/reporters/na.js
```diff
@@ -9,7 +9,25 @@ axe.addReporter('na', function(results, options, callback) {
...
+		testRunner: {
+			name: axe._audit.brand
+		},
+		testEnvironment: {
+			userAgent: navigator.userAgent,
+			windowWidth: window.innerWidth,
+			windowHeight: window.innerHeight,
+			orientationAngle: orientation.angle,
+			orientationType: orientation.type
+		},
```
WilcoFiers: Instead of duplicating this code a bunch of times, it'd be better to create a helper for this in `lib/core/reporters/helpers`, so you could just do:

```js
callback({
  ...getEnvironmentData(),
  toolOptions: options,
  violations: out.violations,
  passes: out.passes,
  incomplete: out.incomplete,
  inapplicable: out.inapplicable
})
```

You could probably have the URL & timestamp returned from it too. That's probably a better place for those two than the processAggregate helper.

stephenmathieson: @WilcoFiers please see https://github.com/dequelabs/axe-core/pull/1353#discussion_r254805702

We've discussed this already, but neither @straker nor I know where to put things like this.

PR #1361, opened 2019-02-11
fix(aria-allowed-role): Allow iframe role=none [#1360]

### LINKED ISSUE #1360: Iframe fails with role presentation
Expectation: The following code should not identify it as a best practice
<code><iframe role="presentation"></iframe></code>
Providing an role presentation on iframe is allowed

Actual: aXe extension throws best practice violation aganist an iframe with role presentation

<pre><code>
axe-core version: 3.1.2
</code></pre>

WilcoFiers:
As reported by Soma, iframes should be allowed role=none|presentation. See:
https://www.w3.org/TR/html-aria/#iframe

Closes #1360 

## Reviewer checks


straker:
LGTM

lib/commons/aria/index.js
```diff
@@ -1407,9 +1407,11 @@ lookupTable.role = {
 					'h5',
 					'h6',
 					'header',
+					'iframe',
 					'li',
+					'ol',
 					'section',
-					'ol'
+					'ul'
```
dylanb: Why this addition?

WilcoFiers: Not sure how this got overlooked, but it the spec says these are allowed.

Issue #1137, opened 2018-09-13
DOMException: Failed to execute 'matches' on 'Element'


stephenmathieson:
Attempting to run axe-core on https://eclkc.ohs.acf.hhs.gov/education-child-development yields the following error:

```
VM60:3472 Uncaught (in promise) DOMException: Failed to execute 'matches' on 'Element': 'a[href="  /child-screening-assessment
"] > .card__image > .card__text' is not a valid selector.
    at Object.matchesSelector (<anonymous>:3472:26)
    at <anonymous>:4002:28
    at Array.filter (<anonymous>)
    at generateSelector (<anonymous>:4001:27)
    at Object.createUniqueSelector [as getSelector] (<anonymous>:4038:14)
    at DqElement.get selector [as selector] (<anonymous>:3429:48)
    at <anonymous>:2399:51
    at Array.map (<anonymous>)
    at <anonymous>:2392:47
    at Array.map (<anonymous>)
```

It looks like this page contains newlines in attribute values, which causes axe-core to produce invalid selectors. Example (note the `href`):

```html

<a href="  /child-screening-assessment
  " class="no-icon" rel="bookmark">
    [...]
</a>
```

This error occurs with axe-core versions 3.0.3 and 3.1.1, but 2.5.0 is fine.

straker:
The bug is in the `axe.utils.getSelector` code where the only thing we don't escape is the `href` attribute. https://github.com/dequelabs/axe-core/blob/320e63380189f507a32c99fd19619f3c6a161764/lib/core/utils/get-selector.js#L44

stephenmathieson:
Nice find, Steven!

PR #1366, opened 2019-02-14
fix: escape href attribute when creating a CSS selector

straker:
When creating a selector through `utils.getSelector`, the `href` attribute was not run through the escapeCSS function and so would produce improper CSS selector strings.

Closes: #1137 

## Reviewer checks


lib/core/utils/get-selector.js
```diff
@@ -41,7 +41,11 @@ function getAttributeNameValue(node, at) {
 				return;
```
jkodu: At L39 above, is it worth doing `escapeSelector(value)` as well?

straker: Probably would be a good idea as well

PR #1343, opened 2019-01-30
chore: remove "linting" from Grunt

stephenmathieson:
This patch removes all of the ESLint stuff from the Grunt configuration. There have been numerous efforts to "fix" the existing Grunt/ESLint setup which all seem unnecessary. Now, if you want to lint your code, just do `npm run eslint`. If you don't want to lint your code, then don't do `npm run eslint`.

Ref: #1342, #1314, #1267, #1255, #1247, and probably others...

## Reviewer checks

**Required fields, to be filled out by PR reviewer(s)**
- [x] Follows the commit message policy, appropriate for next version
- [x] Has documentation updated, a DU ticket, or requires no documentation change
- [x] Includes new tests, or was unnecessary
- [x] Code is reviewed for security by: @WilcoFiers

PR #1381, opened 2019-02-19
chore: add eslint to lint staged

jkodu:
**Note: looking for opinion**

Following this PR - https://github.com/dequelabs/axe-core/pull/1343, `npm run eslint` does not run locally at all, which leads to some basic lint errors getting into CI & then the developer having an Aha moment. 

It should not be developer overhead to remember to run `eslint` before pushing. Hence adding this step to `lint-staged` which should catch lint errors (albeit not only on staged files). 

Verified that it works, locally.

Closes issue: NA

## Reviewer checks

### DIFF
--- a/package.json
+++ b/package.json
@@ -129,8 +129,8 @@
 			"prettier --write",
 			"git add"
 		],
-		"*.{js}": [
-			"npm run eslint --fix",
+		"*.js": [
+			"eslint --fix",
 			"prettier --write",
 			"git add"
 		]
--- a/package.json
+++ b/package.json
@@ -125,7 +125,12 @@
 	},
 	"dependencies": {},
 	"lint-staged": {
-		"*.{md,json,js,ts}": [
+		"*.{md,json}": [
+			"prettier --write",
+			"git add"
+		],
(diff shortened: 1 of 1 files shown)


jkodu:
@stephenmathieson - updated as requested.

package.json
```diff
@@ -125,7 +125,12 @@
 	},
 	"dependencies": {},
 	"lint-staged": {
-		"*.{md,json,js,ts}": [
+		"*.{md,json,ts}": [
+			"prettier --write",
+			"git add"
+		],
+		"*.{js}": [
```
stephenmathieson: These braces are unnecessary. Please replace with `"*.js"`.

package.json
```diff
@@ -125,7 +125,12 @@
...
+			"npm run eslint --fix",
```
stephenmathieson: Thanks for adding `--fix`, but this is still incorrect (for the same reason). We do not want to lint the _entire_ source tree just because a single file was `git add`ed.

Please replace this with:

```
"*.js": [
  "eslint --fix",
  "prettier --write",
  "git add"
]
```

Issue #1216, opened 2018-11-01
False positive: 'aria-valid-attr' fails for using 'aria-roledescription' on role button


iamrafan:
```aria-valid-attr``` fails for aria-roledescription as it was not supported across most of the Browser/ATs

![image](https://user-images.githubusercontent.com/22068434/47879519-09b2ef80-ddde-11e8-8c41-f78d362ab3af.png)

Latest AT testing shows that now there is much broader support for the usage of ```aria-roledescription```

- Link to live demo: https://codepen.io/iamrafan/pen/WYevMx?editors=1000

  | AT/Browser | Announced as | Is aria-roledescription announced?
-- | -- | -- | --
1 | Edge/Narrator | Here Split button | Yes
2 | Edge/JAWS | Here Button | **No**
3 | Edge/NVDA | Split button here | Yes
4 | Chrome/NVDA | Split button here | Yes
5 | Chrome/JAWS | Here Split button | Yes
6 | IE/JAWS | Here split button | Yes
7 | IE/NVDA | Split button here | Yes
8 | Firefox/NVDA | Split button here | Yes

<pre><code>axe-core version: 3.1.2
axe-Coconut version: 3.5.0</code></pre>

mfairchild365:
@WilcoFiers Yep, it works in JAWS+IE. I tested with JAWS 2018.1808.10 and IE 11.

Some other thoughts:

Thought 1: I'm curious how well it works with other elements and ARIA roles. My guess is that if it works fine with the `<button>` element, that it would work fine with other elements too, but that is a big assumption.

Thought 2: The [ARIA 1.2 spec says](https://www.w3.org/TR/wai-aria-1.2/#aria-roledescription) "Assistive technologies SHOULD use the value of aria-roledescription when presenting the role of an element, but SHOULD NOT change other functionality based on the role of an element that has a value for aria-roledescription."

The fact that it says "SHOULD" instead of "MUST" would suggest to me that this must pass the "accessibility supported" requirement even if assistive technologies don't implement it.

attilavago:
This needs to get fixed ASAP. Creates confusion everywhere...

mfairchild365:
A further note. I did some more research and my findings can be found here: https://a11ysupport.io/tech/aria/aria-roledescription_attribute (feel free to contribute any different findings)

In summary: the attribute is well supported for interactive elements like `<button>` but not so much for grouping/sectioning elements like `<section>`. I tested both situations because of this text in the ARIA spec:

[...]

PR #1382, opened 2019-02-19
fix: better unsupported attribute support for aria-roledescription

straker:
Expand the `unsupported` property of an attribute by allowing certain elements to use the attribute.

Closes: #1216

## Reviewer checks


doc/aria-supported.md
```diff
@@ -14,8 +14,7 @@ For a detailed description about how accessibility support is decided, see [How
 
 ## Attributes
 
-| aria-attribute       | axe-core support |
-| -------------------- | ---------------- |
-| aria-describedat     | No               |
-| aria-details         | No               |
-| aria-roledescription | No               |
```
WilcoFiers: We should do something to put the exceptions in here, something like "Everything except button, input[type=button], ...". Even better is we do something like "No [1]" and than under the table we put that list of exceptions as `1. Everything except ...`.

lib/checks/aria/unsupportedattr.js
```diff
@@ -9,8 +9,33 @@ let unsupported = Array.from(node.attributes)
...
-	this.data(unsupported);
+if (!unsupportedAttrs.length) {
+	return false;
+}
+
+const nodeName = node.nodeName.toUpperCase();
+const lookupTable = axe.commons.aria.lookupTable;
+const role = axe.commons.aria.getRole(node);
+
+unsupportedAttrs = unsupportedAttrs.filter(attr => {
```
WilcoFiers: I think this should be handled by `validateAttr`, instead of in the check.

build/tasks/aria-supported.js
```diff
@@ -60,6 +61,33 @@ module.exports = function(grunt) {
 									!subject.hasOwnProperty(key)
 								) {
 									out.push([`${key}`, 'No']);
+								} else if (
```
jkodu: the complexity of this `reduce` fn has become hard to follow, worth splitting this into smaller fn's, with detailed comments.

lib/commons/aria/attributes.js
```diff
@@ -37,13 +37,9 @@ aria.allowedAttr = function(role) {
  * @memberof axe.commons.aria
  * @instance
  * @param  {String} att The attribute name
- * @param  {Object} options Use `flagUnsupported: true` to report unsupported attributes
  * @return {Boolean}
  */
-aria.validateAttr = function(att, { flagUnsupported = false } = {}) {
+aria.validateAttr = function(att) {
```
jkodu: I know you are only removing an argument here, could you name the function here, as it will be easier to debug in the long run.

```
aria.validateAttr = function validateAttr(att)....
```

build/tasks/aria-supported.js
```diff
@@ -32,6 +32,46 @@ module.exports = function(grunt) {
...
+						const value = element.properties[prop];
+
+						// the 'type' property can be a string or an array
+						if (typeof value === 'string') {
+							return `\`<${element.nodeName} ${prop}="${value}">\``;
+						}
+
+						// output format for an array of types:
+						// <input type="button" | "checkbox">
+						let values = value.map(v => `"${v}"`).join(' | ');
```
jkodu: `let` -> `const`

build/tasks/aria-supported.js
```diff
@@ -32,6 +32,46 @@ module.exports = function(grunt) {
...
 			const aQaria = new Set(axe.utils.uniqueArray(roleAriaKeys, ariaKeys));
+			let footnotes;
+
+			/**
+			 * Parse a list of unsupported exception elements and add a footnote
+			 * detailing which HTML elements are supported.
+			 * @param {Array<String|Object>} elements List of supported elements
+			 */
+			const generateSupportFootnote = elements => {
+				let supportedElements = elements.map(element => {
```
jkodu: `let` -> `const`

build/tasks/aria-supported.js
```diff
@@ -32,6 +32,46 @@ module.exports = function(grunt) {
...
 			}, []);
 			const aQaria = new Set(axe.utils.uniqueArray(roleAriaKeys, ariaKeys));
+			let footnotes;
+
+			/**
+			 * Parse a list of unsupported exception elements and add a footnote
+			 * detailing which HTML elements are supported.
+			 * @param {Array<String|Object>} elements List of supported elements
+			 */
+			const generateSupportFootnote = elements => {
```
jkodu: Given that this is an extracted function, I do not see a reason for `footnotes` to be a top level variable.

I think we should scope `footnotes` to be within this function & return it, lying closer to the analogy of a pure function.

jkodu:
@WilcoFiers 
I have updated this, so on your plate to review.

lib/checks/aria/unsupportedattr.js
```diff
@@ -1,16 +1,33 @@
...
+	.filter(({ name }) => {
+		const attribute = lookupTable.attributes[name];
+		if (!axe.commons.aria.validateAttr(name) || !attribute.unsupported) {
+			return false;
+		}
+
+		const unsupported = attribute.unsupported;
+
+		if (typeof unsupported === 'boolean' || !unsupported.exceptions) {
+			return true;
```
WilcoFiers: Nitpick, rather than checking if unsupported is falsey on line 8, I think this code would be a little easier to read if you check on line 14 if unsupported is not an object, and on line 14 do `return !!unsupported`.

lib/checks/aria/unsupportedattr.js
```diff
@@ -1,16 +1,33 @@
...
+		}
+
+		// validate attributes and conditions (if any) from allowedElement to given node
+		let out = axe.commons.matches(node, unsupported.exceptions);
```
WilcoFiers: `let out =` -> `const isException =`

PR #1388, opened 2019-02-21
chore: fix npm spawn child process in windows with grunt run

jkodu:
Invoking `npm scripts` using `grunt-run` resulted in a child process spawn error in windows, as `grunt-run` perhaps could not resolve between `command prompt` vs `shell` internally.

In our case, we are executing a `node` script via `npm script`, so it makes sense to run the `node` script directly using `grunt-run`, thus bypassing the above issue.

The main change is here - https://github.com/dequelabs/axe-core/pull/1388/files#diff-35b4a816e0441e6a375cd925af50752cR334, rest are comments update and prettier going to town.

Checked with Wahaj @waabid, who confirmed that this worked in his machine, also tested locally on windows via bootcamp.

Closes issue: NA

## Reviewer checks


package.json
```diff
@@ -58,7 +58,6 @@
 		"start": "npm run develop",
 		"develop": "grunt dev --force",
 		"api-docs": "jsdoc --configure .jsdoc.json",
-		"imports-gen": "node ./build/imports-generator",
```
WilcoFiers: We need to move work out of grunt, not move more stuff into it.

jkodu: If you look at the changes here - https://github.com/dequelabs/axe-core/pull/1388/files#diff-35b4a816e0441e6a375cd925af50752cR334, all we have done is invoke `node` instead of a `npm` command via an already used plugin `grunt-run`, so nothing has been moved around per se.

stephenmathieson: Perhaps we could do the inverse and not invoke this via `grunt-run` but as an npm script instead.

jkodu:
After discussing with @WilcoFiers, re-added the npm script, which will be leveraged later when moving away from Grunt.

Issue #866, opened 2018-04-26
landmark-unique-name


ssanaul:
# landmark-unique-name

Ensures that landmark elements have unique accessible names

Add a unique accessible name using aria-label or aria-labelledby

**Tags:** best-practice

## Selector

html

### landmark-accessible-name-nav ( any )

using options, this rule could apply to multiple landmarks.

WilcoFiers:
I like this rule. Good idea. I would suggest the following changes:

- Change the ID to "landmark-unique-name"
- Use selector / matcher to select all landmarks
- Have two _any_ checks:
  - _page-no-duplicate-landmark_: Checks that only one landmark of this role exists. This should be a duplicate of "page-no-duplicate-main"
  - _landmark-unique-name_: Checks that the landmark has an accessible name, and that there is no landmark of the same role with the same name

dylanb:
@WilcoFiers duplicate landmarks on not even be a best practice violation per se (for example, you could quite easily have multiple navigation landmarks. Duplicate names for landmarks would be a best practice violation.

WilcoFiers:
That's what I intended for this to day :).

ssanaul:
Wanted to make sure we're on the same page for some of the code here...

Rule JSON:
```
{
  "id": "nav-unique-name",
  "selector": "nav:not([role]), [role=navigation]",
  "tags": [
    "best-practice"
  ],
  "metadata": {
    "description": "Labels help people using screen readers to identify the purpose of the links in the navigation landmark",
    "help": "Navigation landmarks must have unique accessible names"
  },
  "all": [],
  "any": [
    "page-no-duplicate-nav",
    "nav-unique-name"
  ],
  "none": []
}
```
nav-unique-name.js:
```
this.data(node);
return true;
```
nav-unique-name-after.js:
```
let names = [];
let labelVirtual = axe.commons.aria.labelVirtual;

//set results with null labels false
//push non-empty labels into seperate array
results.forEach(result => {
	let name = labelVirtual(result.data);
	!!name ? names.push(name) : result.result = false;
});

//filter duplicates into seperate array
let duplicates = names.filter(name => {
	names.includes(name, names.indexOf(name) + 1);
});

//set results with duplicate labels false
results.forEach(result => {
	result.result = !duplicates.includes(labelVirtual(result.data));
});

return results;
```

Haven't actually tested it yet; moreso wondering about the logic at high level before implementing.

@WilcoFiers 
cc: @jongund

[...]

Issue #1052, opened 2018-08-06
Add axe-core specific eslint rules


WilcoFiers:
Couple of things we should have rules for:

- Don't use node.matches()
- Don't use node.tagName, use node.nodeName instead
- Always follow node.nodeName with a .toUpperCase()

Adding Axe specific ESLint rules will help avoid common mistakes.

mohanraj-r:
Hi @WilcoFiers Is this related to the [eslint-plugin-axe - npm](https://www.npmjs.com/package/eslint-plugin-axe) plugin? Can you please describe the intended goal of the reserved npm package? 

I am wondering if axe rules can be expressed as an eslint plugin? Guess axe needs a DOM to run? Can we do static code analysis with axe?

jkodu:
Hi @mohanraj-r 	

To clarify this issue is related to enhancing the eslint rules within axe-core & not related to the package you have referred above.

mohanraj-r:
Thanks @jkodu, I had thought so too. Do you know that the npm package is for ?

PR #1394, opened 2019-02-28
feat(rule): Adding landmark-is-unique rule

waabid:
The rule checks for whether rules are unique. The structure is:

matches + selectors: identify landmarks.
evaluate: passes all nodes and adds role/label data.
after: fails all landmarks that have duplicate roles and labels.

Linked issue: #866

## Reviewer checks


lib/checks/landmarks/landmark-is-unique-after.js
```diff
@@ -0,0 +1,19 @@
...
+			currentResult.data.role === someResult.data.role &&
+			currentResult.data.label === someResult.data.label
+		);
+	};
+
+	const foundResultIndex = uniqueLandmarks.findIndex(findMatch);
+	if (foundResultIndex === -1) {
+		uniqueLandmarks.push(currentResult);
+	} else {
+		currentResult.result = false;
```
dylanb: We should be adding the information about the currentResult to the matching item as a `relatedNode` that way we only return one fail for each duplicated role and name and we can see the others directly in the result.

dylanb: @WilcoFiers I believe that we can build a list of related nodes for each element and the call the result object's `relatedNodes` helper function to turn it into the correct type of array - can you confirm?

lib/rules/landmark-unique-matches.js
```diff
@@ -0,0 +1,39 @@
+const excludedParentsForHeaderFooterLandmarks = [
```
dylanb: Based on testing with VO I think this filtering is unnecessary but need more data

lib/rules/landmark-unique-matches.js
```diff
@@ -0,0 +1,39 @@
...
+	const landmarkRoles = axe.commons.aria.getRolesByType('landmark');
+	const role = axe.commons.aria.getRole(actualNode);
+	const tagName = actualNode.nodeName.toLowerCase();
+
+	if (tagName === 'header' || tagName === 'footer') {
+		return isHeaderFooterLandmark(actualNode);
+	}
+
+	if (tagName === 'section' || tagName === 'form') {
+		const label = axe.commons.aria.label(actualNode);
```
dylanb: This will be slow on large (deep) pages or with lots of landmarks or when both conditions hold. Need to pass the `virtualNode` into `isLandmarkVirtual` and then into here. Where we do not need the `virtualNode`, we can dereference it there.

lib/rules/landmark-unique-matches.js
```diff
@@ -0,0 +1,39 @@
...
+		return !!label;
+	}
+
+	return (role && landmarkRoles.indexOf(role) >= 0) || role === 'region';
+}
+
```
dylanb: Replace from here on down with

```
return (
    isLandmarkVirtual(virtualNode) && axe.commons.dom.isVisible(node, true)
);
```

lib/rules/landmark-unique-matches.js
dylanb: for the record: we have decided that because this is a best-practice rule, it will follow the [ARIA 1.1 practices](https://www.w3.org/TR/wai-aria-practices-1.1/#aria_landmark) document regardless of treatment by browser/AT combinations

lib/rules/landmark-unique-matches.js
waabid: Replaced actual node usage with virtual node usage where there seemed to be virtual functions.

lib/rules/landmark-unique-matches.js
waabid: Done.

lib/checks/landmarks/landmark-is-unique-after.js
waabid: I followed the pattern set by access-keys; let me know what you think 😄

lib/checks/landmarks/landmark-is-unique-after.js
```diff
@@ -0,0 +1,30 @@
...
+				return index;
+			}
+		}
+		return -1;
+	};
+
+	const foundResultIndex = findIndex(uniqueLandmarks, findMatch);
+	if (foundResultIndex === -1) {
+		uniqueLandmarks.push(currentResult);
+		currentResult.relatedNodes = [];
```
jkodu: perhaps setting the relatedNodes to an empty array in the check itself, can help negate this step here?

lib/checks/landmarks/landmark-is-unique.js
```diff
@@ -0,0 +1,7 @@
+let role = axe.commons.aria.getRole(node);
```
jkodu: is it worth adding `if(!role) return false`?

lib/rules/landmark-unique-matches.js
```diff
@@ -0,0 +1,45 @@
...
+		headerFooterElement,
+		excludedParentsForHeaderFooterLandmarks
+	);
+}
+
+function isLandmarkVirtual(virtualNode) {
+	const { actualNode } = virtualNode;
+	const landmarkRoles = axe.commons.aria.getRolesByType('landmark');
+	const role = axe.commons.aria.getRole(actualNode);
+	const tagName = actualNode.nodeName.toLowerCase();
```
jkodu: as a general practise with in `axe-core`, when using `node.nodeName`, we tend to always use `toUpperCase()`, please replace this with `const nodeName = actualNode.nodeName.toUpperCase()`.

We are looking to enhance our `eslint` rules to catch this at the moment. See https://github.com/dequelabs/axe-core/issues/1052 for more information.

You will have to update your usages as well `nodeName === 'HEADER'` and so on.

lib/rules/landmark-unique-matches.js
```diff
@@ -0,0 +1,45 @@
...
+	if (tagName === 'header' || tagName === 'footer') {
+		return isHeaderFooterLandmark(virtualNode);
+	}
+
+	if (tagName === 'section' || tagName === 'form') {
+		const label = axe.commons.aria.labelVirtual(virtualNode);
+		return !!label;
+	}
+
+	return (role && landmarkRoles.indexOf(role) >= 0) || role === 'region';
```
jkodu: exiting early, when !role, would help with the readability and remove the && here.

lib/checks/landmarks/landmark-is-unique-after.js
```diff
@@ -0,0 +1,30 @@
...
+		return true;
+	} else {
```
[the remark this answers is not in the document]
dylanb: @jkodu I disagree! But there is a problem with it.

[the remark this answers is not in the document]
dylanb: Correction! @wahaj the rule seems to be working fine and adding the `relatedNodes` correctly in single iframe pages.

```
"{
 "id": "landmark-unique",
 "impact": "moderate",
 "tags": [
  "cat.semantics",
  "best-practice"
 ],
 "description": "Ensures landmarks are unique",
 "help": "Landmarks must have a unique role or role/label combination(aria-label OR aria-labelledby)",
 "helpUrl": "https://dequeuniversity.com/rules/axe/3.1/landmark-unique?application=axeAPI",
 "nodes": [
  {
   "any": [
    {
     "id": "landmark-is-unique",
     "data": {
      "role": "region",
      "label": "not unique"
     },
     "relatedNodes": [
      {
       "html": "<div role=\"region\" aria-label=\"not unique\">Second occurence</div>",
       "target": [
        "main > div"
       ]
      }
     ],
     "impact": "moderate",
     "message": "The landmarks do not have a unique role or role/label combination (use aria-label OR aria-labelledby to make landmarks distinguishable)."
    }
   ],
   "all": [],
   "none": [],
   "impact": "moderate",
   "html": "<div role=\"region\" aria-label=\"not unique\">First occurence</div>",
   "target": [
    "body > div"
   ],
   "failureSummary": "Fix any of the following:\n  The landmarks do not have a unique role or role/label combination (use aria-label OR aria-labelledby to make landmarks distinguishable)."
  }
 ]
}"
```

lib/rules/landmark-unique-matches.js
dylanb: @wahaj reason for this is XHTML documents

test/integration/rules/landmark-unique/frame.html
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<meta charset="utf8" />
+		<title>landmark-unique test</title>
+		<script src="/axe.js"></script>
+	</head>
+	<body>
+		<main id="violation-main-2">Second main</main>
```
dylanb: For this test can you add two landmarks, one with only a role and one with a non-empty name and a role. That way we can be certain that the name is actually being passed up the iframe chain.

test/rule-matches/landmark-unique-matches.js
```diff
@@ -0,0 +1,94 @@
...
+			it(`should match because ${elementType} is not contained inside the excluded descendents`, () => {
+				axeFixtureSetup(`<${elementType}>an element</${elementType}>`);
+				const node = fixture.querySelector(elementType);
+				assert.isTrue(rule.matches(node, null));
+			});
+		});
+	});
+
+	(shadowSupport ? it : xit)(
+		'return true for landmarks contained within shadow dom',
```
dylanb: You should add some tests to ensure that the hierarchical exclusions are working across shadow DOM boundaries

lib/rules/landmark-unique-matches.js
dylanb: @waabid I am still showing the old code

waabid: Sorry; I think I misunderstood you. I assume the new changes are what you're looking for. 😄

lib/checks/landmarks/landmark-is-unique.js
waabid: The reason I didn't add it is because we can be certain by the time we get here, we know for sure that it has a role (since matches would filter out other scenarios).

If you still feel strongly that I should add it, I can.

lib/checks/landmarks/landmark-is-unique-after.js
```diff
@@ -0,0 +1,30 @@
+const uniqueLandmarks = [];
+return results.filter(currentResult => {
+	const findMatch = someResult => {
+		return (
+			currentResult.data.role === someResult.data.role &&
+			currentResult.data.label === someResult.data.label
+		);
+	};
+
+	const findIndex = (array, matchingFunction) => {
```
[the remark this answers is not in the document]
waabid: I actually just used find instead (which is there in polyfills). 😄 

This was more for me to validate that this was the issue anyways.

lib/rules/landmark-unique-matches.js
waabid: Done 😄

lib/rules/landmark-unique-matches.js
waabid: Good call! Added.

lib/checks/landmarks/landmark-is-unique-after.js
waabid: I followed the accesskey pattern here; this is because I need info from the node itself (and as such the relatedNodes includes itself when we get to after).

Let me know if this makes sense or I am missing something here. 😄

test/integration/rules/landmark-unique/frame.html
dylanb: @waabid I should have been clearer, we also need an assertion on the actual values. I did not see one. If I missed, it, please point it out

waabid: Sorry, I may not be following?

FTR; I added "violation-form-through-iframe-1", which should only fail if the role and name are being passed through iframes (and match the element in the parent element), as an example of when we are indirectly asserting for the values. For direct assertion, I think that might be another form of test (based off of what I've seen in integration tests): is that what you're asking for?

lib/checks/landmarks/landmark-is-unique-after.js
```diff
@@ -0,0 +1,20 @@
+const uniqueLandmarks = [];
+return results.filter(currentResult => {
```
WilcoFiers: This solution isn't going to work with scoped runs. This check is assuming that the scope of the run is always the entire page, rather than some element or section on the page. Take this example:

```html
<nav aria-label="foo" id="foo"></nav>
<nav aria-label="foo" id="bar"></nav>
<script> axe.run('#foo') </script>
```

Because `#bar` is out of scope when running axe, it will not be in the `results` array in the after method, and so `#foo` won't be seen as a duplicate. The way to deal with this is to have the `evaluate` method look for duplicates in the page. Have a look at the duplicate-id rule for an example. That also avoids the need of updating relatedNodes in the after method, which while it does work, isn't how this is supposed to work.

General rule of thumb is that when you test the page as a whole, the total number of issues should be the same as when you test separate pieces of the page and concatenate the results.

lib/checks/landmarks/landmark-is-unique.json
```diff
@@ -0,0 +1,16 @@
+{
+	"id": "landmark-is-unique",
+	"evaluate": "landmark-is-unique.js",
+	"after": "landmark-is-unique-after.js",
+	"tags": ["best-practice"],
+	"options": {
```
WilcoFiers: What was the reason for this options object? I don't see this being used anywhere so it looks like this should be removed.

lib/checks/landmarks/landmark-is-unique.json
```diff
@@ -0,0 +1,16 @@
...
+		"selector": "[role=banner], [role=complementary], [role=contentinfo], [role=main], [role=navigation], [role=region], [role=search], [role=form], form, footer, header, aside, main, nav, section"
+	},
+	"metadata": {
+		"impact": "moderate",
+		"messages": {
+			"pass": "Landmarks must have a unique role or role/label combination(aria-label OR aria-labelledby).",
+			"fail": "The landmarks do not have a unique role or role/label combination (use aria-label OR aria-labelledby to make landmarks distinguishable)."
```
WilcoFiers: As before, I'm not sure why this isn't considering title. That certainly seems to work as a way to set the accessible name of a landmark.

test/checks/landmarks/landmark-is-unique-after.js
```diff
@@ -0,0 +1,84 @@
...
+	function createResult(result, data) {
+		return {
+			result: result,
+			data: data
+		};
+	}
+
+	function createResultWithSameRelatedNodes(result, data) {
+		return {
+			...createResult(result, data),
```
WilcoFiers: Babel isn't enabled in our tests, so we need to restrict it to ES3 features. This isn't failing because Appveyor is currently disabled, a fix for that landed recently. There's an issue open to enable Babel on our tests, but for now, use `Object.assign`.

test/checks/landmarks/landmark-is-unique.js
```diff
@@ -0,0 +1,44 @@
+describe('landmark-is-unique', function() {
+	'use strict';
+
+	var checkContext = new axe.testUtils.MockCheckContext();
+	let fixture;
```
WilcoFiers: As above, we can't use ES2015 features in tests. Please use `var` instead of `let` and `const`.

test/rule-matches/landmark-unique-matches.js
```diff
@@ -0,0 +1,142 @@
+describe('landmark-unique-matches', function() {
+	'use strict';
+	let rule;
```
WilcoFiers: As above, tests need to be valid ES3.

lib/rules/landmark-unique-matches.js
```diff
@@ -0,0 +1,41 @@
...
+	if (!role) {
+		return false;
+	}
+
+	const nodeName = actualNode.nodeName.toUpperCase();
+	if (nodeName === 'HEADER' || nodeName === 'FOOTER') {
+		return isHeaderFooterLandmark(virtualNode);
+	}
+
+	if (nodeName === 'SECTION' || nodeName === 'FORM') {
```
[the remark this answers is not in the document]
waabid: I believe in those cases their roles would be form and section since they are explicit (unlike nav in that example, which is implicit).

If my understanding is incorrect, I can definitely make that change.

lib/rules/landmark-unique-matches.js
```diff
@@ -0,0 +1,41 @@
...
+		const label = axe.commons.aria.labelVirtual(virtualNode);
```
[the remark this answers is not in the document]
waabid: My bad: this should actually be using the accessible text API as we have previously discussed.

lib/checks/landmarks/landmark-is-unique-after.js
WilcoFiers: After some extensive discussions, I'm going to withdraw this comment. Dylan and I can't come up with a way to have this rule work across frames, and with limited scope. At least not one that doesn't require substantial changes to the engine.

We think testing landmarks across frames is more important than comparing landmarks outside the scope of the run. So lets go with the proposed solution.

There are other comments to resolve though, so please pick those up, and leave this one for what it is.

test/integration/rules/landmark-unique/landmark-unique-fail.html
```diff
@@ -0,0 +1,29 @@
+<main id="violation-main-1">First main</main>
```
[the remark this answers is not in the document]
waabid: Added.

test/rule-matches/landmark-unique-matches.js
waabid: Done.

test/checks/landmarks/landmark-is-unique.js
waabid: Done.

test/checks/landmarks/landmark-is-unique-after.js
waabid: Done.

lib/checks/landmarks/landmark-is-unique.json
waabid: Updated.

lib/checks/landmarks/landmark-is-unique.json
waabid: Removed.

lib/checks/landmarks/landmark-is-unique.js
```diff
@@ -0,0 +1,7 @@
...
+let label = axe.commons.aria.label(node);
```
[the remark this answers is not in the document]
waabid: Updated.

test/integration/rules/landmark-unique/landmark-unique-fail.html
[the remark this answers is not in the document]
waabid: Added.

lib/checks/landmarks/landmark-is-unique-after.js
waabid: Resolving as such.

lib/rules/landmark-unique.json
```diff
@@ -0,0 +1,13 @@
+{
+	"id": "landmark-unique",
+	"selector": "[role=banner], [role=complementary], [role=contentinfo], [role=main], [role=navigation], [role=region], [role=search], [role=form], form, footer, header, aside, main, nav, section",
+	"tags": ["cat.semantics", "best-practice"],
+	"metadata": {
+		"description": "Ensures landmarks are unique",
+		"help": "Landmarks must have a unique role or role/label combination(aria-label OR aria-labelledby)"
```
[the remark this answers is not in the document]
waabid: Done

dylanb:
@waabid @WilcoFiers how can I help move this PR along?

waabid:
> @waabid @WilcoFiers how can I help move this PR along?

I've addressed all comments; had some tests breaking but fixed those. Now the test build is broken due to a dependency I believe.

WilcoFiers:
Closing this as a duplicate of #1394

WilcoFiers:
Whoops, I meant the other way around.

lib/checks/landmarks/landmark-is-unique.json
```diff
@@ -0,0 +1,14 @@
...
+	"id": "landmark-is-unique",
+	"evaluate": "landmark-is-unique.js",
+	"after": "landmark-is-unique-after.js",
+	"tags": ["best-practice"],
+	"selector": "[role=banner], [role=complementary], [role=contentinfo], [role=main], [role=navigation], [role=region], [role=search], [role=form], form, footer, header, aside, main, nav, section",
+	"metadata": {
+		"impact": "moderate",
+		"messages": {
+			"pass": "Landmarks must have a unique role or role/label/title (i.e. accessible name) combination",
+			"fail": "The landmarks do not have a unique role or role/label/title combination (use aria-label, aria-labelledby, or title to make landmarks distinguishable)"
```
WilcoFiers: We don't need to repeat the requirement here. I 

> "fail": "The landmark must have a unique aria-label, aria-labelledby, or title to make landmarks distinguishable"

lib/checks/landmarks/landmark-is-unique-after.js
```diff
@@ -0,0 +1,20 @@
+var uniqueLandmarks = [];
+return results.filter(currentResult => {
```
WilcoFiers: This file could do with a comment or two explaining the unusual filter logic.

jeankaplansky:
Rule help for 3.3 required.

Issue #96, opened 2015-07-31
New Rule: Scrollable pane not keyboard accessible


dsturley:
Scrollable areas (overflow: auto / scroll) need a method for keyboard-only users to scroll the area.  This can be fixed if:
1.  The scrollable area has focusable elements within it.
2.  The scrollable area is added to tab order.

dylanb:
Any idea how we get this to perform well?

WilcoFiers:
We have a list of scrollable elements that we need for restoreScroll. We may need a way to bypass the selector/matches and use that list instead. We'll also need to check that there is hidden content and that there is visible content as well. This may have to be experimental.

marcysutton:
Perhaps the new `valid-scrollable-semantics` check could be leveraged for this? https://github.com/dequelabs/axe-core/commit/ce5f3dc765dfb2079424d23a94c64d9043c5a803

jkodu:
# Rule: scrollable-keyboard-accessible

The rule checks that scrollable elements have keyboard access.

**Tags:** wcag2a, experimental

## Matcher

Element(s) that can be scrolled.  (See - `function getScroll`)

## Checks

### Check 1
"any": focusable-content
check that element has focusable content

pass - 
fail - 

### Check 2
"any": is-focusable
check that element is focusable

pass - 
fail -

jeankaplansky:
documented in INR - 6/17/2019

jeankaplansky:
https://dequeuniversity.com/rules/axe/3.3/scrollable-region-focusable

jeankaplansky:
DQU and WCAG links included creating MVP rule help

chandana7393:
Tested, working as expected.
![scrollable_keyboard_Attest](https://user-images.githubusercontent.com/45285854/62292510-2ae76000-b484-11e9-8edf-dcec5af683fb.PNG)
![scrollable_keyboard_Axe](https://user-images.githubusercontent.com/45285854/62292512-2cb12380-b484-11e9-851c-de6f687ffa23.PNG)
![scrollable_keyboard_axe_coconut](https://user-images.githubusercontent.com/45285854/62292517-2e7ae700-b484-11e9-9075-fac7de3798a9.PNG)
**Tested Environment:**
Attest - 2.5.1.21115v
Axe - 3.8.1.21115v
Axe-coconut - 3.8.1.21304v
Chrome - 76.0.3809.87v
Firefox - 68.0.1v
OS - Windows 10 64 bit.

PR #1396, opened 2019-03-01
feat(rule): Scrollable region focusable

jkodu:
Rule which checks if scrollable region has keyboard access.

Closes issue: https://github.com/dequelabs/axe-core/issues/96

## Reviewer checks


lib/checks/keyboard/focusable-element.js
```diff
@@ -0,0 +1,3 @@
+const isFocusable = virtualNode.isFocusable;
+
+return isFocusable;
```
straker: Why not just return `virtualNode.isFocusable`?

test/checks/keyboard/focusable-content.js
```diff
@@ -0,0 +1,124 @@
...
+		var params = checkSetup(
+			'<div id="target">' + '<input type="text" tabindex="-1">' + '</div>'
+		);
+		var actual = check.evaluate.apply(checkContext, params);
+		assert.isFalse(actual);
+	});
+
+	it('returns false when parent element is focusable (only checks if contents are focusable)', function() {
+		var params = checkSetup(
+			'<div id="target">' + '<p style="height: 200px;"></p>' + '</div>'
```
straker: Should `div` have a tab index on it to make it focusable as the test suggests?

test/checks/keyboard/focusable-element.js
```diff
@@ -0,0 +1,46 @@
...
+		var actual = check.evaluate.apply(checkContext, params);
+		assert.isTrue(actual);
+	});
+
+	it('returns false when element made not focusable by tabindex', function() {
+		var params = checkSetup(
+			'<input id="target" type="checkbox" tabindex="-1">'
+		);
+		var actual = check.evaluate.apply(checkContext, params);
+		assert.isTrue(actual);
```
straker: Should be `assert.isFalse(actual)`

lib/checks/keyboard/focusable-element.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * Note:
+ * Check
+ * - if element is focusable
+ * - if element is in focus order via `tabindex`
+ */
+const isFocusable = virtualNode.isFocusable;
+
+let tabIndex = virtualNode.actualNode.getAttribute('tabindex');
```
straker: Just a minor suggestion: If you `parseInt` when you access the attribute then you shouldn't have to do it twice in the ternary.

dylanb:
@jkodu who has the ball on this PR?

jkodu:
@dylanb it has been reviewed by @WilcoFiers. I am picking up with the changes requested. Should be ready for another round soon.

lib/core/utils/get-scroll.js
```diff
@@ -0,0 +1,36 @@
...
+ * Get the scroll position of given element
+ * @method getScroll
+ * @memberof axe.utils
+ * @param {Element} node
+ * @param {buffer} (Optional) allowed negligence in overflow
+ * @returns {Object | undefined}
+ */
+axe.utils.getScroll = function getScroll(elm, buffer = 0) {
+	if (
+		elm.scrollHeight < elm.clientHeight ||
```
WilcoFiers: This should be `elm.scrollHeight <= elm.clientHeight + buffer`.
1. Use `<=` instead of `<`. scroll is *never* smaller than client
2. If you add your buffer here, you don't need a second if statement. Also much less confusing about what that buffer is supposed to do.

lib/core/utils/get-scroll.js
```diff
@@ -0,0 +1,36 @@
...
+	if (maxOverflow <= buffer) {
+		return;
+	}
+
+	const style = window.getComputedStyle(elm);
+	const visibleOverflowY = style.getPropertyValue('overflow-y') === 'visible';
+	const visibleOverflowX = style.getPropertyValue('overflow-x') === 'visible';
+
+	// See if the element hides overflowing content
+	if (!visibleOverflowY || !visibleOverflowX) {
```
WilcoFiers: This is buggy. At this point, you know that either X or Y overflows, but you don't know which. So if X overflows, but has `visibleOverflowY: false, visibleOverflowX: true`, you are returning this object when you shouldn't.

lib/rules/scrollable-region-focusable-matches.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * Note:
+ * `excludeHidden=true` for this rule, thus considering only elements in the accessibility tree.
+ */
+const { isHidden, querySelectorAll } = axe.utils;
+const nodeAndDescendents = querySelectorAll(virtualNode, '*');
+
+const hasVisibleChildren = nodeAndDescendents.filter(
```
WilcoFiers: Use `nodeAndDescendents.some()`, that way you're not calling `isVisible` a million times.

lib/rules/scrollable-region-focusable-matches.js
```diff
@@ -0,0 +1,15 @@
...
+	({ actualNode }) => !isHidden(actualNode)
```
WilcoFiers: This isn't going to work. `isHidden` returns false for every empty div. If you want to check if something has content, use something like `dom.hasContentVirtual`, although we may need something like that doesn't look at the accessible name.

lib/rules/scrollable-region-focusable-matches.js
```diff
@@ -0,0 +1,15 @@
...
+);
+if (!hasVisibleChildren) {
+	return false;
+}
+
+return !!axe.utils.getScroll(node, 13); //1em ~= 13px
```
WilcoFiers: getScroll is much faster than checking if the thing has visible content. You should run this first.

lib/core/utils/get-scroll.js
```diff
@@ -0,0 +1,36 @@
+/**
+ * Get the scroll position of given element
+ * @method getScroll
+ * @memberof axe.utils
+ * @param {Element} node
+ * @param {buffer} (Optional) allowed negligence in overflow
+ * @returns {Object | undefined}
+ */
+axe.utils.getScroll = function getScroll(elm, buffer = 0) {
```
WilcoFiers: Proposal:
```js
axe.utils.getScroll = function getScroll(elm, buffer = 0) {
  const overflowX = elm.scrollWidth > elm.clientWidth + buffer // true
  const overflowY = elm.scrollHeight > elm.clientHeight + buffer // false
	if (!overflowX || !overflowY) {
		return;
	}

	const style = window.getComputedStyle(elm);
	const scrollableY = style.getPropertyValue('overflow-y') !== 'visible'; // false
	const scrollableX = style.getPropertyValue('overflow-x') !== 'visible'; // true


	// See if the element hides overflowing content
	if ((overflowX && scrollableX) || (overflowY && scrollableY)) {
		return {
			elm,
			top: elm.scrollTop,
			left: elm.scrollLeft
		};
	}
};
```

lib/rules/scrollable-region-focusable-matches.js
WilcoFiers:
```js
		(ignoreAria || !!aria.labelVirtual(elm)) ||
```

kensgists:
This new rule is causing standard accessible text clipping techniques to fail. For example, this CSS:

`.accessible-text {  
  position:` absolute !important;
  clip: rect(0.0625rem 0.0625rem 0.0625rem 0.0625rem);
  clip: rect(0.0625rem, 0.0625rem, 0.0625rem, 0.0625rem);
  padding: 0 !important;
  border: 0 !important;
  height: 0.0625rem !important;
  width: 0.0625rem !important;
  overflow: hidden;
}`

is marked as a violation in axe-coconut.

@WilcoFiers pointed out above that elements with overflow:hidden might be exempted. I think this is a good idea.

We use axe-coconut for testing because of some of the useful experimental rules. But now we are getting vast numbers of false positives due to this new rule.

PR #1406, opened 2019-03-05
fix: Avoid require conflict with Cypress [#1405]

### LINKED ISSUE #1405: [3.2.0] Cannot find module './lib/axios'
Attempts to update `axe-core` from 3.1.2 to 3.2.0 are causing Cypress tests to fail. From the error message, it looks like it can't resolve `axios` anymore? 

Expectation: no errors

Actual: `Error: Cannot find module './lib/axios' from '/root/project/node_modules/axe-core'`

Motivation: N/A

<pre><code>
axe-core version: 3.2.0
axe-webdriver, extension or other integration version: N/A

For Tooling issues:
- Node version: 10  
- Platform: all 
</code></pre>

Here's a reproduction using `create-react-app` and Cypress (test copied from #755):

https://github.com/wKovacs64/cypress-axe-core-320

(Run `yarn cy:run` to start Cypress in headless mode in your CLI or `yarn cy:open` to launch the GUI.)

Thanks.

WilcoFiers:
Use derequire the rename all uses of `require` to `_dereq_` to avoid the problem of Cypress overloading the `require` method.

Closes #1405

## Reviewer checks

### DIFF
--- a/build/imports-generator.js
+++ b/build/imports-generator.js
@@ -1,5 +1,6 @@
 const path = require('path');
 const browserify = require('browserify');
+const derequire = require('derequire');
 const createFile = require('./shared/create-file');
 
 const inputFile = path.join(
@@ -25,6 +26,8 @@ async function run() {
 			throw new Error('Cannot browserify axe.imports', err);
 		}
 		try {
+			// Replace `require` calls with `_dereq_` in order not to confuse Cypress.js
+			result = derequire(result);
 			await createFile(outputFile, result);
 		} catch (error) {
 			throw new Error('Cannot write browserify generated axe.imports', error);
--- a/package.json
+++ b/package.json
@@ -85,6 +85,7 @@
 		"chai": "~4.1.2",
 		"clone": "~2.1.1",
 		"css-selector-parser": "^1.3.0",
+		"derequire": "^2.0.6",
 		"dot": "~1.1.2",
 		"emoji-regex": "7.0.3",
 		"es6-promise": "^4.2.6",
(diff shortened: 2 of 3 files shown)


stephenmathieson:
This seems scary. I'd like to hear more about why this is necessary before OK'ing the code.

WilcoFiers:
@stephenmathieson Cypress seems to be replacing any variable with the name `require` with its own require method. I didn't dig into exactly why that is. I searched, but couldn't find any documentation about this. Renaming all uses of `require` to `_dereq_` gets around that problem.

What is your concern with doing this? I realise that if derequire messes up, this could go wrong, but the same is true for browserify. That's why I used derequire, it's made to do this.

stephenmathieson:
> Cypress seems to be replacing any variable with the name require with its own require method. I didn't dig into exactly why that is. I searched, but couldn't find any documentation about this.

Wow, OK. That's very strange. I wonder why that is 🤔 

> What is your concern with doing this?

I'm just worried about adding _another_ layer of transformation to the codebase, especially since in this PR we're transforming code that we don't own/maintain (`dot`, `axios`, and `es-promise`). It's generally recommended that you **do not** run Babel/etc over your `node_modules`, so I'm of the same mindset with the "derequire" transformation.

However, since this is urgent, and the tests are all green, let's give this a go 😄

stephenmathieson:
> However, since this is urgent, and the tests are all green, let's give this a go 😄

I was wrong. Tests are now failing.

stephenmathieson:
Based on #1427, it seems like using axe-core in Cypress still isn't working as expected.

artemkochnev:
@stephenmathieson I just want to add that the issue I'm seeing only exists when using webpack to preprocess files. If I let Cypress manage things itself, using axe-core works without issue.

YOU54F:
This is an issue with webpack, I have switched to 

https://github.com/cypress-io/cypress-browserify-preprocessor

which has resolved the issue

PR #1408, opened 2019-03-05
chore: Avoid publishing unnecessary files to NPM

WilcoFiers:
Speaks for itself.

Closes issue: none

## Reviewer checks


WilcoFiers:
@stephenmathieson those files are only used for development. This is just removing clutter.

stephenmathieson:
Right. I'd call it a "chore" then, since it doesn't fix a bug.

WilcoFiers:
@stephenmathieson updated.

scurker:
Just a thought: Would it be better to use a [whitelist](https://docs.npmjs.com/files/package.json#files) instead of a blacklist?

stephenmathieson:
@scurker yes, probably, but the path of least resistance is just adding these few entries to the existing `.npmignore`.

PR #1415, opened 2019-03-07
fix: Avoid "screen is not defined" error [#1404]

### LINKED ISSUE #1404: screen is not defined
Expectation: `axe.run` should run without throwing errors

Actual: `axe.run` throws

```
ReferenceError: screen is not defined
      at Object.getEnvironmentData (node_modules/axe-core/axe.js:3694:39)
      at /Users/arobert3/Sites/verso/node_modules/axe-core/axe.js:3880:35
      at /Users/arobert3/Sites/verso/node_modules/axe-core/axe.js:3656:23
      at /Users/arobert3/Sites/verso/node_modules/axe-core/axe.js:3549:11
      at /Users/arobert3/Sites/verso/node_modules/axe-core/axe.js:5992:13
      at /Users/arobert3/Sites/verso/node_modules/axe-core/axe.js:2637:9
      at /Users/arobert3/Sites/verso/node_modules/axe-core/axe.js:5992:13
      at /Users/arobert3/Sites/verso/node_modules/axe-core/axe.js:5992:13
      at rule.run.Object.assign.result (node_modules/axe-core/axe.js:2581:9)
      at /Users/arobert3/Sites/verso/node_modules/axe-core/axe.js:3110:14
      at Timeout._onTimeout (node_modules/axe-core/axe.js:5992:13)
```

Motivation: I want to upgrade my version of `axe-core`

<pre><code>
axe-core version: 3.2.0

For Tooling issues:
- Node version: 8.13.0
- Platform:  Mac
</code></pre>

WilcoFiers:
`helpers.getEnvironmentData` made the assumption that `window` is the name of the global object. This is not the case when running axe-core in a NodeJS environment such as when testing axe-core using JSDOM. Usually all props on window get copied over to the global to avoid this, but especially for something as uncommon as `screen` it is not surprising that this broke somewhere.

Closes #1404

## Reviewer checks

### DIFF
--- a/lib/core/reporters/helpers/get-environment-data.js
+++ b/lib/core/reporters/helpers/get-environment-data.js
@@ -5,6 +5,8 @@
  * @return {Object}
  */
 helpers.getEnvironmentData = function getEnvironmentData(win = window) {
+	// TODO: remove parameter once we are testing axe-core in jsdom and other
+	// supported environments
 	const {
 		screen = {},
 		navigator = {},
--- a/lib/core/reporters/helpers/get-environment-data.js
+++ b/lib/core/reporters/helpers/get-environment-data.js
@@ -4,12 +4,17 @@
  * Add information about the environment axe was run in.
  * @return {Object}
  */
-helpers.getEnvironmentData = function getEnvironmentData() {
-	'use strict';
-	var orientation = window.screen
-		? screen.msOrientation ||
-		  (screen.orientation || screen.mozOrientation || {})
-		: {};
+helpers.getEnvironmentData = function getEnvironmentData(win = window) {
+	const {
+		screen = {},
+		navigator = {},
+		location = {},
+		innerHeight,
+		innerWidth
+	} = win;
+
+	const orientation =
+		screen.msOrientation || screen.orientation || screen.mozOrientation || {};
 
 	return {
 		testEngine: {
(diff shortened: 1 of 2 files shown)


stephenmathieson:
Another thing - I do not believe this fixes #1404, as [we _always_ declare a `window`variable](https://github.com/dequelabs/axe-core/blob/develop/lib/intro.stub#L12).

WilcoFiers:
@stephenmathieson 

Regarding your first point. Yes this isn't an air tight test for the problem. The only way I could test that is in an environment where `window` isn't the global object. Short of setting up an entirely new testing environment, I don't think I can solve that. So I picked the closest thing to a test that I could write.

Point 2, `window` is one of the variables that axe-core requires to have in its scope when it is loaded. There are a few others too. This is already a requirement and is something we can safely assume. Requiring `screen` as a global variable was an oversight, and would be a breaking change if we added that.

I'll create issues for both of these points. I do not think this is something we need to resolve as part of this PR.

stephenmathieson:
@WilcoFiers I disagree that we don't need a proper test here, otherwise how can we be sure this fixes the bug? I do not believe this fixes the problem, and I think there's something else at play here.

Here's my reasoning:

I ran the relevant "orientation" code (copied from `axe-core`) in a non-browser environment (node.js):

```js
var orientation = window.screen
  ? screen.msOrientation ||
    (screen.orientation || screen.mozOrientation || {})
  : {};

console.log({ orientation })
```

As expected, this threw `ReferenceError: window is not defined`.

Now defining `window` in scope (as `axe-core` does) like this:

```js
var window = {}

var orientation = window.screen
  ? screen.msOrientation ||
    (screen.orientation || screen.mozOrientation || {})
  : {};

console.log({ orientation })
```

yields:

```
{ orientation: {} }
```

Note that there is no `screen` global available, nor is there a `screen` property on the `window` object. This tells me there's something else happening in the non-browser environment the #1404 problem is occurring in.

straker:
I added the comments as we discussed

Issue #1424, opened 2019-03-11
document-title rule should not fail framed pages


WilcoFiers:
Got a report of axe-core calling out the content of the following iframe:

```html
<iframe tabindex="-1" role="presentation" src="..."></iframe>
```

Axe-core shouldn't fail iframes for SC3.1.1. It isn't applicable to framed pages, only to the top window. While it might be true that such pages can be displayed without the frame, they aren't on the page that is tested, so I don't think axe-core should ever report them.

WilcoFiers:
There's an easy fix for this. We can add the `window-is-top.js` matcher to the document-title rule.

jkodu:
@WilcoFiers 

I believe you intended to say rule `html-has-lang (sc311)` and not `document-title (sc242)`.

PR #1430, opened 2019-03-13
fix: exclude  iframe for html-has-lang rule (Issue 1424)

jkodu:
Exclude `iframe` elements for SC311 rule `html-has-lang`.

Note: `context` integration tests had to be updated to not use `html-has-lang`.

Closes issue: https://github.com/dequelabs/axe-core/issues/1424

## Reviewer checks


jkodu:
@WilcoFiers 

Before I fix up the `context` tests, wanted to discuss, if perhaps ignoring `iframe` with certain roles, would be a better matches than `window-is-top`

test/integration/full/html-has-lang/html-has-lang-pass.html
```diff
@@ -16,10 +17,15 @@
 		var assert = chai.assert;
 	</script>
 </head>
+
 <body>
-	<iframe id="frame1" src="frames/level1.html"></iframe>
+	<!-- 
+			Note: 
+			This rule does not include `iframe` uses matches "window-is-top.js"
```
WilcoFiers: Yeah, but you do actually need to test that. Don't remove the frames from the examples.

test/integration/full/html-has-lang/html-has-lang-fail.html
```diff
@@ -16,10 +17,15 @@
 		var assert = chai.assert;
 	</script>
 </head>
+
 <body>
-	<iframe id="frame1" src="frames/level1.html"></iframe>
+	<!-- 
+			Note: 
+			This rule does not include `iframe` uses matches "window-is-top.js" 
+		-->
```
WilcoFiers: You're not testing that iframes without `html[lang]` don't fail.

Issue #1397, opened 2019-03-01
TypeError: Cannot read property 'indexOf' of undefined


patrickhulce:
Run axe on https://www.homeyou.com/

Actual: Fails immediately with 
```
TypeError: Cannot read property 'indexOf' of undefined
    at te (chrome-extension://<id>/adapter.bundle.js:76:59210)
    at Array.filter (<anonymous>)
    at a (chrome-extension://<id>/adapter.bundle.js:76:61464)
    at Object.H.utils.getSelectorData (chrome-extension://<id>/adapter.bundle.js:76:61665)
    at Object.q [as _runRules] (chrome-extension://<id>/adapter.bundle.js:76:35171)
    at Object.H.run (chrome-extension://<id>/adapter.bundle.js:76:37333)
    at Object.analyze (chrome-extension://<id>/adapter.bundle.js:99:285715)
    at chrome-extension://<id>/adapter.bundle.js:99:284178
    at onMessage (chrome-extension://<id>/browser-polyfill.js:1056:22)
```

Expectation: Run completes without error

Motivation: https://github.com/GoogleChrome/lighthouse/issues/7355

<pre><code>
axe-core version: 3.1.2
axe-webdriver, extension or other integration version: 3.7.0

Browser and Assistive Technology versions
Chrome 72.0.3626.119
</code></pre>

WilcoFiers:
Howdy! Thank you. I've been aware this existed for a while now but never found a place to reproduce it. We'll be sure to have a look.

straker:
So the issue is caused by one of the form elements on the page clobbering the DOM and overriding `node.attributes`. 

```html
<form>
   <select name="attributes">
</form
```

This causes `node.attributes` to return the `<select>` element instead of a `NamedNodeMap`. To fix this we just have to check that `node.attributes` is an instanceof `NamedNodeMap` before we loop over it.

PR #1432, opened 2019-03-13
fix: Prevent error when using <select name="attributes">

straker:
Any time we access `node.attributes` we should check that it returns a `NameNodeMap` as expected and that the DOM isn't clobbered. 

Closes: #1397 

## Reviewer checks


lib/checks/aria/allowed-attr.js
```diff
@@ -6,7 +6,7 @@ var attr,
 	attrName,
 	allowed,
 	role = node.getAttribute('role'),
-	attrs = node.attributes;
+	attrs = axe.utils.getNodeAttributes(node);
```
jkodu: suggestion:
if we perhaps make `attributes` a getter in `virtualNode`, it does look seamless to access the property, same as what we have done for `isFocusable` or `tabbableElements`.

`node.attributes` can become `virtualNode.attributes`

jkodu: alternatively, add a custom eslint rule, which will stop people from using node.attributes, because it is hard to remember that a util method exists.

Something like this could help - https://github.com/dequelabs/axe-core/blob/develop/.eslintrc#L71

straker: Nice, I like that.

straker: I think we should hold off on that for now since we're still figuring out what we want to put in the virtual node.

stephenmathieson: I'm not sure about this. Adding an ESLint rule which prevents accessing standard DOM properties seems a bit arbitrary.

There are two major reasons I'm against it:

1. the rule will confuse people who aren't aware of it, thus making `axe-core` an even more difficult project to contribute to
2. it promotes adding more rules like it, which will make the first point even worse

Before we build a new rule, can we come to a consensus here and make sure it's the right decision?

straker: I don't know. If the result of accessing the standard DOM property is potential bugs, I see that fitting under the realm of eslint.

stephenmathieson: I understand that it'll potentially prevent bugs, and that's why I'm not saying "no". I just want be sure it's worth the risk.

Can we be sure that this rule doesn't create false positives? If not, I'm of the opinion this will add *a lot* of confusion down the line.

What if a future developer does this:

```js
class NodeImpl {
  attributes = 7
}

var node = new NodeImpl()
var atrrs = node.attributes
```

Will the ESLint rule fail their build?

563712793:
我也不知道怎么来评论这个项目！

lib/checks/aria/allowed-attr.js
straker: To help with user understanding, we can add a comment above the eslint rule that links back to this pr:

```js
{
  // node.attributes can be clobbered so is unsafe to use
  // @see https://github.com/dequelabs/axe-core/pull/1432 
  "selector": "MemberExpression[object.name=node][property.name=attributes]",
  "message": "Don't use node.attributes use node.hasAttirbutes() or axe.utils.getNodeAttributes(node) instead."
}
``

Comment (stephenmathieson) [reply lib/checks/aria/allowed-attr.js]: As discussed on our call, I'm OK with adding the ESLint thing along with the comment. If it becomes an issue in the future, we can revisit it then. Enough yak hair has been removed here.

File: lib/core/utils/get-node-attributes.js [hunk 2]
```diff
@@ -0,0 +1,21 @@
+/* global axe */
+
+/**
+ * Return the list of attributes of a node.
+ * @method getNodeAttributes
+ * @memberof axe.utils
+ * @param {Element} node
+ * @returns {NamedNodeMap}
+ */
+axe.utils.getNodeAttributes = function getNodeAttributes(node) {
```
WilcoFiers: I think this should be on `axe.commons.dom`.

lib/core/utils/get-node-attributes.js
```diff
@@ -0,0 +1,21 @@
...
+	// eslint-disable-next-line no-restricted-syntax
+	if (node.attributes instanceof NamedNodeMap) {
```
WilcoFiers: Can you get this from `window.NamedNodeMap` instead? Have this not be a new global variable we require?

straker: Good point. Going to be a hard habit for me to break.

PR #1437, opened 2019-03-14
chore: test axe examples in ci

scurker:
Get examples in `doc/examples` running on CI:

- [x] ~~Chrome Debugging Protocol~~ _(needs to spin up separate instance of chrome)_
- [x] Jasmine ~~(Existing Issue on CircleCI/Travis [#295](https://github.com/gruntjs/grunt-contrib-jasmine/issues/295))~~ _Swapped to Karma to run Jasmine specs_
- [x] Jest/React
- [x] Mocha
- [x] PhantomJS
- [x] Puppeteer
- [x] QUnit

Added script to run all examples under `doc/examples/*`. Added appveyor config to have examples running on windows. 

## Reviewer checks


doc/examples/jest_react/package.json
```diff
@@ -13,7 +13,7 @@
 		"test": "jest"
 	},
 	"devDependencies": {
-		"axe-core": "^3.1.1",
+		"axe-core": "^3.2.2",
```
stephenmathieson: Keeping this up-to-date seems tedious. Since we're already refactoring this a bit, would a `*` or `file://` be better?

doc/examples/test-examples.js
```diff
@@ -0,0 +1,36 @@
+const { readdirSync, statSync } = require('fs');
+const { join } = require('path');
+const { spawn } = require('child_process');
+const exampleDirs = readdirSync(__dirname)
+	.map(dir => join(__dirname, dir))
+	.filter(dir => statSync(dir).isDirectory());
+
+async function runner(dir) {
+	let child = spawn('npm install && npm test', {
```
stephenmathieson: IIRC `child_process.spawn` has some gnarly Windows issues. Maybe use [`execa`](https://github.com/sindresorhus/execa) instead? It'd also give us the benefit of removing the below Promise wrapper.

doc/examples/test-examples.js
```diff
@@ -0,0 +1,36 @@
...
+				resolve();
+			} else {
+				reject();
+			}
+		});
+	});
+}
+
+(async () => {
+	await Promise.all(exampleDirs.map(runner))
```
stephenmathieson: The outer closure and this `await` are unnecessary. Just do:

```js
Promise.all(exampleDirs.map(runner)).catch(err => {
  console.error(err)
  process.exit(1)
})
```

doc/examples/test-examples.js
```diff
@@ -0,0 +1,36 @@
...
+		.then(() => {
+			// Return successful exit
+			process.exit();
+		})
+		.catch(err => {
+			console.log(err);
```
stephenmathieson: Errors belong on `stderr`, not `stdout`.

doc/examples/test-examples.js
```diff
@@ -0,0 +1,36 @@
+const { readdirSync, statSync } = require('fs');
+const { join } = require('path');
+const { spawn } = require('child_process');
+const exampleDirs = readdirSync(__dirname)
+	.map(dir => join(__dirname, dir))
+	.filter(dir => statSync(dir).isDirectory());
+
+async function runner(dir) {
```
stephenmathieson: Please remove the `async` keyword here, since `await` is not used in this function.

doc/examples/jasmine/karma.conf.js
```diff
@@ -0,0 +1,51 @@
+// Karma configuration
+module.exports = function(config) {
+	config.set({
+		// base path that will be used to resolve all patterns (eg. files, exclude)
```
stephenmathieson: Can we remove the comments here? IMO they're only helpful for people setting up Karma (you), not for people reading this code (us).

appveyor.yml
```diff
@@ -0,0 +1,20 @@
+version: 1.0.{build}-{branch}
```
stephenmathieson: Can we make this change separately? Adding AppVeyor should be in its own PR/commit.

.circleci/config.yml
```diff
@@ -92,11 +99,13 @@ workflows:
           requires:
             - dependencies
             - lint
+      - test_examples
```
stephenmathieson: We should wait until the `dependencies`, `lint` and `test` jobs have finished successfully before running the examples.

We want to "fail fast" on CI so we're not blocking other builds unnecessarily.

This should look something like:

```yaml
- test_examples
    requires:
      - test
```

doc/examples/jest_react/package.json
scurker: It seems since these are primarily intended to be examples to be consumed, neither `*` or `file:...` seem appropriate if we want to promote best practices.

doc/examples/test-examples.js
scurker: I made sure there wasn't any windows issues, but it looks like we're using `execa` elsewhere so I can switch to that.

doc/examples/test-examples.js
scurker: 👍

doc/examples/test-examples.js
scurker: 👍

doc/examples/test-examples.js
scurker: 👍

.circleci/config.yml
scurker: Good call!

appveyor.yml
scurker: I wanted to ensure that appveyor was running successfully on this pr, but this can be implemented on another pr.

doc/examples/jest_react/package.json
stephenmathieson: Sounds good.

doc/examples/test-examples.js
```diff
@@ -0,0 +1,34 @@
+const { readdirSync, statSync } = require('fs');
+const { join } = require('path');
+const execa = require('execa');
+const exampleDirs = readdirSync(__dirname)
+	.map(dir => join(__dirname, dir))
+	.filter(dir => statSync(dir).isDirectory());
+
+function runner(dir) {
+	let child = execa('npm install && npm test', {
```
stephenmathieson: `execa` works a bit differently that `child_process.spawn`. This should probably be something like:

```js
function runner(dir) {
  const install = execa('npm', [ 'install' ], { cwd: dir })
  return install.then(()=> execa('npm', ['test'], { cwd: dir })
}
```

doc/examples/test-examples.js
```diff
@@ -0,0 +1,34 @@
...
+		cwd: dir,
+		shell: true,
+		stdio: 'inherit'
+	});
+
+	return new Promise((resolve, reject) => {
```
stephenmathieson: This is no longer necessary since `execa` returns a Promise.

stephenmathieson:
It's probably not a big deal, but it might be nice to cache the dependencies of the examples in CircleCI. This would speed the builds up a bit.

doc/examples/test-examples.js
scurker: 👍 That's on me for not looking at the documentation intelligently...

doc/examples/test-examples.js
```diff
@@ -0,0 +1,25 @@
...
+	.map(dir => join(__dirname, dir))
+	.filter(dir => statSync(dir).isDirectory());
+
+async function runner(dir) {
+	const config = { cwd: dir, stdio: 'inherit' };
+
+	let install = await execa.shell('npm install', config);
+	let test = await execa.shell('npm test', config);
+
+	return Promise.all([install, test]);
```
stephenmathieson: Since the `install` and `test` promises have already resolved, we don't need the `Promise.all` here 😄

scurker: Yeah, I realized that after I pushed. Sorry about that!

PR #1442, opened 2019-03-18
test: createHTMLDocument needs mandatory title when invoked in IE

jkodu:
fix breaking test in IE.
Reference - https://developer.mozilla.org/en-US/docs/Web/API/DOMImplementation/createHTMLDocument

Closes issue: NA

## Reviewer checks

### DIFF
--- a/test/checks/mobile/css-orientation-lock.js
+++ b/test/checks/mobile/css-orientation-lock.js
@@ -3,7 +3,9 @@ describe('css-orientation-lock tests', function() {
 
 	var checkContext = axe.testUtils.MockCheckContext();
 	var origCheck = checks['css-orientation-lock'];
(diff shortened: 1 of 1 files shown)


test/checks/mobile/css-orientation-lock.js
```diff
@@ -3,7 +3,7 @@ describe('css-orientation-lock tests', function() {
 
 	var checkContext = axe.testUtils.MockCheckContext();
 	var origCheck = checks['css-orientation-lock'];
-	var dynamicDoc = document.implementation.createHTMLDocument();
+	var dynamicDoc = document.implementation.createHTMLDocument('ie is dumb');
```
stephenmathieson: Let's use something else for the page title. How about something that describes the page like "CSS Orientation Lock Tests"?

Issue #1301, opened 2019-01-03
New rule: Avoid inline spacing


WilcoFiers:
This is a rule for https://www.w3.org/TR/WCAG21/#text-spacing. Fail any element with a style attribute that sets line-height, letter-spacing or word-spacing using `!important`. Since this is the highest priority, custom stylesheets can't override these styles, meaning that SC1.4.12 fails.

TODO: Write up a proper proposal.

jkodu:
# avoid-inline-spacing

The rule should fail any element with a style attribute that sets `line-height`, `letter-spacing` or `word-spacing` using `!important`.

**Tags:** wcag21, wcag1412

## Matcher

Element(s) to be tested is deciphered at run-time based on CSSOM and or styles applied.

## Checks

Check that no `!important` style overrides are specified for `line-height`, `letter-spacing` or `word-spacing` properties.

**Pass**
If there are no `!important` styles applied for `line-height`, `letter-spacing` or `word-spacing`.

**Fail**
If there are `!important` styles applied for `line-height`, `letter-spacing` or `word-spacing`.

**Needs Review**
??

jeankaplansky:
documented in INR - 6/17/2019

jeankaplansky:
https://dequeuniversity.com/rules/axe/3.3/avoid-inline-spacing

jeankaplansky:
DQU and Other resources added for MVP.

chandana7393:
Tested, only in Axe extension able to see the violations related to inline spacing but in Attest and Axe-coconut unable to see the violations related to the inline spacing rule.
![no_violations_inline_Attest](https://user-images.githubusercontent.com/45285854/62291429-911eb380-b481-11e9-9a4b-fba59a66eb0d.PNG)
![no_violations_inline_axe_coconut](https://user-images.githubusercontent.com/45285854/62291434-92e87700-b481-11e9-94cb-52e60f838e97.PNG)
**Tested Environment:**
Attest - 2.5.1.21115v
Axe - 3.8.1.21115v
Axe-coconut - 3.8.1.21304v
Chrome - 76.0.3809.87v
Firefox - 68.0.1v
OS - Windows 10 64 bit.

jeankaplansky:
Closing - See #1698.

PR #1446, opened 2019-03-19
feat(rule): Inline text spacing must be adjustable with custom stylesheets

jkodu:
Ensure text spacing is not affected by inline spacing styles that affects CSS specificity

Closes issue: 
- https://github.com/dequelabs/axe-core/issues/1301

**Note: to make sure to make it clear in the failure message that this is due to limited support for custom stylesheets.**

## Reviewer checks


lib/checks/shared/avoid-inline-spacing.js
```diff
@@ -0,0 +1,15 @@
...
+	'letter-spacing',
+	'word-spacing'
+];
+
+const hasImportantPriority = inlineSpacingCssProperties.some(
+	property => node.style.getPropertyPriority(property) === `important`
+);
+
+if (hasImportantPriority) {
+	return false; //-> fail
```
stephenmathieson: Is the "fail" comment necessary here? IIRC `return false` is how _all_ checks "fail".

jkodu:
@straker as you have approved this, you will have to fill out the review checks above.

lib/checks/shared/avoid-inline-spacing.json
```diff
@@ -0,0 +1,11 @@
+{
+	"id": "avoid-inline-spacing",
+	"evaluate": "avoid-inline-spacing.js",
+	"metadata": {
+		"impact": "serious",
+		"messages": {
+			"pass": "No inline styles that affect text spacing has been specified",
```
WilcoFiers: Should be `No inline styles with "!important" ...`

lib/checks/shared/avoid-inline-spacing.json
```diff
@@ -0,0 +1,11 @@
...
+			"fail": "Inline styles that affect text spacing has been specified"
```
WilcoFiers: You'll need to include what style props are failing. Something like:

> Remove `!important` from inline styles {= data =}, as overriding this not supported by most browsers

lib/rules/avoid-inline-spacing.json
```diff
@@ -0,0 +1,12 @@
+{
+	"id": "avoid-inline-spacing",
+	"selector": "[style]",
+	"tags": ["wcag21", "wcag1412"],
+	"metadata": {
+		"description": "Ensure text spacing is not affected by inline spacing styles that affects CSS specificity",
+		"help": "No inline styles that affect text spacing should been specified"
```
WilcoFiers: As above, the key here is "!important".

test/checks/shared/avoid-inline-spacing.js
```diff
@@ -0,0 +1,83 @@
+describe('avoid-inline-spacing tests', function() {
```
WilcoFiers: Should test when an inline style has an invalid value, like `line-height: 3foobar !important;`

lib/checks/shared/avoid-inline-spacing.json
jkodu: I don't think `as overriding this not supported by most browsers` reads right here. I have done the below.

`Remove '!important' from inline style{{=it.data && it.data.length > 1 ? 's' : ''}} {{=it.data.join(', ')}} that affect text spacing`

lib/checks/shared/avoid-inline-spacing.js
```diff
@@ -0,0 +1,19 @@
...
+
+const overriddenProperties = inlineSpacingCssProperties.filter(property => {
+	if (node.style.getPropertyPriority(property) === `important`) {
+		return property;
+	}
+});
+
+if (overriddenProperties.length > 0) {
+	this.data(overriddenProperties);
+	console.log(overriddenProperties);
```
WilcoFiers: Console log?

lib/rules/avoid-inline-spacing.json
```diff
@@ -0,0 +1,12 @@
+{
+	"id": "avoid-inline-spacing",
+	"selector": "[style]",
+	"tags": ["wcag21", "wcag1412"],
+	"metadata": {
+		"description": "Ensure text spacing is not affected by inline spacing styles that affects CSS specificity",
+		"help": "No inline styles with '!important' that affect text spacing should been specified"
```
WilcoFiers: I'm not sure about this "should be specified". I don't think these texts describe the problem well. The point is not to prevent custom stylesheets from overriding whitespace, and the reason that I want to include this "not supported in many browsers" thing somewhere is because this is an accessibility support problem. Our messaging needs to make it clear that this is a support problem, and not something thats inherently inaccessible.

lib/rules/avoid-inline-spacing.json
```diff
@@ -0,0 +1,12 @@
+{
+	"id": "avoid-inline-spacing",
+	"selector": "[style]",
+	"tags": ["wcag21", "wcag1412"],
+	"metadata": {
+		"description": "Ensure text spacing is not affected by inline spacing styles that affects CSS specificity",
+		"help": "Remove inline styles with '!important' that affect text spacing, as overriding this is not supported by most browsers"
```
WilcoFiers: Did a little more editing. I think the pass / fail ones are good. Can you change help/description to:


Description: Ensure that test spacing set through style attributes can be adjusted with custom stylesheets
Help: Inline text spacing must be adjustable with custom stylesheets

The browser thing is part of the fail message, we don't need to repeat that.

PR #1448, opened 2019-03-20
fix: role=radio should not require aria-checked

jkodu:
As per ARIA spec, when an element has `role=radio`, states and properties are derived from the role and do not have be explicitly mentioned.

See - https://www.w3.org/TR/wai-aria-1.1/#radio for more info.

Closes issue:
- NA

## Reviewer checks


test/integration/rules/aria-required-attr/required-attr.html
```diff
@@ -3,6 +3,7 @@
 <div role="spinbutton" id="pass3" aria-valuenow="value" aria-valuemax="value" aria-valuemin="value">ok</div>
 <div role="heading" id="pass4" aria-level="1">ok</div>
 <div role="combobox" id="pass5" aria-expanded="true">ok</div>
+<span role="radio" id="pass6">I am BLUE!</span>
```
WilcoFiers: Add a check to show it doesn't fail when you give it `aria-checked` please.

WilcoFiers: ... in the rule that checks allowed attributes.

jkodu:
@WilcoFiers - wai tools need this, so a review/ approval can be handy.

aweary:
> As per ARIA spec, when an element has role=radio, states and properties are derived from the role and do not have be explicitly mentioned.

If I'm looking at [the spec](https://www.w3.org/TR/wai-aria-1.1/#radio) correctly, `aria-checked` is marked as a required property, not inherited. The spec says that this identifies:

> States and properties specifically required for the role and subclass roles. Content authors MUST provide a non-empty value for required states and properties.

@WilcoFiers @jkodu it seems like this should still be marked as required?

Issue #1449, opened 2019-03-20
New Rule: aria-form-field-label


jkodu:
**Some Context:**
This is a composite rule that will use some pre-existing checks, to extend the current `label` rule for a larger set of semantic roles (checkbox, combobox, listbox, menuitemcheckbox, menuitemradio, radio, searchbox, slider, spinbutton, switch, textbox)

# aria-form-field-label

Ensures every element with semantic roles of a form field (checkbox, combobox (select elements), listbox, menuitemcheckbox, menuitemradio, radio, searchbox, slider, spinbutton, switch, textbox) has a label.

**Tags:** wcag2a, wcag412

## Matches
- all of the above semantic roles excluding ones covered by `label` rule & or any overlapping rules like `area-alt`, `button-name`, `image-alt`.

## Checks

**any:** 
- `aria-label` (Already exists) 
- `aria-labelledby`(Already exists) 
- `non-empty-title` (Already exists) 

**none:**
- `no-implicit-explicit-label`
	- "fail": "Element with ARIA role ... must have label",
	- "incomplete": "Element with ARIA role ... cannot have implicit or explicit label"

### Related issues
- https://github.com/dequelabs/axe-core/issues/1148
- https://github.com/dequelabs/axe-core/issues/1157

dylanb:
we should call this `aria-input-field-name`

jkodu:
@dylanb 

This got split into 2 rules
- `aria-toggle-field-label` and
- `aria-input-field-label`

Do you want to change both of them to below?
- `aria-toggle-field-name`
- `aria-input-field-name`

@WilcoFiers - what do you think?

dylanb:
yes, name and this should probably happen before we ship 3.3

jkodu:
@dylanb - as the PR's (https://github.com/dequelabs/axe-core/pull/1450 & https://github.com/dequelabs/axe-core/pull/1610) have been merged, I am closing this issue. If you believe the rules have to be renamed, please reopen the issue & notify.

dylanb:
Yes, i think we should rename these before releasing 3.3

PR #1450, opened 2019-03-20
feat(rule): New aria-toggle-field-label rule

jkodu:
Ensures elements with one of the below semantic roles has a label.
- checkbox
- menuitemcheckbox
- menuitemradio
- radio
- switch
- textbox

Closes issue:
- https://github.com/dequelabs/axe-core/issues/1449

## Reviewer checks


jkodu:
@WilcoFiers - I have some open questions on this one, so worth a discussion.

test/integration/rules/form-field-label/form-field-label.html
```diff
@@ -0,0 +1,40 @@
+<!-- Inapplicable -->
+<input id='inapplicable1' role="presentation" />
```
jkodu: 1. remove role, they should be inapplicable irrespective

2. add textarea tests

lib/rules/form-field-label.json
```diff
@@ -0,0 +1,12 @@
...
+	"id": "form-field-label",
+	"matches": "form-field-label-matches.js",
+	"tags": ["wcag2a", "wcag412"],
+	"metadata": {
+		"description": "Ensures every element with semantic role that designates as a form element has a label",
+		"help": "Form elements must have labels"
+	},
+	"all": [],
+	"any": ["aria-label", "aria-labelledby", "non-empty-title"],
+	"none": ["help-same-as-label", "hidden-explicit-label"]
```
jkodu: - remove none checks

- create a new check, that raises a needs review (for implicit labels & explicit labels) when there is acc name, otherwise fail (work on good pass & fail msgs).

```
<label>
  foo
  <div role="textbox" aria-label="foo"><div>
</label>
```

jkodu: @WilcoFiers see proposal for new check `no-implicit-explicit-label` - https://github.com/dequelabs/axe-core/issues/1449`

lib/checks/aria/no-implicit-explicit-label.js
```diff
@@ -0,0 +1,10 @@
+const { aria } = axe.commons;
+
+const role = aria.getRole(node, { noImplicit: true });
+this.data(role);
+
+if (aria.isImplicitLabel(virtualNode) || aria.isExplicitLabel(node)) {
+	return undefined;
+}
+
+return false;
```
WilcoFiers: You should use the `text.labelText` method to check for label elements. That way you don't have to create new commons, and you are using the same method that `accessibleText` has.

I also think this needs to check that the label text isn't included in the accessible text, so that something like this passes:

```html
<label>
  foo <div role="textbox" aria-label="Foo field">
</label>
```

```html
<label id="foo">
  foo <div role="textbox" aria-labelleby="foo">
</label>
```

lib/checks/aria/no-implicit-explicit-label.json
```diff
@@ -0,0 +1,11 @@
+{
+	"id": "no-implicit-explicit-label",
+	"evaluate": "no-implicit-explicit-label.js",
+	"metadata": {
+		"impact": "moderate",
+		"messages": {
+			"fail": "Element with ARIA role {{=it.data}} must have label",
```
WilcoFiers: This check should never call out a failure. This rules come back either passed, or with one of the following two messages:

Failed:
```md
Fix at least one of the following:
- ... aria-labelledby ...
- ... aria-label ...
- ... non-empty title ...
```

Incomplete:
```md
Fix the following:
- Check that the <label> does not need be part of the ARIA {=it.data=} field's name 
```

lib/commons/aria/is-explicit-label.js
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+/**
+ * Check if given element has `explicit` label
+ * @method isExplicitLabel
+ * @memberof axe.commons.aria
+ * @param {HTMLElement} node element
+ * @return {Boolean}
+ */
+aria.isExplicitLabel = function(node) {
```
WilcoFiers: I don't think you should pull this into a commons. See my comment about using `text.textLabel`.

lib/commons/aria/is-implicit-label.js
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+/**
+ * Check if given element has `implicit` label
+ * @method isImplicitLabel
+ * @memberof axe.commons.aria
+ * @param {Object} virtualNode virtualNode
+ * @return {Boolean}
+ */
+aria.isImplicitLabel = function(virtualNode) {
```
WilcoFiers: I don't think you should pull this into a commons. See my comment about using `text.textLabel`.

test/integration/rules/aria-form-field-label/aria-form-field-label.html
```diff
@@ -0,0 +1,26 @@
...
+
+<!-- Pass -->
+<div id="pass1" aria-label="country" role="combobox" aria-disabled="true">England</div>
+
+
+<!-- Incomplete -->
+<!-- Implicit label -->
+<label>
+  first name
+  <div id="canttell1" role="textbox"></div>
```
WilcoFiers: This should fail. It has no accessible name. Only if you added something like `aria-label="name"` to the div should it be inapplicable.

test/integration/rules/aria-form-field-label/aria-form-field-label.html
```diff
@@ -0,0 +1,26 @@
...
+</label>
+<!--  Explicit label -->
+<label for="canttell2">first name</label>
+<div role="textbox" id="canttell2"></div>
```
WilcoFiers: As above, should fail.

test/integration/rules/aria-form-field-label/aria-form-field-label.html
```diff
@@ -0,0 +1,26 @@
...
+  <option value="opel">Opel</option>
+</select>
+<textarea id="inapplicable3" title="Label"></textarea>
+
+<!-- Fail -->
+<div id="fail1" aria-label=" " role="combobox">England</div>
+<div id="fail2" aria-labelledby="non-existing" role="combobox">England</div>
+
+<!-- Pass -->
+<div id="pass1" aria-label="country" role="combobox" aria-disabled="true">England</div>
```
WilcoFiers: Why does this have aria-disabled on it?

test/integration/rules/aria-form-field-label/aria-form-field-label.html
```diff
@@ -0,0 +1,26 @@
+<!-- Inapplicable -->
```
WilcoFiers: Should exclude native form fields, buttons, and img elements with role too.

jkodu:
Closing for now, will re-open soon after resolving all changes requested.

lib/rules/aria-form-field-label-matches.js
```diff
@@ -0,0 +1,56 @@
...
+ */
+const { aria } = axe.commons;
+
+const nodeName = node.nodeName.toUpperCase();
+const role = aria.getRole(node, { noImplicit: true });
+
+/**
+ * Ignore elements from rule -> `area-alt`
+ */
+if (nodeName === `AREA` && !!node.getAttribute(`href`)) {
```
stephenmathieson: Just a nitpick, but why not use qoutes here?

test/rule-matches/aria-form-field-label-matches.js
```diff
@@ -0,0 +1,91 @@
...
+
+	it('returns false when node has no role', function() {
+		var vNode = queryFixture('<textarea id="target" title="Label"></textarea>');
+		var actual = rule.matches(vNode.actualNode, vNode);
+		assert.isFalse(actual);
+	});
+
+	it('returns false for node `map area[href]`', function() {
+		var vNode = queryFixture(
+			'<map><area id="target" href="#" role="randomRole"></map>'
```
WilcoFiers: `randomRole` would return false anyway. This test doesn't prove what you're trying to prove.

test/rule-matches/aria-form-field-label-matches.js
```diff
@@ -0,0 +1,91 @@
...
+		);
+		var actual = rule.matches(vNode.actualNode, vNode);
+		assert.isFalse(actual);
+	});
+
+	it('returns false when node is either INPUT, SELECT or TEXTAREA', function() {
+		['INPUT', 'SELECT', 'TEXTAREA'].forEach(function(node) {
+			var vNode = queryFixture(
+				'<' + node + 'role="randomRole" id="target"><' + node + '>'
```
WilcoFiers: As above, you're not proving what you're trying to test.

test/rule-matches/aria-form-field-label-matches.js
```diff
@@ -0,0 +1,91 @@
...
+			);
+			var actual = rule.matches(vNode.actualNode, vNode);
+			assert.isFalse(actual);
+		});
+	});
+
+	it('returns false when node is IMG', function() {
+		var vNode = queryFixture('<img id="target" role="randomRole">');
```
WilcoFiers: Same here

test/rule-matches/aria-form-field-label-matches.js
```diff
@@ -0,0 +1,91 @@
...
+	});
+
+	it('returns false when node is not SVG with role=`img`', function() {
+		var vNode = queryFixture('<div id="target" role="img">');
+		var actual = rule.matches(vNode.actualNode, vNode);
+		assert.isFalse(actual);
+	});
+
+	it('returns false when node is BUTTON', function() {
+		var vNode = queryFixture('<button id="target" role="button"></button>');
```
WilcoFiers: Set the role to something that would actually match, such as spinbutton.

test/rule-matches/aria-form-field-label-matches.js
```diff
@@ -0,0 +1,91 @@
...
+	it('returns false when role=`button`', function() {
+		var vNode = queryFixture('<div id="target" role="button"></div>');
+		var actual = rule.matches(vNode.actualNode, vNode);
+		assert.isFalse(actual);
+	});
+
+	it('returns false for INPUT of type `BUTTON`, `SUBMIT` or `RESET`', function() {
+		['button', 'submit', 'reset'].forEach(function(type) {
+			var vNode = queryFixture(
+				'<input id="target" role="randomRole" type="' + type + '">'
```
WilcoFiers: As above

test/integration/rules/aria-form-field-label/aria-form-field-label.html
```diff
@@ -0,0 +1,41 @@
+<!-- PASS -->
```
WilcoFiers: Add pass and fail tests for all roles you're testing: `checkbox`, `combobox`, `listbox`, `menuitemcheckbox`, `menuitemradio`, `radio`, `searchbox`, `slider`, `spinbutton`, `switch`, `textbox`

lib/checks/aria/no-implicit-explicit-label.js
```diff
@@ -0,0 +1,17 @@
...
+
+if (!accText && labelText) {
+	return undefined;
+}
+
+if (!accText.includes(labelText)) {
+	return undefined;
+}
+
+return false;
```
WilcoFiers: `return true;` This check should never fail a rule.

lib/checks/aria/no-implicit-explicit-label.json
```diff
@@ -0,0 +1,11 @@
+{
+	"id": "no-implicit-explicit-label",
+	"evaluate": "no-implicit-explicit-label.js",
+	"metadata": {
+		"impact": "moderate",
+		"messages": {
+			"fail": "Fix at least one of the following: `aria-labelledby`, `aria-label`, `non-empty title`",
+			"incomplete": "Check that the <label> does not need be part of the ARIA {{=it.data}} field's name"
```
WilcoFiers: Remove `fail`, and add a "pass" instead. Something like "There is no mismatch between a <label> and accessible name"

lib/checks/aria/no-implicit-explicit-label.js
jkodu: This is a `none` check, so returning `false` passes a rule.

test/rule-matches/aria-form-field-label-matches.js
```diff
@@ -0,0 +1,91 @@
...
+	var rule = axe._audit.rules.find(function(rule) {
+		return rule.id === 'aria-form-field-label';
+	});
+
+	afterEach(function() {
+		fixture.innerHTML = '';
+	});
+
+	it('returns false when node has no role', function() {
+		var vNode = queryFixture('<textarea id="target" title="Label"></textarea>');
```
WilcoFiers: Missing a role, such as `role="textbox"`. This element doesn't even match the selector.

test/integration/rules/aria-form-field-label/aria-form-field-label.html
```diff
@@ -0,0 +1,176 @@
...
+<!-- combobox -->
+<div id="pass2" aria-label="country" role="combobox">England</div>
+<!-- listbox -->
+<p id="pass3Label" role="label">Select a color:</p>
+<div id="pass3" role="listbox" aria-labelledby="pass3Label">
+	<div role="option">Orange</div>
+</div>
+<!-- menuitemchecbox -->
+<ul role="menu">
+	<li role="menuitem">Console</li>
```
WilcoFiers: Is there a reason for this li element to be here?

test/integration/rules/aria-form-field-label/aria-form-field-label.html
```diff
@@ -0,0 +1,176 @@
...
+		role="radio"
+		aria-checked="false"
+		tabindex="0"
+		aria-label="Regular crust"
+	>
+		Regular crust
+	</div>
+</div>
+<!-- searchbox -->
+<p id="pass7Label" role="label">Search currency pairs:</p>
```
WilcoFiers: `role="label"` isn't a thing. Please remove.

test/integration/rules/aria-form-field-label/aria-form-field-label.html
```diff
@@ -0,0 +1,176 @@
...
+<!-- combobox -->
+<div id="fail6" role="combobox">England</div>
+<!-- listbox -->
+<div id="fail7" role="listbox" aria-labelledby="label-does-not-exist">
+	<div role="option">Orange</div>
+</div>
+<!-- menuitemchecbox -->
+<ul role="menu">
+	<li role="menuitem">Console</li>
+	<li id="fail8" role="menuitemcheckbox" aria-checked="true">Word wrap</li>
```
WilcoFiers: Darnit... I think we made a mistake Jey. Some of these roles are named from contents, and others aren't. This role should pass, because "word wrap" is its accessible name. We may have to break up this rule into two. One for named from author only, and one for named from contents. Here's the breakdown:

### Named from Contents
checkbox
menuitemcheckbox
menuitemradio
radio
switch

### Named by author only
combobox
listbox
searchbox
slider
spinbutton
textbox

lib/checks/aria/no-implicit-explicit-label.js
```diff
@@ -0,0 +1,21 @@
+const { aria, text } = axe.commons;
+
+const role = aria.getRole(node, { noImplicit: true });
+this.data(role);
+
+const labelText = text.sanitize(text.labelText(virtualNode)).toLowerCase();
+const accText = text.sanitize(text.accessibleText(node)).toLowerCase();
+
+if (!accText && !labelText) {
+	return undefined;
```
WilcoFiers: I think this needs to return false. If there is no label, this check has nothing to report.

test/checks/aria/no-implicit-explicit-label.js
```diff
@@ -0,0 +1,45 @@
...
+		fixture.innerHTML = '';
+		checkContext.reset();
+	});
+
+	it('returns undefined when there is no label text or accessible text', function() {
+		var vNode = queryFixture(
+			'<div id="target" role="searchbox" contenteditable="true"></div>'
+		);
+		var actual = check.evaluate.call(checkContext, vNode.actualNode, {}, vNode);
+		assert.isUndefined(actual);
```
WilcoFiers: This should be false. We don't want people looking at labels when there are none.

test/integration/rules/aria-form-field-label/aria-form-field-label.html
jkodu: @WilcoFiers 

As discussed the other Rule (`aria-input-field-label`) PR is here - https://github.com/dequelabs/axe-core/pull/1610, that handles `named by author only`.

test/integration/rules/aria-toggle-field-label/aria-toggle-field-label.html
```diff
@@ -0,0 +1,87 @@
...
+	id="pass5"
+	role="switch"
+	aria-checked="true"
+	aria-label="Display blue light:"
+>
+	<span>off</span>
+	<span>on</span>
+</div>
+
+<!-- FAIL -->
```
WilcoFiers: All these test cases are wrong. These elements should all pass because they have an accessible name from content.

test/integration/rules/aria-toggle-field-label/aria-toggle-field-label.html
```diff
@@ -0,0 +1,85 @@
+<!-- PASS -->
+<!-- checkbox -->
+<div id="pass1" aria-label="newspaper" role="checkbox">Newspaper</div>
```
WilcoFiers: Remove aria-label, so that we can test visible-text

test/integration/rules/aria-toggle-field-label/aria-toggle-field-label.html
```diff
@@ -0,0 +1,85 @@
...
+	>
+		Word wrap
+	</li>
+</ul>
+<!-- menuitemradio -->
+<ul role="menu">
+	<li
+		id="pass3"
+		role="menuitemradio"
+		aria-label="Sans-serif"
```
WilcoFiers: use `aria-labelledby` to test the `aria-labelledby` check.

test/integration/rules/aria-toggle-field-label/aria-toggle-field-label.html
```diff
@@ -0,0 +1,85 @@
...
+	</li>
+</ul>
+<!-- radio -->
+<div role="radiogroup">
+	<div
+		id="pass4"
+		role="radio"
+		aria-checked="false"
+		tabindex="0"
+		aria-label="Regular crust"
```
WilcoFiers: And maybe a `title` here? Remove the text content though.

Issue #1419, opened 2019-03-07
New Rule: Headings with only whitespace (WCAG 2.0 A)


jeeyyy:
For Rule Specification see 
- https://github.com/auto-wcag/auto-wcag/pull/400

## Rule:
- heading-whitespace-only

## Check (any):
- heading-whitespace-only

jkodu:
# AT Testing of examples from above spec

## Pass Scenario Examples:

| Scenario/ Platform | Firefox (NVDA) | IE (Jaws) | Safari (VoiceOver) | Chrome (Talkback) |
|---|---|---|---|---|
| `<h2></h2>` | Nothing is identified | Nothing is identified | Nothing is identified | Nothing is identified |
| `<div role="heading"></div>` | Nothing is identified | Nothing is identified | Nothing is identified | Nothing is identified |
| `<h2>'</h2>` | Identified as `heading level 2` | Identified as `heading level 2` | Identified as `apostrophe heading level 2` | Identified as `heading level 2` |

## Fail Scenario Examples:

| Scenario/ Platform | Firefox (NVDA) | IE (Jaws) | Safari (VoiceOver) | Chrome (Talkback) |
|---|---|---|---|---|
| `<h2><br /></h2>` | Nothing is identified | Nothing is identified | Identified as `2 heading level 2` | Nothing is identified |
| `<h2>&nbsp;</h2>` | Nothing is identified | Identified as `heading level 2` | Nothing is identified | Nothing is identified |
| `<h2> </h2>` | Nothing is identified | Nothing is identified | Nothing is identified | Nothing is identified |
| `<h2>&#32;</h2>` | Nothing is identified | Nothing is identified | Nothing is identified | Nothing is identified |
| `<h2>&ensp;</h2>` | Identified as `heading level 2` | Nothing is identified | Nothing is identified | Nothing is identified |
| `<h2>&emsp;</h2>` |Identified as `heading level 2` | Nothing is identified | Nothing is identified | Nothing is identified |
| `<h2>&thinsp;</h2>` |Identified as `heading level 2` | Nothing is identified | Nothing is identified | Nothing is identified |
| `<h2><span><br /></span></h2>` | Nothing is identified | Nothing is identified | Nothing is identified | Nothing is identified |

## Inapplicable Examples:

| Scenario/ Platform | Firefox (NVDA) | IE (Jaws) | Safari (VoiceOver) | Chrome (Talkback) |
|---|---|---|---|---|
[...]

PR #1453, opened 2019-03-20
feat(rule): heading-whitespace-only

straker:
New rule for evaluating headings that only contain certain space separator unicode characters or `<br>` and `<wbr>` elements.

Closes: #1419

## Reviewer checks


lib/checks/shared/heading-whitespace-only.json
```diff
@@ -0,0 +1,12 @@
+{
+	"id": "heading-whitespace-only",
+	"evaluate": "heading-whitespace-only.js",
+	"metadata": {
+		"impact": "minor",
+		"messages": {
+			"pass": "Element is empty or does not have whitespace",
+			"fail": "Element does not have text that is visible to screen readers and contains only whitespace characters",
+			"incomplete": ""
```
jkodu: nitpick: remove `incomplete`

lib/rules/heading-whitespace-only.json
```diff
@@ -0,0 +1,14 @@
+{
+	"id": "heading-whitespace-only",
+	"selector": "h1, h2, h3, h4, h5, h6, [role=heading]",
+	"matches": "heading-matches.js",
+	"tags": ["wcag2a", "wcag131"],
+	"metadata": {
+		"description": "Ensure that heading elements do not have Unicode separator characters, `<br>`, or `<wbr>` elements as their only content",
+		"help": "Headings should contain text and not just whitespace or word break elements."
```
jkodu: editorial: heading elements should ...

lib/rules/heading-whitespace-only.json
```diff
@@ -0,0 +1,14 @@
...
+	},
+	"preload": false,
```
jkodu: you should remove `preload` flag

lib/checks/shared/heading-whitespace-only.js
```diff
@@ -0,0 +1,19 @@
+if (axe.commons.text.accessibleTextVirtual(virtualNode)) {
+	return true;
+}
+
+const spaceSeparatorCharCodes = [
```
jkodu: may be worth using something like this - https://www.npmjs.com/package/is-whitespace-character?

lib/checks/shared/heading-whitespace-only.js
```diff
@@ -0,0 +1,19 @@
...
+	160, // non-breaking space
+	8194, // en space
+	8195, // em space
+	8201 // thin space
+];
+
+let hasBreakingSpaceElement = node.querySelectorAll('br, wbr').length > 0;
```
jkodu: could return here immediately, and do the `hasSpaceCharacter` check after.

lib/checks/shared/heading-whitespace-only.js
```diff
@@ -0,0 +1,19 @@
...
+let hasSpaceSeparator =
+	node.textContent
```
jkodu: @WilcoFiers - is there a better util method we can use instead of textContent.

@straker - does the `accessibleTextVirtual` trim out spaces? Perhaps worth passing an argument into it to not trim spaces and use that value from above here?

I remember using `textContent` in the past & it was asked to be changed (can't remember why).

test/checks/shared/heading-whitespace-only.js
```diff
@@ -0,0 +1,192 @@
+describe('heading-whitespace-only', function() {
+	'use strict';
```
jkodu: some more edge cases:
- what happens when heading is given accessible name via `aria-*`? and they end up being space characters?
- what happens if heading element has shadowDOM that then exposes a non space character as heading content?

lib/checks/shared/heading-whitespace-only.js
straker: I had asked Wilco, and he wanted to just start with the 4 failing cases listed in https://github.com/dequelabs/axe-core/issues/1419

lib/checks/shared/heading-whitespace-only.js
straker: Ya,` accessibleTextVirtual` will return a 0 length string for all space separator characters

test/checks/shared/heading-whitespace-only.js
straker: I'll have to figure out how to handle those. Thanks for bringing them up.

dylanb: Definitely need to do some screen reader testing on the aria-label combinations

lib/commons/text/accessible-text-virtual.js
```diff
@@ -56,7 +56,7 @@ text.accessibleTextVirtual = function accessibleTextVirtual(
 		return step(virtualNode, context);
 	}, '');
 
-	if (context.startNode === virtualNode) {
+	if (context.sanitize && context.startNode === virtualNode) {
```
WilcoFiers: Strong preference for using `context.unsanitized`. That way all truthy values do one thing, and all falsey values do another, as opposed to now where undefined behaves as though it's truthy.

lib/commons/text/accessible-text-virtual.js
```diff
@@ -128,6 +128,9 @@ function prepareContext(virtualNode, context) {
 			...context
 		};
 	}
+	if (typeof context.sanitize === 'undefined') {
+		context.sanitize = true;
```
WilcoFiers: Better to treat the context as readonly if you can.

lib/checks/shared/heading-whitespace-only.js
```diff
@@ -0,0 +1,24 @@
...
+	8194, // en space
+	8195, // em space
+	8201 // thin space
+];
+
+const hasSpaceSeparator =
+	text
+		.split('')
+		.map(char => char.charCodeAt(0))
+		.filter(charCode => spaceSeparatorCharCodes.includes(charCode)).length > 0;
```
WilcoFiers: This looks wrong to me. `<h3>Foo&160;bar</h3>` shouldn't fail this rule. I think what this should be is this:

```js
const textCharCodes = text.split('').map(char => char.charCodeAt(0))
const hasOnlySpaceSeparator = textCharCodes.every(charCode => {
  return spaceSeparatorCharCodes.includes(charCode)
})
```

test/checks/shared/heading-whitespace-only.js
```diff
@@ -0,0 +1,214 @@
+describe('heading-whitespace-only', function() {
```
WilcoFiers: You're not testing combinations of things, multiple chars that should fail, chars that should fail together with chars that pass, chars mixed with break elements. Most of these tests can be integration tests if you move the chars and breaking space elements into an option object.

lib/checks/shared/heading-whitespace-only.js
straker: Should `<h3>   &160;   </h3>` also fail the rule?

jkodu:
Upon discussing about the status of the rule in [ACT-Rules](https://github.com/act-rules/act-rules.github.io/pull/400), which is going to go through some changes with `applicability` and `expectation`, it would be good to pause this rule for the time being.

I will re-open the same, when more clarity/ consensus of the above mentioned rule.

PR #1467, opened 2019-04-01
feat(rule,check): add new apis to run a rule synchronously

straker:
**Note:** the files were mostly copied from the asynchronous versions and modified to be run in a synchronous way. A few tests were deleted that no longer applied to a synchronous version.

Linked issue: #1550 

## Reviewer checks


lib/core/base/check.js
```diff
@@ -105,6 +105,58 @@ Check.prototype.run = function(node, options, context, resolve, reject) {
...
+ * Run the check's evaluate function (call `this.evaluate(node, options)`) synchronously
+ * @param  {HTMLElement} node  The node to test
+ * @param  {Object} options    The options that override the defaults and provide additional
+ *                             information for the check
+ */
+Check.prototype.runSync = function(node, options, context) {
+	/* eslint max-statements: ["error", 17] */
+	'use strict';
+	options = options || {};
+	var enabled = options.hasOwnProperty('enabled')
```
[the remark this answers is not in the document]
straker: Nope. I had just copied the async version and removed any async. I can update the copied code to be more es6-like.

lib/core/base/rule.js
```diff
@@ -135,6 +135,28 @@ Rule.prototype.runChecks = function(
...
+	const self = this;
+	let results = [];
+
+	this[type].forEach(function(c) {
+		var check = self._audit.checks[c.id || c];
+		var option = axe.utils.getCheckOption(check, self.id, options);
+		results.push(check.runSync(node, option, context));
+	});
+
+	results = results.filter(function(check) {
```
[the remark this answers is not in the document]
straker: Not sure. The async version of the code does it so I did it too just to be consistent.

lib/core/base/check.js
```diff
@@ -105,6 +105,58 @@ Check.prototype.run = function(node, options, context, resolve, reject) {
...
+				node.actualNode,
+				checkOptions,
+				node,
+				context
+			);
+		} catch (e) {
+			// In the "Audit#run: should run all the rules" test, there is no `node` here. I do
+			// not know if this is intentional or not, so to be safe, we guard against the
+			// possible reference error.
+			if (node && node.actualNode) {
```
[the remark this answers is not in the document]
straker: Based on the comment, if seems `node` doesn't exist for the `Audit#run` test, so the code would throw an error when trying to access `node.actualNode` on L137, so the `catch` would need to ensure `node` actually exists as well.

lib/core/base/rule.js
```diff
@@ -242,20 +244,131 @@ Rule.prototype.run = function(context, options, resolve, reject) {
...
+	this._markChecksStart = 'mark_runchecks_start_' + this.id;
+	this._markChecksEnd = 'mark_runchecks_end_' + this.id;
+};
+
+/**
+ * Log performance of rule.gather
+ * @param {Rule} rule The rule to log
+ * @param {Array} nodes Result of rule.gather
+ */
+function logGatherPerformance(rule, nodes) {
```
WilcoFiers: I think all these new functions should be methods on the Rule class. That makes the implementation consistent.

test/core/base/rule.js
```diff
@@ -755,6 +756,583 @@ describe('Rule', function() {
...
+					isNotCalled(err);
+				}
+			});
+
+			it('should handle an error in #matches', function() {
+				var div = document.createElement('div');
+				div.setAttribute('style', '#fff');
+				fixture.appendChild(div);
+				var success = false,
+					rule = new Rule({
```
WilcoFiers: Strong preference for giving each variable its own `var`.

test/core/base/check.js
```diff
@@ -274,6 +274,124 @@ describe('Check', function() {
 				});
 			});
 		});
+
+		describe('runSync', function() {
```
WilcoFiers: You're not testing the `virtualNode`

paulirish:
> Prepare for running in non-browser environments by allowing a rule to run synchronously.

Just curious, but why not run the rules with their async hops in these environments?

Also, non-browser environments that have DOM nodes?? :) Is this specifically about JSDOM compat?

stephenmathieson:
@paulirish there's a new product we're hacking on that has these requirements. Stay tuned :)

test/core/base/check.js
straker: These test are copy pasted from the async version of the test. Does that mean the async version is not testing `virtualNode`?

WilcoFiers: Jup, sounds like it. What I think is missing is just a test that the virtualNode is passed to the evaluate function correctly.

PR #1469, opened 2019-04-02
fix: Scroll state had top and left properties flipped

mislav:
The arguments for `window.scroll()` are [x-axis, y-axis](https://developer.mozilla.org/en-US/docs/Web/API/Window/scroll); meaning that "top" value (vertical scrolling) was supposed to be passed as 2nd argument.

## Reviewer checks

### DIFF
--- a/lib/core/utils/scroll-state.js
+++ b/lib/core/utils/scroll-state.js
@@ -20,7 +20,7 @@ function getScroll(elm) {
  */
 function setScroll(elm, top, left) {
 	if (elm === window) {
-		return elm.scroll(top, left);
+		return elm.scroll(left, top);
 	} else {
 		elm.scrollTop = top;
 		elm.scrollLeft = left;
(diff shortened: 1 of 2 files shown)


mislav:
@WilcoFiers There was already a test case, but the test case involved a stub of `window.scroll` that was set up in the wrong way. This PR corrects the stub, therefore we do have a test case going forward that ensures this won't regress.

A "proper" test case would be to invoke the non-stubbed version of `window.scroll` and verify the scroll position changes by inspecting `window.scrollY`. However, I'm not sure if it's a good idea for the test suite to change the scroll position of the document executing the test suite. I think the author of the initial test felt this as well; that's why they created a stub in the first place.

Do you have an idea for a better test?

stephenmathieson:
@WilcoFiers thoughts on the above? I think @mislav has a good point.

PR #1472, opened 2019-04-03
feat(rule): add more perf timing metrics to rules

straker:
Add metrics for `rule.gather`, `rule.matches` and `axe.utils.isHidden`. This allows us to see where a rule is taking the most time better. Also rename rules to be sortable based on id. Lastly, have axe only output metrics that were done after axe run, otherwise you get metrics that the website ran before axe did, which makes it really confusing when reading the metric logs.

Example output:

```
Measure rule_aria-valid-attr-value#gather_axe.utils.isHidden took 931ms
Measure rule_aria-valid-attr-value#gather took 931ms
Measure rule_aria-valid-attr-value took 937ms
Measure rule_aria-valid-attr-value#matches took 4ms
Measure rule_aria-valid-attr-value#runchecks took 2ms
Measure rule_aria-valid-attr#gather_axe.utils.isHidden took 960ms
Measure rule_aria-valid-attr#gather took 960ms
Measure rule_aria-valid-attr took 969ms
Measure rule_aria-valid-attr#matches took 9ms
Measure rule_aria-valid-attr#runchecks took 0ms
Measure rule_color-contrasr#gather_axe.utils.isHidden took 0ms
Measure rule_color-contrast#gather took 0ms 
Measure rule_color-contrast#matches took 860ms 
Measure rule_color-contrast took 1264ms 
Measure rule_color-contrast#runchecks took 404ms
```

Closes issue: N/A

## Reviewer checks

### DIFF
--- a/lib/core/base/rule.js
+++ b/lib/core/base/rule.js
@@ -90,16 +90,49 @@ Rule.prototype.matches = function() {
 /**
  * Selects `HTMLElement`s based on configured selector
  * @param  {Context} context The resolved Context object
+ * @param  {Mixed}   options Options specific to this rule
  * @return {Array}           All matching `HTMLElement`s
  */
-Rule.prototype.gather = function(context) {
+Rule.prototype.gather = function(context, options) {
 	'use strict';
+	const markStart = 'mark_gather_start_' + this.id;
+	const markEnd = 'mark_gather_end_' + this.id;
+	const markHiddenStart = 'mark_isHidden_start_' + this.id;
+	const markHiddenEnd = 'mark_isHidden_end_' + this.id;
+
+	if (options.performanceTimer) {
+		axe.utils.performanceTimer.mark(markStart);
+	}
+
 	var elements = axe.utils.select(this.selector, context);
 	if (this.excludeHidden) {
-		return elements.filter(function(element) {
+		if (options.performanceTimer) {
+			axe.utils.performanceTimer.mark(markHiddenStart);
+		}
+
+		elements = elements.filter(function(element) {
 			return !axe.utils.isHidden(element.actualNode);
 		});
+
+		if (options.performanceTimer) {
+			axe.utils.performanceTimer.mark(markHiddenEnd);
+			axe.utils.performanceTimer.measure(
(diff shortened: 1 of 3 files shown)


lib/core/base/rule.js
```diff
@@ -228,6 +259,38 @@ Rule.prototype.run = function(context, options, resolve, reject) {
...
 
+/**
+ * Selects `HTMLElement`s based on configured selector and filters them based on
+ * the rules matches function
+ * @param  {Rule} rule The rule to check for after checks
+ * @param  {Context} context The resolved Context object
+ * @param  {Mixed}   options Options specific to this rule
+ * @return {Array}           All matching `HTMLElement`s
+ */
+function gatherAndMatchNodes(rule, context, options) {
```
WilcoFiers: I don't think this should be a separate function. Feels like a strange architectural decision to pull some logic out of a class method. I get that you're trying to keep this method from bloating, but I'd rather pull line 210 - 239 out into a separate utils function in that case.

I'm not entirely sure about it though, so @stephenmathieson I'd like a second opinion.

stephenmathieson: I agree that moving this logic _out_ of the class doesn't seem like the right decision. Perhaps we just add another method to the class containing this logic and make it more "usable" in all scenarios?

straker: When adding the `runSync` method for #1467 , I tried to pull out as much of the duplicate code as possible to make it easier to maintain the two similar functions. This was mostly just an extension of doing that in prep for that pr (as they will conflict)

stephenmathieson: Right. So I think we can make this an instance method that will work for both cases.

Or am I missing something there?

straker: I figured since it wasn't callable outside the function of `run`, we wouldn't want to make it now callable publicly. But I can make it a public api method of a rule if we want, it just doesn't do a whole lot on it's own.

stephenmathieson: It sounds like I'm missing something. Let's chat about this tomorrow on Slack.

jeankaplansky:
Not rule-help related.

PR #1481, opened 2019-04-08
chore: require polyfills from core-js

jkodu:
- Remove copy pasted polyfills from `lib/core/utils/pollyfills.js` and instead require the same from well-maintained `core-js`, included by `@babel/polyfill` - https://babeljs.io/docs/en/babel-polyfill.

- Remove `es6-promise` dependency and use `core-js/fn/promise` polyfill.

For size(s) of required modules, see below:
image

Closes issue:
- NA

## Reviewer checks


jkodu:
@stephenmathieson 
I have only added the ones that already existed in the `polyfills.js`, 
https://github.com/dequelabs/axe-core/pull/1481/files#diff-44472deac859e4d1ecba96df5affa641

And I opened this PR to get an opinion from everyone if this is the right direction to go, over copy pasted polyfill(s)

package.json
```diff
@@ -88,7 +88,6 @@
 		"derequire": "^2.0.6",
 		"dot": "~1.1.2",
 		"emoji-regex": "8.0.0",
-		"es6-promise": "^4.2.6",
```
WilcoFiers: You'll have to add `core-js` if you're going to use it as a dependency.

WilcoFiers:
Discussed with Jey. The main concern with limiting the size of axe-core is that we need to be able to include several versions of it in a Firefox Addon. The restrictions on that have been loosened quite a bit, now allowing for 200mb addons. That means this is far less of a concern than it used to be.

jkodu:
After discussing with @WilcoFiers, and checking allowed max size for extensions etc, decided to go forward with this PR. Have done the change of adding `core-js` as a dependency.

stephenmathieson:
I think it's important to consider/keep-track-of package weight. I understand the previous requirement was due to a limitation in the Firefox extension and that's no longer a concern. However, as a consumer/user of axe-core, I do not want to be required to download tons and tons of code in my apps/etc if I don't have to.

stephenmathieson:
Closing for now.

Though it makes sense to offload the "ownership" of these polyfills to an npm package, I don't agree that it's worth adding 30k to the overall size of axe-core.

Issue #1346, opened 2019-02-04
Incorrect skiplink error


WilcoFiers:
On the following page, axe-core incorrectly flagged a link with `href="#0"` as a skiplink with no target. The link is something with an event handler on it, the href isn't supposed to reference something. Not sure how to work around that though, except to put it in needs review.
https://css-tricks.com/reacts-experimental-suspense-api-will-rock-for-fallback-ui-during-data-fetches/

scurker:
Are you suggesting that we put all links with an `href="#..."` value that does not point to a target as needs review?

straker:
What if we only looked at the first link on the page and made sure the href started with `#...` and that it pointed to a target that was visible. WebAIM states that

> The horizontal location of the link doesn't matter much. The link can be on the left side, the middle, the right, or in some location in between. The key is to make sure the link is one of the first items that screen readers hear and that keyboard users tab to. Otherwise, users may not realize there is a "skip navigation" link there at all, and may waste time trying to muddle through the extraneous links. Screen reader users especially may get impatient if they don't hear the "skip navigation" link very early in the page.

straker:
As an alternative solution if we want to look at all skip links on a page, we could define a skip link as any anchor element whose href starts with `#...` and which precedes the first link whose href doesn't start with `#...`

```js
let firstPageLink = document.querySelector('a:not([href^="#"])');
let skipLinks = Array.from(document.querySelectorAll('a[href^="#"]')).filter(link => {
  return link.compareDocumentPosition(firstPageLink) === 4
});
```

WilcoFiers:
Cool idea! Yeah, let's do it!

Issue #863, opened 2018-04-26
Skip links and landmark regions


digitala11y:
Ensures all content is contained within a landmark region
 <html lang="en-US" prefix="og: http://ogp.me/ns#">
 URL:https://www.digitala11y.com

Lang attribute cannot  be wrapped inside a landmark region.

marcysutton:
Have a look at the related nodes. Those are the elements that aXe is saying should be wrapped in a landmark region. In this case they're skip links, which we've talked about making exempt.

WilcoFiers:
Our skiplink ignore part of this rule should account for multiple skip links in a page. I think that's the fix here.

PR #1496, opened 2019-04-15
fix(skip-link,region): Allow multiple skiplinks at page top

straker:
Redefine a skip link as any anchor element whose `href` starts with `#...` and which precedes the first anchor element whose `href` doesn't start with `#..`. (that is, a link to a page)

Closes: #1346
Closes: #863

## Reviewer checks


straker:
We should hold off merging this for just a bit. As it stands, it increases the time of the `skip-link` rule on a page of >11k elements from ~30ms to ~140ms. We should figure out a way to cache the `firstPageLink` so we only have to do it once for the entire run instead of for each anchor element.

lib/commons/dom/is-skip-link.js
```diff
@@ -0,0 +1,36 @@
...
+	}
+
+	// define a skip link as any anchor element whose href starts with `#...`
+	// and which precedes the first anchor element whose href doesn't start
+	// with  `#...` (that is, a link to a page)
+	const firstPageLink = axe.utils.querySelectorAll(
+		axe._tree,
+		'a:not([href^="#"]):not([href^="/#"]):not([href^="javascript"])'
+	)[0];
+	if (firstPageLink) {
```
WilcoFiers: Nitpick: In the spirit of keeping the [happy path on the left](https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88), I'd suggest `if (!firstPageLink) { return true }`. Not having a first link is an exception, not the default.

lib/commons/dom/is-skip-link.js
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Determines if element is a skip link
+ * @method isSkipLink
+ * @memberof axe.commons.dom
+ * @instance
+ * @param  {Element} element
+ * @return {Boolean}
+ */
+dom.isSkipLink = function(element) {
```
WilcoFiers: We'll need to have tests for this. Feel free to move some of the tests of the check into a separate tests for this commons.

straker:
When #1503 is merged, I can cache the firstPageLink so we won't have to recalculate it for each node.

test/checks/navigation/skip-link.js
```diff
@@ -7,31 +7,26 @@ describe('skip-link', function() {
...
-		assert.isFalse(checks['skip-link'].evaluate(node));
-	});
-
 	it('should return true if the href points to an element with an ID', function() {
 		fixture.innerHTML =
 			'<a href="#target">Click Here</a><h1 id="target">Introduction</h1>';
+		axe._tree = axe.utils.getFlattenedTree(fixture);
 		var node = fixture.querySelector('a');
-		assert.isTrue(checks['skip-link'].evaluate(node));
+		assert.isTrue(axe.commons.dom.isSkipLink(node));
```
WilcoFiers: Why are you running a commons here, and not the check?

test/checks/navigation/skip-link.js
```diff
@@ -7,31 +7,26 @@ describe('skip-link', function() {
...
 	});
 
 	it('should return true if the href points to an element with an name', function() {
 		fixture.innerHTML = '<a href="#target">Click Here</a><a name="target"></a>';
+		axe._tree = axe.utils.getFlattenedTree(fixture);
 		var node = fixture.querySelector('a');
-		assert.isTrue(checks['skip-link'].evaluate(node));
+		assert.isTrue(axe.commons.dom.isSkipLink(node));
```
WilcoFiers: As above.

test/checks/navigation/skip-link.js
straker: I think I had moved them to the `isSkipLink` test before deciding to move it back, but forgot to change it back to a check test. Good catch.

PR #1511, opened 2019-04-23
test: rollback playground.html

jkodu:
- `playground.html` was updated from a previously authored rule.

Closes issue: NA

## Reviewer checks


jkodu:
@WilcoFiers - I understand that, but the contents of the file has changes from a PR for a rule (avoid-inline-spacing), which should not be the case.

jkodu:
Closing, as it is not a concern. Will re-open after discussion with @WilcoFiers

Issue #1195, opened 2018-10-18
raw reporter shouldn't output DqElements


WilcoFiers:
While working on a new reporter I tried to run `JSON.stringify()` on the results of the `raw` reporter. Turns out that this is outputting DqElements, which can't be stringified after completing a run. Because the selectors are lazily constructed, and in order to construct them `axe._selectorData` is required, they can't be constructed after cleanup of `axe._selectorData`. To fix this we need to change the `raw` reporter to not output DqElement objects.

jkodu:
Note:
Go thru results and any Dq element, toJson() or equivalent for selector

jkodu:
@WilcoFiers - I was unable to find any DqElement in the raw report, so I could run a `toJson` on the same. Need more input.

dbjorge:
We just hit this issue during development of https://github.com/Microsoft/axe-sarif-converter - the problematic elements are at JSON paths along the lines of `root_raw_results[0].violations[0].node` and `root_raw_results[0].violations[0].any[0].relatedNodes[0]`.

One option for a low-churn fix that we tried out successfully would be updating the raw reporter to output `JSON.parse(JSON.stringify(results))`, rather than outputting results directly. I don't know that I'd recommend this, though, since DqElement.toJSON omits some properties from the original DqElement; I think it would be preferable for the output of the raw reporter to be as close as possible to the input that a new reporter library would be receiving (using the output of the raw reporter as test cases for the new reporter we're developing has been our primary use case for the raw reporter so far), so I think it might be better to do a more explicit step to replace the lazy properties in question with non-lazy ones in the existing finalization phase before the results get passed to reporters (https://github.com/dequelabs/axe-core/blob/develop/lib/core/utils/finalize-result.js).

AutoSponge:
@WilcoFiers so, the test at https://github.com/dequelabs/axe-core/blame/320e63380189f507a32c99fd19619f3c6a161764/test/core/utils/dq-element.js#L35 is not a valid test for your case?

That may defeat the purpose of the `raw` reporter. Could I suggest a `string` reporter that uses a `JSON.stringify({}, `**replacer**`)` function to remove or otherwise normalize DQElements?

PR #1513, opened 2019-04-24
fix(raw-reporter): do not output `DqElement`s

stephenmathieson:
This patch updates our "raw" reporter, ensuring it does not pass `DqElement`s to its callback. Instead, it serializes the `DqElement` by using the `DqElement#toJSON()` method.

Unfortunately many of our tests pass non-results to the reporters, so several "guards" were added to the reporter ensuring we don't error while attempting to transform the provided results. In an ideal world, these tests would be rewritten to pass "proper" result objects to our reporters, but this felt out-of-scope for the task at hand.

Closes https://github.com/dequelabs/axe-core/issues/1195

## Reviewer checks

### DIFF
--- a/lib/core/reporters/raw.js
+++ b/lib/core/reporters/raw.js
@@ -1,8 +1,32 @@
 axe.addReporter('raw', function(results, options, callback) {
 	'use strict';
+
 	if (typeof options === 'function') {
 		callback = options;
 		options = {};
 	}
-	callback(results);
+
+	// Guard against tests which don't pass an array as the first param here.
+	if (!results || !Array.isArray(results)) {
+		return callback(results);
+	}
+
+	const transformedResults = results.map(result => {
+		const transformedResult = { ...result };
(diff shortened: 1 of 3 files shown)


stephenmathieson:
> Did you mean to check in the playground?

It was "dirty" and I just did `git commit -am ...`. I'm not sure what the "rules" are here. Should I roll back the changes?

jkodu:
Yup, the rule is, playground is do not commit it, so yes rollback the playground changes.

jkodu:
Which is why I thought perhaps `playground.html` should be bare bones like this PR - https://github.com/dequelabs/axe-core/pull/1511

WilcoFiers:
@stephenmathieson Jup, just roll it back.

stephenmathieson:
OK, I'll roll it back.

What's the purpose of versioning the file if it's not supposed to be updated tho?

Issue #1463, opened 2019-03-29
aria-valid-attr-value possible false positive with aria-controls


bebebebebe:
Assumption: aria-controls can be used for a collapsible element. While the element is in collapsed state, it could be the collapsed content is not in the DOM. In this case, we can have aria-controls="someId", but no element present with id "someId."
https://www.w3.org/WAI/GL/wiki/Using_the_WAI-ARIA_aria-expanded_state_to_mark_expandable_and_collapsible_regions

Expected: The case described above will not lead to an aria-valid-attr-value error.

Actual: The case described above leads to an aria-valid-attr-value error.

Motivation: I'm concerned that the above is a false positive. Alternatively, if my assumption above is incorrect, it would be helpful to know so I can avoid actual errors!

axe-core version: 3.2

WilcoFiers:
Totally agreed, when on a thing with aria-expanded="false" I would think that should be allowed. Thank you for reporting!

iamrafan:
@WilcoFiers We think, it should also be allowed on elements with aria-selected="false". Here is an example pointed out by one of our users: https://www.w3.org/TR/wai-aria-practices-1.1/examples/tabs/tabs-1/tabs.html

The inactive tabs should not fail this rule when the elements referenced to aria-controls(hidden elements in this example) are removed form the DOM.
![image](https://user-images.githubusercontent.com/22068434/55552592-a120b500-5692-11e9-91eb-4dc70709acbd.png)

iamrafan:
@straker @WilcoFiers:  aria-owns also should be passed when ```aria-expanded="false"```. For example, comboboxes are required to have popups only when expanded.

straker:
@iamrafan could you open a separate issue for that one? I'll take a look at it as well

iamrafan:
@straker Sure, I filed #1524

jeankaplansky:
@straker : what, if any documentation is required to complete this ticket?

PR #1515, opened 2019-04-24
fix(aria-valid-attr-value): allow aria-controls to pass when element is not in the DOM

straker:
Closes: #1463

## Reviewer checks

### DIFF
--- a/lib/checks/aria/valid-attr-value.js
+++ b/lib/checks/aria/valid-attr-value.js
@@ -18,9 +18,8 @@ const preChecks = {
 	}
 };
 
-for (var i = 0, l = attrs.length; i < l; i++) {
-	let attr = attrs[i];
-	let attrName = attr.name;
+for (const attr of attrs) {
+	const attrName = attr.name;
 	// skip any attributes handled elsewhere
 	if (
 		!skipAttrs.includes(attrName) &&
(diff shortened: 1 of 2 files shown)


lib/checks/aria/valid-attr-value.js
```diff
@@ -1,26 +1,35 @@
...
+const preChecks = {
+	'aria-controls': function() {
+		return (
+			node.getAttribute('aria-expanded') !== 'false' &&
+			node.getAttribute('aria-selected') !== 'false'
+		);
+	}
+};
 
 for (var i = 0, l = attrs.length; i < l; i++) {
```
stephenmathieson: While you're here and updating `var` to `const` already, maybe update this loop too?

```js
for (const attr of attrs) {
  const attrName = attr.name
  // ...
}
```

lib/checks/aria/valid-attr-value.js
```diff
@@ -1,26 +1,35 @@
...
-			invalid.push(attrName + '="' + attr.nodeValue + '"');
-		}
+	if (
+		!skipAttrs.includes(attrName) &&
+		options.indexOf(attrName) === -1 &&
+		aria.test(attrName) &&
+		(preChecks[attrName] ? preChecks[attrName]() : true) &&
+		!axe.commons.aria.validateAttrValue(node, attrName)
+	) {
+		invalid.push(attrName + '="' + attr.nodeValue + '"');
```
stephenmathieson: A template string would likely be more readable:

```js
invalid.push(`${attrName}="${attr.nodeValue}"`)
```

🤷‍♀️

lib/checks/aria/valid-attr-value.js
straker: I keep forgetting that `for of` is a thing.

Issue #1456, opened 2019-03-22
False positive: color-contrast rule fails when body height set to 0-8px


iamrafan:
Color-contrast rule fails when body height is set to 0-8px, irrespective of the presence of sufficient background/foreground contrast ratio.

Demo: https://codepen.io/iamrafan/pen/WmLJmw?editors=1000#0

<pre><code>axe-core version: 3.1.2
axe-Coconut version: 3.7.0</code></pre>

WilcoFiers:
That's pretty darn obscure, but jup, that's a false positive.

scurker:
I did some digging to see if I could determine the root of the problem.

In the example, there's a link element of which we take the center x/y coordinates in order to determine the background element stack. It just so happens the y coordinate is _17 pixels_ which when we take into account the body height of `8px` and the user agent margin of `8px`, `body` ends up being left out of the background stack with `elementsFromPoint()`.

However, this is a case where the visible canvas has a special background color according to the spec:

[CSS3 Backgrounds Spec](https://www.w3.org/TR/css-backgrounds-3/#special-backgrounds):

> The document canvas is the infinite surface over which the document is rendered. Since no element corresponds to the canvas, in order to allow styling of the canvas CSS propagates the background of the root element (or, in the case of HTML, the `<body>` element) as described below. However, if no boxes are generated for the element whose background would be used for the canvas (for example, if the root element has display: none), then the canvas background is transparent.

Since the visible HTML canvas has a background color but not a intrinsically defined one that's within the coordinates, the rule fails. We should probably have `html` inherit its color value from `<body>` when no background-color exists for `html`.

jeankaplansky:
No docs required.

chandana7393:
Color-contrast violation is not throwing when height is set to 0-8px.
![color_contrast_height](https://user-images.githubusercontent.com/45285854/58779197-c5e1cc00-85f2-11e9-881b-ba1f5f0b3296.PNG)
**Tested Environment:**
Axe-coconut - 3.8.0.15404v
Chrome - 74v
OS - Windows 10 64 bit.

PR #1520, opened 2019-04-25
fix: Include body as part of background color checks when element does not intersect

scurker:
[`sortPageBackground`](https://github.com/dequelabs/axe-core/blob/c7803bfa3fca037ec9cb168853a8c009b3fab602/lib/commons/color/get-background-color.js#L207) did not account for instances where an element did not visually intersect with `document.body` but should inherit the background color from the canvas background (`<html>`).

According to the [CSS3 Background Spec](https://www.w3.org/TR/css-backgrounds-3/#special-backgrounds), the document canvas should propagate the background color of the `<body>` element when all of the following are true:

* The `<html>` element has `background-image: none`
* The `<html>` element has `background-color: transparent`

The following [CodePen](https://codepen.io/iamrafan/pen/WmLJmw) provides a case where axe shows a false positive due to the above missing check.

* Added: Include `body` as part of the element stack when element does not intersect with body
* Moved `#fixture` element in test template as it was interfering with the layout of elements due to to the Mocha results being dynamic.
* When moving `#fixture`, the static display of `#mocha-stats` was interfering with a few tests.
* Included more verbose browser information when running with WebDriver for easier debugging.

I opted to go the route of moving the `#fixture` and resolving the breaking tests, as the alternatives I explored seemed dirty and hacky.

Closes issue: #1456 

## Reviewer checks

### DIFF
--- a/lib/commons/color/get-background-color.js
+++ b/lib/commons/color/get-background-color.js
@@ -207,8 +207,10 @@ color.getRectStack = function(elm) {
 function sortPageBackground(elmStack) {
 	let bodyIndex = elmStack.indexOf(document.body);
 	let bgNodes = elmStack;
+	// prettier-ignore
 	let sortBodyElement =
-		bodyIndex > 1 || bodyIndex === -1; // only if there are negative z-index elements // only when an element is positioned outside of the body's rect
+		bodyIndex > 1 || // only if there are negative z-index elements
+		bodyIndex === -1; // only when an element is positioned outside of the body's rect
 
 	if (
 		sortBodyElement &&
--- a/lib/commons/color/get-background-color.js
+++ b/lib/commons/color/get-background-color.js
@@ -207,10 +207,17 @@ color.getRectStack = function(elm) {
 function sortPageBackground(elmStack) {
 	let bodyIndex = elmStack.indexOf(document.body);
 	let bgNodes = elmStack;
+
+	// Body can sometimes appear out of order in the stack:
+	//   1) Body is not the first element due to negative z-index elements
+	//   2) Elements are positioned outside of body's rect coordinates
+	//      (see https://github.com/dequelabs/axe-core/issues/1456)
+	// In those instances we want to reinsert body back into the element stack
+	// when not using the root document element as the html canvas for bgcolor
 	// prettier-ignore
 	let sortBodyElement =
-		bodyIndex > 1 || // only if there are negative z-index elements
-		bodyIndex === -1; // only when an element is positioned outside of the body's rect
+		bodyIndex > 1 || // negative z-index elements
+		bodyIndex === -1; // element does not intersect with body
 
 	if (
 		sortBodyElement &&
@@ -220,9 +227,10 @@ function sortPageBackground(elmStack) {
 		).alpha === 0
 	) {
 		// Only remove document.body if it was originally contained within the element stack
-		if (bodyIndex !== -1) {
+		if (bodyIndex > 1) {
 			bgNodes.splice(bodyIndex, 1);
 		}
+		// Remove document element since body will be used for bgcolor
 		bgNodes.splice(elmStack.indexOf(document.documentElement), 1);
 
 		// Put the body background as the lowest element
--- a/lib/commons/color/get-background-color.js
+++ b/lib/commons/color/get-background-color.js
@@ -207,17 +207,20 @@ color.getRectStack = function(elm) {
 function sortPageBackground(elmStack) {
 	let bodyIndex = elmStack.indexOf(document.body);
 	let bgNodes = elmStack;
+	let sortBodyElement =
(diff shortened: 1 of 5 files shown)


straker:
Should we add a test to verify that we don't break the check in either direction? So works when there is a color-contrast problem (test we don't have) and when there isn't (test we have now).

scurker:
I believe in this case, I'm just testing that an element returns the correct background stack. An element that visually intersects with `body` should have `[element, ... , body, html]` as its stack while an element that does not should have `[element, ... , body]` when the html element *does not* have an image or opaque background.

I'm not sure what the inverse test would be, although it should be covered by the existing color-contrast tests.

lib/commons/color/get-background-color.js
```diff
@@ -207,17 +207,20 @@ color.getRectStack = function(elm) {
 function sortPageBackground(elmStack) {
 	let bodyIndex = elmStack.indexOf(document.body);
 	let bgNodes = elmStack;
+	let sortBodyElement =
+		bodyIndex > 1 || bodyIndex === -1; // only if there are negative z-index elements // only when an element is positioned outside of the body's rect
 
 	if (
-		// Check that the body background is the page's background
-		bodyIndex > 1 && // only if there are negative z-index elements
+		sortBodyElement &&
```
WilcoFiers: Shouldn't this just be `bodyIndex !== 0`, since `bodyIndex` can never be less than -1.

lib/commons/color/get-background-color.js
```diff
@@ -207,17 +207,20 @@ color.getRectStack = function(elm) {
...
 		!color.elementHasImage(document.documentElement) &&
 		color.getOwnBackgroundColor(
 			window.getComputedStyle(document.documentElement)
 		).alpha === 0
 	) {
-		// Remove body and html from it's current place
-		bgNodes.splice(bodyIndex, 1);
+		// Only remove document.body if it was originally contained within the element stack
+		if (bodyIndex !== -1) {
```
WilcoFiers: This looks wrong to me. My understanding is that the body is **always** at the back. Even if it has a height of 1px, and elementsFromPoint doesn't pick it up, the browser will render it over the whole page anyway. It seems to me like you should also do this splice when `bodyIndex === -1` (so remove the if).

Otherwise what you'll get is (like in your test case) a black background, even though the body has a different color.

test/commons/color/get-background-color.js
```diff
@@ -740,6 +740,28 @@ describe('color.getBackgroundColor', function() {
...
+		document.body.style.height = '1px';
+		document.body.style.background = '#000';
+
+		var target = fixture.querySelector('#target');
+		var actual = axe.commons.color.getBackgroundColor(target, []);
+
+		assert.closeTo(actual.red, 0, 0);
+		assert.closeTo(actual.green, 0, 0);
+		assert.closeTo(actual.blue, 0, 0);
+		assert.closeTo(actual.alpha, 1, 0);
```
WilcoFiers: I don't think this should end up black (see my comment above). It looks to me like even with the body height changed, the background color is still the background color of the body.

lib/commons/color/get-background-color.js
scurker: It doesn't look that way, at least according to the [CSS3 spec](https://www.w3.org/TR/css-backgrounds-3/#special-backgrounds):

> For documents whose root element is an HTML HTML element or an XHTML html element [HTML]: if the computed value of background-image on the root element is none and its background-color is transparent, user agents must instead propagate the computed values of the background properties from that element’s first HTML BODY or XHTML body child element. The used values of that BODY element’s background properties are their initial values, and the propagated values are treated as if they were specified on the root element.

The `<html>` or document canvas just inherits the color from whatever the root element is, and doesn't expand the root element itself.

Here's a test case that demonstrates this: https://jsfiddle.net/o3hjdsy8/1/ Even though the body has an implicit height of `10px`, the canvas still fills the screen with the background color according to the spec.

I tested this in:

* Chrome 73
* Firefox 66
* Safari 12

When getting the center of the element with `document.elementsFromPoint(x, y)` the element stack does not include `<body>` in any of the above browsers.

lib/commons/color/get-background-color.js
scurker: `bodyIndex` can be `-1` in some specific scenarios. See https://jsfiddle.net/o3hjdsy8/1/

lib/commons/color/get-background-color.js
```diff
@@ -207,17 +207,22 @@ color.getRectStack = function(elm) {
...
```
WilcoFiers: OK, finally figured out what's going on :P. Let's flip around this if so it's not a double negative, maybe add in one or two more comments to explain it.

```js
if (bodyIndex > 1) { // If the body is in the stack, take it out
```

Issue #1532, opened 2019-05-01
Error during analyse of  Website - should report


FunThomas424242:
Details following in the text:

Testobject: https://funthomas424242.github.io/honey-toggle-button/
wich contains this webcomponent
https://funthomas424242.github.io/honey-toggle-button/src/components/toggle-button/toggle-button.js

Expectation: I expect a list of accessibillity errors of my webcomponent.

Actual: There is a stacktrace of axe-cocounut addon

Motivation: The stacktrace write i should create an issue

<pre><code>
An error occurred
Please try again. If the error persists, pleasereport the issue  to Deque Systems

TypeError: Cannot read property 'nodeMap' of undefined
    at G (chrome-extension://iobddmbdndbbbfjopjdgadphaoihpojp/adapter.bundle.js:97:74309)
    at Object.utils.getFlattenedTree (chrome-extension://iobddmbdndbbbfjopjdgadphaoihpojp/adapter.bundle.js:97:81423)
    at Object.j [as cleanup] (chrome-extension://iobddmbdndbbbfjopjdgadphaoihpojp/adapter.bundle.js:97:51377)
    at chrome-extension://iobddmbdndbbbfjopjdgadphaoihpojp/adapter.bundle.js:109:321777
    at new Promise (<anonymous>)
    at Object.u [as highlight] (chrome-extension://iobddmbdndbbbfjopjdgadphaoihpojp/adapter.bundle.js:109:321749)
    at chrome-extension://iobddmbdndbbbfjopjdgadphaoihpojp/adapter.bundle.js:85:171163
    at onMessage (chrome-extension://iobddmbdndbbbfjopjdgadphaoihpojp/browser-polyfill.js:1056:22)
    at EventImpl.dispatchToListener (extensions::event_bindings:403:22)
    at Event.publicClassPrototype.(anonymous function) [as dispatchToListener] (extensions::utils:138:26)

Following the chrome extension details site: 

Erweiterungssymbol axe-Coconut An

Beschreibung
Accessibility testing in Chrome Developer Tools
Version
3.8.0.12104 build 50d7d54b96
Größe
3,1 MB
Berechtigungen
Alle Ihre Daten auf von Ihnen besuchten Websites lesen und ändern
Dieser Erweiterung erlauben, alle Ihre Daten auf von Ihnen besuchten Websites zu lesen und zu ändern:
Weitere Informationen zum Websitezugriff

Im Inkognitomodus zulassen
[...]

PR #1535, opened 2019-05-01
fix(utils): make cache global instead of only setup in axe.run

straker:
Turns out that a lot of things can call `axe.utils.getFlattenedTree` without going through `axe.run`, including our own code when we cleanup plugins https://github.com/dequelabs/axe-core/blob/develop/lib/core/public/cleanup-plugins.js#L28. This caused the axe extensions to break. 

Moving the creation of the `nodeMap` to the flatten tree code meant that the cache also had to exist at that point, which meant it was created in two different places (`axe.run` and `getFlattenedTree`). I decided it was cleaner to create a global private cache that always existed and could be used anywhere for any code that needed it. That way you weren't restricted to go through `axe.run` to have the cache available.

Linked issues: #1532 

## Reviewer checks

### DIFF
--- a/lib/core/utils/cache.js
+++ b/lib/core/utils/cache.js
@@ -1,0 +1,33 @@
+(function() {
+	'use strict';
+	let _cache = {};
+
+	const cache = {
+		/**
+		 * Set an item in the cache.
+		 * @param {String} key - Name of the key.
+		 * @param {*} value - Value to store.
+		 */
+		set(key, value) {
+			_cache[key] = value;
+		},
+
+		/**
+		 * Retrieve an item from the cache.
+		 * @param {String} key - Name of the key the value was stored as.
+		 * @returns {*} The item stored
+		 */
+		get(key) {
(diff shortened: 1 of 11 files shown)


lib/core/utils/cache.js
```diff
@@ -0,0 +1,33 @@
...
+
+		/**
+		 * Clear the cache.
+		 */
+		clear() {
+			_cache = {};
+		}
+	};
+
+	axe.utils.cache = cache;
```
stephenmathieson: I'm a little nervous about adding this to "public" namespace. It's very hard for us to deprecate public APIs right now, and I'm worried that this could make that more difficult.

For example, what if a custom rule starts using this to store some sort of arbitrary data?

straker:
Good call. I created an integration test for plugins following our own docs code https://github.com/dequelabs/axe-core/blob/develop/doc/plugins.md. The last test breaks in the develop branch, as was expected.

PR #1536, opened 2019-05-02
fix: check if property exists in cache of flattenedTree

jkodu:
During a pairing session, noticed a bug with checking if a property already exists in the `_vNodeCache` of `flattenedTree`.

Previously because of this line https://github.com/dequelabs/axe-core/pull/1536/files#diff-a4aba2b3f5c65ff940eb4c8060dc57dfL37, if `isFocusable` is false, the utility function will be called. We should check if a cached value exists, rather than check against `true` or `false`, given that the values of some of these cached functions are boolean.

Closes issue: NA.

## Reviewer checks


WilcoFiers:
What's the bug? Please explain.

Issue #689, opened 2018-01-19
False Positive: Form element has multiple <label> elements


jnurthen:
IMO this rule should be retired completely. Multiple labels for a form field are allowed by spec and work in most useragents. I believe it is only IE where they do not. This would be the simplest way to resolve.

However, even if it were to be kept the following should certainly not throw an error
https://codepen.io/anon/pen/MrLgKm

```html
<input type="checkbox" id="A" aria-labelledby="B">
<label class="indicator" aria-hidden="true" for="A">X</label>
<label for="A" id="B" class="ps-label">Yes/No Value</label>
```
Here there are 2 labels for the checkbox (the first is actually a CSS image in the real thing - but here I have added X for simplicity). They are both labels in order to increase the hit area for the checkbox. The first is hidden from screen readers with aria-hidden as it is not meaningful.

Even the addition of aria-labelledby on the checkbox which should override the labels does not seem to remove the aXe error.

WilcoFiers:
@jnurthen Going by the WebAIM survey, IE is still the second most used browser with screen readers. https://webaim.org/projects/screenreadersurvey7/#browsers I don't think we can drop support for IE just yet. Axe is highly configurable, so you can turn off rules / checks that don't fit your accessibility support baseline.

marcysutton:
I think we should at least check the `aria-hidden` case to see if it has any impact in IE. I can look into that today.

jnurthen:
@WilcoFiers If someone is overriding the accessible name using aria-labelledby as in this example, why is the fact that there are multiple labels relevant? Neither of the <label> elements are used in the accessible name calculation. They are ONLY here to act as a larger target area for mouse clicks. 
We would make the UI worse by changing one or both of the label elements to DIVs yet this is the only was we can get a clean aXe run on this.

dylanb:
aria-labelledby should override the double labels

Also, last I tested, VO did not respect double labels

donifer:
Sorry for reviving this thread. In @jnurthen's example one of the labels has `aria-hidden="true"` so now I'm a bit confused. 

Is that not a a valid case? Or should I also use `aria-labelledby` on top of `aria-hidden`?

meghnashah29:
I did not completely understand the conclusion on this thread. I faced the same situation today that WorldSpace attest (axe) is throwing the error "Form element has multiple <label> elements" while using same label multiple times as shown below:
<input type="checkbox" id="A" aria-labelledby="B">
[...]

PR #1538, opened 2019-05-02
fix(multiple-label): no longer raises issue when aria-labelledby overrides how AT views multiple labels

AutoSponge:
When authors use `aria-labelledby` on a labeled input, multiple labels will be consumed by AT. In cases where less than all labels are included, the excluded label must have `aria-hidden="true"`.

issue: #689 (don't autoclose, docs needed)

## Reviewer checks

- [x] Code is reviewed for security by: dylanb

### DIFF
--- a/lib/checks/label/multiple-label.js
+++ b/lib/checks/label/multiple-label.js
@@ -3,16 +3,8 @@ let labels = Array.from(document.querySelectorAll(`label[for="${id}"]`));
 let parent = node.parentNode;
 
 if (labels.length) {
-	// filter out hidden labels because they're fine
-	// except: fail first label if hidden because of VO
-	labels = labels.filter(function(label, index) {
-		if (
-			(index === 0 && !axe.commons.dom.isVisible(label, true)) ||
-			axe.commons.dom.isVisible(label, true)
-		) {
-			return label;
-		}
-	});
+	// filter out CSS hidden labels because they're fine
+	labels = labels.filter(label => axe.commons.dom.isVisible(label));
 }
 
 while (parent) {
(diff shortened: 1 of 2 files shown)


lib/checks/label/multiple-label.js
```diff
@@ -26,4 +19,16 @@ while (parent) {
...
+		axe.commons.dom.isVisible(label, true)
+	);
+	if (visibleLabels.length > 1) {
+		return true;
+	}
+	const labelledby = axe.commons.dom.idrefs(node, 'aria-labelledby');
+	return !labelledby.includes(visibleLabels[0]);
+}
+
 return labels.length > 1;
```
straker: This could be `returns false` since the `true` case is handled above

straker:
The tests pass as expected, but I got a bit lost in the logic trying to figure out was trying to be tested. Maybe a few comments explaining the whys of the logic would be helpful

lib/checks/label/multiple-label.js
AutoSponge: But then it would then pass even if the `aria-labelledby` does not include the correct idRef. I'll update with some comments.

AutoSponge: Good point. I've also added some comments and changed a variable name to make it a little more clear.

straker:
LGTM. 

Because of the extent of the rule change, I just want to make sure @WilcoFiers is Ok with the direction before we merge (see https://github.com/dequelabs/axe-core/issues/689#issuecomment-490176712).

dylanb:
Needs a change to the more info page to explain the logic

dylanb:
@jeankaplansky please take up the documentation requirement and when the ticket is created, add the link here

jeankaplansky:
> @jeankaplansky please take up the documentation requirement and when the ticket is created, add the link here

@dylanb: Please confirm: documentation requirement exists for both Attest AND axe-core?

stephenmathieson:
Merging this. We'll figure out the documentation stuff later on.

Issue #1444, opened 2019-03-18
Remove aria-required-children for role="tree"


robpaveza:
**Expectation**:

```html
<ol role="tree"></ol>
```

should generate no errors.

**Actual**: It generates the following error:

> [aria-required-children](https://dequeuniversity.com/rules/axe/3.1/aria-required-children?application=axeAPI): Certain ARIA roles must contain particular children  [WCAG 1.3.1](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships)

**Motivation**: A tree that has no leaves is still a tree, and therefore `role="treeitem"` should not be a required child of an element of `role="tree"`.

**axe-core version**: 3.0.0-alpha1, 3.1.1

smhigley:
We ran into this issue as well in a product that allows user interaction to insert tree items. The empty `role="tree"` exists before any tree items have been inserted.

There are examples of empty trees existing on desktop applications in Windows applications, so it's not a completely unknown pattern. An example might be in something like this: https://www.lightningdesignsystem.com/components/dueling-picklist/ where instead of listboxes there are selectible tree views, and all items are removed from one side.

WilcoFiers:
I agree. I think generally the aria-required-children rule isn't quite right and should probably be more of an "allowed children" type rule, similar to what we've done with the list rules.

Thank you for reporting. We'll definitely pick this up.

straker:
@WilcoFiers how exactly would you like to handle this? It seems related to #874 and #383

WilcoFiers:
@straker The aria-required-children check has an option for `reviewEmpty`, which when we put a role in there, puts the results into "needs review" rather than failing it. I think it should have the following items in it, so that they are put into needs review:

- doc-bibliography
- doc-endnotes
- list
- listbox
- table
- grid
- treegrid
- tree
- group
- rowgroup
- tablist


That would make only the following roles fail if they are left empty:
- menu
- menubar
- row

chandana7393:
Verified for the required test file, not seeing any voilations, but observed "Needs review" issue.
(Attached the screenshot) 
![need_review_no_children](https://user-images.githubusercontent.com/45285854/58472527-0f01ce00-8164-11e9-8126-0570efc334d2.PNG)
**Tested Environment:**
Axe-coconut - 3.8.0.14704v
Chrome - 74v
OS - Windows 10 64 bit

Issue #383, opened 2017-06-28
Aria listbox and list roles should be allowed to be empty


dylanb:
This example page shows multiple listboxes, some of which are empty https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox.html

These empty listbox elements should not fail the required children rule. We should create a new category of required children that equates to "required if not empty" and put the `listbox` and `list` roles into this category

marcysutton:
The URL above is throwing a 404, as an FYI. This is also relevant to #874.

dylanb:
https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-rearrangeable.html

jkodu:
@dylanb @WilcoFiers 

As per the fix that was merged for the related issue, https://github.com/dequelabs/axe-core/issues/874, empty listbox elements are flagged as incomplete rather than violation.

Does that suffice for this issue?

dylanb:
I think it is wrong to flag them as incomplete but I think that is a debate for a different forum

BrendanMcK:
Looks like the fix for #874 allows `listbox` items to be empty, but not `list` items. [code](https://github.com/dequelabs/axe-core/pull/927/files#diff-9b42f0948ee7ca9d3175ad9821a85847R5)

chandana7393:
Issues related to "Ensures elements with an ARIA role that require child roles contain them" are displaying under Needs Review.
![list_listitem_needs_review](https://user-images.githubusercontent.com/45285854/58696939-4071d800-83b6-11e9-9e8b-6c077653aa87.PNG)
**Tested Environment:**
Axe-coconut - 3.8.0.15004v
Chrome - 74v
OS - Windows 10 64 bit.

padmavemulapati:
@scurker , can you please take look at the comment @chandana7393 kept, if that is the expected behavior, we are good to close the issue.

scurker:
Per https://github.com/dequelabs/axe-core/issues/1444#issuecomment-488253893 this is correct behavior. Only the following roles should fail if empty:

* menu
* menubar
* row

padmavemulapati:
Thank you @scurker for clarifying this, so we are good to close this tickets as it is giving expected results.

PR #1547, opened 2019-05-06
feat(rule): add additional elements to check for incomplete with required children

scurker:
* Added additional elements per https://github.com/dequelabs/axe-core/issues/1444#issuecomment-488253893
* Added check for element to fall under _Needs Review_ only when an element is empty or has no owned nodes.
* Added missing roles to aria lookup table where owner relationship was missing per [D.Pub Aria 1.0](https://www.w3.org/TR/dpub-aria-1.0/) and [Aria 1.1](https://www.w3.org/TR/wai-aria-1.1/).

Closes issue: #1444 
Closes issue: #383

## Reviewer checks

### DIFF
--- a/lib/checks/aria/required-children.js
+++ b/lib/checks/aria/required-children.js
@@ -39,20 +39,23 @@ function ariaOwns(nodes, role) {
 }
 
 function missingRequiredChildren(node, childRoles, all, role) {
-	var i,
-		l = childRoles.length,
+	var index,
+		length = childRoles.length,
 		missing = [],
 		ownedElements = idrefs(node, 'aria-owns');
 
-	for (i = 0; i < l; i++) {
-		var r = childRoles[i];
-		if (owns(node, virtualNode, r) || ariaOwns(ownedElements, r)) {
+	for (index = 0; index < length; index++) {
+		var childRole = childRoles[index];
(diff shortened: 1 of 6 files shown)


jkodu:
@scurker are we now saying raise a `needs review` for an element with no owned elements, as against a failure? eg: `<ul></ul>`?

We are authoring couple of rules in ACT Rules, for owned elements & context elements, and ideally would like to match our implementation with expected results of test cases from here - https://github.com/act-rules/act-rules.github.io/pull/245/files

lib/checks/aria/required-children.js
```diff
@@ -109,7 +112,12 @@ if (!missing) {
 
 this.data(missing);
 
-if (reviewEmpty.includes(role)) {
+// Only review empty nodes when a node is both empty and does not have an aria-owns relationship
+if (
+	reviewEmpty.includes(role) &&
+	node.children.length === 0 &&
+	idrefs(node, 'aria-owns').length === 0
```
WilcoFiers: Can you add a test to show that this returns undefined if there are no children, but false when it has children that don't have the correct rule?

scurker: 👍

PR #1548, opened 2019-05-06
ci: run test tasks with appveyor-retry to allow flaky tests to re-run

scurker:
* Run test scaffolding commands isolated
* Split up test commands between phantom/webdriver to allow for quicker runs if something flakes out
* Only run `webdriver:ie` tests

I did a bit of testing around this. Ran 40+ iterations on CI after the change and there was only a single failure due to flaky tests being unable to run after 3 attempts (the other failures were due to random appveyor issues). This change could potentially make tests run longer (up to ~30 minutes), but will provide greater confidence levels that there aren't breaking windows changes.

It _does not_ resolve any of the underlying issues that may be contributing to potential flakiness on Windows.

## Reviewer checks

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -15,9 +15,12 @@ test_script:
   - node --version
   - npm --version
   - npm run build
+  # Run test scaffolding
+  - grunt testconfig fixture
   # A few tests are flaky, causing occasional random failures
   # When we have time, we should investigate to see if we can eliminate these flakes
-  - appveyor-retry npm run test
+  - appveyor-retry grunt connect mocha
+  - appveyor-retry grunt connect test-webdriver:ie
   - npm run test:examples
 
  # Don't actually build.
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -16,11 +16,11 @@ test_script:
   - npm --version
   - npm run build
   # Run test scaffolding
-  - grunt testconfig fixture
+  - npx grunt testconfig fixture
   # A few tests are flaky, causing occasional random failures
   # When we have time, we should investigate to see if we can eliminate these flakes
-  - appveyor-retry grunt connect mocha
-  - appveyor-retry grunt connect test-webdriver:ie
+  - appveyor-retry npx grunt connect mocha
+  - appveyor-retry npx grunt connect test-webdriver:ie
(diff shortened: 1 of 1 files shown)


straker:
Since my knowledge of grunt is limited, are the `file-exists` and `parallel` tasks needed for the appyveyor tests?

scurker:
It looks like `file-exists` only is used for verifying that language files exist? It doesn't appear to be necessary for running any of the tests.

`parallel` is used to run the webdriver tests in parallel. Since only IE is being run at the moment, it's not needed.

stephenmathieson:
I disagree with @jkodu .any additional coverage we can get here is a win in my book.

scurker:
@jkodu Ideally we should be testing _all_ browsers on Windows, but as it stands we currently aren't testing IE at all. I just happened to stumble upon `appveyor-retry` last week and wanted to see if I could improve our existing coverage. Since some of the Chrome tests can be flaky as well, I wanted to isolate the tests just to IE to at least guarantee some coverage there.

scurker:
The biggest downside to this, is since `appveyor-retry` is prefixed against a whole test suite, Appveyor tests will no longer fail fast since the full suite of tests will run 3 times before finally erring out.

stephenmathieson:
@scurker since Jey is working on updating/modernizing the test suite, and @dequedoug hasn't been able to setup an org-owned AppVeyor account yet, I'm hesitant to let this land. I think we should shelf this until one of the 2 aforementioned projects is completed.

What do you think?

scurker:
@stephenmathieson I'd be okay with holding off until @dequedoug has an organization account as it looks like that's something he's currently chasing. As for modernizing the test suite, I'd be a little concerned we'd be missing some immediate benefit as that's a much larger task to tackle.

stephenmathieson:
> holding off until @dequedoug has an organization account as it looks like that's something he's currently chasing

👍 let's wait then.

> As for modernizing the test suite, I'd be a little concerned we'd be missing some immediate benefit as that's a much larger task to tackle.

Fair enough!

scurker:
Members of the _html-team_ should now have access to Appveyor builds, so we can move forward with this now.

Issue #1534, opened 2019-05-01
Add a rawEnv reporter


WilcoFiers:
This reporter would return the raw results and the env variables in a single object. Something like this:

```js
axe.addReporter('rawEnv', function(raw, options, callback) {
	if (typeof options === 'function') {
		callback = options;
		options = {};
	}
	const env = helpers.getEnvironmentData();
	callback({ raw, env });
});

```

padmavemulapati:
QA Reviewed.

PR #1556, opened 2019-05-09
feat(reporter): adds the rawEnv reporter which wraps raw and env data

AutoSponge:
`rawEnv` reporter calls the callback with `{raw, env}` where `raw` is the output of the `raw` reporter and `env` is the output of `helpers.getEnvironmentData()`.

Closes issue: #1534

## Reviewer checks

### DIFF
--- a/lib/core/reporters/raw-env.js
+++ b/lib/core/reporters/raw-env.js
@@ -1,0 +1,12 @@
+/*global helpers */
+axe.addReporter('rawEnv', function(results, options, callback) {
+	if (typeof options === 'function') {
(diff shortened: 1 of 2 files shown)


straker:
Could you add a test to ensure the `raw.env` contains some expected properties? Right now you could swap out the env call for an empty object and it would still pass.

AutoSponge:
@straker that doesn't cover any of my code.

WilcoFiers:
I'm created an issue #1560 for our lack of documentation about available reporters.

Issue #1385, opened 2019-02-20
Failing Safari tests


jkodu:
The below tests are failing in Safari, on develop branch. Our CI does not test on safari

![image](https://user-images.githubusercontent.com/20978252/53123027-05a70b00-3550-11e9-914d-d55371868a6e.png)

![image](https://user-images.githubusercontent.com/20978252/53123047-135c9080-3550-11e9-863e-f53f7263915c.png)

straker:
Turns out Safari has a bug where `<main>` does not support `attachShadow` even though it's a valid `attachShadow` target https://bugs.webkit.org/show_bug.cgi?id=197726

PR #1557, opened 2019-05-09
test: fix test for Safari

straker:
There were two Safari browser bugs that I had to work around, and one difference in floating point precision, but I got the tests working locally for Safari. Unfortunately I couldn't get it to work in CircleCI, so we'll have to come back to that in a later pr.  

Closes: #1385 

## Reviewer checks


.circleci/config.yml
```diff
@@ -65,6 +65,16 @@ jobs:
       - <<: *restore_dependency_cache
       - run: npm run test:examples
 
+  test_macos:
+    macos:
+      xcode: "10.0"
+    steps:
+      - checkout
+      - run: npm install
```
scurker: You should be able to just restore the dependency cache here.

```
- checkout
- <<: *restore_dependency_cache
```

.circleci/config.yml
```diff
@@ -65,6 +65,16 @@ jobs:
...
+      - run: npm install -g grunt-cli
+      - run: npm run build
```
scurker: Not 100% sure this is required.

.circleci/config.yml
straker: Since the test is on a different platform, I'm not sure the cache is available to restore.

lib/commons/color/get-background-color.js
```diff
@@ -118,6 +118,15 @@ color.filteredRectStack = function filteredRectStack(elm) {
...
 		let isSame;
 
+		// Safari v12.1 does not include labels as part of elementsFromPoint()
+		// if they wrap an input element (UNLESS the label has a background
+		// color). this results in two different rectStacks (since
+		// elm.getClientRects() returns two rects for the element) which
+		// returns null as isSame is false. we can fix this by adding in the
+		// missing label to the boundingStack before checking for isSame
+		// @see https://bugs.webkit.org/show_bug.cgi?id=197743
+		includeMissingElements(boundingStack, elm);
```
WilcoFiers: This is already happening in `getBackgroundStack`. I don't quite understand why you'd call this twice.

straker: So the `includeMissingElements` call in `getBackgroundStack` is not called since `filteredRectStack` returns `null` because it's missing the label. I tried removing the `includeMissingElements` in `getBackgroundStack` but that failed a lot of tests.

.circleci/config.yml
```diff
@@ -65,6 +65,15 @@ jobs:
       - <<: *restore_dependency_cache
       - run: npm run test:examples
 
+  # test_macos:
```
stephenmathieson: I probably missed a thread/conversation somewhere. Why is this commented out?

straker: To save it into the git history for later use. It can be removed any time after this is merged.

stephenmathieson: Please don't do that. We don't need to version commented out code.

If you're worried about this stuff getting "lost", just create a branch with the changes in it, or host as `.patch` file as a Gist for future use.

Issue #1460, opened 2019-03-26
Axe throwing error for role="figure"


scottaohara:
## Expectation:
Not all browser and screen reader combinations correctly expose the `figure` element and its caption.  Using `role="figure"` can help alleviate this gap for the time being.

People should be able to use `role="figure"` without receiving an error.

## Actual: 
axe (and lighthouse) return "ARIA roles used must conform to valid values" when using a `role="figure"` on an element.

e.g.
```html
<div role="figure" aria-label="text goes here"> 
  <img src="my-image.jpg" alt="alt text here"> 
  <p>text goes here.</p> 
</div>
```

## Motivation: 
The rule should be corrected because [`role="figure"` is a valid role](https://www.w3.org/TR/wai-aria-1.1/#figure).

<pre><code>axe-core version: 3.2.2
axe-devtools 3.8.0
</code></pre>

WilcoFiers:
@scottaohara Do you have test information about how your example works in things like VO+Safari, NVDA+FF and JAWS+IE? If this solution does indeed work in those, I fully agree that this needs to be changed in axe.

scottaohara:
As a matter of fact :) 
https://www.scottohara.me/blog/2019/01/21/how-do-you-figure.html 
where I talk about the support for `<figure>` (and `role=figure`) quite a bit.  tldr; `role=figure` is necessary to expose figures in IE11+JAWS, and additionally is exposed correctly when testing with VO+Safari.

Additionally I have the following demo: https://scottaohara.github.io/testing/automated/ for a simple test case.

I've opened an issue with NVDA regarding [`figure` and `role=figure`](https://github.com/nvaccess/nvda/issues/9177) and their lack of communicating the element/role.

With that said, I'd still find it odd to flag it as an "invalid role" when it's not an invalid role. 

A best practice or warning of "this role may not be supported in all browsers/AT" sure, I get that.  But if you take that stance with `role=figure` then why does the `<meter>` element, for example, not cause any flags?  It has abysmal accessibility support.

chandana7393:
Axe-coconut is not throwing error for role="figure"
![figure_role](https://user-images.githubusercontent.com/45285854/58697601-677cd980-83b7-11e9-873d-76e4c305d754.PNG)
**Tested Environment:**
Axe-coconut - 3.8.0.15004v
Chrome - 74v
OS - Windows 10 64 bit.

PR #1558, opened 2019-05-09
fix(aria-allowed-attr): allow figure role

straker:
The `figure` role is needed to properly read the `figcaption` element in some browser and AT combinations. See https://www.scottohara.me/blog/2019/01/21/how-do-you-figure.html.

Linked issue: #1460

## Reviewer checks


WilcoFiers:
@straker did you test this in JAWS? According to Steve Faulkner's docs this doesn't work:
https://freedomscientific.github.io/VFO-standards-support/aria.html#index-aria-figure

straker:
I did not personally. I relied on the extensive testing of Scotts article which showed JAWS needing it in ie11.

> IE11 requires the use of role="figure" and an aria-label or aria-labelledby pointing to the figcaption to mimic native announcements. It’s not surprising that IE11 doesn’t support the native element, as HTML5 Accessibility’s IE11 browser rating will never improve. But at least ARIA can provide the semantics.

I think there are two different use cases that we are trying to cover. `role=figure` on any element may not be supported (Steve Faulkner's test), but `role=figure` on a `<figure>` element is needed in some browsers/AT combinations in order to read the `figcaption` (Scott's test).

I agree with Scott that we should just straight out flag it as unsupported, but maybe we just need to do what we did with `aria-roledescription` and say where it's supported.

I can do some testing for all cases and come up with a support matrix for when/where it is supported.

WilcoFiers:
My main question, which wasn't obvious from Scott's article, was whether or not that figure element is announced as a figure. I'm sure if you put an aria-label(ledby) on it, it'll be announced, but is it announced as a "figure" or "graphic" or whatever.

straker:
Here's the results of my testing with `role=figure` https://codepen.io/straker/pen/zQoQYr. Any support element/browser/AT combination the browser announced the element as `figure` role.   In terms of who announced the caption of the 4 different figures, VO/Safari announced it natively while JAWS/IE11 only announced the caption with `aria-labelledby`. These results are consistent with Scotts findings.

So in terms of what we want to allow. VO/Safari and JAWS/IE11 (with 'role=figure' and 'aria-lablledby') are the only combinations that announce the figure role and the caption. So support isn't great, but there is a benefit for IE11 to add value even if no other browsers get the same value.

I propose we only support `role=figure` on `<figure>` elements as a fallback role.

doc/aria-supported.md
```diff
@@ -10,7 +10,7 @@ For a detailed description about how accessibility support is decided, see [How
 
 | aria-role | axe-core support |
 | --------- | ---------------- |
-| figure    | No               |
+
```
WilcoFiers: Now that this table is empty, can you please remove it?

Issue #1552, opened 2019-05-07
Add required methods to VirtualNode to replace actualNode lookups in QSA


straker:

chandana7393:
Dev Task.

padmavemulapati:
QA Reviewd.

PR #1562, opened 2019-05-10
feat(qsa, flatten-tree): abstract Node and Element apis in virtual tree

straker:
I wanted the abstraction layer on the virtual node to be consistent and predictable so you wouldn't have to look at the virtual node implementation to figure out what the thing was called you wanted. I went with prefixing all api names with `element` and then capitalizing the first letter. Someone suggested mimicking jQuery api names, but they don't have any api names for low-level properties like `nodeType`, and `nodeName`, which would mean an inconsistency in naming convention.

Qsa tests used a mock virtual tree, which made it really difficult to test changes to qsa that depend on the virtual tree implementation. So I changed it to use a real virtual tree and then added a test to ensure qsa does not look at the `actualNode` property (ensuring a DOM-less run).

Linked issue: #1552

## Reviewer checks


lib/core/utils/flattened-tree.js
```diff
@@ -46,6 +46,18 @@ function virtualDOMfromNode(node, shadowId) {
 				);
 			}
 			return vNodeCache._tabbableElements;
+		},
+
+		// abstract Node and Element APIs so we can run axe in DOM-less
+		// environments. these are static properties in the assumption
+		// that axe does not change any of them while it runs.
+		elementNodeType: node.nodeType,
+		elementNodeName: node.nodeName,
```
WilcoFiers: I think we should normalise this, do a `.toLowerCase()` here.

lib/core/utils/flattened-tree.js
```diff
@@ -46,6 +46,18 @@ function virtualDOMfromNode(node, shadowId) {
 				);
 			}
 			return vNodeCache._tabbableElements;
+		},
+
+		// abstract Node and Element APIs so we can run axe in DOM-less
+		// environments. these are static properties in the assumption
+		// that axe does not change any of them while it runs.
+		elementNodeType: node.nodeType,
```
WilcoFiers: VirtualNode objects always reference Element instances. This will always be `nodeType === 1`. I don't think we need this.

lib/core/utils/flattened-tree.js
```diff
@@ -46,6 +46,18 @@ function virtualDOMfromNode(node, shadowId) {
...
+		elementNodeName: node.nodeName,
+		elementClassName: node.className,
+		elementId: node.id,
+
+		elementGetAttribute(attrName) {
```
WilcoFiers: can we do `attr()` rather than `elementGetAttribute()`. It's pretty verbose.

lib/core/utils/flattened-tree.js
```diff
@@ -46,6 +46,18 @@ function virtualDOMfromNode(node, shadowId) {
...
 			}
 			return vNodeCache._tabbableElements;
+		},
+
+		// abstract Node and Element APIs so we can run axe in DOM-less
+		// environments. these are static properties in the assumption
+		// that axe does not change any of them while it runs.
+		elementNodeType: node.nodeType,
+		elementNodeName: node.nodeName,
+		elementClassName: node.className,
```
WilcoFiers: What about using a `hasClass()` type method instead of directly accessing the `className` property?

lib/core/utils/flattened-tree.js
straker: This turns out not to be true. Text elements make it into the virtual tree and excluding them as such result in tons of errors in the tests

straker:
With `hasClass` we were now adding functions with logic, so I decided to create a VirtualNode 
constructor function so we could have tests around the logic. I also moved the getter functions for isFocusable and tabbableElements to actual functions so they could be parted of the shared prototype rather than one off for each virtual node.

lib/core/base/virtual-node.js
```diff
@@ -0,0 +1,71 @@
+/**
+ * Wrap the real node and provide list of the flattened children
+ *
+ * @param node {Node} - the node in question
+ * @param shadowId {String} - the ID of the shadow DOM to which this node belongs
+ */
+function VirtualNode(node, shadowId) {
```
stephenmathieson: Any reason not to do something like:

```js
class VirtualNode {
  constuctor(node, shadowId) {
    // ...
  }

  hasClass(className) {
    // ...
  }
}
```

We've [had Babel here since 2016](https://github.com/dequelabs/axe-core/commit/bcf283d72a17f01fb7c9bbfefe0243d69f65a4eb#diff-e56633f72ecc521128b3db6586074d2c), so there shouldn't be any browser compatibility issues.

lib/core/base/virtual-node.js
```diff
@@ -0,0 +1,71 @@
...
+}
+
+/**
+ * Determine if the actualNode has the given class name.
+ * @see https://j11y.io/jquery/#v=2.0.3&fn=jQuery.fn.hasClass
+ * @param {String} className - The class to check for.
+ * @return {Boolean} True if the actualNode has the given class, false otherwise.
+ */
+VirtualNode.prototype.hasClass = function(className) {
+	if (typeof this.actualNode.className !== 'undefined') {
```
stephenmathieson: Please refactor this to keep the happy-path closer to the left.

```js
if (typeof this.actualNode.className === 'undefined') {
  return
}

const selector = ' ' + className +' '
return ' ' + this.actualNode.className + ' '.indexOf(selector) >= 0
```

lib/core/base/virtual-node.js
```diff
@@ -0,0 +1,71 @@
...
+	if (axe._cache.get('nodeMap')) {
+		axe._cache.get('nodeMap').set(node, this);
+	}
+}
+
+/**
+ * Determine if the actualNode has the given class name.
+ * @see https://j11y.io/jquery/#v=2.0.3&fn=jQuery.fn.hasClass
+ * @param {String} className - The class to check for.
+ * @return {Boolean} True if the actualNode has the given class, false otherwise.
```
stephenmathieson: This function does **not** always return a boolean.

lib/core/base/virtual-node.js
stephenmathieson: Also, it appears we expect `actualNode.className` to be of type `string`. This will cause nasal demons:

```js
this.actualNode.className = { hello: false }
this.hasClass('potato')
```

lib/core/base/virtual-node.js
```diff
@@ -0,0 +1,71 @@
...
+	}
+};
+
+/**
+ * Get the value of the given attribute name.
+ * @param {String} attrName - The name of the attribute.
+ * @returns {String} The value of the attribute
+ */
+VirtualNode.prototype.attr = function(attrName) {
+	if (typeof this.actualNode.getAttribute !== 'undefined') {
```
stephenmathieson: Please refactor to keep the happy path on the left.

lib/core/base/virtual-node.js
```diff
@@ -0,0 +1,71 @@
...
+	if (typeof this.actualNode.className !== 'undefined') {
+		let selector = ' ' + className + ' ';
+		return (' ' + this.actualNode.className + ' ').indexOf(selector) >= 0;
+	}
+};
+
+/**
+ * Get the value of the given attribute name.
+ * @param {String} attrName - The name of the attribute.
+ * @returns {String} The value of the attribute
```
stephenmathieson: This function does not always return a string. If `this.actualNode.getAttribute` is `undefined`, then doing this will cause exceptions:

```js
this.attr('banana').indexOf('ana')
```

lib/core/utils/qsa.js
```diff
@@ -6,32 +6,27 @@ var matchExpressions = function() {};
...
-	);
+	return !exp.classes || exp.classes.every(cl => node.hasClass(cl.value));
 }
 
 function matchesAttributes(node, exp) {
 	return (
 		!exp.attributes ||
 		exp.attributes.reduce((result, att) => {
-			var nodeAtt = node.getAttribute(att.key);
+			var nodeAtt = node.attr(att.key);
```
stephenmathieson: The following line could raise an exception.

lib/core/base/virtual-node.js
stephenmathieson: See https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88 for a good article/primer about happy-path.

lib/core/base/virtual-node.js
straker: Nope, just following the other base files style. If we're comfortable having our first (and only for now) class, I can do that.

lib/core/base/virtual-node.js
straker: `className` always returns a string, even if you set it to an object it will return `[object Object]` since it reflects the class attribute value.

lib/core/utils/qsa.js
straker: None that it wouldn't have raised before

test/core/utils/qsa.js
```diff
@@ -1,116 +1,48 @@
...
+		'<li role="tab" id="one"></li>' +
+		'<li role="button" id="one"></li>' +
+		'</ul>' +
+		'</div>' +
+		'</body>';
+
+	// remove the head node
+	var head = html.querySelector('head');
+	if (head) {
+		head.remove();
```
scurker: [Not supported in IE.](https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/remove)

You could alternatively use `head.parentNode.removeChild(head)` for IE.

straker: I had no idea that wasn't supported in IE. I've always thought it was an older API. Thanks for pointing that out.

lib/core/base/virtual-node.js
```diff
@@ -0,0 +1,79 @@
...
+		if (axe._cache.get('nodeMap')) {
+			axe._cache.get('nodeMap').set(node, this);
+		}
+	}
+
+	/**
+	 * Determine if the actualNode has the given class name.
+	 * @see https://j11y.io/jquery/#v=2.0.3&fn=jQuery.fn.hasClass
+	 * @param {String} className - The class to check for.
+	 * @return {Boolean|undefined} True if the actualNode has the given class, false if it does not have the given class, and undefined if the actualNode does not have a className property.
```
stephenmathieson: Why return `undefined` rather than `false`? Having multiple return types (even tho documented) for a function prefixed with `has` seems like a very strange pattern.

lib/core/base/virtual-node.js
```diff
@@ -0,0 +1,79 @@
...
+		}
+
+		let selector = ' ' + className + ' ';
+		return (' ' + this.actualNode.className + ' ').indexOf(selector) >= 0;
+	}
+
+	/**
+	 * Get the value of the given attribute name.
+	 * @param {String} attrName - The name of the attribute.
+	 * @returns {String|undefined} The value of the attribute or undefined if the actualNode does not have a getAttribute function
```
stephenmathieson: Why not `return null` (like the native `getAttribute`) rather than `undefined`?

lib/core/base/virtual-node.js
```diff
@@ -0,0 +1,79 @@
...
+	 */
+	attr(attrName) {
+		if (typeof this.actualNode.getAttribute === 'undefined') {
```
stephenmathieson: We expect `.getAttribute` to be a function. If it's somehow set to `{ foo: 'bar' }`, L#55 will throw. Why not do:

```js
if (typeof this.actualNode.getAttribute !== 'function') {
  return null
}
```

lib/core/base/virtual-node.js
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	/**
+	 * Determine if the actualNode has the given class name.
+	 * @see https://j11y.io/jquery/#v=2.0.3&fn=jQuery.fn.hasClass
+	 * @param {String} className - The class to check for.
+	 * @return {Boolean|undefined} True if the actualNode has the given class, false if it does not have the given class, and undefined if the actualNode does not have a className property.
+	 */
+	hasClass(className) {
+		if (typeof this.actualNode.className === 'undefined') {
```
stephenmathieson: Why don't we check that `.className` is the type we expect it to be?

lib/core/base/virtual-node.js
```diff
@@ -0,0 +1,79 @@
...
+		this.actualNode = node;
+
+		this._isHidden = null; // will be populated by axe.utils.isHidden
+		this._cache = {};
+
+		// abstract Node and Element APIs so we can run axe in DOM-less
+		// environments. these are static properties in the assumption
+		// that axe does not change any of them while it runs.
+		this.elementNodeType = node.nodeType;
+		this.elementNodeName = node.nodeName.toLowerCase();
```
stephenmathieson: IIRC @jkodu said there was some (undocumented) convention in this codebase to always use upper-case node names. I don't know if this matters (or is true), but if is, we should follow convention.

lib/core/base/virtual-node.js
straker: I guess it depends on how much we care / expect `actualNode` to not be a DOM node. I just need to check that it exists before I call it, but like everywhere else in axe we assume it's a true node and the property won't be overridden with nonsense data.

stephenmathieson: I don't see why adding a guard to protect against unknowns is a bad idea, especially since the `typeof` check is already there.

lib/core/base/virtual-node.js
straker: Wilco asked for this so I'm going to assume the undocumented convention is now set to lowerCase. https://github.com/dequelabs/axe-core/pull/1562#discussion_r283401772

lib/checks/keyboard/focusable-content.js
```diff
@@ -2,7 +2,7 @@
  * Note:
  * Check if given node contains focusable elements (excluding thyself)
  */
-const tabbableElements = virtualNode.tabbableElements;
+const tabbableElements = virtualNode.tabbableElements();
```
WilcoFiers: This potentially breaks custom rules. I don't think we can change this from being a getter.

lib/checks/keyboard/focusable-element.js
```diff
@@ -4,7 +4,7 @@
  * - if element is focusable
  * - if element is in focus order via `tabindex`
  */
-const isFocusable = virtualNode.isFocusable;
+const isFocusable = virtualNode.isFocusable();
```
WilcoFiers: As above, this could break custom rules.

lib/core/base/virtual-node.js
```diff
@@ -0,0 +1,79 @@
...
+	 * @param {String} className - The class to check for.
+	 * @return {Boolean} True if the actualNode has the given class, false otherwise.
+	 */
+	hasClass(className) {
+		if (typeof this.actualNode.className !== 'string') {
+			return false;
+		}
+
+		let selector = ' ' + className + ' ';
+		return (' ' + this.actualNode.className + ' ').indexOf(selector) >= 0;
```
WilcoFiers: Is the className property normalised? As in, if there's a tab or some other whitespace character between the class names, will this work?

stephenmathieson: Sounds like a test should be added :wink:

lib/checks/keyboard/focusable-content.js
straker: Is this something we could change in 4.0? As in, could we add back the 2 getters but keep the functions internally, and then we remove the getters in 4.0? Having them on the class itself rather than getters can save quite a bit of memory for larger pages, especially as we potentially add more of these type of functions to the virtual node.

PR #1570, opened 2019-05-15
test: fix tests failing in IE11

### LINKED ISSUE #1384: Failing IE11 tests
Was testing some updates on Windows 10, IE 11 & in today's develop branch, these tests are failing. Our CI does not test on IE11

image

straker:
[Phantomjs does not support the `transform` css property](https://github.com/ariya/phantomjs/issues/12779) and so all the CSS orientation lock tests were actually false positives with the one failure being correct. Adding `-webkit-transform` fixes the problem.

IE11 also parses CSS rule property order differently so the `testutil` rule string compare would fail. Changed it to create a style on the page and compare the same properties. IE11 also didn't add a space between the CSS media properties of `(layout: portrait)`

Closes #1384

## Reviewer checks


lib/checks/mobile/css-orientation-lock.js
```diff
@@ -87,8 +87,9 @@ Object.keys(rulesGroupByDocumentFragment).forEach(key => {
 				return;
 			}
 
-			// check if transform style exists
-			const transformStyleValue = cssRule.style.transform || false;
+			// check if transform style exists (don't forget vendor prefixes)
+			const transformStyleValue =
+				cssRule.style.transform || cssRule.style.webkitTransform || false;
```
stephenmathieson: Looks like you've covered `-webkit-transform`, but what about `-ms-transform`?

test/checks/mobile/css-orientation-lock.js
```diff
@@ -23,11 +21,11 @@ describe('css-orientation-lock tests', function() {
 		MEDIA_STYLE_ORIENTATION_WITHOUT_TRANSFORM:
 			'@media screen and (min-width: 1px) and (max-width: 2000px) and (orientation: portrait) { #mocha { color: red; } }',
 		MEDIA_STYLE_ORIENTATION_WITH_TRANSFORM_NOT_ROTATE:
-			'@media screen and (min-width: 1px) and (max-width: 3000px) and (orientation: landscape) { #mocha { transform: translateX(10px); } }',
+			'@media screen and (min-width: 1px) and (max-width: 3000px) and (orientation: landscape) { #mocha { transform: translateX(10px); -webkit-transform: translateX(10px); } }',
```
stephenmathieson: Please add a test using `-ms-transform`.

lib/checks/mobile/css-orientation-lock.js
straker: Good catch, forgot about IE9

test/checks/mobile/css-orientation-lock.js
straker: We found out Phantomjs doesn't support `-ms-transform` since it uses a webkit engine under the hood, so we can't test this but we can add support for it in the code.

test/checks/mobile/css-orientation-lock.js
```diff
@@ -221,29 +219,26 @@ describe('css-orientation-lock tests', function() {
 	});
 
 	// This currently breaks in IE11
```
scurker: Is this still true? If not it should be removed.

test/testutils.js
```diff
@@ -338,10 +338,25 @@ testUtils.assertStylesheet = function assertStylesheet(
...
+		var testSheet = document.styleSheets[document.styleSheets.length - 1];
+		var sheetRule = sheet.cssRules[0];
+		var testRule = testSheet.cssRules[0];
+
+		try {
+			for (var i = 0; i < testRule.style.length; i++) {
+				var property = testRule.style[i];
+				assert.equal(sheetRule.style[property], testRule.style[property]);
+			}
+		} finally {
```
scurker: Do we really want a finally without a catch here?

straker: I want the exception to be thrown for the assert, but I always want to clean up the style.

Issue #1501, opened 2019-04-17
aria-valuenow should not fail aria-required-attr when there is a fallback value


WilcoFiers:
The following shouldn't fail:

```html
<input type="range" role="slider" />
```

There are similar cases. Basically any element that has a value (see `commons.text.formControlValue`) should be allowed as an alternative for `aria-requirednow`.

straker:
Should an empty value fail when the input is not populated?

```
<input type="number" role="spinbutton" <="" div="">
```

chandana7393:
Tested, working as expected.
![no_violations](https://user-images.githubusercontent.com/45285854/58951276-4d276f00-87ae-11e9-8fcc-efec77422a14.PNG)

**Tested Environment:**
Axe-coconut - 3.8.0.15604v
Chrome - 74v
OS - Windows 10 64 bit.

PR #1579, opened 2019-05-20
fix(aria-required-attr): allow aria-valuenow to pass on elements with value

straker:
Because `<input type="number" role="spinbutton">` should pass even though it has an empty string value `''`, I had to modify the `formControlValue` function to differentiate between a form control with a value (even if empty string) and one that didn't.

After trying a few things, this solution was the least worst option. I wanted to create a method that returned the true value property (with no empty string default). But since each of the form control value methods are public, I would have had to create a similar method for each that just returned the true value. That resulted in tons of duplicate tests to test the default value and the text method that returned an empty string.

For axe 4.0, I would like to remove the `text.formControlValueMethods` from the public api and only test the `text.formControlValue` function directly. That would allow us to clean this and the unit tests up to make a cleaner api. As it stands, the unit tests mostly test the individual methods but not the main `formControlValue` method that does all the logic. So if the main method had an error but the individual functions were correct, we would never know.

Closes: #1501

## Reviewer checks

### DIFF
--- a/lib/commons/forms/is-aria-range.js
+++ b/lib/commons/forms/is-aria-range.js
@@ -1,0 +1,14 @@
+/* global forms */
+const rangeRoles = ['progressbar', 'scrollbar', 'slider', 'spinbutton'];
+
+/**
+ * Determines if an element is an aria range element
+ * @method isAriaRange
+ * @memberof axe.commons.forms
+ * @param {Element} node Node to determine if aria range
+ * @returns {Bool}
+ */
+forms.isAriaRange = function(node) {
+	const role = axe.commons.aria.getRole(node, { noImplicit: true });
+	return rangeRoles.includes(role) && node.hasAttribute('aria-valuenow');
+};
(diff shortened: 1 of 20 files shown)


lib/commons/forms/is-aria-range.js
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * Determines if an element is an aria range element
+ * @method isAriaRange
+ * @memberof axe.commons.forms
+ * @param {Element} node Node to determine if aria range
+ * @returns {Bool}
+ */
+forms.isAriaRange = function(node) {
+	const role = axe.commons.aria.getRole(node, { noImplicit: true });
+	return rangeRoles.includes(role) && node.hasAttribute('aria-valuenow');
```
WilcoFiers: I don't think we should include the `hasAttribute` test here. Even without aria-valuenow, it's still an aria range element. This check is going to make reuse of this function problematic. Better to move the attribute check part outside this function IMO.

Issue #1559, opened 2019-05-10
Multiple-labels rule not correct in shadow DOM


WilcoFiers:
Noticed this line:
https://github.com/dequelabs/axe-core/blob/develop/lib/checks/label/multiple-label.js#L2

*Expectation*: This rule checks for labels in the same DOM tree that the form field exists in

*Actual*: The rule checks the light DOM for labels with the same ID.

chandana7393:
Tested, working as expected.
![multi_label_shadow_dom](https://user-images.githubusercontent.com/45285854/62131135-3148d100-b2f8-11e9-8eb9-a5a355929eca.PNG)
**Tested Environment:**
Attest - 2.5.1.20021v
Axe - 3.8.1.19820v
Axe-coconut - 3.8.1.21104v
Chrome - 75.0.3770.142v
OS - Windows 10 64 bit.

PR #1584, opened 2019-05-22
fix(multiple-label): considers explicit labels in the same shadow tree

AutoSponge:
If the node is not contained in the ownerDocument it must be in shadow (open). Therefore, we climb the parentNode path to the first document fragment. Closed shadow is not supported. This fixes an issue where a single shadow tree held an input and valid label(s). Crossing shadow tree/document boundaries can not produce an accessible name.

Closes issue: #1559

## Reviewer checks

### DIFF
--- a/lib/checks/label/multiple-label.js
+++ b/lib/checks/label/multiple-label.js
@@ -1,6 +1,8 @@
 const id = axe.utils.escapeSelector(node.getAttribute('id'));
-let labels = Array.from(document.querySelectorAll(`label[for="${id}"]`));
 let parent = node.parentNode;
+let root = axe.commons.dom.getRootNode(node);
+root = root.documentElement || root;
+let labels = Array.from(root.querySelectorAll(`label[for="${id}"]`));
 
 if (labels.length) {
 	// filter out CSS hidden labels because they're fine
--- a/test/checks/label/multiple-label.js
+++ b/test/checks/label/multiple-label.js
@@ -2,7 +2,7 @@ describe('multiple-label', function() {
 	'use strict';
 
 	var fixture = document.getElementById('fixture');
-
+	var shadowSupported = axe.testUtils.shadowSupport.v1;
 	var checkContext = axe.testUtils.MockCheckContext();
 
 	afterEach(function() {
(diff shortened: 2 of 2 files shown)


lib/checks/label/multiple-label.js
```diff
@@ -1,12 +1,21 @@
 const id = axe.utils.escapeSelector(node.getAttribute('id'));
-let labels = Array.from(document.querySelectorAll(`label[for="${id}"]`));
 let parent = node.parentNode;
+let root = document;
+if (!axe.utils.contains(node.ownerDocument, node)) {
```
AutoSponge: I tried `.contains` which fails in the IE integration test (which is strange because it's supported in IE9). So, I changed to the util, which fails in the unit tests. Any ideas?

test/checks/label/multiple-label.js
```diff
@@ -176,4 +176,23 @@ describe('multiple-label', function() {
...
+
+	(shadowSupported ? it : xit)(
+		'should consider labels in the same document/shadow tree',
+		function() {
+			fixture.innerHTML = '<div id="target"></div>';
+
+			var target = document.querySelector('#target');
+			var shadowRoot = target.attachShadow({ mode: 'open' });
+			shadowRoot.innerHTML =
+				'<input id="myinput"></input><label for="myinput">normal</label>';
```
WilcoFiers: Closing input tags (`</input>`) are invalid HTML, please use '<input id="myinput" />`

test/checks/label/multiple-label.js
```diff
@@ -176,4 +176,23 @@ describe('multiple-label', function() {
...
+			var shadowTarget = target.shadowRoot;
+			assert.isFalse(
```
WilcoFiers: Add a second label, and test that this returns true. Right now, this test only proves that there weren't two or more labels anywhere in the DOM.

chandana7393:
Verified behaving as expected.
image
**Tested Environment:**
Axe-coconut - 3.8.0.14704v
Chrome - 74v
OS - Windows 10 64 bit.

PR #1594, opened 2019-05-28
feat(run-virtual-rule): new api to run rules using only virtual nodes

straker:
This api allows a user to run a rule without needing an actual DOM node. It completes the api needed to run a rule in a linter on individual virtual nodes.

```js
const vNode = {
  shadowId: undefined,
  children: [],
  parent: undefined,
  _cache: {},
  _isHidden: null,
  _attrs: {
    type: 'text',
    autocomplete: 'not-on-my-watch',
  },
  props: {
    nodeType: 1,
    nodeName: 'input',
    id: null,
    type: 'text'
  },
  hasClass() {
    return false;
  },
  attr(attrName) {
    return this._attrs[attrName];
  },
  hasAttr(attrName) {
    return !!this._attrs[attrName];
  }
};

const results = axe.runVirtualRule('autocomplete-valid', vNode);
```

Linked issue: #1601

## Reviewer checks


lib/core/public/run-virtual-rule.js
```diff
@@ -0,0 +1,29 @@
...
+ * @param {VirtualNode} vNode  The virtual node to run the rule against
+ * @param {Object} options  (optional) Set of options passed into rules or checks
+ * @return {Object} axe results for the rule run
+ */
+axe.runVirtualRule = function(ruleId, vNode, options) {
+	'use strict';
+	let rule = axe._audit.rules.find(rule => rule.id === ruleId);
+
+	if (!rule) {
+		return;
```
stephenmathieson: I think we should `throw` here rather than failing silently.

Consider the following:

```js
const results = axe.runVirtualRule('snuffaluffagus', someVirtualNode)
console.log(results.violations) // will throw a very unhelpful error
```

However, if we added a `throw new Error('Invalid rule or something helpful here')`, then users would immediately know what they need to do to fix their issue.

test/core/public/run-virtual-rule.js
```diff
@@ -0,0 +1,149 @@
...
+		axe._audit = {
+			data: {}
+		};
+	});
+
+	afterEach(function() {
+		axe._audit = audit;
+	});
+
+	it('should return undefined if the rule does not exist', function() {
```
stephenmathieson: I do not agree with this. See above.

test/core/public/run-virtual-rule.js
```diff
@@ -0,0 +1,149 @@
+describe('axe.runVirtualRule', function() {
```
stephenmathieson: Please add a test verifying the return value is of the expected type/shape.

lib/core/public/run-virtual-rule.js
straker: Good point. Will make it throw instead

lib/core/base/virtual-node.js
```diff
@@ -5,14 +5,15 @@ const whitespaceRegex = /[\t\r\n\f]/g;
...
 	 * Wrap the real node and provide list of the flattened children
-	 *
-	 * @param node {Node} the node in question
-	 * @param shadowId {String} the ID of the shadow DOM to which this node belongs
+	 * @param {Node} node the node in question
+	 * @param {VirtualNode} parent The parent VirtualNode
+	 * @param {String} shadowId the ID of the shadow DOM to which this node belongs
 	 */
-	constructor(node, shadowId) {
+	constructor(node, parent, shadowId) {
```
WilcoFiers: Good point.

lib/core/base/rule.js
```diff
@@ -282,7 +282,9 @@ Rule.prototype.runSync = function(context, options = {}) {
 
 		const result = getResult(results);
 		if (result) {
-			result.node = new axe.utils.DqElement(node.actualNode, options);
+			result.node = node.actualNode
+				? new axe.utils.DqElement(node.actualNode, options)
+				: node;
```
WilcoFiers: We should probably just return null here.

lib/core/utils/select.js
```diff
@@ -103,33 +100,26 @@ function reduceIncludes(includes) {
...
-	for (var i = 0; i < reducedIncludes.length; i++) {
+	const reducedIncludes = reduceIncludes(context.include);
+	for (let i = 0; i < reducedIncludes.length; i++) {
 		candidate = reducedIncludes[i];
-		if (
-			candidate.actualNode.nodeType === candidate.actualNode.ELEMENT_NODE &&
-			axe.utils.matchesSelector(candidate.actualNode, selector) &&
-			curried(candidate)
-		) {
-			result = pushNode(result, [candidate]);
```
WilcoFiers: Can you explain why this is removed?

straker: I needed to convert it to `querySelectorAll` for virtual node only runs, but from all my testing `querySelectorAll` already returns the top node if it matches the selector, so I'm not even sure what this code did that the next line didn't do. 

The only use case I could come up with is if the selector included an element that wasn't in the axe tree but was in the DOM, but then the next line would not return any other nodes. There wasn't any test or comment about why it was there so I just removed it.

PR #1604, opened 2019-05-31
feat(autocomplete-matches): use virtualNode only lookups

straker:
Autocomplete-matches uses the JavaScript DOM property `.type` to look at the type of input. This is trivial when you have a DOM node, but becomes a bit more complicated when you don't (e.g. AST implementations). Created a method on the `VirtualNode` that just returns the DOM property lookup but is expected to be overridden for other implementations.

It also tries to determine if the node is visible in the DOM, but that doesn't work for virtual node only runs, so I made it conditional on there being an `actualNode` to look at.

Please double check the tests that I added all the same properties when converting to `queryFixture`.

## Reviewer checks

### DIFF
--- a/lib/core/base/virtual-node.js
+++ b/lib/core/base/virtual-node.js
@@ -66,7 +66,7 @@ class VirtualNode {
 	/**
 	 * Determine if the element has the given attribute.
 	 * @param {String} attrName The name of the attribute
-	 * @return {Bool} True if the element has the attribute, false otherwise.
+	 * @return {Boolean} True if the element has the attribute, false otherwise.
 	 */
 	hasAttr(attrName) {
 		if (typeof this.actualNode.hasAttribute !== 'function') {
@@ -74,6 +74,15 @@ class VirtualNode {
 		}
 
 		return this.actualNode.hasAttribute(attrName);
+	}
+
+	/**
+	 * Get the input type. This is more complicated when you don't have an
+	 * actualNode to reference.
+	 * @return {String|undefined}
+	 */
+	getType() {
+		return this.actualNode.type;
 	}
 
 	/**
--- a/lib/core/base/virtual-node.js
+++ b/lib/core/base/virtual-node.js
@@ -17,16 +17,22 @@ class VirtualNode {
 		this._isHidden = null; // will be populated by axe.utils.isHidden
 		this._cache = {};
 
-		// abstract Node and Element APIs so we can run axe in DOM-less
-		// environments. these are static properties in the assumption
-		// that axe does not change any of them while it runs.
-		this.elementNodeType = node.nodeType;
-		this.elementNodeName = node.nodeName.toLowerCase();
-		this.elementId = node.id;
-
 		if (axe._cache.get('nodeMap')) {
 			axe._cache.get('nodeMap').set(node, this);
 		}
+	}
+
+	// abstract Node properties so we can run axe in DOM-less environments.
+	// add to the prototype so memory is shared across all virtual nodes
+	get props() {
+		const { nodeType, nodeName, id, type } = this.actualNode;
+
+		return {
+			nodeType,
+			nodeName: nodeName.toLowerCase(),
+			id,
+			type
+		};
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


lib/core/base/virtual-node.js
```diff
@@ -76,6 +76,15 @@ class VirtualNode {
 		return this.actualNode.hasAttribute(attrName);
 	}
 
+	/**
+	 * Get the input type. This is more complicated when you don't have an
+	 * actualNode to reference.
+	 * @return {String|undefined}
+	 */
+	get elementType() {
+		return this.actualNode.type;
```
stephenmathieson: Just a bikeshed, but `|| null` would be preferred to undefined here.

lib/rules/autocomplete-matches.js
```diff
@@ -1,31 +1,34 @@
...
-if (['TEXTAREA', 'INPUT', 'SELECT'].includes(nodeName) === false) {
+const nodeName = virtualNode.elementNodeName;
+if (['textarea', 'input', 'select'].includes(nodeName) === false) {
 	return false;
 }
 
 // The element is an `input` element a `type` of `hidden`, `button`, `submit` or `reset`
 const excludedInputTypes = ['submit', 'reset', 'button', 'hidden'];
-if (nodeName === 'INPUT' && excludedInputTypes.includes(node.type)) {
+if (
```
AutoSponge: `hidden` inputs can still have `autocomplete`, is this exclusion based on some other factor?

lib/core/base/virtual-node.js
stephenmathieson: `null` is explicit, while `undefined` is not. IMO it's better to return a defined falsy value. It shows intention and follows practices put in place by the native DOM:

```js
document.createElement('div').getAttribute('type')
// => null
```

straker: But we're not accessing the `type` attribute, we're accessing the `type` property, which returns `undefined`.

```
document.createElement('div').type
// => undefined
```

lib/core/base/virtual-node.js
```diff
@@ -76,6 +76,15 @@ class VirtualNode {
 		return this.actualNode.hasAttribute(attrName);
 	}
 
+	/**
+	 * Get the input type. This is more complicated when you don't have an
+	 * actualNode to reference.
+	 * @return {String|undefined}
+	 */
+	get elementType() {
```
WilcoFiers: This might get confused with `nodeType`. Can we have a different name for this?

straker: Sure. I'd like it to be obvious that we're accessing JavaScript property values, and be consistent with other properties we might need (like `attributes`). I think we should also make individual property lookup functions and not a generic function like jQuery's `props` so anyone overriding the class definition knows exactly which props are needed. 

Some thoughts:

- `propType` (might get confusing with React)
- `typeProp`
- `getTypeProp` or `getPropType` 
-  just `type`

WilcoFiers: How about `inputType`?

straker: My only concern with that is that `input`s are not the only element with a `type` property. `object` elements also have one https://developer.mozilla.org/en-US/docs/Web/HTML/Element/object.

stephenmathieson: `.type` makes sense to me, since that's the native property too.

straker: How should we address the fact that `nodeType`, `nodeName` and `id` are prefixed with `element` but `type` is not?

stephenmathieson: Yikes. That's no good.

To avoid adding an inconsistent property name, I think `elementType` might be the best option. We probably want to consider changing this naming convention in v4 tho.

lib/core/base/virtual-node.js
stephenmathieson: Good point.

lib/core/base/virtual-node.js
straker: What's wrong with the naming convention? We can change anything right now before 3.3 is released.

straker: We decided to namespace all JavaScript node properties under the `.props` object named the same as their Node counterpart.

Issue #1505, opened 2019-04-18
New Rule: identical-links-same-purpose


jkodu:
# identical-links-same-purpose

This rule checks that links with identical accessible names resolve to the same resource or equivalent resources.

## Tags
- wcag2
- wcag249

## Matches

- semantic role of `link`, or roles that inherit from `link` role.
- is not `aria-hidden`.
- has `accessible name`.

## Checks

**all:**
- identical-links-same-purpose
	- `incomplete` : `Links with identical names that do not serve same purpose`.
	- `pass`: `Links with identical names serve equivalent purpose`
	
- identical-links-same-purpose-after
	- all matching nodes in the results are processed to find any identical nodes (same accessible name) and verified if they resolve to the same resource. Nodes which do not satisfy to the same resource are marked as `incomplete`.

### References:
- [ACT Rule - Links with identical accessible names serve equivalent purpose](https://act-rules.github.io/rules/b20e66)

jkodu:
@WilcoFiers - Please see proposal above, feedback welcome.

jeankaplansky:
framework in place in github and DQU 1/27/2020

jeankaplansky:
Release notes and rule content populated as of 2/03/2020

padmavemulapati:
@jkodu , I am  seeing the voilations instead incomplete for this new rule, identical-links-same-purpose. used test file specified in the `https://github.com/dequelabs/axe-core/blob/develop/test/integration/rules/identical-links-same-purpose/identical-links-same-purpose.html` 
![image](https://user-images.githubusercontent.com/47409579/74122867-ae7fc780-4bf2-11ea-8abc-4f6c0679e2a6.png)
suspecting whether the changes in the axe version updated properly or not.

padmavemulapati:
Working for axe-core's axe.run(), will verify the same with axe and attest extensions once axe-core3.5 deployed in devTools.

PR #1606, opened 2019-06-03
feat(rule): add before function callback to rule spec

jkodu:
**Note: This PR has partial work done in https://github.com/dequelabs/axe-core/pull/1605**

- Whilst authoring proposal for a new rule [identical links serve same purpose](https://github.com/dequelabs/axe-core/issues/1505), there arises a need for running a process only **once**, as a `before` step to running the rule. 
- The results of the `before` callback can be `extended` in the `context` object.
- The `context` object is passed to all `checks` of the `rule`.
- This change will enable adding a callback to rule(`json`) like so: `"before":"my-rule-before.js"`

**_More context:_**
This functionality requirement was discussed in the past, when authoring `preload (cssom)`, and also when refactoring `color-contrast` rule. This change may also enable refactor to multiple rules to gain performance improvements.

Closes issue: 
- NA

## Reviewer checks

### DIFF
--- a/build/configure.js
+++ b/build/configure.js
@@ -96,7 +96,7 @@ function buildRules(grunt, options, commons, callback) {
 		function replaceFunctions(string) {
 			return string
 				.replace(
-					/"(evaluate|after|gather|matches|source|commons)":\s*("[^"]+?")/g,
+					/"(evaluate|after|gather|matches|before|source|commons)":\s*("[^"]+?")/g,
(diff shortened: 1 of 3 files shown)


jkodu:
@WilcoFiers - looking for feedback here, hence the `draft` PR.

jkodu:
After discussion a decision was made, to author a proposal & perhaps pick this work after `3.3` release.

Issue #1326, opened 2019-01-17
button-name rule gives incorrect advise for button elements


WilcoFiers:
Axe-core should not pass the following element:

```html
<button value="foo" tabindex="-1"></button>
```

This is caused by the `value` attribute. The button-name rule has the non-empty-value check for `input[type=button]` elements, and relies on `focusable-no-name` to still fail buttons if they have a value attribute but no accessible name. Firstly this leads to a confusing remediation message, where axe is saying to put a value attribute on buttons. Secondly, because `focusable-no-name` looks at focus, you can trip it up by putting `tabindex="-1"` on the button.

I think the way to fix this is to break the rule up. Have one rule for `button` elements, and a second for `input` buttons. This is similar to the problem we have with `svg` elements where we're recommending to put an `alt` attribute on them. See #1192

straker:
As an alternate solution, could we make the `non-empty-value` check look at the `nodeName` of the element and make sure it's `input` to determine pass/fail?

jeankaplansky:
https://dequeuniversity.com/rules/axe/3.3/button-name
https://dequeuniversity.com/rules/axe/3.3/input-button-name

chandana7393:
Tested, working as expected.
![button_value_rule](https://user-images.githubusercontent.com/45285854/61214664-1f770380-a726-11e9-8c48-c0f0afd7311c.PNG)
**Tested Environment:**
Axe-coconut -3.8.1.19604v
Chrome - 75v
OS - Windows 10 64 bit.

PR #1615, opened 2019-06-05
feat(rules): split button name rule into button only and button input rules

scurker:
Split `button-name` into `button-name` and `input-button-name` rules.

* `button-name` rule should only include `button` and `[role=button]`
* `input-button-name` rule should only include inputs where type is `button`, `reset`, or `submit`.

For each rule, removed any checks that were no longer applicable to the targeted elements.

Closes: #1326 

## Reviewer checks


straker:
LGTM but will need @WilcoFiers approval before we merge

lib/rules/input-button-name.json
```diff
@@ -0,0 +1,26 @@
+{
+	"id": "input-button-name",
```
stephenmathieson: Funky spacing. Did Prettier not run here?

scurker: How is the spacing funky? It looks normal to me.

https://github.com/dequelabs/axe-core/blob/a2c531dd9bc812a639f742cc9b156d2869748b8a/lib/rules/input-button-name.json#L1-L10

stephenmathieson: Weird. Must have been a Github Refined bug. It looked like indentation was set to 1 space. :man_shrugging:

scurker: Huh, I don't think it is. I think something funky is going on with Github's formatting. There's different spacing from when I looked at it earlier. 👀

lib/rules/input-button-name.json
```diff
@@ -0,0 +1,26 @@
...
+	"any": [
+		"non-empty-if-present",
+		"non-empty-value",
+		"aria-label",
+		"aria-labelledby",
+		"role-presentation",
+		"role-none",
+		"non-empty-title"
+	],
+	"none": ["focusable-no-name"]
```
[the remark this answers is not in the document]
scurker: Looks like it's not still necessary, I removed it. Thanks!

Issue #1416, opened 2019-03-07
Redundant issues with image-redundant-alt rule


testingtycoon:
Performed axe analysis on below code snippet and observed 3 redundant issues against image-redundant-alt rule.

```
<table role=presentation>
	<tr>
		<td align=center>Pass/Fail</td>
		<td align=center>Example</td>		
	</tr>
	<tr>
		<td>Fail</td>
		<td>
			<p>
				<a href="index.html"><img src="images/image-redundant-alt.jpg" alt="Home Page" width="24" height="25">Home Page</a>
			</p>
		</td>		
	</tr>		
</table>
```

Expectation: Axe should throw only 1 violation against image-redundant-alt rule.

Actual: Axe throws 3 redundant violations against image-redundant-alt rule. One against the table cell, one against the paragraph and one against the actual element.

<pre><code>
axe-core version: 3.1.2
axe-extension version: 3.7.0
Chrome Browser: 72
</code></pre>

WilcoFiers:
Thanks Soma. We'll take a look.

chandana7393:
Tested, working as expected only one best practice was displayed.
![one_best_practice](https://user-images.githubusercontent.com/45285854/59345119-119c2000-8d2d-11e9-82c7-32fabebe86fd.PNG)
**Tested Environment:**
Axe-coconut - 3.8.0.16304v
Chrome - 75v
OS - Windows 10 64 bit.

PR #1616, opened 2019-06-05
fix(image-redundant-alt): prevent redundant issues of image tree

straker:
The `image-redundant-alt` rule applied to more than just buttons and links (as the help text stated). It also was being applied to `p`, `li`, `td`, and `th` elements. This caused a single image to be reported multiple times if it's parent tree consisted of each of those elements:

```html
<td>
  <p>
    <button>
      <img alt="Alt text">
      Alt text
    </button>
  </p>
</td>
```

This changes the rule to focus only on `img` elements and look up the tree for the `button`, `[role=button]`, `a[href]`, `p`, `li`, `td`, and `th` elements and check the text against the image alt. This way only the image node will throw the error and not each parent node.

Linked issue: #1416

## Reviewer checks


lib/checks/label/duplicate-img-label.js
```diff
@@ -1,19 +1,17 @@
-const text = axe.commons.text.visibleVirtual(virtualNode, true).toLowerCase();
-if (text === '') {
+if (['none', 'presentation'].includes(node.getAttribute('role'))) {
```
WilcoFiers: Could you use `aria.getRole(node)` instead? I'd like us to step away from moving the role attribute directly.

lib/checks/label/duplicate-img-label.js
```diff
@@ -1,19 +1,17 @@
...
-		({ actualNode }) =>
-			axe.commons.dom.isVisible(actualNode) &&
-			!['none', 'presentation'].includes(actualNode.getAttribute('role'))
-	);
+const parent = axe.commons.dom.findUpVirtual(
+	virtualNode,
+	'button, [role="button"], a[href], p, li, td, th'
+);
+const parentVNode = axe.utils.getNodeFromTree(parent);
+const text = axe.commons.text.visibleVirtual(parentVNode, true).toLowerCase();
```
WilcoFiers: For clarity, could you call this `visibleText`?

Issue #1551, opened 2019-05-07
Create a type definition for VirtualNode, possibly a class


straker:
In `axe.d.ts` a type definition must be provided that describes the new VirtualNode class.

stephenmathieson:
Yes please!

chandana7393:
Dev Task.

somaalapati:
@straker Please close this issue if it cannot be testable by QA?

PR #1627, opened 2019-06-10
feat: add AbstractVirtualNode for linting

straker:
Adds a new `AbstractVirtualNode` class that just throws on all required properties and functions. This ensures anyone who implements a `VirtualNode` has everything needed to run it through `axe.runVirtualRule`. I wanted to have it in its own file in the `public` directory but since the `VirtualNode` class extends it it needed to be available before declaring the `VirtualNode` class.

This allows a JSX styled `VirtualNode` implementation such as:

```js
class VirtualNode extends axe.AbstractVirtualNode {
  constructor(jsxAST, parent) {

    if (jsxAST.type === 'JSXText') {
      this.props = {
        nodeType: 3
      };
      return;
    }

    this.children = [];

    this._attrs = {};
    jsxAST.openingElement.attributes.forEach(attr => {
      this._attrs[attr.name.name] = attr.value.value;
    });

    this.shadowId = undefined;
    this.parent = parent;
    this.props = {
      nodeType: 1,
      nodeName: jsxAST.openingElement.name.name.toLowerCase(),
      id: this._attrs.id,
      type: this._attrs.type
    };
  }

  attr(attrName) {
    return this._attrs[attrName];
  }

  hasAttr(attrName) {
    return !!this._attrs[attrName];
  }

  hasClass(className) {
     // ...
  }
}
```

**NOTE:** by throwing in all required properties and functions, any changes that add new properties or functions could potentially break any libraries who have their own `VirtualNode` implementation which extends the `AbstractVirtualNode`. Those libraries should be encouraged to pin to specific versions of axe-core in order to not break on new releases to the `AbstractVirtualNode` class. 

Linked issue: #1551

## Reviewer checks

### DIFF
--- a/axe.d.ts
+++ b/axe.d.ts
@@ -153,12 +153,9 @@ declare namespace axe {
 		}[];
 		cleanup?(callback: Function): void;
 	}
-	class VirtualNode {
-		constructor(node?: HTMLElement, parent?: VirtualNode, shadowId?: string);
-		shadowId?: string;
+	interface VirtualNode {
 		children?: VirtualNode[];
-		actualNode?: HTMLElement;
-		parent?: VirtualNode;
+		parent?: VirtualNode | null;
 		props: VirtualNodeProps;
 		hasClass(className: string): boolean;
 		attr(attrName: string): string | null;
--- a/axe.d.ts
+++ b/axe.d.ts
@@ -22,6 +22,13 @@ declare namespace axe {
 
 	type ElementContext = Node | string | RunOnlyObject;
 
+	type VirtualNodeProps = {
+		nodeType: number;
+		nodeName: string;
+		id?: string;
+		type?: string;
+	};
+
 	interface TestEngine {
 		name: string;
 		version: string;
@@ -145,6 +152,17 @@ declare namespace axe {
 			callback(...args: any[]): void;
 		}[];
 		cleanup?(callback: Function): void;
+	}
+	class VirtualNode {
+		constructor(node?: HTMLElement, parent?: VirtualNode, shadowId?: string);
+		shadowId?: string;
+		children?: VirtualNode[];
+		actualNode?: HTMLElement;
+		parent?: VirtualNode;
+		props: VirtualNodeProps;
+		hasClass(className: string): boolean;
+		attr(attrName: string): string | null;
+		hasAttr(attrName: string): boolean;
 	}
 
 	let plugins: any;
--- a/axe.d.ts
+++ b/axe.d.ts
@@ -22,13 +22,6 @@ declare namespace axe {
 
 	type ElementContext = Node | string | RunOnlyObject;
 
-	type VirtualNodeProps = {
-		nodeType: number;
-		nodeName: string;
-		id?: string;
-		type?: string;
-	};
-
 	interface TestEngine {
 		name: string;
 		version: string;
--- a/axe.d.ts
+++ b/axe.d.ts
@@ -153,14 +153,6 @@ declare namespace axe {
 		}[];
 		cleanup?(callback: Function): void;
 	}
-	interface VirtualNode {
-		children?: VirtualNode[];
-		parent?: VirtualNode | null;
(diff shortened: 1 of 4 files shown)


axe.d.ts
```diff
@@ -146,6 +153,17 @@ declare namespace axe {
 		}[];
 		cleanup?(callback: Function): void;
 	}
+	class VirtualNode {
```
stephenmathieson: Shouldn't this be the abstract class instead?

lib/core/base/virtual-node.js
```diff
@@ -1,15 +1,39 @@
 const whitespaceRegex = /[\t\r\n\f]/g;
 
+class AbstractVirtualNode {
+	constructor() {}
```
stephenmathieson: Can remove this.

straker: I thought so too but something in our babel compiler breaks the `this` reference in the class when being extended and throws `TypeError: Cannot set property 'shadowId' of undefined`. It seems it assumes the parent class will have a `constructor` and that you need to call `super` on it to get the the correct `this` reference.

axe.d.ts
stephenmathieson: I thought we didn't expose the `VirtualNode` class and expected people to `class MyVNodeImpl extends axe.AbstractVirtualNode`. If so, we need to expose the class in the TS definition, otherwise the TS compiler won't know the abstract class exists.

lib/core/base/virtual-node.js
stephenmathieson: The subclass will need to call `super()` before setting instance properties, but the superclass doesn't need a constructor in order for that to happen.

Contrived example:

```js
class AbstractVNode {
  foo() {
    throw new Error('not implemented');
  }
}

class MyVNode extends AbstractVNode {
  constructor() {
    super();
    this.banana = true;
  }
  foo() {
    return 'hi';
  }
}

const n = new MyVNode()
n.foo()
```

axe.d.ts
```diff
@@ -146,6 +153,14 @@ declare namespace axe {
 		}[];
 		cleanup?(callback: Function): void;
 	}
+	interface VirtualNode {
```
stephenmathieson: This isn't exported on `axe-core`. Shouldn't this be `class AbstractVirtualNode`?

axe.d.ts
```diff
@@ -146,6 +153,14 @@ declare namespace axe {
...
+		children?: VirtualNode[];
```
stephenmathieson: This means `children` is either `undefined` or an array of `VirtualNode`s. IMO it should _always_ be an array, even if the array is empty.

This way you don't have to:

```
if (vNode.children && vNode.children.length) {
 // ...
}
```

axe.d.ts
```diff
@@ -146,6 +153,14 @@ declare namespace axe {
...
+		parent?: VirtualNode | null;
```
stephenmathieson: This is saying `parent` is either `undefined`, `null` or a `VirtualNode`. IMO it should only be `null` or `VirtualNode`, never `undefined`.

lib/core/base/virtual-node.js
```diff
@@ -1,15 +1,42 @@
 const whitespaceRegex = /[\t\r\n\f]/g;
 
+class AbstractVirtualNode {
+	constructor() {
+		this.children = [];
```
stephenmathieson: No big deal (no need to change), but this could just be:

```
class AbstractVirtualNode {
  children = []
  parent = null

  // ...
}
```

axe.d.ts
straker: The ts file seems to be taking things from the `axe.d.ts` file and not the `axe.js` file. I could import the `axe.VirtualNodeProps` without problems.

lib/core/base/virtual-node.js
straker: I forget we have babel to correct that for us an it's not just a native class

Issue #1628, opened 2019-06-11
color-contrast message: prevent "pt" vs "px" misreadings


dbjorge:
The color-contrast check includes the computed font size of the element being checked in its failure message in format "13.5pt". Recently, we've seen multiple false false-positive reports due to misreading this as "13.5px". We think it would be much harder to misread if it were instead formatted like "13.5pt (18px)" or "13.5pt / 18px" or similar - `color-contrast.js` already computes both anyway, so this should be a relatively minor change to the check implementation.

**Expectation**: A color-contrast check fail message like *"Element has insufficient color contrast of 4.21 (foreground color: #00809d, background color: #f5f5f5, font size: **13.5pt (18px)**, font weight: bold). Expected contrast ratio of 4.5:1"*

**Actual**: The color-contrast check's fail message currently looks like *"Element has insufficient color contrast of 4.21 (foreground color: #00809d, background color: #f5f5f5, font size: **13.5pt**, font weight: bold). Expected contrast ratio of 4.5:1"*

**Motivation**: We recently had a false report of an Accessibility Insights for Web false positive that turned out to be because a customer (and then a few of our engineers) misread the color-contrast check message's description of font-size in "pt" as "px" instead. A quick glance through similar axe-core issues suggests that a different user had a similar bit of confusion recently in #1585 - if multiple users are reporting similar confusion, we think it's probably worth making the minor update to alleviate it :)

<pre><code>
axe-core version: 3.2.2
Accessibility Insights for Web: 2.4.1
Chrome: 74.0.3729.169
</code></pre>

straker:
That's a good idea. I'll see what it will take to implement.

jeankaplansky:
The rule help is very explicit re: pts and the CSS pixel translations. No further rule help documentation required.

chandana7393:
Tested, appearing as expected.
![axe_coconut_px](https://user-images.githubusercontent.com/45285854/61215202-bbedd580-a727-11e9-88a2-e6eb847244ca.PNG)
**Tested Environment:**
Axe-coconut - 3.8.1.19604v
Chrome - 75v
OS - Windows 10 64 bit.

PR #1634, opened 2019-06-14
feat(color-contrast): add px unit to error messages

straker:
Add the size in px to the color contrast error message to avoid confusion with the pt unit.

```
"Fix any of the following:
  Element has insufficient color contrast of 3.87 (foreground color: #3d7e9a, background color: #eeeeee, font size: 12.0pt / 16px, font weight: normal). Expected contrast ratio of 4.5:1"
```

Linked issue: #1628

## Reviewer checks

- [x] Code is reviewed for security by: @dylanb

### DIFF
--- a/lib/checks/color/color-contrast.js
+++ b/lib/checks/color/color-contrast.js
@@ -42,7 +42,7 @@ const data = {
 	fgColor: fgColor ? fgColor.toHexString() : undefined,
 	bgColor: bgColor ? bgColor.toHexString() : undefined,
 	contrastRatio: cr ? truncatedResult : undefined,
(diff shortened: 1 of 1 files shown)


straker:
@jkodu I did that first, but since we were already in an parenthesis I didn't know if that would be OK

```
(foreground color: #3d7e9a, background color: #eeeeee, font size: 12.0pt (16px), font weight: normal)
```

jkodu:
@straker, it is logical to express `font` as say `300 italic 1.3em/1.7em serif`, which is why I thought parenthesis may be better. But this is a nitpick. Happy with what you decide.

Issue #1540, opened 2019-05-02
Filtering rules, result types doesn't work as expected


mohanraj-r:
Expectation: 
* Disabling a rule should prevent that rule from running. 
* Filtering by result types should present only the selected result type in output. 

Actual: 
* The rule seems to be still run (as per performanceTimer logs). `axe.run()` take the same amount of time with the rule disabled, even though the rule takes non-trivial amount of time.
* result types other than the one specified are also present in output.

Motivation: Ability to turn-off selected rules  for large pages #1464 and result types to improve performance as documented in https://github.com/dequelabs/axe-core/blob/develop/doc/API.md#use-resulttypes

<pre><code>
axe-core version: 3.2.2

Browser and Assistive Technology versions
Google chrome: Version 73.0.3683.103
</code></pre>

```
await axe.run(document, options={
  "rules": {
     "color-contrast": {enabled: false},
   },
  "resultTypes": ["violations"],
  "performanceTimer": true
})
```

Output:
```
PerformanceTimer reports the rule being run. The total time taken for execution is the same as with the rule enabled.
...
Measure rule_color-contrast took 5298.990000009537ms
15:11:51.155 axe.min.js:12 
Measure runchecks_color-contrast took 4943.65499997139ms
...
Measure axe took 14006ms
```

Other result types such as `inapplicable`, `passes` etc are still reported even though `"resultTypes": ["violations"]` is specified
```
[...]

PR #1643, opened 2019-06-19
docs: clarify resultType option

### LINKED ISSUE #1126: Misleading documentation on axe.run options key `resultTypes`
Documentation link: https://github.com/dequelabs/axe-core/blob/develop/doc/API.md#options-parameter

paragraph `6. Only process certain types of results`

It gives example and explanation of expected results:

> ```
> {
>   resultTypes: ['violations', 'incomplete', 'inapplicable']
> }
> ```
> 
> This example will process all of the "violations", "incomplete", and "inapplicable" result types. Since "passes" was not specified, it will only process the first pass for each rule, if one exists. **As a result, the results object's `passes` array will have a length of either `0` or `1`.** On a series of extremely large pages, this would improve performance considerably.

Misleading part is bolded:

> **As a result, the results object's `passes` array will have a length of either `0` or `1`.**

Unfortunately, when tested with given `resultTypes` option, aXe results object has `passes` array length equal to passed rules count, and not just limited to `0` or `1`. 

For example, `https://www.google.com/` with aXe option set to `resultTypes: ['violations', 'incomplete', 'inapplicable']` has `passes` array with length `28`.

### LINKED ISSUE #1183: Config options are not reflected in results
I am not able to get the [config options](https://github.com/dequelabs/axe-core/blob/develop/doc/API.md#options-parameter) working. I have tried with axe-webdriverjs and puppeteer (#1008)

<pre><code>
Versions:
axe-core@3.1.2
axe-webdriver@2.0.1
selenium-webdriver@4.0.0-alpha.1
chromedriver@2.42.0
Google chrome: Version 69.0.3497.100 (Official Build) (64-bit)

For Tooling issues:
- Node version: v10.11.0
- Platform:  Mac v10.13.6 
</code></pre>

Webdriver code:
```js
const AxeBuilder = require("axe-webdriverjs")
const webDriver = require('selenium-webdriver');

// Config options for axe
//  Ref: https://github.com/dequelabs/axe-core/blob/develop/doc/API.md#options-parameter
const axeConfig = {
  runOnly: {
    type: "tag",
    values: ["wcag2a", "wcag2aa"]
  },
  resultTypes: ["violations", "incomplete"]
};

const driver = new webDriver.Builder()
  .forBrowser('chrome')
  .build();

driver.get(url)
  .then(() => {
    // @ts-ignore
    AxeBuilder(driver)
      .configure(axeConfig)
      .analyze((results) => {
        console.log(JSON.stringify(results));
      });
  });
```

Output (only first-level keys are shown below for brevity): 
```json
{
  "inapplicable": [ ...
],
  "incomplete": [ ...
 ],
  "passes": [ ...
],
  "timestamp": "2018-10-09T17:16:58.320Z",
  "url": "https://www.google.com/",
  "violations": [ ...
]
}
```

I tried the following in chrome devtools (the browser instance opened by the above selenium webdriver code):
```js
const axeConfig = {
  runOnly: {
    type: "tag",
    values: ["wcag2a", "wcag2aa"]
  },
  resultTypes: ["violations", "incomplete"]
};
axe.run(document, axeConfig)

Promise {<pending>}
__proto__: Promise
[[PromiseStatus]]: "resolved"
[[PromiseValue]]: Object
inapplicable: Array(22)
0:
description: "Ensures <area> elements of image maps have alternate text"
help: "Active <area> elements must have alternate text"
helpUrl: "https://dequeuniversity.com/rules/axe/3.1/area-alt?application=webdriverjs"
id: "area-alt"
impact: null
nodes: []
tags: (5) ["cat.text-alternatives", "wcag2a", "wcag111", "section508", "section508.22.a"]
... // output snipped 
```

You can see from the above output that even though the config specifies `resultTypes: ["violations", "incomplete"]` the output in both webdriverjs and  direct `axe.run()` invocation in devtools contains other result types

straker:
The `resultType` description in the docs has caused a bit of confusion for users (see #1126, #1183, and #1540). Based on the current read of the docs, it sounds like users expect the option to filter out result types not listed. So if the option was `resultTypes: ['violations']`, the user expected only violation results to be listed. 

Instead what the option really does is limit the number of related nodes to a maximum of one for each result type not listed. Updated the docs to clarify this point. 

Closes #1126
Closes #1183

## Reviewer checks

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -476,7 +476,7 @@ This example first includes all `wcag2a` and `wcag2aa` rules. All rules that are
 
 6. Only process certain types of results
 
-The `resultTypes` options can be used to limit the number of related nodes for a rule to a maximum of one. This can be useful for improving performance on very large or complicated pages when you are only interested in certain types of results.
+The `resultTypes` option can be used to limit the number of related nodes for a rule to a maximum of one. This can be useful for improving performance on very large or complicated pages when you are only interested in certain types of results.
 
 After axe has processed all rules normally, it generates a unique selector for all related nodes in all rules. This process can be time consuming, especially for pages with lots of nodes. By limiting the related nodes to a maximum of one for result types you are not interested in, you can greatly speed up the tail end performance of axe.
 
--- a/doc/API.md
+++ b/doc/API.md
@@ -480,7 +480,7 @@ The `resultTypes` option can be used to limit the number of related nodes for a 
 
 After axe has processed all rules normally, it generates a unique selector for all related nodes in all rules. This process can be time consuming, especially for pages with lots of nodes. By limiting the related nodes to a maximum of one for result types you are not interested in, you can greatly speed up the tail end performance of axe.
 
-Types listed in this option will cause rules that fall under those types to show all related nodes. Types _not_ listed will causes rules that fall under one of the missing types to show a maximum of one related node. This allows you to still see those results and use them to inform the user of them if appropriate.
+Types listed in this option will cause rules that fall under those types to show all related nodes. Types _not_ listed will causes rules that fall under one of the missing types to show a maximum of one related node. This allows you to still see those results and inform the user of them if appropriate.
 
 ```js
 {
--- a/doc/API.md
+++ b/doc/API.md
@@ -478,9 +478,9 @@ This example first includes all `wcag2a` and `wcag2aa` rules. All rules that are
 
 The `resultTypes` options can be used to limit the number of related nodes for a rule to a maximum of one. This can be useful for improving performance on very large or complicated pages when you are only interested in certain types of results.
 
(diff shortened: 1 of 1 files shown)


doc/API.md
```diff
@@ -476,17 +476,19 @@ This example first includes all `wcag2a` and `wcag2aa` rules. All rules that are
 
 6. Only process certain types of results
 
-The `resultTypes` option can be used to limit the result types that axe will process, aggregate, and send to the reporter. This can be useful for improving performance on very large or complicated pages when you are only interested in certain types of results.
+The `resultTypes` option can be used to limit the number of related nodes for a rule to a maximum of one. This can be useful for improving performance on very large or complicated pages when you are only interested in certain types of results.
```
WilcoFiers: I think this should be "number of nodes for a rule to a maximum of one". It doesn't filter the related nodes, but the nodes that is returned. The same is true for all uses of "related nodes" below. If you just take out the word "related", this is correct. See:

https://github.com/dequelabs/axe-core/blob/develop/lib/core/reporters/helpers/process-aggregate.js#L63

Issue #1459, opened 2019-03-26
False positive: Color-contrast rule fails when partially opaque overlay div is placed on top of content


iamrafan:
Expectation: Not fail the color-contrast rule when partially opaque overlay div is placed on top of a section of view port

Actual: Fails color-contrast rule

Motivation: Once the side panel opens, the blurred out content need not be accessible to the user.

Demo link: https://developer.microsoft.com/en-us/fabric#/components/panel

![image](https://user-images.githubusercontent.com/22068434/54966314-1c66c600-4f31-11e9-92ed-3ee8ebbe87b0.png)

![image](https://user-images.githubusercontent.com/22068434/54966342-32748680-4f31-11e9-8195-6c55aad9bc26.png)

<pre><code>axe-core version: 3.1.2
axe-Coconut version: 3.7.0</code></pre>

WilcoFiers:
Interesting. I reran this page with 3.2.2, 3.1.2 and 3.0.3, none of them report it. Is it possible this page has changed since?

iamrafan:
@WilcoFiers No, I can still repro the issue using axe-core 3.1.2.

roman-II:
@iamrafan Can you shed a little light on how this page is handling trapping the focus within the pop out panel? This will help with debugging. Thanks!

iamrafan:
@roman-II I am not on the same team, I am not sure how the focus is being handled here

cnamal:
Pinged folks from OUFR to provide assistance.

dzearing:
@kkjeer can you help take a look?

kkjeer:
Panel uses a [FocusTrapZone](https://github.com/OfficeDev/office-ui-fabric-react/blob/master/packages/office-ui-fabric-react/src/components/Panel/Panel.base.tsx#L167) for focus management. @JasonGore is more familiar with the FocusTrapZone component.

JasonGore:
This issue is reporting a contrast issue with Overlay. I'm not sure what to do here from a FocusTrapZone perspective. FocusTrapZone does not apply any styling. If there is a FocusTrapZone focus issue to look into, I'd recommend creating a new issue.

jeankaplansky:
No docs required.

chandana7393:
Tested, still reproducible.
![overlap_needs_review](https://user-images.githubusercontent.com/45285854/61215030-379b5280-a727-11e9-9f0b-0fc6ac6668f8.PNG)
**Tested Environment:**
[...]

PR #1644, opened 2019-06-20
fix(get-background-color): No longer calculate color from non-opaque overlapping element

roman-II:
…overlapping elm

Prevents calculating color contrast ratio that have an overlapping element

Closes issue: #1459 

## Reviewer checks

### DIFF
--- a/lib/commons/color/get-background-color.js
+++ b/lib/commons/color/get-background-color.js
@@ -92,7 +92,7 @@ color.getBackgroundStack = function getBackgroundStack(elm) {
 
 	// Return all elements BELOW the current element, null if the element is undefined
 	let elmIndex = elmStack.indexOf(elm);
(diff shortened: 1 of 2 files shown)


AutoSponge:
I'm worried this will ignore overlays that cause a problem. Is it possible we can only apply this rule to contents contained by the `intert` attribute or `aria-hidden`?

straker:
Can you give an example of an overlay that causes a problem? One issue is that not all overlays are well written and `inert` or `aria-hidden` the rest of the page. Some will just focus trap the overlay without truly hiding everything else.

AutoSponge:
I'd argue that an incomplete (incorrect) implementation should get the color-contrast rule as-is. If they add `inert`/`aria-hidden` to a container, those elements can be exempt from this rule. The overlay I'm referring to is the one in the linked issue. Without looking at the implementation, I would need to make sure that there are no tab stops in the overlaid section (`inert`) before I said that contrast doesn't matter.

Issue #1504, opened 2019-04-18
New Rule: filename is valid accessible name


jeeyyy:
# filename-is-valid-accessible-name

This rule checks that image elements that use their source filename as their accessible name do so without loss of information to the user.

## Tags
- wcag2
- wcag111
- experimental

## Matches

- any HTML input element with a type of image
- any HTML element with the semantic role of img, 
- is not `aria-hidden`.
- has `accessible name`.

## Checks

**all:**
- filename-is-valid-accessible-name
	- `incomplete` 
		- `Element does not have an accessible name that serves an equivalent purpose to the non-text content`.
	- `pass`: 
		- `Element has an accessible name that serves an equivalent purpose to the non-text content`.

### References:
- [ACT Rule - Filename is valid accessible name](https://act-rules.github.io/rules/9eb3f6)

dylanb:
I suggest that we close this because axe-core does not and cannot infer accurate meaning at this point in time i.e. this rule proposal is out of scope for axe-core

jkodu:
@dylanb, we intend to only return `incomplete` on this rule and not a `failure`. Perhaps can mark this rule as `experimental` too.

dylanb:
ok, I am worried about noise

jkodu:
# Test cases (WIP)

```html
<!-- Pass 1, accessible name, uses filename -->
<img src="https://www.w3.org/WAI/demos/bad/img/w3c" alt="w3c" />

<!-- Pass 2, accessible name, uses filename (along with extension) -->
<img src="https://www.w3.org/WAI/demos/bad/img/w3c.png" alt="Download w3c.png"
/>

<!-- Incomplete 1, accessible name matches file name, but presence of extension in accessible name results in accesible name not describing the image accurately -->
<img src="https://www.w3.org/WAI/demos/bad/img/w3c.png" alt="w3c.png" />

<!-- Incomplete 2, same as above, but different element type -->
<input type="image" src="https://www.w3.org/WAI/demos/bad/before/img/top_weather.gif" alt="top_weather.gif" />

<!-- Inapplicable 1 -->
<img role="presentation" />

<!-- Inapplicable 2 -->
<img style="display:none;" />
```

jkodu:
@jeankaplansky the rule has been merged #1649 and is ready for docs.

WilcoFiers:
Closing this. ACT has deprecated this rule.

PR #1645, opened 2019-06-20
[WIP] feat(rule): filename-is-valid-accessible-name

jkodu:
This PR adds a new rule
- `filename-is-valid-accessible-name`

Further Reading:
- https://act-rules.github.io/rules/9eb3f6

Closes issue:
- https://github.com/dequelabs/axe-core/issues/1504

## Reviewer checks


lib/checks/shared/filename-is-valid-accessible-name.js
```diff
@@ -0,0 +1,29 @@
+const { text } = axe.commons;
+const possibleImageExtensions = [`.jpg`, `.jpeg`, `.png`, `.gif`, `.tif`];
```
straker: Should include WebP images as well

lib/checks/shared/filename-is-valid-accessible-name.js
```diff
@@ -0,0 +1,29 @@
...
+
+const accText = text.accessibleText(node).toLowerCase();
+
+/**
+ * Check if accessible name, uses filename (along with file extension)
+ * eg: Download image of Tajmahal.jpg
+ */
+const fileExtnInAccText = accText.match(/\.[0-9a-z]+$/i);
```
straker: Image file types don't use numbers

lib/rules/filename-is-valid-accessible-name-matches.js
```diff
@@ -0,0 +1,38 @@
...
+	return false;
+}
+
+const srcValue = node.getAttribute('src');
+if (!srcValue) {
+	return false;
+}
+
+const nodeName = node.nodeName.toUpperCase();
+const nodeType = node.getAttribute('type');
```
straker: Should use the property lookup `node.type` so it uses the browsers interpreted type (then you don't have to check for it's existence either)

lib/checks/shared/filename-is-valid-accessible-name.json
```diff
@@ -0,0 +1,11 @@
+{
+	"id": "filename-is-valid-accessible-name",
+	"evaluate": "filename-is-valid-accessible-name.js",
+	"metadata": {
+		"impact": "minor",
+		"messages": {
+			"pass": "Element has an accessible name that serves an equivalent purpose to the non-text content",
+			"incomplete": "Element does not have an accessible name that serves an equivalent purpose to the non-text content"
```
straker: This doesn't help me understand why the check "failed" if the image uses the file extension in the acc name. Should tell me how to fix it so I don't have to click out to the linked help doc.

test/checks/shared/filename-is-valid-accessible-name.js
```diff
@@ -0,0 +1,45 @@
...
+
+	it('returns undefined when element accessible name contains image extension type', function() {
+		var vNode = queryFixture(
+			'<img id="target" src="https://www.w3.org/WAI/demos/bad/img/w3c" alt="w3c.png" />'
+		);
+		var actual = check.evaluate.call(checkContext, vNode.actualNode);
+		assert.isUndefined(actual);
+	});
+
+	it('returns undefined when element accessible name contains image extension type', function() {
```
straker: This is the same text as the prior test. Should update it to reflect that you're testing a different element (`<input type="image">`)

test/checks/shared/filename-is-valid-accessible-name.js
```diff
@@ -0,0 +1,45 @@
...
+
+	it('returns true when img element accessible name uses the filename which describes the image', function() {
+		var vNode = queryFixture(
+			'<img id="target" src="https://www.w3.org/WAI/demos/bad/img/w3c" alt="w3c" />'
+		);
+		var actual = check.evaluate.call(checkContext, vNode.actualNode);
+		assert.isTrue(actual);
+	});
+
+	it('returns true when img element accessible name uses the filename which describes the image', function() {
```
straker: Ditto here about same text as prior test but different element

lib/rules/filename-is-valid-accessible-name-matches.js
```diff
@@ -0,0 +1,38 @@
+/**
+ * Matches the below:
+ * - `HTMLInputElement` with `type=image`
+ * - Element with `role=img`
```
straker: This is confusing as I thought you wanted to match any element with `role=img` and not just make sure the `img` element had the correct role.

test/integration/rules/filename-is-valid-accessible-name/filename-is-valid-accessible-name.html
```diff
@@ -0,0 +1,29 @@
+<!-- Pass -->
```
straker: No test for this example from the dos:

```html
<a href="https://www.w3.org/WAI/demos/bad/img/w3c.png" download
	>Download <img src="w3c.png" alt="w3c.png"
/></a>
```

test/rule-matches/filename-is-valid-accessible-name-matches.js
```diff
@@ -0,0 +1,53 @@
+describe('filename-is-valid-accessible-name-matches', function() {
```
straker: Missing test for `<input type="image">` matching

lib/checks/shared/filename-is-valid-accessible-name.js
dylanb: This should be an option to the check with a default value in the JSON file

lib/checks/shared/filename-is-valid-accessible-name.js
dylanb: this is ok - especially if there is an option for configuring these - but why not simply make the regex the last instance of anything searching backwards from the end of the string, looking for the `.`?

`const fileExtnInAccText = accText.match(/\.(?:.(?!\.))+$/i);`

lib/checks/shared/filename-is-valid-accessible-name.js
```diff
@@ -0,0 +1,29 @@
...
+ */
+const srcValue = node.getAttribute('src').toLowerCase();
+const fileName = text.sanitize(srcValue.split('/').pop());
+const fileNameWithoutExtn = fileName.split('.').shift();
+
+if (accText.includes(fileNameWithoutExtn)) {
+	return true;
+}
+
+return undefined;
```
dylanb: This will return a `needs review` for every alt attribute that is not a file name without the extension. As far as I can tell, this would return a needs review for

`Favorite`

This is absolutely terrible!

lib/rules/filename-is-valid-accessible-name.json
```diff
@@ -0,0 +1,12 @@
+{
```
dylanb: we need a selector for this rule or it will have terrible performance for no reason

test/integration/rules/filename-is-valid-accessible-name/filename-is-valid-accessible-name.html
```diff
@@ -0,0 +1,29 @@
...
+	src="https://www.w3.org/WAI/demos/bad/img/w3c.png"
+	alt="w3c.png"
+/>
+<input
+	id="incomplete2"
+	type="image"
+	src="https://www.w3.org/WAI/demos/bad/before/img/top_weather.gif"
+	alt="top_weather.gif"
+/>
+<img
```
dylanb: This is terrible!!! We will destroy Axe with rules like this.

dylanb:
@WilcoFiers I object strongly to this rule

lib/checks/shared/filename-is-valid-accessible-name.json
jkodu: @jeankaplansky - I need help with better messages here.
Here is more info about the rule here - https://act-rules.github.io/rules/9eb3f6

test/integration/rules/filename-is-valid-accessible-name/filename-is-valid-accessible-name.html
jkodu: Yeah, this is up for debate with in the ACT Rule community. I think we ought to only look for the element and not its context for computing accessible name. That brings in an entire new spectrum. Leave this with me to discuss with concerned peeps.

jkodu:
@dylanb @WilcoFiers closing this rule. @WilcoFiers help make a decision.

test/integration/rules/filename-is-valid-accessible-name/filename-is-valid-accessible-name.html
jkodu: @WilcoFiers will appreciate your input on this.

jkodu: Spoke to @WilcoFiers. The rule will not compute accessible name outside of the context of the given node/ thyself. So for this example, `incomplete` will be returned.

stephenmathieson:
@dylanb using language like "this is terrible" and "not implemented well" on a public forum should be avoided. It sets a very discouraging tone and is not at all welcoming to new contributors. Please try not to be so hostile when reviewing PRs.

jkodu:
Will re-open when picking up the work again after further clarification of the rule.

PR #1649, opened 2019-06-21
feat(rule): identical-links-same-purpose

jkodu:
This PR adds a new rule
- `identical-links-same-purpose`

Further Reading:
- https://act-rules.github.io/rules/b20e66

Closes issue:
- https://github.com/dequelabs/axe-core/issues/1505

## Reviewer checks


straker:
PS. you linked to the Filename is valid accessible name docs

test/integration/full/identical-links-same-purpose/passes.html
```diff
@@ -0,0 +1,53 @@
...
+
+	<body>
+		<!-- Hiding mocha div so that the links generated by mocha stats do not intervene -->
+		<!-- <div id="mocha"></div> -->
+
+		<a href="/about/contact.html">Contact</a>
+		<a href="/about/contact.html">Contact</a>
+
+		<a href="/about/contact.html">Contact</a>
+		<a href="/admissions/contact.html">Contact</a>
```
straker: This is a different url, should it really pass?

jkodu:
@straker I have caught up on some requested changes. For the rest myself and @WilcoFiers are discussing.

lib/checks/navigation/identical-links-same-purpose.json
```diff
@@ -0,0 +1,12 @@
+{
+	"id": "identical-links-same-purpose",
+	"evaluate": "identical-links-same-purpose.js",
+	"after": "identical-links-same-purpose-after.js",
+	"metadata": {
+		"impact": "minor",
+		"messages": {
+			"pass": "Identical links (if any) resolve to same purpose",
```
WilcoFiers: "There are no other links with the same name, that go to a different URL"

lib/checks/navigation/identical-links-same-purpose.js
```diff
@@ -0,0 +1,48 @@
...
+ * Get resource that a element points to (eg: href or location redirects)
+ * @param {HTMLElement} el Element
+ * @returns {String|null}
+ */
+function getLinkResource(el) {
+	if (el.hasAttribute('href')) {
+		return el.getAttribute('href');
+	}
+
+	const resourceAttr = Array.from(axe.utils.getNodeAttributes(node)).find(
```
WilcoFiers: Trying to match a stringified script is not going to be reliable. I don't think we should even try to do this.

lib/checks/navigation/identical-links-same-purpose.js
```diff
@@ -0,0 +1,48 @@
...
+
+return true;
+
+/**
+ * Get resource that a element points to (eg: href or location redirects)
+ * @param {HTMLElement} el Element
+ * @returns {String|null}
+ */
+function getLinkResource(el) {
+	if (el.hasAttribute('href')) {
```
WilcoFiers: use `elm.href`, rather than `getAttribute('href')`. It'll give you the complete URL. That way hrefs that are different, but resolve to the same page are considered the same.

test/checks/navigation/identical-links-same-purpose.js
```diff
@@ -0,0 +1,118 @@
...
+			checkContext,
+			vNode.actualNode,
+			options,
+			vNode
+		);
+		assert.isUndefined(actual);
+		assert.isNull(checkContext._data);
+	});
+
+	it('returns undefined when element does not have a resource (onclick does not change location)', function() {
```
WilcoFiers: If the accessible name of this is unique, we should pass it.

test/checks/navigation/identical-links-same-purpose.js
```diff
@@ -0,0 +1,118 @@
...
+	var checkContext = axe.testUtils.MockCheckContext();
+	var options = {};
+
+	afterEach(function() {
+		fixture.innerHTML = '';
+		checkContext.reset();
+	});
+
+	it('returns undefined when element does not have a resource (empty href)', function() {
+		var vNode = queryFixture('<a id="target" href="">Go to google.com</a>');
```
WilcoFiers: This resolves to the current page. I think that should be considered.

test/checks/navigation/identical-links-same-purpose.js
```diff
@@ -0,0 +1,118 @@
...
+			checkContext,
+			vNode.actualNode,
+			options,
+			vNode
+		);
+		assert.isTrue(actual);
+		assert.hasAllKeys(checkContext._data, ['accessibleText', 'linkResource']);
+	});
+
+	describe('after', function() {
```
WilcoFiers: For the sake of consistency, test the after method in a different file. All other things with `after` do so.

lib/checks/navigation/identical-links-same-purpose.js
```diff
@@ -0,0 +1,30 @@
...
+if (!accNameSanitized) {
+	return undefined;
+}
+
+/**
+ * Set `data` for use in `after` fn
+ */
+const afterData = {
+	name: accNameSanitized,
+	resource: node.href
```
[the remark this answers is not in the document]
jkodu: I have discussed this with @WilcoFiers. When we have two or more such links with identical accessible name, we flag it as `incomplete`.

lib/checks/navigation/identical-links-same-purpose-after.js
```diff
@@ -0,0 +1,151 @@
...
+
+/**
+ * Get list of nodes from results which match a given accessible name
+ * @method getNodesWithIdenticalAccessibleName
+ * @param {String} expectedName accessible name to be matched
+ * @param {Number} excludeIndex exclude `index` of result, that should not be taken into consideration
+ * @returns {Array<Object>}
+ */
+function getNodesWithIdenticalAccessibleName(expectedName, excludeIndex) {
+	return results.filter((result, index) => {
```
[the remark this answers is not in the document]
jkodu: This would mean we do not raise `incomplete` on all the nodes that have identical accessible names. So we let one `pass` and raise `incomplete` on the other nodes. This needs to be discussed.

lib/checks/navigation/identical-links-same-purpose-after.js
```diff
@@ -0,0 +1,151 @@
...
+ */
+function servesSamePurpose(expectedResource, identicalNodes) {
+	const { parts: expected } = parseUri(expectedResource);
+	return identicalNodes.every(({ data }) => {
+		const { resource = undefined } = data;
+		if (!resource) {
+			return false;
+		}
+		const { isFile, parts: actual } = parseUri(resource);
+		return actual[isFile ? 'some' : 'every'](part => expected.includes(part));
```
WilcoFiers: This is very unsafe. `parseUri` should not return an array. There's no telling what you're now comparing. What if the hash is the same as the filename or the protocol? Because you turned an object into an array, your data lost all its integrity. This `? 'some' : 'every'` thing also makes me very nervous. So if the protocol is the same it'll pass? Why?

Please rethink this approach.

lib/checks/navigation/identical-links-same-purpose-after.js
```diff
@@ -0,0 +1,151 @@
...
+ * @method parseUri
+ * @param {String} uri uri resource
+ * @returns {Object}
+ */
+function parseUri(uri = '') {
+	const parser = document.createElement('a');
+	parser.href = uri;
+
+	const curatedPathname = stripLeadingAndTrailingSlash(parser.pathname);
+	const defaults = [
```
WilcoFiers: These are your base values, they're not "defaults". Please use a more appropriate variable name.

lib/checks/navigation/identical-links-same-purpose-after.js
```diff
@@ -0,0 +1,151 @@
...
+		parser.protocol,
+		parser.hostname,
+		parser.port,
+		parser.search
+	];
+	const pathname = parseUriPathnameAndHash(curatedPathname, parser.hash);
+	const file = parseUriPathnameAndFile(curatedPathname);
+
+	let uriParts = [...defaults, ...pathname];
+	if (file) {
```
WilcoFiers: This looks buggy. The only reason you can do this is because `file` can be undefined, which contradicts your documentation that says `parseUriPathnameAndFile` always returns an array. If your function did actually return an empty array, this wouldn't work because `!![] === true`.

lib/checks/navigation/identical-links-same-purpose-after.js
```diff
@@ -0,0 +1,151 @@
...
+		uriParts = [...defaults, ...file];
+	}
+	return {
+		isFile: !!file,
+		parts: uriParts.filter(item => !!item)
```
WilcoFiers: `parts` is not a meaningful name. Call it `uriParts` or something please. Use consistent names for your variables.

lib/checks/navigation/identical-links-same-purpose-after.js
```diff
@@ -0,0 +1,151 @@
...
+		parser.hostname,
+		parser.port,
+		parser.search
+	];
+	const pathname = parseUriPathnameAndHash(curatedPathname, parser.hash);
+	const file = parseUriPathnameAndFile(curatedPathname);
+
+	let uriParts = [...defaults, ...pathname];
+	if (file) {
+		uriParts = [...defaults, ...file];
```
WilcoFiers: It took me 5 reads to realise you were overriding uriParts rather than concatenating `file` to it. You have forking logic here. Either file, or something else. That's what `if / else` is for. Not using `if / else` here sets incorrect expectations, and lead to someone having to read this code 5 times before they understand what's happening. You're also using memory that shouldn't be, because you create an array and than immediately override it.

test/checks/navigation/identical-links-same-purpose-after.js
```diff
@@ -0,0 +1,42 @@
+describe('identical-links-same-purpose-after tests', function() {
```
WilcoFiers: Please test the URI logic.

lib/checks/navigation/identical-links-same-purpose-after.js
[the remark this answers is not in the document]
straker: `getNodesWithIdenticalAccessibleName` returns all nodes that are identical, we can mark all of them as incomplete in the same go instead of just the one you started on.

test/integration/rules/identical-links-same-purpose/identical-links-same-purpose.html
```diff
@@ -0,0 +1,187 @@
+<!-- pass -->
```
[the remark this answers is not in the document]
jkodu: > Note: As discussed this rule is scoped to be with in the same document tree or shadow tree. ACT Rule has a PR open for updating this in the applicability. I have added shadow Dom tests.

lib/checks/navigation/identical-links-same-purpose-after.js
```diff
@@ -0,0 +1,151 @@
...
+}
+
+/**
+ * Parse a given URI and return to parts
+ * @method parseUri
+ * @param {String} uri uri resource
+ * @returns {Object}
+ */
+function parseUri(uri = '') {
+	const parser = document.createElement('a');
```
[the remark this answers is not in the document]
stephenmathieson: I don't know how frequently we run this function, but creating a detached node is pretty quick:

```
console.time('create_anchor'); document.createElement('a'); console.timeEnd('create_anchor')
// => create_anchor: 0.02197265625ms
```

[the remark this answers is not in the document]
jkodu: I have optimised this, by bypassing this step if given node is `a` or `area`.

lib/checks/navigation/identical-links-same-purpose.js
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	const [pathname, filename] = getPathnameAndFilename(parser.pathname);
+	return {
+		protocol: parser.protocol,
+		hostname: parser.hostname,
+		port: parser.port,
+		pathname: pathname.replace(/^\/|\/$/g, ''), // revmove lead/trial(ing) slashes
+		search: parser.search,
+		hash: getHash(parser.hash),
```
WilcoFiers: Can we rename these to hashRoute and getHashRoute? This isn't just the hash after all.

lib/checks/navigation/identical-links-same-purpose.js
```diff
@@ -0,0 +1,151 @@
...
+if (!name) {
+	return undefined;
+}
+
+/**
+ * Set `data` and `relatedNodes` for use in `after` fn
+ */
+const afterData = {
+	name,
+	parsedResource: getParsedResource(node, 'href')
```
WilcoFiers: You're now parsing the HREF in `<div href="foo" role="link">`. Probably shouldn't do that.

lib/checks/navigation/identical-links-same-purpose.js
```diff
@@ -0,0 +1,151 @@
...
+		parser = document.createElement('a');
+		parser.href = value;
+	}
+
+	const [pathname, filename] = getPathnameAndFilename(parser.pathname);
+	return {
+		protocol: parser.protocol,
+		hostname: parser.hostname,
+		port: parser.port,
+		pathname: pathname.replace(/^\/|\/$/g, ''), // revmove lead/trial(ing) slashes
```
WilcoFiers: Leading slashes matter. Why are you removing those?

lib/checks/navigation/identical-links-same-purpose.js
```diff
@@ -0,0 +1,151 @@
...
+	if (filename.indexOf('.') === -1) {
+		return [pathname];
+	}
+
+	const hasIndexInFilename = /index./.test(filename);
+	if (hasIndexInFilename) {
+		return [pathname.replace(filename, '')];
+	}
+
+	return [filename];
```
WilcoFiers: Please return a `{pathname, filename}` object instead of an array. That way you don't accidentally return the wrong value, like you do here, where you're returning the filename in the position that the pathname is expected. I can't tell if that's on purpose, but code that looks buggy. Even if it isn't, this is likely to trip someone up in the future.

lib/rules/identical-links-same-purpose-matches.js
```diff
@@ -0,0 +1,79 @@
...
+	if (!mapEl) {
+		return false;
+	}
+
+	const mapElName = mapEl.getAttribute('name');
+	if (!mapElName) {
+		return false;
+	}
+
+	return isMapReferredByAnyImgNode(rootEl, mapElName);
```
WilcoFiers: As mentioned before, I think this should be dealt with by accessibleText, through its use of `isVisible`. I don't think we should handle this here.

lib/rules/identical-links-same-purpose-matches.js
```diff
@@ -0,0 +1,79 @@
...
+if (['A', 'AREA'].includes(nodeName) && !node.hasAttribute('href')) {
+	return false;
+}
+
+if (nodeName === 'AREA') {
+	const rootEl = dom.getRootNode(node);
+	/**
+	 * Ensure map is used by `img` witin the document tree
+	 */
+	const mapEl = getMapForAreaElement(node, rootEl);
```
WilcoFiers: I have my doubts that you can put `area` elements in a different tree from the `map` element. Please test that. In which case `node.closest('map')` should do fine. If not, use `dom.findUpVirtual(virtualNode, 'map')`. This new `getMapForAreaElement`  seems unnecessary.

lib/rules/identical-links-same-purpose-matches.js
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Check if a given name is referred by <img> node within the `usemap` attribute
+ * @param {HTMLElement} rootEl root node (document or shadowRoot)
+ * @param {String} mapElName element name
+ * @reurns {Boolean}
+ */
+function isMapReferredByAnyImgNode(rootEl, mapElName) {
+	const mapName = escapeSelector(mapElName);
+	const refs = rootEl.querySelectorAll(`[usemap="#${mapName}"]`);
```
WilcoFiers: You should probably just limit this to `img[usemap="..."]`. More accurate and speeds up the selector.

test/checks/navigation/identical-links-same-purpose-after.js
```diff
@@ -0,0 +1,99 @@
...
+	'use strict';
+
+	var fixture = document.getElementById('fixture');
+	var check = checks['identical-links-same-purpose'];
+
+	afterEach(function() {
+		fixture.innerHTML = '';
+	});
+
+	function assertResult(
```
WilcoFiers: Abstracting in tests is not a good practice. Please just use the three asserts where you need to. The point of tests is that they should be easy to read. Tests don't need to be DRY.

lib/checks/navigation/identical-links-same-purpose-after.js
```diff
@@ -0,0 +1,104 @@
...
+}
+
+/**
+ * Verify if parsed resource match against all given identical nodes
+ * @method isIdenticalResource
+ * @param {Object} expectedResource parsed resource to be matched against
+ * @param {Array<Object>} identicalNodes nodes with identical accessible name whose parsed resource is checked for identity
+ * @returns {Boolean}
+ */
+function isIdenticalResource(expectedResource, identicalNodes) {
```
WilcoFiers: Personally, not a fan of filter methods. The reuse value of this is far less then if you had a `isIdenticalResource` method, and you did:

```js
identicalNodes.every(({ data }) => isIdenticalResource(data.parsedResource, parsedResource))
```

test/checks/navigation/identical-links-same-purpose.js
```diff
@@ -0,0 +1,266 @@
...
+		var actual = check.evaluate.call(
+			checkContext,
+			vNode.actualNode,
+			options,
+			vNode
+		);
+		assert.isTrue(actual);
+		assert.hasAllKeys(checkContext._data, ['name', 'parsedResource']);
+		assert.equal(checkContext._data.name, 'Pass 1'.toLowerCase());
+		assert.equal(checkContext._data.parsedResource.hash, '#/foo');
```
WilcoFiers: Simpler way to test this would be just to test that whatever is on `.parsedResource` is deepEqual to what you get from `dom.parsedResource`.

test/checks/navigation/identical-links-same-purpose.js
```diff
@@ -0,0 +1,266 @@
...
+		assert.equal(checkContext._data.parsedResource.pathname, '/home');
+	});
+
+	it('returns undefined for `AREA` without closest `MAP` element', function() {
```
WilcoFiers: This stuff is handled by the selector. I don't think the check needs to duplicate that.

test/checks/navigation/identical-links-same-purpose.js
```diff
@@ -0,0 +1,266 @@
...
+			checkContext,
+			vNode.actualNode,
+			options,
+			vNode
+		);
+		assert.isUndefined(actual);
+		assert.isNull(checkContext._data);
+	});
+
+	it('returns undefined when native link has only emoji as accessible name', function() {
```
WilcoFiers: You can probably just combine those unicode filter things into a single test. You have other tests that ensure that's sufficiently covered. That's not what this check's test should be about.

test/checks/navigation/identical-links-same-purpose.js
```diff
@@ -0,0 +1,266 @@
...
+		assert.hasAllKeys(checkContext._data, ['name', 'parsedResource']);
+		assert.equal(
+			checkContext._data.name,
+			'The is orange the is white'.toLowerCase()
+		);
+		assert.equal(checkContext._data.parsedResource.filename, 'foo.html');
+	});
+
+	(shadowSupported ? it : xit)(
+		'returns undefined for native link (in shadowDOM) with `href` but no accessible name',
```
WilcoFiers: Seems unnecessary. The check does nothing with shadow DOM.

test/checks/navigation/identical-links-same-purpose-after.js
```diff
@@ -0,0 +1,98 @@
+describe('identical-links-same-purpose-after tests', function() {
```
WilcoFiers: Add a test for when `parsedResource` is missing., and one for when they don't have the same keys.

lib/commons/dom/get-parsed-resource.js
```diff
@@ -0,0 +1,100 @@
...
+	};
+};
+
+/**
+ * Resolve if a given pathname has filename & resolve the same as parts
+ * @method getPathnameAndFilename
+ * @param {String} pathname pathname part of a given uri
+ * @returns {Array<String>}
+ */
+function getPathnameAndFilename(pathname) {
```
WilcoFiers: getPathnameAndFilename -> getPathnameOrFilename, you're never returning both.

lib/commons/dom/get-parsed-resource.js
```diff
@@ -0,0 +1,100 @@
...
+	if (!pathname) {
+		return;
+	}
+	const filename = pathname.split('/').pop();
+	if (!filename || filename.indexOf('.') === -1) {
+		return { pathname };
+	}
+
+	const hasIndexInFilename = /index./.test(filename);
+	if (hasIndexInFilename) {
```
WilcoFiers: Put the test in the if.

lib/checks/navigation/identical-links-same-purpose-after.js
```diff
@@ -0,0 +1,104 @@
...
+	return identicalNodes.every(({ data }) => {
+		const { parsedResource } = data;
+		if (!parsedResource) {
+			return false;
+		}
+
+		// remove key's whose values are undefined
+		const keysWithValues = Object.keys(parsedResource).filter(
```
WilcoFiers: This causes asymmetrical equality. Because of that, changing the order of nodes impacts whether or not they pass. Here's an example:

```html
<a href="foo.html#/foo">foo</a>
<a href="foo.html">foo</a>

<a href="bar.html">bar</a>
<a href="bar.html#/bar">bar</a>
```
In this case, foo passes, whereas bar is incomplete. I'm not sure why you are doing this filtering. I'm guessing there are cases where you want to allow certain props to be specified and others not to be. I don't think that should always be the case. For instance, if one thing has a port and the other does not, those two aren't equal (except for port 80, you may want to set that as a default).

lib/commons/dom/get-parsed-resource.js
```diff
@@ -0,0 +1,100 @@
...
+		return {
+			pathname: pathname.replace(filename, '')
+		};
+	}
+
+	return {
+		filename
```
WilcoFiers: I think this needs to return the pathname too. We shouldn't treat `deque.com/axe/api.html` the same as `deque.com/attest/api.html`

test/commons/dom/get-parsed-resource.js
```diff
@@ -0,0 +1,105 @@
...
+
+	it('returns parsed resource when `A` has empty `HREF`', function() {
+		var vNode = queryFixture(
+			'<a id="target" href="">Follow us on Instagram</a>'
+		);
+		var actual = axe.commons.dom.getParsedResource(vNode.actualNode, 'href');
+		assert.isDefined(actual);
+		assert.isUndefined(actual.hash);
+		assert.isUndefined(actual.filename);
+		assert.equal(actual.hostname, 'localhost');
```
WilcoFiers: this is error prone. Can you get this from the window.location instead?

test/commons/dom/get-parsed-resource.js
```diff
@@ -0,0 +1,105 @@
...
+	});
+
+	it('returns parsed resource for `A` with `HREF`', function() {
+		var vNode = queryFixture(
+			'<a id="target" href="https://facebook.com">follow us on Facebook</a>'
+		);
+		var actual = axe.commons.dom.getParsedResource(vNode.actualNode, 'href');
+		assert.isDefined(actual);
+		assert.equal(actual.protocol, 'https:');
+		assert.equal(actual.hostname, 'facebook.com');
```
WilcoFiers: I think you should do a `deepEqual` on actual, so that you don't get unexpected props from your test.

lib/checks/navigation/identical-links-same-purpose-after.js
```diff
@@ -0,0 +1,104 @@
...
+			key => !!parsedResource[key]
+		);
+
+		// ensure every key/ value in resources match
+		const result = keysWithValues.every(key => {
+			const actual = parsedResource[key];
+			const expected = expectedResource[key];
+			return actual === expected;
```
WilcoFiers: You can't compare equality on query strings. `?foo=foo&bar=bar` is functionally the identical to `?bar=bar&foo=foo`.

lib/commons/dom/get-parsed-resource.js
```diff
@@ -0,0 +1,100 @@
+/* global dom */
+
+/**
+ * Parse resource object for a given node from a specified attribute
+ * @method getParsedResource
+ * @param {HTMLElement} node given node
+ * @param {String} attribute attribute of the node from which resource should be parsed
+ * @returns {Object}
+ */
+dom.getParsedResource = function getParsedResource(node, attribute) {
```
WilcoFiers: This name doesn't really describe that it does. How about `urlPropsFromAttribute`?

lib/checks/navigation/identical-links-same-purpose-after.js
jkodu: Extract this functionality to a new fn under namespace `dom` & call the fn -> `hasSameUrl`

lib/checks/navigation/identical-links-same-purpose.js
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Set `data` and `relatedNodes` for use in `after` fn
+ */
+const afterData = {
+	name,
+	parsedResource: dom.getParsedResource(node, 'href')
+};
+this.data(afterData);
+this.relatedNodes([node]);
```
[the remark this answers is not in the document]
jkodu: Created issue - https://github.com/dequelabs/axe-core/issues/1858

lib/commons/dom/url-props-from-attribute.js
```diff
@@ -0,0 +1,154 @@
...
+
+
+/**
+ * Resolve given port excluding default port(s)
+ * @param {String} port port
+ * @returns {String}
+ */
+function getPort(port) {
+	/**
+	 * Note: IE11 resolves port to `443` for any URL with `https:`
```
stephenmathieson: This comment is confusing. The default HTTPS port is 443. All browsers do this.

lib/commons/dom/is-visible.js
```diff
@@ -39,6 +39,39 @@ function isClipped(style) {
...
+	const mapEl = dom.findUp(el, 'map');
+	if (!mapEl) {
+		return false;
+	}
+
+	const mapElName = mapEl.getAttribute('name');
+	if (!mapElName) {
+		return false;
+	}
+	const refs = document.querySelectorAll(
```
WilcoFiers: This is the wrong way around. This checks that the `img` is in the light DOM, instead of checking that the `map` is in the light DOM. The `img` can be anywhere in the tree. You should add some shadow DOM tests for this to prevent regression.

lib/commons/dom/url-props-from-attribute.js
WilcoFiers: Please fix this.

lib/commons/dom/url-props-from-attribute.js
```diff
@@ -0,0 +1,154 @@
...
+	return !excludePorts.includes(port)
+		? port
+		: ``
+}
+
+/**
+ * Resolve if a given pathname has filename & resolve the same as parts
+ * @method getPathnameOrFilename
+ * @param {String} pathname pathname part of a given uri
+ * @returns {Array<String>}
```
WilcoFiers: This is incorrect

lib/commons/dom/url-props-from-attribute.js
```diff
@@ -0,0 +1,154 @@
...
+ */
+function getPathnameOrFilename(pathname) {
+	if (!pathname) {
+		return;
```
WilcoFiers: This is going to throw. Because you're doing destructive assignment on the result, this must return an object. This code path isn't executed in your tests.

lib/commons/dom/url-props-from-attribute.js
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	const filename = pathname.split('/').pop();
+	if (!filename || filename.indexOf('.') === -1) {
+		return {
+			pathname
```
WilcoFiers: Why not return filename `""`?

lib/commons/dom/url-props-from-attribute.js
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	const { pathname = '', filename = '' } = getPathnameOrFilename(
+		parser.pathname
+	);
+	return {
+		protocol: parser.protocol,
+		hostname: parser.hostname,
+		port: getPort(parser.port),
+		pathname: pathname.replace(/\/$/, ''), // remove trialing slashes
```
WilcoFiers: I'd recommend always adding the trailing slash, instead of removing it. It's not a path if it doesn't have the slash.

lib/commons/dom/url-props-from-attribute.js
```diff
@@ -0,0 +1,154 @@
...
+ * @returns {Object}
+ */
+function getSearchPairs(searchStr) {
+	const query = {};
+
+	if (!searchStr || !searchStr.length) {
+		return query
+	}
+
+	const curatedSearchStr = searchStr[0] === `?` ? searchStr.substring(1) : searchStr
```
WilcoFiers: When would this ever not start with `?`? Please add a comment.

lib/commons/dom/url-props-from-attribute.js
```diff
@@ -0,0 +1,154 @@
...
+	for (let index = 0; index < pairs.length; index++) {
+		const pair = pairs[index];
+		const [key, value = ''] = pair.split(`=`)
+		query[decodeURIComponent(key)] = decodeURIComponent(value);
+	}
+
+	/**
+	 * Sort query `keys` alphabetically
+	 */
+	return Object.keys(query)
```
WilcoFiers: Objects in JS do not have ordered keys. This is not reliable.

lib/commons/dom/url-props-from-attribute.js
```diff
@@ -0,0 +1,154 @@
...
+}
+
+/**
+ * Interpret a given hash
+ * if `hash`
+ * -> is `hashbang` -or- `hash` is followed by `slash`
+ * -> it resolves to a different resource
+ * @method getHashRoute
+ * @param {String} hash hash component of a parsed uri
+ * @returns {String|undefined}
```
WilcoFiers: These methods are inconsistent about what happens when they didn't find anything. Some return undefined, other return an empty string. I don't really care which one is used, but it should be consistent.

test/commons/dom/url-props-from-attribute.js
```diff
@@ -0,0 +1,168 @@
...
+		var vNode = queryFixture('<a id="target">Follow us on Instagram</a>');
+		var actual = axe.commons.dom.urlPropsFromAttribute(vNode.actualNode, 'href');
+		assert.isUndefined(actual);
+	});
+
+	it('returns URL properties when `A` has empty `HREF`', function () {
+		var vNode = queryFixture(
+			'<a id="target" href="">Follow us on Instagram</a>'
+		);
+		var expected = {
```
WilcoFiers: Please provide a URL for this. This will break if we run this on a different port, or use an IP address instead of localhost.

test/commons/dom/url-props-from-attribute.js
```diff
@@ -0,0 +1,168 @@
...
+		}
+		var actual = axe.commons.dom.urlPropsFromAttribute(vNode.actualNode, 'href');
+		assert.deepEqual(actual, expected)
+	});
+
+	it('returns URL properties for `A` with `HREF` which has filename', function () {
+		var vNode = queryFixture(
+			'<a id="target" href="directory/widgets/calendar.html">Book tour</a>'
+		);
+		var expected = {
```
WilcoFiers: As before. the port and hostname can change on this.

test/commons/dom/url-props-from-attribute.js
```diff
@@ -0,0 +1,168 @@
...
+			hostname: "mysite.com",
+			pathname: "/search",
+			port: "",
+			protocol: "http:",
+			search: {
+				a: "123",
+				z: "XYZ",
+				name: "Axe",
+				values: "[1,2,3]",
+				version: "1.2.3"
```
WilcoFiers: You didn't test what happens if there's no value.

test/checks/navigation/identical-links-same-purpose.js
```diff
@@ -0,0 +1,137 @@
...
+			vNode
+		);
+		assert.isTrue(actual);
+		assert.hasAllKeys(checkContext._data, ['name', 'urlProps']);
+		assert.equal(checkContext._data.name, 'Pass 1'.toLowerCase());
+		assert.equal(checkContext._data.urlProps.hash, '#/foo');
+		assert.equal(checkContext._data.urlProps.pathname, '/home');
+	});
+
+	it('returns undefined for `AREA with closest `MAP` with no name attribute', function() {
```
WilcoFiers: These AREA things are done by the matcher. These tests don't really do anything.

lib/commons/dom/url-props-from-attribute.js
```diff
@@ -0,0 +1,154 @@
...
+	if (!['A', 'AREA'].includes(nodeName)) {
+		parser = document.createElement('a');
+		parser.href = value;
+	}
+
+	const { pathname = '', filename = '' } = getPathnameOrFilename(
+		parser.pathname
+	);
+	return {
+		protocol: parser.protocol,
```
WilcoFiers: I'm not so sure we should include this. `http://deque.com` is the same URL as `https://deque.com` I'd say.

jkodu: Upon further reading, the protocol section of the URI is usually called a scheme & I found this long list of schemes - https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml that can form the part of the URI.

At this juncture I am not sure if keeping a finite list of protocols and tweaking it is a good idea (as in convert `https` to `http`). Please provide your thoughts.

WilcoFiers: Even so, "http" and "https" should be treated as the same, same for ftp and ftps. Please add that.

lib/commons/dom/is-visible.js
```diff
@@ -39,6 +39,50 @@ function isClipped(style) {
...
+
+	const refs = axe.utils.querySelectorAll(
+		axe._tree,
+		`img[usemap="#${axe.utils.escapeSelector(mapElName)}"]`
+	);
+	if (!refs || !refs.length) {
+		return false;
+	}
+
+	return refs.some(({ actualNode }) =>
```
WilcoFiers: There's no test for this.

lib/checks/navigation/identical-links-same-purpose-after.js
```diff
@@ -0,0 +1,102 @@
...
+ * Skip, as no results to curate
+ */
+if (results.length < 2) {
+	return results;
+}
+
+/**
+ * Filter results for which `result` is undefined & thus `data`, `relatedNodes` are undefined
+ */
+const filteredResults = results.filter(
```
WilcoFiers:
```suggestion
const incompleteResults = results.filter(
```
Use descriptive variable names instead of comments.

lib/checks/navigation/identical-links-same-purpose-after.js
```diff
@@ -0,0 +1,102 @@
...
+	({ result, data }) => result !== undefined && !!data
```
WilcoFiers:
```suggestion
	({ result, data }) => result !== undefined
```

That second bit does nothing

Issue #741, opened 2018-02-22
Is this really a WCAG defect: All <th> elements and elements with role=columnheader/rowheader must have data cells they describe


melaniephilipp:
Could you please clarify how the table used in the example on this page: https://dequeuniversity.com/rules/axe/3.0/th-has-data-cells is in violation of WCAG? The table is constructed properly - every `<th>` has a column of corresponding `<td>`s. How does having empty data cells in the last column make this table a failure of WCAG? And, at a Serious level as my testing indicates that the table is accessibility supported across many platforms. There are certainly many legitimate use cases for having a table that has a column of empty data cells.
Thank you!

WilcoFiers:
Hey Melanie. Thanks for the report. I think you are entirely correct. I would suggest that we make the following changes:
1. Allow headers associated with empty cells - while still flagging empty with no associated cells
2. Flag headers in tables without data cells as "needs review", since the table could simply not be populated

What do you think?

anutin544:
Ok

WilcoFiers:
@anutin544 Please stop responding with meaningless "ok" messages. I'll treat this as spam going forward.

anutin544:
Srie

straker:
After looking into this, it looks like the code correctly marks the "Bad Example" as needs review. It seems like the example should be updated to show a missing column rather than an empty column.

jeankaplansky:
Immediate change doesn't require specific documentation.

chandana7393:
@straker Can you please specify the expected behaviour.

straker:
@chandana7393 The rule should return needs review only if a table heading row or table heading column has 0 cells (so is missing the entire row/column after the header). Otherwise it should always pass.

```html
<table>
  <tr>
    <th>Header</th>
  </tr>
  <tr></tr>
</table>
```

chandana7393:
Tested, working as expected.
![Attest_table_empty](https://user-images.githubusercontent.com/45285854/61268404-f5baec80-a7b8-11e9-93ad-452b7917c6bd.PNG)
![Uploa
![axe_coconut_table_empty](https://user-images.githubusercontent.com/45285854/61268410-f94e7380-a7b8-11e9-8502-de272fd53475.PNG)
ding Axe_table_empty.PNG…]()
**Tested Environment:**
Attest - 2.5.1.19317v
Axe - 3.8.1.19317v
Axe-coconut - 3.8.1.19604v
Chrome - 75v
Firefox - 68v
[...]

PR #1659, opened 2019-06-26
fix(th-has-data-cells): empty cells will now pass

straker:
Allow empty cells to pass the check. Only if there are no header cells will the check return `undefined`.

Linked issue: #741

## Reviewer checks


lib/checks/tables/th-has-data-cells.js
```diff
@@ -40,25 +40,21 @@ headers.forEach(header => {
...
-		hasCell = tableUtils
-			.traverse('right', pos, tableGrid)
-			.some(
-				cell =>
-					axe.commons.dom.hasContent(cell) && !tableUtils.isRowHeader(cell)
-			);
+		hasCell =
+			tableUtils
+				.traverse('right', pos, tableGrid)
+				.filter(cell => !tableUtils.isRowHeader(cell)).length !== 0;
```
scurker: Wouldn't `.find()` be preferable here? `find` will halt after the first match where `filter` will always run through the full array.

test/integration/full/incomplete/th-has-data-cells.html
```diff
@@ -30,17 +30,6 @@
 				</tr>
 			</table>
 
-			<table id="table2">
```
WilcoFiers: Please move this table to the passes, so we don't lose this as a test.

PR #1684, opened 2019-07-08
test: make the output of `test:examples` readable

stephenmathieson:
This patch rewrites the program which runs the examples' tests with something more human (and coincidentally machine) readable. Rather than piping `npm install` and `npm test` output to `stdio` directly (and in parallel), we suppress the output (unless an error is raised) and run the tests serially. We give visual feedback letting the user running the tests know what's happening, but don't flood them with too much information (this is done via [`Listr`](http://npmjs.org/listr)).

This has the unfortunate side effect of increasing the time the `test:examples` script takes, but will make debugging the tests significantly easier.

output from `npm run test:examples`

A failed test will look something like:

failed test output

## Reviewer checks

### DIFF
--- a/doc/examples/test-examples.js
+++ b/doc/examples/test-examples.js
@@ -7,28 +7,33 @@ const exampleDirs = readdirSync(__dirname)
 	.map(dir => join(__dirname, dir))
 	.filter(dir => statSync(dir).isDirectory());
 
-const tasks = new Listr([
-	{
-		title: 'Install dependencies',
-		task: () =>
-			new Listr(
-				exampleDirs.map(dir => ({
-					title: basename(dir),
-					task: () => execa('npm install', { cwd: dir, shell: true })
-				}))
-			)
-	},
-	{
-		title: 'Run tests',
-		task: () =>
-			new Listr(
-				exampleDirs.map(dir => ({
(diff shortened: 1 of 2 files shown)


straker:
the output for this looks way worse when it keeps all prior screens https://circleci.com/gh/dequelabs/axe-core/7597?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

stephenmathieson:
> the output for this looks way worse when it keeps all prior screens

Wow, yes it does. Looks like this isn't going to work after all.

Maybe we should just buffer output and not output to stdio until each `npm test` is "done"?

straker:
We could just not asyc the tests. It would make them run slower, but each example is only one or two tests + `npm install`, so it should still be decently fast.

scurker:
> We could just not asyc the tests. It would make them run slower, but each example is only one or two tests + `npm install`, so it should still be decently fast.

I originally scaffolded them this way, but I wouldn't be opposed to having them run synchronously. These tests should run fairly fast and shouldn't hold up builds significantly since they run after the full test suite.

stephenmathieson:
Will pick this back up later.

PR #1686, opened 2019-07-08
test(examples): do not fail tests if deque.com has violations

stephenmathieson:
This patch updates the `phantomjs` example test, making it not fail if an axe-core violation is found on deque.com.

While it's extremely embarrassing that deque.com has violations, it's not something we control, so we shouldn't block merges based on it.

## Reviewer checks


stephenmathieson:
@straker can you please fill in the checklist?

straker:
Sorry, done

PR #1700, opened 2019-07-11
test: require new release rule help docs to be active before creating release

straker:
This should block a release pr (a pr into `master` branch only) if the newest axe version rule help docs are not active.

## Reviewer checks

### DIFF
--- a/test/test-rule-help-version.js
+++ b/test/test-rule-help-version.js
@@ -1,0 +1,16 @@
+var https = require('https');
+var path = require('path');
+var assert = require('assert');
+var packageJSON = require(path.join(__dirname, '../package.json'));
+
+var version = packageJSON.version.substr(
+	0,
+	packageJSON.version.lastIndexOf('.')
+);
+
+it('latest axe version rule help docs should be active', function(done) {
+	https.get('https://dequeuniversity.com/rules/axe/' + version, function(res) {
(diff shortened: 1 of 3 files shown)


test/test-rule-help-version.js
```diff
@@ -0,0 +1,16 @@
+var https = require('https');
+var path = require('path');
+var assert = require('assert');
+var packageJSON = require(path.join(__dirname, '../package.json'));
+
+var version = packageJSON.version.substr(
```
stephenmathieson: This looks a little more complicated and error-prone than necessary. What about:

```js
var [major,minor] = packageJSON.version.split('.')
var url = `https://dequeuniversity.com/rules/axe/${major}.${minor}`
// ...
```

stephenmathieson: Also, since this test is run **in a Node.js context**, we can use "modern" JS syntax. We don't have PhantomJS getting in the way here.

straker: Except that we our jslint blocks all es6 syntax in the `test` dir.

stephenmathieson: Bummer OK

stephenmathieson:
> LGTM. Nitpick, the test/* folder allows for es6 syntax.

Per https://github.com/dequelabs/axe-core/pull/1700#discussion_r302638483, we'd need to update ESLint to allow for it.

Issue #1688, opened 2019-07-09
Do not require textbox role in combobox if input[type=search] exists


ldanet:
This is effectively a duplicate of https://github.com/dequelabs/axe-core/issues/549. That issue is marked as closed but the issue persists.

Expectation: 
The code below should pass.
It is a copy paste of Example 1 of [ARIA 1.1 Combobox with Listbox Popup Examples](https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html) except I replaced input[type=text] with input[type=search]
```html
<label for="ex1-input"
       id="ex1-label"
       class="combobox-label">
  Choice 1 Fruit or Vegetable
</label>
<div class="combobox-wrapper">
  <div role="combobox"
       aria-expanded="false"
       aria-owns="ex1-listbox"
       aria-haspopup="listbox"
       id="ex1-combobox">
    <input type="search"
           aria-autocomplete="list"
           aria-controls="ex1-listbox"
           id="ex1-input">
  </div>
  <ul aria-labelledby="ex1-label"
      role="listbox"
      id="ex1-listbox"
      class="listbox hidden">
  </ul>
</div>
```

Actual: aXe audit fails with `Required ARIA child role not present: textbox` for the `#ex1-combobox` element.

Motivation: As per the [WAI-ARIA 1.1 recommendations for combobox](https://www.w3.org/TR/wai-aria-1.1/#combobox):
> Authors MUST ensure an element with role combobox contains or owns a text input element with role textbox or searchbox 

input[type=search] should satisfy this requirement.

Here is a minimal page containing the above snippet: https://srvpf.codesandbox.io/

axe-core version: 3.2.2-canary.2390925
axe-coconut extension version 3.8.1

straker:
Thanks for reporting this, we'll take a look at the code to see what's going on.

jeankaplansky:
Added a link to the rule help back to the ARIA documentation specific to role="combobox" which explains the intent of the spec.

chandana7393:
Tested, working as expected.
![role_textbox](https://user-images.githubusercontent.com/45285854/67200473-7d9c3c80-f421-11e9-9bfb-9b9d8553effa.PNG)

**Tested Environment:**
Axe-coconut - 4.0.0.29404v
Chrome - 77.0.3865.120v
OS - Windows 10 64 bit.

Issue #1677, opened 2019-07-08
`listitem` rule fails with roles inheriting from the 'list' or 'listitem' roles


rdeltour:
Expectation: Axe would not report a failure to the rule `listitem` when native list elements (`ol`,`ul`,`li`) declare a more specific role _inheriting_ from their native role (e.g. [`role="directory"` on `ul`](https://www.w3.org/TR/wai-aria-1.1/#directory), [`role="doc-biblioentry` on `li`](https://www.w3.org/TR/dpub-aria-1.0/#doc-biblioentry), etc.).

Actual: the rule `listitem` fails.

Motivation: even if some of those roles may have limited accessibility support (like the DPUB roles), using them on elements with native list semantics is an appropriate fallback and should pass the rule.

<pre><code>
axe-core version: 3.2.2
</code></pre>

Example 1 of markup that will fail the rule `listitem`
```html
<ul>
  <li role="doc-biblioentry">an entry</li>
  …
</ul>
```

Example 2 of markup that will fail the rule `listitem`
```html
<ol role="directory">
   <li><a href="preface_001.xhtml">Original Transcriber’s Notes:</a></li>
   <li><a href="introduction_001.xhtml">ETYMOLOGY.</a></li>
   <li><a href="epigraph_001.xhtml">EXTRACTS (Supplied by a Sub-Sub-Librarian).</a></li>
   <li><a href="chapter_001.xhtml">Chapter 1. Loomings.</a></li>
   …
</ol>
```

rdeltour:
If you agree this is a false-positive, I can submit at PR 😊

straker:
Looks good to me, a PR would be great. Just to be clear, did you mean that 

```html
<ul>
  <li role="doc-biblioentry">an entry</li>
  …
</ul>
```

fails the `list` rule?

WilcoFiers:
Totally agreed, although I'm not sure there's an obvious fix for it. Axe-core doesn't have a good representation of the ARIA role hierarchy. If you find it's going to be a complicated fix, please put together a proposal. Don't want to waste a whole bunch of time and then find out there's some (possibly unrelated) edge case that makes it not work.

rdeltour:
@straker 
> did you mean that (…) fails the `list` rule?

Yes, that one would fail the `list` rule (`only-list-items` check), and the other example I gave in the OP would fail the `listitem` rule (`listitem` check).

@WilcoFiers 
[...]

PR #1708, opened 2019-07-17
fix(aria-required-children): allow combobox to own a searchbox

straker:
Allow `aria-required-children` to pass when the `role=combobx` owns a native `input[type=search]`

Linked issue: #1688

## Reviewer checks

- [x] Code is reviewed for security by: @dylanb


lib/checks/aria/required-children.js
```diff
@@ -62,13 +63,13 @@ function missingRequiredChildren(node, childRoles, all, role) {
...
+		// remove 'textbox' from missing roles if combobox is a native text-type input or owns a 'searchbox'
 		var textboxIndex = missing.indexOf('textbox');
-		var textTypeInputs = ['text', 'search', 'email', 'url', 'tel'];
 		if (
 			textboxIndex >= 0 &&
-			node.nodeName.toUpperCase() === 'INPUT' &&
-			textTypeInputs.includes(node.type)
+			(getRole(node) === 'textbox' ||
+				owns(node, virtualNode, 'searchbox') ||
+				ariaOwns(ownedElements, 'searchbox'))
```
WilcoFiers: I don't think you correctly identified the problem. This exception was correct the way it was. If the combobox is an input element with text input, an additional `textbox` is not required. The real problem is that we're not looking at inheritance in this rule.

In ARIA 1.1, `searchbox` inherits from `textbox`. So any `searchbox` that is owned, should count as a `textbox`. That's not too different from the issue #1677. A problem there is that `directory` inherits from `list`, and so any `listitem` element can be owned by a `directory`.

You'll probably have to ditch the `owns` and `ariaOwns` functions in this method to fix this problem. There's an `aria.getOwnedVirtual` method that I got started on a while ago, which I intended to at some point be a replacement of this code. We should maybe take the time to finish it. That way you don't have to run multiple QSA's for everything that inherits from another role.

dylanb:
@straker Have we checked #1688 for accessibility supported?

lib/commons/aria/index.js
```diff
@@ -1974,6 +1975,7 @@ lookupTable.role = {
 			'input:not([type])',
 			'textarea'
 		],
+		subclassRoles: ['searchbox'],
```
WilcoFiers: IMO we should derive subclasses from the `superclassRole` that you've introduced, rather than duplicate this information.

lib/commons/aria/subclass-implicit-nodes.js
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Get a list of CSS selectors of subclass nodes that have an implicit role
+ * @method subclassImplicitNodes
+ * @memberof axe.commons.aria
+ * @instance
+ * @param {String} role The role to check
+ * @return {Mixed} Either an Array of CSS selectors or `null` if there are none
+ */
+aria.subclassImplicitNodes = function(role) {
```
WilcoFiers: I'd propose we create a `getRoleInheritance` function, which returns the role + recursively return any subclass roles.

lib/checks/aria/required-children.js
```diff
@@ -10,11 +11,15 @@ function owns(node, virtualTree, role, ariaOwned) {
 		return false;
 	}
 	var implicit = implicitNodes(role),
```
WilcoFiers: There's a bug here. This should be `implicitNodes(role) + ':not([role])'`

straker:
Does that mean reverting to the first commit https://github.com/dequelabs/axe-core/pull/1708/commits/cb3fbab18fa13caded8240e9275c363dd6875414?

lib/checks/aria/required-children.js
```diff
@@ -9,7 +9,9 @@ function owns(node, virtualTree, role, ariaOwned) {
 	if (node === null) {
 		return false;
 	}
-	var implicit = implicitNodes(role),
+	var implicit = implicitNodes(role).map(
```
jkodu: CI is failing saying `expected [TypeError: null is not an object (evaluating 'implicitNodes(role).map')] to equal null`.

PR #1721, opened 2019-07-24
test: run example tests in sync

straker:
This will better let us see which example dir is failing the test as the tests now run in sync. To try to speed up the test though the `npm install` command is run async for each dir.

## Reviewer checks

### DIFF
--- a/doc/examples/test-examples.js
+++ b/doc/examples/test-examples.js
@@ -12,15 +12,15 @@ function install(dir) {
 	return execa('npm install', { cwd: dir, ...config });
 }
 
-// but run tests synchronously so we can see which one threw an error
+// run tests synchronously so we can see which one threw an error
 function test(dir) {
 	return execa('npm test', { cwd: dir, ...config });
 }
 
 Promise.all(exampleDirs.map(install))
 	.then(async () => {
-		for (let i = 0; i < exampleDirs.length; i++) {
-			await test(exampleDirs[i]);
+		for (const dir of exampleDirs) {
+			await test(dir);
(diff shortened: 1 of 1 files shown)


doc/examples/test-examples.js
```diff
@@ -5,14 +5,24 @@ const exampleDirs = readdirSync(__dirname)
...
-	await execa('npm install', config);
-	return execa('npm test', config);
+const config = { stdio: 'inherit', shell: true };
+
+// run npm install in parallel
+function install(dir) {
+	return execa('npm install', { cwd: dir, ...config });
+}
+
+// but run tests synchronously so we can see which one threw an error
```
stephenmathieson: Please remove "but". The sentence reads better without it.

doc/examples/test-examples.js
```diff
@@ -5,14 +5,24 @@ const exampleDirs = readdirSync(__dirname)
...
+function test(dir) {
+	return execa('npm test', { cwd: dir, ...config });
 }
 
-Promise.all(exampleDirs.map(runner))
-	.then(() => {
+Promise.all(exampleDirs.map(install))
+	.then(async () => {
+		for (let i = 0; i < exampleDirs.length; i++) {
```
stephenmathieson: `for (const dir of exampleDirs) {...}` seems nicer to me.

straker:
That would be nice, though it would be nicer if somehow we autogenerated the circlci config for it so we didn't need to update it each time we added a new example (being easy to forget and miss).

Issue #1712, opened 2019-07-19
Why require video tags to have a track element with type=description? 


DavidMacDonald:
Expectation: Should not flag video tags that are missing ```<track kind="description"...>```
Actual: Flags them as missing audio description

Motivation: There is no evidence that I've found that a kind=description attribute can be used by any existing tool to provide audio description

WilcoFiers:
Hey David, I think you've got a point. I only know of a single media player that supports kind=description, and it's not one of the common ones. We are working on a plan on how we might be able to deprecate rules. We'll take this into consideration too.

jeankaplansky:
Deprecate rule in 3.4 rule help.

chandana7393:
Tested, the issue is still appearing.
![video_type_description](https://user-images.githubusercontent.com/45285854/67281583-5ca11d80-f4ed-11e9-939b-f8628dca996a.PNG)

**Tested Environment:**
Axe-coconut - 4.0.0.29504v
Chrome -  77.0.3865.120v
OS - Windows 10 64 bit.

PR #1737, opened 2019-07-30
feat(video-description): deprecate video-description rule

straker:
Per Wilco's suggestion in the linked issue, only 1 minor media player supports `<track kind="description"...>`. 

Closes issue: #1712

## Reviewer checks


jkodu:
This looks good, but according to this doc - https://github.com/dequelabs/axe-core/blob/develop/doc/backwards-compatibility-doc.md#table-summary-of-what-deque-guarantees-with-public-axe-core-api, need to keep some track of all the deprecated stuff so we know we do the right version bump when time comes to it.

Issue #1733, opened 2019-07-30
Deprecate "group-labelledby" and "fieldset" checks


WilcoFiers:
With #1640, these two checks are now no longer in use. We need to put some solution in place to deprecate these checks as well.

jeankaplansky:
Deprecated radiogroup, checkboxgroup, and video-description rules.

PR #1738, opened 2019-07-30
test: don't run integration tests for deprecated rules

straker:
When [trying to deprecate](https://github.com/dequelabs/axe-core/issues/1733) the `fieldset` and `group-labelledby` checks by disabling them using `enabled: false`, the deprecated rules using these checks would fail their integration tests since the check was no longer running for the rule. This change allows us to disable checks completely only if either all rules that use the check have been deprecated or if the check is removed from all the rules metadata.

## Reviewer checks

### DIFF
--- a/test/integration/rules/runner.js
+++ b/test/integration/rules/runner.js
@@ -29,7 +29,13 @@
 
 	var fixture = document.getElementById('fixture');
 	Object.keys(tests).forEach(function(ruleId) {
-		describe(ruleId, function() {
+		// don't run deprecated rules
+		var rule = axe._audit.rules.find(function(rule) {
+			return rule.id === ruleId;
+		});
+		var deprecated = rule.tags.indexOf('deprecated') !== -1;
+
+		(deprecated ? describe.skip : describe)(ruleId, function() {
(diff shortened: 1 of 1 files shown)


test/integration/rules/runner.js
```diff
@@ -29,7 +29,13 @@
 
 	var fixture = document.getElementById('fixture');
 	Object.keys(tests).forEach(function(ruleId) {
-		describe(ruleId, function() {
+		// don't run deprecated rules
+		var rule = axe._audit.rules.find(function(rule) {
+			return rule.id === ruleId;
+		});
+		var deprecated = rule.tags.indexOf('deprecated') !== -1;
```
stephenmathieson: If `rules.find()` somehow doesn't match, this will throw, since `rule` will be `undefined`. Do we want to guard against this?

straker: I thought about it, but the entire test script assumes there is a rule with that id, otherwise the `runOnly` option for `axe.run` would also through. It's still probably not a bad idea though.

stephenmathieson: Fair enough!

PR #1740, opened 2019-07-31
feat(fieldset, group-labelledby): deprecate fieldset and group-labelledby checks

straker:
Deprecating checks used only in the newly deprecated `radiogroup` and `checkboxgroup` rules. The `deprecated` property doesn't really do anything in the code, but I figured it would be nice to have so we can search for the term and get hits on rules and checks.

Closes issue: #1733

## Reviewer checks

### DIFF
--- a/lib/checks/forms/group-labelledby.json
+++ b/lib/checks/forms/group-labelledby.json
@@ -2,6 +2,8 @@
 	"id": "group-labelledby",
 	"evaluate": "group-labelledby.js",
 	"after": "group-labelledby-after.js",
+	"enabled": false,
+	"deprecated": true,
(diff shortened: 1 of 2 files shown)


lib/checks/forms/group-labelledby.json
```diff
@@ -2,6 +2,8 @@
 	"id": "group-labelledby",
 	"evaluate": "group-labelledby.js",
 	"after": "group-labelledby-after.js",
+	"enabled": false,
```
WilcoFiers: Don't think these should be disabled. They aren't running currently, but they might get in the way of custom rules.

Issue #1731, opened 2019-07-29
False positive: 'color-contrast' rule fails for disabled control's label


iamrafan:
Per WCAG 1.4.3 the text that is part of an inactive user interface component do not have contrast requirement.

> Incidental: Text or images of text that are part of an inactive user interface component, that are pure decoration, that are not visible to anyone, or that are part of a picture that contains significant other visual content, have no contrast requirement.

Expectation: 'color-contrast' would not fail for labels related to disabled controls

Actual: 'color-contrast' rule fails for a label(using ```for``` attribute) of a disabled control

Motivation: Text related to the disabled control is exempt from the contrast requirements

Demo link: https://developer.microsoft.com/en-us/fabric#/controls/web/combobox

![image](https://user-images.githubusercontent.com/22068434/62076429-7be52300-b1fc-11e9-935b-c459ac6e24d0.png)

<pre><code>axe-core version: 3.3.1-canary.ed28298
axe-Coconut version: 3.8.1</code></pre>

straker:
Thanks for reporting the issue. We'll take a look into what's causing the issue.

chandana7393:
Tested, working fine on Axe-coconut. Able to reproduce the issue in Axe and Attest.
![disabled_color_contrast](https://user-images.githubusercontent.com/45285854/64001705-0b486380-cb26-11e9-95bf-a3237904a8b0.PNG)
![disabled_color_contrast_attest](https://user-images.githubusercontent.com/45285854/64001713-0c799080-cb26-11e9-8180-2f9113b86c13.PNG)
**Tested Environment:**
Attest - 2.6.1.24122v
Axe - 3.9.0.24122v
Axe-coconut - 3.9.0.24122v
Chrome - 76.0.3809.132v
Firefox - 68.0.2v
OS - Windows 10 64 bit.

PR #1741, opened 2019-07-31
fix(color-contrast-matches): don't check aria-disabled explicit label element

straker:
We check if the node or its parent is `aria-disabled`, but not the element an explicit label pointed to. Went ahead and updated the code to use es6.

Closes issue: #1731

## Reviewer checks

### DIFF
--- a/test/rule-matches/color-contrast-matches.js
+++ b/test/rule-matches/color-contrast-matches.js
@@ -238,7 +238,7 @@ describe('color-contrast-matches', function() {
 		assert.isFalse(rule.matches(target, axe.utils.getNodeFromTree(target)));
 	});
 
-	it("should not match a aria-disabled input's label - explicit label", function() {
+	it("should not match an aria-disabled input's label - explicit label", function() {
 		fixture.innerHTML =
 			'<label for="t1">Test</label><input type="text" id="t1" aria-disabled="true">';
 		var target = fixture.querySelector('label');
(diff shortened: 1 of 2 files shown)


test/rule-matches/color-contrast-matches.js
```diff
@@ -238,6 +238,22 @@ describe('color-contrast-matches', function() {
...
+		fixture.innerHTML =
+			'<label for="t1">Test</label><input type="text" id="t1" aria-disabled="true">';
+		var target = fixture.querySelector('label');
+		axe.testUtils.flatTreeSetup(fixture);
+		assert.isFalse(rule.matches(target, axe.utils.getNodeFromTree(target)));
+	});
+
+	it("should not match a parent aria-disabled input's label - explicit label", function() {
+		fixture.innerHTML =
+			'<label for="t1">Test</label><div aria-disabled="true"><input type="text" id="t1"></div>';
```
WilcoFiers: Can you explain why this should fail? I don't think aria-disabled cascades.

straker: I just it from the first few lines of the code that checks that the node or its parent has `aria-disabled` https://github.com/dequelabs/axe-core/blob/develop/lib/rules/color-contrast-matches.js#L6-L11.

straker: Opened #1766 to look into this

PR #1744, opened 2019-08-02
test: add test showing axe runs under strict csp

straker:
Integration test showing that axe can parse and run under a strict csp. If you want to see that this works, you can open the `strict-csp.html` file and change the source of axe to [3.3.0](https://unpkg.com/axe-core@3.3.0/axe.js), then `npm run start` and navigate to the test to see it fail.

image

## Reviewer checks

### DIFF
--- a/test/integration/full/strict-csp/strict-csp.html
+++ b/test/integration/full/strict-csp/strict-csp.html
@@ -6,10 +6,13 @@
 			// phantomjs requires running in non-strict csp
 			if (!window.PHANTOMJS) {
 				var thisScript = document.getElementById('check-phantomjs');
-				var csp = document.createElement('div');
-				csp.innerHTML =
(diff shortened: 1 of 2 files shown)


test/integration/full/strict-csp/strict-csp.html
```diff
@@ -0,0 +1,47 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<meta charset="utf8" />
+		<script id="check-phantomjs">
+			// phantomjs requires running in non-strict csp
+			if (!window.PHANTOMJS) {
+				var thisScript = document.getElementById('check-phantomjs');
+				var csp = document.createElement('div');
```
stephenmathieson: Why not `var meta = document.createElement('meta')`?

straker: Was just being lazy to not have to add 2 different attributes through `createAttribute` :)

Issue #1517, opened 2019-04-24
aria-allowed-attr not returning correct results for aria-roledescription


straker:
Expectation: 
Aria-roledescription should pass for elements/roles listed in https://github.com/dequelabs/axe-core/issues/1216#issuecomment-454104190, return incomplete for all other roles, and fail for an element without a role

Actual: 
* Fails for `<a href="foobar.com" aria-roledescription="foo">foo</a>` (valid role but not supported on anchor elements, should incomplete)
* Fails for `<div role="button" aria-roledescription="foo">foo</button>` (valid role and supported on button elements/roles, should pass) 

Motivation: https://github.com/GoogleChrome/web.dev/issues/763

<pre><code>
axe-core version: 3.2.2
</code></pre>

dylanb:
@straker Have you thought through the accessibility supported nuances here?

straker:
Only in terms of the testing I did in https://github.com/dequelabs/axe-core/issues/1216#issuecomment-45410419

jeankaplansky:
New rule help here: https://dequeuniversity.com/rules/axe/3.4/aria-roledescription

chandana7393:
Tested, working fine.
![aria-roledescription](https://user-images.githubusercontent.com/45285854/67088316-788d8200-f1c2-11e9-9b1f-a856aba38f24.PNG)
**Tested Environment:**
Axe-coconut - 4.0.0.29104v
Chrome - 77.0.3865.120v
OS - Windows 10 64 bit.

Issue #1597, opened 2019-05-29
Add exception for ARIA role="text"


greenc05:
Invalid ARIA `role="text"` can be used to stop 'text splitting' in VoiceOver iOS https://axesslab.com/text-splitting/ This improves the user experience and does not negatively affect accessibility. 

Please add an exception, so that the presence of ARIA `role="text"` does not cause the following failure: '[role] values are not valid - ARIA roles must have valid values in order to perform their intended accessibility functions.'

utandon:
This exception would be very useful. Please add this asap :)

straker:
Thanks for bringing this to our attention. I've started internal discussions about this and we'll see what it would take to implement.

greenc05:
Wonderful thank you @straker

straker:
**tl;dr;** I propose we create a new rule for `role=text` and only support it on `span` and `div` elements, and only if the element does not contain any interactive children.

After doing further investigation I believe we can support `role=text` but in a very strcit capacity. Since it didn't make it into ARIA 1.1, we can only make our best judgment guess as to its uses. Based on the original discussion of `role=text` https://github.com/w3c/html-aria/issues/26 and the linked article about text-splitting, here are my thoughts.

First, we should only allow `role=text` on `span` and `div` elements. As `role=text` would override the native semantic of any element, we should only allow it for elements without implicit roles. The limited uses cases we have show it mostly on these two elements to fix the split-text problem of VoiceOver.  

However, this does not stop someone from wrapping an anchor element with an element with `role=text`, thus removing the anchors semantics:

```html
<!-- no longer announces as link, just text -->
<span role="text"><a href="/site.html">Click here</a></span> 
```

There was talk about UAs ignoring `role=text` if the [descendants of the element contain interactive elements](https://www.w3.org/WAI/ARIA/track/issues/1011). I certainly think we should do something similar. We already have plans for a new rule to [flag nested interactive elements as a problem](https://github.com/dequelabs/axe-core/issues/601), so we could share that logic here as well. 

Adding the exception about interactive children takes it outside the scope of how we normally handle allowed aria roles, so we would probably have to create a new rule just `role=text` where we add the checks we need.

greenc05:
Sounds good @straker I can think of one other use case for role text in addition to allowing this on div and span elements, that is to also allow this on paragraph elements as that's their native role (as long as they don't have nested interactive elements)? Such as:

[...]

PR #1745, opened 2019-08-02
feat(rule): aria-roledescription

straker:
Aria-roledescription required [too many nuances](https://github.com/dequelabs/axe-core/issues/1216#issuecomment-454104190) that couldn't adequately be captured in just an `unsupported` object:

- required matching a list of supported roles
- matcher objects don't allow matching of elements with implicit roles
- required failing on any element without a role

As such, it was easier to turn it into its own rule that could capture these nuances. This is the same problem we'll encounter when trying to [support `role="text"`](https://github.com/dequelabs/axe-core/issues/1597) as we'd want to make sure that it isn't used on an element with interactive children (something else that can't be captured in an `unsupported` object).

Closes issue: #1517

## Reviewer checks


lib/rules/aria-roledescription.json
```diff
@@ -0,0 +1,13 @@
+{
+	"id": "aria-roledescription",
+	"selector": "[aria-roledescription]",
+	"excludeHidden": false,
+	"tags": ["cat.aria", "best-practice"],
```
dylanb: Need to engage with the Jedi Council to determine whether this should be categorized under WCAG 4.1.2

My initial reaction is that it should because if used incorrectly, it means that the role is not being described correctly.

lib/checks/aria/aria-roledescription.js
```diff
@@ -0,0 +1,20 @@
+const role = axe.commons.aria.getRole(node);
+const supportedRoles = [
```
dylanb: Should we not make these configurable options? This way it can be changed in custom rules to

1. Support different AT environments (e.g. internal employee application with a controlled AT environment), and
1. Be changed easily in custom rules if support changes without having to wait for an axe-core release

straker: Good point

lib/rules/aria-roledescription.json
```diff
@@ -0,0 +1,13 @@
...
+	"metadata": {
+		"description": "Ensures aria-roledescription attribute is used correctly",
+		"help": "Aria-roledescription attribute must be used on elements with an implicit or explicit role"
```
WilcoFiers: The help text is supposed to be the shorter text. I also don't think this conveys quite enough information. Suggestion:

> description: Ensure aria-roledescription is only used on elements with an implicit or explicit role
> help: Use aria-roledescription on elements with a semantic role

lib/checks/aria/aria-roledescription.json
```diff
@@ -0,0 +1,23 @@
...
+			"combobox",
+			"menuitemcheckbox",
+			"menuitemradio"
+		]
+	},
+	"metadata": {
+		"impact": "serious",
+		"messages": {
+			"pass": "Used on a known supported role",
+			"incomplete": "Check that the aria-roledescription is announced",
```
WilcoFiers: can you add " by supported screen readers"?

lib/checks/aria/aria-roledescription.json
```diff
@@ -0,0 +1,23 @@
...
+			"radio",
+			"combobox",
+			"menuitemcheckbox",
+			"menuitemradio"
+		]
+	},
+	"metadata": {
+		"impact": "serious",
+		"messages": {
+			"pass": "Used on a known supported role",
```
WilcoFiers: Use full sentences please.

lib/checks/aria/aria-roledescription.json
```diff
@@ -0,0 +1,23 @@
...
+			"incomplete": "Check that the aria-roledescription is announced",
+			"fail": "Used on an element without an implicit or explicit role"
```
WilcoFiers: I suggest "Give the element a role that supports aria-roledescription"

lib/checks/aria/aria-roledescription.js
```diff
@@ -0,0 +1,14 @@
+options = options || {};
+
+const role = axe.commons.aria.getRole(node);
+const supportedRoles = options.supportedRoles || [];
+
+if (supportedRoles.includes(role)) {
+	return true;
+}
+
+if (role) {
```
WilcoFiers: Should we check if the role is `presentation|none` and fail those?

lib/rules/aria-roledescription.json
WilcoFiers: I'd say keep the tags consistent with what it was before we split out: `"cat.aria", "wcag2a", "wcag412"`

test/integration/rules/aria-roledescription/aria-roledescription.html
```diff
@@ -0,0 +1,9 @@
+<img aria-roledescription="my img" src="foo.png" id="pass1" />
+<button aria-roledescription="my button" id="pass2">button</button>
+<div role="combobox" aria-roledescription="my combobox" id="pass3"></div>
```
WilcoFiers: Can you put in cases for each of the check options?

lib/rules/aria-roledescription.json
```diff
@@ -0,0 +1,13 @@
+{
+	"id": "aria-roledescription",
+	"selector": "[aria-roledescription]",
+	"excludeHidden": false,
```
WilcoFiers: This is different from `aria-allowed-attr`, why did you change this? I don't think this is likely to cause false positives, but probably not any less than `aria-allowed-attr`. IMO it's probably safer to not do this and rely on people testing axe in multiple page states instead.

lib/checks/aria/aria-roledescription.js
```diff
@@ -0,0 +1,14 @@
+options = options || {};
```
jkodu: nitpick, destructuring will be easier to read & will remove line 4.

```
const { supportedRoles = [] } = options
```

lib/checks/aria/aria-roledescription.js
```diff
@@ -0,0 +1,14 @@
...
+
+const role = axe.commons.aria.getRole(node);
+const supportedRoles = options.supportedRoles || [];
+
+if (supportedRoles.includes(role)) {
+	return true;
+}
+
+if (role && role !== 'presentation' && role !== 'none') {
```
jkodu: nitpick, exit early again easier to read.

```
if(!role) {
  return false
}

if(!['presentation', 'none'].includes(role)) {
 return undefined
}
```

PR #1758, opened 2019-08-07
feat(l10n): add spanish translation

nvdaes:
Translate all messages, adding full spanish localization for Ace CLI and app by @daisy, suggested by @danielweck

Related PR: https://github.com/daisy/ace/pull/261

<< Describe the changes >>

Closes issue:

## Reviewer checks


nvdaes:
I've fixed a bug in es.json file.
Sorry for the inconveniences and thanks for reviewing this when you want.

nvdaes:
Hi, thanks for your attention. I think we can leave parent and child not translated. I have used single quotes trying to indicate that the contained expression is not translated. Parent and child can be translated as "padre" and "hijo", but the word role related to aria is writen as 'role' in code, so here is not translated. It can be translated as "rol", without "e", but in this case I think that 'parent roles' or 'child roles' is a shorten expression than the right translation, that in fact should include more words to build a sentence according to Spanish grammar, since the literal translation would be something like (roles padre" or "roles hijo", or "roles para el padre" and "roles para el hijo", and the "contain" verb is related to the main name 'parent' or 'child', not to roles, but in English it's OK that parent and child here aren't in plural, but in Spanish, imo, the sentence would be longer or grammatically weird.
I would say you can merge this like that. Anyway, if spanish speakers have suggestion in a later moment, or if something needs to be updated, we can do it.
Thanks.

straker:
No problem. I had only suggested it because in the other translations we have (including Portuguese) it's been translated. I'll try to get this merged after the weekend.

nvdaes:
Thanks. Sometimes I see, not in axe-core, since I havent seen translated messages, that certain guides about computing are not well translated into Spanish. I havent found a translation for ARIA (maybe it is done) to follow it as a reference, so sometimes I have left English expressions, thinking that this is mainly for techinical profiles knowing basics about DOM and English documents. I have never read in Spanish the expression child roles or parent roles translated, so, in dubdt, I prefer not to invent a new translation for this. For now you can merge like that. If later this needs to be changeded, we can use something like "parent with roles" (padres con roles), considering the meaning of the rule. I will spread this when you merge it, for example on my Twitter account or mailing lists (I have just about 30 o 40 followers, but some of them are interested in accessibility an programming), so that people can provide feedback for Spanish.

nvdaes:
Thanks, hope more localizations can be added. I will share the link to this PR so that people can provide feedback if they want.

Issue #1701, opened 2019-07-12
False positive: 'scrollable-region-focusable' failure for a scrollable region containing focusable content positioned using CSS


iamrafan:
Expectation: Not fail the 'scrollable-region-focusable' rule

Actual: There is a 'scrollable-region-focusable' failure here

Motivation: The scrollable region contains focusable elements, hence should pass this rule

Demo: https://www.bing.com/

<pre><code>axe-core version: 3.3.0-canary.5b05eb5
axe-Coconut version: 3.8.1</code></pre>

straker:
I'm unable to reproduce this on the bing page using axe-core v3.3.0 nor axe-coconut v3.8.1 (axe-core 3.3.0-canary.e9f9c18).

iamrafan:
@straker I am still able to reproduce this bug, please re-try
![image](https://user-images.githubusercontent.com/22068434/61327227-42a8be00-a7cd-11e9-92a1-761f8d7b3d91.png)

straker:
![image](https://user-images.githubusercontent.com/2433219/61331519-64ab3c00-a7df-11e9-92b9-3272435bcdc7.png)

dylanb:
I am seeing it, did you scroll down?

![image](https://user-images.githubusercontent.com/79245/61378464-72120600-a873-11e9-8131-90667e7a011d.png)

straker:
I did. I also tried to change the window size and still am not getting it.

The difficulty is that Bing seems to be sending different markup based on unknown factors. I'm getting completely different violations. Are either of you logged into Bing? Maybe language settings or location is determining it?

![image](https://user-images.githubusercontent.com/2433219/61382583-4c343380-a86a-11e9-8d04-72aa74578263.png)

WilcoFiers:
I'm having the same issues as Steve. @iamrafan can you put together something so we can reliably reproduce the issue?

iamrafan:
@iamrafan I have created an example similar to the one on bing. Please let me know if it helps

https://codepen.io/iamrafan/pen/LwbPvz

straker:
@iamrafan  I'm still not seeing the error. I even forked your code and set it up in debug mode just to make sure. Running `axe.run` manually shows that `scrollable-region-focus` is in the `inapplicable` section of the results. 

![image](https://user-images.githubusercontent.com/2433219/62572615-589c2280-b851-11e9-9a5c-c8c8f1cdf624.png)

[...]

PR #1788, opened 2019-08-27
fix(scrollable-region-focus): ignore scrollable regions without content

straker:
`hasContentVirtual` would always run the aria label check even when the `ignoreAria` flag was set to `true`, it also resulted in the entire check always returning `true`. 

The test wanted false but was asserting `true`.

Closes issue: #1701 (will now properly ignore the scrollable region that doesn't have content)

## Reviewer checks

### DIFF
--- a/test/rule-matches/scrollable-region-focusable-matches.js
+++ b/test/rule-matches/scrollable-region-focusable-matches.js
@@ -82,6 +82,14 @@ describe('scrollable-region-focusable-matches', function() {
 		assert.isTrue(actual);
 	});
 
+	it('returns false when element overflow is scroll but has no content', function() {
+		var target = queryFixture(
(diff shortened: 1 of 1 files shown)


test/commons/dom/has-content-virtual.js
```diff
@@ -41,7 +41,7 @@ describe('dom.hasContentVirtual', function() {
 	it('is false if the element has an aria label but `ignoreAria=true`', function() {
 		fixture.innerHTML = '<div id="target" aria-label="my-label">	</div>';
 		tree = axe.utils.getFlattenedTree(fixture);
-		assert.isTrue(
+		assert.isFalse(
```
WilcoFiers: Yikes... That was probably me.

Issue #1762, opened 2019-08-09
aria-required-children false positive for almost empty nodes


WilcoFiers:
The `aria-required-children` rule has a `reviewEmpty` option, which tells axe-core what roles to fail when they are empty, and which ones not to. For example, `listbox` is allowed to be empty.

```html
<div role="listbox" aria-label="Select nothing"></div>
```

That element shouldn't fail. However, if we put something innocuous in it, like this:

```html
<div role="listbox" aria-label="Select nothing">
  <div></div>
</div>
```

Axe-core fails the element. This happens because axe-core only looks if there are any children at all. I think axe-core needs to be a bit smarter about when to consider an item "empty".

I think the best way to do this, is to see if there are any "owned elements". I'm not entirely sure about the best way to solve this. We could maybe look at the owned elements and check if they have a role or an accessible name? That may create other gaps.

straker:
At what point does the element become non-empty? If it has a role but no text content? When it has just text content? What about implicit role elements like `<nav>`? Or if the `<nav>` element has an `aria-label` on it but no text content?

jeankaplansky:
@straker : technically, an element becomes non-empty when you add any text or markup within the start and end tags (XML definition). Attributes aren't considered content. Here's Mozilla's official: https://developer.mozilla.org/en-US/docs/Glossary/Empty_element

chandana7393:
Tested, working as expected.
![aria_role_children](https://user-images.githubusercontent.com/45285854/64336685-32f26c80-cffb-11e9-88e0-8d4d07fae99a.PNG)

**Tested Environment:**
Attest - 2.6.1.24714v
Axe - 3.9.0.24714v
Axe-coconut - 3.9.0.24804v
Chrome - 76.0.3809.132v
Firefox - 68.0.2v
OS - Windows 10 64 bit.

PR #1791, opened 2019-08-28
fix(aria-required-children): allow reviewEmpty nodes to have empty children

straker:
Allow `reviewEmpty` roles to have empty children so long as the child does not have a role.

Closes issue: #1762

## Reviewer checks

### DIFF
--- a/lib/checks/aria/required-children.js
+++ b/lib/checks/aria/required-children.js
@@ -95,11 +95,15 @@ function missingRequiredChildren(node, childRoles, all, role) {
 	return null;
 }
 
-function hasChildWithRole(node) {
+function hasDecendantWithRole(node) {
 	return (
 		node.children &&
 		node.children.some(child => {
-			return axe.commons.aria.getRole(child) || hasChildWithRole(child);
+			const role = axe.commons.aria.getRole(child);
+			return (
+				!['presentation', 'none', null].includes(role) ||
+				hasDecendantWithRole(child)
+			);
 		})
 	);
 }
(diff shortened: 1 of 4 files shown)


lib/checks/aria/required-children.js
```diff
@@ -94,6 +95,15 @@ function missingRequiredChildren(node, childRoles, all, role) {
 	return null;
 }
 
+function hasChildWithRole(node) {
```
WilcoFiers: This is a bit of a misnomer. Can you rename this to `hasDescendantsWithRole`?

lib/checks/aria/required-children.js
```diff
@@ -94,6 +95,15 @@ function missingRequiredChildren(node, childRoles, all, role) {
...
+	return (
+		node.children &&
+		node.children.some(child => {
+			return axe.commons.aria.getRole(child) || hasChildWithRole(child);
```
WilcoFiers: This should ignore role=none/presentation. Should we add an option to `getRole` so that it returns `null` if the role is `presentation` or `none`? We seem to be doing a lot of this: `!['presentation', 'none'].includes(role)`

Maybe we can make this:

```js
getRole(elm, { noPresentation: true })
```

lib/checks/aria/required-children.js
```diff
@@ -119,7 +129,8 @@ this.data(missing);
 // Only review empty nodes when a node is both empty and does not have an aria-owns relationship
 if (
 	reviewEmpty.includes(role) &&
-	node.children.length === 0 &&
+	!hasContentVirtual(virtualNode) &&
```
WilcoFiers: I think this should have `ignoreAria: true`, otherwise I think this will be picked up:

```html
<div role="list" aria-label="My list"><span></span></div>
```

Issue #1780, opened 2019-08-21
Check the "axeVersion" property for axe.configure for compatibility


WilcoFiers:
When calling `axe.configure()` with a "axeVersion" property, axe-core should check if that version of the custom rule is compatible with the axe-core version that's used. Custom rules are compatible with all patch and minor updates that are the same as, or newer than those in "axeVersion". So:

```js
axe.configure({
  rulesetVersion: '2.5.1',
  rules,
  checks
})
```

- 2.5.1 is compatible with 2.5.1
- 2.5.1 is compatible with 2.6.2
- 2.5.1 is compatible with 2.5.1-canary.2664bae
- 2.5.1-canary.2664bae is compatible with 2.5.2
- 2.5.1 is not compatible with 2.5.0
- 2.5.1 is not compatible with 2.4.3
- 2.5.1 is not compatible with 3.x.x
- 2.5.1-canary.2664bae is not compatible with 2.5.1
- 2.5.1-canary.2664bae is not compatible with 2.5.1-canary.a5d727c (since we can't tell which was released first, canary versions are only compatible with themselves, and any future stable release).

If axe.configure is called with an incompatible version, axe-core should throw a **meaningful** error.

**Edit**: The `ver` property has been renamed to `axeVersion`. We will add support for `ver`, but have it only be available as a deprecated feature, that will be removed as of the next major release.

aellsey:
@WilcoFiers @awpearlm will we need design for the "meaningful error"? How do you see this change impacting the user experience?

stephenmathieson:
Why "ver"? It's not a word.

WilcoFiers:
See update above, renamed `ver` to `rulesetVersion`.

isner:
Corresponding Comply ticket to accommodate this change:

**Accommodate axe 3.4 version-compatibility checking**
https://dequesrc.atlassian.net/browse/COM-1865

WilcoFiers:
I've updated the above example to include canary versions.

chandana7393:
@straker what needs to be QA here?? Can you please provide the required steps.

straker:
Just that setting different `axeVersion` values of against the used version of axe either throws an error or doesn't in accordance with @WilcoFiers description.

somaalapati:
@chandana7393 steps to verify this ticket:

1. Launch Attest extension.
2. Navigate to Rules panel.
3. Select axe-version '3.0.3' from axe versions dropdown.
[...]

PR #1793, opened 2019-08-29
feature(configure): add ver property that checks compatibility axe.version

straker:
Closes issue: #1780

## Reviewer checks

### DIFF
--- a/lib/core/public/configure.js
+++ b/lib/core/public/configure.js
@@ -8,6 +8,21 @@ function configureChecksRulesAndBranding(spec) {
 		throw new Error('No audit configured');
 	}
 
+	if (spec.ver) {
+		let [major, minor, patch] = spec.ver.split('.').map(Number);
+		let [axeMajor, axeMinor, axePatch] = axe.version.split('.').map(Number);
+
+		if (
+			major !== axeMajor ||
+			axeMinor < minor ||
+			(axeMinor === minor && axePatch < patch)
+		) {
+			throw new Error(
+				'Configured version is not compatible with current axe version'
+			);
+		}
+	}
+
 	if (
 		spec.reporter &&
 		(typeof spec.reporter === 'function' || reporters[spec.reporter])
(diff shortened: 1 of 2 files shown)


doc/API.md
```diff
@@ -164,7 +164,8 @@ axe.configure({
 	reporter: 'option',
 	checks: [Object],
 	rules: [Object],
-	locale: Object
+	locale: Object,
+	ver: String
```
stephenmathieson: Why `ver` rather than `version`?

lib/core/public/configure.js
```diff
@@ -8,6 +8,21 @@ function configureChecksRulesAndBranding(spec) {
...
 	}
 
+	if (spec.ver) {
+		let [major, minor, patch] = spec.ver.split('.').map(Number);
+		let [axeMajor, axeMinor, axePatch] = axe.version.split('.').map(Number);
+
+		if (
+			major !== axeMajor ||
+			axeMinor < minor ||
+			(axeMinor === minor && axePatch < patch)
```
stephenmathieson: Should this be `axePatch <= patch`?

test/core/public/configure.js
```diff
@@ -555,4 +557,64 @@ describe('axe.configure', function() {
 			});
 		});
 	});
+
+	describe('given a version property', function() {
```
stephenmathieson: Given a "version" property or a "ver" property?

lib/core/public/configure.js
straker: we shouldn't error if the patches are equal

stephenmathieson: Good point. My bad.

stephenmathieson:
I think we should also update the TypeScript definition.

straker:
@WilcoFiers did you have any particular reason why you wanted the name of the prop to be `ver` instead of `version`?

doc/API.md
WilcoFiers: Discussed during standup today. Proposal is to deprecate `ver` and to use `axeVersion`. We'll support both for a while, but will eventually drop `ver`.

@downzer0 WDYT?

WilcoFiers: Got feedback, yup, we're going with `axeVersion`, and have `ver` be an alternative, only if `axeVersion` isn't set. Please mark `ver` as deprecated in the docs as well.

Issue #1784, opened 2019-08-23
Add lang query string to help URLs


WilcoFiers:
To support DQU's localisation, axe-core's help URLs should include a language code, for example:

```
https://dequeuniversity.com/rules/axe/3.3/html-has-lang?lang=jp
```

This can be set when `axe.configure()` is called with a new locale.

jeankaplansky:
@straker: This isn't a new rule, but is it something we should include in the release notes?

jeankaplansky:
Added to release notes 10/14/2019

padmavemulapati:
Verified this ticket, I am seeing the help URL, for language attribute is:  https://dequeuniversity.com/rules/axe/3.3/html-has-lang?lang=ja (not lang=jp) 
@WilcoFiers , can you please confirm that we are expecting "japanese" language , is n't it?

straker:
Opened a separate issue #1861.

padmavemulapati:
As this issue is tracking in #1861, closing this ticket

PR #1794, opened 2019-08-29
feat(audit): add lang query paramenter to help urls

straker:
Closes issue: #1784

## Reviewer checks

### DIFF
--- a/lib/core/base/audit.js
+++ b/lib/core/base/audit.js
@@ -37,6 +37,7 @@ function Audit(audit) {
 	this.brand = 'axe';
 	this.application = 'axeAPI';
 	this.tagExclude = ['experimental'];
+	this.lang = 'en';
 
 	this.defaultConfig = audit;
 	this._init();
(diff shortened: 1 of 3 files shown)


straker:
As a side note, I noticed that a lot of the function in audit don't have any tests. For example, `addCheck`, `addRule`, and `applyLocale` don't have any specific tests.

lib/core/base/audit.js
```diff
@@ -651,7 +657,8 @@ function getHelpUrl({ brand, application }, ruleId, version) {
 		'/' +
 		ruleId +
 		'?application=' +
-		application
+		application +
+		(lang && lang !== 'en' ? '&lang=' + lang : '')
```
WilcoFiers: Can you escape `lang` here, and maybe while you're at it, do the same for `application`?

test/core/base/audit.js
```diff
@@ -236,6 +236,24 @@ describe('Audit', function() {
 				'https://dequeuniversity.com/rules/axe/1.0/target?application=axeAPI'
 			);
 		});
+		it('sets the lang query if locale has been set', function() {
```
WilcoFiers: Can you add a test to make sure this is passed in from `axe.configure`? That's the API that we actually want to have direct this.

PR #1805, opened 2019-09-09
[WIP] test: switch to karma for testing

straker:
Switching out grunt for running tests to using karma. 

**NOTE:** tests will fail until I swap out `npm test` to use karma

## Reviewer checks


test/core/base/audit.js
```diff
@@ -604,100 +604,96 @@ describe('Audit', function() {
...
-					id: 'no-preload-check',
-					evaluate: function(node, options, vNode, context) {
-						noPreloadCheckedInvokedTime = new Date();
-						noPreloadRuleCheckEvaluateInvoked = true;
-						var ready = this.async();
-						setTimeout(function() {
-							ready(true);
-						}, 1000);
-					}
+		it('should run rules (that do not need preload) and preload assets simultaneously', function(done) {
```
straker: showing that we can remove all our phantomjs conditionals

test/core/index.js
```diff
@@ -1,10 +0,0 @@
-describe('index', function() {
```
straker: This file wasn't doing anything as `src/core/index.js` wasn't redefining these things

test/core/public/run-rules.js
```diff
@@ -344,7 +344,7 @@ describe('runRules', function() {
 		});
 
 		fixture.innerHTML =
-			'<div id="t1"><span></span></div><div id="t2"><em></em></div>';
+			'<div id="t1"><em></em></div><div id="t2"><em></em></div>';
```
straker: We no longer have karma output in the test file so these selectors were now unique and the target didn't need a child selector. Updated to force non-unique nodes

test/core/utils/preload-cssom.js
```diff
@@ -39,7 +39,7 @@ describe('axe.utils.preloadCssom', function() {
 		var actual = axe.utils.preloadCssom({ treeRoot: treeRoot });
 		actual
 			.then(function(cssom) {
-				assert.isAtLeast(cssom.length, 2);
+				assert.isAtLeast(cssom.length, 1);
```
straker: We no longer have additional CSS files in test file so there is only the 1 file stylesheet now

test/testutils.js
```diff
@@ -10,6 +10,10 @@ if (window.__AXE_EXTENSION__) {
 /*eslint indent: 0*/
 var testUtils = {};
 
+var fixture = document.createElement('div');
```
straker: We don't use the `runner.tmpl` file anymore so need to create the fixture element

jkodu: You could use something like [karma fixture](https://www.npmjs.com/package/karma-fixture). This will help reduce the number of changes that you have to do to make integration tests work.

test/core/base/audit.js
```diff
@@ -604,100 +604,96 @@ describe('Audit', function() {
...
+			/**
+			 * Note:
+			 * overriding and resolving both check and preload with a delay,
```
jkodu:
```suggestion
			 * overridden and resolving both check and preload with a delay,
```

straker:
Closing for now. Will try again after 4.0

Issue #525, opened 2017-09-15
getElmScrollRecursive fails on SVG in some browsers


fofr:
When recursing through the DOM to calculate the scroll position here: https://github.com/dequelabs/axe-core/blob/83daaa53023c8ad5ea927244617cf7807bf1fd37/lib/core/utils/scroll-state.js#L32-L40

If the code encounters an SVG it essentially runs, `document.querySelector('svg').children`. This doesn't always return an HTMLCollection object which in turn causes `Array.from` to error.

JS fiddle demonstrating issue:
https://jsfiddle.net/uvftjtre/2/ (version 3.0.0.alpha.4)

Running the above example in EDGE:
```
Array.from: argument is not an Object
```

Running in IE11 will return:
```
TypeError: Array.from requires an array-like object - not null or undefined
```

This error surfaced when running automated tests through Phantom JS 1.9.7

More context:
https://stackoverflow.com/questions/7935689/what-is-the-difference-between-children-and-childnodes-in-javascript#7935719

WilcoFiers:
Problem turns out to be that .children is undefined. The work around is to use .childNodes instead and filter them for elements.

benbcai:
@WilcoFiers This issue causes our accessibility tests to fail on any page that contains an SVG element. Looks like this is scheduled to be fixed in axe-core 3.4. Do you have a timeframe when 3.4 is scheduled to be released?

emilyrohrbough:
Do you expect the changes be to add fallback for childNodes like:
```diff
- return Array.from(root.children).reduce((scrolls, elm) => { 
+ return Array.from(root.children || root.childNodes || []).reduce((scrolls, elm) => { 
```
or are the scroll position of SVG children necessary to worry about?

benbcai:
Another option could something like [this](https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/children#Polyfill).

```diff
+ var i = 0
+ var node;
+ var nodes = root.childNodes;
+ var children = [];

+ while (node = nodes[i++]) {
+ 	if (node.nodeType === Node.ELEMENT_NODE) {
+ 		children.push(node);
+ 	}
+ }
- return Array.from(root.children).reduce((scrolls, elm) => { 
+ return Array.from(children).reduce((scrolls, elm) => { 
```

@WilcoFiers What are your thoughts?

benbcai:
[...]

PR #1820, opened 2019-09-20
fix(utils): Fix error in IE when getting scroll state on page with SVG elements

benbcai:
For any page in IE that has an SVG element, accessing `.children` on an SVGElement when getting the scroll state would return `undefined` and ends up throwing an exception. Other browsers treat SVGs as HTMLCollections but IE treats them as SVGElement.

Closes issue: #525

## Reviewer checks

### DIFF
--- a/lib/core/utils/scroll-state.js
+++ b/lib/core/utils/scroll-state.js
@@ -14,13 +14,16 @@ function setScroll(elm, top, left) {
  * Create an array scroll positions from descending elements
  */
 function getElmScrollRecursive(root) {
-	return Array.from(root.children).reduce((scrolls, elm) => {
-		const scroll = axe.utils.getScroll(elm);
-		if (scroll) {
-			scrolls.push(scroll);
-		}
-		return scrolls.concat(getElmScrollRecursive(elm));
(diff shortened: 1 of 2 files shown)


benbcai:
@WilcoFiers @straker Could you please take a look at this pull request? This issue is breaking my tests and blocking by project.

lib/core/utils/scroll-state.js
```diff
@@ -14,13 +14,16 @@ function setScroll(elm, top, left) {
...
  */
 function getElmScrollRecursive(root) {
-	return Array.from(root.children).reduce((scrolls, elm) => {
-		const scroll = axe.utils.getScroll(elm);
-		if (scroll) {
-			scrolls.push(scroll);
-		}
-		return scrolls.concat(getElmScrollRecursive(elm));
-	}, []);
+	return Array.from(root.children || root.childNodes || []).reduce(
```
stephenmathieson: I'm late to the party, but it would be nice to see a "SVGs in IE don't have .children" (or alike) comment here. Otherwise a future developer may not know what this is for.

PR #1822, opened 2019-09-24
fix(label-content-name-mismatch): ignore private space unicode

### LINKED ISSUE #1635: false positive on label-content-name-mismatch rule
The rule seems to evaluate the innerText value & compares it with the aria-label value. This results in false positive, when we have icon as a child element that has some content, even though visually the user only sees icon instead of text.
eg where it fails:
```
<a id="link1" tabindex="0" href="./some-link" aria-label="Favorites">
  <div><span aria-hidden="true" class="some-icon"></span></div>
  <div><span aria-hidden="true">Favorites</span></div>
</a>
```
https://codepen.io/devaradhanm/pen/xoZdmd

including @iamrafan , for his inputs too.

straker:
Part 1 for #1635 to fix using private use unicode characters. 

## Reviewer checks

### DIFF
--- a/lib/commons/text/unicode.js
+++ b/lib/commons/text/unicode.js
@@ -91,6 +91,7 @@ function getUnicodeNonBmpRegExp() {
 		'\u25A0-\u25FF' + // Geometric Shapes
 		'\u2600-\u26FF' + // Misc Symbols
 		'\u2700-\u27BF' + // Dingbats
+		'\uE000-\uF8FF' + // Private Use
 			']'
 	);
 }
--- a/lib/commons/text/unicode.js
+++ b/lib/commons/text/unicode.js
@@ -48,6 +48,7 @@ text.removeUnicode = function removeUnicode(str, options) {
 	}
 	if (nonBmp) {
 		str = str.replace(getUnicodeNonBmpRegExp(), '');
+		str = str.replace(getSupplementaryPrivateUseRegExp(), '');
 	}
 	if (punctuations) {
 		str = str.replace(getPunctuationRegExp(), '');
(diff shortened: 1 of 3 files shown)


lib/commons/text/unicode.js
```diff
@@ -91,6 +91,7 @@ function getUnicodeNonBmpRegExp() {
 		'\u25A0-\u25FF' + // Geometric Shapes
 		'\u2600-\u26FF' + // Misc Symbols
 		'\u2700-\u27BF' + // Dingbats
+		'\uE000-\uF8FF' + // Private Use
```
jkodu: Going with definition from here:

>  Does Unicode have private-use characters?

> A: Yes. There are three ranges of private-use characters in the standard. The main range in the BMP is U+E000..U+F8FF, containing 6,400 private-use characters. That range is often referred to as the Private Use Area (PUA). But there are also two large ranges of supplementary private-use characters, consisting of most of the code points on Planes 15 and 16: U+F0000..U+FFFFD and U+100000..U+10FFFD. Together those ranges allocate another 131,068 private-use characters. Altogether, then, there are 137,468 private-use characters in Unicode.

It looks like we are only ignoring ones in BMP and not the supplementary private-use characters. Shouldn't we consider the supplementary ones too?

straker: I thought about it, but I haven't been able to find a 100% reliable way to test them. I wasn't able to find anything directly on the internet, but was able to piece together how we might be able to using surrogate pairs of the [high](https://www.unicode.org/charts/PDF/UD800.pdf) and [low](https://www.unicode.org/charts/PDF/UDC00.pdf) surrogates together. Theoretically this regex should catch them, but I can't be sure.

```
/[\uDB80-\uDBBF][\uDC00-\uDFFD]/
```

WilcoFiers: Given that nothing in that range is likely to be something Dragon knows how to handle, I think it's a pretty safe bet to just exclude the entire range.

PR #1829, opened 2019-09-27
fix(label-content-name-mismatch): ignore ligature fonts

straker:
Part 2 for #1635 to fix using ligature icon fonts.

## Reviewer checks


lib/commons/text/is-icon-ligature.js
```diff
@@ -0,0 +1,189 @@
...
+		// ignore color differences, only deal with pixel placement differences
+		let compareDataHasPixel = false;
+		let compareWithHasPixel = false;
+		for (let j = 0; j < 4; j++) {
+			compareDataHasPixel = compareDataHasPixel || !!compareData[i + j];
+			compareWithHasPixel = compareWithHasPixel || !!compareWith[i + j];
+		}
+
+		if (compareDataHasPixel !== compareWithHasPixel) {
+			differences++;
```
AutoSponge: it looks like you can check differences against the pixelThreshold here and break if it exceeds before completing the nested for-loops.

lib/commons/text/is-icon-ligature.js
```diff
@@ -0,0 +1,189 @@
...
+	}
+	font.occurances++;
+
+	// 30px was chosen to account for common ligatures in normal fonts
+	// such as fi, ff, ffi. If a ligature would add a single column of
+	// pixels to a 30x30 grid, it would not meet the statistical significance
+	// threshold of 10% (30x30 = 900; 30/900 = 3.333%). this also allows for
+	// more than 1 column differences (60/900 = 6.666%) and things like
+	// extending the top of the f in the fi ligature.
+	let fontStyle = `30px ${fontFamily}`;
```
AutoSponge: Just a note in case we have perf issues: a 10px grid would take nearly an order of magnitude less time to diff.

lib/commons/text/is-icon-ligature.js
AutoSponge: we can also shrink the number of iterations if we use a view, combining the 8-bit values into larger ones. In this case, since we're only looking at pixel existence, it seems we should be able to turn these arrays into 32-bit arrays expecting the result to be 0 or non-zero (because we ignore color).

lib/commons/text/is-icon-ligature.js
straker: A 10px grid would not give you enough leeway to calculate the difference between `f` and `fi`. Just changing a single column of pixels to a 10px grid would create a 10% diff.

lib/commons/text/is-icon-ligature.js
straker: The performance hit from this function will be from the two `getImageData` calls and not the double for loop which is just iterating over every pixel once (3600 iterations). Optimizing the loop seems like a micro-optimization when we are already short circuiting the entire algorithm if the font has been seen 3 times.

AutoSponge: When I looked at it, you're not iterating over each pixel, but each color. I think iterating each pixel would be clearer. Having fewer iterations is a side-effect.

AutoSponge: something like this:
```js
const compareData = new Uint32Array(context.getImageData(0, 0, width, width).data.buffer);
// ...
const compareWith = new Uint32Array(context.getImageData(0, 0, width, width).data.buffer);
const differences = compareData.reduce((diff, pixel, i) => {
  if (pixel === 0 && compareWith[i] === 0) return diff;
  if (pixel !== 0 && compareWith[i] !== 0) return diff;
  return diff++
}, 0)
```

test/commons/text/is-icon-ligature.js
```diff
@@ -0,0 +1,191 @@
...
+		function() {
+			var target = queryFixture(
+				'<div id="target" style="font-family: \'Material Icons\'">delete</div>'
+			);
+			assert.isTrue(isIconLigature(target.children[0]));
+		}
+	);
+
+	(fontApiSupport ? it : it.skip)(
+		'should return true after the 3rd time the font is an icon',
```
AutoSponge: "3rd time" is arbitrary? It's not explained. I'm not sure that works for all cases. Is it possible to have a "normal" font with added ligatures?

straker: The use cases we are optimizing for:
- a non-icon ligature font (95% of cases)
- an exclusive icon ligature font (4.999%)
- a mix of regular text and icon ligatures (should be extremely rare. Material Icon doesn't even have data to draw latin characters)

For the rare case, we have [no choice but to check it every time](https://github.com/dequelabs/axe-core/pull/1829/files#diff-d286a2c961c899ecddcb0ee4aaf2ef48R124-R127). For the other two, we can identify them quickly (using the arbitrary 3 times seen as chosen by Wilco). If we encounter problems we can tweek things when we know about them.

lib/commons/text/is-icon-ligature.js
```diff
@@ -0,0 +1,189 @@
...
+		// to know if it's an icon or not
+	}
+	font.occurances++;
+
+	// 30px was chosen to account for common ligatures in normal fonts
+	// such as fi, ff, ffi. If a ligature would add a single column of
+	// pixels to a 30x30 grid, it would not meet the statistical significance
+	// threshold of 10% (30x30 = 900; 30/900 = 3.333%). this also allows for
+	// more than 1 column differences (60/900 = 6.666%) and things like
+	// extending the top of the f in the fi ligature.
```
AutoSponge: common ligatures should not get a pass on this rule. In these cases, the visible text (w/ `fi` ligature) "final" should match an `aria-label`. Only "final" converted to an icon would get a pass since the label/accName could be anything associated visually with the icon.

lib/commons/text/is-icon-ligature.js
straker: In terms of performance, I don't think converting from an Uint8Array to a Uint32Array (twice) and then iterating over it is significantly faster than looping over a Uint8Array once (I would even counter that it would be slower). 

I also don't think it makes things more clearer as you have to understand the same information to know why you'd want to do the conversion in the first place.

lib/commons/text/is-icon-ligature.js
```diff
@@ -0,0 +1,210 @@
...
+	context.fillText(firstChar, 0, 0);
+	const compareData = new Uint32Array(
+		context.getImageData(0, 0, width, fontSize).data.buffer
+	);
+
+	// if the font doesn't even have character data for a single char then
+	// it has to be an icon ligature (e.g. Material Icon)
+	if (!compareData.some(pixel => pixel)) {
+		font.numLigatures++;
+		return true;
```
AutoSponge: Good point, I like this.

lib/commons/text/is-icon-ligature.js
```diff
@@ -0,0 +1,210 @@
...
+		return width + context.measureText(char).width;
+	}, 0);
+	const actualWidth = context.measureText(nodeValue).width;
+
+	const pixelDifference = differences / compareData.length;
+	const sizeDifference = 1 - actualWidth / expectedWidth;
+
+	if (
+		pixelDifference >= differenceThreshold &&
+		sizeDifference >= differenceThreshold
```
AutoSponge: right now it's fine but later we might find we need to split pixel and size diff thresholds.

PR #1836, opened 2019-10-02
chore: ensures axe Pro api

AutoSponge:
This adds tests that will help prevent breaking changes in axe-core that axe Pro depends on.

Closes issue: N/A (Q3 rock)

## Reviewer checks

### DIFF
--- a/build/test/config.js
+++ b/build/test/config.js
@@ -26,7 +26,10 @@ exports = module.exports = function(grunt, options) {
 					'http://' + host + ':<%= connect.test.options.port %>/test/commons/',
 					'http://' +
 						host +
-						':<%= connect.test.options.port %>/test/integration/rules/'
+						':<%= connect.test.options.port %>/test/integration/rules/',
+					'http://' +
+						host +
+						':<%= connect.test.options.port %>/test/integration/api/external/'
 				],
 				run: true,
 				growlOnSuccess: false,
--- a/test/integration/api/external/index.html
+++ b/test/integration/api/external/index.html
@@ -1,4 +1,4 @@
-<!DOCTYPE html PUBLIC "-//W3C//DTD HTML+ARIA 1.0//EN" "http://www.w3.org/WAI/ARIA/schemata/html4-aria-1.dtd">
+<!DOCTYPE html>
 <html lang="en">
 	<head>
 		<title>External API Test</title>
--- a/test/integration/api/external/index.html
+++ b/test/integration/api/external/index.html
@@ -1,0 +1,28 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD HTML+ARIA 1.0//EN" "http://www.w3.org/WAI/ARIA/schemata/html4-aria-1.dtd">
+<html lang="en">
+	<head>
+		<title>External API Test</title>
+
+		<link
+			rel="stylesheet"
+			type="text/css"
+			href="/node_modules/mocha/mocha.css"
+		/>
+		<script src="/node_modules/mocha/mocha.js"></script>
+		<script src="/node_modules/chai/chai.js"></script>
+		<script src="/axe.js"></script>
+		<script>
+			mocha.setup({
+				timeout: 10000,
+				ui: 'bdd'
+			});
+			var assert = chai.assert;
+		</script>
+		<script src="/test/integration/no-ui-reporter.js"></script>
+	</head>
+	<body>
+		<div id="mocha"></div>
+		<script src="index.js"></script>
+		<script src="/test/integration/adapter.js"></script>
+	</body>
+</html>
--- a/test/integration/api/external/index.js
+++ b/test/integration/api/external/index.js
@@ -1,0 +1,216 @@
+// describes the API used by axe Pro
+describe('external API', function() {
+	'use strict';
+
+	afterEach(function() {
+		// setup _tree as needed, always reset
+		axe._tree = null;
+	});
+
+	describe('axe.commons.text.sanitize', function() {
+		it('must be a function with the signature String -> String', function() {
+			assert.isString(axe.commons.text.sanitize(''));
+			assert.isString(axe.commons.text.sanitize('not empty'));
+		});
+	});
+
+	describe('axe.utils.getBaseLang', function() {
+		it('must be a function with the signature String -> String', function() {
+			assert.isString(axe.utils.getBaseLang(''));
+			assert.isString(axe.utils.getBaseLang('not empty'));
+		});
+	});
+
+	describe('axe.utils.validLangs', function() {
+		it('be a function with the signature * -> [String]', function() {
+			assert.isArray(axe.utils.validLangs());
+			axe.utils.validLangs().forEach(assert.isString);
+			assert.isArray(axe.utils.validLangs(document));
+			assert.isArray(axe.utils.validLangs({}));
+		});
+	});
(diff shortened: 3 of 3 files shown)


test/integration/api/external/index.js
```diff
@@ -0,0 +1,245 @@
...
+	document.createElement('article')
+];
+
+var inTree = [];
+var walker = collectNodes();
+var next = walker.iterate().next();
+while (!next.done) {
+	inTree.push(next.value);
+	next = walker.iterate().next();
+}
```
AutoSponge: This worked as a spreadable array until I realized we can't use Symbol or rest/spread.

straker:
Question: Most of these checks are just checking that the API exists, is a function, and returns the correct thing. Should these be unit tests in the respected files rather than an integration test all in one?

test/integration/api/external/index.js
```diff
@@ -0,0 +1,245 @@
...
+describe('external API', function() {
+	'use strict';
+
+	afterEach(function() {
+		// setup _tree as needed, always reset
+		axe._tree = null;
+	});
+
+	describe('axe.commons.text.sanitize', function() {
+		it('typeof fn', function() {
```
stephenmathieson: What about `it('is a function')`? The test name should be human-readable IMO.

test/integration/api/external/index.js
```diff
@@ -0,0 +1,245 @@
...
+			assert.isFunction(axe.commons.text.sanitize);
```
stephenmathieson: Is this test actually useful?

test/integration/api/external/index.js
```diff
@@ -0,0 +1,245 @@
...
+		});
+		it('str => str', function() {
```
stephenmathieson: What?

test/integration/api/external/index.js
```diff
@@ -0,0 +1,245 @@
...
+			assert.isArray(axe.utils.validLangs(document));
+			assert.isArray(axe.utils.validLangs({}));
+		});
+	});
+
+	describe('axe.commons.dom.isVisible', function() {
+		it('typeof fn', function() {
+			assert.isFunction(axe.commons.dom.isVisible);
+		});
+		it('el => bool', function() {
```
stephenmathieson: Maybe: `it('returns a boolean when given an element reference')`?

test/integration/api/external/index.js
```diff
@@ -0,0 +1,245 @@
...
+			var el = randomNodeInTree(isElement);
+			console.log(el);
```
stephenmathieson: `console.log()`

test/integration/api/external/index.js
```diff
@@ -0,0 +1,245 @@
...
+			var role = axe.commons.aria.implicitRole(randomNodeInTree());
+			assert.oneOf(role, implicitRolesOrNull);
+		});
+	});
+
+	describe('axe.commons.aria.lookupTable.role', function() {
+		it('typeof obj', function() {
+			assert.isObject(axe.commons.aria.lookupTable.role);
+		});
+		it('str <=> str (role.type)', function() {
```
stephenmathieson: I have no idea what this means.

test/integration/api/external/index.js
```diff
@@ -0,0 +1,245 @@
...
+	});
+
+	describe('axe.utils.getFlattenedTree', function() {
+		it('typeof fn', function() {
+			assert.isFunction(axe.utils.getFlattenedTree);
+		});
+		it('el => arr', function() {
+			assert.isArray(axe.utils.getFlattenedTree(document.body));
+		});
+		it('el => [{root}]', function() {
```
stephenmathieson: What is this?

test/integration/api/external/index.js
```diff
@@ -0,0 +1,245 @@
...
+			assert.isFunction(axe.utils.getNodeFromTree);
+		});
+		it('node => vnode', function() {
+			axe._tree = axe.utils.getFlattenedTree(document.body);
+			assert.oneOf(
+				axe.utils.getNodeFromTree(randomNodeInTree()),
+				flat(axe._tree[0])
+			);
+		});
+		it('!node => ()', function() {
```
stephenmathieson: I have no idea what this means.

test/integration/api/external/index.js
```diff
@@ -0,0 +1,245 @@
...
+			assert.isUndefined(axe.utils.getNodeFromTree(randomElement()));
+		});
+	});
+
+	describe('axe.commons.dom.isOpaque', function() {
+		it('typeof fn', function() {
+			assert.isFunction(axe.commons.dom.isOpaque);
+		});
+		it('el inTree => bool', function() {
```
stephenmathieson: More confusing test names.

AutoSponge:
> Question: Most of these checks are just checking that the API exists, is a function, and returns the correct thing. Should these be unit tests in the respected files rather than an integration test all in one?

This file should be marked for special review once it lands, the same way we protect manifest. Changes in this file *only* could represent a shift in the axe-core api that breaks axe Pro. For that reason, spreading these tests into their respective modules would make it harder to keep an "eye" on the api axe Pro requires to operate.

test/integration/api/external/index.js
stephenmathieson: I don't think we need this test, as the test below it requires `axe.commons.sanitize` to be a function in order to pass.

WilcoFiers:
@AutoSponge +1 on the decision to only test call signatures. We're testing those functions in depth elsewhere. I agree with the idea of adding you and/or Harris as code owner for these tests.

build/test/config.js
```diff
@@ -26,7 +26,10 @@ exports = module.exports = function(grunt, options) {
 					'http://' + host + ':<%= connect.test.options.port %>/test/commons/',
 					'http://' +
 						host +
-						':<%= connect.test.options.port %>/test/integration/rules/'
+						':<%= connect.test.options.port %>/test/integration/rules/',
+					'http://' +
+						host +
+						':<%= connect.test.options.port %>/test/integration/api/external/'
```
jkodu: I like the new directory `/api`, what is the thinking behind `/external`?

test/integration/api/external/index.html
```diff
@@ -0,0 +1,28 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD HTML+ARIA 1.0//EN" "http://www.w3.org/WAI/ARIA/schemata/html4-aria-1.dtd">
```
jkodu: enlighten me as to why such verbosity in doctype is essential for these tests?

AutoSponge: This was copied from other axe-core test harnesses. It's not essential to this test.

build/test/config.js
AutoSponge: In the future, someone may want to document the API of a submodule. It also makes it clear this only refers to the API surface area with external dependents.

stephenmathieson:
Is there anything that needs to happen based on https://github.com/dequelabs/axe-core/pull/1836#issuecomment-538650345?

AutoSponge:
@stephenmathieson I assume a code owner of axe-core will need to update the reviewer settings when this file lands.

stephenmathieson:
I think this can land without a "codeowners" update and when @WilcoFiers's back, he can decide what to do moving forward.

Issue #315, opened 2017-04-26
Ensure non-printable characters aren't reported for color-contrast


marcysutton:
Our color contrast rule is reporting incomplete for non-printable characters, such as:

```html
<span>&rlm;</span>
``` 

As a result, it's really difficult to debug them in the new Needs Review UI (to be released) because they have no visible text content. The related node ends up as:

```html
<span></span>
```

We should exclude these elements in `color-contrast-matches.js`, along with any other non-printable characters. Do some research to build up a list of candidates in addition to `&rlm;`.

A live example can be found in the Notifications page of Twitter.com.

WilcoFiers:
This might be a good starting point: http://www.fileformat.info/info/unicode/category/index.htm

straker:
So I dug into this. JavaScript doesn't really understand unicode categories, so we couldn't use a Regex to just test for certain categories. There's a proposal to implement [category aware regex syntax](https://github.com/tc39/proposal-regexp-unicode-property-escapes). It has made it to stage 4 and will be implemented in ES2018, but that wouldn't help us with IE11. 

It does list 2 other libraries that fill in support for unicode: [XRegExp](https://github.com/slevithan/xregexp) and [Regenerate](https://github.com/mathiasbynens/regenerate). Regenerate works with IE6+, while XRegExp works with all ES5+ browsers and node. Both have their pros and cons which are mentioned in the proposal.

There's also the problem that what is "printable" in unicode is not always obvious. From a [stack overflow answer](https://stackoverflow.com/a/53450014/2124254):

> The more you learn about Unicode, the more you realise how unexpectedly diverse and unfathomably weird human writing systems are.
>
> In particular whether a particular "character" is printable is not always obvious.
> 
> Is a zero-width space printable? When is a hyphenation point printable? Are there characters whose printability depends on their position in a word or on what characters are adjacent to them? Is a combining-character always printable?

Another problem would be an element with just an emoji in it. An emojis color isn't determined by the style of the element, but axe still reports it as a color contrast error.

```html
<div style=" background: black; color: #3e3e3e; ">😋</div>
```

If we also want to exclude emoji from the color checker, we might run into a Sisyphus problem with a blacklist approach as the emoji library is constantly growing. 

[...]

PR #1837, opened 2019-10-02
fix(color-contrast): ignore nodes that don't contain text

straker:
Ignore nodes of only emoji, nonBMP unicode, and punctuation. Turns out this was a lot easier than I thought since I discovered we had the unicode functions. 

Closes issue: #315

## Reviewer checks


straker:
`text.sanitize` should remove any whitespace characters from the text

PR #1838, opened 2019-10-05
perf: Improve performance of various functions

sorin-davidoi:
Optimize various functions. Most of the optimizations fall in one of the following categories:
 - Switch the order of conditions in `if` statements to order to execute the cheapest operation first
 - Short-circuit searches in arrays
 - Avoid allocating intermediary arrays

`axe https://www.theguardian.com/international` shows an improvement of ~200ms.

## Reviewer checks

### DIFF
--- a/lib/core/utils/get-scroll.js
+++ b/lib/core/utils/get-scroll.js
@@ -19,29 +19,17 @@ axe.utils.getScroll = function getScroll(elm, buffer = 0) {
 	}
 
 	const style = window.getComputedStyle(elm);
-
-	// Checking Y before X since we assume that
-	// most scrolling is across the Y axis.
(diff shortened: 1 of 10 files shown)


lib/core/utils/get-scroll.js
```diff
@@ -19,17 +19,29 @@ axe.utils.getScroll = function getScroll(elm, buffer = 0) {
 	}
 
 	const style = window.getComputedStyle(elm);
-	const overflowXStyle = style.getPropertyValue('overflow-x');
+
+	// Checking Y before X since we assume that
+	// most scrolling is across the Y axis.
```
straker: These are just property lookups so I don't think splitting out the if statement with two different returns that both return the same thing is worth the millisecond savings.

PR #1842, opened 2019-10-09
feat(get-element-stack): performant api to replace document.elementsFromPoint

straker:
Just making a pr with the new api for now since it's quite big in terms of code size. Once this is merged I'll make another pr to use it.

The code essentially does the following:

1. It creates a 2d grid of the page and adds every element to every cell of the grid it belongs to. If the element creates a scroll region, then a new grid is created for that element and all its children
2. As it adds elements to the grid, it calculates the stack order of the node using an array of numbers. Each entry in the array is a new stacking context and the number represents the z-index of the element at that level.
3. When the stack is requested for a node, it starts at the grid the node belongs to and calculates the row/col of the center point of the element. It then gets all nodes that are part of that row/col and performs a collision detection against the element and the point. If the grid is a sub-grid, it then goes to the parent grid and gets the stack at that level
4. Finally it sorts all the returned nodes based on their visual order

I also added cache functions to virtual node as we do these lookups a ton throughout the codebase.

In terms of performance. Running this new function on every node on https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template compared to `document.elementsFromPoint` and our `color.getRectStack` function (which is what uses `document.elmentsFromPoint` under the hood for color-contrast checks), the average of 5 runs are:

* `color.getRectStack`: 143.47900002030656ms
* `dom.getElementStack`: 53.457000001799314ms

Running it on https://web.archive.org/web/20190613132353/https://giveawaylisting.com/ (have to use web archive result now that the site is going down) results in:

* `color.getRectStack`: 143147.4000000162ms
* `dom.getElementStack`: 2190.5150000122376ms

_NOTE: Jason's suggestion saved ~20% time (500ms on giveawaylistings.com)_

## Reviewer checks


lib/commons/dom/get-element-stack.js
```diff
@@ -0,0 +1,456 @@
...
+	// html node is always sorted to the end of the stack
+	// TODO: how do we test this?
+	if (a.actualNode.nodeName.toLowerCase() === 'html') {
+		return 1;
+	}
+	if (b.actualNode.nodeName.toLowerCase() === 'html') {
+		return -1;
+	}
+
+	/*eslint no-bitwise: 0 */
```
stephenmathieson: Please move this to the relevant LOC.

lib/commons/dom/get-element-stack.js
```diff
@@ -0,0 +1,456 @@
...
+			return -1;
+		}
+
+		if (a._stackingOrder[i] === b._stackingOrder[i]) {
+			continue;
+		}
+
+		if (b._stackingOrder[i] > a._stackingOrder[i]) {
+			return 1;
+		} else {
```
stephenmathieson: Remove `else`.

lib/commons/dom/get-element-stack.js
```diff
@@ -0,0 +1,488 @@
...
+		node.nodeName === '#document-fragment'
+	) {
+		return true;
+	}
+
+	const computedStyle = vNode.computedStyle;
+
+	// position: fixed or sticky
+	if (
+		computedStyle.getPropertyValue('position') === 'fixed' ||
```
scurker: Lookups for the same property value should probably be cached. See https://jsperf.com/get-property-value

lib/commons/dom/get-element-stack.js
```diff
@@ -0,0 +1,488 @@
...
+		computedStyle.getPropertyValue('transform') ||
+		'none';
+
+	if (transform !== 'none') {
+		return true;
+	}
+
+	// elements with a mix-blend-mode value other than "normal"
+	if (
+		computedStyle.getPropertyValue('mix-blend-mode') &&
```
scurker: See https://github.com/dequelabs/axe-core/pull/1842#discussion_r335615350

lib/commons/dom/get-element-stack.js
```diff
@@ -0,0 +1,488 @@
...
+		computedStyle.getPropertyValue('mix-blend-mode') !== 'normal'
+	) {
+		return true;
+	}
+
+	// elements with a filter value other than "none"
+	if (
+		computedStyle.getPropertyValue('filter') &&
```
scurker: See https://github.com/dequelabs/axe-core/pull/1842#discussion_r335615350

lib/commons/dom/get-element-stack.js
```diff
@@ -0,0 +1,488 @@
...
+		computedStyle.getPropertyValue('filter') !== 'none'
+	) {
+		return true;
+	}
+
+	// elements with a perspective value other than "none"
+	if (
+		computedStyle.getPropertyValue('perspective') &&
```
scurker: See https://github.com/dequelabs/axe-core/pull/1842#discussion_r335615350

lib/commons/dom/get-element-stack.js
```diff
@@ -0,0 +1,488 @@
...
+	// elements with -webkit-overflow-scrolling set to "touch"
+	if (computedStyle.webkitOverflowScrolling === 'touch') {
+		return true;
+	}
+
+	// a flex item with a z-index value other than "auto", that is the parent element display: flex|inline-flex,
+	if (computedStyle.getPropertyValue('z-index') !== 'auto' && vNode.parent) {
+		const parentStyle = vNode.parent.computedStyle;
+		if (
+			parentStyle.getPropertyValue('display') === 'flex' ||
```
scurker: See https://github.com/dequelabs/axe-core/pull/1842#discussion_r335615350

lib/commons/dom/get-element-stack.js
```diff
@@ -0,0 +1,488 @@
...
+/**
+ * Determine if a node is a scroll region.
+ * @param {VirtualNode}
+ * @return {Boolean}
+ */
+function isScrollRegion(vNode) {
+	const style = vNode.computedStyle;
+
+	return (
+		style.getPropertyValue('overflow-x') === 'auto' ||
```
scurker: See https://github.com/dequelabs/axe-core/pull/1842#discussion_r335615350

lib/commons/dom/get-element-stack.js
```diff
@@ -0,0 +1,488 @@
...
+	const rootGrid = {
+		container: null,
+		cells: []
+	};
+
+	axe.utils
+		.querySelectorAll(axe._tree[0], '*')
+		.filter(vNode => vNode.actualNode.parentElement !== document.head)
+		.forEach(vNode => {
+			if (vNode.actualNode.nodeType !== 1) {
```
scurker: Minor nitpick, is `Node.ELEMENT_NODE` not an option here?

straker: It probably is, but we haven't used it before. If we did, we'd need to use it under `window` so we don't relay on `Node` as a global.

scurker: Having magic numbers do make it less approachable, especially if I don't know what `nodeType` is coming from. Since this is sticking with existing conventions I'm not going to suggest a one-off change here, but it would be nice to eventually either alias node types to have more readable code.

AutoSponge: Node.ELEMENT_NODE should exist everywhere. If not, there are polyfills (like https://github.com/inexorabletash/polyfill/blob/HEAD/dom.js)

lib/commons/dom/get-element-stack.js
```diff
@@ -0,0 +1,489 @@
...
+
+			const transform =
+				vNode.getComputedStylePropertyValue('-webkit-transform') ||
+				vNode.getComputedStylePropertyValue('-ms-transform') ||
+				vNode.getComputedStylePropertyValue('transform') ||
+				'none';
+
+			// if rect is rotated, rotate point around center of rect and then
+			// perform collision detection as though the rect was not rotated
+			if (
```
AutoSponge: Similarly, if rect is clipped, treat it as not clipped. It could "stop" the stack or allow the stack to continue depending on whether it's a masking or keyhole technique. We shouldn't try to handle this case, just note that it's an area we can't address without a canvas to plot on.

lib/commons/dom/get-element-stack.js
```diff
@@ -0,0 +1,489 @@
+/* global dom */
+
+// split the page cells to group elements by the position
+const gridSize = 200; // arbitrary size, increase to reduce memory (less cells) use but increase time (more nodes per grid to check collision)
+
+// https://developer.mozilla.org/en-US/docs/Web/CSS/Replaced_element
+const replacedElements = ['iframe', 'video', 'embed', 'img'];
```
AutoSponge: Should an image in the `:before` or `:after` of an element create a separate, replaced layer? Or is it considered the same as the host? What if the bounding rect is different?

lib/commons/dom/get-element-stack.js
```diff
@@ -0,0 +1,489 @@
...
+			// if rect is rotated, rotate point around center of rect and then
+			// perform collision detection as though the rect was not rotated
+			if (
+				transform.indexOf('matrix') === 0 &&
+				// only non-replaced inline elements can be rotated
+				// see https://developer.mozilla.org/en-US/docs/Web/CSS/transform
+				(replacedElements.includes(nodeName) ||
+					(vNode.getComputedStylePropertyValue('display') !== 'inline' &&
+						vNode.getComputedStylePropertyValue('display') !== 'contents'))
+			) {
```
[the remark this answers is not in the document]
straker: We can do this at another time. The only problem is that axe my not consider transforms, but the browsers `document.elementsFromPoint` does, so when we move this over to color contrast we may see some regression.

lib/commons/dom/get-element-stack.js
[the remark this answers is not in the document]
dylanb: `input[type="image"]` also needs to be on the list
Also, these elements can be replaced under certain circustances

```<option>
<audio>
<canvas>
<object>
<applet>```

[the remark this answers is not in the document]
Comment (dylanb) [reply lib/commons/dom/get-element-stack.js]: See https://developer.mozilla.org/en-US/docs/Web/CSS/Replaced_element

File: lib/commons/dom/get-element-stack.js [hunk 13]
```diff
@@ -0,0 +1,489 @@
...
+
+	// a flex item with a z-index value other than "auto", that is the parent element display: flex|inline-flex,
+	if (
+		vNode.getComputedStylePropertyValue('z-index') !== 'auto' &&
+		vNode.parent
+	) {
+		const parent = vNode.parent;
+		if (
+			parent.getComputedStylePropertyValue('display') === 'flex' ||
+			parent.getComputedStylePropertyValue('display') === 'inline-flex'
```
scurker: I just noticed the latest [CSS3 display spec](https://www.w3.org/TR/css-display-3/#the-display-properties) allows `display: <display-outside> <display-inside>` so `display: inline-flex` is equivalent to `display: inline-flex`. Do we also want to account for a future inside value of `flex`?

straker:
We decided to hold off on supporting CSS rotation for now and add it as a separate PR / function.

lib/commons/dom/get-element-stack.js
straker: Probably not for now, but we'll have to keep an eye on it as browsers start adopting it.

test/commons/dom/get-element-stack.js
```diff
@@ -0,0 +1,353 @@
...
+				return vNode.actualNode.id;
+			});
+			assert.deepEqual(stack, ['target', '3', '2', '1', 'fixture']);
+		});
+
+		it('should not return elements that do not fully cover the target', function() {
+			var vNode = queryFixture(
+				'<div id="1" style="position:relative;">' +
+					'<div id="2" style="position:absolute;width:300px;height:20px;"></div>' +
+					'<p id="target" style="position:relative;z-index:1;height:40px;">Text oh heyyyy <a href="#" id="target">and here\'s <br>a link</a></p>' +
```
WilcoFiers: You should probably give this a width too, to ensure that this test won't fail on <600px viewport width.

lib/commons/dom/get-element-stack.js
```diff
@@ -0,0 +1,432 @@
...
+
+/**
+ * Determine if node produces a stacking context.
+ * References:
+ * https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context
+ * https://github.com/gwwar/z-context/blob/master/devtools/index.js
+ * @param {VirtualNode} vNode
+ * @return {Boolean}
+ */
+function isStackingContext(vNode) {
```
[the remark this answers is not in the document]
straker: You're right. Added what was missing.

lib/commons/dom/get-element-stack.js
straker: Decided to support this since I was already updating the code

lib/commons/dom/get-element-stack.js
```diff
@@ -0,0 +1,432 @@
...
+
+	return stackingOrder;
+}
+
+/**
+ * Determine if a node is a scroll region.
+ * @param {VirtualNode}
+ * @return {Boolean}
+ */
+function isScrollRegion(vNode) {
```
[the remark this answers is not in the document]
straker: I didn't know about it. Switched to use it.

Issue #1717, opened 2019-07-22
Audit never finishes on page with large SVG


mccahan:
Expectation: Expected axe to finish the audit and show results when running against https://futureofhousing.aarp.org/, which contains a large SVG in-line at the top. If it's unable to complete an audit due to an issue, I would expect it to time out and give up after a reasonable amount of time, or to give some kind of warning or error message indicating where it's hung up.

The exact same page but without the large city SVG at the top (https://futureofhousing.aarp.org/?skipCity=true) completes the audit without issue.

Actual: The pane sits on "Analyzing... please wait." but never appears to finish or time out or give any error messages or other indications.

Motivation: Would expect the product to time out after a reasonable amount of time or give any kind of error indication to help troubleshoot the issue. As-is, I had to trial-and-error remove elements from the page until I found the SVG that was causing the issue.

<pre><code>
axe-core version: 3.2.2
axe-webdriver, extension or other integration version: 3.8.0 (Chrome Extension, Chrome 75.0.3770.100, Windows 10)
</code></pre>

straker:
Thanks for reporting this. I'm able to reproduce and it seems to be stuck on the rule `duplicate-id-active`. My guess is because 3550 nodes in the SVG use the `id` of `Path`.

sorin-davidoi:
I think this is solvable by caching the result of `isHiddenWithCSS` for the traversed nodes. From what I understand the code looks something like

```javascript
for(const node of document.querySelectorAll('[id]')) {
  const (const otherWithSameId of document.querySelectorAll(`*[id="${node.getAttribute('id')}"]`)) {
    // The following is inside isHiddenWithCSS
    const parent = otherWithSameId;
    while(parent) {
       // Some expensive computations involving getComputedStyle and getPropertyValue
      parent = parent.parentElement;
    }
  }
}
```

The problem is that `isHiddenWithCSS` can be called hundreds of time for the same element, but since we know that there have been no DOM mutations that could be avoided. The following patch makes `axe https://futureofhousing.aarp.org` finish in 1 minute 20 seconds, but I'm not sure if it's applied at the right level (or if `WeakMap`/`Map` can be used). @straker What is your take on this?

```diff
index 7c0d2f15..efe6ae9b 100644
--- a/lib/commons/dom/is-hidden-with-css.js
+++ b/lib/commons/dom/is-hidden-with-css.js
@@ -10,6 +10,24 @@
  * @return {Boolean} the element's hidden status
  */
[...]

PR #1846, opened 2019-10-14
perf(is-hidden-with-css): Cache result

sorin-davidoi:
Caches the computation result.

Closes issue:https://github.com/dequelabs/axe-core/issues/1717

## Reviewer checks

### DIFF
--- a/lib/commons/dom/is-hidden-with-css.js
+++ b/lib/commons/dom/is-hidden-with-css.js
@@ -10,11 +10,13 @@
  * @return {Boolean} the element's hidden status
  */
 dom.isHiddenWithCSS = function isHiddenWithCSS(el, descendentVisibilityValue) {
-	if (el._isHiddenWithCSS === void 0) {
-		el._isHiddenWithCSS = _isHiddenWithCSS(el, descendentVisibilityValue);
+	const vNode = axe.utils.getNodeFromTree(el);
+
+	if (vNode._isHiddenWithCSS === void 0) {
+		vNode._isHiddenWithCSS = _isHiddenWithCSS(el, descendentVisibilityValue);
 	}
 
-	return el._isHiddenWithCSS;
+	return vNode._isHiddenWithCSS;
 };
 
 function _isHiddenWithCSS(el, descendentVisibilityValue) {
--- a/lib/commons/dom/is-hidden-with-css.js
+++ b/lib/commons/dom/is-hidden-with-css.js
@@ -10,6 +10,14 @@
  * @return {Boolean} the element's hidden status
  */
 dom.isHiddenWithCSS = function isHiddenWithCSS(el, descendentVisibilityValue) {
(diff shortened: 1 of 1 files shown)


lib/commons/dom/is-hidden-with-css.js
```diff
@@ -10,6 +10,14 @@
  * @return {Boolean} the element's hidden status
  */
 dom.isHiddenWithCSS = function isHiddenWithCSS(el, descendentVisibilityValue) {
+	if (el._isHiddenWithCSS === void 0) {
```
straker: So `el` is an HTMLElement, which we try not to add properties to as it can be unreliable. Instead, you should get the VirtualNode from the element and put the cache on that.

```js
const vNode = axe.utils.getNodeFromTree(el);

if (vNode._isHiddenWithCss === void 0) {

}
```

sorin-davidoi:
> The current changes will only cache the result for the current node.

This patch does the same, albeit in a different way (`dom.isHiddenWithCSS` and `_isHiddenWithCSS` are calling each other).

sorin-davidoi:
Not sure why using the VirtualNode breaks the tests.

straker:
Ah yes. So the test are using actual DOM nodes (and in one of them even a fake node) to run the test. Axe isn't being setup with virtual nodes before each test, so `axe.utils.getNodeFromTree` returns `null`.

To fix, in each test you'll need to use `testUtils.fixtureSetup()` and pass it the node, which should set up the cache with vNodes.

lib/commons/dom/is-hidden-with-css.js
```diff
@@ -10,6 +10,16 @@
  * @return {Boolean} the element's hidden status
  */
 dom.isHiddenWithCSS = function isHiddenWithCSS(el, descendentVisibilityValue) {
+	const vNode = axe.utils.getNodeFromTree(el);
+
+	if (vNode._isHiddenWithCSS === void 0) {
+		vNode._isHiddenWithCSS = _isHiddenWithCSS(el, descendentVisibilityValue);
```
WilcoFiers: @straker should this not push to `vnode._cache.isHiddenWithCSS`?

straker: Only if we were to update all the calls to `isHiddenWithCSS` so they called the vNodes cache function. We did this same thing for [`utils.isVisible`](https://github.com/dequelabs/axe-core/blob/develop/lib/core/utils/is-hidden.js#L23-L25) so we could get all the performance gains without having to update every use of the function.

sorin-davidoi:
@straker Turns out that `body` doesn't have a virtual node. Maybe that's the case only in the test environment?

straker:
@sorin-davidoi Correct. We scope the virtual node tree to just the `#fixture` element so it ignores everything from Mocha.

sorin-davidoi:
@straker I've added a check for that, let me know if there's anything else that's missing here.

PR #1879, opened 2019-11-04
[WIP] chore: update to es module syntax

straker:
Modernize the codebase by using ES Module imports and exports. Really helps find where things are coming from and removes all magic globals. Also removes the need for grunt to build and concat axe as we can just run rollup or webpack and it will do it for us.

Also since we no longer concat files together, each file doesn't have to worry about name collisions with other files in the same group (so any top level `const` in a file in the commons directory won't collide with the same named variable in another file of the same directory).

Changed how the reporters worked because they didn't export and i needed to somehow bring them into the index file so they would be added to the end file. So I made them export the reporter function and had the index file call add reporter for each one.

Also split functions into their own files if multiple export functions were defined in the same file (within reason).

Lastly, I renamed all the checks/rule `.json` to  be the `.js` file of the same name, and moved the check evaluate `.js` to be named with `evaluate.js` at the end. So you'll see *a lot* of file deletions (e.g. over half the files) for the check and rule `.json` files and prior evaluate `.js`, with new files being added in their place. The contents should be the same just with imports and exports.

## Reviewer checks


lib/core/imports/index.js
```diff
@@ -1,4 +1,14 @@
-/* global axe */
+/**
+ * Namespace `axe.imports` which holds required external dependencies
+ *
+ * @namespace imports
+ * @memberof axe
+ */
+export { default as axios } from 'axios';
```
stephenmathieson: This is my favorite part of this PR. Being able to `import` 3rd party tools will hugely improve our workflows. :heart:

straker: I know! Webpack is pretty awesome at that.

.huskyrc
```diff
@@ -1,5 +1,5 @@
 {
 	"hooks": {
-		"pre-commit": "grunt configure && lint-staged"
```
straker: This is temporary as I converted the code, will be reverted once eslint and grunt are setup properly

test/playground.html
```diff
@@ -4,13 +4,13 @@
 
 	<div class="foo" id="foo">foo</div>
 
-	<script src="/axe.js"></script>
```
straker: Also temporary

package.json
```diff
@@ -135,7 +136,6 @@
 			"git add"
 		],
 		"*.js": [
-			"eslint --fix",
```
straker: Also temporary

webpack.config.js
```diff
@@ -0,0 +1,11 @@
+const path = require('path');
```
straker: Please feel free to suggest / write a better Webpack config, I'm still new to it.

straker:
Closing as well out of date now. Will start again for 4.0

PR #1882, opened 2019-11-06
fix: remove heading from list of widget roles

schne324:
see https://www.w3.org/WAI/PF/aria/roles#widget_roles

## Reviewer checks


straker:
LGTM but I'd like @WilcoFiers to do a 1 over just to make sure we're not missing something.

Issue #1171, opened 2018-10-02
<ul role="menubar"> not recognizing <li><a role="menuitem"></a></li> as a descendent


bbAdam:
In the version 3.5 chrome extension update from 9/25/18, a menu structure like the following: <pre>\<ul role="menuitem">\<li>\<a role="menuitem">Menu Item\</a>\</li>\</ul></pre>

Throws the following flag: <pre>\<li> elements must be contained in a \<ul> or \<ol></pre>

This is inconsistent with the specs layed out here: https://www.w3.org/TR/wai-aria-practices/examples/menubar/menubar-1/menubar-1.html

If the menuitem role is moved to the list item, the flag goes away. 

It appears as though the AXE algorithm is only looking for Direct Decedents to have the role of menuitem, instead of any descendent of the parent menubar. 

https://www.w3.org/WAI/PF/aria/roles#menuitem states that "Authors MUST ensure that menu items are owned by an element with role menu or menubar in order to identify that they are related widgets. Authors MAY separate menu items into sets by use of a separator or an element with an equivalent role from the native markup language."

and an owned element is defined as "any DOM descendant of the element, any element specified as a child via aria-owns, or any DOM descendant of the owned child." here: https://www.w3.org/WAI/PF/aria/terms#def_owned_element

Issue can be seen here: https://templatelibrary.schoolwires.net/academy

I am a web developer for Blackboard, Inc. and we have hundreds of sites out there that are now getting this flag. Thank you for looking into it!

WilcoFiers:
Thanks for reporting. I'm quite surprised that this seems to work. I've seen different cases where, unless there is a role=presentation or role=none on the element between the owner and the owned element, the screen reader can't work out how many items are in that group. Testing the example in VoiceOver gets me a correct count, so it seems something more is going on there. Have you tried testing this in NVDA or JAWS? Do they give you a correct count of menu items?

bbAdam:
@WilcoFiers I don't have access to NVDA or JAWS, so I'm not sure... but I have seen what you mentioned in VoiceOver.

WilcoFiers:
Update: I've done a bunch of research. There is quite a bit going on that I wasn't aware of. I'm working to reverse engineer the algorithms used to come up with this list of owned elements, so that axe can correctly identify what elements belong where. It'll take some time to get this done, but it's high on my priorities list.

WilcoFiers:
[...]

PR #1894, opened 2019-11-12
fix(listitem): clarify that li elements must be contained in a list or role=list

straker:
There was some confusion about lists having overriding roles and `li` children.

Closes issue: #1171

## Reviewer checks

### DIFF
--- a/lib/rules/listitem.json
+++ b/lib/rules/listitem.json
@@ -5,7 +5,7 @@
 	"tags": ["cat.structure", "wcag2a", "wcag131"],
 	"metadata": {
 		"description": "Ensures <li> elements are used semantically",
-		"help": "<li> elements must be contained in a <ul> or <ol>"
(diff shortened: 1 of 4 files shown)


lib/rules/listitem.json
```diff
@@ -5,7 +5,7 @@
 	"tags": ["cat.structure", "wcag2a", "wcag131"],
 	"metadata": {
 		"description": "Ensures <li> elements are used semantically",
-		"help": "<li> elements must be contained in a <ul> or <ol>"
+		"help": "<li> elements must be contained in a <ul> or <ol> without a role or an element with role=list."
```
WilcoFiers: Can you swap around the description and help text. The help text needs to be shorter than the description. Also, I think this needs to be done in the remediation message of the check. Preferably, this is put in conditionally. If it's not in a UL, it just has the message it has now, if it is in a UL with a role, the message add something like `without a role, or with role="list"`.

Issue #1033, opened 2018-07-30
Assess feasibility to swap check for `aria-dpub-role-fallback` to `aria-allowed-role`


jkodu:
- see comments: https://github.com/dequelabs/axe-core/pull/945/files#r198097140
- to swap the underlying checks used by `aria-dpub-role-fallback` rule, from `implicit-role-fallback` to `aria-allowed-role rule`.
- aim to deprecate `implicit-role-fallback`

jeankaplansky:
Included in 3.5 release notes doc

padmavemulapati:
Verified with the latest axe-coconut, seeing the "aria-dpub-role-fallback" rules in deprecated tag and moved to "aria role"
![image](https://user-images.githubusercontent.com/47409579/69849754-8e06c900-12a3-11ea-8364-5e33e67795da.png)

![image](https://user-images.githubusercontent.com/47409579/69849790-abd42e00-12a3-11ea-9d90-fb2291a4a9dc.png)

PR #1899, opened 2019-11-13
feat(aria-dpub-role-fallback): depreacte aria-dpub-role-fallback and move into aria-allowed-role

straker:
Aria-allowed-role can handle dpub role fallbacks, so no longer need the aria-dpub-role-fallback rule for that. See https://github.com/dequelabs/axe-core/pull/945/files#r198097140

Closes issue: #1033

## Reviewer checks


lib/checks/aria/implicit-role-fallback.json
```diff
@@ -1,6 +1,7 @@
 {
 	"id": "implicit-role-fallback",
 	"evaluate": "implicit-role-fallback.js",
+	"deprecated": true,
```
WilcoFiers: Did you mean to put this on the other rule?

straker: Nope, implicit-role-fallback is only used by aria-dpub-role-fallback

lib/commons/aria/get-element-unallowed-roles.js
```diff
@@ -1,5 +1,15 @@
...
 
+const dpubRoles = [
+	'doc-backlink',
+	'doc-biblioentry',
+	'doc-biblioref',
+	'doc-cover',
+	'doc-endnote',
+	'doc-glossref',
+	'doc-noteref'
+];
```
rdeltour: Just to clarify: these are not all the DPUB ARIA roles, but the ones which are subclassing roles that are implicit on some native HTML elements (specifically: `img`, `link`, and `listitem`).

Maybe this can be clarified in a comment for later reference?

PR #1900, opened 2019-11-14
fix(typings): add proper return value to getRule

michael-siek:
Working on a typescript project and would like proper typings for `getRules()`

## Reviewer checks


axe.d.ts
```diff
@@ -155,6 +155,13 @@ declare namespace axe {
 		}[];
 		cleanup?(callback: Function): void;
 	}
+	interface GetRuleResult {
```
straker: I think `RuleMetadata` would be appropriate as that's what we call the JSON files that create this data.

PR #1901, opened 2019-11-14
test: disable appveyor

straker:
They broke our IE builds so we're turning off for now.

## Reviewer checks


straker:
It would seem so. I was hoping just removing the appveyor config would disable the checks but that doesn't seem to be the case. Will look into it later.

scurker:
See #1903

Issue #1228, opened 2018-11-07
Some contrast violations are not being reported


mfairchild365:
Running axe in chrome yields contrast errors on the last tab example here https://material.angular.io/components/tabs/examples - however contrast errors in other tab examples (that appear to have the same problem) are not flagged.

<pre><code>
axe-core version: 3.1.1
Chrome extension 3.6.0 on Mac
</code></pre>

straker:
The page may have been updated since this ticket was created, but the color contrast rule still passes over some nodes it shouldn't. Digging into it, it seems the 2nd to last row passes because our foreground color function checks for the opacity of the element, which returns 1, however the parent defines an opacity of 0.6. Because we do not check the parent tree for opacity, we incorrectly calculate the foreground color and return a ratio of 15.52 when in fact the true ratio is 4.26.

The solution seems to be to scan the parent tree and calculate the stacking opacities as opacity is not inherited but is determined by any parent opacities (similar to z-index).

jeankaplansky:
Added to release notes as a fix since it was reported as an stuff not being reported that should have been.

padmavemulapati:
Verified with the latest axe-coconut, seeing the more color-contrast issues(88)  than earlier(66). 
@straker , Is this the one expected. 

![image](https://user-images.githubusercontent.com/47409579/69857621-52c1c580-12b6-11ea-908b-878f517fe26d.png)

jeankaplansky:
release notes fix.

PR #1902, opened 2019-11-14
fix(color-contrast): take into account parent opacity for foreground color

straker:
Our foreground color check did not take into account the parent opacity, so would incorrectly calculate the final color. Added a function that would gather the parent opacity level and cache it for quicker lookups.

Also did a quick performance optimization by passing the already calculated bgColor from the parent into the foreground color check so we didn't have to do that expensive operation again.

Closes issue: #1228

## Reviewer checks


lib/commons/color/get-foreground-color.js
```diff
@@ -1,5 +1,30 @@
...
+
+	const vNode = axe.utils.getNodeFromTree(node);
+
+	if (vNode && vNode._opacity !== undefined && vNode._opacity !== null) {
+		return vNode._opacity;
+	}
+
+	const nodeStyle = window.getComputedStyle(node);
+	const opacity = nodeStyle.getPropertyValue('opacity');
+	const finalOpacity = opacity * getOpacity(node.parentElement);
```
WilcoFiers: Add a test to check the recursive opacity lookup works.

Issue #1834, opened 2019-10-01
tabindex rule throws error when it shouldn't


corrieann:
Seems to be a recurrence of https://github.com/dequelabs/axe-core/issues/1714

Snippet of HTML of the Element | &lt;paper-input class="colab-replace-input" type="text" label="Replace with" tabindex="0" aria-disabled="false"&gt;
-- | --
Impact | serious
Targets | .colab-replace-row > paper-input.colab-replace-input
FailureSummary | Fix any of the following: Element has a tabindex greater than 0
At least one check must have passed | CheckResult ID: tabindexImpact: seriousMessage: Element has a tabindex greater than 0

Can view the element in question on 
https://colab.research.google.com by opening Edit -> Find and replace...
although if I do that and run axe via a Chrome extension, it's not triggering this error.  It only seems to from axe-webdriver.

<pre><code>
axe-core version: 2.5
axe-webdriver, extension or other integration version: 2.0

For Tooling issues:
- Platform:  Linux
</code></pre>

straker:
The bug is that `paper-input` redefines the `tabIndex` property of the element, so when we check for `node.tabIndex <= 0`, it returns `undefined` instead of a value. I'm not sure if this is a Polymer bug or something we have to start accounting for (custom elements overriding native properties of the HTMLElement).

WilcoFiers:
Conclusion. This is a bug in paper-input, which does not allow accessing .tabIndex. That said, to avoid this type of issue, I do think we should drop using properties to get at attribute values. Using `getAttribute()` is more reliable, we've been burned by this before, so let's just solve it.

straker:
Fix this problem, then work it in over time for the rest of the code base.

padmavemulapati:
Not seeing the now, as we have done fix. Closing the ticket as issue not replicating

PR #1910, opened 2019-11-19
fix(tabindex): don't error when tabindex property is overridden

straker:
Closes issue: #1834

## Reviewer checks

### DIFF
--- a/lib/checks/keyboard/tabindex.js
+++ b/lib/checks/keyboard/tabindex.js
@@ -1,4 +1,4 @@
-const tabIndex = parseInt(node.getAttribute('tabindex'));
+const tabIndex = parseInt(node.getAttribute('tabindex'), 10);
 
 // an invalid tabindex will either return 0 or -1 (based on the element) so
 // will never be above 0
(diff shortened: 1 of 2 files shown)


lib/checks/keyboard/tabindex.js
```diff
@@ -1 +1 @@
-return node.tabIndex <= 0;
+return node.getAttribute('tabindex') <= 0;
```
jkodu: given we are not reading [tabIndex](https://developer.mozilla.org/en-US/docs/Web/API/HTMLOrForeignElement/tabIndex) which would return an integer, and instead an attribute it would make sense to `parseInt` this & handle `NaN` given someone can set `tabindex='something'` that is not a number

test/checks/keyboard/tabindex.js
```diff
@@ -22,4 +22,13 @@ describe('tabindex', function() {
 
 		assert.isTrue(checks.tabindex.evaluate(node));
 	});
+
+	it('should look at the attribute and not the property', function() {
```
jkodu: should add test for `NaN` cases

straker:
I'll open a new pr for that one since it's out of scope for what this pr is fixing.

PR #1922, opened 2019-11-28
feat(aria-label): deprecate Element arg; use virtualNode

WilcoFiers:
One of many upcoming PRs that will start to transition axe-core away from directly working on the DOM.

## Reviewer checks

### DIFF
--- a/lib/commons/aria/arialabel-text.js
+++ b/lib/commons/aria/arialabel-text.js
@@ -3,13 +3,16 @@
 /**
  * Get the text value of aria-label, if any
  *
+ * @deprecated Do not use Elemnet directly. Pass VirtualNode instead
  * @param {VirtualNode|Element} element
  * @return {string} ARIA label
  */
 aria.arialabelText = function arialabelText(node) {
-	node = node.actualNode || node;
-	if (node.nodeType !== 1) {
-		return '';
+	if (node instanceof axe.AbstractVirtualNode === false) {
+		if (node.nodeType !== 1) {
+			return '';
+		}
+		node = axe.utils.getNodeFromTree(node);
 	}
-	return node.getAttribute('aria-label') || '';
+	return node.attr('aria-label') || '';
 };
--- a/lib/commons/aria/arialabel-text.js
+++ b/lib/commons/aria/arialabel-text.js
@@ -3,7 +3,7 @@
 /**
  * Get the text value of aria-label, if any
  *
- * @deprecated Do not use Elemnet directly. Pass VirtualNode instead
+ * @deprecated Do not use Element directly. Pass VirtualNode instead
  * @param {VirtualNode|Element} element
  * @return {string} ARIA label
  */
--- a/lib/core/base/virtual-node/serial-virtual-node.js
+++ b/lib/core/base/virtual-node/serial-virtual-node.js
@@ -84,3 +84,5 @@ function normaliseAttrs({ attributes = {} }) {
 		return attrs;
 	}, {});
 }
+
+axe.SerialVirtualNode = SerialVirtualNode;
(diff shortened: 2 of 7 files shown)


lib/commons/aria/arialabel-text.js
```diff
@@ -3,13 +3,16 @@
 /**
  * Get the text value of aria-label, if any
  *
+ * @deprecated Do not use Elemnet directly. Pass VirtualNode instead
```
straker: Typo in `Elemnet`

lib/core/base/virtual-node/serial-virtual-node.js
```diff
@@ -84,3 +84,5 @@ function normaliseAttrs({ attributes = {} }) {
 		return attrs;
 	}, {});
 }
+
+axe.SerialVirtualNode = SerialVirtualNode;
```
straker: We shouldn't need to add this to the axe namespace. All our files are loaded directly into the test, so `SerialVirtualNode` should be available as a global, just like our `Audit` is in `test/core/base/audit.js`

lib/core/base/virtual-node/virtual-node.js
```diff
@@ -124,3 +124,5 @@ class VirtualNode extends axe.AbstractVirtualNode {
 		return this._cache.boundingClientRect;
 	}
 }
+
+axe.VirtualNode = VirtualNode;
```
straker: Ditto for not adding this to the axe namespace.

test/commons/aria/arialabel-text.js
```diff
@@ -3,32 +3,44 @@ describe('aria.arialabelText', function() {
 	var aria = axe.commons.aria;
 
 	it('returns "" if there is no aria-label', function() {
-		var node = document.createElement('div');
-		assert.equal(aria.arialabelText(node), '');
+		var vNode = new axe.SerialVirtualNode({ nodeName: 'div' });
```
straker: I don't think we need to specifically test `SerialVirtualNode` and instead just stick to `VirtualNode`. We already have tests for `SerialVirtualNode` that ensures it gives the correct results back to it's API, so we should just stick to testing the `arialabel-text` function and not worry about different VirtualNode types.

lib/core/base/virtual-node/serial-virtual-node.js
WilcoFiers: That's not correct. `core` and `commons` are in an isolated scope. Only files in `test/core` have access to the scope of `src/core`. Files in `test/commons`, `tests/checks` and `tests/integrations` don't have access. Same with `src/commons`. Only files in `tests/commons` have access to the `src/commons` scope.

Axe-core isn't just one big global scope.

test/commons/aria/arialabel-text.js
WilcoFiers: VirtualNode already works. The purpose of this change is to not just ensure VirtualNode works, but that anything extending from AbstractVirtualNode works (such as Serial). VirtualNode has access to a DOM node, tests need to show it can work entirely without accessing the DOM. Passing Serial does that.

lib/core/base/virtual-node/serial-virtual-node.js
straker: Ah I didn't know that. Other than for the test, would there be a reason to have 'VirtualNode' and 'SerialVirtualNode' on the public API?

test/commons/aria/arialabel-text.js
straker: But that would mean we duplicate every test for common functions that go though this change to ensure it works with both types. That seems like overkill and makes things much harder to maintain when we can just write a much more robust test suite for the types in one place. As long as the APIs return what we expect, we shouldn't need to write a test per type.

PR #1923, opened 2019-11-28
chore(has-alt): use virtualNode instead of node

WilcoFiers:
Stop using DOM node in has-alt check.

## Reviewer checks

### DIFF
--- a/lib/checks/shared/has-alt.js
+++ b/lib/checks/shared/has-alt.js
@@ -1,6 +1,6 @@
 const { nodeName } = virtualNode.props;
-if (['img', 'input', 'area'].includes(nodeName) === false) {
+if (!['img', 'input', 'area'].includes(nodeName)) {
 	return false;
 }
 
-return typeof virtualNode.attr('alt') === 'string';
+return virtualNode.hasAttr('alt');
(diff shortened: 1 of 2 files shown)


lib/checks/shared/has-alt.js
```diff
@@ -1,4 +1,6 @@
-var nn = node.nodeName.toLowerCase();
-return (
-	node.hasAttribute('alt') && (nn === 'img' || nn === 'input' || nn === 'area')
-);
+const { nodeName } = virtualNode.props;
+if (['img', 'input', 'area'].includes(nodeName) === false) {
+	return false;
+}
+
+return typeof virtualNode.attr('alt') === 'string';
```
straker: VirtualNode has a `hasAttr` function, any reason why you're not using it?

lib/checks/shared/has-alt.js
```diff
@@ -1,4 +1,6 @@
-var nn = node.nodeName.toLowerCase();
-return (
-	node.hasAttribute('alt') && (nn === 'img' || nn === 'input' || nn === 'area')
-);
+const { nodeName } = virtualNode.props;
+if (['img', 'input', 'area'].includes(nodeName) === false) {
```
straker: Why not `!['img', 'input', 'area'].includes(nodeName)`?

WilcoFiers: Easier to spot the negation this way. Doesn't really matter I guess.

lib/checks/shared/has-alt.js
WilcoFiers: Right you are!

PR #1932, opened 2019-12-06
chore: remove CircleCI button from readme

WilcoFiers:
There's no good reason for this button to be here. Anyone who needs to see this knows where to find it, and having the occasional "FAIL" sit at the top of our readme isn't a good look.

## Reviewer checks


stephenmathieson:
We should fix our build rather than hide the fact that it's broken.

I find the CI link very helpful.

straker:
I'm indifferent on this one. I agree that we should fix the build though.

Issue #1778, opened 2019-08-20
Rule "css-orientation-lock" should consider matrix transforms


jkodu:
Having gone through how the [transform rendering](https://drafts.csswg.org/css-transforms/#transform-rendering) specification works, the rule needs to be updated to consider the outcome values of [current transformation matrix](https://drafts.csswg.org/css-transforms/#current-transformation-matrix) in order to assess if an orientation lock has been applied.

PR #1937, opened 2019-12-11
fix: compute orientation lock  from various transformation functions

jkodu:
This PR tackles computing of orientation lock when transform is applied using any of the below transformation properties, that could affect rotation along the z-axis
- rotate
- rotateZ
- rotate3d
- matrix
- matrix3d

Closes issue:
- https://github.com/dequelabs/axe-core/issues/1778

> Note: ACT Rules has authored a rule for the same - https://act-rules.github.io/rules/b33eff

## Reviewer checks


lib/checks/mobile/css-orientation-lock.js
```diff
@@ -1,134 +1,251 @@
...
+
+			// if locked and not root HTML, preserve as relatedNodes
+			if (locked && cssRule.selectorText.toUpperCase() !== 'HTML') {
+				const elms =
+					Array.from(root.querySelectorAll(cssRule.selectorText)) || [];
+				relatedElements = relatedElements.concat(elms);
+			}
+
+			// set locked boolean
+			isLocked = locked;
```
WilcoFiers: You are repeatedly setting this. If there are two orientation media queries and the first one has the lock, the second one is going to set this to false again. 

```suggestion
			isLocked = isLocked || locked;
```

lib/checks/mobile/css-orientation-lock.js
```diff
@@ -1,134 +1,251 @@
...
-			// ensure selectorText exists
-			if (!cssRule.selectorText) {
-				return;
-			}
-			// ensure the given selector has styles declared (non empty selector)
-			if (cssRule.style.length <= 0) {
-				return;
-			}
+	const locked = degrees % 90 === 0 && degrees % 180 !== 0;
+	return !!locked;
```
WilcoFiers: This is already a boolean. You can just return  the previous line.

lib/checks/mobile/css-orientation-lock.js
```diff
@@ -1,134 +1,251 @@
...
-		Array.from(r.cssRules).forEach(cssRule => {
-			// ensure selectorText exists
-			if (!cssRule.selectorText) {
-				return;
-			}
-			// ensure the given selector has styles declared (non empty selector)
-			if (cssRule.style.length <= 0) {
-				return;
-			}
+	const locked = degrees % 90 === 0 && degrees % 180 !== 0;
```
WilcoFiers: We should allow for some margin of error. Especially when converting rad, we could be dealing with rounding errors. I think this needs to be an option in the check, how much of an error margin is on this rotation.

lib/checks/mobile/css-orientation-lock.js
```diff
@@ -1,134 +1,251 @@
...
-			const rotate = transformStyleValue.match(/rotate\(([^)]+)deg\)/);
-			const deg = parseInt((rotate && rotate[1]) || 0);
-			const locked = deg % 90 === 0 && deg % 180 !== 0;
+/**
+ * Get angle in degrees from a transform value by interpolating the unit of measure
+ * @param {String} angleWithUnit value applied to a transform function (Eg: 1turn)
+ * @returns{Number|undefined}
+ */
+function getAngleInDegrees(angleWithUnit) {
+	const [unit] = angleWithUnit.match(/(deg|grad|rad|turn)/);
```
WilcoFiers: This may throw if the regex doesn't match anything. Can't destruct on null.

lib/checks/mobile/css-orientation-lock.js
```diff
@@ -130,8 +123,14 @@ function getIsOrientationLocked({ selectorText, style }) {
 		return false;
 	}
 
-	const locked = degrees % 90 === 0 && degrees % 180 !== 0;
-	return !!locked;
+	/**
+	 * When degree is a multiple of 180, it is not considered an orientation lock
+	 */
+	if (Math.abs(degrees) % 180 <= degreeThreshold) {
```
WilcoFiers: The threshold only works if the number is slightly higher than 180, not if it ends up slightly below it.

```js
181 % 180 <= 1 // true
179 % 180 <= 1 // false
```

jkodu:
@WilcoFiers - I had to merge from develop to get one of the status checks to run, so will need your approval again

Issue #1588, opened 2019-05-25
Elements with display:table-row not included in background color test


WilcoFiers:
**Expected**: The following should pass the color-contrast rule

```html
<div style="background-color: #003; color: #eee; display: table-row;">
  <p>Good contrast</p>
</div>
```

**Actual**: The rule fails to pick up the background-color because it is on a `display: table-row` element. This is similar to how `<tr>` and `<thead>` elements aren't picked up by `elementsFromPoint`.

dbjorge:
Note that this issue repros on Chromium (stable and dev) but not Firefox, as of writing. We are reasonably confident that the root cause is [Chromium bug 647594](https://bugs.chromium.org/p/chromium/issues/detail?id=647594). We already notified the Chromium team of the axe-core impact and they have already escalated the bug to P1 accordingly, tagged for release M-78.

jeankaplansky:
Include as part of the overall improvements/fixes to color-contrast

jeankaplansky:
https://docs.google.com/document/d/14PI0e3X9g6CBWtl86HReHrRxl-_zapZBE8HaKoOHrmE/edit#heading=h.cy2pxqhzo2cm

padmavemulapati:
Throwing color-contrast voilation.
![image](https://user-images.githubusercontent.com/47409579/73331276-468fbf80-4289-11ea-8d80-8ae93345c5b3.png)

atotic:
FYI: Chrome will not change elementFromPoint behavior for rows/sections. I've tested it in all modern browsers, and it seems that current Firefox matches Chrome's behavior.

Background color of a table cell is also determined by columns. This makes it impossible to determine true table background cell from javascript in general case. (unless there is an api I am not aware of). I suggest filling a csswg issue if this is a capability you need for accessiblity.

Issue #1920, opened 2019-11-27
Incompatible with smooth scroll-behavior


simonberube:
Axe doesn't identify all issues on a page if it has CSS smooth-scroll:
```css
html { scroll-behavior: smooth; }
```

I'm not sure which issues require Axe to scroll, but color contrast is one of them.

Expectation: Axe should identify issues regardless of scroll-behavior. Scroll-behavior could temporarily set to auto and then returned to its previous value. It might be more appropriate to have Axe delay analysis until each scroll position has been reached.

```css
html { scroll-behavior: auto; }
```
```javascript
document.body.parentElement.style.scrollBehavior = 'auto';
```

Actual: Axe appears to go through window scroll positions quickly, and it never reaches the destination while CSS smooth scrolling is enabled.

Motivation: (self-evident - to have all results)

<pre><code>
axe-core version: 3.3.2
axe-webdriver, extension or other integration version: v4.1.0

Browser and Assistive Technology versions

For Tooling issues:
- Platform:  Windows 10
</code></pre>

simonberube:
How's this for a solution?

```javascript
// get initial scroll behavior
var initial_scroll_behavior = document.body.parentElement.style.scrollBehavior;

// set normal scroll behavior
document.body.parentElement.style.scrollBehavior = 'auto';

/*
  run Axe here
*/

// restore initial scroll behavior
document.body.parentElement.style.scrollBehavior = initial_scroll_behavior;
```

Depending on how Axe works, this concept might have to be repeated for all elements it scrolls through.

WilcoFiers:
Thanks for reporting! Yup that is definitely something to consider. Axe-core only scrolls in the color-contrast rule at the moment. We have a rewrite of the color contrast check in the works that wouldn't involve scrolling, so that should hopefully fix the issue.

padmavemulapati:
As we no longer scroll the page it doesn't affect smooth-scroll behavior, tested it with larger test file which is `https://web.archive.org/web/20180625151005/https://users.math.msu.edu/users/wongwwy/expository/torsion.html` , on running the performance of this larger fine found that its not taking more than 11s for total rendering and analyzing , confirming as everything looks good.
![image](https://user-images.githubusercontent.com/47409579/74041933-9fc7c380-49ec-11ea-92e8-4bd3434d2031.png)

Issue #1685, opened 2019-07-08
MathJax issue ?


damien-git:
aXe seems to have a hard time analyzing https://users.math.msu.edu/users/wongwwy/expository/torsion.html . I noticed the issue first with axe-webdriverjs. I also tried the aXe plugin in both Firefox and Chromium, and had to force stop it in both cases.
Possibly because of the use of MathJax ?

straker:
Looking at the page, it seems the color-contrast rule is taking 23 seconds to complete. Axe-core's color-contrast rule is one of the slowest rules we have, and it looks like all the individual math symbols are being checked for color contrast.

damien-git:
Interesting. Would there be any way to accelerate the process for similar pages ?

straker:
Right now our only workaround is to turn off the color-contrast rule for such pages.

```js
axe.run({
  "rules": {
    "color-contrast": { enabled: false },
  }
});
```

WilcoFiers:
It is also worth noting that we're working on a major performance improvement to the color-contrast rule. That will hopefully help to solve this.

damien-git:
This is good news.
I wrote a [domain accessibility audit](https://github.com/MSU-NatSci/DomainAccessibilityAudit) web app based on axe that can check many websites at once and report statistics. Improved axe-core performance would be a must for that.

jeankaplansky:
ID'd as color-contrast issue. We're going to make one entry into the release notes for all color-contrast fixes/enhancements.

jeankaplansky:
Added to release notes as of 2/04/2019

padmavemulapati:
As the reported url no longer working, verified on the another suggested url, `http://web.archive.org/web/20180625151005/https://users.math.msu.edu/users/wongwwy/expository/torsion.html ` time taking here is less than 2secs
![image](https://user-images.githubusercontent.com/47409579/73907923-0efac600-48ce-11ea-947a-5b983c820ee3.png)

Issue #1539, opened 2019-05-02
Color-contrast rule returns incomplete and claims node is partially obscured


straker:
Run Axe on https://straker.github.io/kontra/.

Expectation: Color contrast should pass

Actual: Color contrast returns incomplete for all nodes under the fold claiming the elements are partially obscured, even though no element obscures it.

I tried to figure out why the rule fails with @scurker, but we both gave up after we couldn't reproduce a reliable test case that worked consistently. 

The best we could determine is that when the color-contrast rule tries to determine the background color of the element, it thinks `document.body` does not visually contain the element. We thought it was because there is an inner scroll element `<main>` that captures the scroll instead of the body itself, so body never scrolls. But any time we tried to create a test case to reproduce that, the color-contrast rule would pass.

In summary: we think the code for `color.getBackgroundColor` scrolling the element with `scrollIntoView` is not scrolling the body element, so when checking `dom.visuallyContains`, the element fails for `document.body`.

<pre><code>
axe-core version: 3.2.2
axe-devtools 3.8.0
browser: Chrome
</code></pre>

robertvrabel:
I'm having a similar issue. It looks like when putting a background-color on the body element it throws off the color-contrast check.

If you move it to the html element instead it seems to resolve the issue.

jeankaplansky:
ID'd as color-contrast issue. We're going to make one entry into the release notes for all color-contrast fixes/enhancements.

jeankaplansky:
ID'd as color-contrast issue. We're going to make one entry into the release notes for all color-contrast fixes/enhancements.

padmavemulapati:
@straker , ran axe on, https://straker.github.io/kontra/ , I am seeing "all issues (17)" count, and all 17 were under "Needs to review" section. Verified with the latest axe. What is the expected behaviour here?

padmavemulapati:
Now with the latest axe, getting all issues(17) in "Incomplete", working as expected.

padmavemulapati:
![image](https://user-images.githubusercontent.com/47409579/74318327-052c0380-4da3-11ea-899c-c7e3fe43851f.png)

PR #1943, opened 2019-12-16
feat(color-contrast): greatly improve performance for very large sites

straker:
Remove the scroll requirement from color-contrast and use the new performance api of `dom.getElementStack` (#1842). I also tried to milk the performance of the `dom.getElementStack` api for all it's worth (using `TreeWalker`, caching lookups, etc.), but I think it's the best I can get it. Lastly, to get correct color contrast, I needed every node on the page to be part of the 2d grid, even if they weren't in the axe context. So I had to create temporary VirtualNodes for any node that was missing from the virtual tree so I could use them in the stack.

This returns my favorite site https://web.archive.org/web/20190613132353/https://giveawaylisting.com/ in about 5s. Slower than I wanted but way better than never returning. All other sites seem to return in about the same time as the old code.

Looking at the flame chart, you can see that the initial setting up of the 2d grid takes 2s, and then it's very tiny stacks for each of the 50906 nodes. So nothing is taking too long to run, it's just processing that many nodes is the bottleneck and something that probably can't be overcome.

image

For #1685, the performance went from 23s to just a few seconds to run color-contrast (had to use an archived view again https://web.archive.org/web/20180625151005/https://users.math.msu.edu/users/wongwwy/expository/torsion.html).

Closes issue: #1588, #318, #1358, #1920, #1730, #1685, #1539, #1951

## Reviewer checks


lib/commons/dom/visually-contains.js
```diff
@@ -58,4 +73,39 @@ dom.visuallyContains = function(node, parent) {
...
+ * Checks whether a parent element visually contains its child, either directly or via scrolling.
+ * Assumes that |parent| is an ancestor of |node|.
+ * @method visuallyContains
+ * @memberof axe.commons.dom
+ * @instance
+ * @param {Element} node
+ * @param {Element} parent
+ * @return {boolean} True if node is visually contained within parent
+ */
+dom.visuallyContains = function visuallyContains(node, parent) {
```
emilyrohrbough: This should reset the scroll position after the check has finished. See #1951

straker: With these changes the code no longer needs to scroll the page to check for color-contrast.

PR #1944, opened 2019-12-16
test: prevent timeout in parseCrossOriginStylesheet test

straker:
This test was always timing out in Phantom in the terminal which was preventing running `npm test` and getting past it. Turns out axios was taking longer than the PhantomJS timeout, so I increased it and used the preload timeout for axios.

## Reviewer checks

### DIFF
--- a/lib/core/utils/parse-crossorigin-stylesheet.js
+++ b/lib/core/utils/parse-crossorigin-stylesheet.js
@@ -19,7 +19,7 @@ axe.utils.parseCrossOriginStylesheet = function parseCrossOriginStylesheet(
 ) {
 	const axiosOptions = {
 		method: 'get',
-		timeout: 5000,
+		timeout: axe.constants.preload.timeout,
 		url
 	};
 
--- a/lib/core/utils/parse-crossorigin-stylesheet.js
+++ b/lib/core/utils/parse-crossorigin-stylesheet.js
@@ -19,6 +19,7 @@ axe.utils.parseCrossOriginStylesheet = function parseCrossOriginStylesheet(
 ) {
 	const axiosOptions = {
 		method: 'get',
+		timeout: 5000,
 		url
 	};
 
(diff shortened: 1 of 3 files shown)


lib/core/utils/parse-crossorigin-stylesheet.js
```diff
@@ -19,6 +19,7 @@ axe.utils.parseCrossOriginStylesheet = function parseCrossOriginStylesheet(
 ) {
 	const axiosOptions = {
 		method: 'get',
+		timeout: 5000,
```
jkodu: Rather than an arbitrary number, we should use https://github.com/dequelabs/axe-core/blob/develop/lib/core/utils/preload.js#L32

Because if I want my preload to kick the bucket in `3000` instead & it does not then that is not ideal

test/core/utils/parse-crossorigin-stylesheet.js
```diff
@@ -60,6 +60,8 @@ describe('axe.utils.parseCrossOriginStylesheet', function() {
 	});
 
 	it('rejects when given url to fetch is not found', function(done) {
+		this.timeout(6000);
```
jkodu: Test needs to pass the `preload.timeout` option

lib/core/utils/parse-crossorigin-stylesheet.js
```diff
@@ -19,6 +19,7 @@ axe.utils.parseCrossOriginStylesheet = function parseCrossOriginStylesheet(
 ) {
 	const axiosOptions = {
 		method: 'get',
+		timeout: axe.constants.preload.timeout,
```
jkodu: Defaulting to `preload.timeout` constant means something passed as a configuration will be ignored.

Eg:
```
axe.run({
	runOnly: {
		...
	},
	preload: {
		assets: ['cssom', ...],
		timeout: 10000
	}
})
```

In this case I specifically want timeout to be after x seconds and not a fallback constant.

> Note: This is still not right, but I will open a PR for this.

Issue #1506, opened 2019-04-18
New Rule: no-autoplay-audio


jkodu:
# no-autoplay-audio

- Description: `This rule checks that `autoplay` audio does not last for more than 3 seconds, or it has a control mechanism`
- Help: `<video> or <audio> elements does not autoplay audio`

## Tags
- wcag2
- wcag142

## Matches

- `"selector": "audio[autoplay], video[autoplay]"`
- `"excludeHidden": false`, because some user agents hide media elements when controls attribute is missing.
- exclude nodes that are either `paused` or `muted`.

### Check 1:
  - name: `no-autoplay-audio-below-allowed-duration`
    - impact: serious
    - messages: 
      - pass: `<video> or <audio> does not output audio for more than allowed duration`
      - fail: `<video> or <audio> outputs audio for more than allowed duration`
      - incomplete: `Check that the <video> or <audio> does not output audio for more than allowed duration`

### Check 2: 
  - name: `no-autoplay-audio-has-controls`
    - impact: serious
    - messages: 
      - pass: `<video> or <audio> has controls mechanism`
      - fail: `<video> or <audio> does not have controls mechanism`
      - incomplete: `Check that the <video> or <audio> has controls mechanism`

## ACT Rule Reference:
- [Video or audio has no auto-play audio](https://act-rules.github.io/rules/80f0bf)

jkodu:
@WilcoFiers - Please see proposal above, feedback welcome.

jeankaplansky:
rule help infrastructure created 1/27/2020

jeankaplansky:
New rule help created, captured in release notes 2/03/2020

padmavemulapati:
Verified with 'no-autoplay-audio.html' test file, getting 'needs review' issues.
![image](https://user-images.githubusercontent.com/47409579/74039925-b8ce7580-49e8-11ea-9733-67f2ac029b5e.png)

padmavemulapati:
Verified with axe-coconut , 
![image](https://user-images.githubusercontent.com/47409579/74312162-2981e300-4d97-11ea-8e82-af0cee39bb07.png)

PR #1946, opened 2019-12-18
feat(rule): no-autoplay-audio

jkodu:
This rule checks that automatically playing audio/video does not last for more than 3 seconds, or the audio/video has a control mechanism to stop or mute it.

Closes issue:
- #1506 

## Reviewer checks


lib/core/utils/preload-media.js
```diff
@@ -7,10 +7,28 @@
  * @property {Object} options.treeRoot (optional) the DOM tree to be inspected
  */
 axe.utils.preloadMedia = function preloadMedia({ treeRoot = axe._tree[0] }) {
-	const mediaVirtualNodes = axe.utils.querySelectorAll(
-		treeRoot,
-		'video, audio'
-	);
+	const mediaVirtualNodes = axe.utils
+		.querySelectorAll(treeRoot, 'video, audio')
```
straker: could use `querySelectorAllFilter()`

lib/core/utils/is-hidden.js
```diff
@@ -15,6 +15,11 @@ axe.utils.isHidden = function isHidden(el, recursed) {
 		return false;
 	}
 
+	// do not exclude `video` or `audio` el's
```
straker: Why should audio and video elements always return `false` for `isHidden`?

lib/checks/media/no-autoplay-audio.json
```diff
@@ -0,0 +1,15 @@
...
+	"id": "no-autoplay-audio",
+	"evaluate": "no-autoplay-audio.js",
+	"options": {
+		"allowedDuration": 3
+	},
+	"metadata": {
+		"impact": "moderate",
+		"messages": {
+			"pass": "<video> or <audio> does not output audio for more than allowed duration or has controls mechanism",
+			"fail": "<video> or <audio> either outputs audio for more than allowed duration or does not have a controls mechanism",
```
straker:
```diff
- "<video> or <audio> either outputs audio for more than allowed duration or does not have a controls mechanism"
+ "<video> or <audio> outputs audio for more than allowed duration and does not have a controls mechanism"
```

test/checks/media/no-autoplay-audio.js
```diff
@@ -0,0 +1,90 @@
...
+		var checkArgs = checkSetup(
+			'<audio id="target" src="/test/assets/moon-speech.mp3" autoplay="true" controls></audio>'
+		);
+		axe.utils.preload(preloadOptions).then(function() {
+			assert.isTrue(check.evaluate.apply(checkContext, checkArgs));
+			done();
+		});
+	});
+
+	it('returns true when <video> can autoplay and has no controls mechanism', function(done) {
```
straker:
```diff
- returns true when <video> can autoplay and has no controls mechanism
+ returns true when <video> can autoplay and has controls mechanism
```

test/integration/rules/no-autoplay-audio/no-autoplay-audio.html
```diff
@@ -0,0 +1,42 @@
+<!-- Passed  -->
```
straker: Add test for audio/video element with `autoplay` attribute but no value: `<audio autoplay>`

test/rule-matches/no-autoplay-audio-matches.js
```diff
@@ -0,0 +1,41 @@
...
+		var vNode = queryFixture(
+			'<video id="target" autoplay="true" muted="true"><source src="/test/assets/video.mp4" type="video/mp4" /></video>'
+		);
+		var actual = rule.matches(vNode.actualNode);
+		assert.isFalse(actual);
+	});
+
+	it('returns true for <audio> element that has loop attribute', function() {
+		var vNode = queryFixture(
+			'<video id="target" loop="true"><source src="/test/assets/video.mp4" type="video/mp4" /></video>'
```
straker: This is a video element (test says audio)

test/rule-matches/no-autoplay-audio-matches.js
```diff
@@ -0,0 +1,41 @@
+describe('no-autoplay-audio-matches', function() {
```
straker: Add test case for media with duration <3 seconds and no loop attribute (should return false)

test/integration/rules/no-autoplay-audio/no-autoplay-audio.html
straker: Add test case for audio/video element with `autoplay` attribute, a duration / playableDuration <3, and loop attribute (should fail)

test/rule-matches/no-autoplay-audio-matches.js
jkodu: This is tested in the check, rather than rule matches.
The rule matches fn is not aware of any allowedDuration & this is a part of the check.

straker: https://github.com/dequelabs/axe-core/pull/1946/files#diff-fbe82cba4f16ea544ec75aecbaedef3fR18-R23

jkodu: I still have not pushed my changes yet or dismissed your reviews. I am only recording my thoughts here. So hold off from commenting or reviewing until all my changes drop in.

lib/core/utils/is-hidden.js
```diff
@@ -15,6 +15,14 @@ axe.utils.isHidden = function isHidden(el, recursed) {
 		return false;
 	}
 
+	/**
+	 * Handle media elements that can be hidden by the absence of an attribute
+	 * Eg: `audio` without `controls` attribute is hidden by default,
+	 */
+	if (el.nodeName.toUpperCase() === `AUDIO` && !el.hasAttribute('controls')) {
+		return false;
```
straker: This should return true as the element is hidden

test/core/utils/is-hidden.js
```diff
@@ -40,6 +40,27 @@ describe('axe.utils.isHidden', function() {
...
+	});
+
+	it('should return false on a `audio` element with `controls` attribute', function() {
+		fixture.innerHTML =
+			'<audio id="target" src="/test/assets/moon-speech.mp3" controls></audio>';
+		var el = document.getElementById('target');
+		assert.isFalse(axe.utils.isHidden(el));
+	});
+
+	it('should return false on a `audio` element with `controls` attribute but `display:none`', function() {
```
straker: The test returns `true` but the title says it returns `false`

test/core/utils/is-hidden.js
```diff
@@ -40,6 +40,27 @@ describe('axe.utils.isHidden', function() {
 		assert.isFalse(axe.utils.isHidden(document));
 	});
 
+	it('should return false on a `audio` element without `controls` attribute', function() {
+		fixture.innerHTML =
+			'<audio id="target" src="/test/assets/moon-speech.mp3"></audio>';
+		var el = document.getElementById('target');
+		assert.isFalse(axe.utils.isHidden(el));
```
straker: This should return `true` as the node is hidden without the `controls` attribute

test/integration/rules/no-autoplay-audio/no-autoplay-audio.html
```diff
@@ -0,0 +1,57 @@
...
+<audio
+	id="fail4"
+	src="/test/assets/moon-speech.mp3#t=1,3"
+	autoplay="true"
+	loop="true"
+></audio>
+
+<!-- Inapplicable  -->
+<!-- inapplicable 1 -->
+<video id="inapplicable1" autoplay="true" muted="true">
```
straker: Add test case for being `paused` as well

lib/core/utils/is-hidden.js
jkodu: We are trying to negate what is by default hidden.

`excludeHidden` is a flag we have on rules to aggregate the nodes necessary for each rule to run. By default excludeHidden is true. When excludeHidden is true, the isHidden fn is called to filter the nodes.

We only have the option to say `excludeHidden: false` in a rule's config, which basically means the nodes are not filtered via the isHidden fn.

Here we are manipulating HTMLMediaElement `audio` which by default is hidden when there is no controls attribute in browsers, we want that node to run through our check, so we want this to return false, so it gets added as a candidate for the check.

test/core/utils/is-hidden.js
jkodu: See explanation above - https://github.com/dequelabs/axe-core/pull/1946#discussion_r367033403

Issue #978, opened 2018-06-27
'landmark-no-duplicate-banner' &  'landmark-no-duplicate-contentinfo' return incorrect element source


iamrafan:
All the elements that fail these two rules are return as related nodes unlike in the case of duplicate-id failures, where the first matching node is returned as a failure node and the rest of the nodes are returned as related nodes 

![image](https://user-images.githubusercontent.com/22068434/41998784-6c399ae6-7a10-11e8-8a16-fe0399ed9de7.png)

<pre><code>axe-core version: 3.0.3
axe-Coconut version: 3.4.0-alpha.2</code></pre>

WilcoFiers:
This is the same issue as #878.

WilcoFiers:
Proposal to use the after function to turn the related nodes into matched nodes, so that the result would go from look like this:

```js
[{
   target: 'html'
   relatedNode: [{
     target: '#div1'
   }, {
     target: '#div2'
   }, {
     target: '#div3'
   }]
}]
```

to looking like this:

```js
[{
  target: '#div1',
  relatedNodes: [{
    target: 'html'
  }]
}, {
  target: '#div2',
  relatedNodes: [{
    target: 'html'
  }]
}, {
  target: '#div3',
  relatedNodes: [{
    target: 'html'
  }]
}]
```

marcysutton:
We also talked about limiting results to common ancestors, so we don't get a bunch of extra noise. This will take the region rule from one result to many, to make it easier to understand.

WilcoFiers:
Me and Jey have been looking into this, and it isn't as simple as tweaking the data in "after". Unfortunately axe-core doesn't let you flip related nodes and selected nodes. We'd have to rewrite these rules to do this I think, which is a bummer. I do think that's probably the right way to do it.

WilcoFiers:
Talking to Steve about this. The way to deal with this is to select all the landmarks in the rule, and use the after method to remove any duplicates after the first. This would be similar to how we do it in #1787 and #1649.

jeankaplansky:
new rule landmark-no-duplicate-main infrastructure created 1/27/2020

jeankaplansky:
landmark-no-duplicate-main rule create in axe-core rule help and release notes 2/03/2020

padmavemulapati:
Ensuring no duplicate landmarks , verified with the latest axe.
[...]

PR #1949, opened 2019-12-18
feat(landmark-no-duplicate-*): add rule landmark-no-duplicate-main, don't use html as element source for all duplicate rules

straker:
Changes the element returned from `html` to whichever is the first duplicate element. However, doing this broke the `landmark-one-main` rule since it did two things at once: ensured there was a `main` element and that there was only one. Since I couldn't do both with a non-html selector, I had to split the rule into two rules to cover both cases

Closes issue: #978

## Reviewer checks


lib/checks/keyboard/page-no-duplicate.js
```diff
@@ -4,7 +4,16 @@ if (!options || !options.selector || typeof options.selector !== 'string') {
 	);
 }
 
-let elms = axe.utils.querySelectorAll(virtualNode, options.selector);
+// only look at the first node and it's related nodes
+const key = 'page-no-duplicate;' + options.selector;
+if (axe._cache.get(key)) {
+	return 'ignored';
```
WilcoFiers: Hmm, not too keen on having strings as the return value of checks. I think we should use date() for this.

lib/checks/keyboard/page-no-duplicate.js
```diff
@@ -4,7 +4,16 @@ if (!options || !options.selector || typeof options.selector !== 'string') {
...
+}
+axe._cache.set(key, true);
+
+let elms = axe.utils
+	.querySelectorAll(axe._tree[0], options.selector)
+	.filter(elm => elm !== virtualNode);
```
WilcoFiers: You could also use `axe.querySelectorAllFilter()`. Just a thought.

lib/rules/landmark-one-main.json
```diff
@@ -3,10 +3,10 @@
 	"selector": "html",
 	"tags": ["cat.semantics", "best-practice"],
 	"metadata": {
-		"description": "Ensures the document has only one main landmark and each iframe in the page has at most one main landmark",
+		"description": "Ensures the document has a main landmark and each iframe in the page has a main landmark",
```
WilcoFiers: This has changed a little. This rule should check that the top window has at least one main landmark. What's in iframes shouldn't matter for this rule. Then the new rule should check that there is no more than one main in any given document. That way the top window must have 1 main, and iframes can have 0 or 1 main.

test/integration/full/landmark-no-duplicate-main/landmark-no-duplicate-main-pass.js
```diff
@@ -0,0 +1,43 @@
...
+
+	describe('passes', function() {
+		it('should find 1', function() {
+			assert.lengthOf(results.passes[0].nodes, 1);
+		});
+
+		it('should find #frame1, #pass2', function() {
+			assert.deepEqual(results.passes[0].nodes[0].target, [
+				'#frame1',
+				'#pass2'
```
WilcoFiers: Why is this pass2 if there's only 1 pass?

Issue #1009, opened 2018-07-16
Aria-required-children violation on role=combobox not valid


ShawnChenXHC:
Hello, I would like to raise an issue around the `aria-required-children` violation.

Consider the following example:
```HTML
<div role="combobox" aria-expanded="true" aria-haspopup="grid" aria-owns="ex-grid">
  <input type="text" aria-controls="ex-grid" />
</div>
<div role="grid" id="ex-grid">
</div>
```

When I run an aXe test on the above set of elements, the test would fail because of a violation of the `aria-required-children` check. The details reveal that the check failed because the `<input>`, having being defined with `role="combobox"`, must have a child with `role="listbox"`.

However, according to the WAI-ARIA guidelines for [comboboxes](https://www.w3.org/TR/wai-aria-1.1/#combobox), we can see that aside from `listbox`, there are three other permitted children roles: `tree`, `grid`, and `dialog`. The above example satisfies this condition because the `combobox` specifies its popup as being a `grid` and uses `aria-owns` to point to the specific `grid` tied to this `combobox`.

Looking at the role lookup table inside of the aXe engine, there are no provisions for these alternatives under the `combobox` section. If I am interpreting the WAI-ARIA guidelines correctly, then I believe this should be updated.

Keerthi-Penukonda:
@WilcoFiers , tested the given code snippet with role="listbox" and also with role="grid". In both the cases screen reader behavior is same.	
Tested in 
1. Chrome - NVDA on Windows
2. Firefox - NVDA on Windows
3. IE - JAWS on Windows
Please let me know if anything needed.

marcysutton:
We need to test all of the combinations in our support matrix, there are a few more: https://github.com/dequelabs/axe-core/blob/develop/doc/accessibility-supported.md

Keerthi-Penukonda:
Below are the screen reader behaviors with different combinations:

JAWS and IE on Windows: 
1.	With role grid: Screen reader is announcing as “Edit combo use JAWS key+alt+m to move to the controlled element”.  To set the value use the arrows keys or type the value.
2.	With role listbox: Screen reader is announcing as “Edit combo use JAWS key+alt+m to move to the 
controlled element”.  To set the value use the arrows keys or type the value.

 NVDA and FF on Windows:
1.	With role grid: Screen reader is announcing as “combobox expanded Edit has autocomplete blank” and sometimes announcing as just  “Edit”
[...]

PR #1950, opened 2019-12-18
fix(aria-required-children): allow comboboxes with more popup roles

AdnoC:
Adds the other allowed expanded child types to the `combobox` owned requirements. If the box is open it checks what the expected role of the expanded element should be and only makes that required. If the box is closed it doesn't make any of them required.

Also modernized the rest of the file (`var` -> `let` or `const`)

Closes issue: https://github.com/dequelabs/axe-core/issues/1009

## Reviewer checks

### DIFF
--- a/lib/checks/aria/required-children.js
+++ b/lib/checks/aria/required-children.js
@@ -78,7 +78,9 @@ function missingRequiredChildren(node, childRoles, all, role) {
 		const expandedChildRoles = ['listbox', 'tree', 'grid', 'dialog'];
 		const expandedValue = node.getAttribute('aria-expanded');
 		const expanded = expandedValue && expandedValue !== 'false';
-		const popupRole = node.getAttribute('aria-haspopup') || 'listbox';
+		const popupRole = (
+			node.getAttribute('aria-haspopup') || 'listbox'
+		).toLowerCase();
 
 		for (let index = 0; index < expandedChildRoles.length; index++) {
 			const expandedChildRole = expandedChildRoles[index];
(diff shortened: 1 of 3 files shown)


lib/checks/aria/required-children.js
```diff
@@ -78,11 +75,23 @@ function missingRequiredChildren(node, childRoles, all, role) {
...
-			missing.splice(listboxIndex, 1);
+		const expandedChildRoles = ['listbox', 'tree', 'grid', 'dialog'];
+		const expandedValue = node.getAttribute('aria-expanded');
+		const expanded = expandedValue && expandedValue !== 'false';
+		const popupRole = node.getAttribute('aria-haspopup') || 'listbox';
+
+		for (let index = 0; index < expandedChildRoles.length; index++) {
+			const expandedChildRole = expandedChildRoles[index];
+			// keep the specified popup type required if expanded
+			if (expanded && expandedChildRole === popupRole) {
```
WilcoFiers: should probably make this case insensitive.

lib/checks/aria/required-children.js
```diff
@@ -78,7 +78,9 @@ function missingRequiredChildren(node, childRoles, all, role) {
 		const expandedChildRoles = ['listbox', 'tree', 'grid', 'dialog'];
 		const expandedValue = node.getAttribute('aria-expanded');
 		const expanded = expandedValue && expandedValue !== 'false';
-		const popupRole = node.getAttribute('aria-haspopup') || 'listbox';
+		const popupRole = (
+			node.getAttribute('aria-haspopup') || 'listbox'
+		).toLowerCase();
```
WilcoFiers: Please test this too.

PR #1953, opened 2019-12-19
feat(svg-img-alt): rule for when svg needs a title

WilcoFiers:
This new svg-img-alt rule checks that the `svg` elements with `role=graphics-document`, and any element in the SVG namespace that has `role=img` or `role=graphics-symbol` has an accessible name, either through ARIA label, the `title` attribute or by having a non-empty child `title` element.

This rule limits the role-img-alt rule to elements in the HTML namespace to avoid overlap with the svg-img-alt rule.

Closes issue: #20

## Reviewer checks


lib/rules/role-img-alt.json
```diff
@@ -1,6 +1,7 @@
 {
 	"id": "role-img-alt",
-	"selector": "[role='img']:not(svg):not(img):not(area):not(input):not(object)",
+	"selector": "[role='img']:not(img):not(area):not(input):not(object)",
```
straker: Should add a test to this rules integration test to ensure svg elements aren't captured by this rule now.

WilcoFiers: gd catch.

lib/checks/shared/svg-non-empty-title.js
```diff
@@ -0,0 +1,4 @@
+const titleNode = virtualNode.children.find(({ props }) => {
+	return props.nodeName === 'title';
```
jkodu: I know this is a nitpick, but don't we always go `nodeName.toUpperCase`?

straker: The `VirtualNode` interface does that for us now https://github.com/dequelabs/axe-core/blob/develop/lib/core/base/virtual-node/virtual-node.js#L27-L36

Issue #1919, opened 2019-11-27
Clarify reporting of ul > li[role=button] and other list rules with roles


WilcoFiers:
Similar to #1894, axe-core reports the following code this way:

```html
<ul>
  <li role="button">Stuff</li>
</ul>
```

> Fix the following:
> List element has direct children that are not allowed inside <li> elements

This message is a little misleading. We should make sure it's clear that this is failing because of the role.  Axe-core should be clear that whenever something is failed for having it's implicit role changes, that the role attribute is the reason for the failure.

straker:
@jeankaplansky I don't believe the rule-help docs need to be updated since we just added a new remediation message to clarify the problem.

padmavemulapati:
Showing a message "List element has direct children that are not allowed inside <li> elements", which is the fix seeing.

![image](https://user-images.githubusercontent.com/47409579/73372423-3275ad80-42dd-11ea-9371-beedf0e523f8.png)

PR #1954, opened 2019-12-19
fix(only-listitem): add message about invalid role on li elements

straker:
Add a new message when an `<li>` element has a role that is not `listitem`. I also heavily refactored and simplified the code because it was super confusing to read through and try to figure out what was going on (lots of negation statements, a `reduce` that wasn't needed, confusing logic, didn't make use of `aria.getRole`, etc.).

Closes issue: #1919

## Reviewer checks

### DIFF
--- a/lib/checks/lists/only-listitems.js
+++ b/lib/checks/lists/only-listitems.js
@@ -1,64 +1,45 @@
-const { dom } = axe.commons;
-const getIsListItemRole = (role, tagName) => {
-	return role === 'listitem' || (tagName === 'LI' && !role);
-};
+const { dom, aria } = axe.commons;
 
-const getHasListItem = (hasListItem, tagName, isListItemRole) => {
-	return hasListItem || (tagName === 'LI' && isListItemRole) || isListItemRole;
-};
+let hasNonEmptyTextNode = false;
+let atLeastOneListitem = false;
+let isEmpty = true;
+let badNodes = [];
 
-let base = {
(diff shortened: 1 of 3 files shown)


lib/checks/lists/only-listitems.json
```diff
@@ -5,7 +5,10 @@
 		"impact": "serious",
 		"messages": {
 			"pass": "List element only has direct children that are allowed inside <li> elements",
-			"fail": "List element has direct children that are not allowed inside <li> elements"
+			"fail": {
+				"default": "List element has direct children that are not allowed inside <li> elements",
+				"roleNotValid": "List element has direct children with a role that is not allowed"
```
WilcoFiers: I think it's worth mentioning which role it is. Would make this message clearer.

lib/checks/lists/only-listitems.js
```diff
@@ -1,64 +1,45 @@
...
-		if (tagName !== 'LI' && !isListItemRole) {
-			out.badNodes.push(actualNode);
-		}
-	}
-	if (actualNode.nodeType === 3) {
-		if (actualNode.nodeValue.trim() !== '') {
-			out.hasNonEmptyTextNode = true;
-		}
+	if (!isLi && !isListItemRole) {
+		badNodes.push(actualNode);
```
WilcoFiers: We should put `li` elements with its role changed in the related nodes too.

PR #1955, opened 2019-12-19
test: using headless chrome (and remove PhantomJs)

jkodu:
> I came across this package [mocha-headless-chrome](https://www.npmjs.com/package/mocha-headless-chrome) and that gave an idea to put this PR together, for bug day. 

**Some Context:**
For a while now PhantomJs has been deprecated & we have tried various attempts (like authoring our own puppeteer wrapper and or use karma) in order to lose `grunt-mocha` which internally uses PhantomJs.

**Some more History (Backstory):**
Currently axe uses `grunt` as a task runner. The `test` task runs all tests under `tests/**` directory in a 2 step fashion
- IN HEADLESS MODE
	-> using `grunt-mocha` which under the hood uses `phantomjs`
- IN PARALLEL TESTS
	-> uses `selenium-webdriver` and runs tests in a variety of browser configurations. Refer grunt task `parallel` for more information on the same.

This PR aims to swap out the tooling used in HEADLESS MODE & thus finally losing phantomjs :).

Also note, although the `npm run test:headless` script can run on it's own, I have wired it to be called via a grunt task, as there are some preparation tasks to be completed (like building, generating fixtures etc.,) which needs to happen before headless tests or tests can be run in general.

> I have not removed all usages of `window.PHANTOMJS` to keep the number of file changes in this PR to a minimum.

Closes issue:
- NA

## Reviewer checks


test/integration/no-ui-reporter.js
```diff
@@ -1,6 +1,11 @@
 /*global mocha, console */
 (function() {
 	'use strict';
+
+	if (!mocha || !mocha.reporter || !mocha.reporter('base')) {
```
jkodu: > Note: Non-headless tests (that run on selenium-webdriver) need to wire up to mocha reporter to collect all the results and report them. (See `test-webdriver.js`). But for other tests this can be skipped.

straker:
Somethings not right. I wanted to test what a failure would look like, so I changed a single assert to fail and now I have 100s of errors in random tests:

```diff
diff --git a/test/commons/dom/is-visible.js b/test/commons/dom/is-visible.js
index 34674387..90ee39d5 100644
--- a/test/commons/dom/is-visible.js
+++ b/test/commons/dom/is-visible.js
@@ -20,7 +20,7 @@ describe('dom.isVisible', function() {
                        window.getComputedStyle = function() {
                                return null;
                        };
-                       assert.isFalse(axe.commons.dom.isVisible(fakeNode));
+                       assert.isTrue(axe.commons.dom.isVisible(fakeNode));
                        window.getComputedStyle = orig;
                });
```

image


image

image

jkodu:
Interesting. I made the same change and ran `mocha-headless-chrome --file http://localhost:9876/test/commons/` on it's own & got the same problem. It looks like something to do with the module being used.

jkodu:
@straker It was all due to binding process stdout with the child process. I have now updated the script to use the runner instance via node, instead of executing a command.

Issue #1945, opened 2019-12-16
dom.isFocusable returns true for the summary element


mitchellwills:
Expectation: axe.commons.dom.isFocusable should return true for `<summary>` elements and false for `<details>` elements with `<summary>` elements inside of them

Actual: It always returns true for `<details>` elements and returns false for `<summary>` elements

Motivation: doesn't match browser behavior (tested in recent Chrome, Firefox & Safari)

A simple example (prints focused element to the console): https://next.plnkr.co/edit/5lxR5V2jJWa44oND?open=index.html&preview

<pre><code>
axe-core version: 3.1.2 (appears to be a problem in the current code too)
</code></pre>

WilcoFiers:
Right you are. Thank you for reporting. We'll be sure to fix that!

jeankaplansky:
in release notes as of 2/3/2020

PR #1957, opened 2019-12-19
fix(isFocusable): return true for summary element and false for details element with summary child

straker:
Closes issue: #1945

## Reviewer checks


lib/commons/dom/is-focusable.js
```diff
@@ -65,9 +65,11 @@ dom.isNativelyFocusable = function(el) {
 			return el.type !== 'hidden';
 		case 'TEXTAREA':
 		case 'SELECT':
-		case 'DETAILS':
+		case 'SUMMARY':
 		case 'BUTTON':
 			return true;
+		case 'DETAILS':
+			return !el.querySelector('summary');
```
WilcoFiers: This should test the flattened tree instead. details > summary works across shadow tree boundaries.

test/commons/dom/is-focusable.js
```diff
@@ -360,11 +376,27 @@ describe('is-focusable', function() {
...
+			assert.isTrue(axe.commons.dom.isFocusable(el));
+		});
+
+		it('should return false for a details element with a summary element', function() {
+			fixture.innerHTML =
+				'<details id="target"><summary>Summary</summary><p>Detail</p></details>';
+			var el = document.getElementById('target');
+
+			assert.isFalse(axe.commons.dom.isFocusable(el));
 		});
```
WilcoFiers: Add a shadow DOM test, where details is in the light DOM and summary is in a shadow DOM tree.

lib/commons/dom/is-focusable.js
straker: How does details > summary work across shadow boundaries? Details is not one of the elements you can [attach shadow](https://developer.mozilla.org/en-US/docs/Web/API/Element/attachShadow) to, so you would have to do it to a child of the details. That doesn't property create a summary for the details to use.

```html
<details>
  <div id="shadow"></div>
</details>

<script>
  let details = document.getElementById('shadow');
  let shadow = details.attachShadow({mode: 'open'});
  shadow.innerHTML = `<summary>A summary</summary>
  Some content`;
</script>
```

PR #1958, opened 2019-12-20
feat(core): add preload configuration option for media files

jkodu:
This PR adds an additional type 'media' to the `preload` configuration.
The intention of this configuration is to await running some rules which requires some metadata from HTMLMediaElement to run checks.

An example of one such rule would be [no-autoplay-audio](https://github.com/dequelabs/axe-core/pull/1946), which relies on duration property of a media file.

Closes issue:
- NA

## Reviewer checks


lib/core/utils/preload-media.js
```diff
@@ -0,0 +1,53 @@
...
+		 */
+		if (elm.readyState > 0) {
+			resolve(elm);
+		}
+
+		function onMediaReady() {
+			elm.removeEventListener('loadedmetadata', onMediaReady);
+			if (elm.readyState <= 0) {
+				reject(
+					`Preload media element has wrong readyState of HAVE_NOTHING (0)`
```
WilcoFiers: Put an Error in the reject rather than a string.

lib/core/utils/preload-media.js
```diff
@@ -0,0 +1,53 @@
+/**
+ * Given a rootNode
+ * -> get all HTMLMediaElement's and ensure their metadata is loaded
+ *
+ * @method preloadMedia
+ * @memberof axe.utils
+ * @property {Object} options.treeRoot (optional) the DOM tree to be inspected
+ */
+axe.utils.preloadMedia = function preloadMedia({ treeRoot = axe._tree[0] }) {
```
WilcoFiers: This should have a timeout option. Users need to be able to specify how long they want to wait for this.

lib/core/utils/preload.js
```diff
@@ -81,7 +81,8 @@ axe.utils.getPreloadConfig = function getPreloadConfig(options) {
  */
 axe.utils.preload = function preload(options) {
 	const preloadFunctionsMap = {
-		cssom: axe.utils.preloadCssom
+		cssom: axe.utils.preloadCssom,
+		media: axe.utils.preloadMedia
```
WilcoFiers: We need to document this feature.

test/core/utils/preload-media.js
```diff
@@ -0,0 +1,76 @@
+/* global Promise */
+
+describe('axe.utils.preloadMedia', function() {
+	'use strict';
+
+	var origFn = axe.utils.preloadMedia;
+	var fixture = document.getElementById('fixture');
+	var fixtureSetup = axe.testUtils.fixtureSetup;
+
+	function rebuildTree() {
```
WilcoFiers: Please take this out. You can just as easily put that line in the tests, and have a much more readable test because of it. Having readable tests is more important then having DRY tests.

test/core/utils/preload-media.js
```diff
@@ -0,0 +1,76 @@
...
+
+		axe.utils.preloadMedia({ treeRoot: axe._tree[0] }).then(function(result) {
+			assert.equal(result.length, 0);
+			done();
+		});
+	});
+
+	it('returns media node (audio) after their metadata has been preloaded', function(done) {
+		fixtureSetup(
+			'<audio src="https://act-rules.github.io/test-assets/moon-audio/moon-speech.mp3" autoplay="true" controls></audio>'
```
WilcoFiers: Put this file into the repo so we're not relying on a remote server that's in no way synched with this repo. Same goes for the rest of the tests below.

test/core/utils/preload-media.js
```diff
@@ -0,0 +1,76 @@
...
+				'<video>' +
+				'<source src="https://act-rules.github.io/test-assets/rabbit-video/video.mp4" type="video/mp4" />' +
+				'<source src="https://act-rules.github.io/test-assets/rabbit-video/video.webm" type="video/webm" />' +
+				'</video>'
+		);
+
+		axe.utils.preloadMedia({ treeRoot: axe._tree[0] }).then(function(result) {
+			assert.equal(result.length, 2);
+			assert.isTrue(result[0].readyState > 0);
+			assert.notEqual(result[1].duration, NaN);
```
WilcoFiers: Why aren't you checking the actual duration, instead of what it isn't?

test/core/utils/preload-media.js
```diff
@@ -0,0 +1,76 @@
+/* global Promise */
+
+describe('axe.utils.preloadMedia', function() {
```
WilcoFiers: You're not testing what happens when something errors out.

lib/core/utils/preload-media.js
jkodu: Each of the preload(able) items do not have a special timeout. As in there is a global timeout specified on the `preload` itself, over specifying individual timeouts for each item. 

[This is how we timeout now](https://github.com/dequelabs/axe-core/pull/1958/files/27c18bd54f051116fcf4710c21f17ccc89687b0c#diff-34951a6fe41c0b8543bc30e1c882fe9aR100)

Example we cannot say timeout is 10000 for CSSOM and 2000 for MEDIA. We can simple say all preload expires in a given timeout.

lib/core/utils/preload-media.js
jkodu: There is no error per-se at this point, rejecting when readyState is identified as 0.

test/core/utils/preload-media.js
jkodu: Removed the only reject that was here - More info - https://github.com/dequelabs/axe-core/pull/1958#discussion_r364315256.

lib/core/utils/preload-media.js
WilcoFiers: Thanks for the clarification.

lib/core/utils/preload-media.js
```diff
@@ -0,0 +1,48 @@
...
+	const mediaVirtualNodes = axe.utils.querySelectorAll(
+		treeRoot,
+		'video, audio'
+	);
+	if (!mediaVirtualNodes || !mediaVirtualNodes.length) {
```
WilcoFiers: this is statement can be removed. `mediaVirtualNodes` is will always be an array, and an empty array will work fine with your map function. Promise.all will resolve.

PR #1963, opened 2020-01-02
chore: Add searchbox to widget lookup table

schne324:
Maybe I'm missing some context but it appears that [searchbox was intended to be in the widget lookup table](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/aria/index.js#L2391). In addition, [`commons.aria.lookupTable.role['searchbox'].type`](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/aria/index.js#L1709-L1710) returns "widget"

Closes issue:

## Reviewer checks


straker:
Other than you're spacing being off for some reason (wonder why prettier didn't fix it), seems fine to me. I just want @WilcoFiers or @dylanb to verify we're not missing something before merging.

schne324:
> Other than you're spacing being off for some reason

Ooops! That is what I get for being lazy and editing right here in github 🤦‍♂ I will get that spacing fixed!

schne324:
spacing has been fixed

Issue #1935, opened 2019-12-10
Version of axe-core without polyfills


straker:
To help reduce our bundled size, we should release a version of axe-core without our polyfills. With 3.4.0, we added 60KB minified to our codebase, mostly from the addition of [core-js Unit32Array polyfills](https://github.com/GoogleChrome/lighthouse/pull/10056#issuecomment-562360821) needed for the ligature icon detection. This can get us under size more quickly while we determine a better long term solution.

connorjclark:
I confirmed that a minified webpack bundled w/ just these [polyfills](https://github.com/dequelabs/axe-core/pull/1829/files#diff-91383c5b06b40827c141c40fd2a28787R21) is ~32 KB.

straker:
We decided that we will make `axe.js` in 4.0.0 be polyfill free instead of introducing a new file. For older browsers that require a polyfill, we will document how to bring in the necessary polyfill.

connorjclark:
@paulirish and I are interested in this feature so that we can reduce Lighthouse's bundle size. Is there anything we could do to help?

straker:
Glad you asked. We tried to get ahold of Paul awhile ago in regards to #2357, specifically trying to figure out how large axe-core's gizzped size should be, especially when bundled into Lighthouse. If you could help us figure that out that'd be most helpful.

Once we have a target we can figure out the best approach from there.

connorjclark:
We're attempting to bring Lighthouse's bundle for DevTools down to 1MB unzipped. We have a document tracking this: https://docs.google.com/document/d/15DyId8C9bGnk1ZpgaIekYG_RDEC_KYNQW3Tsp6vs11Q/edit#

1 MB is a long shot for us, but the number isn't particular important, it just serves as a guide. In the same mindset, we can't provide a "real" target number for aXe, because dropping axe from Lighthouse just isn't an option, and at the end of the day we'd accept whatever size as the cost of integration.

What's most important to us is not bundling anything extraneous, like polyfills, as those are usually low-pickings for improvements.

You mentioned 100KB gzipped–that sounds like a great goal to me. If after removing polyfills aXe is still above that number, I'd love to help with some of the tougher ideas you might have.

Is there anything we can do to assist in the polyfill effort? And does that also include axios?

____

[...]

PR #1968, opened 2020-01-06
[WIP] feat(core): add -noPolyfill option to `grunt build`

straker:
Reduces the file size of axe from 893KB to 731KB unminified, and 450KB to 384KB minified.

Partially closes issue: #1935 

## Reviewer checks


WilcoFiers:
Having the build option doesn't fully resolve the issue. We should have these no-poly files published to NPM (and included in sri-history.json)

build/polyfills-generator.js
```diff
@@ -0,0 +1,39 @@
...
+const outputFile = path.join(
+	__dirname,
+	'..',
+	'tmp',
+	'core',
+	'polyfills',
+	'index.js'
+);
+
+async function run() {
```
jkodu: This is a clone of `imports-generator` - https://github.com/dequelabs/axe-core/blob/develop/build/imports-generator.js

I see no reason it has to be though. You should make this a common function, that when given a source and destination and the namespace to mount, it creates these files.

Why have two files with the same code one for creating `axe.imports` and one for creating `axe.polyfills`?

lib/core/polyfills/index.js
```diff
@@ -0,0 +1,262 @@
...
+require('weakmap-polyfill');
+
+/* eslint-disable */
+/*
+ These pollyfills came directly from the ES Specification it's self
+ Contained within:
+  - Object.assign
+  - Array.prototype.find
+*/
+if (typeof Object.assign !== 'function') {
```
jkodu: another thought is, given we have `core-js` already above for `reduce`, can we remove the copy pasted polyfills below and swap them out to features from `core-js`

Eg: `core-js(-pure)/es|stable|features/object/assign`

stephenmathieson: I agree here, but don't think that needs to change for this PR to land

Issue #1906, opened 2019-11-15
Ligature color-contrast checks result as incomplete


davyrwoo:
Expectation: Ligature with failing contrast ratios should result as violations.

Actual: Ligature with failing contrast ratios result as incomplete.

Motivation: Ligature with inaccessible contrast should be considered violations, similar to how text color-contrast issues are violations.

![a11y ligature](https://user-images.githubusercontent.com/56053686/68982203-eecdd600-07ba-11ea-8d90-bb4454d3c772.PNG)

Example code:

```
<html>
	<style>
		.pass {
			color: #D7D7D7;
		}
		.fail {
			color: #AD2F42;
		}
		.dark {
			background: #201F1F
		}
	</style>
	<body class="dark">
		<main>
			<h1 class="pass">A11y test</h1>
			<div>
				<p class="pass"> &AElig </p>
				<p class="fail"> &AElig </p>
			</div>
		</main>
	</body>
</html>
```
axe-core version: 3.4.0
cypress-axe: 0.5.1
cypress: 3.6.1

jeankaplansky:
No customer-facing docs required.

padmavemulapati:
Verified this issue with the latest axe-coconut, I am still seeing the color-contrast issue in the "needs review" not in the voilations. 
@straker , can you please suggest me here, where can I see the fix.

straker:
So the issue is our color-contrast check returns incomplete for the node because the content is too short: "Element content is too short to determine if it is actual text content." Not sure we can do anything about it as the code is working as intended.

padmavemulapati:
As per @straker , nothing else here to do , so closing the issue

KyleBastien:
@padmavemulapati and/or @straker would it be possible to check if the node only contains ligature's and then not care about the content length?

Seems like a big miss if all lone ligature's are unchecked for color contrast, since it's quite common to use ligature's for icon's, especially :heavy_check_mark: or :no_entry_sign: type icon's who might be green or red and should really be checked for color contrast.

padmavemulapati:
@KyleBastien , I ran axe and axe-coconut on the given test file, when I observed with color-contrast analyzer, realised that red colored ligature is having low color-contrast ratio than the standard ratio.. so that issue is coming under needs review, which is expected as per the ticket. Please find below screenshots and please let me know if anything else need to be taken care inorder to evaluate this ticket.
[...]

PR #1969, opened 2020-01-06
feat(rule): add color-contrast check for unicode characters, behind flags.

KyleBastien:
Adding two new config flags to color-contrast so it can check unicode character. Sometimes ligature based icons utilize unicode characters as stand-in's, so these flags become useful to be able to do color-contrast checks on these icons.

ignoreUnicode, defaults to true and retains the behavior of ignoring all unicode characters when doing color contrast. This can be turned off to start checking unicode characters for color contrast.

ignoreLength, defaults to false and retains the behavior that single character nodes do not contain enough information to say whether or not they have color contrast issues. This can be turned on to ignore this length check and always check if a node has color contrast issues.

Closes issue: #1906 

## Reviewer checks

### DIFF
--- a/test/checks/color/color-contrast.js
+++ b/test/checks/color/color-contrast.js
@@ -327,12 +327,69 @@ describe('color-contrast', function() {
 	it('should return true for a single character text with insufficient contrast', function() {
 		var params = checkSetup(
 			'<div style="background-color: #FFF;">' +
-				'<div style="color:#000;" id="target">X</div>' +
+				'<div style="color:#DDD;" id="target">X</div>' +
 				'</div>'
 		);
 
 		var actual = contrastEvaluate.apply(checkContext, params);
+		assert.isUndefined(actual);
+		assert.equal(checkContext._data.messageKey, 'shortTextContent');
+	});
+
+	it('should return undefined when the text only contains nonBmp unicode by default', function() {
+		var params = checkSetup(
+			'<div style="background-color: #FFF;">' +
+				'<div style="color:#DDD;" id="target">◓</div>' +
+				'</div>'
+		);
+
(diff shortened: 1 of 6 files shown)


test/checks/color/color-contrast.js
```diff
@@ -324,7 +324,7 @@ describe('color-contrast', function() {
 		assert.equal(checkContext._data.messageKey, 'shortTextContent');
 	});
 
-	it('should return true for a single character text with insufficient contrast', function() {
+	it('should return true for a single character text with sufficient contrast', function() {
```
straker: I think the description of this test is correct but the code setup is wrong. It seems to me that it should show that a failing contrast of a single character still passes the check (due to being ignored by short text content). Otherwise the test itself would pass as the contrast is sufficient.

test/checks/color/color-contrast.js
```diff
@@ -335,6 +335,61 @@ describe('color-contrast', function() {
 		assert.isTrue(actual);
 	});
 
+	it('should return true when the text only contains nonBmp unicode by default', function() {
+		var params = checkSetup(
+			'<div style="background-color: #FFF;">' +
```
straker: Should probably have the test setup make the contrast invalid so a passing test isn't due to a passing contrast but instead due to being ignored by the check.

test/checks/color/color-contrast.js
KyleBastien: Changed so the test setup has failing contrast for the single character, and asserted that undefined is returned due to short text content. Let me know if that isn't what you were thinking.

test/checks/color/color-contrast.js
KyleBastien: Done!

Issue #1916, opened 2019-11-21
Be explicit about fallback roles being unsupported


WilcoFiers:
Currently, when you do something like this:

```html
<div role="application tablist">
```

Axe-core reports "ARIA role is valid". That's a confusing message. Instead it should say something like "Use only one role value, since fallback roles are not widely supported".

## Proposal on how to do this

1. support fallback roles in the existing *role tools, have them look at each role in the role="" attribute
2. Add a new check that looks just for use of fallback roles

That way something like `role="foobar textbox select"` would be reported as:

```
Fix all of the following:
- Use only one role value, since fallback roles are not widely supported"
- Abstract role 'select' cannot be directly used
- Role 'foobar' must be a valid ARIA roles
```

We'd have to update the message to indicate which role is invalid / abstract, and report multiples. (e.g. "abstract roles 'select', and 'landmark' cannot be directly used".

jnurthen:
Fallback roles are pretty well supported except on IE. For those of us writing products which don't support IE is there any chance we could have a way of allowing fallback roles.
I'm trying to write a meter right now and am using `role="meter progressbar"` causing aXe errors

WilcoFiers:
Getting axe-core to run entirely with support for fallback roles is going to take a while. But I'm pretty confident some time next year this will change. This change will be the first step, by creating a separate check for this, it also makes it possible to turn the check off, separate from the rest of the rule.

I also think it's still worth pointing out that IE still makes up 15% of the WebAIM survey market. That's a significant enough portion that I'm comfortable saying we should flag it for a while still.

jeankaplansky:
Release notes updated 2/04/2020

padmavemulapati:
Instead Aria-role is valid, if displayed some non-confusing accessible name could be better , verified with aria-role.html rule test file, 
![image](https://user-images.githubusercontent.com/47409579/74037502-6c813680-49e4-11ea-888d-2f9c1b66ec2f.png)

PR #1970, opened 2020-01-06
fix(aria-roles): report error for fallback roles

straker:
Closes issue: #1916 

## Reviewer checks


lib/checks/aria/fallbackrole.js
```diff
@@ -0,0 +1 @@
+return virtualNode.attr('role').split(' ').length > 1;
```
WilcoFiers: This should probably use `axe.utils.tokenList`. Otherwise `role="  button  "` would fail the rule.

lib/checks/aria/fallbackrole.json
```diff
@@ -0,0 +1,11 @@
+{
+	"id": "fallbackrole",
+	"evaluate": "fallbackrole.js",
+	"metadata": {
+		"impact": "serious",
+		"messages": {
+			"pass": "Only one role value used",
+			"fail": "Use only one role value, since fallback roles are not widely supported"
```
WilcoFiers:
```suggestion
			"fail": "Use only one role value, since fallback roles are not supported in older browsers"
```

lib/rules/aria-roles.json
```diff
@@ -8,5 +8,5 @@
 	},
 	"all": [],
 	"any": [],
-	"none": ["invalidrole", "abstractrole", "unsupportedrole"]
+	"none": ["fallbackrole", "invalidrole", "abstractrole", "unsupportedrole"]
```
WilcoFiers: You'll also have to update `invalidrole`, have it look at each token in a role attribute rather than the entire string. Otherwise that check will still return an issue for `role="spinbutton button"` for example. That should only fail `fallbackrole`.

lib/checks/aria/invalidrole.js
```diff
@@ -1,3 +1,7 @@
-return !axe.commons.aria.isValidRole(node.getAttribute('role'), {
-	allowAbstract: true
-});
+return (
+	axe.utils.tokenList(virtualNode.attr('role')).filter(role => {
+		return !axe.commons.aria.isValidRole(role, {
+			allowAbstract: true
+		});
+	}).length > 0
```
WilcoFiers: Can you put the output of this into data, and list which roles are invalid? Same thing with abstract role. I know that's not strictly part of this PR, but it'd make the messaging about this a bunch better.

straker:
@jeankaplansky just new remediation messages about fallback roles not being supported

Issue #878, opened 2018-05-03
Region rule results for related nodes are confusing


marcysutton:
From the web-a11y Slack:
> Running aXe on my webpage I am getting an error that the encompassing `html` tag must be in a landmark region. What regions have people put these under? I haven’t seen any examples online where the landmarks encompass the entire page. I have my `main` landmark nested within the `body`

>https://dequeuniversity.com/rules/axe/3.0/region?application=AxeChrome is the relevant error and the element its noting is `<html lang="en">` which is the root element

I've told folks repeatedly to "look at the related nodes", as the HTML tag is throwing them off. Users think the HTML tag itself is the problem, rather than the content contained within. I'm not sure what exactly we can do here, but given I've seen it come up more than once it seemed worth documenting in an issue.

This UX need is somewhat related to https://github.com/dequelabs/axe-core/issues/863 for skip links, but is more generic than that.

WilcoFiers:
I have a proposal for this. Instead of having a single failure on the page, we fail all "outer" elements. That way we'll not end up with hundreds of violations. The way we do this is with an "after" method that takes what is now the related nodes, it filters out all of the nodes `contained` in other nodes, and sets those as violations. That should give us a nice short list.

padmavemulapati:
Verified with content outside of a <main> landmark, then the violation no longer points to html element, working as expected.
![image](https://user-images.githubusercontent.com/47409579/74053178-4b7b0e80-4a01-11ea-9ad3-37099becf95b.png)

PR #1980, opened 2020-01-13
fix(region): return outermost regionless node instead of html

straker:
Have region rule return the outermost container element as the failure node rather than `html`. 

Since the rule returns outer elements, we cannot fail on a specific target easily. So the test cases look at the `data` property to look at which container element was returned. The end result is that all outer container elements will return as violations while every other node on the page will return as passes.

Closes issue: #878 

## Reviewer checks


lib/checks/navigation/region.js
```diff
@@ -47,6 +52,13 @@ function findRegionlessElms(virtualNode) {
 			dom.getElementByReference(virtualNode.actualNode, 'href')) ||
 		!dom.isVisible(node, true)
 	) {
+		// Mark each parent node as having region descendant
+		let vNode = axe.utils.getNodeFromTree(node);
```
WilcoFiers: From the looks of it, you can use `virtualNode`, which is passed into this function.

lib/checks/navigation/region.js
```diff
@@ -62,7 +74,26 @@ function findRegionlessElms(virtualNode) {
 	}
 }
 
-var regionlessNodes = findRegionlessElms(virtualNode);
-this.relatedNodes(regionlessNodes);
+regionlessNodes = findRegionlessElms(virtualNode)
+	// Find first parent marked as having region descendant (or body) and
+	// return the node right before it as the "outer" element
+	.map(node => {
+		let vNode = axe.utils.getNodeFromTree(node);
```
WilcoFiers: Why don't we just have `findRegionlessElms` return vitual nodes, instead of having to look them up again?

test/integration/full/region/region-pass.js
```diff
@@ -13,12 +13,10 @@ describe('region pass test', function() {
 	});
 
 	describe('passes', function() {
-		it('should find one', function() {
-			assert.lengthOf(results.passes[0].nodes, 1);
```
WilcoFiers: Lets test the number of elements we expect from the fixture, just to make sure.

test/integration/full/region/region-fail.js
```diff
@@ -17,22 +17,8 @@ describe('region fail test', function() {
 			assert.lengthOf(results.violations[0].nodes, 1);
 		});
 
-		it('should find html', function() {
-			assert.deepEqual(results.violations[0].nodes[0].target, ['html']);
```
WilcoFiers: Lets test the number of elements we expect from the fixture, just to make sure.

lib/checks/navigation/region.js
```diff
@@ -62,7 +74,26 @@ function findRegionlessElms(virtualNode) {
...
+			vNode = vNode.parent;
+		}
+
+		return vNode.actualNode;
+	})
+	// Remove duplicate containers
+	.filter((node, index, array) => {
+		return array.indexOf(node) === index;
+	});
+this.data(regionlessNodes);
```
WilcoFiers: Why are you putting these into data? Other than for testing, I don't see what this is useful for. And since this will end up in our output data, I don't think we should use it this way.

test/checks/navigation/region.js
```diff
@@ -12,241 +12,263 @@ describe('region', function() {
 		checkContext.reset();
 	});
 
-	it('should return true when all content is inside the region', function() {
+	it('should return no regionless nodes when all content is inside the region', function() {
 		var checkArgs = checkSetup(
 			'<div id="target"><div role="main"><a href="a.html#mainheader">Click Here</a><div><h1 id="mainheader" tabindex="0">Introduction</h1></div></div></div>'
 		);
 
-		assert.isTrue(checks.region.evaluate.apply(checkContext, checkArgs));
```
WilcoFiers: I think we should keep these outcome asserts.

test/integration/full/region/region-pass.js
straker: The problem I had with that is that it was inconsistent. You can see from this test that the [number failed in the unix test](https://app.circleci.com/jobs/github/dequelabs/axe-core/11652) but will pass in localhost.

image

Screen Shot 2020-01-24 at 11 45 55 AM

test/checks/navigation/region.js
```diff
@@ -1,9 +1,10 @@
-describe('region', function() {
+describe.only('region', function() {
```
WilcoFiers:
```suggestion
describe('region', function() {
```

Issue #1482, opened 2019-04-08
aria-hidden-focus trigger false positive


luarmr:
Expectation: I expect `aria-hidden-focus` only show as an error when the content is in the viewport.

Actual: `aria-hidden-focus` is showing as an error in elements that are outside the viewport (intentionally not showing in the screen) or behind another layer. 

Motivation: I am getting this error in carousels or modal window
for example:
https://react-bootstrap.github.io/components/modal/#modals-live
or code like (nonsense code but I think made a point): 
https://codepen.io/luarmr/pen/ROoMWQ
 

<pre><code>
axe-core version: 3.2.2
axe-puppeteer version:1.0.0 (but the relevant is axe-core)
</code></pre>

WilcoFiers:
@luarmr I am not sure why you would want axe-core to exclude visually hidden content. If you have a link or button or something else focusable inside an aria-hidden=true region, that element can still be focused, and it can be pulled up in things like link lists. When you put an element in an aria-hidden=true region, its accessibility name is removed. This results in something that's effectively a link/button/whatever without a name.

ktsangop:
I am facing a similar problem, working with ng-bootstrap library
https://ng-bootstrap.github.io/#/components/modal/examples

I believe that when a modal has the `role="dialog"` and `aria-modal="true"` then AXE could omit showing this error because those attributes explicitly set the modal as the only area of focus.

https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html
https://www.w3.org/TR/wai-aria-practices/#dialog_roles_states_props

The fact that in ng-bootstrap modals, removing the `aria-hidden="true"` from the body/root app element, is fixing the AXE error, without breaking screen reader support, shows that this could be an issue of ng-bootstrap.
But, since not all browsers/screen readers currently support the above attributes, i believe that `aria-hidden="true"` is intentionally left, to prevent screen readers from focusing out of the dialog.

In fact another related example here with the same error, which is i believe from AXE's own documentation website :

https://dequeuniversity.com/library/aria/popups-dialogs/sf-simple-dialog

**_FYI:_** The react-bootstrap library indeed seems to have another bug. Except for adding `aria-hidden="true"` to the body, it also adds it to the modal container, along with `role="dialog"` and `aria-modal="true"`. That's why AXE shows 2 errors. On the second case i think that AXE is correct.

[...]

PR #1995, opened 2020-01-22
fix(aria-hidden-focus): mark as needs review if a modal is open

straker:
The rule will now return as needs review if there are focusable elements present while what we believe to be a modal is open. On the react example, it will now return as needs review rather than failing when the modal is open that traps focus https://react-bootstrap.github.io/components/modal/#modals-live.

Closes issue: #1482

## Reviewer checks


lib/commons/dom/is-modal-open.js
```diff
@@ -0,0 +1,108 @@
...
+	let fullWidthElement;
+	stacks.forEach(stack => {
+		// the modal element may not be the top-most element of the stack due to
+		// how it's written, so we'll look for it in the first half
+		for (let i = 0; i < stack.length / 2; i++) {
+			const style = window.getComputedStyle(stack[i]);
+			const modal =
+				parseInt(style.width, 10) >= percentWidth &&
+				parseInt(style.height, 10) >= percentHeight &&
+				(style.position === 'absolute' || style.position === 'fixed');
```
WilcoFiers: Should we look at pointer-events here too? If this has pointer-events:none on it it's likely not an overlay.

lib/commons/dom/is-modal-open.js
```diff
@@ -0,0 +1,108 @@
...
+	// four corners and one from the middle of the viewport (total of 5). if each
+	// stack contains an element whose width/height is >= 75% of the screen as
+	// the top most element, we found a "could be a modal"
+	const viewport = dom.getViewportSize(window);
+	const percentWidth = viewport.width * modalPercent;
+	const percentHeight = viewport.height * modalPercent;
+	const x = (viewport.width - percentWidth) / 2;
+	const y = (viewport.height - percentHeight) / 2;
+
+	const points = [
```
WilcoFiers: Wondering if we shouldn't also look at the center point.

lib/commons/dom/is-modal-open.js
```diff
@@ -0,0 +1,108 @@
...
+
+	const stacks = points.map(point => {
+		return document.elementsFromPoint(point.x, point.y);
+	});
+
+	let fullWidthElement;
+	stacks.forEach(stack => {
+		// the modal element may not be the top-most element of the stack due to
+		// how it's written, so we'll look for it in the first half
+		for (let i = 0; i < stack.length / 2; i++) {
```
WilcoFiers: This "half the stack" thing feels pretty arbitrary. Why did you choose this?

test/commons/dom/is-modal-open.js
```diff
@@ -0,0 +1,96 @@
...
+	);
+
+	it('returns true if there is a visible absolutely positioned element with >= 75% width/height', function() {
+		fixtureSetup(
+			'<div style="position: absolute; top: 0; bottom: 0; left: 0; right: 0">Modal</div>'
+		);
+		assert.isTrue(isModalOpen());
+	});
+
+	it('returns true if there is a visible absolutely positioned element with >= 75% width/height', function() {
```
WilcoFiers: This test is a duplicate of the one above it.

test/commons/dom/is-modal-open.js
```diff
@@ -0,0 +1,96 @@
...
+		fixtureSetup('<div>Modal</div>');
+		assert.isUndefined(isModalOpen());
+	});
+
+	it('returns undefined if there is a hidden element with role=dialog', function() {
+		fixtureSetup('<div role="dialog" style="display: none">Modal</div>');
+		assert.isUndefined(isModalOpen());
+	});
+
+	it('returns undefined if there is a hidden element with aria-modal=undefined', function() {
```
WilcoFiers:
```suggestion
	it('returns undefined if there is a hidden element with aria-modal=true', function() {
```

lib/commons/dom/is-modal-open.js
```diff
@@ -0,0 +1,108 @@
...
+	if (
+		fullWidthElement &&
+		stacks.every(stack => Array.from(stack).includes(fullWidthElement))
+	) {
+		axe._cache.set('isModalOpen', true);
+		return true;
+	}
+
+	axe._cache.set('isModalOpen', undefined);
+	return undefined;
```
WilcoFiers: Why is this returning true | undefined instead of just true | false?

straker: I thought about returning false but I didn't want it to seem like the test was accurate and could say "false" meant a modal was not open. Undefined kinda help make the check feel more uncertain that we're not even sure if a modal is open or not.

Issue #1992, opened 2020-01-21
ARIA live region triggers region rule violation


madalynrose:
## Description: 
Gatsby [updated](https://github.com/gatsbyjs/gatsby/pull/19290) navigation to announce when client-side routes are changed. This is accomplished through an ARIA live region with `role="alert"` and `aria-live="assertive"`. Based on #1073 this should not cause any accessibility violations. However, technologies that rely on axe-core, including [cypress-axe](https://github.com/avanslaars/cypress-axe) and the aXe browser extension are reporting a violation on the element. An [issue](https://github.com/gatsbyjs/gatsby/issues/20697) in Gatsby was opened but I believe this is an issue with aXe itself.

### Expectation: 
No accessibility violations are found

### Actual: 
aXe reports that the region rule "Some page content is not contained by landmarks" has been violated after navigation has occurred

### Motivation: 
This live region is intended to increase the accessibility of client-side applications. Adding a landmark to this live region does not make semantic sense because it does not have user impact. This rule should allow this element to exist outside of landmarks. 

axe-core version: 3.4.0

This can be reproduced on Gatsby's [pages](www.gatsbyjs.org) by running the aXe extension.

straker:
I am able to reproduce this. I think we just need to rearrange an if statement in our code to get this to work. We appear to assume that an aria-live region will not have an explicit role set and so fail before we check if the aria-live attribute is set https://github.com/dequelabs/axe-core/blob/develop/lib/checks/navigation/region.js#L15-L21

padmavemulapati:
on making testfile,  `<div aria-live="assertive" role="alert" id="target"><p>This is random content.</p></div>` , not finding any voilations

![image](https://user-images.githubusercontent.com/47409579/73369613-e0cb2400-42d8-11ea-80ea-336d1a2a80d1.png)

PR #1999, opened 2020-01-27
fix(region-rule): allow live regions with explicit roles

madalynrose:
The region rule should always allow content in an ARIA live region. The current code fails the rule based on role before it gets to the `aria-live` check. This fix moves that check up so a region with `role="alert"` and `aria-live` is allowed.

Closes issue: #1992 

## Reviewer checks


straker:
Awesome, thanks for the pr. Just want @WilcoFiers to take a look to make sure I didn't miss anything

Issue #1985, opened 2020-01-14
color-contrast takes >200s


connorjclark:
URL: https://james-muriithi.github.io/jCheckBox/

Expectation: `color-contrast` should finish in ~2s (duration of the second-longest check for the same page).

Actual: ~230s

Motivation: Prevent Lighthouse run from failing to finish.

See: https://github.com/GoogleChrome/lighthouse/issues/10183#issuecomment-573986202

<pre><code>axe-core version: 3.3.0
Chrome 79
</code></pre>

Previous issues: https://github.com/dequelabs/axe-core/issues/716

___

![image](https://user-images.githubusercontent.com/4071474/72380939-545f1580-36cb-11ea-9215-e479b0fbaf9c.png)

![image](https://user-images.githubusercontent.com/4071474/72381523-62616600-36cc-11ea-945e-e805860f4365.png)

[Profile-20200114T124246.json.zip](https://github.com/dequelabs/axe-core/files/4061366/Profile-20200114T124246.json.zip)

straker:
Thanks for reporting this. We're actually working on a huge improvement to the color-contrast rule right now. I tested the page using the new code and it can run color-contrast in 10seconds. That's a bit slower than running it on a page with 50k+ nodes, so I'll do some digging to see if I can boost the performance some more.

connorjclark:
Super great! Thanks for the quick response :)

straker:
Alright, I dug into the problem and it turns out that issue is the large syntax highlighted code blocks. When we do color contrast, we look at the `clientRects` of an element and find each element stack for each rect.

Well it turns out that the large code block has 1981 client rects and so we take a very long time to process this. Even in the latest code it takes 7 seconds to loop over each rect and get all the elements in the stacks. The other code blocks have between 26 and 90 client rects.

We should probably find some way to filter out elements like these since they don't need to be checked for color contrast, just their children.

padmavemulapati:
I am seeing , the time taking is between 1.6 to 1.74 secs.

PR #2003, opened 2020-01-28
fix(color-contrast): improve speed and accuracy of code blocks with syntax highlighting

straker:
Code blocks with syntax highlighting could have large amounts of client rects, which would then cause it to take a few seconds each to process. Code blocks could also cause us lots of incomplete results with "partially obscured by another element." 

The code changes fix the speed by focusing only on the client rects of the text nodes of the element rather than all client rects. Since each of the child nodes that have text nodes will be checked individually, there is no need for the main parent to also check those nodes as well. This reduced the large code block from 2000+ client rects to just <20, taking color contrast from running in 10+ seconds to <1 seconds. 

It also adjusts the collision check to account for elements which lie exactly on the middle of the element being checked. This solved all the incomplete issues and they now return as proper pass/fail.

Closes issue: #1985

## Reviewer checks

### DIFF
--- a/lib/commons/dom/get-element-stack.js
+++ b/lib/commons/dom/get-element-stack.js
@@ -548,7 +548,9 @@ dom.getTextElementStack = function(node) {
 				const rect = rects[i];
 
 				// filter out 0 width and height rects (newline characters)
-				if (rect.width && rect.height) {
+				// ie11 has newline characters return 0.00998, so we'll say if the
+				// line is < 1 it shouldn't be counted
+				if (rect.width >= 1 && rect.height >= 1) {
 					clientRects.push(rect);
 				}
 			}
--- a/lib/commons/dom/get-element-stack.js
+++ b/lib/commons/dom/get-element-stack.js
@@ -337,7 +337,10 @@ function addNodeToGrid(grid, vNode) {
 
 			for (let col = startCol; col <= endCol; col++) {
 				grid.cells[row][col] = grid.cells[row][col] || [];
-				grid.cells[row][col].push(vNode);
+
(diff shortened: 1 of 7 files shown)


test/commons/color/get-foreground-color.js
```diff
@@ -14,8 +14,8 @@ describe('color.getForegroundColor', function() {
 
 	it('should return the blended color if it has alpha set', function() {
 		fixture.innerHTML =
-			'<div style="height: 40px; width: 30px; background-color: #800000;">' +
-			'<div id="target" style="height: 20px; width: 15px; color: rgba(0, 0, 128, 0.5);' +
+			'<div style="height: 40px; background-color: #800000;">' +
```
straker: I had to update these tests as my code changes made them more accurate than before. Before my changes, we would test only the first line of this fixture. However, if we look at the result of the fixture, we can see that there should be 3 background colors and 3 foreground colors, but we only calculate 1.

image

So I updated the tests to just focus on the single color with alpha/opacity.

lib/commons/dom/get-element-stack.js
```diff
@@ -506,13 +509,13 @@ dom.getElementStack = function(node) {
...
- * Return all elements that are at the center of each client rect of the passed in node.
- * @method getClientElementStack
+ * Return all elements that are at the center of each text client rect of the passed in node.
+ * @method getTextElementStack
  * @memberof axe.commons.dom
  * @param {Node} node
  * @return {Array<Node[]>}
  */
-dom.getClientElementStack = function(node) {
+dom.getTextElementStack = function(node) {
```
WilcoFiers: Renaming a commons is a breaking change. We can't do that.

Issue #1993, opened 2020-01-22
color contrast nonMbp charset isn't global and isn't sanitised


WilcoFiers:
If I pass something like this into the color contrast rule, it is tested because only the first char is stripped out, and the spaces aren't removed:

```html
<div style="color: #ccc">
  &#x20A0;
  &#x20A1;
  &#x20A2;
  &#x20A3;
</div>
```

**Expected**: This element is incomplete for color-contrast
**Actual**: This element fails color-contrast

jeankaplansky:
Added to release notes 2/04/2020;

https://docs.google.com/document/d/14PI0e3X9g6CBWtl86HReHrRxl-_zapZBE8HaKoOHrmE/edit#

padmavemulapati:
non BMP chanrecter color-contrast test still seeing voilations, not needs review.

straker:
Confirmed. The checks options are not be passed in due to a [misconfiguration of the check.json](https://github.com/dequelabs/axe-core/blob/develop/lib/checks/color/color-contrast.json#L6-L10).

straker:
@padmavemulapati Should be fixed now in the latest axe-core

padmavemulapati:
will verify the same with axe and attest extensions once axe-core3.5 deployed in devTools.

PR #2005, opened 2020-01-30
fix(color-contrast): mark as needs review for text that contains only non-BMP characters

straker:
The test for non-BMP characters was actually broken as it was returning `undefined` because of short text content and not because of the non-BMP characters. Also fixed the non-BMP regex to globally ignore characters.

Closes issue: #1993

## Reviewer checks

### DIFF
--- a/lib/checks/color/color-contrast.json
+++ b/lib/checks/color/color-contrast.json
@@ -23,7 +23,7 @@
 				"outsideViewport": "Element's background color could not be determined because it's outside the viewport",
 				"equalRatio": "Element has a 1:1 contrast ratio with the background",
 				"shortTextContent": "Element content is too short to determine if it is actual text content",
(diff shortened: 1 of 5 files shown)


lib/checks/color/color-contrast.json
```diff
@@ -22,7 +22,8 @@
 				"elmPartiallyObscuring": "Element's background color could not be determined because it partially overlaps other elements",
 				"outsideViewport": "Element's background color could not be determined because it's outside the viewport",
 				"equalRatio": "Element has a 1:1 contrast ratio with the background",
-				"shortTextContent": "Element content is too short to determine if it is actual text content"
+				"shortTextContent": "Element content is too short to determine if it is actual text content",
+				"nonBmp": "Element content contains only non-BMP unicode characters"
```
WilcoFiers:
```suggestion
				"nonBmp": "Element content contains only non-text characters"
```

PR #2009, opened 2020-02-03
fix(preload): reject promise `axe.utils.preload` when XHR fails

jkodu:
Preload assets did not reject with `axios.get` failed.
This meant that before this fix, axe.run waited till preload timeout kicked in.

Closes issue:
- NA

## Reviewer checks

### DIFF
--- a/lib/core/utils/preload.js
+++ b/lib/core/utils/preload.js
@@ -66,7 +66,7 @@ axe.utils.getPreloadConfig = function getPreloadConfig(options) {
 	if (
 		options.preload.timeout &&
 		typeof options.preload.timeout === 'number' &&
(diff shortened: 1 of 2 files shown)


lib/core/utils/preload.js
```diff
@@ -66,7 +66,7 @@ axe.utils.getPreloadConfig = function getPreloadConfig(options) {
 	if (
 		options.preload.timeout &&
 		typeof options.preload.timeout === 'number' &&
-		!Number.isNaN(options.preload.timeout)
+		!isNaN(options.preload.timeout)
```
straker: `Number.isNaN` does not exist in IE11, and was causing the timeout test to fail, but not because of the timeout

Issue #2019, opened 2020-02-04
[Docs] Update rule `label-content-name-mismatch` to filter by widget roles


jkodu:
Expectation: 
Below example should be `inapplicable`.
```
<div role="tooltip" aria-label="OK">Next</div>
```

Actual: Axe fails the above example

Motivation: 
Needed for ACT Rules implementation. Rule in question: https://act-rules.github.io/rules/2ee8b8

<pre><code>
axe-core version: X.Y.Z
axe-webdriver, extension or other integration version: X.Y.Z

Browser and Assistive Technology versions

For Tooling issues:
- Node version: XX  
- Platform:  
</code></pre>

straker:
We already filter by widget roles, but `tooltip` is included in the lookup tale in `rolesOfType` (it shouldn't be)

jkodu:
PR has been merged. Closing issue.

padmavemulapati:
Verified with latest axe-coconut(v4.3.0.06000), the test file
` <a id="inapplicable1" aria-label="OK">Next</a>
<input id="inapplicable2" type="email" aria-label="E-mail" value="Contact" />
<div id="inapplicable3" role="tooltip" aria-label="OK">Ok</div>
<div id="inapplicable4" role="feed" aria-label="RSS Feed">RSS Feed</div>
<div id="inapplicable5" role="marquee" aria-label="Foo">Foo Restaurant</div>`
giving inapplicable .
![image](https://user-images.githubusercontent.com/47409579/75665780-b6cf9d80-5c9a-11ea-9eec-1d4d487bc3e4.png)

jkodu:
No documentation necessary. Closing issue. QA verified.

Issue #1930, opened 2019-12-06
Consider having a size budget for axe-core


straker:
From the discussion on https://github.com/GoogleChrome/lighthouse/pull/10056, axe 3.4.0 added 60KB minified to our codebase, mostly from the addition of [core-js Unit32Array polyfills](https://github.com/GoogleChrome/lighthouse/pull/10056#issuecomment-562360821) needed for the ligature icon detection. This pushed lighthouse over their size budget, causing them to fail their build. 

For 4.0.0 we should consider introducing a size budget ourselves so we don't continue to increase in size (as we have done as shown by @patrickhulce chart)

![image](https://user-images.githubusercontent.com/2433219/70284829-af275800-1782-11ea-8005-906d448bdc8b.png)

scurker:
@straker I've put some of this in place already with the extension, so hit me up if you want to pair on this for axe-core.

WilcoFiers:
Dropped axe-core's filesize from 138kb in 4.0 to 108kb in 4.1! For 4.2 we should put tests in place to enforce a maximum size.

dylanb:
Can we add a performance budget too? Some of the techniques for reducing size, may increase execution time.

straker:
I like the idea of a performance budget. We'll probably want to separate our metrics into different size pages or buckets though as some rules perform worse on very large pages but this may be acceptable for the rule itself.

PR #2022, opened 2020-02-05
fix(label-content-name-mismatch): ignore non `widget` aria role(s) & do not use deprecated `lookupTable.rolesOfType`

jkodu:
Updated lookUpTable with roles of type `widget`.

Closes issue:
- https://github.com/dequelabs/axe-core/issues/2019

## Reviewer checks


straker:
Thanks for posting the new sizes. With axe-core 4.0 being the next phase, I want to introduce a size budget (#1930). Adding 100 KB of / 60 KB  (min) to axe-core seems a bit much just to fix a hard-coded list we have already.

I know we've talked about aria-query before and using it in place of our lookup table entirely. @WilcoFiers  is that something we could do for 4.0?

lib/core/imports/index.js
```diff
@@ -47,5 +47,6 @@ axe.imports = {
 	CssSelectorParser: require('css-selector-parser').CssSelectorParser,
 	doT: require('@deque/dot'),
 	emojiRegexText: require('emoji-regex'),
-	memoize: require('memoizee')
+	memoize: require('memoizee'),
+	ariaQuery: require('aria-query')
```
WilcoFiers: I don't think we should pull aria-query into axe-core at this point. There's a much simpler fix for this. Please don't overthink it.

stephenmathieson:
I agree with @WilcoFiers and @straker. Adding 100k just to fix a single bug probably isn't worth it.

lib/commons/aria/index.js
```diff
@@ -2387,43 +2387,54 @@ lookupTable.evaluateRoleForElement = {
...
  * 	"alert", "alertdialog", "button", "checkbox", "dialog", "gridcell", "link", "log", "marquee", "menuitem", "menuitemcheckbox",
  * 	"menuitemradio", "option", "progressbar", "radio", "scrollbar", "searchbox", "slider", "spinbutton", "status", "switch", "tab", "tabpanel",
  * 	"textbox", "timer", "tooltip", "treeitem"
  * ]
- * There are some differences against specs, hence the below listing was made
+ * There are some differences against specs, hence the below listing was made by cross referring with `aria-query`
  */
 lookupTable.rolesOfType = {
 	widget: [
+		'command',
```
straker: command is an abstract role, not a widget role. Same for a lot of the ones you've added. Axe doesn't allow abstract roles on elements so we don't need to add them here.

Also don't we need to change `tooltip` in the a `lookupTable.role` to not be a widget role? I'm not sure how we differentiate between `widget` roles and `rolesOfType.widget` (such as `alertdialog` being a widget in the lookupTable but not in this list)

jkodu: Thanks @straker. 

The reason I did not use the above code snippet is our lookupTable is out of date & a small change to accommodate something like this usually breaks other tests. 

We have had conversations to update the lookupTable in the past. I used aria-query to generate this list of roles with a widget role type.

Let me discuss with @WilcoFiers for the right way forward to be lookupTable.

jkodu:
Ended up removing the array `rolesOfType.widget` and changed the type of a role in the lookup table.

History points to this [snapshot](https://github.com/dequelabs/axe-core/commit/a4255da17c736bfa740790e61f8f69635866d4da#diff-ec67bb6113bfd9a900ee27ecef942f74R2372) when this listing was added. From memory I believe there was an ask to not touch/ alter the lookup table when this PR was authored.

straker:
@WilcoFiers probably needs to approve removing the list as he'll have the most knowledge of why it was created in the first place.

lib/commons/aria/index.js
```diff
@@ -2385,45 +2385,3 @@ lookupTable.evaluateRoleForElement = {
...
- * The current lookupTable.role['widget'] widget, yeilds
- * ->
- * [
- * 	"alert", "alertdialog", "button", "checkbox", "dialog", "gridcell", "link", "log", "marquee", "menuitem", "menuitemcheckbox",
- * 	"menuitemradio", "option", "progressbar", "radio", "scrollbar", "searchbox", "slider", "spinbutton", "status", "switch", "tab", "tabpanel",
- * 	"textbox", "timer", "tooltip", "treeitem"
- * ]
- * There are some differences against specs, hence the below listing was made
- */
-lookupTable.rolesOfType = {
```
WilcoFiers: Taking this out could break custom rules. Please leave it in, label it as deprecated, and make sure to include the word "deprecated" in the title of the PR so it appears in the changelog.

Issue #2017, opened 2020-02-04
Rule `identical-links-same-purpose` should fail below example


jkodu:
Expectation: 
Below example should fail
```
<html lang="en">
	<svg viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
		<a href="http://facebook.com" aria-label="Follow us">
			<circle cx="50" cy="40" r="35" />
		</a>

		<a href="http://twitter.com">
			<text x="50" y="90" text-anchor="middle">
				Follow us
			</text>
		</a>
	</svg>
</html>
```

Actual: Returns as `inapplicable`

Motivation: Needed for implementation in ACT Rules - https://act-rules.github.io/rules/b20e66

<pre><code>
axe-core version: X.Y.Z
axe-webdriver, extension or other integration version: X.Y.Z

Browser and Assistive Technology versions

For Tooling issues:
- Node version: XX  
- Platform:  
</code></pre>

straker:
Anchors inside SVGs are different than normal anchors so don't have the `protocol` and `pathname` properties (required to get the `getPathnameOrFilename`)

padmavemulapati:
Seeing the code change impact..
![image](https://user-images.githubusercontent.com/47409579/76214148-28c35c00-6232-11ea-96e9-3a7a3613437d.png)

jkodu:
Nothing to document here. Closing issue.

PR #2025, opened 2020-02-06
fix(commons): handle node(s) contained by SVG document when deciphering url properties

jkodu:
Nodes contained inside SVG document were a special case when constructing url properties.

Closes issue:
- https://github.com/dequelabs/axe-core/issues/2017

## Reviewer checks


jkodu:
After debugging in IE/ Windows machine, this is going down a rabbit hole. So will re-work this PR in a different approach

Issue #2031, opened 2020-02-10
Parse font-weight values as number in color-contrast check


takenspc:
Expectation: variable fonts offer the ability to use any weight such 750 or 850 which should be considered as bold. I expected that any weight greater than or equal to 700 are considered as bold in color-contrast check.

Actual: axe-core reports following error when `font-weight` is `750`.

```html
<p style="color: gray; background: white; font-size: 14pt; font-weight: 750">Hello</p>
```

> Element has insufficient color contrast of 3.94 (foreground color: #808080, background color: #ffffff, font size: 14.0pt (18.6667px), font weight: normal). Expected contrast ratio of 4.5:1

While it doesn't report any errors if `font-weight` is `700`.

```html
<p style="color: gray; background: white; font-size: 14pt; font-weight: 700">Hello</p>
```

Motivation: we would like to change

1. Parse `font-weight` values as number in color-contrast check
2. Make color-contrast check consider any weight greater than or equal to 700 are bold (current threshold is 600)

As far as I know, "semi" of semi-bold (600) means “partially" so that it is not considered as bold ("partially passed test" is not considered as "passed test"). If we need more discussions, I can split 2 as another issue. 

<pre><code>
axe-core version: 3.4.2
custom auditing script using puppeteer 2.1.1

- Node version: v12.15.0
- Platform:  Windows 10 version 1903
</code></pre>

straker:
Thanks for letting us know. We just released 3.5 so this probably won't get into a release for a month or two.

takenspc:
Thank you for your triaging.

>  We just released 3.5 so this probably won't get into a release for a month or two.

It is OK for me. I have a workaround for this.

padmavemulapati:
Verified on latest axe-coconut(
with test file - 
`<div style="background-color: #1c1f21; color: #ffffff;  opacity: 0.2 ">
		<div style="color: gray; background-color: white; font-size: 14pt; font-weight: 600" id="target"> + ` 
Not seeing any color-contrast issues - replaced font-wright: bold with font-weight:'numbervalue'...
![image](https://user-images.githubusercontent.com/47409579/74656494-3bf38680-51b4-11ea-8e8c-ae7734f92b36.png)

PR #2040, opened 2020-02-12
fix(color-contrast): parse font-weight value as number (#2031)

takenspc:
The computed values of `font-weight` are numeric string so that we don't need to check keywords such as `bold`.

The `font-weight` values are continuous. They can be `123` or `123.4` so that we should parse them as number.

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Fonts/Variable_Fonts_Guide

<< Describe the changes >>

Closes issue: #2031 

## Reviewer checks


WilcoFiers:
@tabushaban Thank you for the contribution!

PR #2041, opened 2020-02-12
fix(sri-history): correct SRI for various axe files

WilcoFiers:
Explanation for why this went wrong is in #2042.

## Reviewer checks

### DIFF
--- a/sri-history.json
+++ b/sri-history.json
@@ -140,35 +140,35 @@
 		"axe.min.js": "sha256-c+LGvM2VcJVUXwaqzmDeKLDBMJSZgKUhnW9GK3JkZ5o="
 	},
 	"3.3.0": {
-		"axe.js": "sha256-IPwXSv4YYXnxEsNQj30armipJWS7+ZFgF+UbQnHZEJc=",
-		"axe.min.js": "sha256-K5dF2aKG6IIZpZZDQrZrJi0XDDjz0mMbHcAkVtj5+ic="
+		"axe.js": "sha256-9qKFcrJohPVSRdWg+nYMf1cnxUz00rojnrcDbRpofP4=",
+		"axe.min.js": "sha256-t2QDE4+vMfgX52JfOztdvXHLxDJLc9Gc1O583kc4l/s="
 	},
 	"3.3.1": {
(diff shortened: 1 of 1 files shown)


stephenmathieson:
Why were these incorrect?

straker:
These were auto generated from the release script, so fixing them here won't fix the problem for the next release either. We need to figure out what is wrong with the script that generates these if it's producing invalid SRIs

sri-history.json
```diff
@@ -140,35 +140,35 @@
...
+		"axe.min.js": "sha256-z1eK+dbxVx44jv/XqW4xYJlyA9T9zXuxMlJbNJfTvOg="
 	},
 	"3.4.2": {
 		"axe.js": "sha256-1O2tV4TBC/8pYjxbIwj192fUX1SJvsjn8KQH3PdJUuU=",
 		"axe.min.js": "sha256-XXgQWSZNsMKcqEQnEU2BDkIfCzV6gCjOjrwnrzioYm4="
 	},
 	"3.5.0": {
-		"axe.js": "sha256-MPhNDpZVnZlXBeRKLHjKbiN4Lz4RA/Cb6FHqSsOQR/g=",
-		"axe.min.js": "sha256-Snr0xNyrSfcrWKoPPk/5335i7Aj5nrflncNblRZMMqM="
+		"axe.js": "sha256-w/zYk9v7UUawn15gNmZHGrh8WdzCI1bSxmRM2N7Uk4o=",
```
straker: This number is different than the one I generate when on the #2042 branch

straker:
```js
"3.5.0": {
		"axe.js": "sha256-+7Q5f49NA9ko9730M1oM2Tj3aXXwIvx5qaG92uETNp8=",
		"axe.min.js": "sha256-H4Py7XBch38wDECBdMX7eHsyHuTHUiz6DPbeFOoEzzs="
	}
```

WilcoFiers: Running npm ci might not be correct either. I got these SRIs from https://cdnjs.com/libraries/axe-core. I've used it before, and other SRIs seem correct, so I trust it. If you want to know for sure, pull down axe.js from NPM, put it in the repo and validate that one.

straker: You are correct. Have to use the exact axe-core from npm to build the same hash.

PR #2042, opened 2020-02-12
chore(ci): prevent publishing invalid SRIs

WilcoFiers:
An SRI is essentially a hash of axe.js. As of version 3.1 (I think) , we are bundling a few external dependencies into axe-core. This hash is created when we create the release branch, and put into sri-history.json. If any of the dependencies is different when we build locally for the release branch, from when we build in CI right before executing `npm publish`, axe-core will not be identical, and the SRI in sri-history.json will be incorrect.

To fix this, this PR does:
1. add a package-lock file, and use `npm ci` to ensure the server build has the dependencies in the lock file
2. make `npm ci` part of the `npm run release` script, to ensure dependencies are in sync with package-lock
3. Add `npm run sri-validate` before releasing to test the sri matches what is in sri-history.json

## Reviewer checks

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,7 +39,7 @@ jobs:
       - checkout
       - <<: *set_npm_auth
       - <<: *restore_dependency_cache_unix
-      - run: npm install
+      - run: npm ci
       - save_cache:
           key: v8-cache-unix-{{ checksum "package.json" }}
           paths:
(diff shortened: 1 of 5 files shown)


straker:
We should merge #2041 first so the sri-validate can work locally

straker:
Nm, won't work anyway. merge away

Issue #1886, opened 2019-11-07
contrast issue on buttons with negative text-indent


WilcoFiers:
This example throws a false positive, it shouldn't:

```html
<link href="https://stackpath.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" rel="stylesheet" />
<style>
	.form-submit {
		background: #d1c2a6;
	}
	.form-submit > input {
		height: 30px;
		background: transparent;
		margin-left: -18px;
		text-indent: -9999px;
		color: white;
		width: 30px;
	}
</style>
<span class="fa fa-search form-submit">
	<input type="submit" value="Search" />
</span>
```

Problem seems to have something to do with the combination of `input[type=button]` and `text-indent`. The problem goes away when using a `button` element.

padmavemulapati:
Verified wit latest axe-coconut, didnot find any voilations on color-contrast for different input types and text-indent.

![image](https://user-images.githubusercontent.com/47409579/75784956-0096b180-5d89-11ea-88ad-311ce05e9d84.png)

straker:
No docs needed

PR #2044, opened 2020-02-12
fix(color-contrast): ignore form elements that move text outside of node using text-indent

straker:
Closes issue: #1886 

## Reviewer checks

### DIFF
--- a/lib/rules/color-contrast-matches.js
+++ b/lib/rules/color-contrast-matches.js
@@ -14,23 +14,25 @@ if (
 // the text indent has not been changed and moved the text away from the
 // control
 const formElements = ['INPUT', 'SELECT', 'TEXTAREA'];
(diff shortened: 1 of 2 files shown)


lib/rules/color-contrast-matches.js
```diff
@@ -10,6 +10,30 @@ if (
 	return false;
 }
 
+// form elements that don't have direct child text nodes need to check that
+// the text indent has not been changed and moved the text away from the
+// control
+const formElements = ['INPUT', 'SELECT', 'TEXTAREA'];
+const style = window.getComputedStyle(node);
```
WilcoFiers: Minor point, but I'm thinking we should only call getComputedStyle after we've tested nodeName, to minimise the number of style computations we're doing here.

Issue #2012, opened 2020-02-04
Update rule `aria-required-parent`


jeeyyy:
Expectation: 
For examples below, the rule should fail because required parent role is not correct.

```
<div role="list">
	<div role="tabpanel">
		<div role="listitem">List item 1</div>
	</div>
</div>
```

```
<div role="list" aria-owns="item"></div>

<div id="host"></div>

<script>
	const host = document.querySelector('#host')
	const root = host.attachShadow({ mode: 'open' })
	root.innerHTML = '<div id="item" role="listitem">List item 1</div>'
</script>
```

Actual:
The rule only looks for parent with an allowed role, without intercepting nodes in between which can have over-ridding roles.

Motivation:
Required for implementation of rule in ACT Rule - https://act-rules.github.io/rules/ff89c9

<pre><code>
axe-core version: X.Y.Z
axe-webdriver, extension or other integration version: X.Y.Z

Browser and Assistive Technology versions

For Tooling issues:
- Node version: XX  
- Platform:  
</code></pre>

straker:
The check only uses `findUp` to find the ancestor role but doesn't check if there are any roles in between the ancestor and the current node.

jkodu:
Note: 

This may also need an update to the rule `aria-required-children` for the inverse scenario of the above example.

jkodu:
Debugged `aria-required-children` rule and have created a new issue for this - https://github.com/dequelabs/axe-core/issues/2076.

As the approach to solve this between these two rules will vary.

straker:
This should be resolved now with latest axe-core.

padmavemulapati:
aria-required-parent rule should fail when there is an intermediate role between the child and parent
`<div role="list"><div role="tabpanel"><p role="listitem" id="target">Nothing here.</p></div></div>`
aria-required-parent rule should pass when intermediate node is role=presentation
`<div role="list"><div role="presentation"><p role="listitem" id="target">Nothing here.</p></div></div>`
aria-required-parent rule should pass when intermediate node is role=none'
`<div role="list"><div role="none"><p role="listitem" id="target">Nothing here.</p></div></div>`
aria-required-parent rule should pass when intermediate node is not owned by parent
`<div role="list" aria-owns="target"><div role="navigation"><p role="listitem" id="target">Nothing here.</p></div></div>`

![image](https://user-images.githubusercontent.com/47409579/93731789-2f248e80-fbec-11ea-8bed-277626af39ac.png)

[...]

PR #2046, opened 2020-02-13
[WIP] fix(required-parent): validate parent nodes for desired context role

jkodu:
Closes issue:
- https://github.com/dequelabs/axe-core/issues/2012

## Reviewer checks


lib/checks/aria/required-parent.js
```diff
@@ -62,26 +79,26 @@ function getAriaOwners(element) {
...
-	return true;
-}
+/**
+ * Recursively attempt to find parent with a desired role
+ * @param {HTMLElement} elm element
+ * @param {String} role aria role
+ * @returns {Boolean}
+ */
+function hasParentWithRole(elm, role) {
+	const parent = dom.getComposedParent(elm);
```
straker: Wouldn't it be easier to pass `virtualNode` into the function? Then you wouldn't have to call `getComposedParent` and could instead use `virtalNode.parent` which is already the composed parent.

lib/checks/aria/required-parent.js
```diff
@@ -62,26 +79,26 @@ function getAriaOwners(element) {
...
+	if (!parent) {
+		return false;
+	}
 
-var owners = getAriaOwners(node);
+	const parentRole = aria.getRole(parent, { noImplicit: true });
```
straker: Implicit roles should probably be considered. Current axe does not fail this:

```
<ul>
  <div role="listitem">List item 1</div>
</ul>
```

but these changes will.

straker:
Could you add a test ensuring this passes so we don't get regressions?

```html
<ul>
  <div role="listitem">List item 1</div>
</ul>
```

lib/checks/aria/required-parent.js
```diff
@@ -1,5 +1,29 @@
+const { aria, dom } = axe.commons;
```
WilcoFiers: I'd really like us to clean this up... my initial request for this was to create an `aria.getOwnerVirtual()`. I still think we should do this. Here's what this check would look like if we do this, which I think is far easier to read then what we have now:

```js
const role = aria.getRole(virtualNode)
const requiredContext = aria.requiredContext(role);
if (!requiredContext) {
  return true
}

// Accounts for role=presentation, as well as aria-owns
const owner = aria.getOwnerVirtual(virtualNode)
if (!owner) {
  return undefined;
}

const ownerRole = aria.getRole(owner)
if (requiredContext.some((contextRole) => contextRole === ownerRole)) {
  return true;
} else {
  this.relatedNode(owner);
  this.data(requiredContext);
  return false
}
```

jkodu: I would really like to keep this PR scope to fix the bug only & not go down the route of introducing nice to have's and helper function.

I will look into what the `getOwnerVirtual` entails.

lib/commons/aria/get-owner-virtual.js
```diff
@@ -0,0 +1,54 @@
...
+ * @returns {HTMLElement | undefined}
+ */
+function getAriaOwner(vNode) {
+	if (!vNode) {
+		return;
+	}
+
+	const id = vNode.attr('id');
+	if (!id) {
+		return getAriaOwner(vNode.parent);
```
WilcoFiers: aria-owns doesn't inherit descendants. This should just be:
```suggestion
		return;
```

lib/commons/aria/get-owner-virtual.js
```diff
@@ -0,0 +1,54 @@
...
+
+	const doc = dom.getRootNode(vNode.actualNode);
+	const owner = doc.querySelector(
+		`[aria-owns~=${axe.utils.escapeSelector(id)}]`
+	);
+	if (!owner) {
+		return getAriaOwner(vNode.parent);
+	}
+	if (['presentation', 'none'].includes(aria.getRole(owner))) {
+		return;
```
WilcoFiers: aria-owns is a global attribute, role=none should get ignored. We should probably test that to be sure, but I suspect this if statement needs to be remove.

lib/commons/aria/get-owner-virtual.js
```diff
@@ -0,0 +1,54 @@
...
+	if (!id) {
+		return getAriaOwner(vNode.parent);
+	}
+
+	const doc = dom.getRootNode(vNode.actualNode);
+	const owner = doc.querySelector(
+		`[aria-owns~=${axe.utils.escapeSelector(id)}]`
+	);
+	if (!owner) {
+		return getAriaOwner(vNode.parent);
```
WilcoFiers: As above, I aria-owns doesn't cause all descendants to change owner, just the element that's directly referenced.

lib/commons/aria/get-owner-virtual.js
```diff
@@ -0,0 +1,54 @@
+/* global aria, dom */
+
+/**
+ * Get owner element of a provided virtualNode
+ * @param {Object} virtualNode virtual node
+ * @returns {HTMLElement | undefined}
+ */
+aria.getOwnerVirtual = function getOwnerVirtual(virtualNode) {
+	return getAriaOwner(virtualNode) || getOwner(virtualNode);
```
WilcoFiers: Before doing this, can you test that the thing doesn't have role=presentation or none. Those elements can't be owned.

lib/commons/aria/get-owner-virtual.js
```diff
@@ -0,0 +1,54 @@
+/* global aria, dom */
+
+/**
+ * Get owner element of a provided virtualNode
+ * @param {Object} virtualNode virtual node
+ * @returns {HTMLElement | undefined}
```
WilcoFiers: This needs to return virtual nodes. We have to make sure this works on SerialVirtualNode too.

lib/commons/aria/get-owner-virtual.js
```diff
@@ -0,0 +1,54 @@
...
+	if (!vNode) {
+		return;
+	}
+
+	const id = vNode.attr('id');
+	if (!id) {
+		return getAriaOwner(vNode.parent);
+	}
+
+	const doc = dom.getRootNode(vNode.actualNode);
```
WilcoFiers: Please check if vNode.actualNode exists first.

lib/checks/aria/required-parent.js
```diff
@@ -1,87 +1,41 @@
...
-				getSelector(requiredContext[index])
-			)
-		) {
-			//if one matches, it passes
-			return null;
-		} else {
-			missing.push(requiredContext[index]);
-		}
+	const role = aria.getRole(elm);
+	if (['presentation', 'none', null].includes(aria.getRole(elm))) {
```
WilcoFiers: You've already tested for presentation/none in getOwnerVirtual, why are you doing this again? Also, why are you ignoring null? Firefox doesn't ignore null roles, so I don't think we should either. It'd be good to add an option to this check to make it configurable whether or not null is ignored, but that should probably be handled by getOwnerVirtual. Can you either add that or open an issue for it?

lib/commons/aria/get-owner-virtual.js
WilcoFiers: Also thinking we should add options to this, allowing for the following:
- noAriaOwns
- ignoreNullRoleOwner

These could be passed in from the check options. Either add these now or open an issue for it.

lib/checks/aria/required-parent.js
```diff
@@ -1,87 +1,41 @@
...
 }
 
-function getMissingContext(virtualNode, requiredContext, includeElement) {
-	var index,
-		length,
-		role = virtualNode.actualNode.getAttribute('role'),
-		missing = [];
+const owner = aria.getOwnerVirtual(virtualNode);
+if (!owner) {
+	return undefined;
```
WilcoFiers: add a message to the check, and add tests.

test/integration/rules/aria-required-parent/aria-required-parent.html
```diff
@@ -1,16 +1,51 @@
...
 <div role="menu" id="pass6" aria-owns="fail3"></div>
+
 <div role="listitem" id="fail3"></div>
+
 <div role="list" id="pass7" aria-owns="parent"></div>
-<div id="parent"><div role="listitem" id="pass8"></div></div>
-<div role="grid" id="pass9"><div role="rowgroup" id="pass10">Item 1</div></div>
+
+<div id="parent">
+	<div role="listitem" id="pass8"></div>
```
WilcoFiers: I don't think this should pass. aria-owns needs to point directly to the owned element, not one of its descendants. Can you test this to make sure?

test/commons/aria/get-owner-virtual.js
```diff
@@ -0,0 +1,80 @@
...
+		axe._tree = undefined;
+	});
+
+	it('returns owner element - `div` which is a direct parent (not `aria-owns`)', function() {
+		var vNode = queryFixture(
+			'<div><p role="listitem" id="target">Nothing here.</p></div>'
+		);
+
+		var owner = aria.getOwnerVirtual(vNode);
+		assert.equal(owner.nodeName.toUpperCase(), 'DIV');
```
WilcoFiers: There can be any number of divs on this page. This test doesn't prove what you're trying to prove.

test/commons/aria/get-owner-virtual.js
```diff
@@ -0,0 +1,80 @@
...
+
+	it('returns owner element which references via `aria-owns`', function() {
+		var vNode = queryFixture(
+			'<div role="list"><div aria-owns="parent"></div></div>' +
+				'<div id="parent"><p role="listitem" id="target">Nothing here.</p></div>'
+		);
+
+		var owner = aria.getOwnerVirtual(vNode);
+		assert.equal(owner.nodeName.toUpperCase(), 'DIV');
+		assert.equal(owner.getAttribute('aria-owns'), 'parent');
```
WilcoFiers: the id=parent element should be the owner here, not the element with aria-owns.

jkodu:
Given the change to es modules, it may be worth to re-do this after the conversion to es modules, to avoid a lot of conflict resolution.

PR #2054, opened 2020-02-18
fix(commons):  handle node(s) contained by SVG document when deciphering url properties

jkodu:
Closes issue:
- https://github.com/dequelabs/axe-core/issues/2017

## Reviewer checks

### DIFF
--- a/lib/commons/dom/url-props-from-attribute.js
+++ b/lib/commons/dom/url-props-from-attribute.js
@@ -8,10 +8,7 @@
  * @returns {Object}
  */
 dom.urlPropsFromAttribute = function urlPropsFromAttribute(node, attribute) {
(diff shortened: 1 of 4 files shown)


lib/commons/dom/url-props-from-attribute.js
```diff
@@ -8,7 +8,9 @@
  * @returns {Object}
  */
 dom.urlPropsFromAttribute = function urlPropsFromAttribute(node, attribute) {
-	const value = node[attribute];
+	const value = !node.ownerSVGElement
+		? node[attribute]
+		: node.getAttribute(attribute);
```
WilcoFiers: Didn't fix the problem. `href=""` for SVG will still result in `undefined` getting returned by this function.

Issue #2060, opened 2020-02-25
Breaking change between 3.5.1 and 3.4.2


matijs:
A breaking change seems to have slipped in between `axe-core` 3.5.1 and 3.4.2. 

`eslint-plugin-jsx-a11y` uses `axe-core` in [tests](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/__tests__/__util__/axeMapping.js#L5).

Tests in `eslint-plugin-jsx-a11y` were failing for me with `axe-core` 3.5.1 installed. After ['fixing'](https://github.com/evcohen/eslint-plugin-jsx-a11y/pull/665) the tests, they were passing locally but now failed in CI. I manually downgraded a few times and did some digging and found [this commit](https://github.com/dequelabs/axe-core/commit/a2ddba305003db0540c093f4ad9015c8854cd98d) which looks like it's the cause of the failing tests.

**Expectation**: no breaking changes between patch and minor version bumps.

**Actual**: a breaking change was introduced between two patch version bumps.

**axe-core version**: 3.4.2, 3.5.1

For Tooling issues:
- **Node version**: [several](https://travis-ci.org/evcohen/eslint-plugin-jsx-a11y/builds/654015156)
- **Platform**:  Travis CI

straker:
I'm sorry the change has broken eslint-plugin-jsx-a11y. Unfortunately `axe._audit` is a private axe property and is not covered by our [backward compatibility guidelines](https://github.com/dequelabs/axe-core/blob/develop/doc/backwards-compatibility-doc.md) and thus is not considered a breaking change. 

That being said, we don't want to prevent eslint-plugin-jsx-a11y from working. We introduced a new public function to handle messages `axe.utils.processMessage(message, data)`. This should allow you to pass the message and handle inserting data into it. It still uses the private API to access the messages, so the structure may break again in a future change.

```js
export function axeFailMessage(checkId, data) {
  return axe.utils.processMessage(axe._audit.data.checks[checkId].messages.fail, data);
}
```

straker:
You know, I just noticed that @WilcoFiers added that file, and so I guess we just shot ourselves in the foot... Due to that I think we might have to considered this a breaking change, but I'm not sure we can revert it back to the former API style. I think our best course of action may be to create a new public API that can be used to access check messages.

ljharb:
[...]

PR #2066, opened 2020-02-26
fix(getCheckMessage): add API to return check message

straker:
Technically this is a `feat`, but we had a breaking change for 3.5.1 that broke eslint-plugin-jsx-a11y, so we need to release this as a patch release so we don't go to 3.6.0. 

With this, they can use a public API rather than accessing a private interface:

```js
export function axeFailMessage(checkId, data) {
  return axe.utils.getCheckMessage(checkId, 'fail', data);
}
```

Closes issue: #2060 

## Reviewer checks

### DIFF
--- a/lib/core/utils/get-check-message.js
+++ b/lib/core/utils/get-check-message.js
@@ -1,0 +1,18 @@
+/* global axe*/
+
+/**
+ * Get the pass, fail, or incomplete message for a check.
+ * @param {String} checkId The check id
+ * @param {String} type The message type ('pass', 'fail', or 'incomplete')
+ * @param {Object} [data] The check data
+ * @return {String}
+ */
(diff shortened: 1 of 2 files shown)


lib/core/utils/get-check-message.js
```diff
@@ -0,0 +1,18 @@
...
+ * @param {String} checkId The check id
+ * @param {String} type The message type ('pass', 'fail', or 'incomplete')
+ * @param {Object} [data] The check data
+ * @return {String}
+ */
+axe.utils.getCheckMessage = function getCheckMessage(checkId, type, data) {
+	const check = axe._audit.data.checks[checkId];
+
+	if (!check || !check.messages[type]) {
+		return '';
```
WilcoFiers: I think it might be better if this just threw an error.

PR #2067, opened 2020-02-27
feat(axe.sansPolyfills): add axe version without polyfills

straker:
Reduces the file size of axe from 1.1 MB to 905 KB unminified, and 471 KB to 396 KB minified.

We had discussed [moving to core-js for our polyfills](https://github.com/dequelabs/axe-core/pull/1968), but decided against it in another pr [due to the file size increase](https://github.com/dequelabs/axe-core/pull/1481). However, since we have a version without polyfills, we can use that version in our extensions which do not need the polyfills. (FYI, core-js polyfills only add .05 MB to axe.js, going from 1.06 MB to 1.11 MB)

Also, I'm not sure why we were bothering to beautify the code before minifying it. Our code is already run through prettier so we don't need to beautify that, which means we were beautifying the babel output which seems unnecessary since we're never debugging that code. I also simplified the uglify code options.

Closes issue: #1935 

## Reviewer checks


straker:
@WilcoFiers for v4, should we just make `axe.js` default not have polyfills and then tell IE 11 users how to polyfill what we need? Something like https://polyfill.io/v3/ or brining in `core-js` themselves?

stephenmathieson:
I'm in favor of delaying this until v4 and shipping without any polyfills by default.

WilcoFiers:
This saves a lot less space then I thought it would.

PR #2072, opened 2020-03-02
docs(readme): use axe.run() promise API

WilcoFiers:
I noticed in a recent article someone was using the callback API instead of the promise API. The promise API is the preferred way to run axe, so it doesn't make sense that we use the callback API in the readme.

## Reviewer checks

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,9 +50,15 @@ Now include the javascript file in each of your iframes in your fixtures or test
 Now insert calls at each point in your tests where a new piece of UI becomes visible or exposed:
 
 ```js
-axe.run().then(results => {
-  assert.isEmpty(results.violations, 'Has no accessibility issues');
-}).catch(done);
+axe.run()
+  .then(results => {
+    if (results.violations.length) {
+      throw new Error('Accessibility issues found')
+    }
+  })
+  .catch(err => {
+    console.error('Something bad happened:', err.message)
+  })
 ```
 
 ## Supported Browsers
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -50,12 +50,9 @@ Now include the javascript file in each of your iframes in your fixtures or test
...
 ```js
-axe.run(function (err, results) {
-  if (err) throw err;
-    ok(results.violations.length === 0, 'Should be no accessibility issues');
-    // complete the async call
-    ...
-});
+axe.run().then(results => {
+  assert.isEmpty(results.violations, 'Has no accessibility issues');
+}).catch(done);
```
stephenmathieson: What is `done`?

WilcoFiers: when using it in a test, you often get a `done` function to call. Open to alternative suggestions. Any thoughts?

straker:
We should also update the API.md file to not use the callback parameter in all the examples

README.md
stephenmathieson: IMO *ALL* code snippets in documentation should be copy/paste/run-able. What about this instead:

```js
axe.run()
  .then(results => {
    if (results.violations.length) {
      throw new Error('Accessibility issues found')
    }
  })
  .catch(err => {
    console.error('Something bad happened:', err.message)
  })
```

stephenmathieson:
> We should also update the API.md file to not use the callback parameter in all the examples

I agree, but think this can land without including that change.

Issue #2073, opened 2020-03-02
Update check `invalidrole` to fail when all given roles are invalid


jkodu:
**Expectation:** 
Check `invalidrole` should only fail when all given roles are invalid.

**Actual:** 
Currently `invalidrole` fails when even one role is invalid.

**Motivation:** 
Noticed wrong outcomes when validating ACT rule against this example - https://act-rules.github.io/rules/674b10#passed-example-3

padmavemulapati:
Added one invalid aria role and two pass/valid roles and tested. 
Invalid role only failing , not all the roles if invalid role is there in the test file. Working as expected.
Verified with axe-4.3.0.06416 with chrome
![image](https://user-images.githubusercontent.com/47409579/75980673-fac7da00-5f08-11ea-9a25-aa0e5d252608.png)

jkodu:
Nothing to document here. Closing issue.

PR #2075, opened 2020-03-02
fix: check `invaildrole` fails only when all roles are invalid

jkodu:
`invalidrole` should fail only when all roles on a given element are invalid.

Closes issue:
- https://github.com/dequelabs/axe-core/issues/2073

## Reviewer checks

### DIFF
--- a/lib/checks/aria/invalidrole.js
+++ b/lib/checks/aria/invalidrole.js
@@ -10,7 +10,7 @@ const allInvalid = allRoles.every(
  * Only fail when all the roles are invalid
  */
 if (allInvalid) {
-	this.data(invalidRoles);
+	this.data(allRoles);
 	return true;
 }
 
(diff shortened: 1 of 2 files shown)


0K471N0C4WK:
....

jkodu:
@WilcoFiers As I was working on a fix for something else, I pushed into this branch by mistake. I had to reset it, which dismissed the approval. Please do the needy.

lib/checks/aria/invalidrole.js
```diff
@@ -1,12 +1,15 @@
...
-if (invalidRoles.length > 0) {
+const allRoles = tokenList(virtualNode.attr('role'));
+const invalidRoles = allRoles.filter(
+	role => !aria.isValidRole(role, { allowAbstract: true })
+);
+
+/**
+ * Only fail when all the roles are invalid
+ */
+if (invalidRoles.length > 0 && invalidRoles.length === allRoles.length) {
```
straker:
```suggestion
const allInvalid = allRoles.every(
	role => !aria.isValidRole(role, { allowAbstract: true })
);

/**
 * Only fail when all the roles are invalid
 */
if (allInvalid) {
```

Issue #2069, opened 2020-02-28
Potential regression with skip-link rule


thepassle:
Hi!

At open-wc we provide a project generator, that scaffolds a basic starter application using web components. In this setup, we also scaffold automated accessibility tests that relies on axe-core.

We recently ran into an issue where the accessibility tests started failing where they didnt before, from the changelog, we're not quite sure at which version exactly this was introduced, but our best guess is this code here: https://github.com/dequelabs/axe-core/commit/6a1bcba#diff-ee6303bd768770ef023bc5e2ace0c650R23-R25

The HTML it's failing on is:
```html
<header>
  <ul>
    <li>
      <a href="#main" class=${this.__navClass('main')} @click=${this.__onNavClicked}>
        Main
      </a>
    </li>
    <li>
      <a href="#pageOne" class=${this.__navClass('pageOne')} @click=${this.__onNavClicked}>
        Page One
      </a>
    </li>
    <li>
      <a href="#about" class=${this.__navClass('about')} @click=${this.__onNavClicked}>
        About
      </a>
    </li>
  </ul>
</header>
```

Which you can find [here](https://github.com/open-wc/open-wc/blob/master/packages/create/src/generators/app-lit-element/templates/_MyApp.js#L86), and is a very naive implementation of hash based routing that renders a component once a link is clicked. (The project setup is intentionally simple so users can easily make it their own)

Axe seems to have started identifying these as skiplinks though, and fails with: 

```
Rule: skip-link
Impact: moderate
The skip-link target should exist and be focusable (https://dequeuniversity.com/rules/axe/3.5/skip-link?application=axeAPI)
```

You can find the issue on our repo [here](https://github.com/open-wc/open-wc/issues/1378). We think this may have been a regression in axe-core, and we'd be interested to hear your thoughts on this and how to move forward with this 😊

straker:
Thanks for letting us know. I believe this is a similar problem from #2055

straker:
Going to close this and track the work in #2055. A pr has been merged and we're about to create a release with the changes.

Issue #2055, opened 2020-02-18
Error for skip links to other pages.


KeironLowe:
I have some links which are displayed on all pages which link to an anchor on the homepage. These sections are focusable and when I run the scanner on the homepage I don't see any errors, but when I scan a different page they throw the skip links have focusable target error. The links look like `/#about`, `/#services`, `/#contact`.

Expectation: If scanning the page the links go to, then the links should be treated as skip links and scanned as much. If not scanning the page the links go to, then these be ignored.

Actual: The links always throw an error, even if linking to another page

Motivation: Reduce errors in scans

<pre><code>axe-core version: 3.5.1</code></pre>

straker:
So if I understand correctly, you have a few links at the top of the page that  look like:

```html
<a href="/#about"></a>
<a href="/#services"></a>
<a href="/#contact"></a>
```

which fails the skip-link rule for pages that are not the root page (`/`)?

KeironLowe:
@straker yes that's correct.

padmavemulapati:
For three of the above links - givng fail (/#about, /#services, /#contact)
![image](https://user-images.githubusercontent.com/47409579/76215799-79888400-6235-11ea-9fde-bdf5ca553f96.png)
![image](https://user-images.githubusercontent.com/47409579/76215833-8efdae00-6235-11ea-9aa6-d1aeabd9798c.png)
![image](https://user-images.githubusercontent.com/47409579/76215888-a6d53200-6235-11ea-99f9-94e39d88c687.png)

straker:
No docs needed

PR #2079, opened 2020-03-02
fix(skip-link): identify as skip-link only if the link is offscreen

straker:
We had a few false positives with normal links (`<a href="/#target"></a>`) being identified as skip links. This changes the definition of a skip link to only include links which are offscreen. 

Closes issue: #2069, #2055

## Reviewer checks

### DIFF
--- a/lib/commons/dom/is-skip-link.js
+++ b/lib/commons/dom/is-skip-link.js
@@ -20,9 +20,9 @@ dom.isSkipLink = function(element) {
 	if (typeof axe._cache.get('firstPageLink') !== 'undefined') {
 		firstPageLink = axe._cache.get('firstPageLink');
 	} else {
-		// define a skip link as any offscreen anchor element whose href starts with
-		// `#...` and which precedes the first anchor element whose href doesn't
-		// start with  `#...` (that is, a link to a page)
+		// define a skip link as any anchor element whose href starts with `#...`
+		// and which precedes the first anchor element whose href doesn't start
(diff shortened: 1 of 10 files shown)


test/playground.html
```diff
@@ -2,7 +2,10 @@
 <html lang="en">
 	<title>O hai</title>
 
-	<div class="foo" id="foo">foo</div>
+	<button aria-description="Undefined global">myVar</button>
+
+	<div id="aria-describedby">Undefined global</div>
+	<button aria-describedby="aria-describedby">myVar</button>
```
WilcoFiers: Did you mean to make this change?

straker: No... I always forget to revert the playground changes.

lib/commons/dom/is-skip-link.js
```diff
@@ -32,14 +32,14 @@ dom.isSkipLink = function(element) {
 		axe._cache.set('firstPageLink', firstPageLink || null);
 	}
 
-	// if there are no page links then all all links will need to be
+	// if there are no page links then all offscreen links will need to be
 	// considered as skip links
 	if (!firstPageLink) {
-		return true;
+		return dom.isOffscreen(element);
```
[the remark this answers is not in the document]
straker: I'm not sure we could solve the two issues if the skip link is currently visible. Hidden skip link was the only solution that would identify them. Otherwise I think we have to move this rule to a best-practice and not fail it.

PR #2087, opened 2020-03-09
chore: merge master into develop

straker:
Something went wrong with the merging (probably cherry-picking) so we have conflicts from master merge as well as too many files being changed. So had to merge develop into master and merge back into develop.

## Reviewer checks


straker:
This merge is not being squashed so we don't make master/develop out of sync again. The non-conforming commit messages will appear in the `git log`, but standard-version should ignore them for the next release.

Issue #2039, opened 2020-02-11
td-has-header should not always fail <td headers="">


WilcoFiers:
The `td-has-header` rule seems to consistently fail `<td headers="">` elements. We have some test data that shows that AT will fall back to column / rowheaders in this situation.

```html
<table>
  <tr>
    <th scope="col">Projects</th>
    <th scope="col">Progress</th>
  </tr>
  <tr>
    <td headers="">My Project</td>
    <td>15%</td>
  </tr>
</table>
```

**IE + JAWS**: Works
**Chrome + JAWS**: Works
**Firefox + NVDA**: Works
**Safari + VO**: Works

I think we should report this for review. Axe-core should probably still fail if the headers attribute references something that doesn't exist, or something outside the table.

jkodu:
@WilcoFiers did you mean the rule `td-headers-attr` here?

WilcoFiers:
Nope. The problem here is that td-has-header stops looking if it finds a headers attribute without associated headers. It should continue looking for column/rowheaders. This is not according to spec, but we have test data showing this happens consistently.

padmavemulapati:
Its verified as part of the verification of `https://github.com/dequelabs/axe-core/issues/2013` , yes we are seeing the results as `incomplete` which is expected as per the change done here in this ticket.
![image (11)](https://user-images.githubusercontent.com/47409579/77031885-ebe12d00-69c8-11ea-8918-250f22f0380f.png)

straker:
No docs needed

PR #2095, opened 2020-03-11
fix(td-has-headers): don't fail for empty headers attribute

straker:
Part 1 of 2 (part 2: https://github.com/dequelabs/axe-core/pull/2096)

Partially closes issue: #2039

## Reviewer checks

### DIFF
--- a/lib/commons/table/get-headers.js
+++ b/lib/commons/table/get-headers.js
@@ -57,7 +57,7 @@ function traverseForHeaders(headerType, position, tableGrid) {
  * @return {Array<HTMLTableCellElement>} Array of headers associated to the table cell
  */
 table.getHeaders = function(cell, tableGrid) {
-	if (cell.hasAttribute('headers')) {
+	if (cell.getAttribute('headers')) {
 		return commons.dom.idrefs(cell, 'headers');
 	}
 	if (!tableGrid) {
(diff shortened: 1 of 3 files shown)


lib/commons/table/get-headers.js
```diff
@@ -57,7 +57,7 @@ function traverseForHeaders(headerType, position, tableGrid) {
  * @return {Array<HTMLTableCellElement>} Array of headers associated to the table cell
  */
 table.getHeaders = function(cell, tableGrid) {
-	if (cell.hasAttribute('headers')) {
+	if (cell.getAttribute('headers')) {
 		return commons.dom.idrefs(cell, 'headers');
```
WilcoFiers: That doesn't fix the whole problem. The issue lays in this line right here, not the one above. There are two problems with this line:

1. It finds things that aren't cells in the table
2. if it doesn't find anything, it shouldn't return empty here, but continue down to look for row/ column headers.

straker: 1 is already covered by td-headers-attr rule so we don't need to fail it in this rule, thus we can just assume if it has an attribute (we can trim) then it can pass this rule. td-headers-attr (part 2 change) then can validate the headers attr is correct.

WilcoFiers: I don't follow... td-headers-attr fails if the headers attribute is incorrect. We shouldn't also fail td-has-headers if there are fallback headers, like this:

```html
<tr> <th>Hello</th> <td headers="non-existant">goodbye</td> </tr>
```

I think this should pass `td-has-headers` and it should fail `td-headers-attr`.

PR #2096, opened 2020-03-11
fix(td-headers-attr): mark as needs review if headers attr is empty

straker:
Part 2 of 2 (part 1: https://github.com/dequelabs/axe-core/pull/2095). Also modernized and cleaned up the code.

Closes issue: #2039

## Reviewer checks


lib/checks/tables/td-headers-attr.js
```diff
@@ -1,40 +1,36 @@
...
-var badCells = cells.reduce(function(badCells, cell) {
-	var isSelf, notOfTable;
+cells.forEach(cell => {
+	let isSelf, notOfTable;
+
+	if (!cell.hasAttribute('headers')) {
+		return;
+	}
+
+	const headersAttr = cell.getAttribute('headers');
```
WilcoFiers: Might be better to trim here, so that `headers=" "` is put for review too.

lib/checks/tables/td-headers-attr.js
```diff
@@ -43,20 +39,22 @@ var badCells = cells.reduce(function(badCells, cell) {
 		}
 
 		// Check if the headers are of cells inside the table
-		notOfTable = headers.reduce(function(fail, header) {
-			return fail || ids.indexOf(header) === -1;
-		}, false);
+		notOfTable = headers.find(header => !ids.includes(header));
```
WilcoFiers: Might be better to use `some` instead of `find` to make notOfTable a boolean value. I misread it because of this the first time around.

```suggestion
		notOfTable = headers.some(header => !ids.includes(header));
```

lib/checks/tables/td-headers-attr.js
```diff
@@ -1,40 +1,36 @@
...
 	if (cell.getAttribute('id')) {
 		ids.push(cell.getAttribute('id'));
 	}
 	return ids;
 }, []);
 
-var badCells = cells.reduce(function(badCells, cell) {
-	var isSelf, notOfTable;
+cells.forEach(cell => {
+	let isSelf, notOfTable;
```
WilcoFiers: Lets give these the defaults we expect them to have, instead of relying on type casting of undefined.
```suggestion
	let isSelf = false;
	let notOfTable = false;
```

lib/checks/tables/td-headers-attr.js
```diff
@@ -43,20 +39,22 @@ var badCells = cells.reduce(function(badCells, cell) {
...
 
 		if (isSelf || notOfTable) {
 			badCells.push(cell);
 		}
 	}
-	return badCells;
-}, []);
+});
+
+if (reviewCells.length) {
```
WilcoFiers: If there are bad cells, this check needs to return false, not undefined. Best to move this if down.

Issue #2101, opened 2020-03-18
Color contrast checking of unicode private use characters


ahuth:
Axe version: 3.5.2
OS: MacOS and Linux
Browsers: Firefox and Chrome

My company uses a custom font to render various icons for certain unicode characters. It looks like #2005 fixed the color-contrast rule for non-BMP characters, and now I'm getting color-contrast failures that I don't expect.

I believe this is caused by the non-BMP regex not matching characters in Supplementary Private Use Area-A

Changing this would solve our issue, and I'd be happy contribute fixes for either or both of them, if this seem reasonable.

## Our setup

We use a custom font to show icons for unicode characters in Supplementary Private Use Area-A. For example:

<img width="368" alt="airmoji" src="https://user-images.githubusercontent.com/2503289/77013257-eb935400-692c-11ea-8fe0-505df492a75f.png">

The red star, checkmark, and medal icons are the characters in question, and are now failing color-contrast checks. I believe they meet the "incidental" exception to the minimum contrast success criteria.

## Possible fix

The characters in question are in the range F0000–FFFFD, and outside the Basic Multilingual Plane range. We could add this range to the [regex for testing non-BMP characters](https://github.com/dequelabs/axe-core/blob/7952a377b960c2403c8d4758a9f35ebe525549fe/lib/commons/text/unicode.js#L100).

Let me know if this makes sense, and I'd be happy to contribute a fix.

ahuth:
Also, thanks for all your work on axe-core!

We use it extensively at Airbnb. For example, we've ran axe on around 16,840 CI builds so far this year. It's been instrumental to helping scale our accessibility efforts.

ahuth:
Yikes. Updating the regex and testing it is turning out to be harder than anticipated in #2102 😞 . Not sure if there are any other potential solutions.

pieralexandreb:
Hi, it looks like my team is also having an issue here because of the nonBmp rule. We are also using a custom font to render icons and some of our tests are broken. Is there a way to turn off the rule or to modify the rule?

straker:
You can disable color-contrast completely, buy I don't think there's a way to modify the rule to flag icons. Would you be able to provide an example of the problem? Would love to see what's going on and maybe we can look into fixing it.

pieralexandreb:
[...]

PR #2102, opened 2020-03-19
fix(unicode) detect supplementary private use area A

ahuth:
This PR adds unicode characters in Supplementary Private Use Area-A to the `getUnicodeNonBmpRegExp` regex.

Closes issue: https://github.com/dequelabs/axe-core/issues/2101

## Reviewer checks

### DIFF
--- a/lib/commons/text/unicode.js
+++ b/lib/commons/text/unicode.js
@@ -135,5 +135,5 @@ function getSupplementaryPrivateUseRegExp() {
 	 * https://www.unicode.org/charts/PDF/UDC00.pdf
 	 * https://www.unicode.org/charts/PDF/UF0000.pdf
 	 */
-	return /[\uDB80-\uDBBF][\uDC00-\uDFFD]|[\u{F0000}-\u{FFFFD}]/gu;
+	return /[\uDB80-\uDBBF][\uDC00-\uDFFD]|(?:[\uDB80-\uDBBE][\uDC00-\uDFFF]|\uDBBF[\uDC00-\uDFFD])/g;
 }
--- a/lib/commons/text/unicode.js
+++ b/lib/commons/text/unicode.js
@@ -19,7 +19,10 @@ text.hasUnicode = function hasUnicode(str, options) {
 		return axe.imports.emojiRegexText().test(str);
 	}
 	if (nonBmp) {
-		return getUnicodeNonBmpRegExp().test(str);
+		return (
+			getUnicodeNonBmpRegExp().test(str) ||
+			getSupplementaryPrivateUseRegExp().test(str)
+		);
 	}
 	if (punctuations) {
 		return getPunctuationRegExp().test(str);
@@ -93,12 +96,11 @@ function getUnicodeNonBmpRegExp() {
 	 * '\u2600-\u26FF'  Misc Symbols
 	 * '\u2700-\u27BF'  Dingbats
 	 * '\uE000-\uF8FF'  Private Use
-	 * '\u{F0000}-\u{FFFFD}' Supplementary Private Use Area A
 	 *
 	 * Note: plane '\u2000-\u206F' used for General punctuation is excluded as it is handled in -> getPunctuationRegExp
 	 */
 
-	return /[\u1D00-\u1D7F\u1D80-\u1DBF\u1DC0-\u1DFF\u20A0-\u20CF\u20D0-\u20FF\u2100-\u214F\u2150-\u218F\u2190-\u21FF\u2200-\u22FF\u2300-\u23FF\u2400-\u243F\u2440-\u245F\u2460-\u24FF\u2500-\u257F\u2580-\u259F\u25A0-\u25FF\u2600-\u26FF\u2700-\u27BF\uE000-\uF8FF\u{F0000}-\u{FFFFD}]/gu;
+	return /[\u1D00-\u1D7F\u1D80-\u1DBF\u1DC0-\u1DFF\u20A0-\u20CF\u20D0-\u20FF\u2100-\u214F\u2150-\u218F\u2190-\u21FF\u2200-\u22FF\u2300-\u23FF\u2400-\u243F\u2440-\u245F\u2460-\u24FF\u2500-\u257F\u2580-\u259F\u25A0-\u25FF\u2600-\u26FF\u2700-\u27BF\uE000-\uF8FF]/g;
 }
 
 /**
@@ -131,6 +133,7 @@ function getSupplementaryPrivateUseRegExp() {
 	/**
 	 * Reference: https://www.unicode.org/charts/PDF/UD800.pdf
 	 * https://www.unicode.org/charts/PDF/UDC00.pdf
+	 * https://www.unicode.org/charts/PDF/UF0000.pdf
 	 */
-	return /[\uDB80-\uDBBF][\uDC00-\uDFFD]/g;
+	return /[\uDB80-\uDBBF][\uDC00-\uDFFD]|[\u{F0000}-\u{FFFFD}]/gu;
 }
--- a/lib/commons/text/unicode.js
+++ b/lib/commons/text/unicode.js
@@ -93,11 +93,12 @@ function getUnicodeNonBmpRegExp() {
 	 * '\u2600-\u26FF'  Misc Symbols
 	 * '\u2700-\u27BF'  Dingbats
 	 * '\uE000-\uF8FF'  Private Use
+	 * '\u{F0000}-\u{FFFFD}' Supplementary Private Use Area A
 	 *
 	 * Note: plane '\u2000-\u206F' used for General punctuation is excluded as it is handled in -> getPunctuationRegExp
 	 */
 
-	return /[\u1D00-\u1D7F\u1D80-\u1DBF\u1DC0-\u1DFF\u20A0-\u20CF\u20D0-\u20FF\u2100-\u214F\u2150-\u218F\u2190-\u21FF\u2200-\u22FF\u2300-\u23FF\u2400-\u243F\u2440-\u245F\u2460-\u24FF\u2500-\u257F\u2580-\u259F\u25A0-\u25FF\u2600-\u26FF\u2700-\u27BF\uE000-\uF8FF]/g;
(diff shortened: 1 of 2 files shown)


test/commons/text/unicode.js
```diff
@@ -69,6 +69,13 @@ describe('text.hasUnicode', function() {
 			});
 			assert.isTrue(actual);
 		});
+
+		it('returns true for a srting with characters in the Supplementary Private Use Area-A', function() {
+			var actual = axe.commons.text.hasUnicode('\u{F0019}', {
```
ahuth: Unfortunately, the unicode escape sequence is causing a syntax error in eslint.

Not seeing that in other projects using eslint, and the test and build steps happen just fine. Not sure what's going on there, and this will take some more debugging.

lib/commons/text/unicode.js
```diff
@@ -93,11 +93,12 @@ function getUnicodeNonBmpRegExp() {
 	 * '\u2600-\u26FF'  Misc Symbols
 	 * '\u2700-\u27BF'  Dingbats
 	 * '\uE000-\uF8FF'  Private Use
+	 * '\u{F0000}-\u{FFFFD}' Private Use
 	 *
 	 * Note: plane '\u2000-\u206F' used for General punctuation is excluded as it is handled in -> getPunctuationRegExp
 	 */
 
-	return /[\u1D00-\u1D7F\u1D80-\u1DBF\u1DC0-\u1DFF\u20A0-\u20CF\u20D0-\u20FF\u2100-\u214F\u2150-\u218F\u2190-\u21FF\u2200-\u22FF\u2300-\u23FF\u2400-\u243F\u2440-\u245F\u2460-\u24FF\u2500-\u257F\u2580-\u259F\u25A0-\u25FF\u2600-\u26FF\u2700-\u27BF\uE000-\uF8FF]/g;
+	return /[\u1D00-\u1D7F\u1D80-\u1DBF\u1DC0-\u1DFF\u20A0-\u20CF\u20D0-\u20FF\u2100-\u214F\u2150-\u218F\u2190-\u21FF\u2200-\u22FF\u2300-\u23FF\u2400-\u243F\u2440-\u245F\u2460-\u24FF\u2500-\u257F\u2580-\u259F\u25A0-\u25FF\u2600-\u26FF\u2700-\u27BF\uE000-\uF8FF\u{F0000}-\u{FFFFD}]/ug;
```
ahuth: Note the `u` flag added to the regex. This enables using unicode escape sequences in the pattern.

One side-effect of this is that the build step actually compiles the regex to this in the final output:

```js
function getUnicodeNonBmpRegExp() {
  return /(?:[\u1D00-\u1DFF\u20A0-\u27BF\uE000-\uF8FF]|[\uDB80-\uDBBE][\uDC00-\uDFFF]|\uDBBF[\uDC00-\uDFFD])/g;
}
```

The `u` flag is an ES6 feature, and I'm assuming that [preset-env is compiling to a regex that will work in older ES versions](https://babeljs.io/docs/en/babel-plugin-transform-unicode-regex).

This is kind of unfortunate, but the tests do pass locally.

test/commons/text/unicode.js
ahuth: Oh, just realized that the test directory is parsed as ES5. That explains it. 🤦‍♂

test/commons/text/unicode.js
```diff
@@ -69,6 +69,13 @@ describe('text.hasUnicode', function() {
 			});
 			assert.isTrue(actual);
 		});
+
+		it('returns true for a srting with characters in the Supplementary Private Use Area-A', function() {
+			var actual = axe.commons.text.hasUnicode('\uDB80\uDC19', {
```
ahuth: The string here is the same as the unicode escape sequence `'\u{F0019}'`. However, ES5 doesn't support those, and the tests are parsed by eslint as ES5. Instead this is the corresponding surrogate pair.

lib/commons/text/unicode.js
```diff
@@ -93,11 +93,12 @@ function getUnicodeNonBmpRegExp() {
 	 * '\u2600-\u26FF'  Misc Symbols
 	 * '\u2700-\u27BF'  Dingbats
 	 * '\uE000-\uF8FF'  Private Use
+	 * '\u{F0000}-\u{FFFFD}' Supplementary Private Use Area A
 	 *
 	 * Note: plane '\u2000-\u206F' used for General punctuation is excluded as it is handled in -> getPunctuationRegExp
 	 */
 
-	return /[\u1D00-\u1D7F\u1D80-\u1DBF\u1DC0-\u1DFF\u20A0-\u20CF\u20D0-\u20FF\u2100-\u214F\u2150-\u218F\u2190-\u21FF\u2200-\u22FF\u2300-\u23FF\u2400-\u243F\u2440-\u245F\u2460-\u24FF\u2500-\u257F\u2580-\u259F\u25A0-\u25FF\u2600-\u26FF\u2700-\u27BF\uE000-\uF8FF]/g;
+	return /[\u1D00-\u1D7F\u1D80-\u1DBF\u1DC0-\u1DFF\u20A0-\u20CF\u20D0-\u20FF\u2100-\u214F\u2150-\u218F\u2190-\u21FF\u2200-\u22FF\u2300-\u23FF\u2400-\u243F\u2440-\u245F\u2460-\u24FF\u2500-\u257F\u2580-\u259F\u25A0-\u25FF\u2600-\u26FF\u2700-\u27BF\uE000-\uF8FF\u{F0000}-\u{FFFFD}]/gu;
```
ahuth: Note the `u` flag added to the regex. This enables using unicode escape sequences in the pattern.

One side-effect of this is that the build step actually compiles the regex to this in the final output:

```js
function getUnicodeNonBmpRegExp() {
  return /(?:[\u1D00-\u1DFF\u20A0-\u27BF\uE000-\uF8FF]|[\uDB80-\uDBBE][\uDC00-\uDFFF]|\uDBBF[\uDC00-\uDFFD])/g;
}
```

The `u` flag is an ES6 feature, and I'm assuming that [preset-env is compiling to a regex that will work in older ES versions](https://babeljs.io/docs/en/babel-plugin-transform-unicode-regex).

straker: I think we should add this range to the `getSupplementaryPrivateUseRegExp` function. I also think the `text.hasUnicode` should check this range for the `nonBMP` flag (currently it's missing and that's probably a bug since `removeUnicode` with `nonBMP` uses it).

ahuth: Makes sense 👍

ahuth:
Made the changes, and the build looks green @straker .

I noticed yesterday that test_examples was green on one build, and then red on the next (where I only changed a comment).

Anyway, let me know if there's anything else we need to do here.

lib/commons/text/unicode.js
```diff
@@ -130,6 +133,7 @@ function getSupplementaryPrivateUseRegExp() {
 	/**
 	 * Reference: https://www.unicode.org/charts/PDF/UD800.pdf
 	 * https://www.unicode.org/charts/PDF/UDC00.pdf
+	 * https://www.unicode.org/charts/PDF/UF0000.pdf
 	 */
-	return /[\uDB80-\uDBBF][\uDC00-\uDFFD]/g;
+	return /[\uDB80-\uDBBF][\uDC00-\uDFFD]|[\u{F0000}-\u{FFFFD}]/gu;
```
straker: So the `u` flag might pose a problem in some situations in IE11 as it's [not supported](https://kangax.github.io/compat-table/es6/#test-RegExp_y_and_u_flags). We've had quite the issue with these regexs before in IE11, so probably best to change this to use the surrogate pair codes as the `[\uDB80-\uDBBF][\uDC00-\uDFFD]` is doing.

ahuth: Babel's preset-env is compiling this to a regex without the `u` flag, and we should be good with regards to IE11 support.

The compiled axe.js has:

```js
function getSupplementaryPrivateUseRegExp() {
    return /(?:[\uDB80-\uDBBF](?![\uDC00-\uDFFF]))(?:(?:[^\uD800-\uDBFF]|^)[\uDC00-\uDFFD])|(?:[\uDB80-\uDBBE][\uDC00-\uDFFF]|\uDBBF[\uDC00-\uDFFD])/g;
}
```

That's obviously a little bit gnarlier than it was before (especially with the added non-capturing groups).

Manually writing the surrogate pair I think would give us something like:

```js
function getSupplementaryPrivateUseRegExp() {
    return /[\uDB80-\uDBBF][\uDC00-\uDFFD]|(?:[\uDB80-\uDBBE][\uDC00-\uDFFF]|\uDBBF[\uDC00-\uDFFD])/g;
}
```

Because the regex gets compiled to a regex compatible with ES5, it may be preferable to keep the unicode escape sequences in the regex, since it's easier to see what the actual range is.

Do the windows tests run in IE11, or on Edge?

Or, if you still prefer manually writing the surrogate pair, I'll do it. Let me know! Thanks

ahuth: Actually, I'll just update it to

```js
/[\uDB80-\uDBBF][\uDC00-\uDFFD]|[\udb80-\udbbe][\udc00-\udfff]|\udbbf[\udc00-\udffe]]/g
```

to be safe. The `u` flag does get compiled to a compatible regex, but with the surrogate pairs there's no ambiguity as to what's happening.

lib/commons/text/unicode.js
```diff
@@ -127,9 +130,11 @@ function getPunctuationRegExp() {
...
-	 * Reference: https://www.unicode.org/charts/PDF/UD800.pdf
-	 * https://www.unicode.org/charts/PDF/UDC00.pdf
-	 */
-	return /[\uDB80-\uDBBF][\uDC00-\uDFFD]/g;
+	// 1. High surrogate area (https://www.unicode.org/charts/PDF/UD800.pdf)
+	// 2. Low surrogate area (https://www.unicode.org/charts/PDF/UDC00.pdf)
+	// 3. Supplementary private use area A (https://www.unicode.org/charts/PDF/UF0000.pdf)
+	//
+	//             1              2                                  3
+	//      ┏━━━━━━┻━━━━━━┓┏━━━━━━┻━━━━━━┓ ┏━━━━━━━━━━━━━━━━━━━━━━━━━┻━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
```
ahuth: I got a little clever documenting this. Say the word and I'll remove this.

ahuth:
Thanks!

PR #2106, opened 2020-03-22
chore: simplify the regex getting supplementary private use characters

ahuth:
I added an additional range to the regex in #2102. The purpose of it was to match unicode characters in Supplementary Private Use Area A, and I got the regex itself from https://apps.timwhitlock.info/js/regex.

However, it randomly occurred to me that the part of the regex I added is actually a superset of what was already there. And the real fix to #2101 was not the addition to the regex, but was solely due to [adding `getSupplementaryPrivateUseRegExp` to `hasUnicode`](https://github.com/dequelabs/axe-core/pull/2102/files#diff-e46b3ba4cd0e3b35ba002d5305ee074b) 🤦‍♂.

The existing part was:

```regex
[\uDB80-\uDBBF][\uDC00-\uDFFD]
```

The part I previously added (with an "or") was:

``` regex
[\uDB80-\uDBBE][\uDC00-\uDFFF]|\uDBBF[\uDC00-\uDFFD]
```

Note the overlap in the ranges. In fact, as far as I can tell, they're equivalent, except the previous one excluded anything where the low surrogate was DFFE or DFFF.

I've gone ahead and:
1. Removed the part of the regex I added
2. Expanded the range of the low surrogate to include DFFE and DFFF

## Reviewer checks

### DIFF
--- a/lib/commons/text/unicode.js
+++ b/lib/commons/text/unicode.js
@@ -134,6 +134,10 @@ function getSupplementaryPrivateUseRegExp() {
 	// characters between F0000 and FFFFF. Because ES5 doesn't have a syntax for regular expressions
 	// of such characters, search instead for the corresponding surrogate pairs.
 	//
+	// Code points FFFFD and FFFFF are "noncharacters", but the regex still matches them, because its
+	// intent is to match things we don't want to check color contrast for. This is why the low
+	// surrogate range in the regex ends at DFFF, not DFFD.
+	//
 	// 1. High surrogate area (https://www.unicode.org/charts/PDF/UD800.pdf)
 	// 2. Low surrogate area (https://www.unicode.org/charts/PDF/UDC00.pdf)
 	//
--- a/lib/commons/text/unicode.js
+++ b/lib/commons/text/unicode.js
@@ -130,11 +130,14 @@ function getPunctuationRegExp() {
  * @returns {RegExp}
  */
 function getSupplementaryPrivateUseRegExp() {
+	// Supplementary private use area A (https://www.unicode.org/charts/PDF/UF0000.pdf) contains
+	// characters between F0000 and FFFFF. Because ES5 doesn't have a syntax for regular expressions
+	// of such characters, search instead for the corresponding surrogate pairs.
+	//
 	// 1. High surrogate area (https://www.unicode.org/charts/PDF/UD800.pdf)
 	// 2. Low surrogate area (https://www.unicode.org/charts/PDF/UDC00.pdf)
-	// 3. Supplementary private use area A (https://www.unicode.org/charts/PDF/UF0000.pdf)
 	//
-	//             1              2                                  3
-	//      ┏━━━━━━┻━━━━━━┓┏━━━━━━┻━━━━━━┓ ┏━━━━━━━━━━━━━━━━━━━━━━━━━┻━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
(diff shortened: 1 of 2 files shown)


lib/commons/text/unicode.js
```diff
@@ -130,11 +130,14 @@ function getPunctuationRegExp() {
...
 	// 1. High surrogate area (https://www.unicode.org/charts/PDF/UD800.pdf)
 	// 2. Low surrogate area (https://www.unicode.org/charts/PDF/UDC00.pdf)
-	// 3. Supplementary private use area A (https://www.unicode.org/charts/PDF/UF0000.pdf)
 	//
-	//             1              2                                  3
-	//      ┏━━━━━━┻━━━━━━┓┏━━━━━━┻━━━━━━┓ ┏━━━━━━━━━━━━━━━━━━━━━━━━━┻━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
-	return /[\uDB80-\uDBBF][\uDC00-\uDFFD]|(?:[\uDB80-\uDBBE][\uDC00-\uDFFF]|\uDBBF[\uDC00-\uDFFD])/g;
+	//             1              2
+	//      ┏━━━━━━┻━━━━━━┓┏━━━━━━┻━━━━━━┓
+	return /[\uDB80-\uDBBF][\uDC00-\uDFFF]/g;
```
ahuth: Note that the low surrogate has been expanded to include DFFF. Without that, I think we exclude characters such as [F03FF (\uDB80\uDFFF)](https://www.fileformat.info/info/unicode/char/f03ff/index.htm), which seems to be a valid character in the private use area. I've updated the tests to catch this.

Technically, FFFFE and FFFFF are in the private use area, but are "non characters".

I think it's okay to still match those, but if we think that will be an issue, we can replace this with the part I originally added in #2102:

```js
return /[\uDB80-\uDBBE][\uDC00-\uDFFF]|\uDBBF[\uDC00-\uDFFD]/g;
```

which would exclude FFFFE/FFFFF.

test/commons/text/unicode.js
```diff
@@ -71,7 +71,7 @@ describe('text.hasUnicode', function() {
 		});
 
 		it('returns true for a string with characters in supplementary private use area A', function() {
-			var actual = axe.commons.text.hasUnicode('\uDB80\uDC19', {
+			var actual = axe.commons.text.hasUnicode('\uDB80\uDFFE', {
```
ahuth: Updated these tests so that they pass if the regex is:

```js
// The change, as of this PR
/[\uDB80-\uDBBF][\uDC00-\uDFFF]/g
```

but fail if it's

```js
/[\uDB80-\uDBBF][\uDC00-\uDFFD]/g
```

lib/commons/text/unicode.js
straker: I think it should be ok if we include FFFFE and FFFFF. The regex is used to strip out characters that we shouldn't check for color contrast so non characters fit that bill.

straker: Could you add this as a comment to the code? That way when we look at this in the future we'll remember why the low surrogate doesn't end at uDFFD (which the doc points to)?

ahuth: Will do

ahuth: And done

straker:
Thanks again!

ahuth:
Thank you!

Issue #2109, opened 2020-03-23
Setup grunt-webpack


straker:
Setup grunt-webpack and configure it to handle the migration to ES Modules.

straker:
QA: nothing to test (dev work)

PR #2124, opened 2020-03-23
chore: add grunt-webpack

straker:
Webpacked the `commons/utils` directory as it has nothing in it really. So made it an easy test case to see this all working. The export is pretty useless in this case but we can fix it later once we convert the entire library in ES Modules (the file itself should go away).

We'll need to be extra sure this is all working. So reviewer should pull down the changes and run `npm run build` and make sure the compiled output (both `axe.js` and `axe.min.js`) still runs in the browser and that `axe.commons.utils` is still there.

Closes issue: #2109 

## Reviewer checks

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -371,6 +371,7 @@ module.exports = function(grunt) {
 	grunt.registerTask('translate', [
 		'pre-build',
 		'validate',
+		'webpack',
 		'concat:commons',
 		'add-locale'
 	]);
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,3 +1,5 @@
+var path = require('path');
+
 /*eslint
 camelcase: ["error", {"properties": "never"}]
 */
@@ -15,6 +17,7 @@ module.exports = function(grunt) {
 	grunt.loadNpmTasks('grunt-contrib-watch');
 	grunt.loadNpmTasks('grunt-parallel');
 	grunt.loadNpmTasks('grunt-run');
+	grunt.loadNpmTasks('grunt-webpack');
 	grunt.loadTasks('build/tasks');
 
 	var langs;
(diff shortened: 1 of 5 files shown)


Gruntfile.js
```diff
@@ -137,11 +140,29 @@ module.exports = function(grunt) {
...
+					'tmp/commons/**/*.js',
+
 					'lib/commons/outro.stub'
 				],
 				dest: 'tmp/commons.js'
 			}
 		},
+		webpack: {
+			commonsUtils: {
+				devtool: 'source-map',
```
stephenmathieson: Do we do anything with the source map generated here? I think we should either be using an "inline" source map or none at all (unless there's some build magic elsewhere that I'm not aware of).

straker: I'm not sure? I just read it prevents webpack from turning it into a string and evaling it.

stephenmathieson: I am **strongly** against copy/pasting Webpack configuration without having a full understanding of what it's doing. If you want help with Webpack stuff, please LMK (send a Slack message).

Per [the `devtool` docs](https://webpack.js.org/configuration/devtool/):

> source-map - A full SourceMap is emitted as a separate file. It adds a reference comment to the bundle so development tools know where to find it.

This means Webpack will create a source-map in a separate file for the bundle you're creating here (as well as every other bundle created). Since we're not going to do anything with this source map, I do not believe we want this. It's going to slow down the build and provide no benefit.

If we want source maps for development purposes, we should use one Webpack's "inline" options (so there's no extra build step to include the map in our bundle(s)). However, I do not recall source maps being discussed in your proposal for this change, so I think we can safely leave them out for now.

Gruntfile.js
```diff
@@ -137,11 +140,29 @@ module.exports = function(grunt) {
...
+				mode: 'development',
+				entry: path.resolve(__dirname, 'lib/commons/utils/index.js'),
```
stephenmathieson: Since this file contains ES6+, I believe we'll want to run it thru Babel, but I don't see any custom loaders added here. How does this work?

straker: For now this will just concat things together. The build pipeline already babels so we don't want to do it twice. Once this is all converted over we can add the babel load to our webpack config.

Issue #2113, opened 2020-03-23
Convert commons/forms to ES Modules


straker:

straker:
QA: nothing to test (dev work)

PR #2129, opened 2020-03-24
chore: update commons.forms to ES Modules

straker:
Closes issue: #2113 

## Reviewer checks

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -5,10 +5,17 @@ camelcase: ["error", {"properties": "never"}]
 */
 var testConfig = require('./build/test/config');
 
-var defaultWebpackConfig = {
-	devtool: false,
-	mode: 'development'
-};
+function createWebpackConfig(input, output) {
+	return {
+		devtool: false,
+		mode: 'development',
+		entry: path.resolve(__dirname, input),
+		output: {
+			filename: 'index.js',
+			path: path.resolve(__dirname, output)
+		}
+	};
+}
 
 module.exports = function(grunt) {
 	'use strict';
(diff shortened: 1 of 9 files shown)


Gruntfile.js
```diff
@@ -5,6 +5,11 @@ camelcase: ["error", {"properties": "never"}]
 */
 var testConfig = require('./build/test/config');
 
+var defaultWebpackConfig = {
```
stephenmathieson: Thoughts on creating a "config factory" instead?

```js
const createWebpackConfig = (input, output) => ({
  devtool: false,
  mode: 'development',
  entry: path.resolve(__dirname, input),
  output: {
    filename: 'index.js',
    path: output
  }
})
```

Then we'd just call it like this:

```js
{
  webpack: {
    commonsUtils: createWebpackConfig('lib/commons/utils/index.js', 'tmp/commons/utils'),
    commonsForms: createWebpackConfig('lib/commons/forms/index.js', 'tmp/commons/forms'),
  }
}
```

lib/commons/forms/index.js
```diff
@@ -1,8 +1,23 @@
...
+import isNativeTextbox from './is-native-textbox';
+
 /**
  * Namespace for forms-related utilities.
  * @namespace commons.forms
  * @memberof axe
  */
-
-var forms = {};
+const forms = {
```
stephenmathieson: This was probably discussed in the proposal and I'm just not remembering the answer, but why aren't we `export`ing the `forms` object?

Should this be:

```
export { isAriaCombobox, isAriaListbox, isAriaRange, ... }

// or:

export default forms
```

straker: I don't think I specifically mentioned it, but I figured we would do

```js
export { default as isAriaCombobox } from './is-aria-combobox.js'
export { default as isAriaListbox } from './is-aria-listbox.js'
```

Then the main commons `index.js` file would group these as "forms"

stephenmathieson: Should we make this change now?

straker: We can't yet. I would need all the common subdirectors converted first so I can add them to the main commons object. The way we concat now is that concat the commons index file first and then every file adds their function to it. So I can't webpack the commons index file yet as that would break the code that expects commons to be a global variable.

stephenmathieson: Ahh, ok, that makes sense. Thanks for clarifying.

Issue #2090, opened 2020-03-10
False positive: 'color-contrast' rule fails disabled control's label added using "aria-labelledby" attribute


iamrafan:
Same as #1731, but here the label is added using ```aria-labelledby``` attribute

Per WCAG 1.4.3 the text that is part of an inactive user interface component do not have contrast requirement.

> Incidental: Text or images of text that are part of an inactive user interface component, that are pure decoration, that are not visible to anyone, or that are part of a picture that contains significant other visual content, have no contrast requirement.

Expectation: 'color-contrast' would not fail for labels related to disabled controls

Actual: 'color-contrast' rule fails for a label( added using ```aria-labelledby``` attribute) of a disabled control

Motivation: Text related to the disabled control is exempt from the contrast requirements

Demo link: https://developer.microsoft.com/en-us/fabric#/controls/web/combobox

![image](https://user-images.githubusercontent.com/22068434/76367465-5ec41700-62ea-11ea-8aa5-332c37d00193.png)

<pre><code>axe-core version: 3.5.1-canary.0ec5091
axe-Coconut version: 4.3.0</code></pre>

WilcoFiers:
Here's another example of this that shouldn't fail axe:

source: https://act-rules.github.io/rules/afw4f7#inapplicable-example-9
```html
<label id="my_pets_name" style="color:#888; background: white;">
	My pet's name
</label>
<div role="textbox" aria-labelledby="my_pets_name" aria-disabled="true"
	style="height:20px; width:100px; border:1px solid black;">
	test
</div>
```

padmavemulapati:
Verified with latest axe (axe-core3.5.5), color-contrast is failing which has no `aria-labelledby` attribute. and its passing which has `aria-labelledby`
Failed-snippet in the  Screenshot 
![image](https://user-images.githubusercontent.com/47409579/87685204-c0b4f180-c7a0-11ea-863c-153afbcc94cd.png)
Pass-snippet in the screenshot:
![image](https://user-images.githubusercontent.com/47409579/87685420-feb21580-c7a0-11ea-8779-47a72535bd77.png)

jeankaplansky:
doc'd in internal release notes on 7/20/2020

PR #2130, opened 2020-03-25
fix(color-contrast): ignore aria-disabled labels

WilcoFiers:
If an element is a descendent of an aria-labelledby reference, and any of the elements referencing it has disabled, the label must be ignored by the color-contrast rule.

Closes issue: #2090

## Reviewer checks


lib/rules/color-contrast-matches.js
```diff
@@ -30,107 +42,80 @@ if (formElements.includes(nodeName)) {
...
-// label of disabled control associated w/ aria-labelledby
-if (node.getAttribute('id')) {
-	const id = axe.utils.escapeSelector(node.getAttribute('id'));
-	const doc = axe.commons.dom.getRootNode(node);
-	const candidate = doc.querySelector('[aria-labelledby~=' + id + ']');
-	if (candidate && candidate.disabled) {
-		return false;
+const ariaLabelledbyControls = [];
+let ancestorNode = virtualNode;
+while (ancestorNode) {
```
straker: This is going to very slow on large pages as it will check the entire parent tree for every node (include subnodes of nodes already checked). The result of this gathering should be cached on each of the nodes checked so the next subnode only has to look up 1 level to find the answer.

See https://github.com/dequelabs/axe-core/blob/develop/lib/core/utils/is-hidden.js for a similar approach.

test/commons/forms/is-disabled.js
```diff
@@ -0,0 +1,125 @@
...
+		});
+
+		it('returns true when disabled has a value', function() {
+			fixtureSetup('<input type="text" disabled="yes" />');
+			var node = axe.utils.querySelectorAll(axe._tree[0], 'input')[0];
+
+			assert.isTrue(isDisabled(node));
+		});
+
+		it('returns true for in a disabled fieldset', function() {
```
straker:
```suggestion
		it('returns true when in a disabled fieldset', function() {
```

test/commons/forms/is-disabled.js
```diff
@@ -0,0 +1,125 @@
...
+			fixtureSetup(
+				'<fieldset disabled>' + '<span>Hello world</span>' + '</fieldset>'
+			);
+			var node = axe.utils.querySelectorAll(axe._tree[0], 'span')[0];
+
+			assert.isTrue(isDisabled(node));
+		});
+
+		it('returns true for in a disabled button', function() {
```
straker:
```suggestion
		it('returns true when in a disabled button', function() {
```

test/commons/forms/is-disabled.js
```diff
@@ -0,0 +1,125 @@
...
+
+	describe('with aria-disabled attr', function() {
+		it('returns true for an element with aria-disabled=true', function() {
+			fixtureSetup('<span aria-disabled="true">hello</span>');
+			var node = axe.utils.querySelectorAll(axe._tree[0], 'span')[0];
+
+			assert.isTrue(isDisabled(node));
+		});
+
+		it('returns true for an element with aria-disabled is not true', function() {
```
straker:
```suggestion
		it('returns false for an element when aria-disabled is not true', function() {
```

test/commons/forms/is-disabled.js
```diff
@@ -0,0 +1,125 @@
...
+				'<section aria-disabled="false">' +
+					'<div aria-disabled="true"><span>hello</span></div>' +
+					'</section>'
+			);
+			var node = axe.utils.querySelectorAll(axe._tree[0], 'span')[0];
+
+			assert.isTrue(isDisabled(node));
+		});
+
+		it('returns true if the closest ancestor with aria-disabled is set to false', function() {
```
straker:
```suggestion
		it('returns false if the closest ancestor with aria-disabled is set to false', function() {
```

test/rule-matches/color-contrast-matches.js
```diff
@@ -337,6 +337,26 @@ describe('color-contrast-matches', function() {
 		assert.isFalse(rule.matches(target, axe.utils.getNodeFromTree(target)));
 	});
 
+	it("should match not all input's are disabled - aria-labelledby", function() {
```
straker:
```suggestion
	it("should match when at least one input is not disabled - aria-labelledby", function() {
```

test/rule-matches/color-contrast-matches.js
```diff
@@ -351,6 +371,15 @@ describe('color-contrast-matches', function() {
 		assert.isFalse(rule.matches(target, axe.utils.getNodeFromTree(target)));
 	});
 
+	it("should not match a aria-disabled input's label - aria-labelledby", function() {
```
straker:
```suggestion
	it("should not match an aria-disabled input's label - aria-labelledby", function() {
```

test/rule-matches/color-contrast-matches.js
```diff
@@ -337,6 +337,26 @@ describe('color-contrast-matches', function() {
...
+		fixture.innerHTML =
+			'<div id="t1">Test</div>' +
+			'<input type="text" aria-labelledby="t1" disabled />' +
+			'<input type="text" aria-labelledby="t1" />';
+		var target = fixture.querySelector('div');
+		axe.testUtils.flatTreeSetup(fixture);
+		assert.isTrue(rule.matches(target, axe.utils.getNodeFromTree(target)));
+	});
+
+	it("should match not input's disabled by an ancestor - aria-labelledby", function() {
```
straker:
```suggestion
	it("should match when inputs are disabled by an ancestor - aria-labelledby", function() {
```

straker:
Also, we just merged a PR that updated the entire `commons.forms` to ES Modules. We'll probably need to update the new `forms.isDisabled` function to ES Modules before we merge this as well.

lib/commons/forms/is-disabled.js
```diff
@@ -0,0 +1,40 @@
...
+ * identifies an element as disabled, but not if all of them do.
+ *
+ * @method isDisabled
+ * @memberof axe.commons.forms
+ * @param {VirtualNode} virtualNode
+ * @return {boolean} whether or not the element is disabled in some way
+ */
+function isDisabled(virtualNode) {
+	let disabledState = virtualNode._isDisabled;
+	if (typeof disabled === 'boolean') {
```
straker, line 17: Did you mean `disabledState`?

```suggestion
	if (typeof disabledState === 'boolean') {
```

lib/rules/color-contrast-matches.js
```diff
@@ -43,109 +61,74 @@ function colorContrastMatches(node, virtualNode) {
...
+			explicitControl && getNodeFromTree(explicitControl);
+
+		if (explicitControlVirtual && isDisabled(explicitControlVirtual)) {
 			return false;
 		}
-
-		candidate = querySelectorAll(
-			relevantVirtualNode,
+		// implicit label of disabled control
+		const query =
```
straker, line 84: If we have an explicit control, do we need to check for an implicit one as well if the explicit control is not disabled?

test/rule-matches/color-contrast-matches.js
```diff
@@ -337,6 +337,26 @@ describe('color-contrast-matches', function() {
...
+		fixture.innerHTML =
+			'<div id="t1">Test</div>' +
+			'<input type="text" aria-labelledby="t1" disabled />' +
+			'<input type="text" aria-labelledby="t1" />';
+		var target = fixture.querySelector('div');
+		axe.testUtils.flatTreeSetup(fixture);
+		assert.isTrue(rule.matches(target, axe.utils.getNodeFromTree(target)));
+	});
+
+	it("should match when inputs are disabled by an ancestor - aria-labelledby", function() {
```
straker, line 350: Shouldn't this example not match as every `aria-lablledby` input is disabled by the parent fieldset so the `div` label should be treated as disabled?

lib/rules/color-contrast-matches.js
WilcoFiers: I don't think so. If an elm has an explicit label, the implicit one will not be applied.

test/rule-matches/color-contrast-matches.js
WilcoFiers: Jup, it should, and it would have if I hadn't messed up the test in two places :P

lib/rules/color-contrast-matches.js
straker: The way this `if` statement above is written, if the element has an explicit label and it's not disabled, it will continue to look at the implicit label. Based on your answer we would get into a state where if the explicit label is not disable but the implicit one is, we would ignore the element.

WilcoFiers: Thought about this, I think fixing this is well outside of the scope of this PR. I'll open an issue for it instead.

Issue #2076, opened 2020-03-02
Update rule `aria-required-children`


jkodu:
Expectation: 
For examples below, the rule should fail because required children role is not correct.

```
<div role="list">
	<div role="tabpanel">
		<div role="listitem">List item 1</div>
	</div>
</div>
```

```
<div role="list" aria-owns="item"></div>

<div id="host"></div>

<script>
	const host = document.querySelector('#host')
	const root = host.attachShadow({ mode: 'open' })
	root.innerHTML = '<div id="item" role="listitem">List item 1</div>'
</script>
```

Actual:
The rule only looks for children with an allowed role, without intercepting nodes in between which can have over-ridding roles.

Motivation:
Required for implementation of rule in ACT Rule - https://act-rules.github.io/rules/bc4a75

padmavemulapati:
Verified with latest axe-core, and with the above specified test example, aria-required-children rule is failing , which is expected.
![image](https://user-images.githubusercontent.com/47409579/78759799-d2b31700-799d-11ea-9bfd-0bb8dbade1e8.png)

straker:
No docs

PR #2131, opened 2020-03-25
fix(required-children): consider overriding descendant role(s) when validating required child role(s)

jkodu:
Closes issue:
- https://github.com/dequelabs/axe-core/issues/2076

## Reviewer checks

### DIFF
--- a/lib/checks/aria/required-children.js
+++ b/lib/checks/aria/required-children.js
@@ -65,7 +65,11 @@ function missingRequiredChildren(node, childRoles, all, role) {
 			 * `<div role="list"><div role="tabpanel"><div role="listitem">List item 1</div></div></div>`
 			 * should fail because `listitem` role not under `list` but has `tabpanel` between them, so although `listitem` is owned by `list` this is a failure.
 			 */
-			if (!descendantRoles.includes(childRole) && role !== 'combobox') {
+			if (
+				!ariaOwnsRole &&
+				role !== 'combobox' &&
+				!descendantRoles.includes(childRole)
+			) {
 				missing.push(childRole);
 			}
 		} else {
--- a/lib/checks/aria/required-children.js
+++ b/lib/checks/aria/required-children.js
@@ -202,7 +202,7 @@ function getDescendantRole(node, ownedEls) {
 	return Array.from(items).reduce((out, child) => {
 		const role = getRole(child);
 		if (['presentation', 'none', null].includes(role)) {
-			out.push(...getDescendantRole(child));
+			out = out.concat(getDescendantRole(child));
 		} else {
 			out.push(role);
 		}
--- a/lib/checks/aria/required-children.js
+++ b/lib/checks/aria/required-children.js
@@ -18,7 +18,16 @@ if (!childRoles) {
 	childRoles = required.all;
 }
 
-const missing = missingRequiredChildren(node, childRoles, all, role);
+const ownedElements = idrefs(node, 'aria-owns');
+const descendantRole = getDescendantRole(node, ownedElements);
+const missing = missingRequiredChildren(
+	node,
+	childRoles,
+	all,
+	role,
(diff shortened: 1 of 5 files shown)


straker:
Do we need to fail on this test case?

```html
<div role="list" aria-owns="list"></div>

<div id="list">
  <div role="tabpanel">
    <div role="listitem"></div>
  </div>
</div>
```

test/checks/aria/required-children.js
```diff
@@ -273,6 +315,24 @@ describe('aria-required-children', function() {
...
+	it('should pass one existing aria-owned child when one required', function() {
+		var params = checkSetup(
+			'<ul id="target"  role="tablist"><li role="tab">Tab 1</li><li role="tab">Tab 2</li></ul>'
+		);
+		assert.isTrue(
+			checks['aria-required-children'].evaluate.apply(checkContext, params)
+		);
+	});
+
+	it('should pass one existing aria-owned child when one required', function() {
```
straker: Probably should name these test cases so they are different. I had to look at the code to see what exactly made them different.

lib/checks/aria/required-children.js
```diff
@@ -106,49 +135,85 @@ function missingRequiredChildren(node, childRoles, all, role) {
...
-	reviewEmpty.includes(role) &&
-	!hasContentVirtual(virtualNode, false, true) &&
-	!hasDecendantWithRole(virtualNode) &&
-	idrefs(node, 'aria-owns').length === 0
-) {
-	return undefined;
-} else {
-	return false;
+	const items = isOwns
+		? Array.from(el.children).reduce(
```
straker, line 201: What exactly is this reduce doing? It looks like it's just pushing the children onto an array with the parent element. If that's the case it would be cleaner to just use concat.

```js
const items = (isOwns ? [el] : []).concat(Array.from(el.children));
```

lib/checks/aria/required-children.js
```diff
@@ -1,58 +1,87 @@
...
+}
 
-	selector = selector.join(',');
-	return ariaOwned
-		? matchesSelector(node, selector) ||
-				!!axe.utils.querySelectorAll(virtualTree, selector)[0]
-		: !!axe.utils.querySelectorAll(virtualTree, selector)[0];
+const ownedElements = idrefs(node, 'aria-owns');
+const descendantRole = getDescendantRole(node, ownedElements);
+const missing = missingRequiredChildren(
```
straker, line 23: Instead of upping the eslint `max-params` limit you could pass all these as a single object and destructure it.

```js
missingRequiredChildren({
  node,
  childRoles,
  all,
  role,
  ownedElements,
  descendantRole
});

function missingRequiredChildren({ 
  node,
  childRoles,
  all,
  role,
  ownedElements,
  descendantRole
}) {
  // ...
}
```

JustSlone:
@jeeyyy hey, question on this change. I'm not 100% sure, but this change *seems* to result in axe-core to flagging issues where `role="group"` is used to group `menuitems`. Was this intentional? 

Example that seems like it should be OK, but fails (Example pulled from: [The MDN Example #2 on using group role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_group_role))

```html
<div role="menu">
  <ul role="group">
    <li role="menuitem">Inbox</li>
    <li role="menuitem">Archive</li>
    <li role="menuitem">Trash</li>
  </ul>
  <ul role="group">
    <li role="menuitem">Custom Folder 1</li>
    <li role="menuitem">Custom Folder 2</li>
    <li role="menuitem">Custom Folder 3</li>
  </ul>
  <ul role="group">
    <li role="menuitem">New Folder</li>
  </ul>
</div>
```
[Repro gist for convience](https://gistpreview.github.io/?66ccb25097b58e80ac4f2f5a053f5a04)

PR #2158, opened 2020-04-09
refactor(core/base): Migrate to ES6 modules

stephenmathieson:
Ref #2117 

## Reviewer checks


lib/commons/aria/arialabel-text.js
```diff
@@ -6,8 +6,10 @@
  * @return {string} ARIA label
  */
 function arialabelText(node) {
+	// TODO: this is a ridiculous hack since webpack is making these two
+	// separate functions
 	// TODO: es-module-AbstractVirtualNode
-	if (node instanceof axe.AbstractVirtualNode === false) {
+	if (!node.constructor.name.includes('VirtualNode')) {
```
stephenmathieson: Does this still work when minfied?

straker: It didn't work in IE so I had to remove it and try something else

PR #2161, opened 2020-04-13
fix: typo in tags for avoid-inline-spacing rule

dylanb:
Changed `wcag1412` tag to the correct value: `wcag412`

## Reviewer checks


straker:
Is wcag 4.1.2 the right tag for this rule though? I'm not sure what inline spacing for stylesheet overrides has to do with name role value?

dylanb:
Shoot - correct - it is 1.4.12 - let me close this

PR #2162, opened 2020-04-13
feat: support check evaluate functions as an id string

straker:
Creating the stub for the map that will allow checks to use a string id to access shared evaluate and after functions. Once checks have been converted to es modules we can add them all to the map and have the feature working.

Closes issue:

## Reviewer checks

### DIFF
--- a/lib/core/base/metadata-function-map.js
+++ b/lib/core/base/metadata-function-map.js
@@ -1,0 +1,9 @@
+/*eslint no-unused-vars:0*/
+// TODO: es-modules-checks
+// TODO: es-modules-rules
+// import all check evaluate and after functions, and all rule matches functions
+
+const metadataFunctionMap = new Map();
+
+// TODO: es-modules-core
+// export default metadataFunctionMap;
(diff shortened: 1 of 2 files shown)


lib/core/base/metadata-function-map.js
```diff
@@ -0,0 +1,9 @@
+/*eslint no-unused-vars:0*/
+// TODO: es-modules-checks
+// TODO: es-modules-rules
+// import all check evaluate and after functions, and all rule matches functions
+
+const metadataFunctionMap = new Map();
```
stephenmathieson, line 6: I thought we were going to have a map for checks and a map for rules? How will we differenciate between the different types of functions here?

For example:

```
map.set('some-fn', (some, args) => { /* ... */ })
```

Is the above for a check `eval`, a check `after` or a rule `matches`?

straker: The id for each function will be unique and include the type in the name. E.g. `aria-allowed-attr-matches`, `aria-allowed-attr-evaluate`, `aria-allowed-attr-after`

PR #2167, opened 2020-04-17
refactor(core/public): Convert to ES modules

jkodu:
Refactoring `core/public` to modules.

> Note: I have skipped 5 tests which were attempting to mock some axe globals. These tests have to b refactored to actually test the unit of code that it is supposed to test. (Tried to fix these, but some of these depend on other globals that are also being refactored currently, i have added a todo:jey to remind myself that these needs updating).

Closes issue:
- NA

## Reviewer checks


lib/core/public/reporter.js
```diff
@@ -1,7 +1,11 @@
-var reporters = {};
-var defaultReporter;
+const reporters = {};
+let defaultReporter;
 
-axe.getReporter = function(reporter) {
+export const hasReporter = reporterName => {
```
WilcoFiers, line 4: This needs a test.

test/core/public/cleanup-plugins.js
```diff
@@ -30,7 +30,7 @@ describe('cleanupPlugins', function() {
 	it('should throw if no audit is configured', function() {
 		assert.throws(
 			function() {
-				cleanupPlugins(document, {});
+				axe.cleanup(document, {});
```
WilcoFiers, line 33: I think this file should be renamed to just cleanup.js

lib/core/public/load.js
```diff
@@ -1,6 +1,9 @@
-/*global Audit, runRules, cleanupPlugins */
+/*global Audit */
```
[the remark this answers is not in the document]
jkodu: Because `Audit` is not modularised yet - https://github.com/dequelabs/axe-core/pull/2158. I will add a todo note, so we do not forget to do such changes.

PR #2176, opened 2020-04-23
refactor(core): single webpack for core files

straker:
core-js-pure for some reason has an [empty file for the `some` and `reduce` functions](https://github.com/zloirock/core-js/blob/master/packages/core-js-pure/override/modules/es.typed-array.some.js) for a typed-array, so had to change them to an [MDN polyfill](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/some).

## Reviewer checks


test/core/public/run.js
```diff
@@ -101,21 +101,6 @@ describe('axe.run', function() {
 		});
 	});
 
-	it('should clear axe._tree', function(done) {
```
straker, line 104: This test is not part of the `run` code

test/core/utils/preload-media.js
```diff
@@ -17,26 +14,9 @@ describe('axe.utils.preloadMedia', function() {
 	});
 
 	afterEach(function() {
-		axe.utils.preloadMedia = origFn;
 		fixture.innerHTML = '';
 	});
 
-	it('invokes utils.preloadMedia and passes the treeRoot property', function(done) {
```
straker, line 24: This test does not test anything. It mocks a function and asserts that the function receives the passed in property

PR #2181, opened 2020-04-28
refactor: convert aria/checks to es modules

straker:
I figured out how to do the checks in piecemeal and still have the code work. So here is the aria checks converted to use the function map instead of build time compilation of the function. I also was able to reduce the webpack to just the single `core.js` entry point. 

I'm not sure why `build/test/engine.js` decided to override the `_load` function to set the check data instead of just getting it from axe directly. Needed to fix this so the tests would work in the new format.

Lastly, for some reason `commons` was able to be set from the Audit, but that would no longer work with importing functions directly (and would break our code anyway if it didn't have functions we expected). It's not documented anywhere in our code comments nor public documentation, so I removed it. @WilcoFiers will need your OK on that one.

## Reviewer checks

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -75,7 +75,8 @@
     {
       "files": [
         "lib/core/**/*.js",
-        "lib/commons/**/*.js"
+        "lib/commons/**/*.js",
+        "lib/checks/aria/**/*.js"
       ],
       "parserOptions":{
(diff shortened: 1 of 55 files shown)


WilcoFiers:
Steve asked me to OK that we take commons out of axe._load. It is :)

Issue #2195, opened 2020-04-28
Provide more options for non-generic check evaluate functions


WilcoFiers:
Adding these options will make it easier to configure these checks in custom rules:

* [x] Label/duplicate-img-label
    * Options for parent selector
* [x] Language/has-lang
    * Options for which attributes to look at
* [x] Language/valid-lang
    * Options for which attributes to look at
* [x] Navigation/region
    * Option for what elements are considered regions using a matcher
* [x] Shared/avoid-inline-spacing
    * Option for inline spacing css props
* [x] Tables/scope-value
    * Option for invalid values of scope

Other checks either already have options, they will be replaced with more general checks, or they will be configurable through commons.

straker:
QA: you'll need to look at each of the individual prs listed here to know what the options for each check is called.

padmavemulapati:
Seeing the changes on all Options (Options.value) and instead normal methods, observed getter methods and call funtions. Please let me know is there any specific impacts on tests/test results.

PR #2216, opened 2020-05-06
feat(duplicate-img-label): add option for parentSelector

straker:
Part of #2195. [findUpVirtual is deprecated](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/dom/find-up-virtual.js#L8) so switched it to axe.utils.closest.

## Reviewer checks


lib/checks/label/duplicate-img-label-evaluate.js
```diff
@@ -1,23 +1,21 @@
...
 	if (['none', 'presentation'].includes(getRole(node))) {
 		return false;
 	}
 
-	const parent = findUpVirtual(
-		virtualNode,
-		'button, [role="button"], a[href], p, li, td, th'
-	);
+	const {
+		parentSelector = 'button, [role="button"], a[href], p, li, td, th'
```
WilcoFiers, line 11: I think we should move all the defaults into the options object in the JSON, so that it is always in the same place and not some of it in options and some of it in evaluate.

PR #2238, opened 2020-05-18
fix(meta-viewport): don't throw error if viewport property doesn't have a value

mattzeunert:
Currently checking this meta viewport tag throws an error:

    <meta name="viewport" content="user-scalable=1, minimal-ui">

> TypeError: Cannot read property 'toLowerCase' of undefined

I think the issue was [introduced by this change](https://github.com/dequelabs/axe-core/pull/2137).

This PR fixes the issue by tolerating meta viewport properties without a value.

## Reviewer checks

### DIFF
--- a/lib/checks/mobile/meta-viewport-scale-evaluate.js
+++ b/lib/checks/mobile/meta-viewport-scale-evaluate.js
@@ -14,7 +14,7 @@ function metaViewportScaleEvaluate(node, options) {
 
 		const [key, value] = contentValue.split('=');
 		const curatedKey = key.toLowerCase().trim();
-		let curatedValue = value.toLowerCase().trim();
+		let curatedValue = value && value.toLowerCase().trim();
 
 		// convert `yes` to `1`
 		if (curatedKey === 'maximum-scale' && curatedValue === 'yes') {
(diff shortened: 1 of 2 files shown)


lib/checks/mobile/meta-viewport-scale-evaluate.js
```diff
@@ -14,7 +14,7 @@ function metaViewportScaleEvaluate(node, options) {
 
 		const [key, value] = contentValue.split('=');
 		const curatedKey = key.toLowerCase().trim();
-		let curatedValue = value.toLowerCase().trim();
+		let curatedValue = value && value.toLowerCase().trim();
```
stephenmathieson, line 17: This now makes this variable either hold a boolean or a string. This will be confusing when attempting to consume it later on.

Maybe we should do something like this instead:

```suggestion
		let curatedValue = (value || "").toLowerCase().trim();
```

mattzeunert: Thanks, good point 👍

lib/checks/mobile/meta-viewport-scale-evaluate.js
```diff
@@ -14,7 +14,7 @@ function metaViewportScaleEvaluate(node, options) {
 
 		const [key, value] = contentValue.split('=');
 		const curatedKey = key.toLowerCase().trim();
```
jkodu, line 16: Rather than creating a map (result) of key's with empty values, we should exit early when either key or value is `undefined` or `null` or has no length.

```
if(!key || !value) {
	return out
}
```

mattzeunert: Made that change now.

WilcoFiers:
@mattzeunert Thanks for the contribution!

Issue #2236, opened 2020-05-15
<aside> aria-labelledby a child <header> does not make it a unique landmark


gretencord:
Expectation: `<aside>` labeled by child `<header>` should pass. This fails:
```
<aside aria-labelledby="header-1">
  <header id="header-1">Header 1</header>
</aside>
<aside aria-labelledby="header-2">
  <header id="header-2">Header 2</header>
</aside>
```
Replacing `<header>` with `<div>` or `<aside>` with `<section`> passes. Using `aria-label="something"` instead of `aria-labelledby` also passes.

Actual: [`landmark-unique`](https://dequeuniversity.com/rules/axe/3.5/landmark-unique?application=AxeChrome) rule is thrown: "Landmarks must have a unique role or role/label/title (i.e. accessible name) combination"

Motivation: The issue appears to be with the `<header>` tag. From what I understand "The header element is not a banner landmark when it is a descendant of ... HTML sectioning elements" including `<aside>` [source](https://www.w3.org/TR/wai-aria-practices/examples/landmarks/HTML5.html). Regardless, I'm not sure why aXe will not accept it as the referenced label for its parent `<aside>`.

<pre><code>
axe-core version: 3.5.3
axe-webdriver, extension or other integration version: Web Accessibility Testing Chrome Extension 4.5.0

Browser and Assistive Technology versions: Windows 10 Chrome 81
</code></pre>

WilcoFiers:
Jup, you seem to be right. The issue is that axe-core still treats these header elements as landmarks; it shouldn't because they are a descendent of an aside element.

padmavemulapati:
Verified with axe-core3.5.5 , `<aside> aria-labelledby ..<header>` is not failing now and showing it in passes. Attached the screenshots describing passes and also voilations list(not having `<aside>`)
![image](https://user-images.githubusercontent.com/47409579/85378464-3807b480-b558-11ea-80eb-40b2d4ce68c4.png)
![image](https://user-images.githubusercontent.com/47409579/85378551-5372bf80-b558-11ea-8675-a750d3d1cff1.png)

jeankaplansky:
no rule-help changes required.

Issue #2212, opened 2020-05-06
build role conflict resolution into axe.commons.aria.getRole


straker:
This should replace the exception we build into axe.commons.text.titleText.

https://www.w3.org/TR/wai-aria-1.1/#conflict_resolution_presentation_none

straker:
See also https://github.com/w3c/aria/issues/1270

padmavemulapati:
when role=presentation and aria-attributes are properly defined no failures seeing: 
for  ex: `<h1 role="presentation" aria-haspopup="true"> Sample Content </h1> ` is passing no failure found
but ` <h1 role="presentation" aria-level="2"> Sample  Content </h1>` is failing for `aria-allowed-attr` - ruleid

jeankaplansky:
no user-facing docs required.

Issue #2144, opened 2020-04-02
Convert most APIs to accept a virtual node and return a virtual node


straker:
To help simplify the API, most functions should accept a virtual node and return a virtual node (as opposed to a DOM node). This deprecates functions like `findUp` which just get passed a DOM node and find the virtual node for it and continue to `findUpVirtual`. It also avoids problems such as `findUpVirtual` accepting a virtual node but returning a DOM node.

Most `commons` functions can probably do this, but functions in `utils` might not be able to (such as `getFlattenedTree` which creates the virtual tree). But as a general rule of thumb, a function should only take and return virtual nodes.

dylanb:
@straker is this still valid

straker:
We've been doing a lot of this work in https://github.com/dequelabs/axe-core/issues/2734, so we can close this one.

PR #2242, opened 2020-05-20
fix(implicit-roles): add proper implicit role calculation

straker:
We've had quite a few bugs in our implicit role calculations being incorrect so we wanted to rewrite it to be correct. Once this is in, we can finalize the role computations by adding [role conflict resolution](https://github.com/dequelabs/axe-core/issues/2212) and `role=presentation` inheritance.

The implicit roles were taking from https://www.w3.org/TR/html-aam-1.0/ and https://www.w3.org/TR/html-aria/.

Note: this probably could replace the `implicit` property from the roles table, but that is used by `implicit-nodes` and I didn't feel comfortable touching that yet. So we may need to make a tech debt ticket for that.

Closes issue: #2236

## Reviewer checks


lib/commons/aria/lookup-table.js
```diff
@@ -2079,6 +2094,132 @@ lookupTable.role = {
...
+	},
+	input: vNode => {
+		// Source: https://www.w3.org/TR/html52/sec-forms.html#suggestions-source-element
+		const listElement = idrefs(vNode.actualNode, 'list').filter(
+			node => !!node
+		)[0];
+		const suggestionsSourceElement =
+			listElement && listElement.nodeName.toLowerCase() === 'datalist';
+
+		switch (vNode.attr('type')) {
```
WilcoFiers, line 2160: We should toLowerCase() this if in an HTML doc.

lib/commons/aria/lookup-table.js
```diff
@@ -2079,6 +2094,132 @@ lookupTable.role = {
...
+				return !suggestionsSourceElement ? 'textbox' : 'combobox';
+			case 'number':
+				return 'spinbutton';
+			case 'radio':
+				return 'radio';
+			case 'range':
+				return 'slider';
+			case 'search':
+				return !suggestionsSourceElement ? 'searchbox' : 'combobox';
+		}
```
WilcoFiers, line 2182: What about if the type isn't any of these? I think that defaults to text, no?

lib/commons/aria/lookup-table.js
```diff
@@ -2079,6 +2094,132 @@ lookupTable.role = {
...
+			vNode,
+			'article, aside, main, nav, section, [role=article], [role=complementary], [role=main], [role=navigation], [role=region]'
+		);
+
+		return !sectioningElement ? 'contentinfo' : null;
+	},
+	form: vNode => {
+		// accessible-text-virtual uses getRole, so we have to only
+		// use the name calculation steps we know apply to a form
+		// element
```
WilcoFiers, line 2129: This isn't going to be sustainable. For us to properly implement presentational role conflict resolution, titleText is going to have to call getRole(), not just on the `alwaysTitleElements` list, but on all elements with a title. This is also relying on us continuing to have arialabelledby-text not work recursively, which could potentially change.

I think we should instead figure out a way to ensure this can never get into an infinite loop. How do browsers do this? I think it's worth doing some reverse engineering to figure out how browsers dealt with this.

lib/commons/aria/lookup-table.js
```diff
@@ -2079,6 +2094,132 @@ lookupTable.role = {
...
+		return arialabelledbyText(vNode) || arialabelText(vNode) || titleText(vNode)
```
WilcoFiers, line 2130: These results need to be sanitized.

lib/commons/aria/implicit-role.js
```diff
@@ -9,86 +9,25 @@ import lookupTable from './lookup-table';
...
-		const lookup = lookupTable.role[role];
-		return { name: role, implicit: lookup && lookup.implicit };
-	});
+	if (!vNode) {
+		return null;
+	}
 
-	/* Build a list of valid implicit roles for this node */
-	const availableImplicitRoles = roles.reduce(isValidImplicitRole, []);
+	const nodeName = vNode.props.nodeName.toLowerCase();
```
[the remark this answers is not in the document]
straker: If we are in an XHTML document then the nodeName may return as uppercase, which would not find anything in the structure.

lib/commons/aria/implicit-role.js
```diff
@@ -9,86 +9,25 @@ import lookupTable from './lookup-table';
  * @param {HTMLElement} node The node to test
  * @return {Mixed} Either the role or `null` if there is none
  */
-// TODO: axe 4.0 - rename to `getImplicitRole` so we can use
-// a variable named `implicitRole` when assigning the output of
-// the function
 function implicitRole(node) {
-	'use strict';
+	const vNode = getNodeFromTree(node);
```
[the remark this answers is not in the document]
straker: I was just going to make the accessible name lookup work with only vNodes to work with https://github.com/dequelabs/axe-core/issues/2144

lib/commons/aria/implicit-role.js
```diff
@@ -9,86 +9,25 @@ import lookupTable from './lookup-table';
...
-
-	/*
-	 * Create a list of { name / implicit } role mappings to filter on
-	 */
-	const roles = Object.keys(lookupTable.role).map(function(role) {
-		const lookup = lookupTable.role[role];
-		return { name: role, implicit: lookup && lookup.implicit };
-	});
+	if (!vNode) {
+		return null;
```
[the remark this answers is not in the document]
straker: Maybe? Some of our tests don't include the full tree (<body> got here for some of our tests and wasn't included in the tree). But it could also be the element is excluded from the context when doing a parent/child lookup.

lib/commons/aria/lookup-table.js
straker: `case null:` is that case.

lib/commons/aria/implicit-role.js
```diff
@@ -9,86 +9,25 @@ import lookupTable from './lookup-table';
...
+	}
 
-	/* Build a list of valid implicit roles for this node */
-	const availableImplicitRoles = roles.reduce(isValidImplicitRole, []);
+	const nodeName = vNode.props.nodeName.toLowerCase();
+	const role = lookupTable.implicitRole[nodeName];
```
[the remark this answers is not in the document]
straker: Can we add this after this pr is merged? SVG is a complicated beast that I'd rather tackle in a separate pr.

lib/commons/aria/implicit-role.js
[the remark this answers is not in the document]
WilcoFiers: What do you mean with "excluded from the context when doing a parent/child lookup"? I definitely don't think it's right to conclude that if we don't have it in the tree the role is null. Better to have it throw I think.

lib/commons/aria/implicit-role.js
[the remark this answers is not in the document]
WilcoFiers: If we are in XHTML, an `` is not the same as an ``. We shouldn't treat them as equivalent.

lib/commons/aria/implicit-role.js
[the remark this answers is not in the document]
WilcoFiers: The issue isn't that this doesn't include the SVG mapping. It's that the way this is down now will map SVG elements to things that it shouldn't be mapped to. An SVG `title` element shouldn't map to the HTML `title` element. At the very least this needs to be renamed to `implicitHtmlRole` and we need to check the namespace.

test/commons/aria/implicit-role.js
```diff
@@ -270,20 +285,57 @@ describe('aria.implicitRole', function() {
...
+		var node = fixture.querySelector('#target');
+		flatTreeSetup(fixture);
+		assert.equal(implicitRole(node), 'region');
+	});
+
+	it('should return region for section with accessible name title', function() {
+		fixture.innerHTML = '<section id="target" title="foo"></section>';
+		var node = fixture.querySelector('#target');
+		flatTreeSetup(fixture);
+		assert.equal(implicitRole(node), 'region');
```
WilcoFiers, line 307: This didn't work in Safari, right? If so I think we probably don't want to do that

lib/commons/aria/implicit-role.js
```diff
@@ -9,86 +9,37 @@ import lookupTable from './lookup-table';
...
+	// until we have proper implicit role lookups for svgs we will
+	// avoid giving them one
+	if (node.namespaceURI === 'http://www.w3.org/2000/svg') {
+		return null;
+	}
 
-	/* Build a list of valid implicit roles for this node */
-	const availableImplicitRoles = roles.reduce(isValidImplicitRole, []);
+	const nodeName = vNode.props.nodeName;
+	const role = lookupTable.implicitRole[nodeName];
```
WilcoFiers, line 32: I still think this is going to be a problem with SVG later. I'd think we need to rename this prop:

```suggestion
	const role = lookupTable.implicitHtmlRole[nodeName];
```

lib/commons/aria/lookup-table.js
```diff
@@ -2079,6 +2094,151 @@ lookupTable.role = {
...
+			(vNode.hasAttr('aria-labelledby') &&
+				sanitize(arialabelledbyText(vNode))) ||
+			(vNode.hasAttr('aria-label') && sanitize(arialabelText(vNode)))
+		)
+	);
+}
+
+// Source: https://www.w3.org/TR/html-aam-1.0/#element-mapping-table
+// Source: https://www.w3.org/TR/html-aria/
+lookupTable.implicitRole = {
```
WilcoFiers, line 2131:
```suggestion
lookupTable.implicitHtmlRole = {
```

Issue #2217, opened 2020-05-07
<ul> with role presentation being flagged by <li> rule


Brynman:
Expectation: << <li> elements must be contained in a `<ul>` or `<ol>`, should not flag elements with role `presentation` >>

Actual: << `<li>` elements nested in in parent `<ul>` with `role="presentation"` are failing the above check onhttps://www.waitrose.com/ecom/products/essential-waitrose-semi-skimmed-milk/053457-26759-26760

![image](https://user-images.githubusercontent.com/60389544/81285974-42c5c300-9058-11ea-9171-b06d58cbd5f5.png)
>>

Motivation: << Unless I am missing something this element is not exposed to AT and so it doesn't matter that the `<ul>` doesn't have a list role>>

<pre><code>
axe-core version: 4.4.1 (Plan: Professional axe-core 3.5.1)
axe-webdriver, extension or other integration version: NA

Browser and Assistive Technology versions: Safari 13.1.1 + VoiceOver

straker:
Thanks for the issue. It seems that we do not take into account `role=presentation | none` when determining the parent of an `<li>`.

padmavemulapati:
Verified with the fix - Not seeing Parent- children failures for `ui` `li` elements even with `role=presentation | none ` 
![image](https://user-images.githubusercontent.com/47409579/82628339-2f932580-9c0a-11ea-895f-0856f133e774.png)

jeankaplansky:
No user-facing rule help docs required.

PR #2281, opened 2020-06-08
feat(get-role): add presentation role resolution and inheritance

straker:
With proper conflict resolution, we can now set up rule `selector` to be a matcher which would allow us to only find elements that match their role. This would allow simpler resolutions, such as the bug #2217 could be fixed by having the `selector` be:

```json
{
  "selector": {
    "nodeName": "li",
    "role": "listitem"
  }
}
```

and since the presentation role on the `ul` or `ol` parent would change the role to `presentation` for the child, we would not have to look at the parent role in the check (the same for similar bug for `dl` elements).

Closes issue: #2212 

## Reviewer checks

### DIFF
--- a/lib/commons/aria/get-role.js
+++ b/lib/commons/aria/get-role.js
@@ -61,7 +61,15 @@ function getPresentationalAncestorRole(vNode, explicitRoleOptions) {
 		return ['presentation', 'none'].includes(role) ? role : null;
 	}
 
-	if (!vNode.parent || !chain.includes(vNode.parent.props.nodeName)) {
+	// if we can't look at the parent then we can't know if the node
+	// inherits the presentational role or not
+	if (!vNode.parent) {
+		throw new ReferenceError(
+			'Cannot determine role presentational inheritance of a required parent outside the current scope.'
+		);
+	}
+
+	if (!chain.includes(vNode.parent.props.nodeName)) {
 		return null;
 	}
 
--- a/lib/commons/aria/get-role.js
+++ b/lib/commons/aria/get-role.js
@@ -35,30 +35,20 @@ const inheritsPresentationChain = {
 	thead: ['table'],
 	tbody: ['table'],
 	tfoot: ['table'],
-	table: [],
 	li: ['ol', 'ul'],
-	ol: [],
-	ul: [],
-	dt: ['dl'],
-	dd: ['dl'],
-	dl: []
+	// dts and dds can be wrapped in divs and the div will pass through
+	// the presentation role
+	dt: ['dl', 'div'],
+	dd: ['dl', 'div'],
+	div: ['dl']
 };
 
 // role presentation inheritance.
 // Source: https://www.w3.org/TR/wai-aria-1.1/#conflict_resolution_presentation_none
-function getPresentationalAncestorRole(vNode, explicitRoleOptions) {
-	const chain = inheritsPresentationChain[vNode.props.nodeName];
-
-	if (!chain) {
+function getInheritedRole(vNode, explicitRoleOptions) {
+	const parentNodeNames = inheritsPresentationChain[vNode.props.nodeName];
+	if (!parentNodeNames) {
 		return null;
-	}
-
-	const role = getExplicitRole(vNode, explicitRoleOptions);
-	if (role && !hasConflictResolution(vNode)) {
-		// an explicit role of anything other than presentational will
-		// prevent any children from inheriting a presentational role
-		// from a valid ancestor
-		return ['presentation', 'none'].includes(role) ? role : null;
 	}
 
 	// if we can't look at the parent then we can't know if the node
@@ -69,11 +59,27 @@ function getPresentationalAncestorRole(vNode, explicitRoleOptions) {
 		);
 	}
 
-	if (!chain.includes(vNode.parent.props.nodeName)) {
+	// parent is not a valid ancestor that can inherit presentation
+	if (!parentNodeNames.includes(vNode.parent.props.nodeName)) {
 		return null;
 	}
 
-	return getPresentationalAncestorRole(vNode.parent);
+	const parentRole = getExplicitRole(vNode.parent, explicitRoleOptions);
+	if (
+		['none', 'presentation'].includes(parentRole) &&
+		!hasConflictResolution(vNode.parent)
+	) {
+		return parentRole;
+	}
+
+	// an explicit role of anything other than presentational will
+	// prevent any children from inheriting a presentational role
+	// from a valid ancestor
+	if (parentRole) {
+		return null;
+	}
+
+	return getInheritedRole(vNode.parent, explicitRoleOptions);
 }
 
 function resolveImplicitRole(vNode, explicitRoleOptions) {
@@ -83,10 +89,7 @@ function resolveImplicitRole(vNode, explicitRoleOptions) {
 		return null;
 	}
 
-	const presentationalRole = getPresentationalAncestorRole(
-		vNode,
-		explicitRoleOptions
-	);
+	const presentationalRole = getInheritedRole(vNode, explicitRoleOptions);
 	if (presentationalRole) {
 		return presentationalRole;
 	}
@@ -132,12 +135,9 @@ function getRole(node, { noImplicit, ...explicitRoleOptions } = {}) {
 	}
 
 	const explicitRole = getExplicitRole(vNode, explicitRoleOptions);
-	const implicitRole = noImplicit
-		? null
-		: resolveImplicitRole(vNode, explicitRoleOptions);
 
 	if (!explicitRole) {
-		return implicitRole;
+		return noImplicit ? null : resolveImplicitRole(vNode, explicitRoleOptions);
 	}
 
 	if (!['presentation', 'none'].includes(explicitRole)) {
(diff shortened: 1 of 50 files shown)


lib/commons/aria/get-role.js
```diff
@@ -1,5 +1,83 @@
...
+	tfoot: ['table'],
+	th: ['table'],
+	thead: ['table'],
+	tr: ['table']
+};
+
+function hasGlobalAriaAttributes(vNode) {
+	return lookupTable.globalAttributes.find(
+		attr => vNode.hasAttr(attr) && sanitize(vNode.attr(attr))
+	);
```
WilcoFiers, line 37: I tested this. The value doesn't matter. Even just doing `<button disabled role="none" aria-label>` triggers this behavior.

```suggestion
	return lookupTable.globalAttributes.find(attr => vNode.hasAttr(attr));
```

lib/commons/aria/get-role.js
```diff
@@ -32,14 +113,26 @@ function getRole(node, { noImplicit, fallback, abstracts, dpub } = {}) {
...
+
+	// role conflict resolution
+	// Source: https://www.w3.org/TR/wai-aria-1.1/#conflict_resolution_presentation_none
+	// See also: https://github.com/w3c/aria/issues/1270
+	if (
+		!explicitRole ||
+		(['presentation', 'none'].includes(explicitRole) &&
+			(hasGlobalAriaAttributes(vNode) || isFocusable(node)))
+	) {
+		return implicitRole;
```
WilcoFiers, line 132: So, here's a problem.... Chrome and Safari don't return the implicit role for certain elements. For example:

```html
<ul role="none" aria-label> <li>hello</li> </ul>
```

This is returned as "generic". Only Firefox actually returns this as a list. Buttons and links do get their role back. This seems to be exclusively based on the nodeName. We're going to need to look into what we want to do here.

lib/commons/text/title-text.js
```diff
@@ -1,15 +1,7 @@
...
-const alwaysTitleElements = [
-	'button',
-	'iframe',
-	'a[href]',
-	{
-		nodeName: 'input',
-		properties: { type: 'button' }
-	}
-];
+const alwaysTitleElements = ['iframe'];
```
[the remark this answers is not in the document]
straker: That's fine, but I think that's outside the scope of this pr. We should create another pr that fixes the iframe issue of isFocusable.

lib/commons/aria/get-role.js
```diff
@@ -22,17 +106,37 @@ import AbstractVirtuaNode from '../../core/base/virtual-node/abstract-virtual-no
...
+	}
+
+	// role conflict resolution
+	// note: Chrome returns a list with resolved role as "generic"
+	// instead of as a list
+	// (e.g. <ul role="none" aria-label><li>hello</li></ul>)
+	// we will return it as a list as that is the best option.
+	// Source: https://www.w3.org/TR/wai-aria-1.1/#conflict_resolution_presentation_none
+	// See also: https://github.com/w3c/aria/issues/1270
+	const hasGlobalAria = lookupTable.globalAttributes.find(attr =>
```
WilcoFiers, line 126: array.find returns the string, this returns a boolean.

```suggestion
	const hasGlobalAria = lookupTable.globalAttributes.some(attr =>
```

lib/commons/aria/get-role.js
```diff
@@ -22,17 +106,37 @@ import AbstractVirtuaNode from '../../core/base/virtual-node/abstract-virtual-no
...
+		vNode.hasAttr(attr)
+	);
+	if (hasGlobalAria || isFocusable(node)) {
+		// return null if there is a conflict resolution but no implicit
+		// has been set as the explicit role is not the true role
+		return noImplicit ? null : resolveImplicitRole(vNode);
+	}
+
+	if (!noImplicit && !explicitRole) {
+		return resolveImplicitRole(vNode);
```
WilcoFiers, line 136: Just for the sake of readability
```suggestion
	if (noImplicit && !explicitRole) {
		return null;
```

lib/commons/aria/get-role.js
```diff
@@ -22,17 +106,37 @@ import AbstractVirtuaNode from '../../core/base/virtual-node/abstract-virtual-no
...
 	}
 
-	return explicitRole || null;
+	return explicitRole;
```
WilcoFiers, line 139:
```suggestion
	return resolveImplicitRole(vNode);
```

lib/commons/aria/get-role.js
```diff
@@ -1,10 +1,94 @@
...
+// role presentation inheritance.
+// Source: https://www.w3.org/TR/wai-aria-1.1/#conflict_resolution_presentation_none
+function getPresentationalAncestorRole(vNode) {
+	const chain = inheritsPresentationChain[vNode.props.nodeName];
+
+	if (!chain) {
+		return null;
+	}
+
+	const role = getRole(vNode, { noImplicit: true });
```
WilcoFiers, line 56: We deprecated "noImplicit"
```suggestion
	const role = getExplicitRole(vNode);
```

lib/commons/aria/get-role.js
```diff
@@ -1,10 +1,94 @@
...
+	if (role) {
+		if (['presentation', 'none'].includes(role)) {
+			return role;
+		}
+
+		// an explicit role of anything other than presentational will
+		// prevent any children from inheriting a presentational role
+		// from a valid ancestor
+		return null;
```
WilcoFiers, line 65: Personal opinion: I think a ternary would read better here:

```js
return ['presentation', 'none'].includes(role)) ? role : null
```

lib/commons/aria/get-role.js
```diff
@@ -1,10 +1,94 @@
...
+	}
+
+	if (vNode.parent && chain.includes(vNode.parent.props.nodeName)) {
+		return getPresentationalAncestorRole(vNode.parent);
```
WilcoFiers, line 69: Swap this out with the return null at the bottom of the function. Recursion calls should generally always be at the end of the function. Kudos on the recursion idea BTW. Definitely prefer that over what I suggested.

lib/commons/aria/get-role.js
```diff
@@ -1,10 +1,94 @@
...
+			return role;
+		}
+
+		// an explicit role of anything other than presentational will
+		// prevent any children from inheriting a presentational role
+		// from a valid ancestor
+		return null;
+	}
+
+	if (vNode.parent && chain.includes(vNode.parent.props.nodeName)) {
```
WilcoFiers, line 68: I think this should throw if there's no parent. If we're testing an unmounted `li`, we can't determine the role. We shouldn't make assumptions about this.

lib/commons/aria/get-role.js
```diff
@@ -1,10 +1,115 @@
...
+	ol: [],
+	ul: [],
+	dt: ['dl'],
+	dd: ['dl'],
+	dl: []
+};
+
+// role presentation inheritance.
+// Source: https://www.w3.org/TR/wai-aria-1.1/#conflict_resolution_presentation_none
+function getPresentationalAncestorRole(vNode, explicitRoleOptions) {
```
WilcoFiers, line 49: This method did my head in. I had to rewrite it to understand it. Here's my version:

```js
function getInheritedRole(vNode, explicitRoleOptions) {
  const parentNodeNames = inheritsPresentationChain[vNode.props.nodeName];
  if (!parentNodeNames) {
    return null;
  }

  if (!vNode.parent) {
    throw new ReferenceError(
			'Cannot determine role presentational inheritance of a required parent outside the current scope.'
		);
  }

  // parent can't cause inherited presentation
  if (!parentNodeNames.includes(vNode.parent.props.nodeName)) {
    return null
  }

  // Check if parent is presentational:
  const parentRole = getExplicitRole(parent, explicitRoleOptions)
  if (['none', 'presentation'].includes(parentRole) && !hasConflictResolution(parent)) {
    return parentRole
  }

  return getInheritedRole(vNode.parent, explicitRoleOptions)
}
```

Couple improvements:
- it doesn't need the odd empty arrays above
- It doesn't do explicit role lookup on the first node. That lookup, more than anything else made this confusing to me.

lib/commons/aria/get-role.js
```diff
@@ -1,10 +1,115 @@
...
+	thead: ['table'],
+	tbody: ['table'],
+	tfoot: ['table'],
+	table: [],
+	li: ['ol', 'ul'],
+	ol: [],
+	ul: [],
+	dt: ['dl'],
+	dd: ['dl'],
+	dl: []
```
WilcoFiers, line 44: I just looked these up, as far as I can tell this stuff doesn't work with dl, which is good because otherwise we'd have to deal with div being allowed to group dd / dt elements.

```suggestion
```

test/commons/aria/get-role.js
```diff
@@ -219,12 +219,23 @@ describe('aria.getRole', function() {
 			var node = fixture.querySelector('#target');
 			assert.equal(aria.getRole(node), 'listitem');
 		});
+
+		it('throws an error if the tree is incomplete', function() {
+			fixture.innerHTML =
+				'<ul role="presentation"><li id="target">foo</li></ul>';
```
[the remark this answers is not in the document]
straker: Because I specifically made the tree have just the `<li>` to show that it will fail even if the parent exists in the DOM.

```js
flatTreeSetup(node);
```

lib/commons/aria/get-role.js
straker: It doesn't work in Chrome but does in Firefox, Safari/VO, and IE11/JAWS:

image

image

straker: I had assumed `<div>` would stop the chain as we've seen in other things, but after testing it does in fact carry through the role inheritance so we'll need to add that.

Issue #2261, opened 2020-06-03
"opacity: 0" elements are treated as visible


straker:
`utils.isHidden` and `dom.isVisible` do not take into account opacity. As such, a node with `opacity: 0` is visible and not hidden according to axe.

```html
<h1 id="target" style="opacity: 0">Title</h1>

<script>
axe.utils.isHidden(target);  // false
axe.commons.dom.isVisible(target);  // true
</script>
```

This leads to some interesting results when a combox dropdown menu is "hidden" using opacity. 

```html
<div style="color: gray; background-color: white; font-size: 14pt; font-weight: 100;">
  <span id="target" style="font-weight:bolder; opacity: 0;">My text</span>
</div>
```

Color contrast runs on the nodes and reports incomplete for nodes that are not visible on screen.

padmavemulapati:
Verified with the test file - 
`<div style="color: gray; background-color: white; font-size: 14pt; font-weight: 100;">
  <span id="target" style="font-weight:bolder; opacity: 0;">My text</span>
</div>`
when a combox dropdown menu is "hidden" using opacity. Color contrast runs on the nodes and reports incomplete for nodes that are not visible on screen. (attached  screenshot)

![image](https://user-images.githubusercontent.com/47409579/84401547-05d58900-ac21-11ea-8eef-618d6fc9f5e2.png)

Issue #2262, opened 2020-06-03
children of scroll regions with 0 height are considered visbile


straker:
A scroll region with 0 height will not display any of the contents, but all children (and the scroll region itself) are considered visible and so all rules run against those nodes even when not visible on screen (especially color contrast).

```html
<div style="color: gray; background-color: white; font-size: 14pt; font-weight: 100; overflow: scroll; height: 0">
  <span id="target" style="font-weight:bolder">My text</span>
</div>
```

This throws an error in color-contrast and returns as incomplete.

padmavemulapati:
Verified with the test file -
` <div style="color: gray; background-color: white; font-size: 14pt; font-weight: 100; overflow: scroll; height: 0">
  <span id="target" style="font-weight:bolder">My text</span>
</div> `
This throws an error in color-contrast and returns as incomplete.
![image](https://user-images.githubusercontent.com/47409579/84402105-ad52bb80-ac21-11ea-8816-e516d99052eb.png)

PR #2283, opened 2020-06-09
fix(is-visible): return false for opacity: 0 and 0 height scrollable regions

straker:
Also cleaned up some imports while I was there.

Closes issue: #2261 and #2262 

## Reviewer checks

### DIFF
--- a/lib/commons/dom/is-visible.js
+++ b/lib/commons/dom/is-visible.js
@@ -1,6 +1,12 @@
 import getRootNode from './get-root-node';
 import isOffscreen from './is-offscreen';
 import findUp from './find-up';
+import {
+	getScroll,
+	getNodeFromTree,
+	querySelectorAll,
+	escapeSelector
+} from '../../core/utils';
 
 const clipRegex = /rect\s*\(([0-9]+)px,?\s*([0-9]+)px,?\s*([0-9]+)px,?\s*([0-9]+)px\s*\)/;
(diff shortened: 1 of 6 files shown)


lib/commons/dom/is-visible.js
```diff
@@ -133,7 +136,10 @@ function isVisible(el, screenReader, recursed) {
 		 */
 		(nodeName !== 'AREA' && style.getPropertyValue('display') === 'none') ||
 		['STYLE', 'SCRIPT', 'NOSCRIPT', 'TEMPLATE'].includes(nodeName) ||
-		(!screenReader && isClipped(style)) ||
+		(!screenReader &&
+			(isClipped(style) ||
+				style.getPropertyValue('opacity') === '0' ||
+				(getScroll(el) && parseInt(style.getPropertyValue('height')) === 0))) ||
```
WilcoFiers, line 142: We need to clean up this if statement. A 10 line if condition seems a bit much, no?

lib/commons/dom/is-visible.js
```diff
@@ -101,8 +105,7 @@ function isAreaVisible(el, screenReader, recursed) {
  */
 function isVisible(el, screenReader, recursed) {
```
WilcoFiers, line 106: I'd like to see a few tests added either for the check or for the rule please.

Issue #2282, opened 2020-06-09
Error out when axe.configure is given bad values


WilcoFiers:
QA stumbled upon this issue by accident. The following code creates a rule with an undefined ID, which is really odd:

```js
axe.configure({ rules: ['foo'] })
```

The content of the rules array should be an object, not a string, but that's never caught. At the very least axe-core should throw when the id of a check / rule is undefined. The ID has to be a string.

jeankaplansky:
no user-facing does required.

PR #2293, opened 2020-06-15
fix(configure): validate rules and checks properties

straker:
Prevents passing `axe.configure({ rules: ['hello'] });`. Also checked the `checks` property as it had the same problem.

Closes issue: #2282

## Reviewer checks

### DIFF
--- a/lib/core/public/configure.js
+++ b/lib/core/public/configure.js
@@ -46,7 +46,7 @@ function configure(spec) {
 
 	if (spec.checks) {
 		if (!Array.isArray(spec.checks)) {
-			throw new Error('Checks property must be an array');
+			throw new TypeError('Checks property must be an array');
 		}
 
 		spec.checks.forEach(function(check) {
--- a/lib/core/public/configure.js
+++ b/lib/core/public/configure.js
@@ -68,7 +68,7 @@ function configure(spec) {
 
 		spec.rules.forEach(function(rule) {
 			if (!rule.id) {
-				throw new Error(
+				throw new TypeError(
 					`Configured rule "${rule}" is invalid. Rules must be an object with at least an id property`
 				);
 			}
--- a/lib/core/public/configure.js
+++ b/lib/core/public/configure.js
@@ -45,14 +45,34 @@ function configure(spec) {
 	}
 
 	if (spec.checks) {
+		if (!Array.isArray(spec.checks)) {
+			throw new Error('Checks property must be an array');
+		}
+
 		spec.checks.forEach(function(check) {
+			if (!check.id) {
+				throw new Error(
+					`Configured check "${check}" is invalid. Checks must be an object with at least an id property`
+				);
+			}
+
 			audit.addCheck(check);
 		});
 	}
 
 	const modifiedRules = [];
 	if (spec.rules) {
+		if (!Array.isArray(spec.rules)) {
+			throw new Error('Rules property must be an array');
+		}
+
 		spec.rules.forEach(function(rule) {
+			if (!rule.id) {
+				throw new Error(
+					`Configured rule "${rule}" is invalid. Rules must be an object with at least an id property`
+				);
+			}
+
 			modifiedRules.push(rule.id);
 			audit.addRule(rule);
 		});
--- a/lib/core/public/configure.js
+++ b/lib/core/public/configure.js
@@ -51,8 +51,10 @@ function configure(spec) {
 
 		spec.checks.forEach(function(check) {
 			if (!check.id) {
+				const checkStr =
+					typeof check === 'object' ? JSON.stringify(check) : check;
 				throw new TypeError(
-					`Configured check "${check}" is invalid. Checks must be an object with at least an id property`
+					`Configured check "${checkStr}" is invalid. Checks must be an object with at least an id property`
 				);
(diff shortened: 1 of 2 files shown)


lib/core/public/configure.js
```diff
@@ -45,14 +45,34 @@ function configure(spec) {
 	}
 
 	if (spec.checks) {
+		if (!Array.isArray(spec.checks)) {
+			throw new Error('Checks property must be an array');
```
stephenmathieson, line 49:
```suggestion
			throw new TypeError('Checks property must be an array');
```

lib/core/public/configure.js
```diff
@@ -45,14 +45,34 @@ function configure(spec) {
...
+		}
+
 		spec.checks.forEach(function(check) {
+			if (!check.id) {
+				throw new Error(
```
stephenmathieson, line 54:
```suggestion
				throw new TypeError(
```

lib/core/public/configure.js
```diff
@@ -45,14 +45,34 @@ function configure(spec) {
...
+			}
+
 			audit.addCheck(check);
 		});
 	}
 
 	const modifiedRules = [];
 	if (spec.rules) {
+		if (!Array.isArray(spec.rules)) {
+			throw new Error('Rules property must be an array');
```
stephenmathieson, line 66:
```suggestion
			throw new TypeError('Rules property must be an array');
```

lib/core/public/configure.js
```diff
@@ -45,14 +45,34 @@ function configure(spec) {
...
+		}
+
 		spec.rules.forEach(function(rule) {
+			if (!rule.id) {
+				throw new Error(
```
stephenmathieson, line 71:
```suggestion
				throw new TypeError(
```

lib/core/public/configure.js
straker: Good idea. Will fail the build after i merge so ill fix it with another commit once I merge these in.

lib/core/public/configure.js
```diff
@@ -45,14 +45,34 @@ function configure(spec) {
...
 
 	if (spec.checks) {
+		if (!Array.isArray(spec.checks)) {
+			throw new TypeError('Checks property must be an array');
+		}
+
 		spec.checks.forEach(function(check) {
+			if (!check.id) {
+				throw new TypeError(
+					`Configured check "${check}" is invalid. Checks must be an object with at least an id property`
```
WilcoFiers, line 55: This is going to stringify to "[object object]", that doesn't seem meaningful.

lib/core/public/configure.js
```diff
@@ -45,14 +45,34 @@ function configure(spec) {
...
 	const modifiedRules = [];
 	if (spec.rules) {
+		if (!Array.isArray(spec.rules)) {
+			throw new TypeError('Rules property must be an array');
+		}
+
 		spec.rules.forEach(function(rule) {
+			if (!rule.id) {
+				throw new TypeError(
+					`Configured rule "${rule}" is invalid. Rules must be an object with at least an id property`
```
WilcoFiers, line 72: As above.

lib/core/public/configure.js
```diff
@@ -45,14 +45,37 @@ function configure(spec) {
...
+		if (!Array.isArray(spec.rules)) {
+			throw new TypeError('Rules property must be an array');
+		}
+
 		spec.rules.forEach(function(rule) {
+			if (!rule.id) {
+				const ruleStr = typeof rule === 'object' ? JSON.stringify(rule) : rule;
+				throw new TypeError(
+					`Configured rule "${ruleStr}" is invalid. Rules must be an object with at least an id property`
+				);
```
stephenmathieson, line 76:
```suggestion
				throw new TypeError(
					`Configured rule ${JSON.stringify(rule)} is invalid. Rules must be an object with at least an id property`
				);
```

lib/core/public/configure.js
```diff
@@ -45,14 +45,37 @@ function configure(spec) {
...
+			throw new TypeError('Checks property must be an array');
+		}
+
 		spec.checks.forEach(function(check) {
+			if (!check.id) {
+				const checkStr =
+					typeof check === 'object' ? JSON.stringify(check) : check;
+				throw new TypeError(
+					`Configured check "${checkStr}" is invalid. Checks must be an object with at least an id property`
+				);
```
stephenmathieson, line 58: NBD, but we don't need the above ternary. Can just always stringify the "check":

```suggestion
				throw new TypeError(
					`Configured check ${JSON.stringify(check)} is invalid. Checks must be an object with at least an id property`
				);
```

lib/core/public/configure.js
```diff
@@ -45,14 +45,40 @@ function configure(spec) {
...
 	if (spec.checks) {
+		if (!Array.isArray(spec.checks)) {
+			throw new TypeError('Checks property must be an array');
+		}
+
 		spec.checks.forEach(function(check) {
+			if (!check.id) {
+				throw new TypeError(
+					// eslint-disable-next-line max-len
+					`Configured check ${JSON.stringify(
```
stephenmathieson, line 56: Prettier made this really ugly 🙃

PR #2294, opened 2020-06-15
feat(commons/aria): deprecate getRole({ noImplicit }) for getExplicitRole()

straker:
While working on #2281, one of the tests was failing because it wanted the explicit role but role conflict resolution was getting in the way and returning `null`. So created a function to get just the explicit role with no conflict resolution and made `getRole` the semantic role function. Deprecated the `noImplicit` option of `getRole` as that functionality is now covered by `getExplicitRole`.

Also made sure `getRole` could work with virtual nodes. 

Once this is merged, will pull it into #2281 to finish it off.

## Reviewer checks

### DIFF
--- a/lib/commons/aria/get-explicit-role.js
+++ b/lib/commons/aria/get-explicit-role.js
@@ -13,7 +13,7 @@ function getExplicitRole(vNode, { fallback, abstracts, dpub } = {}) {
 	const roleList = fallback ? tokenList(roleAttr) : [roleAttr];
 
 	// Get the first valid role:
-	const validRoles = roleList.filter(role => {
+	const firstValidRole = roleList.find(role => {
 		if (!dpub && role.substr(0, 4) === 'doc-') {
 			return false;
 		}
--- a/lib/commons/aria/get-explicit-role.js
+++ b/lib/commons/aria/get-explicit-role.js
@@ -1,0 +1,26 @@
+import isValidRole from './is-valid-role';
+import { getNodeFromTree, tokenList } from '../../core/utils';
+import AbstractVirtuaNode from '../../core/base/virtual-node/abstract-virtual-node';
+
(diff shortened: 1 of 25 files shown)


lib/commons/aria/get-explicit-role.js
```diff
@@ -0,0 +1,26 @@
...
+
+	if (vNode.props.nodeType !== 1) {
+		return null;
+	}
+
+	const roleAttr = (vNode.attr('role') || '').trim().toLowerCase();
+	const roleList = fallback ? tokenList(roleAttr) : [roleAttr];
+
+	// Get the first valid role:
+	const validRoles = roleList.filter(role => {
```
WilcoFiers, line 16:
```suggestion
	const firstValidRole = roleList.find(role => {
```

lib/commons/aria/get-explicit-role.js
```diff
@@ -0,0 +1,26 @@
...
+		if (!dpub && role.substr(0, 4) === 'doc-') {
+			return false;
+		}
+		return isValidRole(role, { allowAbstract: abstracts });
+	});
+
+	return validRoles[0] || null;
```
WilcoFiers, line 23:
```suggestion
	return firstValidRole || null;
```

lib/commons/aria/get-role.js
```diff
@@ -1,42 +1,35 @@
...
- * Return the accessible role of an element
+ * Return the semantic role of an element
  *
  * @method getRole
  * @memberof axe.commons.aria
  * @instance
- * @param {Element} node
+ * @param {Element|VirtualNode} node
  * @param {Object} options
  * @param {boolean} options.noImplicit  Do not return the implicit role
```
WilcoFiers, line 14:
```suggestion
 * @param {boolean} options.noImplicit  Do not return the implicit role // @deprecated
```

Issue #2105, opened 2020-03-21
False positive: page must have means to bypass...


dylanb:
Expectation: Not raise the "page must have a means to bypass repeated blocks" violation on a page with an iframe with a heading level 1 inside the iframe

Actual: False positive violation is raised

Motivation: False positive goes against the axe manifesto

Page that exhibits the behavior https://lion-web-components.netlify.com/?path=/docs/forms-form-examples--default-story

<pre><code>
axe-core version: 3.5.1

Browser and Assistive Technology versions

For Tooling issues:
Chrome extension v4.4.1, Mac OS X
</code></pre>

WilcoFiers:
It shouldn't matter what heading level it is, other than that, I agree.

padmavemulapati:
Verified with the iframe ` <!doctype html>
<html lang="bob">
<head>
	<title>Hello</title>
	<meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=0">
  <p>metaviewport</p>
</div>
</head>
</html>` and calling this into another iframe -- not throwing voilations or incomplete

jeankaplansky:
comment in release notes 7/28/2020

PR #2310, opened 2020-06-18
fix(bypass): find headings in iframes

straker:
I needed to modify the bypass matches function to only look for anchor elements on the top level. Inside iframes we should still run and find any bypassable blocks.

The iframe tests were mostly copied from `page-has-heading-one` integration tests.

Closes issue: #2105

## Reviewer checks


lib/rules/bypass.json
```diff
@@ -1,7 +1,6 @@
 {
 	"id": "bypass",
 	"selector": "html",
-	"pageLevel": true,
```
jkodu, line 4: Based on our discussion (pairing), this should not be removed?

See - https://github.com/dequelabs/axe-core/blob/4b99965b40ccb0d6ecfda620c1bbd697a504aa2e/doc/API.md#L214

lib/rules/bypass-matches.js
```diff
@@ -1,5 +1,14 @@
+import windowIsTopMatches from './window-is-top-matches';
+
 function bypassMatches(node) {
-	return !!node.querySelector('a[href]');
+	// the top level window should have an anchor
+	if (windowIsTopMatches(node)) {
+		return !!node.querySelector('a[href]');
```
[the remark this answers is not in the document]
straker: The original matches function required a page link before testing the entire page. Without one, the iframes (which don't have anchors) would not be tested for headings or landmarks. As to why we require the link in the first place, I don't know why we required that in the firest place

.github/axe-linter.yml
```diff
@@ -1,3 +1,3 @@
 exclude:
-  - ./CHANGELOG.md
-  - ./test/**/*
\ No newline at end of file
+  - CHANGELOG.md
+  - test/**/*
```
WilcoFiers, line 3: This needs a separate PR.

Issue #2303, opened 2020-06-17
Verify lookup table is up-to-date


straker:
The `aria.lookupTable` keeps having issues with [missing attributes](https://github.com/dequelabs/axe-core/issues/2170) or [allowed roles on elements](https://github.com/dequelabs/axe-core/issues/2288). We should do a one over pass on the lookup table with current ARIA spec and verify all the information is there.

straker:
Closed via the 3 standards prs.

Issue #2243, opened 2020-05-21
Organise the lookup table


WilcoFiers:
The ARIA lookup table has become massive. It is over 2000 lines long at this point. We should split that file up, put each table into its own file instead.

straker:
Closed via standards object

Issue #2108, opened 2020-03-23
Make global ARIA attributes configurable in aria-allowed-attr


WilcoFiers:
To make axe-core more flexible and allow it to support different environments it would be helpful to add more configuration options. One such option is to be able to set which global attributes are available in aria-allowed-attr, see https://github.com/dequelabs/axe-core/issues/2078 for details.

straker:
qa: to test this you can take a rule such as `aria-allowed-attrs` and test any role in the new [standards object](https://github.com/dequelabs/axe-core/blob/develop/lib/standards/aria-roles.js) that has an `allowedAttrs` property. First test that you can't add an aria-attribute that is not one 1) global 2) part of the `reuqiredAttrs` and 3) part of the `allowedAttrs`, then configure the role to allow the aria-attribute using `axe.configure`.

For example, the [checkbox role](https://github.com/dequelabs/axe-core/blob/develop/lib/standards/aria-roles.js#L59-L68) allows attributes `aria-checked`, `aria-readonly`, and `aria-required`. To test, you can add `aria-expanded` to an element that has `role=checkbox` and see that it should fail `aria-allowed-attrs` rule. Then configure the standards object to allow it and test it again, and it should pass.

```html
<div role="checkbox" aria-expanded="true" aria-checked="true"></div>

<script>
axe.configure({
  standards: {
    // the standards object to configure https://github.com/dequelabs/axe-core/blob/develop/lib/standards
    ariaRoles: {
      // the role to configure
      checkbox: {
        // the property to configure. with arrays you'll need to copy what's in the standards
       //  object and add the new values, otherwise the previous values will be removed
       // (meaning `aria-checked`, etc. would no longer be allowed)
        allowedAttrs: ['aria-checked', 'aria-readonly', 'aria-required', 'aria-expanded']
      }
    }
  }
});
</script>

await axe.run();  // so not show violation for `aria-allowed-attrs` for `aria-expanded` on the checkbox role
```

padmavemulapati:
Verified as per QA notes using latest axe-core(3.5.5 pulled code)
Step1: for the test file ' <div role="checkbox" aria-expanded="true" aria-checked="true">Hello</div>
Simply running axe - would result - failure for `aria-allowed-attr` (attached the Screenshot)
![image](https://user-images.githubusercontent.com/47409579/88069007-89728600-cb8e-11ea-9fb9-576c042f2c83.png)

Step2: configured with `axe.configure({
  standards: {
[...]

PR #2315, opened 2020-06-18
feat(standards): create standards object and ariaAttrs

straker:
I went through the spec for aria attributes and verified they were correct and updated anything that was different. Changes I made:

* Added the `global` property to any global aria attributes
* Made aria-hidden an `mntoken` that allows `undefined` (https://www.w3.org/TR/wai-aria-1.1/#aria-hidden)
* Removed boolean values as true/false are the only valid values (this will be fixed in `validate-attr-value` to check for `true/false` when we adopt the standards object in the codebase)
* Alphabetized values (the spec lists them alphabetically, except for `aria-haspopup` for some reason)
* Added `undefined` to `aria-orientation` (https://www.w3.org/TR/wai-aria-1.1/#aria-orientation)
* Changed `reference` token to `link` in `aria-dropeffect` (https://www.w3.org/TR/wai-aria-1.1/#aria-dropeffect)
* Removed `aria-describedat` as [it was removed from the spec in 2016](https://www.w3.org/TR/wai-aria-1.1/#changelog). Users can configure it back into the standards object if they need it.

To help keep the pr small, I am only creating the standards object here. I will create another pr to update the codebase to use it.

Part of issues: #2108, #2303, and #2243 

## Reviewer checks


lib/core/core.js
```diff
@@ -76,6 +77,7 @@ axe.reset = reset;
 axe._runRules = runRules;
 axe.runVirtualRule = runVirtualRule;
 axe.run = run;
+axe.standards = standards;
```
stephenmathieson, line 80: This will conflict with the Attest browser script, as it exposes its rulesets as `attest.standards`. Can we use a different name here?

stephenmathieson: Also, why make this part of the public API? Now that we have a module system, we can just `import` it when necessary.

stephenmathieson: For context, the Attest script does something like this (contrived/pseudocode):

```
import axe
import standards

var attest = {
  ...axe,
  standards,
  foo: 'bar'
}

export default attest
```

WilcoFiers: Agreed, there's no reason to put this on the global object.

test/core/public/configure.js
```diff
@@ -908,4 +908,54 @@ describe('axe.configure', function() {
 			});
 		});
 	});
+
+	describe('given a standards object', function() {
```
WilcoFiers, line 912: We should test that configuring something doesn't replace the other values on the standards object.

lib/core/public/standards.js
```diff
@@ -0,0 +1,18 @@
+import ariaAttrs from '../../standards/aria-attrs';
+import { clone } from '../utils';
+
+const standards = {
+	ariaAttrs
+};
+
+export function configureStandards(config) {
+	if (config.ariaAttrs) {
+		standards.ariaAttrs = Object.assign({}, clone(ariaAttrs), config.ariaAttrs);
```
WilcoFiers, line 10: Why are you cloning here?

lib/core/public/standards.js
```diff
@@ -0,0 +1,18 @@
+import ariaAttrs from '../../standards/aria-attrs';
+import { clone } from '../utils';
+
+const standards = {
+	ariaAttrs
+};
+
+export function configureStandards(config) {
```
WilcoFiers, line 8: I think this needs to do a deep copy. We don't want to lose props on objects. For example if you configure like this, I don't think that should cause "type" to get dropped.

```js
axe.configure({ standards: { ariaAttrs: { 'aria-disabled': { global: false } } } });
```

lib/core/public/standards.js
straker: To get a deep copy

lib/standards/index.js
```diff
@@ -0,0 +1,31 @@
...
+
+const origAriaAttrs = clone(ariaAttrs);
+const standards = {
+	ariaAttrs
+};
+
+// @see https://stackoverflow.com/a/59008477/2124254
+function merge(current, updates) {
+	for (const key of Object.keys(updates)) {
+		if (!current.hasOwnProperty(key) || typeof updates[key] !== 'object') {
```
WilcoFiers, line 12: We don't want to merge arrays. That'd get confusing. Stuff like:

```js
merge(['foo', 'bar'], ['bar']) // returns ['bar', 'bar'], but should return ['bar']
```

```suggestion
		if (!current.hasOwnProperty(key) || typeof updates[key] !== 'object' || Array.isArray(current[key])) {
```

lib/standards/index.js
```diff
@@ -0,0 +1,31 @@
...
+		} else {
+			merge(current[key], updates[key]);
+		}
+	}
+	return current;
+}
+
+export function configureStandards(config) {
+	if (config.ariaAttrs) {
+		standards.ariaAttrs = merge(ariaAttrs, config.ariaAttrs);
```
WilcoFiers, line 23: For safety, I think we should clone and freeze:

```suggestion
		const ariaAttrs = merge(ariaAttrs, clone(config.ariaAttrs))
		standards.ariaAttrs = deepFreeze(ariaAttrs);
```

lib/standards/index.js
```diff
@@ -0,0 +1,31 @@
+import ariaAttrs from './aria-attrs';
+import { clone } from '../core/utils';
+
+const origAriaAttrs = clone(ariaAttrs);
+const standards = {
+	ariaAttrs
```
WilcoFiers, line 6: For safety, I think we should freeze the contents of these things:

```suggestion
	ariaAttrs: deepFreeze(ariaAttrs)
```

lib/standards/index.js
```diff
@@ -0,0 +1,31 @@
...
+}
+
+export function configureStandards(config) {
+	if (config.ariaAttrs) {
+		standards.ariaAttrs = merge(ariaAttrs, config.ariaAttrs);
+	}
+}
+
+export function resetStandards() {
+	standards.ariaAttrs = origAriaAttrs;
```
WilcoFiers, line 28:
```suggestion
	standards.ariaAttrs = deepFreeze(origAriaAttrs);
```

straker: If we deep freeze then we can't merge in new properties. The standards `ariaAttrs` need to allow users to change it so we can pull it into other parts of the code and use the modified object.

lib/standards/index.js
straker: If we deep freeze the object then we can't add in new properties. I cloned the original so we could override `ariaAttrs` without worries.

lib/standards/index.js
```diff
@@ -9,7 +9,7 @@ const standards = {
 // @see https://stackoverflow.com/a/59008477/2124254
 function merge(current, updates) {
 	for (const key of Object.keys(updates)) {
-		if (!current.hasOwnProperty(key) || typeof updates[key] !== 'object') {
+		if (!current.hasOwnProperty(key) || typeof updates[key] !== 'object' || Array.isArray(current[key])) {
```
WilcoFiers, line 12: The array thing needs a test.

Issue #601, opened 2017-11-03
New rule: flag nested interactive roles


marcysutton:
# Nested Interactive Roles

Ensures interactive controls are not nested so they work with keyboards and screen readers

**Tags:** best-practice, experimental, cat.keyboard

## Help text

Ensure interactive controls are not nested.

## Selector

`a, button, [role="link"], [role="button"]`

## More info
The code sample below shows a few places where aXe should have flagged nested interactive roles. The developer in this case added a role of link with tabindex, nested a link with role=button and tabindex inside of it, and then put an input with type of image inside. That's a triple-whammy for nested controls, that we should be able to catch pretty easily since it doesn't even use click events.

```
<div role="link" tabindex="40" title="link label">
  <a href="/url-went-here" title="link label" role="button" tabindex="12">
    Link label
    <div class="link-icon">
        <input type="image" alt="add to stop throwing error">
    </div>
  </a>
</div>
```

### Checks

- none: nested-interactives

WilcoFiers:
Can you write this in the rules format?
https://github.com/dequelabs/axe-core/blob/develop/doc/rule-proposal.md

marcysutton:
I updated it partially, but it still needs the checks to be thought through. I wanted to capture it now so it wasn't forgotten.

marcysutton:
Here's a Codepen showing what happens with nested links. The browser automatically reformats nested anchor tags, but not when ARIA role="link" is wrapping an anchor tag. https://codepen.io/marcysutton/pen/YaxJVm?editors=1010#

Alternative rule name: `dont-do-it` 😂

dylanb:
What is the accessibility impact of this? The keyboard focus can move from the one element to the next. The user can interact with both.

If there is no real impact, then this should be a best practice.

marcysutton:
The accessibility impact is that it's extremely confusing to screen reader users to hear "Link label button link" or an empty "button" because of misused HTML and ARIA. Mixing of interface controls makes them extremely hard to understand from a screen reader user's perspective, even if they are coincidentally still operable.

As one example, when the browser changes your markup from `<button><button>Text</button</button>` to `<button></button><button>Text</button>`, you end up with an unlabeled button. We could nip these problems in the bud with a dedicated rule.

dylanb:
The unlabelled button problem will be found by the `button-name` rule (the beauty of testing the DOM)

Potential confusion problems as well as event handler propagation problems are not candidates for zero-false-positive detection

[...]

PR #2325, opened 2020-06-25
feat(standards): add html elems object

straker:
**Note:** the object is incomplete and this pr is here to start discussing the object and structure. Since the object is so large multiple smaller prs that can be easier to verify the information is correct would be better.

The lookup table has a few different objects that try to convey HTML spec related things. `elementsAllowedAnyRole`, `elementsAllowedNoRole`, and `evaluateRoleForElement`. `evaluateRoleForElement` is a bunch of functions that can't even be serialized. Also, there are a few commons functions that need HTML spec information such as `text.nativeElementTypes` and `subtree-text.js` needing to know phrasing content type elements.

The aria `roles` object has a property `allowedElements` which is an inverse property lookup to the spec table found in https://www.w3.org/TR/html-aria/#docconformance. This means that to generate the list of allowed elements, you need to find each element the role is allowed on rather than looking at the element row and seeing which roles are allowed. This makes it extremely easy to miss things, for example:

* https://github.com/dequelabs/axe-core/pull/2326
* https://github.com/dequelabs/axe-core/pull/2300

This object puts all that information into a serializable standards object that can be configured as well. It removes the need for `evaluateRoleForElement` and makes that information configurable. 

The object is defined as such and the information was taken from https://www.w3.org/TR/html-aria/#docconformance:

```ts
type Spec = {
  matches?: matcherObj, // if the spec should only apply if it matches the element
  allowedRole: boolean, // if the element is allowed a role
  allowedAttrs: boolean, // if the element is allowed global aria attrs and role attrs
  roles?: string[], // if only specific roles are allowed. if missing implies any role is allowed when allowedRole:true
  implicitAttrs:?: string[], // implicitly defined attrs per https://www.w3.org/TR/html-aam-1.0/#html-element-role-mappings
  namingMethods?: string | string[] // native-element-type.js naming methods
}

htlmEmls = {
  nodeName:string : Spec | {
    features: Spec[] 
  }
}
```

It also simplifies and enhances the check for `aria-allowed-attr` and `aria-allowed-role`. For example, the `is-aria-role-allowed-on-element` function has to perform 3 different matches functions looking at the `elementsAllowedAnyRole`, `elementsAllowedNoRole`, `allowedElements` and `evaluateRoleForElement`. Instead, it can just look at the HTML spec table and see which roles are allowed on the element:

```js
isAriaRoleAllowedOnElement(vNode, role) {
  const nodeName = vNode.props.nodeName; 
  const implicitRole = getImplicitRole(vNode);

  let spec = standards.htlmElms[nodeName]; 
  if (spec.features) {
    spec = matches(vNode, spec.features);
  }

  // always allow the explicit role to match the implicit role
  if (role === implicitRole) {
    return true;
  }

  // element is not allowed a role
  if (!spec.allowedRole) {
    return false;
  }

  // element is allowed only certain roles
  if (spec.roles) {
    return spec.roles.includes(role);
  }

  // element is allowed any role
  return true;
}
```

The table also allows us to catch elements which aren't allowed global aria elements in `aria-allowed-attr`. For example, `<col aria-label="foo"></col>` should not pass as `col` is not allowed any aria attributes. 

```js
ariaAllowedAttrEvaluate(node) {
  let spec = standards.htlmElms[nodeName]; 
  if (spec.features) {
    spec = matches(vNode, spec.features);
  }

  if (!spec.allowedAttrs) {
    return false;
  }

  // ...
}
```

Lastly, the object defines `implicitAttrs` which allow us to remove the ["hacks" in `ariaRequiredAttrEvaluate` ](https://github.com/dequelabs/axe-core/blob/develop/lib/checks/aria/aria-required-attr-evaluate.js#L17-L40) by defining which HTML elements define default aria attributes. This means `<h1 role="heading">foo</h1>` should pass since the `h1` element defaults `aria-level` [to the level of the element](https://www.w3.org/TR/html-aam-1.0/#html-element-role-mappings).

We'll need to talk about how to [handle the content types of each element](https://html.spec.whatwg.org/multipage/dom.html#kinds-of-content). Having them in place would greatly help for #601 as knowing which elements are interactive content types would let us know to flag them. However, they get a bit messy in their requirements of when it's interactive or not and duplicates a some of the properties when that's the only difference between them. For example, `audio` is an interactive element only when it has a `controls` attribute, but the other properties of `allowedRole` and `allowedAttrs` is the same regardless.

Personally, I think it should be in it's own object so we don't have to conflict with this object. For example: `img` is an interactive content model if the `usemap` attribute is present, but should be part of the 3 other features of allowed roles or attrs. Doing this would require 6 different objects to ensure the combinations are intact, which is a bit much.

## Reviewer checks


lib/standards/html-elms.js
```diff
@@ -0,0 +1,507 @@
...
+	a: {
+		features: [
+			{
+				matches: {
+					attributes: {
+						href: '/.*/'
+					}
+				},
+				allowedRole: true,
+				allowedAttrs: true,
```
WilcoFiers, line 13: I think a simpler way to do this is to use "noAriaAttrs", we'd only have to specify the prop on a few roles.

lib/standards/html-elms.js
```diff
@@ -0,0 +1,507 @@
...
+const htmlElms = {
+	a: {
+		features: [
+			{
+				matches: {
+					attributes: {
+						href: '/.*/'
+					}
+				},
+				allowedRole: true,
```
WilcoFiers, line 12: Lets combine this with the roles prop:

```ts
type allowedRoles = boolean | string[]
```

lib/standards/html-elms.js
```diff
@@ -0,0 +1,415 @@
...
+						type: 'radio'
+					}
+				},
+				allowedRoles: ['menuitemradio'],
+				implicitAttrs: ['aria-checked']
+			},
+			{
+				matches: {
+					attributes: {
+						type: 'text',
```
WilcoFiers, line 384: This is a bit of a problem... something with an invalid type, or without a type will default to type text. That's not captured with this logic.

straker:
This pr got messed up somewhere in the merging and branching. Going to open another clean one.

PR #2332, opened 2020-06-29
feat(standards): add dpub-roles spec

straker:
I generating the list automatically using a custom script on the 1.0 spec page. I then created a script to diff the generated file against our current lookup table and modified the output as needed to match (putting comments where and why changes were made).

The final output of the diff contains things I believe we are OK with but am putting here for reference. They are mostly additional aria allowed attrs that the current lookup table did not have but the spec allows. Seems we missed 2 name from content roles and a single allowed attr.

```
====== doc-glossref.nameFromContent ======
* doc-glossref.nameFromContent does not exist in the lookupTable
{ AssertionError [ERR_ASSERTION]: doc-glossref.nameFromContent does not match
    at file:///Users/stevenlambert/deque/axe-core/diff.mjs:49:14
    at ModuleJob.run (internal/modules/esm/module_job.js:95:12)
  generatedMessage: false,
  name: 'AssertionError [ERR_ASSERTION]',
  code: 'ERR_ASSERTION',
  actual: true,
  expected: undefined,
  operator: 'deepEqual' }


====== doc-noteref.nameFromContent ======
* doc-noteref.nameFromContent does not exist in the lookupTable
{ AssertionError [ERR_ASSERTION]: doc-noteref.nameFromContent does not match
    at file:///Users/stevenlambert/deque/axe-core/diff.mjs:49:14
    at ModuleJob.run (internal/modules/esm/module_job.js:95:12)
  generatedMessage: false,
  name: 'AssertionError [ERR_ASSERTION]',
  code: 'ERR_ASSERTION',
  actual: true,
  expected: undefined,
  operator: 'deepEqual' }


====== doc-pagebreak.allowedAttrs ======
{ AssertionError [ERR_ASSERTION]: doc-pagebreak.allowedAttrs does not match
    at file:///Users/stevenlambert/deque/axe-core/diff.mjs:49:14
    at ModuleJob.run (internal/modules/esm/module_job.js:95:12)
  generatedMessage: false,
  name: 'AssertionError [ERR_ASSERTION]',
  code: 'ERR_ASSERTION',
  actual: [ 'aria-expanded', 'aria-orientation' ],
  expected: [ 'aria-expanded' ],
  operator: 'deepEqual' }
```

Part of issues: #2108, #2303, and #2243

## Reviewer checks


straker:
Ha. I didn't know how to load it as a js file using imports in my node script so had them as `.mjs` extensions so it would work. Forgot to name it back.

PR #2333, opened 2020-06-29
feat(standards): add html-elms spec

straker:
Continued from #2325 to get a clean tree instead of a botched file merge of 34 file changes. 

This table replaces some hardcoded values in our code and allows us to work on #1597 and #601 by allowing us to identify interactive content.

Replaces these values: 
* `phrasingElements` in [subtree-text](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/text/subtree-text.js#L31-L72)
* `htmlTags` in [is-html-element](https://github.com/dequelabs/axe-core/blob/develop/lib/core/utils/is-html-element.js#L1-L120) (which is why there are empty objects)
* `possibleShadowRoots` in [is-shadow-root.js](https://github.com/dequelabs/axe-core/blob/develop/lib/core/utils/is-shadow-root.js#L1-L20)
* part of the `sectioningElementSelector` in [lookup-table.js](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/aria/lookup-table.js#L2101-L2102)
* all of [native-element-type.js](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/text/native-element-type.js)

This will also allow us to replace `svg-namespace-matches` and `html-namespace-matches` with this table by looking up the element name in the table to get us 90% of the way there. Two elements such as `a` and `svg` belong to both tables, so we will have a special case for those.

There might be a few more but those were the ones I could find in a quick element search.

Just like we plan to do for the `aria-attributes` spec, we should create a commons helper that aggregates information from this table into useful forms, such as getting all elements that are of a certain content type.
## Reviewer checks

### DIFF
--- a/lib/standards/html-aria.js
+++ b/lib/standards/html-aria.js
@@ -1,0 +1,715 @@
+// Source: https://www.w3.org/TR/html-aria/#allowed-aria-roles-states-and-properties
+// Source: https://www.w3.org/TR/html-aam-1.0/#html-element-role-mappings
+const htmlElms = {
+	a: {
+		variant: {
+			href: {
+				matches: '[href]',
+				allowedRoles: [
+					'button',
+					'checkbox',
+					'menuitem',
+					'menuitemcheckbox',
+					'menuitemradio',
+					'option',
+					'radio',
+					'switch',
+					'tab',
+					'treeitem',
+					'doc-backlink',
+					'doc-biblioref',
+					'doc-glossref',
+					'doc-noteref'
+				]
+			},
+			noHref: {
+				matches: {
+					attributes: {
+						href: null
+					}
+				},
+				allowedRoles: true
+			}
+		}
+	},
+	abbr: {
+		allowedRoles: true
+	},
+	addres: {
+		allowedRoles: true
+	},
+	area: {
+		allowedRoles: false
+	},
+	article: {
+		allowedRoles: [
+			'feed',
+			'presentation',
+			'none',
+			'document',
+			'application',
+			'main',
+			'region'
+		]
+	},
+	aside: {
+		allowedRoles: [
+			'feed',
+			'note',
+			'presentation',
+			'none',
+			'region',
+			'search',
+			'doc-dedication',
+			'doc-example',
+			'doc-footnote',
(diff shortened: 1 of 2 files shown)


lib/standards/html-aria.js
```diff
@@ -0,0 +1,715 @@
...
+			'region',
+			'search',
+			'doc-dedication',
+			'doc-example',
+			'doc-footnote',
+			'doc-pullquote',
+			'doc-tip'
+		]
+	},
+	audio: {
```
WilcoFiers, line 70: This one looks wrong. From what I can tell the controls don't matter here.

lib/standards/html-aria.js
```diff
@@ -0,0 +1,715 @@
...
+			},
+			radio: {
+				matches: '[type="radio"]',
+				allowedRoles: ['menuitemradio'],
+				implicitAttrs: {
+					'aria-checked': 'false'
+				}
+			},
+			textWithList: {
+				matches: '[type="text"][list]',
```
WilcoFiers, line 377: This won't work if it's an input without a type, or with an invalid type.

lib/standards/html-aria.js
```diff
@@ -0,0 +1,715 @@
...
+				allowedRoles: false
+			},
+			// Note: this covers type=text and any other input type
+			default: {
```
WilcoFiers, line 381: :+1:

lib/standards/html-aria.js
WilcoFiers: This doesn't seem like it got resolved.

lib/standards/html-aria.js
```diff
@@ -0,0 +1,707 @@
+// Source: https://www.w3.org/TR/html-aria/#allowed-aria-roles-states-and-properties
+// Source: https://www.w3.org/TR/html-aam-1.0/#html-element-role-mappings
+const htmlAria = {
```
WilcoFiers, line 3: I preferred this when it was "htmlElms", see: https://github.com/dequelabs/axe-core/pull/2337#pullrequestreview-440760860

Issue #246, opened 2016-10-30
Contrast analyser doesn't factor in CSS shadow


DavidMacDonald:
A client who met contrast minimums by creating a 360 degree shadow around text. aXe measured foreground text and background but did not factor in the shadow... this would be tricky programming to remedy. It would require ensuring the shadow surrounds the text and that the shadow is sufficient, but I think it's worth investigating further, because I've had to over ride the aXe report.

marcysutton:
We should be able to check for a CSS text-shadow if it's directly on the element (and not a pseudo-element). But how much blur, spread, etc. makes the shadow readable? We'd have to compare those values along with shadow color to the text and background colors to see if there is enough of a difference.

Volker-E:
We've been running into this question ourselves without finding a distinct boundary.
Blur < 5px and offsets < 2px seemed to be approximations.

dylanb:
@DavidMacDonald how did you determine that the contrast was sufficient using the shadow?

@Volker-E (side note: disabled elements do not have to meet the contrast minimums)

DavidMacDonald:
That's where I use the Colour Contrast Analyser eyedroppers :)

dylanb:
@DavidMacDonald What I mean by that question is, what relationship between the font-size/weight and the shadow-size leads to the same ability to read the text at the threshold contrast ratio? Do you have any research that backs up the claim that the shadow contrast is equivalent to the same background color?

DavidMacDonald:
I have no research ... just the plain fact that it passes WCAG. There is no distinction between shadow and background. The text needs to have sufficient colour with its adjacent color (or outline) which we call background. When CSS shadow is used, it becomes the background as long as it is at 1px then it is background.

dylanb:
@DavidMacDonald how do you determine whether the shadow definition actually results in a background that is at least 1px wide around the entire letter?

WilcoFiers:
@DavidMacDonald I don't think it's quite as easy as you're saying it. Check out this codepen:
https://codepen.io/wilcofiers/pen/Kmxrra

In theory, all of these meet the color contrast, if you take the color of the text-shadow in place of the background. But offset and blur radius impact that quite a bit. Can you make a suggestion on how to deal with those?

DavidMacDonald:
I agree, it would take some clever programming, and no one has done that before... some investigation of threshold blur etc... it would make you guys rock stars :)

[...]

PR #2334, opened 2020-06-30
fix(color-contrast): account for text-shadow

WilcoFiers:
Take text-shadow into account when computing the background color of a text node.

Related #246

## Reviewer checks


lib/commons/color/color.js
```diff
@@ -83,6 +107,31 @@ function Color(red, green, blue, alpha) {
 		}
 	};
 
+	this.parseHexString = function(colorString) {
```
stephenmathieson, line 110: NBD either way, but now that we can `import` things, I'm curious why you didn't grab something from npm rather than building this yourself.

lib/standards/css-colors.js
```diff
@@ -0,0 +1,147 @@
+// Source: https://www.w3.org/TR/css-color-3/#svg-color
+const cssColors = {
```
stephenmathieson, line 2: There are a few packages available from npm that have these listed in them already. Do we want to be responsible for keeping this list up-to-date, or should we offload that responsibility to the maintainers of one of the existing packages?

stephenmathieson: Also, this is out-of-date. For example ["rebeccapurple" has been added](https://drafts.csswg.org/css-color/#valdef-color-rebeccapurple).

lib/commons/color/color.js
WilcoFiers: Security / package size. I know this does exactly what we need, and nothing else.

lib/standards/css-colors.js
WilcoFiers: Yer right, we should use CSS 4 stuff here. As for why not NPM. Partially because if we do this ourselves we get exactly what we want, and it is more secure.

lib/commons/color/color.js
stephenmathieson: Our bundler should be tree-shaking any unused dependency code out for us, and we would "bundle" it at build-time so it'd be easier to analyze the sources. I understand the security concern, but am just worried about adding more to the list of complex things we're responsible for maintaining. Not worth blocking this PR, but I thought once we'd migrated to modules, we'd stop "reinventing the wheel".

lib/standards/css-colors.js
stephenmathieson: Are we going to remember to update this when changes are made to the spec?

WilcoFiers: I'm cued into W3C publications, I do keep an eye out for things like this.

test/commons/color/get-text-shadow-colors.js
```diff
@@ -0,0 +1,128 @@
+describe('axe.commons.color.getTextShadowColors', function() {
```
straker, line 1: Need tests for each of the thrown cases

lib/commons/color/get-text-shadow-colors.js
```diff
@@ -0,0 +1,97 @@
...
+
+	while (str) {
+		let colorMatch =
+			str.match(/^rgba?\([0-9,.\s]+\)/i) ||
+			str.match(/^[a-z]+/i) ||
+			str.match(/^#[0-9a-f]+/i);
+		let pixelMatch = str.match(/^([0-9.]+)px/i) || str.match(/^(0)/);
+
+		if (colorMatch) {
+			assert(
```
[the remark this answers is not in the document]
WilcoFiers: I disagree. This isn't like the "element overlaps" or whatever case. If for some reason there are multiple color strings in a text-shadow, whatever environment this is running is has gotten things wrong. That's not a "we don't know, test it yourself" scenario, it's more of a "your test environment is buggy, please fix it" type of situation.

That's why there are no tests for this. It's impossible to get a browser into this erroring state. Axe-core isn't just used in browsers, so we can't really rule out issues like that. This ensures the issue isn't ignored.

lib/commons/color/get-text-shadow-colors.js
```diff
@@ -0,0 +1,97 @@
...
+	shadowColor.parseString(colorStr);
+	shadowColor.alpha *= blurRadiusToAlpha(blurRadius);
+
+	return shadowColor;
+}
+
+function blurRadiusToAlpha(blurRadius) {
+	// This formula is an estimate based on various tests.
+	// Different people test this differently, so opinions may vary.
+	return 3.7 / (blurRadius + 8);
```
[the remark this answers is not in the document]
straker: Why was this resolved? I wouldn't be able to change this formula if we had to adjust it based on user feedback. Where does `3.7` and `blurRadius + 8` come from? I know it was based on testing, if they are truly magic then maybe include the results of those tests so we can better gauge how to adjust this in the future.

test/integration/full/contrast/code-highlighting.js
```diff
@@ -1,5 +1,18 @@
 describe('color-contrast code highlighting test', function() {
 	'use strict';
+	var results;
+
+	before(function(done) {
```
straker, line 5: This will undo the changes I made to fix the flaky test. The before that waits for the window load event needs to happen first before we run axe on the file.

lib/commons/color/get-text-shadow-colors.js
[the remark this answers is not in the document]
WilcoFiers: Afraid I don't have the data anymore, sorry. These estimates are pretty rough.

PR #2337, opened 2020-06-30
feat(standards): add html-dom spec

straker:
This table is new and not part of the previous lookup table. It replaces some hardcoded values in our code and allows us to work on #1597 and #601 by allowing us to identify interactive content.

Replaces these values: 
* `phrasingElements` in [subtree-text](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/text/subtree-text.js#L31-L72)
* `htmlTags` in [is-html-element](https://github.com/dequelabs/axe-core/blob/develop/lib/core/utils/is-html-element.js#L1-L120) (which is why there are empty objects)
* `possibleShadowRoots` in [is-shadow-root.js](https://github.com/dequelabs/axe-core/blob/develop/lib/core/utils/is-shadow-root.js#L1-L20)
* part of the `sectioningElementSelector` in [lookup-table.js](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/aria/lookup-table.js#L2101-L2102)
* all of [native-element-type.js](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/text/native-element-type.js)

There might be a few more but those were the ones I could find in a quick element search.

Just like we plan to do for the `aria-attributes` spec, we should create a commons helper that aggregates information from this table into useful forms, such as getting all elements that are of a certain content type.

## Reviewer checks

### DIFF
--- a/lib/standards/html-dom.js
+++ b/lib/standards/html-dom.js
@@ -1,0 +1,429 @@
+// Source https://html.spec.whatwg.org/multipage/dom.html#content-models
+const htmlDom = {
+	a: {
+		variant: {
+			href: {
+				matches: '[href]',
+				contentTypes: ['interactive', 'phrasing', 'flow']
+			},
+			noHref: {
+				matches: {
+					attributes: {
+						href: null
+					}
+				},
+				contentTypes: ['phrasing', 'flow']
+			}
+		}
+	},
+	abbr: {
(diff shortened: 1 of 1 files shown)


straker:
Merged in with https://github.com/dequelabs/axe-core/pull/2333

Issue #2183, opened 2020-04-28
Make accessible name rules run exclusively on VirtualNode


WilcoFiers:
This is part of our larger project to create a better abstraction layer in axe-core. For the 4.0, we want to have the following rules run exclusively on the virtualNode object (e.g. without anything in matches / evaluate / after, or any of the commons it uses touch the `.actualNode` property, or use the `node` parameter.

## Rules

The following rules will be enabled to run exclusively on virtualNode:

- [x] image-alt
- [x] area-alt
- [x] frame-title
- [x] input-image-alt
- [x] link-name
- [x] object-alt
- [x] button-name
- [x] input-button-name
- [x] role-img-alt
- [x] svg-img-alt
- [x] aria-toggle-field-name
- [x] aria-input-field-name
- [x] label

## Checks

To enable this, the following checks will have to be converted:

- [x] alt-space-value
- [x] aria-label
- [x] aria-labelledby
- [x] focusable-no-name (maybe integrate this into role-none / role-presentation?)
- [x] has-alt
- [x] non-empty-alt
- [x] non-empty-title
- [x] has-visible-text
- [x] role-none
- [x] role-presentation
- [x] button-has-visible-text
- [x] non-empty-if-present
- [x] non-empty-value
- [x] svg-non-empty-title
- [x] no-implicit-explicit-label
- [x] implicit-label
- [x] explicit-label
- [x] help-same-as-label
- [x] hidden-explicit-label

## Match functions

Additionally, a few match functions will have to be converted as well:

- [x] html-namespace-matches.js
- [x] svg-namespace-matches.js
- [x] aria-form-field-name-matches.js
- [x] label-matches.js

PR #2341, opened 2020-07-02
fix(aria-lablledby): work with virtual and serial virtual nodes

straker:
If we have a `SerialVritualNode` we cannot properly run id lookups on it since in linting we only have the one node to look at. In these cases we will throw an error and resolve the error to incomplete in `aria-labelledby`.

Part of issue: #2183 

## Reviewer checks


lib/checks/shared/aria-labelledby.json
```diff
@@ -5,7 +5,8 @@
 		"impact": "serious",
 		"messages": {
 			"pass": "aria-labelledby attribute exists and references elements that are visible to screen readers",
-			"fail": "aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty"
+			"fail": "aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty",
+			"incomplete": "aria-labelledby cannot run on non-DOM nodes"
```
WilcoFiers, line 9:
```suggestion
			"incomplete": "ensure aria-labelledby references an existing element"
```

PR #2343, opened 2020-07-02
fix(forms/*): allow all form control value checks to work with virtual nodes

straker:
Part of issue: #2183 of checks that use accessible-text-virtual code path (for example,`focusable-no-name`)

## Reviewer checks

### DIFF
--- a/lib/commons/forms/is-native-textbox.js
+++ b/lib/commons/forms/is-native-textbox.js
@@ -26,7 +26,8 @@ function isNativeTextbox(node) {
 	const nodeName = node.props.nodeName;
 	return (
 		nodeName === 'textarea' ||
-		(nodeName === 'input' && !nonTextInputTypes.includes(node.attr('type')))
+		(nodeName === 'input' &&
+			!nonTextInputTypes.includes((node.attr('type') || '').toLowerCase()))
 	);
 }
 
--- a/lib/commons/forms/is-native-textbox.js
+++ b/lib/commons/forms/is-native-textbox.js
@@ -1,3 +1,6 @@
+import { getNodeFromTree } from '../../core/utils';
+import AbstractVirtuaNode from '../../core/base/virtual-node/abstract-virtual-node';
+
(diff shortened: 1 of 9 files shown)


lib/commons/forms/is-native-textbox.js
```diff
@@ -15,14 +18,15 @@ const nonTextInputTypes = [
...
  */
 function isNativeTextbox(node) {
-	const nodeName = node.nodeName.toUpperCase();
+	node = node instanceof AbstractVirtuaNode ? node : getNodeFromTree(node);
+	const nodeName = node.props.nodeName;
 	return (
-		nodeName === 'TEXTAREA' ||
-		(nodeName === 'INPUT' && !nonTextInputTypes.includes(node.type))
+		nodeName === 'textarea' ||
+		(nodeName === 'input' && !nonTextInputTypes.includes(node.attr('type')))
```
WilcoFiers, line 29: you'll need to lowercase the type.

lib/commons/forms/is-native-textbox.js
```diff
@@ -15,14 +18,16 @@ const nonTextInputTypes = [
...
 function isNativeTextbox(node) {
-	const nodeName = node.nodeName.toUpperCase();
+	node = node instanceof AbstractVirtuaNode ? node : getNodeFromTree(node);
+	const nodeName = node.props.nodeName;
 	return (
-		nodeName === 'TEXTAREA' ||
-		(nodeName === 'INPUT' && !nonTextInputTypes.includes(node.type))
+		nodeName === 'textarea' ||
+		(nodeName === 'input' &&
+			!nonTextInputTypes.includes((node.attr('type') || '').toLowerCase()))
```
WilcoFiers, line 30: May not be necessary for this PR, but I think we should consider adding `virtualNode.props.type` by default, in the same way that we set nodeType. That way we don't have to do this type casting, and it'd give us an easier way to deal with inputs defaulting to type=text.

PR #2345, opened 2020-07-06
tests: ensure checks can be run using only serial virtual nodes

straker:
Noticed that we don't have any tests that ensure we can run a rule/check/matches using `SerialVritualNode`. This caught the error in `aria-lablledby-evaluate` where we passed `node` into `arialabelledbyText` instead of passing `virtualNode`. 

## Reviewer checks

### DIFF
--- a/test/integration/virtual-rules/index.js
+++ b/test/integration/virtual-rules/index.js
@@ -1,0 +1,315 @@
+describe('virtual rules', function() {
+	describe('checks', function() {
+		describe('autocomplete-valid', function() {
+			it('should pass', function() {
+				var results = axe.runVirtualRule(
+					'autocomplete-valid',
+					new axe.SerialVirtualNode({
+						nodeName: 'input',
+						attributes: {
+							type: 'text',
+							autocomplete: 'country-name'
+						}
+					})
+				);
+
+				assert.lengthOf(results.passes, 1);
+				assert.lengthOf(results.violations, 0);
+				assert.lengthOf(results.incomplete, 0);
+			});
+
+			it('should fail', function() {
+				var results = axe.runVirtualRule(
+					'autocomplete-valid',
+					new axe.SerialVirtualNode({
+						nodeName: 'input',
+						attributes: {
+							type: 'text',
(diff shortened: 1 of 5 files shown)


test/integration/virtual-rules/index.js
```diff
@@ -0,0 +1,315 @@
+describe('virtual rules', function() {
+	describe('checks', function() {
+		describe('autocomplete-valid', function() {
+			it('should pass', function() {
+				var results = axe.runVirtualRule(
+					'autocomplete-valid',
+					new axe.SerialVirtualNode({
```
WilcoFiers, line 7: We don't need to create a virtual node, this is done by runVirtualRule.

```suggestion
					{
```

test/integration/virtual-rules/index.js
```diff
@@ -0,0 +1,315 @@
...
+						nodeName: 'input',
+						attributes: {
+							type: 'text',
+							autocomplete: 'country-name'
+						}
+					})
```
WilcoFiers, line 13:
```suggestion
					}
```

test/integration/virtual-rules/index.js
```diff
@@ -0,0 +1,315 @@
+describe('virtual rules', function() {
```
WilcoFiers, line 1: I think this file should be split up, give each rule its own .js file. I'm afraid this is going to get messy quickly if we don't.

test/integration/virtual-rules/index.js
```diff
@@ -0,0 +1,315 @@
...
+	describe('checks', function() {
```
WilcoFiers, line 2: I don't think we should split this up into checks. For integrations, the checks don't matter. Just have rules with different passes and fails.

test/integration/virtual-rules/object-alt.js
```diff
@@ -0,0 +1,103 @@
+describe('object-alt', function() {
```
WilcoFiers, line 1: I don't think we should include this until it can actually produce violations.

PR #2350, opened 2020-07-08
fix(virtual-node): default and lowercase type property

straker:
Allows us to do input property matches with case insensitivity for serial and virtual nodes.

```js
const input = document.createElement('input');
input.setAttribute('type', 'BUTTON');
const vNode = axe.utils.getFlattenedTree(fixture)[0];

commons.matches(vNode, {
  properties: {
    type: 'button'
  }
}); // true

const serialNode = new axe.SerialVritualNode({
  nodeName: 'input',
  attributes: {
    type: 'BUTTON'
  }
});

commons.matches(serialNode, {
  properties: {
    type: 'button'
  }
}); // true
```

## Reviewer checks

### DIFF
--- a/lib/core/base/virtual-node/serial-virtual-node.js
+++ b/lib/core/base/virtual-node/serial-virtual-node.js
@@ -52,11 +52,25 @@ function normaliseProps(serialNode) {
 		`nodeName has to be a string, got '${nodeName}'`
 	);
 
+	nodeName = nodeName.toLowerCase();
+	let type = null;
+	if (nodeName === 'input') {
+		type = (
+			serialNode.type ||
+			(serialNode.attributes && serialNode.attributes.type) ||
+			'text'
+		).toLowerCase();
+	}
+
 	const props = {
 		...serialNode,
 		nodeType,
-		nodeName: nodeName.toLowerCase()
+		nodeName
 	};
+	if (type) {
+		props.type = type;
+	}
+
 	delete props.attributes;
 	return Object.freeze(props);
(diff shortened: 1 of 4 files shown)


lib/core/base/virtual-node/serial-virtual-node.js
```diff
@@ -52,11 +52,25 @@ function normaliseProps(serialNode) {
 		`nodeName has to be a string, got '${nodeName}'`
 	);
 
+	nodeName = nodeName.toLowerCase();
+	let type = null;
+	if (nodeName === 'input') {
+		type = (
+			serialNode.type ||
+			(serialNode.attributes && serialNode.attributes.type) ||
```
WilcoFiers, line 60: I think we should check that type actually exists. `type="foo"` should be `props.type = 'text'`.

lib/core/base/virtual-node/virtual-node.js
```diff
@@ -26,6 +26,18 @@ class VirtualNode extends AbstractVirtualNode {
...
 
+		this._nodeName = this._isXHTML
+			? node.nodeName
+			: node.nodeName.toLowerCase();
+
+		// we will normalize the type prop for inputs by looking strictly
+		// at the attribute and not what the browser resolves the type
+		// to be
+		if (node.nodeName.toLowerCase() === 'input') {
+			const type = node.getAttribute('type') || 'text';
```
WilcoFiers, line 37: As above, lets make sure type exists.

PR #2353, opened 2020-07-08
fix(aria-toggle-field-name): work with virtual nodes

straker:
Part of #2183

## Reviewer checks


lib/commons/text/label-text.js
```diff
@@ -50,6 +49,12 @@ function labelText(virtualNode, context = {}) {
  * @return {HTMLElement} The label element, or null if none is found
  */
 function getExplicitLabels({ actualNode }) {
+	if (!actualNode) {
```
WilcoFiers, line 52: If there's no ID, we don't need to throw.

PR #2354, opened 2020-07-08
fix(label): work with virtual nodes

straker:
This one was a lot of changes to be able to get the rule to work with serial virtual nodes.

Part of #2183 

## Reviewer checks


lib/checks/label/implicit-evaluate.js
```diff
@@ -1,12 +1,16 @@
...
-import { accessibleText } from '../../commons/text';
+import { closest } from '../../core/utils';
+import { accessibleTextVirtual } from '../../commons/text';
 
 function implicitEvaluate(node, options, virtualNode) {
-	const label = findUpVirtual(virtualNode, 'label');
-	if (label) {
-		return !!accessibleText(label, { inControlContext: true });
+	try {
+		const label = closest(virtualNode, 'label');
```
WilcoFiers, line 6: This is going to create false positives. If we don't have access to the parent of `input`, we can't just decide it has no implicit label as an ancestor. I think "closest" may need to throw if it stops before it reaches a root node.

lib/commons/text/label-virtual.js
```diff
@@ -12,28 +12,37 @@ import findUpVirtual from '../dom/find-up-virtual';
...
 		if (candidate) {
 			return candidate;
 		}
 	}
 
-	ref = findUpVirtual(node, 'label');
-	candidate = ref && visible(ref, true);
+	if (!virtualNode.actualNode) {
+		throw new TypeError('Cannot determine if non-DOM node is visible');
+	}
```
WilcoFiers, line 42: I'm not sure a generic "always throw" is right. Instead lets have "closest" and "visibleVirtual" decide if throwing is necessary.

lib/commons/text/label-virtual.js
```diff
@@ -12,28 +12,37 @@ import findUpVirtual from '../dom/find-up-virtual';
...
+
+	ref = closest(virtualNode, 'label');
+	candidate = ref && visible(ref.actualNode, true);
```
WilcoFiers, line 45:
```suggestion
	candidate = ref && visibleVirtual(ref, true);
```

straker:
Turns out `help-same-as-label` has been disabled for at least 2 years so no need to convert it over. Not sure why we still have it as part of the check at this point.

PR #2358, opened 2020-07-09
feat(commons/standards): create the commons/standards object for helper functions against the standards table

straker:
This will allow us to start using the standards table in our code base.

## Reviewer checks

### DIFF
--- a/lib/commons/standards/get-html-element-spec.js
+++ b/lib/commons/standards/get-html-element-spec.js
@@ -6,7 +6,7 @@ import matchesFn from '../../commons/matches';
  * @param {VirtualNode} vNode The VirtualNode to get the spec for.
  * @return {Object} The html-elms standard spec
  */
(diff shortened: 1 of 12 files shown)


lib/commons/standards/get-html-element-spec.js
```diff
@@ -0,0 +1,49 @@
+import standards from '../../standards';
+import matchesFn from '../../commons/matches';
+
+/**
+ * Return the spec for an HTML element from the html-elms standard. Since the spec is determined by the node and what attributes it has, a node is required.
+ * @param {VirtualNode} vNode The VirtualNode to get the spec for.
+ * @return {Object} The html-elms standard spec
+ */
+function getHtmlElementSpec(vNode) {
```
WilcoFiers, line 9: I suspect we're going to need this for SVG too, lets not limit ourselves in the name

```suggestion
function getElementSpec(vNode) {
```

PR #2361, opened 2020-07-10
fix(aria/get-role-type): work with standards object

straker:
Refactored `get-role-type` to have a better happy path and return early rather than doing the `&&` and `||` checks.

Part of #2108

## Reviewer checks

### DIFF
--- a/lib/commons/aria/get-role-type.js
+++ b/lib/commons/aria/get-role-type.js
@@ -1,4 +1,4 @@
-import lookupTable from './lookup-table';
+import standards from '../../standards';
 
 /**
  * Get the "type" of role; either widget, composite, abstract, landmark or `null`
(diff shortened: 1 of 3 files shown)


lib/commons/aria/get-role-type.js
```diff
@@ -9,8 +9,13 @@ import lookupTable from './lookup-table';
...
 function getRoleType(role) {
-	var r = lookupTable.role[role];
-	return (r && r.type) || null;
+	const roleDef = standards.ariaRoles[role];
+
+	if (!roleDef) {
+		return null;
+	}
+
+	return roleDef.type;
```
stephenmathieson, line 18: Previously, this would return `null` if there wasn't a truthy `.type`. Do we need to maintain consistently here, or is the possibility of returning `undefined` here ok?

straker: Every role should have a type so it shouldn't be possible to return `undefined` or `null`. A role without a type should be seen as a bug.

PR #2370, opened 2020-07-13
chore(validate-attr-value): work with standards object

straker:
Added true/false support and modified the integration test as discussed in https://github.com/dequelabs/axe-core/pull/2315.

Part of #2108 

## Reviewer checks


lib/commons/aria/validate-attr-value.js
```diff
@@ -27,6 +27,8 @@ function validateAttrValue(node, attr) {
 
 	switch (attrInfo.type) {
 		case 'boolean':
+			return ['true', 'false'].includes(value.toLowerCase());
```
WilcoFiers, line 30: Didn't realise, good call though. Nice cleanup.

test/commons/aria/validate-attr-value.js
```diff
@@ -65,36 +73,12 @@ describe('aria.validateAttrValue', function() {
 
 		node.setAttribute('pigs', '');
 		assert.isTrue(axe.commons.aria.validateAttrValue(node, 'pigs'));
+
+		node.setAttribute('horses', '');
+		assert.isTrue(axe.commons.aria.validateAttrValue(node, 'pigs'));
```
WilcoFiers, line 78: Should this not be:

```suggestion
		assert.isTrue(axe.commons.aria.validateAttrValue(node, 'horses'));
```

test/commons/aria/validate-attr-value.js
```diff
@@ -1,52 +1,60 @@
...
 	'use strict';
 
-	var orig = axe.commons.aria.lookupTable.attributes,
-		fixture = document.getElementById('fixture');
-
+	var fixture = document.getElementById('fixture');
 	var shadowSupport = axe.testUtils.shadowSupport;
 
+	beforeEach(function() {
+		axe.configure({
```
WilcoFiers, line 8: Instead of setting this each time, I think we should put these back into the tests. A test should be self-descriptive. I get that's a bit repetitive but that's a good thing. It makes it easy to see if the test is correct.

test/commons/aria/validate-attr-value.js
```diff
@@ -113,16 +97,52 @@ describe('aria.validateAttrValue', function() {
...
 		node.setAttribute('sheep', '  \r\n\t  ');
 		assert.isTrue(axe.commons.aria.validateAttrValue(node, 'pigs'));
+
+		node.setAttribute('horses', '  \r\n\t  ');
+		assert.isTrue(axe.commons.aria.validateAttrValue(node, 'pigs'));
 	});
+
 	describe('schema defintions', function() {
+		beforeEach(function() {
+			axe.configure({
```
WilcoFiers, line 107: As above, I don't think this should be part of setup.

test/commons/aria/validate-attr-value.js
```diff
@@ -191,219 +186,275 @@ describe('aria.validateAttrValue', function() {
...
 
 			it('should return false when a single referenced node is not found', function() {
-				node.setAttribute('cats', 'invalid');
+				node.setAttribute('goats', 'invalid');
 				// target2 not found
-				assert.isFalse(axe.commons.aria.validateAttrValue(node, 'cats'));
+				assert.isFalse(axe.commons.aria.validateAttrValue(node, 'goats'));
 			});
 
 			it('should return false when at no referenced element is found', function() {
```
WilcoFiers, line 218:
```suggestion
			it('should return false when no referenced element is found', function() {
```

PR #2377, opened 2020-07-14
feat(required-attrs): deprecate options to pass more required attrs

straker:
Slider, spinbutton, and scrollbar all have implicit values so we didn't make them required anymore.

Part of issue #2108 

## Reviewer checks

### DIFF
--- a/lib/checks/aria/aria-required-attr-evaluate.js
+++ b/lib/checks/aria/aria-required-attr-evaluate.js
@@ -1,48 +1,18 @@
-import {
-	isNativeTextbox,
-	isNativeSelect,
-	isAriaTextbox,
-	isAriaListbox,
-	isAriaCombobox,
-	isAriaRange
-} from '../../commons/forms';
 import { requiredAttr } from '../../commons/aria';
+import { getElementSpec } from '../../commons/standards';
 import { uniqueArray, getNodeFromTree } from '../../core/utils';
-import matches from '../../commons/matches';
 
 function ariaRequiredAttrEvaluate(node, options = {}) {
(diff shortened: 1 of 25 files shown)


lib/checks/aria/aria-required-attr-evaluate.js
```diff
@@ -1,48 +1,18 @@
...
-			});
-		}
-	};
-
 	if (node.hasAttributes()) {
 		const role = node.getAttribute('role');
 		let required = requiredAttr(role);
+		const elmSpec = getElementSpec(vNode);
 
+		// @deprecated: required attr options to pass more attrs.
```
WilcoFiers, line 14: Please include the deprecation notice in your PR body "DEPRECATED: ..."

test/integration/rules/aria-required-attr/required-attr.html
```diff
@@ -34,9 +34,10 @@
 <span role="radio" id="pass8" aria-checked="true">I am GREEN!</span>
 <input type="number" role="spinbutton" id="pass9" value="10" />
 <input type="number" role="spinbutton" id="pass10" />
+<div role="scrollbar" id="pass11">ok</div>
+<div role="slider" id="pass12">ok</div>
+<div role="spinbutton" id="pass13">ok</div>
```
WilcoFiers, line 39: This is because of aria-valuenow, correct? I'd prefer to leave this as is. ARIA 1.2 has taken the default valuenow off scrollbar and slider because it makes very little sense, and wasn't well supported. I suspect the same is true for spinbutton, but we can test that if you want.

PR #2379, opened 2020-07-14
chore(aria-required-parent): work with standards object

straker:
Implicit nodes is being deprecated in favor of just getting the role of the node in question. 

Part of issue #2108 

## Reviewer checks

### DIFF
--- a/lib/commons/aria/required-context.js
+++ b/lib/commons/aria/required-context.js
@@ -15,7 +15,7 @@ function requiredContext(role) {
 		return null;
 	}
 
-	return roleDef.requiredContext;
+	return [...roleDef.requiredContext];
 }
 
 export default requiredContext;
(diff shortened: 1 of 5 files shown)


lib/commons/aria/required-context.js
```diff
@@ -9,15 +9,13 @@ import lookupTable from './lookup-table';
...
 
-	if (roles) {
-		// TODO: es-module-utils.clone
-		context = axe.utils.clone(roles.context);
+	if (!roleDef || !Array.isArray(roleDef.requiredContext)) {
+		return null;
 	}
-	return context;
+
+	return roleDef.requiredContext;
```
WilcoFiers, line 18: We should return a copy here. The return value isn't immutable, we can't assume some rule isn't going to accidentally modify this object.

PR #2380, opened 2020-07-14
chore(region): work with standards object

straker:
`title` no longer provides a `<section>` element an accessible name as we went with the [Safari implementation](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/aria/lookup-table.js#L2117-L2125).

Part of issue #2108 

## Reviewer checks

### DIFF
--- a/test/integration/full/region/region-fail.html
+++ b/test/integration/full/region/region-fail.html
@@ -35,6 +35,9 @@
 			<section>
 				<p>Content</p>
 			</section>
+			<section title="section 3">
+				<p>Content</p>
+			</section>
 		</div>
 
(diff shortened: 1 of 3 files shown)


test/integration/full/region/region-fail.html
```diff
@@ -35,6 +35,9 @@ <h1 id="mainheader" tabindex="0">This is a header.</h1>
 			<section>
 				<p>Content</p>
 			</section>
+			<section title="section 3">
+				<p>Content</p>
+			</section>
```
WilcoFiers, line 40: Why should that fail?

Issue #2347, opened 2020-07-06
Bug: running JSDOM (without setting globals) results in error


straker:
```js
let dom = jsom('<div></div>');
axe.run(dom.window.document.documentElement);  // runs fine

dom = jsom('<div></div>');
axe.run(dom.window.document.documentElement);  // errors with "axe.run arguments are invlaid"
```

The issue is that axe-core now sets the global `document` and `window` but does not unset it after the run. JSDom apparently creates a new `window` and/or `document` every new instance, so they are no longer the same. When we check if the `context` is a context, we use the `window` element to determine things, which is what fails on the 2nd run.

straker:
Pairing with @stephenmathieson  we discovered the problem only happens in jsdom v16 (v15 does not fail). The fix is to null `window` and `document` in the `cleanup` function of `run-rules.js` if we set the globals ourselves.

stephenmathieson:
<details>
<summary>Failing test case</summary>

```js
const axe = require('axe-core'); // ^3.5.5-canary.67d2dca
const jsdom = require('jsdom'); // ^16

const test = async () => {
  const one = new jsdom.JSDOM(`
    <!DOCTYPE html>
    <html lang="en">
      <head>
        <meta charset="UTF-8">
        <meta name="viewport" content="width=device-width, initial-scale=1.0">
        <title>Document</title>
      </head>
      <body>
        Hello
      </body>
    </html>
  `);

  let r = await axe.run(one.window.document.documentElement, {
    rules: { 'color-contrast': { enabled: false } },
  });
  console.log('one: %d violations', r.violations.length);

  const two = new jsdom.JSDOM(`
    <!DOCTYPE html>
    <html lang="en">
      <head>
        <meta charset="UTF-8">
        <meta name="viewport" content="width=device-width, initial-scale=1.0">
        <title>Document</title>
      </head>
      <body>
        Hello
      </body>
    </html>
  `);

  r = await axe.run(two.window.document.documentElement, {
    rules: { 'color-contrast': { enabled: false } },
  });
  console.log('two: %d violations', r.violations.length);
};

test();
```
</details>

padmavemulapati:
Installed "JSDOM" and ran 
1. without defining jsdom - `dom = jsom('<div></div>');
axe.run(dom.window.document.documentElement);` - got error - "Uncaught ReferenceError: jsom is not defined
    at playground.html:10"
and 
[...]

PR #2387, opened 2020-07-15
fix(run): cleanup globals if set from context

straker:
As discussed in the issue, just needed to clean up the set globals as jsdom creates new globals when called again.

Closes issue: #2347 

## Reviewer checks


stephenmathieson:
> Please include a test to prevent regression.

I came here just to say that ☝️ 

We had discussed updating the [`jsdom` example](https://github.com/dequelabs/axe-core/tree/develop/doc/examples/jsdom) in order to accomplish this, as [the examples are run by CI](https://github.com/dequelabs/axe-core/blob/develop/.circleci/config.yml#L103-L111).

straker:
I tried updating the example, but I would have to do some convoluted stuff to make it break. For example, just trying to recreate the DOM in the `it` gives me an error `ReferenceError: Node is not defined`. I could probably get it to break the with example provided in this issue, but it doesn't fit very well in a public facing example of how to use JSOM

stephenmathieson:
> I tried updating the example, but I would have to do some convoluted stuff to make it break.

It's not that bad. Here's what I did:

```
npm run build
cd doc/examples/jsdom
# Use `"axe-core": "file://../../../"` in package.json, then
npm install jsdom@latest --save-dev
npm install
npm test
```


<details>
<summary>Updated JSDOM example</summary>

```js
/* eslint-env mocha */
const axe = require('axe-core');
const { JSDOM } = require('jsdom');
const assert = require('assert');

const config = {
  rules: {
    // `color-contrast` does not work with JSDOM.
    'color-contrast': { enabled: false },
    // In this example, we are not concerned with the following rules:
    'landmark-one-main': { enabled: false },
    'page-has-heading-one': { enabled: false },
    region: { enabled: false }
  }
};

const createDOM = (working = true) => {
  const { window } = new JSDOM(`<!DOCTYPE html>
  <html lang="en">
    <head>
      <title>JSDOM Example</title>
    </head>
    <body>
    ${
      working
        ? `<div id="working">
            <label for="has-label">Label for this text field.</label>
            <input type="text" id="has-label">
          </div>`
        : `<div id="broken">
            <p>Not a label</p><input type="text" id="no-label">
          </div>`
    }
    </body>
  </html>`);
  return window;
};

describe('axe', () => {
  it('should report that good HTML is good', async () => {
    const window = createDOM(true);
    const result = await axe.run(window.document.documentElement, config);
    assert.equal(result.violations.length, 0, 'Violations is not empty');
  });

  it('should report that bad HTML is bad', async () => {
    const window = createDOM(false);
    const result = await axe.run(window.document.documentElement, config);
    assert.equal(result.violations.length, 1, 'Violations.length is not 1');
  });
});
```
</summary>

straker:
Except we determined in our last discussion of this file being both an example and a test file, that it should remain an example. So I couldn't change the main package.json file to use the axe file from the repo, so had to edit https://github.com/dequelabs/axe-core/blob/develop/doc/examples/test-examples.js#L14-L20 as a workaround so we could leave it as an example and still test the examples using the latest axe-core.

stephenmathieson:
If you don't want to use the example as the test case, I'm fine with that. However, I [agree with Wilco that this needs to be tested](https://github.com/dequelabs/axe-core/pull/2387#pullrequestreview-449694974).

Would you rather create a separate JSDOM test elsewhere?

straker:
I guess we'll have to so this problem of example/test doesn't prevent us from testing fixes we need to

test/node/jsdom.js
```diff
@@ -0,0 +1,54 @@
+var axe = require('../../');
+var jsdom = require('jsdom');
```
stephenmathieson, line 2: Since we can use "modern JS" here, I really think we should. Thoughts on refactoring this to make it look like "today's JavaScript"?

test/node/jsdom.js
```diff
@@ -0,0 +1,54 @@
...
+describe('jsdom axe-core', function() {
+	it('should run without setting globals', function(done) {
+		var dom = new jsdom.JSDOM(domStr);
+
+		axe
+			.run(dom.window.document.documentElement, {
+				rules: { 'color-contrast': { enabled: false } }
+			})
+			.then(function(results) {
+				assert.equal(results.violations.length, 2);
```
stephenmathieson, line 27: Why check that there are 2 violations? This number is subject to change any time we add/remove rules.

test/node/jsdom.js
```diff
@@ -0,0 +1,54 @@
...
+
+				axe
+					.run(dom.window.document.documentElement, {
+						rules: { 'color-contrast': { enabled: false } }
+					})
+					.then(function(results) {
+						assert.equal(results.violations.length, 2);
+						done();
+					});
+			});
```
stephenmathieson, line 52: Without a `.catch`, this could cause an unhandled rejection and crash the process.

test/node/jsdom.js
straker: I'd love to but we can't. The entire `test` dir is dictated by the [eslint file](https://github.com/dequelabs/axe-core/blob/develop/test/.eslintrc) in there that disallows any modern javascript... It's a bit of a pain.

stephenmathieson: That doesn't sound like a "we can't", it just sounds like some ESLint settings need to be modified. We could add `test/node/.eslintrc` and override the rules added there, or move that file into the directory(ies) that those rules are relevant for.

I don't consider this blocking tho. If it's too much work to mess with the ESLint stuff, I understand. We can always improve on this later.

test/node/jsdom.js
WilcoFiers: Yeah, doesn't matter, just check there is a result.

test/node/jsdom.js
```diff
@@ -0,0 +1,54 @@
...
+		axe
+			.run(dom.window.document.documentElement, {
+				rules: { 'color-contrast': { enabled: false } }
+			})
+			.then(function(results) {
+				assert.equal(results.violations.length, 2);
+
+				var dom = new jsdom.JSDOM(domStr);
+
+				axe
```
WilcoFiers, line 44: If you return the promise, you can chain it, instead of nesting another promise.

stephenmathieson: Also, if we chain and return the promise, we don't have to add a `.catch()` or mess with `done`. Mocha has had support for promises for awhile now.

test/node/jsdom.js
```diff
@@ -0,0 +1,54 @@
...
+
+		axe
+			.run(dom.window.document.documentElement, {
+				rules: { 'color-contrast': { enabled: false } }
+			})
+			.then(function(results) {
+				assert.equal(results.violations.length, 2);
+				done();
+			});
+	});
```
stephenmathieson, line 30:
```suggestion
	it('should run without setting globals', function() {
		var dom = new jsdom.JSDOM(domStr);

		return axe
			.run(dom.window.document.documentElement, {
				rules: { 'color-contrast': { enabled: false } }
			})
			.then(function(results) {
				assert.notEqual(results.violations.length, 0);
			});
	});
```

test/node/jsdom.js
```diff
@@ -0,0 +1,54 @@
...
+				axe
+					.run(dom.window.document.documentElement, {
+						rules: { 'color-contrast': { enabled: false } }
+					})
+					.then(function(results) {
+						assert.equal(results.violations.length, 2);
+						done();
+					});
+			});
+	});
```
stephenmathieson, line 53:
```suggestion
	it('should unset globals so it can run with a new set of globals', function() {
		var dom = new jsdom.JSDOM(domStr);

		return axe
			.run(dom.window.document.documentElement, {
				rules: { 'color-contrast': { enabled: false } }
			})
			.then(function(results) {
				assert.notEqual(results.violations.length, 0);

				var dom = new jsdom.JSDOM(domStr);

				return axe
					.run(dom.window.document.documentElement, {
						rules: { 'color-contrast': { enabled: false } }
					})
					.then(function(results) {
						assert.notEqual(results.violations.length, 0);
					});
			});
	});
```

Issue #2209, opened 2020-05-04
Add "ancestry" property to nodes


WilcoFiers:
Currently, axe-core has the "selector" property, which gives a CSS selector that is designed to be easy for developers to read, and understand the element from. This selector is helpful, but costly to compute and somewhat prone to change.

A selector that is more stable across axe-core versions could be an "ancestry" selector, which just lists all the node names down to the root, including only which child was selected. The ancestry selector would look something like this:

```
['html > body > iframe', 'html > body > main > div:nth-child(2) > p > img']
```

The "ancestry" property would be in addition to the existing "selector" property.

WilcoFiers:
We should check for performance, if this adds a significant amount of time, make this an option that's off by default.

padmavemulapati:
devTask

jeankaplansky:
Added to internal release notes 7/27/2020

PR #2389, opened 2020-07-16
feat(options): add ancestry CSS selector to nodes

WilcoFiers:
Adds an option for axe.run() to return an "ancestry" css selector. Similar to "selector" but instead including just the node names and child index up to the root node of the tree. Handles iframe and shadow DOM nesting in the same way .selector does.

Closes issue: #2209

## Reviewer checks


lib/core/utils/get-ancestry.js
```diff
@@ -0,0 +1,31 @@
+import getShadowSelector from './get-shadow-selector';
+
+function generateAncestry(node) {
```
straker, line 3: I think we should cache these results on the virtual node as we process them so we aren't recalculating the entire ancestry tree of every descendant of a node.

lib/core/utils/get-ancestry.js
```diff
@@ -0,0 +1,31 @@
...
+	const parent = node.parentElement;
+	if (!parent) {
+		return nodeName;
+	}
+
+	let nthChild = '';
+	if (
+		nodeName !== 'head' &&
+		nodeName !== 'body' &&
+		parent.children.length > 1
```
straker, line 14: Would it be better if this was only checking if the parent had multiple of the nodeName? I know `getSelector` only adds nth-child stuff if it needs to.

Right now if the parent has two children but they are both different nodeNames, you still get the nth-child selector which doesn't do much but add noise:

```
<html>
  <body>
    <div>
      <a href="/foo">Foo</a>
      <span>Hello</span>
    </div> 
  </body>
</html>
```

Results in `html > body > div:nth-child(1) > a:nth-child(1)`

test/core/reporters/helpers/process-aggregate.js
```diff
@@ -69,6 +73,7 @@ describe('helpers.processAggregate', function() {
 							selector: '#dopel',
 							source: '<input id="dopel"/>',
 							xpath: '/main/input[@id="dopel"]',
+							ancestry: 'html > body > main > input:nth-child(0)',
```
straker, line 76: how do we get an nth-child of 0?

lib/core/utils/get-ancestry.js
WilcoFiers: What would the benefit of that be? This seems faster.

PR #2391, opened 2020-07-16
feat(object-alt,accessible-text): object-alt rule and accessible text to work with serial virtual nodes with children

straker:
There is a lot going on here but it was all necessary to get a serial virtual node with children to run through the accessbile-text functions. This allows `object-alt` and other rules that look at the subtree text to pass/fail with a full descendant tree (before it would just incomplete).

```js
var node = new axe.SerialVirtualNode({
  nodeName: 'object'
});
var child = new axe.SerialVirtualNode({
  nodeName: '#text',
  nodeType: 3,
  textContent: 'foobar'
});
node.children.push(child);

var results = axe.runVirtualRule('object-alt', node);  // passes
```

We decided that if a virtual node does not have an `actualNode` property, we will always consider it visible so we don't have to enter the `isVisible` function which checks visibility with CSS. 

We also needed serial virtual nodes to have text children, so they needed to accept `nodeType` of 3.

Note: I wanted to use the html-elm spec to know if the element was as phrasing element rather than keep that hardcoded list. Unfortunately our hardcoded list is not a complete list and so something like and `input`, which is a [phrasing element](https://html.spec.whatwg.org/multipage/dom.html#phrasing-content), [would no longer add space](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/text/subtree-text.js#L81-L90) between the label text and the input value such as this acceptance test:

```html
<label for="test">
  foo
  <input type="text" value="David"/>
</label>
```

Which results in `fooDavid` rather than `foo David`. So I had to leave it in, but had to lowercase the names, and also decided to sort the list while I was there (much easier to compare against the list in the spec).

This also lead to discovering some of the html-elm spec was incorrect, like `kbd` being misspelled and `p` and `summary` being phrasing elements when they should not have been. Also made `namingMethods` an array because it made more sense than being a string or array.

Part of issue #2183

## Reviewer checks


lib/commons/text/accessible-text-virtual.js
```diff
@@ -60,11 +60,11 @@ function accessibleTextVirtual(virtualNode, context = {}) {
...
  * @return {String} textContent value
  */
-function textNodeContent({ actualNode }) {
-	if (actualNode.nodeType !== 3) {
+function textNodeContent(virtualNode) {
+	if (virtualNode.props.nodeType !== 3) {
 		return '';
 	}
-	return actualNode.textContent;
+	return virtualNode.props.textContent;
```
WilcoFiers, line 67: I think nodeValue is more appropriate here. In this particular case it does the same thing, but what we're getting at here is the value of the text node, not the text content of the node and all its descendants, which is what textContent does.

```suggestion
	return virtualNode.props.nodeValue;
```

lib/commons/text/accessible-text-virtual.js
```diff
@@ -60,11 +60,11 @@ function accessibleTextVirtual(virtualNode, context = {}) {
  * @param {VirtualNode} element
  * @return {String} textContent value
```
WilcoFiers, line 61:
```suggestion
 * @return {String} nodeValue
```

lib/core/base/virtual-node/virtual-node.js
```diff
@@ -49,14 +49,15 @@ class VirtualNode extends AbstractVirtualNode {
 	// abstract Node properties so we can run axe in DOM-less environments.
 	// add to the prototype so memory is shared across all virtual nodes
 	get props() {
-		const { nodeType, nodeName, id, multiple } = this.actualNode;
+		const { nodeType, nodeName, id, multiple, textContent } = this.actualNode;
```
WilcoFiers, line 52: I suspect this is going to be faster, as it won't require the browser to compute the text content of every node in the page.

```suggestion
		const { nodeType, nodeName, id, multiple, nodeValue } = this.actualNode;
```

lib/core/base/virtual-node/virtual-node.js
```diff
@@ -49,14 +49,15 @@ class VirtualNode extends AbstractVirtualNode {
...
 
 		return {
 			nodeType,
 			nodeName: this._isXHTML ? nodeName : nodeName.toLowerCase(),
 			id,
 			type: this._type,
-			multiple
+			multiple,
+			textContent
```
WilcoFiers, line 60:
```suggestion
			nodeValue
```

lib/standards/html-elms.js
```diff
@@ -504,15 +504,15 @@ const htmlElms = {
...
+				namingMethods: ['labelText']
 			}
 		}
 	},
 	ins: {
 		contentTypes: ['phrasing', 'flow'],
 		allowedRoles: true
 	},
-	kdb: {
+	kbd: {
```
WilcoFiers, line 515: Oof! Good catch.

test/integration/virtual-rules/object-alt.js
```diff
@@ -0,0 +1,152 @@
...
+	});
+
+	it('should pass for visible text content', function() {
+		var node = new axe.SerialVirtualNode({
+			nodeName: 'object'
+		});
+		var child = new axe.SerialVirtualNode({
+			nodeName: '#text',
+			nodeType: 3,
+			textContent: 'foobar'
```
WilcoFiers, line 74:
```suggestion
			nodeValue: 'foobar'
```

test/integration/virtual-rules/object-alt.js
```diff
@@ -0,0 +1,152 @@
...
+	});
+
+	it('should fail children contain no visible text', function() {
+		var node = new axe.SerialVirtualNode({
+			nodeName: 'object'
+		});
+		var child = new axe.SerialVirtualNode({
+			nodeName: '#text',
+			nodeType: 3,
+			textContent: ''
```
WilcoFiers, line 142:
```suggestion
			nodeValue: ''
```

test/core/base/virtual-node/virtual-node.js
```diff
@@ -37,6 +37,7 @@ describe('VirtualNode', function() {
 			assert.equal(vNode.props.nodeName, 'input');
 			assert.equal(vNode.props.id, 'monkeys');
 			assert.equal(vNode.props.type, 'text');
+			assert.equal(vNode.props.textContent, '');
```
WilcoFiers, line 40:
```suggestion
			assert.equal(vNode.props.nodeValue, '');
```

test/checks/shared/has-visible-text.js
```diff
@@ -52,12 +52,28 @@ describe('has-visible-text', function() {
...
+		});
+
+		it('should return true if there is visible text', function() {
+			var node = new axe.SerialVirtualNode({
+				nodeName: 'object'
+			});
+			var child = new axe.SerialVirtualNode({
+				nodeName: '#text',
+				nodeType: 3,
+				textContent: 'hello!'
```
WilcoFiers, line 72:
```suggestion
				nodeValue: 'hello!'
```

test/checks/shared/has-visible-text.js
```diff
@@ -52,12 +52,28 @@ describe('has-visible-text', function() {
 			);
 		});
 
-		it('should return undefined if element is named from contents', function() {
+		it('should return false if there is no visible text', function() {
```
WilcoFiers, line 55: As above, this is a false positive. I'm OK with a tech debt ticket if this is too difficult to solve in one PR.

test/checks/shared/button-has-visible-text.js
```diff
@@ -54,13 +54,37 @@ describe('button-has-visible-text', function() {
...
+		});
+
+		it('should return true if a button element has text', function() {
+			var node = new axe.SerialVirtualNode({
+				nodeName: 'button'
+			});
+			var child = new axe.SerialVirtualNode({
+				nodeName: '#text',
+				nodeType: 3,
+				textContent: 'Text'
```
WilcoFiers, line 78:
```suggestion
				nodeValue: 'Text'
```

lib/standards/html-elms.js
```diff
@@ -811,10 +811,9 @@ const htmlElms = {
 		allowedRoles: true
 	},
 	summary: {
-		contentTypes: ['phrasing', 'flow'],
```
WilcoFiers, line 814: Why this change?

lib/standards/html-elms.js
```diff
@@ -647,10 +647,10 @@ const htmlElms = {
 		contentTypes: ['phrasing', 'flow'],
 		allowedRoles: true,
 		// 5.6 output Element
-		namingMethods: 'subtreeText'
+		namingMethods: ['subtreeText']
 	},
 	p: {
-		contentTypes: ['phrasing', 'flow'],
```
WilcoFiers, line 653: Why this change?

lib/commons/text/native-element-type.js
```diff
@@ -1,3 +1,4 @@
+// @deprecated use the standards object instead
```
WilcoFiers, line 1: This needs its own PR, so we can properly include a deprecation notice in the commit message.

lib/commons/text/native-text-alternative.js
```diff
@@ -1,6 +1,5 @@
 import getRole from '../aria/get-role';
-import matches from '../matches/matches';
-import nativeElementType from './native-element-type';
+import getElementSpec from '../standards/get-element-spec';
```
WilcoFiers, line 2: Changes in this file should go into a separate PR.

lib/commons/text/native-text-alternative.js
```diff
@@ -38,12 +37,8 @@ function nativeTextAlternative(virtualNode, context = {}) {
  * @return {Function[]} Array of native accessible name computation methods
  */
 function findTextMethods(virtualNode) {
-	const nativeType = nativeElementType.find(type => {
-		return matches(virtualNode, type.matches);
-	});
-
-	// Use concat because namingMethods can be a string or an array of strings
-	const methods = nativeType ? [].concat(nativeType.namingMethods) : [];
+	const elmSpec = getElementSpec(virtualNode);
```
WilcoFiers, line 40: I love this utility. This is so useful!

lib/standards/html-elms.js
straker: like `kdb` misspelling, it's a correct to the file as summary is not phrasing or flow content.

lib/standards/html-elms.js
straker: like `kdb` misspelling, it's a correct to the file as 'p' is not phrasing content.

lib/commons/text/native-text-alternative.js
straker: Need the property change for virtual node so i think changing it to use `elmSpec` is fine

Issue #693, opened 2018-01-23
Impact being bound to checks is limiting when composing new rules


rdeltour:
The _impact_ property ("minor", "moderate", "serious", or "critical") is currently bound to a _check_. It means that when composing a new _rule_, this rule will necessarily inherit the impacts of the underlying checks.

In my understanding however, a check is an "abstract" atomic piece of logic. A _rule_ on the contrary is the "concrete" implementation of a given accessible guideline. The actual impact of a violation only makes sense for a given rule.

For instance, the `non-empty-title` check has a `serious` impact. But there may be cases where a rule checks the presence of empty titles but isn’t about a serious violation (we have such rules in the digital publishing context). In that case, we have no other choice than implementing a custom check with a lesser impact, even if the implementation logic is exactly the same as `non-empty-title`.

A possible improvement IMO would be to associate the impact to the check only when the rule is composed, but not bake it in the check.

WilcoFiers:
Can you clarify what you mean with:
> associate the impact to the check only when the rule is composed

rdeltour:
> Can you clarify what you mean with:
>>    associate the impact to the check only when the rule is composed

If the rule’s violation impact depends on which check actually failed, then the checks’ impacts can be bound (or overridden) when the rule is configured. Concretely, the `any`, `all`, and `none` fields of the rule, could be set to (in addition to arrays of string IDs) arrays of tuples (ID,impact).

For instance:

```
{
  "id": "aria-roles",
  "selector": "[role]",
   (…)
  "all": [],
  "any": [],
  "none": [
    ["invalidrole", "critical"],
    ["abstractrole", "serious"],
    ["unsupportedrole", "moderate"]
  ]
}
```

WilcoFiers:
Had a quick talk with the team. I think your proposal overcomplicates things. It would certainly take a bunch of reworking of internal tools, so I'm not a fan of that. I think we should simplify this instead.  we think it would be better to simply allow an `impact` property on the rule `metadata` object. This can override whatever comes out of the checks.

rdeltour:
> we think it would be better to simply allow an impact property on the rule metadata object. This can override whatever comes out of the checks.

Yes, I think it's a good intermediary workaround.

In fact my first answer draft included something similar to what you describe, as an alternative to my comment above :-) but since the clarification was specifically about check/impact association, I ended up pruning that part…

[...]

PR #2393, opened 2020-07-17
feat(rule): optional impact on rules

WilcoFiers:
Allows an impact property on rules, which if used overrides the impact of all checks for that rule.

Includes the deletion of the unreleased `has-accessible-text` check, which was added because axe-core did not support this feature.

Closes issue: #693

## Reviewer checks


doc/API.md
```diff
@@ -202,7 +202,7 @@ axe.configure({
   - `rules` - Used to add rules to the existing set of rules, or to override the properties of existing rules
     - The rules attribute is an Array of rule objects
     - each rule object can contain the following attributes
-      - `id` - string(required). This uniquely identifies the rule. If the rule already exists, it will be overridden with any of the attributes supplied. The attributes below that are marked required, are only required for new rules.
+      - `id` - string(required). This uniquely identifies the rule. If the rule already exists, it will be overridden with any of the attributes supplied. The attributes below that are marked required, are only required for new rules. - `impact` - string(optional). Override the impact defined by checks
```
straker, line 205: Looks like a newline is needed before the `impact` list item. Right now it's part of the same paragraph as `id`.

PR #2394, opened 2020-07-17
chore(required-children): work with standards object

straker:
Huge refactor of the code since we aren't using `implicitNodes` anymore. This is the last bit of code that uses the `lookupTable`

Part of issue #2108 

## Reviewer checks

### DIFF
--- a/lib/commons/aria/required-owned.js
+++ b/lib/commons/aria/required-owned.js
@@ -11,11 +11,11 @@ import standards from '../../standards';
 function requiredOwned(role) {
 	const roleDef = standards.ariaRoles[role];
 
-  if (!roleDef || !Array.isArray(roleDef.requiredOwned)) {
-    return null;
-  }
+	 if (!roleDef || !Array.isArray(roleDef.requiredOwned)) {
+	 	 return null;
+	}
 
-  return [...roleDef.requiredOwned];
+	 return [...roleDef.requiredOwned];
 }
 
 export default requiredOwned;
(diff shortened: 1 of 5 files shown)


lib/standards/aria-roles.js
```diff
@@ -85,7 +85,7 @@ const ariaRoles = {
 	},
 	combobox: {
 		type: 'composite',
-		requiredOwned: ['textbox', 'listbox', 'tree', 'grid', 'dialog'],
+		requiredOwned: ['listbox', 'tree', 'grid', 'dialog', 'textbox'],
```
straker, line 88: This was to allow the tests to pass when the order of the required was different. Silly, but we don't have an easy way to check for "equal but order doesn't matter" in chai.

lib/commons/aria/required-owned.js
```diff
@@ -9,15 +9,13 @@ import lookupTable from './lookup-table';
...
-	if (roles) {
-		// TODO: es-module-utils.clone
-		owned = axe.utils.clone(roles.owned);
-	}
-	return owned;
+  if (!roleDef || !Array.isArray(roleDef.requiredOwned)) {
+    return null;
+  }
+
+  return [...roleDef.requiredOwned];
```
WilcoFiers, line 18: Think your editor went wrong:

```suggestion
	 if (!roleDef || !Array.isArray(roleDef.requiredOwned)) {
	 	 return null;
	}

	 return [...roleDef.requiredOwned];
```

PR #2399, opened 2020-07-18
fix(focusable-no-name): work with serial virtual nodes

straker:
Part of issue #2183

## Reviewer checks


test/checks/keyboard/focusable-no-name.js
```diff
@@ -76,4 +76,79 @@ describe('focusable-no-name', function() {
 			);
 		}
 	);
+
+	describe('Serial Virtual Node', function() {
```
WilcoFiers, line 80: This should also be able to get undefined if the children are not set.

PR #2400, opened 2020-07-20
feat: add layout-table-matches method

WilcoFiers:
TECH DEBT: There are no tests for this. Because this matches isn't used in any rule, it can't be accessed on the `axe` object. This can be solved once we have Babel enabled on our tests.

Closes issue:

## Reviewer checks

### DIFF
--- a/lib/rules/layout-table-matches.js
+++ b/lib/rules/layout-table-matches.js
@@ -1,6 +1,7 @@
 import { isDataTable } from '../commons/table';
 import { isFocusable } from '../commons/dom';
 
+// TODO: es-modules add tests. No way to access this on the `axe` object
 function dataTableMatches(node) {
 	return !isDataTable(node) && !isFocusable(node);
 }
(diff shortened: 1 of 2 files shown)


lib/rules/layout-table-matches.js
```diff
@@ -0,0 +1,9 @@
+import { isDataTable } from '../commons/table';
+import { isFocusable } from '../commons/dom';
+
+// TODO: es-modules add tests. No way to access this on the `axe` object
```
stephenmathieson, line 4: This looks like an "ES Module" to me. I do not understand this TODO.

stephenmathieson: I see, thanks for clarifying.

straker: I created a ticket for it so we don't forget https://github.com/dequelabs/axe-core/issues/2403

PR #2410, opened 2020-07-22
ci: Add "sync branches" workflow

straker:
## Reviewer checks

### DIFF
--- a/.github/workflows/sync-master-develop.yml
+++ b/.github/workflows/sync-master-develop.yml
@@ -11,6 +11,6 @@ jobs:
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           pr-title: "chore: merge master into develop"
-          pr-reviewers: straker,WilcoFiers
+          pr-reviewers: straker,WilcoFiers,stephenmathieson
           pr-labels: chore
           pr-template: .github/PULL_REQUEST_TEMPLATE.md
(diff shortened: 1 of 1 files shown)


.github/workflows/sync-master-develop.yml
```diff
@@ -0,0 +1,16 @@
+name: Sync master/develop branches
+on:
+  pull_request:
+    types: [closed]
+    branches: master
```
stephenmathieson, line 5: @scurker suggested we use the workflow like this instead:

```suggestion
on:
  push:
    branches:
      - master
```

So rather than running when _any_ PR into `master` is closed, only run when a commit lands in `master`.

.github/workflows/sync-master-develop.yml
```diff
@@ -0,0 +1,16 @@
...
+jobs:
+  create_sync_pull_request:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: dequelabs/action-sync-branches@v1.0.0
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          pr-title: "chore: merge master into develop"
+          pr-reviewers: straker,WilcoFiers
```
stephenmathieson, line 14: May as well add me here too. I'm happy to look at these PRs.

```suggestion
          pr-reviewers: straker,WilcoFiers,stephenmathieson
```

.github/workflows/sync-master-develop.yml
```diff
@@ -0,0 +1,16 @@
...
+          pr-labels: chore
```
stephenmathieson, line 15: I don't think this is useful. Why do you want to add a "chore" label to the PR?

```suggestion
```

PR #2406, opened 2020-07-21
docs: add docs for check options

straker:
I only included checks whose options modified how the check behaved. Checks which used generic checks (such as `non-empty-title`) were not added to list as their options are not really there for the user to customize. Generic checks will be described in another doc about creating custom rules.

Note: I do not understand what all the options for `p-as-heading` do even from looking at the code. If anyone can provide that understanding that'd be great.

Rendered view https://github.com/dequelabs/axe-core/blob/62e559fd7c65b8f426d01d99803854b3b9dcef6c/doc/check-options.md

Closes issue: #2187 

## Reviewer checks

**Required fields, to be filled out by PR reviewer(s)**

- [x] Follows the commit message policy, appropriate for next version
- [x] Code is reviewed for security

PR #2414, opened 2020-07-23
docs: document check evaluate functions

straker:
These are just the `aria` checks and if we approve what I've done here I'll continue to the rest.

The goal of documenting the evaluate functions is to be able to auto-generate the documentation for them using the JSDoc comments. This will be easier to maintain than a hard-coded list such as https://github.com/dequelabs/axe-core/pull/2406.

Also the reason to so explicitly document them as I have done is so anyone can understand what the function does, how to modify the check (such as editing the standards objects), what the data object will contain, and what the return types are and under what conditions.

This way someone unfamiliar with the code base should be able to create a custom check or understand what to do in order to modify an existing check and have all the information needed without having to look at the code base and figure it out.

Note: I know `@data` isn't a valid JSDoc tag, but since these aren't written for the JSDoc api docs, I figured we could parse the comment ourselves and get the information we needed.

@jkodu  I didn't quite understand the purpose of the aria check `valid-scrollable-semantics`. Would you be able to provide me with a description of its purpose?

## Reviewer checks


stephenmathieson:
@jkodu, not `@jey` ;)

lib/checks/aria/abstractrole-evaluate.js
```diff
@@ -1,6 +1,14 @@
 import { tokenList } from '../../core/utils';
 import { getRoleType } from '../../commons/aria';
 
+/**
+ * Check if an elements `role` attribute uses any abstract role values.
+ *
+ * Abstract roles are taken from the `ariaRoles` standards object from the roles `type` property.
+ *
+ * @data {String[]} List of all abstract role.
```
WilcoFiers, line 9: This doesn't seem to work. I generated the docs and this is not showing up in the definition of this function.

lib/checks/aria/abstractrole-evaluate.js
```diff
@@ -1,6 +1,14 @@
 import { tokenList } from '../../core/utils';
 import { getRoleType } from '../../commons/aria';
 
+/**
+ * Check if an elements `role` attribute uses any abstract role values.
+ *
+ * Abstract roles are taken from the `ariaRoles` standards object from the roles `type` property.
+ *
```
WilcoFiers, line 8: Should we put these into a namespace?

```suggestion
 *
 * @memberof checks
```

lib/checks/aria/abstractrole-evaluate.js
```diff
@@ -1,6 +1,14 @@
 import { tokenList } from '../../core/utils';
 import { getRoleType } from '../../commons/aria';
 
+/**
+ * Check if an elements `role` attribute uses any abstract role values.
```
WilcoFiers, line 5:
```suggestion
 * Check if an element's `role` attribute uses any abstract role values.
```

WilcoFiers (dismissed):

lib/checks/aria/abstractrole-evaluate.js
```diff
@@ -1,6 +1,14 @@
 import { tokenList } from '../../core/utils';
```
jkodu, line 1: The `validate-scrollable-semantics` evaluate function simply checks, if a given node when scrollable, has the allowed html tag type and has a valid [landmark](https://www.w3.org/TR/wai-aria/#landmark_roles) role. The mapping of html tags vs allowed roles, I am unsure how this was arrived but can dig deeper to find out if necessary. I have seen this in WCAG docs and also in HTML Spec, when reading for ACT rules. This is a best practice rule.

In terms of documenting it, I would keep this simple:

```js
/**
 * Check if a given scrollable element has the valid semantics 
 * 
 * @param {HTMLElement} node
 * @return {Boolean}
 */
```

lib/checks/aria/fallbackrole-evaluate.js
```diff
@@ -1,5 +1,10 @@
 import { tokenList } from '../../core/utils';
 
+/**
+ * Check that an element does not use more than one explicit role.
+ *
+ * @return {Boolean} True if the element uses only one explicit role. False otherwise.
```
straker, line 6:
```suggestion
 * @return {Boolean} True if the element uses more than one explicit role. False otherwise.
```

lib/checks/aria/has-widget-role-evaluate.js
```diff
@@ -1,5 +1,12 @@
 import { getRoleType } from '../../commons/aria';
 
+/**
+ * Check if an elements `role` attribute uses any widget role values.
```
straker, line 4:
```suggestion
 * Check if an elements `role` attribute uses any widget or composite role values.
```

straker:
Testing out these changes https://github.com/dequelabs/axe-core/pull/2414/commits/4f50fe4302bb1274e6eba2922e788dc8bd98c137

The output looks like this:

image

I looked into adding custom tag support, but doing so would mean we'd have to fork the [jsdoc template we use](https://github.com/nijikokun/minami) and add the custom tag information into it so it outputs.

WilcoFiers (approved):

WilcoFiers (approved):

Issue #2052, opened 2020-02-18
Consider not exporting global if AMD or CommonJS module definition succeeded.


engelsdamien:
Expectation: When loading axe as an AMD module, I don't expect to also have a global variable named 'axe'

Actual: Axe always export unconditionally on window when in browser context.

Motivation: Exporting on window can cause axe to override an existing value assigned to this name ('axe' is quite short and could be realistically picked as a name by a bundler/minifier for large enough bundles)

Code is here: https://github.com/dequelabs/axe-core/blob/deed0f150680dccf20c3199410a6f00960a202dd/lib/core/index.js#L26

straker:
Thanks for letting us know. We're actually in the works of converting the library to ES Modules and having something create that AMD / CommonJS code for us. Hopefully that will solve this problem.

engelsdamien:
Sounds good, do you have a pointer to an issue I could follow ?

straker:
We don't have an issue for it. Mostly has been tracked internally.

Andarist:
@straker are there any particular challenges regarding that in the axe-core codebase? I potentially could help you migrate the codebase to ES modules - usually, most of the work can be automated, but maybe your codebase uses some unusual patterns?

straker:
@Andarist The codebase is actually already converted to ES modules. Thanks for reminding me of this issue,  it might be too late to get into the 4.0 release but I'll take a look.

Andarist:
If I could do anything to help you ship this in 4.0 - let me know. I'm pretty proficient with modern bundling tools, build systems etc. I haven't worked with Grunt for a while though so I would propose replacing it with something else as it's not a popular tool nowadays.

straker:
Alright, so since this was a breaking change it's too late to go into the 4.0 release. We'll have to revisit this issue when looking at 5.0.

Andarist:
What's the estimate for v4? From what I see it has not been shipped yet, so maybe there still would be time for an external contributor to work on this one?

straker:
4.0 is schedule to go out on Monday. Since this is a breaking change (global axe no longer being loaded if using AMD or CommonJS) we had our code freeze for breaking changes last week. So even with a pr, we wouldn't be able to merge it.

Andarist:
With such a schedule I totally understand the code freeze. Thanks for letting me know!

PR #2419, opened 2020-07-24
fix(core): add proper UMD loader

straker:
Our UMD loader wasn't really a true UMD loader, so fixed it by using the template from https://github.com/umdjs/umd/blob/master/templates/returnExports.js. Due to our conditional loading of `axe.source` I had to do some additional wrapping of the UMD loader.

Note: this now prevents the `axe` object from going onto the global namespace if AMD or CommonJS are defined. This might be a breaking change, I'm not sure how to judge it based on the example puppeteer use. If so I can update it to always export, which means we close the linked issue as will not fix.

Closes issue: https://github.com/dequelabs/axe-core/issues/2052

## Reviewer checks

### DIFF
--- a/doc/examples/puppeteer/axe-puppeteer.js
+++ b/doc/examples/puppeteer/axe-puppeteer.js
@@ -27,7 +27,7 @@ const main = async url => {
 		// Inject and run axe-core
 		const handle = await page.evaluateHandle(`
 			// Inject axe source code
-			${axeCore.source}
+			var axe = ${axeCore.source}
 			// Run axe
 			axe.run()
 		`);
--- a/test/integration/full/umd/umd-module-exports.js
+++ b/test/integration/full/umd/umd-module-exports.js
@@ -2,8 +2,13 @@
 describe('UMD module.export', function() {
 	'use strict';
 
+	var axe;
+	beforeEach(function() {
+		axe = module.exports;
+	});
+
 	it('registers axe to module.exports', function() {
-		assert.strictEqual(module.exports, axe);
+		assert.hasAnyKeys(axe, ['utils', 'commons', 'core']);
 	});
 
 	it('does not use `require` functions', function() {
@@ -18,11 +23,5 @@ describe('UMD module.export', function() {
 
 	it('should ensure axe source includes axios', function() {
 		assert.isTrue(axe.source.includes(axe.imports.axios.toString()));
-	});
-
-	it('should include doT', function() {
-		var doT = axe.imports.doT;
-		assert(doT, 'doT is registered on axe.imports');
-		assert.equal(doT.name, 'doT');
 	});
 });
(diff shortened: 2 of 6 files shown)


straker:
As much as I agree we cannot as Wepback is not the final target of the build. Webpack is used to gather all the imports, but we still have to build the final `axe._load` call from the JSON metadata files into `core/index.js`, which is then concatenated with the Webpack output. Until we stop using JSON metadata files to build things we cannot use Webpack to generate the UMD output.

stephenmathieson:
> Until we stop using JSON metadata files to build things we cannot use Webpack to generate the UMD output.

I see... thanks for clarifying. I suppose this works for now then. 😢

test/integration/full/umd/umd-module-exports.js
```diff
@@ -19,10 +24,4 @@ describe('UMD module.export', function() {
 	it('should ensure axe source includes axios', function() {
 		assert.isTrue(axe.source.includes(axe.imports.axios.toString()));
 	});
-
-	it('should include doT', function() {
```
stephenmathieson, line 23: Why was this removed?

doc/examples/puppeteer/axe-puppeteer.js
```diff
@@ -27,7 +27,7 @@ const main = async url => {
 		// Inject and run axe-core
 		const handle = await page.evaluateHandle(`
 			// Inject axe source code
-			${axeCore.source}
+			var axe = ${axeCore.source}
```
stephenmathieson, line 30: Why is this necessary? I don't think I understand what's happening here.

straker: Puppeteer loads [CommonsJS by default](https://github.com/puppeteer/puppeteer/issues/3685#issuecomment-453229607) so there is no global object after the source runs.

test/integration/full/umd/umd-module-exports.js
straker: Mistake

doc/examples/puppeteer/axe-puppeteer.js
stephenmathieson: This is not true. The comment you linked to is discussing something completely different.

Within the `Page#evaluateHandle` context, there is no `module` defined. Here's a script proving this:

```js
const puppeteer = require('puppeteer')

const test = async () => {
  const browser = await puppeteer.launch()
  const page = await browser.newPage()

  const handle = await page.evaluateHandle(`
    Promise.resolve({
      module: typeof module,
      define: typeof define,
      window: typeof window
    })
  `)

  const types = await handle.jsonValue()
  console.log('typeof module: "%s"', types.module)
  console.log('typeof define: "%s"', types.define)
  console.log('typeof window: "%s"', types.window)

  await page.close()
  await browser.close()
}

test().catch(err => {
  console.error(err)
  process.exit(1)
})
```

This will output:

```
typeof module: "undefined"
typeof define: "undefined"
typeof window: "object"
```

I do not understand why we need to add `var axe = ` in front of `axe.source`, and believe it's due to a bug.

straker:
This is a breaking change since axe is removed from the global scope when using AMD or CommonJS. We can't do this now and will have to wait for 5.0 to do this if we still want to.

PR #2423, opened 2020-07-27
fix(rule,check): allow function ids for matches property in rule.configure

straker:
@downzer0 discovered this one when trying to configure a rule with a `matches` function ID. I added tests to ensure this should work now as well as throw a more helpful message as the default was eval the string as code and you'd get a `thing is not defined` error which was super unhelpful. So now you should get an error with the name of the string you used.

## Reviewer checks

### DIFF
--- a/lib/core/base/check.js
+++ b/lib/core/base/check.js
@@ -10,7 +10,7 @@ export function createExecutionContext(spec) {
 		}
 
 		// execution contexts can only be functions
-		if (spec.includes('function')) {
+		if (/^\s*function[\s\w]*\(/.test(spec)) {
 			return new Function('return ' + spec + ';')();
 		}
 
--- a/lib/core/base/check.js
+++ b/lib/core/base/check.js
@@ -9,7 +9,14 @@ export function createExecutionContext(spec) {
 			return metadataFunctionMap[spec];
 		}
 
-		return new Function('return ' + spec + ';')();
+		// execution contexts can only be functions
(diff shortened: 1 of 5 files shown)


lib/core/base/check.js
```diff
@@ -9,7 +9,14 @@ export function createExecutionContext(spec) {
 			return metadataFunctionMap[spec];
 		}
 
-		return new Function('return ' + spec + ';')();
+		// execution contexts can only be functions
+		if (/^\s*function[\s\w]*\(/.test(spec)) {
```
stephenmathieson, line 13: What about "arrow functions"?

test/core/base/check.js
```diff
@@ -98,6 +98,62 @@ describe('Check', function() {
...
+				axe._load({});
+				axe._audit.metadataFunctionMap['custom-evaluate'] = function() {
+					return 'fong';
+				};
+
+				Check.prototype.test = function() {
+					return this.evaluate();
+				};
+				var check = new Check({
+					evaluate: 'function () { return "foo"; }'
```
stephenmathieson, line 111: What happens if we do this:

```js
evaluate: '() => "foo"'
```

straker: As far as I know we don't allow that. Although it is possible, we only support doT.js template output which uses `functions`.

@WilcoFiers do you find this a problem (especially since we deprecated this?)

Issue #2421, opened 2020-07-27
Call out missing accessible name on various ARIA roles


eps1lon:
Expectation: `<div role="tree" />` triggers a violation because it does not have an accessible name. Example: https://2m3yw.csb.app/

Actual: No violation reported

Motivation:

- `role="tree"` requires an accessible name in [ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/#tree) and [1.2](https://www.w3.org/TR/wai-aria-1.2/#tree).
- the tree will be announced as a generic "tree" in NVDA (possibly any screen reader) making it hard to understand the context of a specific item (though not as bad as having a generic `<button />`).
- Required to add an audit to lighthouse (https://github.com/GoogleChrome/lighthouse/issues/11155)

<pre><code>
axe-core version: 3.5.5
chrome extension: 4.5.3
</code></pre>

axe-core should probably report any element without a name that has a role that requires an accessible name.

Rules to add:
Taken from https://github.com/dequelabs/axe-core/issues/2421#issuecomment-667045198

- [x] ~~aria-structure-name (NEW) - tag: best-practice (https://github.com/dequelabs/axe-core/pull/2431)~~
- [x] aria-dialog-name (NEW) - tag: best-practice #2609
- [x] ~~aria-table-header-name (NEW) - tag: 1.3.1~~
- [x] aria-treeitem-name (NEW) - tag: best-practice #2615
- [x] aria-meter-name (NEW) - tag: sc111 #2607
- [x] aria-toggle-field-name: add "option" #2605 
- [x] ~~aria-input-field-name: add "radiogroup" - skipped, not sure if this is right~~
- [x] aria-command-name: (NEW) - tag: sc412 #2571 
- [x] aria-progressbar-name (NEW) - tag: sc111 #2555
- [x] aria-tooltip-name (NEW) - tag: best-practice #2548
- [x] ~~aria-marquee-name - not happening because of w3c/aria#1339~~

straker:
Thanks for the proposal. It's up to @WilcoFiers if we want the new rule, but I'll add some information on how we would write it:

```json
// rules/aria-required-name.json
{
	"id": "aria-required-name",
	"selector": "[role]",
	"metadata": {
		"description": "Ensures elements with ARIA roles that require an accessible name have one,
		"help": "Required accessible name must be present"
	},
	"all": [],
	"any": ["aria-required-name"],
	"none": []
}
```

[...]

PR #2431, opened 2020-07-30
feat: Add aria-structure-name rule

eps1lon:
Adds two new rules that catch elements which require an accessible name but are unnamed. Two separate rules since not every naming technique applies to all elements. If we can give actionable advise based on the role without having to create multiple rules then I'd be in favor of that.

Closes issue: None. But it is part of #2421.

## Reviewer checks

### DIFF
--- a/lib/rules/name-author-contents.json
+++ b/lib/rules/name-author-contents.json
@@ -1,0 +1,17 @@
+{
+	"id": "name-author-contents",
+	"selector": "[role=\"treeitem\"]",
+	"tags": [],
+	"metadata": {
+		"description": "Ensures elements with ARIA roles that require an accessible name have one from author or contents.",
+		"help": "Required accessible name must be present (author or contents)"
+	},
+	"all": [],
+	"any": [
+		"aria-label",
+		"aria-labelledby",
+		"has-visible-text",
+		"non-empty-title"
+	],
+	"none": []
+}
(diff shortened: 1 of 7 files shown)


lib/rules/name-author-contents.json
```diff
@@ -0,0 +1,17 @@
+{
+	"id": "name-author-contents",
+	"selector": "[role=\"treeitem\"]",
+	"tags": [],
+	"metadata": {
+		"description": "Ensures elements with ARIA roles that require an accessible name have one from author or contents.",
```
straker, line 6: I would avoid talking about author or content here for the same reasons as stated above.

WilcoFiers:
@eps1lon Would you be able to get this updated in the next few days? We're working towards a 4.1 release and would need to have this merged by November 2nd.

eps1lon:
Can't make any promises with regards to a timeline. As far as I can tell this PR is outdated anyway since we went back and forth on the implementation.

WilcoFiers:
I have received some feedback about these ARIA name rules. See my update here: https://github.com/dequelabs/axe-core/issues/2421#issuecomment-717871355. This particular rule will not be added. There are a few more rules in this batch that needs to be done. @eps1lon would you be up for taking on the aria-dialog-name rule?

eps1lon:
> would you be up for taking on the aria-dialog-name rule?

Will do :+1: I'm closing this PR since most of the conversation is obsolete anyway.

PR #2435, opened 2020-08-04
fix(checks): do not normalize options for custom checks

straker:
Prevent breaking change to custom check options. Moved the `metadataFunctionMap` to the private test object as it was suppose to be there to begin with and not on the private `audit` object.

#Reviewer checks

### DIFF
--- a/lib/core/base/check.js
+++ b/lib/core/base/check.js
@@ -180,12 +180,23 @@ Check.prototype.runSync = function(node, options, context) {
  */
 
 Check.prototype.configure = function(spec) {
+	// allow test specs (without evaluate functions) to work as
+	// internal checks
+	if (!spec.evaluate || metadataFunctionMap[spec.evaluate]) {
+		this._internalCheck = true;
+	}
+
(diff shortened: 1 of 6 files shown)


lib/core/base/check.js
```diff
@@ -180,21 +180,37 @@ Check.prototype.runSync = function(node, options, context) {
...
 	['evaluate', 'after']
 		.filter(prop => spec.hasOwnProperty(prop))
 		.forEach(prop => (this[prop] = createExecutionContext(spec[prop])));
 };
 
-Check.prototype.getOptions = function getOptions(options = {}) {
-	return deepMerge(this.options, normalizeOptions(options));
+Check.prototype.getOptions = function getOptions(options) {
+	// only merge and normalize options for internal checks
+	if (this._internalCheck) {
```
WilcoFiers, line 209: I think we allow deep merge for any non-array object

```suggestion
	// only merge and normalize options for internal checks
	if (typeof this.options === 'object' && Array.isArray(this.options)) {
```

Issue #2284, opened 2020-06-10
Create best practice rule to flag elements with role conflict resolution


straker:
We should create a best practice rule that flags elements whose role is `none` or `presentation` and which cause the [role conflict resolution to trigger](https://github.com/dequelabs/axe-core/issues/2212). Role conflict resolution is not a well known concept and getting into that situation causes inconsistent results between browsers. It's also not something you should rely on for accessibility. 

The issue probably shouldn't fail but instead return needs review only.

```html
<!-- this should be flagged as incomplete as the user probably 
     expects this to not be a heading when it in fact is -->
<h1 role="none" aria-label="My Heading"></h1>
```

WilcoFiers:
So, to summarise how to approach this:

1. Needs a new rule, which has a selector of [role=none], [role=presentation]
2. Create two new checks, which go in "all"
    - First check just looks if the element is not focusable. It can use dom.isFocusable()
    - Second check should look if the virtualNode has a global ARIA attribute. There is no commons for this, but it's fairly straight forward:

```js
import lookupTable from '../../commons/aria/lookup-table';

lookupTable.globalAttributes.some(attr =>
	virtualNode.hasAttr(attr)
);
```

Make sure to add tests to `test/checks` and to `test/integration/rules`.

straker:
Note: It might be easier to just check if the explicit role matches the semantic role as the semantic role does all the checking for us :)

padmavemulapati:
Verified new best-practice rule: presentation-role-conflict with axe-core4.0.2 latest code:
1. When the rule passes:
it passes with `<div id="pass1" role="presentation">Test</div>` and `<div id="pass2" role="presentation">Test Button</div>`
![image](https://user-images.githubusercontent.com/47409579/95037897-3cf90a00-06ea-11eb-9a50-70eba02b494f.png)
and the details about the pass condition: 
![image](https://user-images.githubusercontent.com/47409579/95037980-86495980-06ea-11eb-87f9-381ca50d12d7.png)

2. When the rule fails:
it fails with  `<li id="violation1" role="presentation" aria-label="My Heading">Hello</li>` and 
`<h1 id="violation2" role="none" aria-label="My Heading">Hello</h1>`

![image](https://user-images.githubusercontent.com/47409579/95037999-9f520a80-06ea-11eb-91b1-1407efa25838.png)

details on failure:
[...]

PR #2440, opened 2020-08-06
feat(presentation-role-conflict): create rule to flag elements with role conflict resolution #2284

jlin95:
Create best practice rule to flag elements with role conflict resolution #2284

## Reviewer checks


lib/checks/aria/has-global-aria-attribute-evaluate.js
```diff
@@ -0,0 +1,7 @@
+import lookupTable from '../../commons/aria/lookup-table';
```
straker, line 1: We recently released 4.0 in which we deprecated the lookup-table. You'll want to use the new `standards` object to get the [list of global attributes](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/standards/get-global-aria-attrs.js)

doc/rule-descriptions.md
```diff
@@ -62,10 +62,11 @@
...
+| :----------------------------------------------------------------------------------------------------------------- | :--------------------------------------------------------------------------------------------------------- | :------ | :--------------------------- | :--------- |
+| [autocomplete-valid](https://dequeuniversity.com/rules/axe/3.5/autocomplete-valid?application=RuleDescription)     | Ensure the autocomplete attribute is correct and suitable for the form field                               | Serious | cat.forms, wcag21aa, wcag135 | failure    |
+| [avoid-inline-spacing](https://dequeuniversity.com/rules/axe/3.5/avoid-inline-spacing?application=RuleDescription) | Ensure that text spacing set through style attributes can be adjusted with custom stylesheets              | Serious | wcag21aa, wcag1412           | failure    |
+| [flag-role](https://dequeuniversity.com/rules/axe/3.5/flag-role?application=RuleDescription)                       | Flags elements whose role is none or presentation and which cause the role conflict resolution to trigger. | Minor   | cat.aria                     | failure    |
```
straker, line 69: The rule name is a bit vague. Ideally it would tell me what the rule is checking, like `button-name` or `image-alt`. `flag-role` doesn't really tell me what the rule is checking.

Maybe something like `role-conflict`?

lib/checks/aria/is-element-focusable-evaluate.js
```diff
@@ -0,0 +1,7 @@
+import { isFocusable } from '../../commons/dom';
+
+function isElementFocusableEvaluate(node) {
+	return isFocusable(node);
```
straker, line 4: Could you pass the `virtualNode` here instead of the `node`? We've been trying to use `node` less.

```suggestion
function isElementFocusableEvaluate(node, options, virtualNode) {
	return isFocusable(virtualNode);
```

test/checks/aria/has-global-aria-attribute.js
```diff
@@ -0,0 +1,37 @@
...
+		axe._tree = undefined;
+		checkContext.reset();
+	});
+
+	afterEach(function() {
+		fixture.innerHTML = '';
+		axe._tree = undefined;
+		checkContext.reset();
+	});
+
```
straker, line 21: You have two `afterEach` which are the same.

```suggestion
```

lib/rules/flag-role.json
```diff
@@ -0,0 +1,13 @@
+{
+	"id": "flag-role",
+	"selector": "[role=\"none\"], [role=\"presentation\"]",
+	"tags": ["cat.aria"],
+	"metadata": {
+		"description": "Flags elements whose role is none or presentation and which cause the role conflict resolution to trigger.",
+		"help": "Elements of role none or presentation should be flagged"
+	},
+	"preload": false,
+	"all": ["is-element-focusable", "has-global-aria-attribute"],
```
straker, line 10: So this will pass the rule if the element is both focusable and has a global aria attribute. Instead it should fail the rule, so I believe these should be in the `none` array.

test/integration/rules/flag-role/flag-role.html
```diff
@@ -0,0 +1,8 @@
+<!-- HTML Snippets-->
+<div role="presentation" id="pass1" tabindex="1">
+	ok
+</div>
+<h1 role="none" id="pass2" aria-label="My Heading"></h1>
+
+<div role="scrollbar" id="violation1">fail</div>
```
straker, line 7: This should pass the rule as it does not have a conflict resolution.

lib/rules/flag-role.json
```diff
@@ -0,0 +1,13 @@
+{
+	"id": "flag-role",
+	"selector": "[role=\"none\"], [role=\"presentation\"]",
+	"tags": ["cat.aria"],
```
straker, line 4: Please add the `best-practice` tag.
```suggestion
	"tags": ["cat.aria", "best-practice"],
```

lib/rules/flag-role.json
```diff
@@ -0,0 +1,13 @@
+{
+	"id": "flag-role",
```
WilcoFiers, line 2: How about:
```suggestion
	"id": "presentation-role-confict",
```

lib/checks/aria/has-global-aria-attribute-evaluate.js
jlin95: Resolved in PR

lib/checks/aria/has-global-aria-attribute.json
```diff
@@ -0,0 +1,11 @@
+{
+	"id": "has-global-aria-attribute",
+	"evaluate": "has-global-aria-attribute-evaluate",
+	"metadata": {
+		"impact": "minor",
+		"messages": {
+			"pass": "Element has global ARIA attribute",
```
straker, line 7: It would be helpful to the user to know which attribute(s) are the global ones. In the evaluate function, if you call `this.data()` and pass it an array of the global aria attributes that the element has, you can then use `${data.values}` and the [singular/plural structure in the message](https://github.com/dequelabs/axe-core/blob/develop/doc/check-message-template.md#singular-and-plural-messages) to have them listed out for the user:

```js
// has-global-aria-attribute-evaluate.js 
const globalAttrs = getGlobalAriaAttrs().filter(attr => virtualNode.hasAttr(attr));
this.data(globalAttrs);
return globalAttrs.length > 0;

// has-global-aria-attribute.json
{
  "messages": {
    "pass": {
      "singular": "Element has global ARIA attribute: ${data.vaules}",
      "plural": "Element has global ARIA attributes: ${data.vaules}"
    }
  }
}
```

straker (changes requested):
Great work! The changes all look great I only have 1 suggestion that should help the end user know which attributes are global.

straker:
Oh, and also to fix the tests you'll need to update to the latest develop again (we disabled chrome-mobile because it started giving us problems).

lib/checks/aria/has-global-aria-attribute.json
jlin95: Fixed! Thanks for this :)

straker (approved):
Fantastic work. Creating new rules isn't an easy task, so thanks for taking this on

jlin95:
Thank you @WilcoFiers @straker for guidance and review!

PR #2444, opened 2020-08-11
fix(color-contrast): dont error for floating element

straker:
Found this as a bug when testing on a site. The variable is reassigned to an empty array under certain conditions.

## Reviewer checks


straker:
Probably, I'll have to go dig around the color-contrast test code for where this code would be triggered.

Issue #2475, opened 2020-08-20
Update aria-hidden-body-evaluate to work with virtualNodes


straker:
The [aria-hidden-body-evaluate](https://github.com/dequelabs/axe-core/blob/develop/lib/checks/aria/aria-hidden-body-evaluate.js) function should work with `virtualNode` rather than `node`. Update the function call to use `virtualNode` and use [virtual-node](https://github.com/dequelabs/axe-core/blob/develop/lib/core/base/virtual-node/virtual-node.js) methods to get the attribute.

You'll also need to update the [tests](https://github.com/dequelabs/axe-core/blob/develop/test/checks/aria/aria-hidden-body.js) to pass the virtual node rather than just the node.

See [non-empty-if-presenet-evaluate](https://github.com/dequelabs/axe-core/blob/develop/lib/checks/shared/non-empty-if-present-evaluate.js) and its [test file](https://github.com/dequelabs/axe-core/blob/develop/test/checks/shared/non-empty-if-present.js) as an example.

shankarshastri:
@straker , I was trying to fix this.

**aria-hidden-body.js**
```javascript
describe("aria-hidden", function () {
  "use strict";

  var checkContext = axe.testUtils.MockCheckContext();
  var queryFixture = axe.testUtils.queryFixture;

  afterEach(function () {
    checkContext.reset();
  });

  it("should not be present on document.body", function () {
    var vNode = queryFixture('<body></body>');
    assert.isTrue(
      axe.testUtils
        .getCheckEvaluate("aria-hidden-body")
        .call(checkContext, null, {}, vNode)
    );
  });

  it("fails appropriately if aria-hidden=true on document.body", function () {
    var vNode = queryFixture('<body aria-hidden="true"></body>');
    assert.isFalse(
      axe.testUtils
        .getCheckEvaluate("aria-hidden-body")
        .call(checkContext, null, {}, vNode)
    );
  });

  it("passes if aria-hidden=false on document.body", function () {
    var vNode = queryFixture('<body aria-hidden="false"></body>');
    assert.isTrue(
      axe.testUtils
        .getCheckEvaluate("aria-hidden-body")
        .call(checkContext, null, {}, vNode)
    );
  });
});
```

**aria-hidden-body-evaluate.js**
```javascript
function ariaHiddenBodyEvaluate(node, options, virtualNode) {
	return virtualNode.hasAttr('aria-hidden') !== 'true';
}

[...]

PR #2487, opened 2020-08-26
refactor(aria-hidden-body): work with virtual nodes

shankarshastri:
Closes issue: #2475 

## Reviewer checks

- [X] Code is reviewed for security

### DIFF
--- a/lib/checks/aria/aria-hidden-body-evaluate.js
+++ b/lib/checks/aria/aria-hidden-body-evaluate.js
@@ -1,8 +1,5 @@
 function ariaHiddenBodyEvaluate(node, options, virtualNode) {
-	return (
-		virtualNode.hasAttr('aria-hidden') !== true ||
-		virtualNode.attr('aria-hidden') === 'false'
-	);
+	return virtualNode.attr('aria-hidden') !== 'true';
 }
 
 export default ariaHiddenBodyEvaluate;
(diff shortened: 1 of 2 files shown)


shankarshastri:
@straker , I have added the draft PR for the same, even after adding the id=target, its having issues. Could you please let me know how I can proceed?

straker:
Ohhhh. Sorry, I see the problem now. `queryFixture` tries to append the HTML content to the body, but that isn't allowed as you cant add `body` to itself. So nothing is being added and thus nothing returned. We'll have to go around that and setup the virtual tree for every test and get the body element to pass to it.

```js
it('should not be present on document.body', function() {
  var tree = axe.testUtils.flatTreeSetup(document.body);
  assert.isTrue(
    axe.testUtils
      .getCheckEvaluate('aria-hidden-body')
      .call(checkContext, null, {}, tree[0])
  );
});
```

lib/checks/aria/aria-hidden-body-evaluate.js
```diff
@@ -1,5 +1,8 @@
-function ariaHiddenBodyEvaluate(node) {
-	return node.getAttribute('aria-hidden') !== 'true';
+function ariaHiddenBodyEvaluate(node, options, virtualNode) {
+	return (
+		virtualNode.hasAttr('aria-hidden') !== true ||
+		virtualNode.attr('aria-hidden') === 'false'
+	);
```
straker, line 5: Interesting change. Did this solve a problem you ran into? Was `virtualNode.attr('aria-hidden') !== 'true'` not working?

shankarshastri: My bad, fixed it. I was actually exploring the virtual-node js file and tried multiple functions of the same.

Issue #2454, opened 2020-08-14
role-none / role-presentation should consider conflict resolution


WilcoFiers:
**expectation**:

The role-none / role-presentation checks should fail if the element is focusable or if the element has a global ARIA attribute. It should provide a specific fail message (using `messageKey`) that explains that the role is ignored because it is focusable / has a global aria attribute.

This failure is because those roles get ignored by browsers, as describe in [presentation role conflict resolution](https://www.w3.org/TR/wai-aria-1.1/#conflict_resolution_presentation_none).

**Actual**:
Axe-core should fail the following elements, because their role is ignored by browsers:

```html
<button role="none">Still a button</button>
```

```html
<img role="none" aria-live="assertive" />
```

Issue #2502, opened 2020-09-01
image with empty alt and global aria-attribute should not pass the image-alt rule


straker:
The following code snippet should not pass the `image-alt` rule. The problem is that our `has-alt` check passes as the image has an empty alt, but the image is not considered presentational because of role conflict resolution of the global `aria-label` attribute.

```html
<img src="img.jpg" id="violation11" alt="" aria-label="" />
```

straker:
Tested this in:

* Safari/VO
* IE11/JAWS
* Chrome/JAWS
* Firefox/NVDA

and all report the image as "blank" but not as an image element (just as in a newline character or empty tag). Thus this isn't an accessibility problem.

PR #2503, opened 2020-09-01
feat(checks): deprecate role-none and role-presentation for presentational-role

straker:
The `presentational-role` check is written in such a way that as long as the end role of the element is presentation/none the check will pass. This enables it to be used for things that don't use explicit role to make it presentation, such as an images empty `alt` attribute. That should help when fixing #2502 so we don't have to check the role in that check.

Also, I removed the presentational check on the `link-name` rule as there is no way to disable a link from being focusable without a) making it hidden which prevents the rule from running on it or b) removing the `href` attribute which goes against the `selector` of the rule.

Closes issue: #2454 

## Reviewer checks

### DIFF
--- a/build/configure.js
+++ b/build/configure.js
@@ -87,6 +87,7 @@ function buildRules(grunt, options, commons, callback) {
 		var tags = options.tags ? options.tags.split(/\s*,\s*/) : [];
 		var rules = result.rules;
 		var checks = result.checks;
+		parseChecks(checks);
 
 		// Translate checks
 		if (locale && locale.checks) {
--- a/lib/checks/shared/presentational-role.json
+++ b/lib/checks/shared/presentational-role.json
@@ -9,7 +9,7 @@
 				"default": "Element's default semantics were not overridden with role=\"none\" or role=\"presentation\"",
 				"globalAria": "Element's role is not presentational because it has a global ARIA attribute",
 				"focusable": "Element's role is not presentational because it is focusable",
-				"both": "Element's role is not presentational because it has a global aria attribute and is focusable"
+				"both": "Element's role is not presentational because it has a global ARIA attribute and is focusable"
 			}
 		}
 	}
--- a/lib/checks/shared/presentational-role.json
+++ b/lib/checks/shared/presentational-role.json
@@ -1,0 +1,16 @@
+{
+	"id": "presentational-role",
+	"evaluate": "presentational-role-evaluate",
+	"metadata": {
+		"impact": "minor",
+		"messages": {
+			"pass": "Element's default semantics were overriden with role=\"${data.role}\"",
+			"fail": {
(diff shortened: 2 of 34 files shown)


build/configure.js
```diff
@@ -87,6 +87,7 @@ function buildRules(grunt, options, commons, callback) {
 		var tags = options.tags ? options.tags.split(/\s*,\s*/) : [];
 		var rules = result.rules;
 		var checks = result.checks;
+		parseChecks(checks);
```
straker, line 90: From the code, a check's metatdata was only added to `axe._load` if a rule used it. Since `role-none` and `role-presentation` were no longer used in any rule, their metadata was never added. This caused any translation file that passed translations for those checks to fail `axe.configure` with 

> "Locale provided for unknown check: "role-none""

To fix I added this line to parse all checks regardless of if a rule uses it or not.

lib/checks/shared/presentational-role.json
```diff
@@ -0,0 +1,16 @@
+{
+	"id": "presentational-role",
+	"evaluate": "presentational-role-evaluate",
+	"metadata": {
+		"impact": "minor",
+		"messages": {
+			"pass": "Element's default semantics were overriden with role=\"${data.role}\"",
+			"fail": {
+				"default": "Element's default semantics were not overridden with role=\"none\" or role=\"presentation\"",
+				"globalAria": "Element's role is not presentational because it has a global aria attribute",
```
WilcoFiers, line 10:
```suggestion
				"globalAria": "Element's role is not presentational because it has a global ARIA attribute",
```

lib/checks/shared/presentational-role.json
```diff
@@ -0,0 +1,16 @@
...
+				"focusable": "Element's role is not presentational because it is focusable",
+				"both": "Element's role is not presentational because it has a global aria attribute and is focusable"
```
WilcoFiers, line 12:
```suggestion
				"both": "Element's role is not presentational because it has a global ARIA attribute and is focusable"
```

test/integration/rules/frame-title/frame-title.html
```diff
@@ -32,3 +32,28 @@
 	role="presentation"
 ></iframe>
 <iframe src="frame-title/frames/level1a.html" id="pass7" role="none"></iframe>
+
+<iframe
+	src="frame-title/frames/level1a.html"
+	id="violation4"
+	role="none"
+	aria-live="assertive"
+></iframe>
```
WilcoFiers, line 41: We may have to do more work for iframes here later. I don't think this is correct, but we'll get to it when Jey finishes his work on iframes in ACT. Acceptable for now.

WilcoFiers (changes requested):
minor stuff

WilcoFiers (approved):

Issue #1618, opened 2019-06-06
Some rules not running with Cypress


alisd23:
**Expectation**: Run the `document-title` and `html-has-lang` rules when running axe-core inside Cypress (Which runs the whole app inside an iFrame)

**Actual**: The rules are identified as *inapplicable*

**Motivation**:
The `document-title` and `html-has-lang` rules should be able to run correctly when the whole app being tested is mounted inside an iFrame, which is what happens in the case of a Cypress test.

Currently I *think* what's happening is that these 2 rules use the `window-is-top.js` check, which causes the rule to only run for the **root/top** window object - i.e. not the actual window to be tested in a Cypress test.

Could there be a way to define a "root" or "top" window/document, so that this `window-is-top` code checks against a specific window object, instead of the default top window?

```
axe-core version: 3.2.2
Browser: Chrome
```

straker:
Thanks for the issue. We'll have to investigate to see what can be done.

dylanb:
@straker perhaps we could add an option that forces the running of the pageLevel rules, overriding the automatic detection

straker:
So I've done some research. I verified that `window-is-top` is preventing the rules from running inside an iframe, which is what it [was suppose to do](https://github.com/dequelabs/axe-core/issues/1424). However, this prevents it from ever running inside an iframe even if that is what is desired (as is the case with Cypress). 

I think an option would work well. We can either provide an option to that tells axe it's being run inside an iframe and set it when axe injects itself into iframes, or an option that tells axe it's the top frame even if it's really not. The latter is more backwards compatible.

```js
axe.run({ isTopFrame: true });
```

alisd23:
Sounds like a good solution to me. 👍 

Could you please give some quick advice/tips on where to start in adding this feature? If I have time in the future, and if no one picks it up before me - I might try give this a go.

straker:
We'll need to get the passed in option to the `window-is-top` matcher so we can check for it as well. We currently don't pass any options into matcher functions, so you'd need to do this.

[...]

PR #2520, opened 2020-09-15
feat(rule-matches): deprecate window-is-top-matches for is-initiator-matches

straker:
This allows Cypress to run `document-title` and `html-has-lang` rules inside the iframe by making whichever context `axe.run` was called the "top-level" context (be it an iframe or the top-most window).

Closes issue: #1618 

## Reviewer checks

### DIFF
--- a/lib/rules/window-is-top-matches.js
+++ b/lib/rules/window-is-top-matches.js
@@ -1,3 +1,4 @@
+// @deprecated
 function windowIsTopMatches(node) {
 	return (
 		node.ownerDocument.defaultView.self === node.ownerDocument.defaultView.top
--- a/.jsdoc.json
+++ b/.jsdoc.json
@@ -1,5 +1,6 @@
 {
 	"tags": {
+		"allowUnknownTags": true,
 		"dictionaries": ["jsdoc"]
 	},
 	"source": {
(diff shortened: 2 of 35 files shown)


lib/rules/window-is-top-matches.js
```diff
@@ -1,5 +1,8 @@
+import cache from '../core/base/cache';
+
 function windowIsTopMatches(node) {
 	return (
+		cache.get('isTopWindow') ||
```
WilcoFiers, line 5: Why not pass the options around, or record what we think is the top level in the context object and pass that around? This is basically using the cache as a way to create global variables. Global variables is something we're trying to get out of, lets not add more of them.

WilcoFiers (changes requested):

lib/rules/window-is-top-matches.js
stephenmathieson: I agree, but think this might be a naming problem. If `cache` was called `globalState`, I don't think we'd have any issue putting "global data" here.

straker: The main issue is that we pass both the `context` and the `options` directly to the iframes, so any property we put into either of those will also trigger in the iframes.

straker: Just say that `initiator` variable on the context. Instead of an options, we could just always run those rules if `initiator: true` is set (which is always the top-most `run` call).

straker: Deprecated `window-is-top` matches for `is-initiator-matches` and just check if the context is the initiator to run those rules.

Issue #2524, opened 2020-09-21
Objects should not allow fallback content


WilcoFiers:
The following should fail axe-core. Fallback content on objects is not used to compute an accessible name.

```html
<object data="...">
  Fallback content
</object>
```

For details, see the discussion about this in the WCAG issues: https://github.com/w3c/wcag/issues/1138

padmavemulapati:
Verified with the test content
Pass criteria: 
 `<object id="pass1" title="This object has text"></object>` ,
`<object id="pass2" aria-label="this object has text"></object>` , 
and failure: 
	`<object id="violation1"></object>` , 
 `<object id="violation2"><div></div></object>` and 
	`<object id="violation3">This object has text.</object>`
not allowing fallback content on the data , properly pass and fail meeting the requirements
![image](https://user-images.githubusercontent.com/47409579/95086648-5d52b400-073e-11eb-97c9-664e5caa9a3c.png)

yecril71pl:
Please change the summary to something less horrible, like *Object fallback content marks it accessible*.

PR #2525, opened 2020-09-21
fix(object-alt): do not allow fallback content for objects

WilcoFiers:
Fallback content can not be used to give an object an accessible name. Instead us aria-label(ledby) or title.

Closes issue: #2524

## Reviewer checks

### DIFF
--- a/lib/rules/object-alt.json
+++ b/lib/rules/object-alt.json
@@ -14,7 +14,6 @@
 	},
 	"all": [],
 	"any": [
-		"has-visible-text",
 		"aria-label",
 		"aria-labelledby",
 		"non-empty-title",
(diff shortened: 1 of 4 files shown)


test/playground.html
```diff
@@ -2,7 +2,68 @@
 <html lang="en">
 	<title>O hai</title>
 
-	<div class="foo" id="foo">foo</div>
+	<div class="listbox-area">
```
straker, line 5: Glad I'm not the only one who submits changes to the playground file :)

straker (dismissed):
Looks good minus the playground change

test/playground.html
WilcoFiers: whahaha, fair enough :P

straker (approved):

PR #2526, opened 2020-09-21
chore: remove CircleCI badge

dylanb:
Remove the CircleCI badge because it is no longer needed now that we have Slack integration and it changes status multiple times over the day making us look bad on npm

## Reviewer checks


stephenmathieson:
👎 from me

I find the badge helpful. I think the problem is that our tests are unstable, not that the CI badge is "stupid".

WilcoFiers (approved):

dylanb:
@stephenmathieson if the badge did not show up on npmjs.com, then that would be ok, but it does and it makes us look bad on npmjs.com

Issue #2357, opened 2020-07-09
Axe-core file size


straker:
Below is a table of the sizes of `axe.min.js` for the last few releases since v3.0.

| Release | Minified | Gzipped |
| ------- |:--------:|:-------:|
| 3.0.0 | 236 kB | 67.1 kB |
| 3.1.1 | 284 kB | 78.7 kB |
| 3.2.0 | 318 kB | 89.8 kB |
| 3.3.0 | 342 kB | 94.9 kB |
| 3.4.0 | 405 kB | 108 kB |
| 3.5.0 | 461 kB | 126 kB |

If we were to release axe-core v4.0 today (which is not feature complete), we would add ~160 kB (~30 kB gzipped) to the minified file size over v3.4.1 [which broke Lighthouses budget](https://github.com/GoogleChrome/lighthouse/pull/10056) when they tried to update to it. From v3.5.0, that's an increase of ~90 kb (~12 kB gzipped).

| Release | Minified | Gzipped |
| ------- |:--------:|:-------:|
| 3.4.0 | 405 kB | 108 kB |
| 3.5.0 | 461 kB | 126 kB |
| 4.0.0 | 553 kB | 138 kB |

Once v4.0 is feature complete, we will probably add another another 5 kB gzipped (just a guess). What this means is we've *doubled* the size of axe-core since releasing the last major release v3.0.

We should figure out some way to at least slow our rate of file size increase, if not figure out a max size we want the library to be.

WilcoFiers:
I've discussed this with @dylanb, he is on board with introducing the size budget. To address this, I propose we do the following:

1. Before we publish axe-core 4.1, we'll define a size budget for axe-core
1. Every major / minor release, starting with axe-core 4.1 we will reduce the size of axe-core by at least 5kb gzipped since its last major/minor release, until it reaches the size budget.
1. We will put tests in place which will enforce these requirements are met

To reach the size budget, we **may** decide to create a new .js file from which deprecated features and polyfills are removed. This will be a one-time thing, features deprecated after this file is introduced will be included to avoid breaking changes.

We are looking for input on what would be a preferable maximum GZipped file size for axe-core.

straker:
I propose we should aim for axe-core to be no larger than 100kB gippzed.

straker:
Note: I tried switching to terser and didn't see any improvement to our minified / gzip file size using the same options as uglify.

WilcoFiers:
[...]

PR #2527, opened 2020-09-21
feat(valid-langs): deprecate validLangs, add isValidLangs, reduce file size

straker:
Reduce the minified file size of the `valid-langs.js` file by 17 kB gzipped. I did this by transforming the data into a trie and stored as a nested array of integers. The comment in the file explains the technique and gives the code in case we need to update the list of valid langs.

Deprecated the `valdLangs` function as there's no good reason to need the entire list when just knowing if the string itself is valid. With a trie, it's O(n) (where `n` is the length of the string [2-3]) to know if the string is valid whereas generating the list of valid langs and then seeing if the string is included in that list is O(2m) (where `m` is the number of all valid langs [8000])

### Before axe-core size:

__valid-lang.js:__ 47.87 kB (gzipped: 19.75 kB)
__axe.min.js:__ 582 kB (gzipped: 144 kB)

### After axe-core size:

__valid-lang.js:__ 23.27 kB (gzipped: 2.72 kB) 
__axe.min.js:__ 558 kB (gzipped: 124 kB)

Part of issue: #2357 

## Reviewer checks


stephenmathieson (commented):
This is amazingly clever. Super good work, Steve! The drop in file size is great.

However, I wonder if it's "too clever" in its current state. For example, how do we add a new language? How do we remove one? Should we write little "utility programs" for managing the list of langs?

straker:
I thought about an automated script, but then I figured it was not needed. The reason being is that the valid list of ISO 639-1/2 codes should not change (or change very rarely), so we shouldn't have a need to update the list. However, I did leave the entire code at the bottom in a comment so if we do need to update/remove an item we can.

The other option is I take this out and make it it's own npm module that axe-core imports :)

stephenmathieson:
> The reason being is that the valid list of ISO 639-1/2 codes should not change (or change very rarely), so we shouldn't have a need to update the list.

I see. Probably not a valid concern then 😄 

> However, I did leave the entire code at the bottom in a comment so if we do need to update/remove an item we can.

Good point. This is probably enough for now, as long as another maintainer is comfortable updating the list based on your comments.

> The other option is I take this out and make it it's own npm module that axe-core imports :)

I was going to suggest this, but figured you wouldn't like that. If you think the list of languages would be helpful outside of axe-core, then IMO it'd be great to have it be its own package.

lib/core/utils/valid-langs.js
```diff
@@ -1,4 +1,71 @@
...
+
+	// padEnd is not supported in IE11
+	while (lang.length < 3) {
+		lang += '`';
+	}
+	const chars = lang.split('');
+
+	for (let i = 0; i <= chars.length - 1; i++) {
+		const index = chars[i].charCodeAt(0) - 96;
+		array = array[index];
```
WilcoFiers, line 31: Seems like we could just do this:

```suggestion
	for (let i = 0; i <= lang.length - 1; i++) {
		const index = lang.charCodeAt(i) - 96;
		array = array[index];
```

lib/core/utils/valid-langs.js
```diff
@@ -1,4 +1,71 @@
...
+		}
+	});
+
+	return codes;
+}
+
+export default isValidLang;
+
+/*
+If we need to edit the list of langs, use the code below and edit the array of langs:
```
WilcoFiers, line 67: I think we should get rid of the actual list and just point to the registry:
https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry

You may want to rerun this on the latest version of the registry, so that we have an up to date version. We haven't updated this list in several years.

WilcoFiers (changes requested):

dylanb:
In reading the code, I see that this does not affect the ability to pass in a custom list of valid languages - great!!

Nice work!

lib/core/utils/valid-langs.js
```diff
@@ -1,8153 +1,105 @@
...
+ */
+function isValidLang(lang) {
+	let array = langs;
+
+	// padEnd is not supported in IE11
+	while (lang.length < 3) {
+		lang += '`';
+	}
+	for (let i = 0; i <= lang.length - 1; i++) {
+		const index = lang.charCodeAt(i) - 96;
```
WilcoFiers, line 71: Yesterday morning I didn't know about charCodeAt. Last night when working on a personal project, I actually needed it to deal with emoji chars. :) Funny how these things can go.

WilcoFiers (approved):

lib/core/utils/valid-langs.js
paulirish, line 71: Just curious.. what does the `96` here refer to?

--------- 

(awesome PR. so pumped for this.)

I had also found this regex-based approach last week: https://github.com/SafetyCulture/bcp47/blob/develop/src/index.js  but it's more of a "does it look roughly like a bcp47 code" as opposed to being specific.

straker: the 96 gets a 1 index on the letter 'a' and a 0 index on letter '`'

j-mendez:
I think the gzip bytes on the after are off. Was able to get the bundle to `436kb` on this pr https://github.com/dequelabs/axe-core/pull/4013.

PR #2532, opened 2020-09-25
cli: wait before verifying

straker:
Our verify-next-release script keeps failing because of a race condition between npm having it and when our verify script tries to pull it down. Fixing the condition by putting in a sleep command to make the verify script wait before trying to pull it down from npm.

## Reviewer checks

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -197,8 +197,6 @@ jobs:
     steps:
       - checkout
       - <<: *restore_dependency_cache_unix
-      # Verifying the release keeps failing due to race condition
-      - run: sleep 10
       - run: .circleci/verify-release.sh post
 
   # Verify canary released package has all required files
@@ -208,8 +206,6 @@ jobs:
     steps:
       - checkout
       - <<: *restore_dependency_cache_unix
-      # Verifying the release keeps failing due to race condition
-      - run: sleep 10
       - run: npm run next-release
       - run: .circleci/verify-release.sh post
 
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -197,6 +197,8 @@ jobs:
     steps:
       - checkout
       - <<: *restore_dependency_cache_unix
+      # Verifying the release keeps failing due to race condition
+      - run: sleep 10
```
stephenmathieson, line 201: There's still a race condition here, it's just a little more likely we won't hit the issue. I've seen npm take minutes before a package is made available.

TLDR; this may not work

straker: Lovely... any recommendations on how to actually fix this?

stephenmathieson: I don't have any _good_ recommendations. I think `sleep` is the best bet here. Maybe we just add retries to what you've already got? If the package hasn't been published after 10 tries to verify it, we can just give up and manually verify it.

stephenmathieson: I suppose we could do this:

```sh
VERSION="x.y.z"

function wait_for_publish() {
  npm install "axe-core@$VERSION" < /dev/null
  if [ $? -eq 0 ]; then
    echo "Successfully installed axe"
    return
  else
    sleep 10
    wait_for_publish
  fi
}

wait_for_publish
normal_verify_script # call the existing tool to verify
```

WilcoFiers (approved):

PR #2542, opened 2020-10-06
feat(imports): deprecate axios

straker:
Deprecate our use of axios for a single request, which when we can remove it will reduce our minified and gzipped file size by 5 kB. This will also help our transfer to Rollup as axios caused some issues when trying to use the node resolve rollup plugin.

Part of #2357 

## Reviewer checks


stephenmathieson (commented):
LGTM!

I always thought `axios` was overkill here, and figured replacing it with [`unfetch`](https://www.npmjs.com/package/unfetch) would be an easy performance win. Using the native `XMLHttpRequest` makes sense too tho :)

While I'm hugely on-board with this change, I think it's best to wait for review from @WilcoFiers or @dylanb, as it involves changing the API.

WilcoFiers (approved):

Issue #2107, opened 2020-03-23
Add roles added in ARIA 1.2


WilcoFiers:
WAi-ARIA 1.2 has added a few new roles. Those should be added to axe-core. These roles aren't likely to effect accessibility when they're not supported so I think it's safe to add these right away.

straker:
See [the ariaRole standards docs](https://github.com/dequelabs/axe-core/blob/develop/doc/standards-object.md) for how to do this

straker:
New roles for ARIA 1.2

* [blockquote](https://www.w3.org/TR/wai-aria-1.2/#blockquote)
* [caption](https://www.w3.org/TR/wai-aria-1.2/#caption)
* [code](https://www.w3.org/TR/wai-aria-1.2/#code)
* [deletion](https://www.w3.org/TR/wai-aria-1.2/#deletion)
* [emphasis](https://www.w3.org/TR/wai-aria-1.2/#emphasis)
* [insertion](https://www.w3.org/TR/wai-aria-1.2/#insertion)
* [meter](https://www.w3.org/TR/wai-aria-1.2/#meter)
* [paragraph](https://www.w3.org/TR/wai-aria-1.2/#paragraph)
* [strong](https://www.w3.org/TR/wai-aria-1.2/#strong)
* [subscript](https://www.w3.org/TR/wai-aria-1.2/#subscript)
* [superscript](https://www.w3.org/TR/wai-aria-1.2/#superscript)
* [time](https://www.w3.org/TR/wai-aria-1.2/#time)

timmybytes:
Hey there! I'm trying to gain more experience with accessibility, and I'd love to work on this issue!

straker:
All yours. Let me know if you need anything.

straker:
PS. I find running the following script while the WCAG 1.2 spec is open helps seeing which ARIA attributes to add to the `requiredAttrs` and `allowedAttrs` properties by hiding any global ones:

```js
const globalAttrs = Array.from(
    document.querySelectorAll('#global_states li')
  ).map(li => li.textContent.replace(/\s*\(.*\)/, ''));

  const globalRoleAttrs = Array.from(
    document.querySelectorAll('.role-inherited li, .role-properties li')
  ).filter(li => globalAttrs.includes(li.textContent.replace(/\s*\(.*\)/, '')))

  globalRoleAttrs.forEach(li => li.style.display = 'none');
```

straker:
dev task

PR #2544, opened 2020-10-07
fix(aria-roles): Add WAI-ARIA 1.2 roles

timmybytes:
Adds new roles from WAI-ARIA 1.2 to lib/standards/aria-roles.js

Closes issue: #2107 

## Reviewer checks

- [x] Code is reviewed for security by @WilcoFiers


timmybytes:
Hey there. This is my first time working with a codebase of this size and complexity. I followed the contributing guidelines correctly (I think), only altered the single `aria-roles.js` file, but run into the following error:
```bash
  1 failing

  1) standards.getAriaRolesByType
       should return a list of role names by type:

      AssertionError: expected [ Array(37) ] to deeply equal [ Array(25) ]
      + expected - actual

       [
         "article"
      -  "blockquote"
      -  "caption"
         "cell"
      -  "code"
         "columnheader"
         "definition"
      -  "deletion"
         "directory"
         "document"
      -  "emphasis"
         "feed"
         "figure"
         "group"
         "heading"
         "img"
      -  "insertion"
         "list"
         "listitem"
         "math"
      -  "meter"
         "none"
         "note"
      -  "paragraph"
         "presentation"
         "row"
         "rowgroup"
         "rowheader"
         "separator"
      -  "strong"
      -  "subscript"
      -  "superscript"
         "table"
         "term"
      -  "time"
         "toolbar"
         "tooltip"
       ]
      
      at Context.<anonymous> (standards/get-aria-roles-by-type.js:15:10)
```

I've gone through the get-aria-roles-by-type.js file, which imports from lib/standards/index.js, but I'm at a bit of a loss as how to proceed. It seems like these new roles aren't being added in somewhere else for comparison checks, but like I said I'm a bit green with a project this complex. Could you point me in the right direction to get this fixed for you all?

straker:
Sure. The [test that is failing])https://github.com/dequelabs/axe-core/blob/develop/test/commons/standards/get-aria-roles-by-type.js#L15-L41) takes the roles directly from the standards object. So the fix should just be to add all the roles (looks like you added them all as structure roles) to the list from that test and it should work.

lib/standards/aria-roles.js
```diff
@@ -284,6 +303,10 @@ const ariaRoles = {
 		],
 		nameFromContent: true
 	},
+	meter: {
+		type: 'structure',
+		requiredAttrs: ['aria-valuemax', 'aria-valuemin', 'aria-valuenow']
```
straker, line 308: We'll need to add `aria-valuetext` in the `allowedAttrs` since we don't have inherited attrs.

straker (changes requested):
Awesome work. Just need to add 1 attribute and should be good to go after the tests are fixed.

lib/standards/aria-roles.js
timmybytes: Okay, I think I got it fixed.

straker (commented):
LGTM. But I want @WilcoFiers to look it over really quick make sure this is what he intended.

timmybytes:
Absolutely, no worries. Thanks so much for the help!

lib/standards/aria-roles.js
```diff
@@ -39,11 +39,18 @@ const ariaRoles = {
...
+		type: 'structure'
+	},
 	button: {
 		type: 'widget',
 		allowedAttrs: ['aria-expanded', 'aria-pressed'],
 		nameFromContent: true
 	},
+	caption: {
+		type: 'structure',
+		requiredContext: ['figure', 'table', 'grid']
```
WilcoFiers, line 52:
```suggestion
		requiredContext: ['figure', 'table', 'grid', 'treegrid']
```

WilcoFiers (changes requested):

WilcoFiers (approved):

WilcoFiers:
@timmybytes Thank you for the pull request! Much appreciated.

PR #2547, opened 2020-10-08
fix(implilcit-role): use type property instead of attribute to resolve input role

### LINKED ISSUE #2514: axe should be aware of input type default of text
regarding input type...

> If this attribute is not specified, the default type adopted is text.

_from https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input_

**example markup**: 

```html
<input ng-repeat="input in inputCtrl" ng-style="{'width': input.css + '%'}" class="input-md _1DEW order-number landing-input-1 sidebar-input-number ng-touched" ng-class="{ 'input-error': $ctrl.showError }" type="input.type" aria-label="Order Number" aria-describedby="loginErrorTooltip" data-test="input-number" placeholder="" ng-required="input.is_required" ng-model="$ctrl.modelObject[input.name]" ng-change="$ctrl.updateLandingForm(input.name, $ctrl.modelObject[input.name])" style="width: 98%;">
```
(from https://returns.narvar.com/reionline/returns)

Expectation: `getRole(document.querySelector('input'))` should return `textbox`

Actual: `getRole(document.querySelector('input'))` returns `null`

I can tell this is because the above input, has an invalid type of `input.type` 🤦 but it appears that all browsers treat that invalid `type` attribute value as the same as not having it present at all (which defaults to text). I've tested this [rei page](https://returns.narvar.com/reionline/returns) out with safari + vo and inspected a11y trees in the 3 major browsers.

Motivation: 🤷 

<pre><code>
axe-core version:4.0.2

For Tooling issues:
- Node version: XX  
- Platform:  
</code></pre>

shankarshastri:
Closes issue: #2514 (axe should be aware of input type default of text )

## Reviewer checks

- [X] Code is reviewed for security

### DIFF
--- a/lib/commons/standards/implicit-html-roles.js
+++ b/lib/commons/standards/implicit-html-roles.js
@@ -109,7 +109,7 @@ const implicitHtmlRoles = {
 				listElement && listElement.nodeName.toLowerCase() === 'datalist';
 		}
 
-		switch ((vNode.attr('type') || '').toLowerCase()) {
+		switch ((vNode.props.type || '').toLowerCase()) {
 			case 'button':
 			case 'image':
 			case 'reset':
--- a/lib/commons/standards/implicit-html-roles.js
+++ b/lib/commons/standards/implicit-html-roles.js
@@ -109,7 +109,7 @@ const implicitHtmlRoles = {
 				listElement && listElement.nodeName.toLowerCase() === 'datalist';
 		}
 
-		switch ((vNode.props.type || '').toLowerCase()) {
+		switch (vNode.props.type) {
 			case 'button':
 			case 'image':
 			case 'reset':
--- a/test/commons/aria/implicit-role.js
+++ b/test/commons/aria/implicit-role.js
@@ -301,6 +301,13 @@ describe('aria.implicitRole', function() {
 		assert.equal(implicitRole(node), 'combobox');
 	});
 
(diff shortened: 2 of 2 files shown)


shankarshastri:
@straker , Please review this.

lib/commons/standards/implicit-html-roles.js
```diff
@@ -109,7 +109,7 @@ const implicitHtmlRoles = {
 				listElement && listElement.nodeName.toLowerCase() === 'datalist';
 		}
 
-		switch ((vNode.attr('type') || '').toLowerCase()) {
+		switch ((vNode.props.type || '').toLowerCase()) {
```
WilcoFiers, line 112: props.type is already converted to the appropriate case.

```suggestion
		switch (vNode.props.type) {
```

WilcoFiers (changes requested):
Minor change, also can you please update the issue title to describe the fix?

lib/commons/standards/implicit-html-roles.js
shankarshastri: Thanks @WilcoFiers.

WilcoFiers (approved):
Reviewed for security.

WilcoFiers:
@shankarshastri Thank you for the pull request!

PR #2548, opened 2020-10-09
feat(new-rule): aria-tooltip-name, check that tooltips have an accessible name

WilcoFiers:
Part of the rule batch listed in #2421

## Reviewer checks


lib/rules/aria-form-field-name-matches.js
```diff
@@ -9,11 +9,19 @@ function ariaFormFieldNameMatches(node, virtualNode) {
 	 */
 	const nodeName = virtualNode.props.nodeName;
 	const role = getExplicitRole(virtualNode);
+	const hasHrefAttr = !!virtualNode.attr('href');
+
+	/**
+	 * Ignore elements from rule -> 'link-name'
+	 */
+	if (nodeName === 'a' && hasHrefAttr) {
```
straker, line 17: I presume this is to ignore anchor elements with the `tooltip` attribute? In that case it would be better to ignore it in the selector of the rule and not in the matches: `[role="\tooltip\"]:not(a)` (since anchor is not really a form field element)

straker (changes requested):

lib/rules/aria-form-field-name-matches.js
WilcoFiers: The reason I did this is because we're going to have half a dozen more rules where we have to do this. This matcher has a poor name, but it's purpose is to ignore elements that we natively check have an accessible name.

straker (approved):

Issue #1792, opened 2019-08-29
Add option to aria.getRole to ignore presentation and none roles


straker:
We do a lot of `!['presentation', 'none'].includes(role)` in the code. It would be nice to add an option to [aria.getRole](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/aria/get-role.js) to ignore these roles. An option named `noPresentational` should suffice. If the option is set, the function should return `null` if the role is either `presentation` or `none`.

Please also [create a test](https://github.com/dequelabs/axe-core/blob/develop/test/commons/aria/get-role.js) to show that the new option works.

Loganras:
If it's not already being worked on, I think I'll give this one a shot.

straker:
All yours. Let me know if you need anything.

PR #2549, opened 2020-10-09
feat(get-role): add noPresentational option

Loganras:
Add option to getRole to return null if 'presentation' or 'none'
would be returned

Closes issue #1792

<< Describe the changes >>

Closes issue:

## Reviewer checks

### DIFF
--- a/lib/commons/aria/get-role.js
+++ b/lib/commons/aria/get-role.js
@@ -110,22 +110,15 @@ function hasConflictResolution(vNode) {
 }
 
 /**
- * Return the semantic role of an element.
  *
- * @method getRole
- * @memberof axe.commons.aria
- * @instance
+ * @method resolveRole
  * @param {Element|VirtualNode} node
  * @param {Object} options
- * @param {boolean} options.noImplicit  Do not return the implicit role // @deprecated
- * @param {boolean} options.fallback  Allow fallback roles
- * @param {boolean} options.abstracts  Allow role to be abstract
- * @param {boolean} options.dpub  Allow role to be any (valid) doc-* roles
+ * @see getRole for option details
  * @returns {string|null} Role or null
- *
  * @deprecated noImplicit option is deprecated. Use aria.getExplicitRole instead.
  */
-function getRole(node, { noImplicit, ...explicitRoleOptions } = {}) {
+function resolveRole(node, { noImplicit, ...explicitRoleOptions } = {}) {
 	const vNode =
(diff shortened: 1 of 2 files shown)


straker:
@WilcoFiers you'll need to review this as I consulted on it.

lib/commons/aria/get-role.js
```diff
@@ -152,4 +145,31 @@ function getRole(node, { noImplicit, ...explicitRoleOptions } = {}) {
...
+ * @returns {string|null} Role or null
+ *
+ * @deprecated noImplicit option is deprecated. Use aria.getExplicitRole instead.
+ */
+function getRole(node, { noPresentational, ...options } = {}) {
+	const role = resolveRole(node, options);
+
+	if (noPresentational && ['presentation', 'none'].includes(role)) {
+		return null;
+	}
```
WilcoFiers, line 170: I think this would be better done as an "else if" here https://github.com/dequelabs/axe-core/pull/2549/files#diff-1bf6cc2da8acf89f4e8559d8254862d7L143 instead of in a separate function. WDYT?

WilcoFiers (changes requested):

lib/commons/aria/get-role.js
Loganras: Just to be sure I understand the suggestion clearly, I'll explain the reasoning behind my decisions.

I created the resolveRole(...) function from the understanding that the new option needs to handle _whatever role gets returned from getRole(...)_.  Since the original getRole(...) function had so many early returns, this was a great way for me to catch "whatever gets returned" as the role, and then apply the new option's behavior.

Let me know if that helps clarify why I did what I did.  Thank you @WilcoFiers for your review

Loganras:
@WilcoFiers , have you had a chance to review the comment I left earlier on your requested changes?  I look forward to hearing from you with any suggestions you may have.

lib/commons/aria/get-role.js
WilcoFiers: @Loganras You are absolutely right. @straker reminded me that implicitRole can return presentational too. My suggestion doesn't work. Let's go with this.

WilcoFiers (approved):

WilcoFiers:
@Loganras Thank you for the contribution! Much appreciated.

Loganras:
Thank you @WilcoFiers and @straker , it was my pleasure to contribute!

Issue #2460, opened 2020-08-16
aria-valid-attr-value mark missing aria-errormessage for review


WilcoFiers:
Instead of failing the the aria-valid-attr-value rule, the following should be marked for review. This should happen in the aria-errormessage check.

```html
<div role="textbox" aria-errormessage="my-error" aria-label="A textbox"></div>
```

Essentially, what needs to change is that we check for aria-errormessage if the target of the attribute exists in the DOM, and if it does not, return undefined to mark it for review, and add an incomplete message to the JSON file.

imccausl:
I'd like to give this one a shot 🙂

straker:
Sounds good. Let me know if you need anything.

padmavemulapati:
Verified with the latest axe-core develop code (dated 20th Oct 2020),
With the test file `<div role="textbox" aria-errormessage="my-error" aria-label="A textbox"></div> `
Seeing the `aria-errormessage` check result as  " ARIA attribute name is valid"

![image](https://user-images.githubusercontent.com/47409579/96597060-01c22200-130b-11eb-949a-16f4b81ab889.png)

Also verified with the snippet:
`<label for="startTime"> Please enter a start time for the meeting: </label>`
`<input id="startTime" type="text" aria-errormessage="msgID" value="" aria-invalid="false">`
`<span id="msgID" aria-live="off" style="visibility:hidden"> Invalid time:  the time must be between 9:00 AM and 5:00 PM" </span>`
and seeing the `aria-errormessage` check result as "aria-errormessage value `msgID` must use a technique to announce the message (e.g., aria-live, aria-describedby, role=alert, etc.)"
![image](https://user-images.githubusercontent.com/47409579/96596480-60d36700-130a-11eb-8a57-e007ff75e34c.png)

patrickhulce:
Hello fine axe folks! 👋 

This recently came up in Lighthouse (https://github.com/GoogleChrome/lighthouse/issues/12379), and I'm wondering if you could provide a bit more context for why invalid IDs aren't violations anymore (especially when they *also* fail the criteria for when the ID *is* valid). Lighthouse doesn't surface any manual review items, and naively it seems like a failing element when the ID does match should still be failing when the ID does not match. Is that not the case?

dylanb:
[...]

PR #2550, opened 2020-10-11
fix(checks/aria): mark elements missing from aria-errormessage for review

imccausl:
If IDs in aria-errormessage do not correspond to existing elements, return undefined to mark for review.

Closes issue #2460

## Reviewer checks

### DIFF
--- a/lib/checks/aria/aria-errormessage-evaluate.js
+++ b/lib/checks/aria/aria-errormessage-evaluate.js
@@ -3,7 +3,7 @@ import { getRootNode } from '../../commons/dom';
 import { tokenList } from '../../core/utils';
 
 /**
- * Check if an element with `aria-errormessage` also uses a technique to announce the message (aria-live, aria-describedby, etc.).
+ * Check if `aria-errormessage` references an element that also uses a technique to announce the message (aria-live, aria-describedby, etc.).
  *
  * ##### Data:
  * <table class="props">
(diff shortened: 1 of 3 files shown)


lib/checks/aria/aria-errormessage-evaluate.js
```diff
@@ -45,14 +45,13 @@ function ariaErrormessageEvaluate(node, options) {
 					-1
```
imccausl, line 45: I think this is most likely a separate issue (I didn't make any change because I felt it was out of scope, and I wasn't 100% sure about it), but right now it looks like the `ariaErrormessageEvaluate` check will fail in a valid case such as this (from https://www.w3.org/TR/2017/REC-wai-aria-1.1-20171214/#aria-errormessage)
```html
<label for="startTime"> Please enter a start time for the meeting: </label>
<input id="startTime" type="text" aria-errormessage="msgID" value="" aria-invalid="false">
<span id="msgID" aria-live="off" style="visibility:hidden"> Invalid time:  the time must be between 9:00 AM and 5:00 PM" </span>
```
where `aria-live` is `off` because `aria-invalid` is `false`. Maybe this check should check for the existence of `aria-invalid="true"` before getting strict about announcers?

Sidenote, w3 also seems to suggest that the announcer is optional on the element referenced by `aria-errormessage`:
> Authors MAY use live regions for the error message element applying either an aria-live property or using one of the live region roles, for example, alert. A live region scenario is when an error message is displayed to users only after they have provided invalid information. 

Just curious about your thoughts regarding this 🙂

WilcoFiers: That's a totally valid point. Would you mind opening an issue for this?

WilcoFiers (approved):
Looks good to me. I'll let @straker have the final word.

lib/checks/aria/aria-errormessage-evaluate.js
imccausl: Absolutely! I wouldn't mind working on it as well.

PR #2571, opened 2020-10-20
feat(new-rule): ARIA links, buttons, menuitems have an accessible name

WilcoFiers:
Move ARIA link and button testing to this new aria-command-name rule, out of link-name / button-name. This closes issue #2582 and #2577.

Requires #2584 to be merged.

Part of #2421

## Reviewer checks


test/integration/virtual-rules/index.html
```diff
@@ -21,21 +21,24 @@
 	</head>
 	<body>
 		<div id="mocha"></div>
+		<script src="area-alt.js"></script>
+		<script src="aria-input-field-name.js"></script>
+		<script src="aria-menuitem-name.js"></script>
+		<script src="aria-progressbar-name.js"></script>
+		<script src="aria-toggle-field-name.js"></script>
+		<script src="aria-tooltip-name.js"></script>
```
WilcoFiers, line 29: We need to think of a way to make sure we don't forget to do this. Any thoughts?

straker: We could generate the file like we do the other test html files. I've not looked into that part of the build but it should be possible.

straker (approved):

build/tasks/validate.js
```diff
@@ -199,7 +199,9 @@ function createSchemas() {
 				items: {
 					type: 'string'
 				},
-				conform: function hasCategoryTag(tags) { return tags.some(tag => tag.includes('cat.')) },
+				conform: function hasCategoryTag(tags) {
+					return tags.some(tag => tag.includes('cat.'));
+				},
```
WilcoFiers, line 204: Hmm, not sure how that didn't make it in last PR. Shouldn't matter.

straker (approved):

PR #2579, opened 2020-10-21
chore: convert to esbuild

straker:
[Esbuild](https://github.com/evanw/esbuild) is a "newer" javascript bundler that I've been interested in trying out. It runs super fast (benchmarks on their repo) and looking at the output does Rollup style resolving where it concats the functions by use instead of using webpack style resolve (resulting in cleaner and smaller output). Even better, esbuild works out of the box to resolve node_module style webpack paths and doesn't have the same problem I was facing with Rollup and having to add a few plugins for that (and still ran into issues with Axios).

Doing this resulted in a filesize decrease of 15 kB gzipped.

## Webpack stats: 

| File | Minified | Gzip |
|-|-|-|
| axe.js | 1.29 MB | 188 kB |
| axe.min.js | 561 kB | 125 kB |

## Esbuild stats

| File | Minified | Gzip |
|-|-|-|
| axe.js | 940 kB | 164 kB |
| axe.min.js | 421 kB | 110 kB |

Part of issue: #2357 

## Reviewer checks

### DIFF
--- a/build/tasks/esbuild.js
+++ b/build/tasks/esbuild.js
@@ -26,7 +26,7 @@ module.exports = function(grunt) {
 						bundle: true
 					})
 						.then(done)
-						.catch(() => process.exit(1));
+						.catch(done);
 				});
 			});
 		}
--- a/build/tasks/esbuild.js
+++ b/build/tasks/esbuild.js
@@ -6,6 +6,7 @@ module.exports = function(grunt) {
 		'esbuild',
 		'Task to run the esbuild javascript bundler',
 		function() {
+			const done = this.async();
 			const files = grunt.task.current.data.files;
 
 			files.forEach(file => {
@@ -23,7 +24,9 @@ module.exports = function(grunt) {
 						outfile: path.join(dest, name),
 						minify: false,
 						bundle: true
-					}).catch(() => process.exit(1));
+					})
+						.then(done)
+						.catch(() => process.exit(1));
 				});
 			});
 		}
--- a/build/tasks/esbuild.js
+++ b/build/tasks/esbuild.js
@@ -1,0 +1,31 @@
+const { build } = require('esbuild');
+const path = require('path');
+
+module.exports = function(grunt) {
+	grunt.registerMultiTask(
+		'esbuild',
+		'Task to run the esbuild javascript bundler',
+		function() {
+			const files = grunt.task.current.data.files;
+
+			files.forEach(file => {
+				const src = Array.isArray(file.src) ? file.src : [file.src];
+				const dest = file.dest;
+
+				src.forEach(entry => {
+					const name = path.basename(entry);
+					if (file.cwd) {
+						entry = path.join(file.cwd, entry);
+					}
+
+					build({
+						entryPoints: [entry],
+						outfile: path.join(dest, name),
+						minify: false,
+						bundle: true
+					}).catch(() => process.exit(1));
+				});
+			});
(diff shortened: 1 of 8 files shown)


build/tasks/esbuild.js
```diff
@@ -0,0 +1,31 @@
...
+					if (file.cwd) {
+						entry = path.join(file.cwd, entry);
+					}
+
+					build({
+						entryPoints: [entry],
+						outfile: path.join(dest, name),
+						minify: false,
+						bundle: true
+					}).catch(() => process.exit(1));
```
stephenmathieson, line 26: IMO we should output an error here, not just fail silently.

stephenmathieson: If Grunt handles this and outputs a useful error, then ignore this. I don't know how the Grunt architecture works.

straker: I think esbuild does that for us? Their docs just exit https://github.com/evanw/esbuild/blob/master/docs/js-api.md#node-api. I had an issue with the config at one point and they messaged the error, so I assume a problem with the build will as well?

stephenmathieson: That looks like example code, not necessarily what should be used in an actual build process.

Have you seen what happens when an exception is raised here?

straker: Looks like it outputs the error:

image

stephenmathieson: Ah, perfect! Thanks for looking into this.

build/tasks/esbuild.js
```diff
@@ -23,7 +24,9 @@ module.exports = function(grunt) {
 						outfile: path.join(dest, name),
 						minify: false,
 						bundle: true
-					}).catch(() => process.exit(1));
+					})
+						.then(done)
+						.catch(() => process.exit(1));
```
stephenmathieson, line 29: [Based on the Grunt docs](https://gruntjs.com/creating-tasks#why-doesn-t-my-asynchronous-task-complete), it looks like we should pass `false` to `done` here:

```suggestion
						.catch(() => done(false));
```

Don't know how much this matters tho, as you pointed out ESBuild itself emits a "nice" error message prior to exiting the build.

stephenmathieson (commented):
LGTM, but should get approval from Wilco or Dylan on this one.

build/tasks/esbuild.js
WilcoFiers: Yeah, better to just let Grunt do the work.

jeeyyy:
Nice one @straker

build/tasks/esbuild.js
```diff
@@ -0,0 +1,34 @@
...
+					}
+
+					build({
+						entryPoints: [entry],
+						outfile: path.join(dest, name),
+						minify: false,
+						bundle: true
+					})
+						.then(done)
+						.catch(done);
```
stephenmathieson, line 29: Should be calling `done(false)` here according to the Grunt docs

straker: [Error also is valid](https://gruntjs.com/inside-tasks):

> Either false or an Error object may be passed to the done function to instruct Grunt that the task has failed.

WilcoFiers (approved):

build/tasks/esbuild.js
stephenmathieson: Ah cool, ok 👍

PR #2584, opened 2020-10-22
feat: deprecate aria-form-field-name-matches for no-name-method-matches

WilcoFiers:
Rather than exclude some elements based on if we have another rule for them, we need to exclude elements based on if there's a name method for the node type. This is more robust.

Closes issue: #2585

## Reviewer checks


lib/core/base/metadata-function-map.js
```diff
@@ -129,7 +129,6 @@ import noAutoplayAudioEvaluate from '../../checks/media/no-autoplay-audio-evalua
 // rule matches
 import ariaAllowedAttrMatches from '../../rules/aria-allowed-attr-matches';
 import ariaAllowedRoleMatches from '../../rules/aria-allowed-role-matches';
-import ariaFormFieldNameMatches from '../../rules/aria-form-field-name-matches';
```
straker, line 132: This cannot be a `fix` commit and shouldn't be removed (this would be a breaking change).

Instead, this should be a `feat` commit and the prior `ariaFromFieldNameMatches` should be left as it and only deprecated.

straker (dismissed):
I think the idea works, but we can't remove code until 5.0.

straker (approved):

lib/core/base/metadata-function-map.js
straker: After discussing with Wilco: the `id` is left in the code so can still be used. The new function does everything the old function did but in a more efficient way, so it's not considered a breaking change. OK to remove the old code.

PR #2590, opened 2020-10-23
fix(aria-errormessage): allow aria-live="polite" on aria-errormessage target

nantunes:
Spec (https://www.w3.org/TR/wai-aria-1.1/#aria-errormessage) only states that "authors MAY use live regions for the error message element applying either an aria-live property or using one of the live region roles".

So we probably shouldn't enforce any role, unless we consider it a bad practice. In that case we should at least allow "polite" announcements, as "authors SHOULD NOT use the assertive value unless the interruption is imperative".

## Reviewer checks

### DIFF
--- a/lib/checks/aria/aria-errormessage-evaluate.js
+++ b/lib/checks/aria/aria-errormessage-evaluate.js
@@ -19,6 +19,7 @@ function ariaErrormessageEvaluate(node, options) {
 			return (
 				idref.getAttribute('role') === 'alert' ||
 				idref.getAttribute('aria-live') === 'assertive' ||
+				idref.getAttribute('aria-live') === 'polite' ||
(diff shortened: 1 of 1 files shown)


straker (approved):
Thanks or the pr. I'll bring up with the team about the role checking.

straker:
Sooooo. I just realized that this pr pointed to `master`. It should have been branched from `develop` and merged into `develop`. I'm going to have to revert this change since we use `master` as the current release target, do you mind opening it again and pr against the `develop` branch? Sorry for the inconvenience.

nantunes:
Sure, I'll do it. It is my fault, as I didn't notice `develop` was the default branch. But CONTRIBUTING.md needs some improvements and details.

PR #2597, opened 2020-10-27
fix(aria-errormessage): allow aria-live="polite" on aria-errormessage target

nantunes:
Spec (https://www.w3.org/TR/wai-aria-1.1/#aria-errormessage) only states that "authors MAY use live regions for the error message element applying either an aria-live property or using one of the live region roles".

So we probably shouldn't enforce any role, unless we consider it a bad practice. In that case we should at least allow "polite" announcements, as "authors SHOULD NOT use the assertive value unless the interruption is imperative".

Replaces PR #2590, that was wrongly opened targeting the master branch.

Closes issue:

## Reviewer checks


straker (approved):

straker:
Thanks again for doing this. I'll get the contributor file updated.

Issue #2330, opened 2020-06-26
iframe giving "All page content must be contained by landmarks" when iframe itself is contained by a landmark


jnurthen:
Expectation: This best practice should only occur when content isn't contained by a landmark

Actual: The best practice gets reported whenever an iframe contains content without landmarks even when the iframe itself is contained in a landmark.

Motivation: We use a UI shell around some pages. The shell will look at the included page and, if they do not provide landmarks, will add a main around the iframe. axe still reports this as an error and it is confusing to developers

Testcase: https://jnurthen.github.io/testcases/axe/iframeLandmarks/index.html

<pre><code>
axe-core version: 3.5.4
extension: 4.5.2

</code></pre>

WilcoFiers:
Yeah I agree, that should pass. Thanks for reporting James.

padmavemulapati:
Verified using Test File - on axe-coconut(v4.6.1.30800)
[landmark-unique-test.txt](https://github.com/dequelabs/axe-core/files/5494968/landmark-unique-test.txt)
Seeing the "region" rule for iframes:
![image](https://user-images.githubusercontent.com/47409579/98253658-55866980-1fa1-11eb-8fa5-edd313e46ae6.png)

PR #2612, opened 2020-10-29
fix(region): dont fail contents of iframe if iframe is in a region

straker:
Closes issue: #2330

## Reviewer checks

### DIFF
--- a/lib/checks/navigation/region-after.js
+++ b/lib/checks/navigation/region-after.js
@@ -1,0 +1,73 @@
+import cache from '../../core/base/cache';
+import { getNodeFromTree } from '../../core/utils';
+
+function regionAfter(results) {
+	const regionlessNodes = cache.get('regionlessNodes');
+	const frameResults = [];
+
+	// adjust iframe results based on if the iframe is within a
+	// region or if the iframe has a region inside of it.
+	// basically we need to perform the same logic in region-evaluate
+	// but across iframes
+	results.forEach(result => {
+		const vNode = getNodeFromTree(result.node._element);
+		const isIframe = vNode.props.nodeName === 'iframe';
+
+		// look at failing iframes after we process their results
+		if (isIframe && !result.node._fromFrame && !result.result) {
+			frameResults.push(result);
+		} else if (result.node._fromFrame) {
+			// if the iframe is inside of a region then all content
+			// within the iframe passes
+			if (!regionlessNodes.includes(vNode)) {
+				result.result = true;
+			}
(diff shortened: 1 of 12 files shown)


lib/checks/navigation/region-after.js
```diff
@@ -0,0 +1,73 @@
+import cache from '../../core/base/cache';
+import { getNodeFromTree } from '../../core/utils';
+
+function regionAfter(results) {
+	const regionlessNodes = cache.get('regionlessNodes');
```
WilcoFiers, line 5: I don't think we should use the cache as a way to pass information between the evaluate and after functions. That seems problematic. If we need info about a node, lets put it in `data`.

lib/checks/navigation/region-after.js
```diff
@@ -0,0 +1,73 @@
...
+					numPass: 0,
+					numFail: 0
+				};
+				const key = result.result ? 'numPass' : 'numFail';
+				vNode._regionResults[key]++;
+			}
+		}
+	});
+
+	frameResults.forEach(result => {
```
WilcoFiers, line 40: this seems fairly complex, wouldn't it be easier to always pass iframes, and if there's content that fails have whatever inside that frame is missing the region fail?

test/integration/full/region/region-pass.js
```diff
@@ -2,13 +2,15 @@ describe('region pass test', function() {
...
-			r
-		) {
-			assert.isNull(err);
-			results = r;
-			done();
+		axe.testUtils.awaitNestedLoad(function() {
+			axe.run({ runOnly: { type: 'rule', values: ['region'] } }, function(
+				err,
+				r
+			) {
```
WilcoFiers, line 9: This should work too:

```suggestion
			axe.run({ runOnly: ['region'] }, function(err, r) {
```

test/integration/full/region/region-fail.js
```diff
@@ -2,23 +2,36 @@ describe('region fail test', function() {
...
-			r
-		) {
-			assert.isNull(err);
-			results = r;
-			done();
+		axe.testUtils.awaitNestedLoad(function() {
+			axe.run({ runOnly: { type: 'rule', values: ['region'] } }, function(
+				err,
+				r
+			) {
```
WilcoFiers, line 9: This should work too:

```suggestion
			axe.run({ runOnly: ['region'] }, function(err, r) {
```

WilcoFiers (changes requested):

PR #2614, opened 2020-10-30
fix(region): treat iframes as regions

straker:
Treat iframes as regions. If content inside the iframe does not have a region, we'll get a report on just that content.

Closes issue: #2330 

## Reviewer checks


WilcoFiers (approved):
Wow, that did turn out to be much easier.

Issue #1461, opened 2019-03-26
link-name rule fails when it contains landmarks


WilcoFiers:
Found the following example which seemed to fail axe-core:

```html
<a href="...">
  <article><div>
    <div><img src="..." alt=""></div>
    <p class="card-description">Some card</p>
  </div></article>
</a>
```

As far as I can tell this link should and does not have an accessible name, but in a quick test with VO that doesn't seem to be a problem.

axe-core: 3.2.2

drewlee:
We've noticed this issue as well after upgrading to v3.2.2. Links which had previously passed are now getting flagged as violations. Are there any workarounds aside from excluding the element or rule from the audit?

WilcoFiers:
We've done a little more testing. Here are the results so far:

| AT + UA     | method    | announcement |
|-------------|-----------|--------------|
| VO + Sarari | Tabbing   | "Some card"  |
| VO + Sarari | Reading   | "Some card"  |
| VO + Sarari | Link list | "Some card"  |
| NVDA + FF   | Tabbing   | "Some card"  |
| NVDA + FF   | Reading   | "Some card"  |
| JAWS + IE   | Tabbing   | ""           |
| JAWS + IE   | Reading   | "Some card"  |


@straker can you do some more testing for this, complete this table? Tabbing to the link with JAWS does seem to cause problems, so that suggests there is at least some sort of issue here. I want to get a better sense of the problem before we decide whether or not we let axe fail this, warn about it, or let it pass.

WilcoFiers:
@drewlee What element are you using inside links that is causing this behaviour? I think the way to fix it is to not use that element.

drewlee:
@WilcoFiers, that's much easier said than done. The contention seems to be around our usage of a definition list, but unordered/ordered lists are failing as well. From a semantic perspective, our usage is completely justifiable and makes more sense than a hierarchy of divs. I'd also like to point out that in accordance with the HTML5 specification, it is perfectly acceptable to encompass complex semantics within anchors: [https://www.w3.org/TR/html52/textlevel-semantics.html#elementdef-a](). To paraphrase: "the `<a>` element may be wrapped around entire paragraphs, lists, tables, and so forth, even entire sections, so long as there is no interactive content within (e.g., buttons or other links)."

[...]

PR #2617, opened 2020-11-02
fix(link-name): pass landmark content as link text

WilcoFiers:
To figure this one out I tested the following code:

```html
<label for="bar" id="foo">
	<a href="/" id="baz">
		<div role="alert" aria-label="Goodbye">Hello</div>
		<div role="textbox" aria-label="mars">earth</div>
		<div role="img">people</div>
	</a>
</label>
<input type="button" aria-labelledby="foo" id="bar" />
```

The surprising thing is that AT gave different names than the browsers do. The button works fairly consistently, but most AT seem to have something in place to deal with this `a[href] > article:not([aria-label])` scenario. Here are my test results:

(either aria-labelledby or explicit label)
| Browser (+AT)  | Link                | Button
|----------------|---------------------|-----------------
| Chrome name    | Goodbye mars        | Goodbye mars
| NVDA + Chrome  | Hello mars          | Goodbye mars
| JAWS + Chrome  | Hello slash         | Goodbye mars
| Firefox name   | Goodbye mars earth  | Goodbye mars earth
| NVDA + Firefox | Hello mars          | Goodbye mars
| Safari name    | Goodbye mars people | Goodbye mars
| VO + Safari    | Goodbye earth mars  | Goodbye mars
| JAWS + IE      | Hello earth people  | Goodbye mars

Versions: Firefox 82, Chrome 86, Safari 14, IE11, VO (MacOS) Catalina, JAWS 2020, NVDA 2020.3

Results are completely all over the place, so to avoid false positives what I've done is chosen the most optimistic scenario, which is that when looking for name from content of an element that isn't a label, all descendants should be considered, even elements that are named from author, except if the element that is named from author has an accessible name already through some other means.

**This PR does not guarantee the outcome will work everywhere.** Instead it avoids the false positive. I'm opening a ticket to look at this more closely when we're not days away from a code freeze.

Closes issue: #1461

## Reviewer checks

### DIFF
--- a/lib/commons/text/subtree-text.js
+++ b/lib/commons/text/subtree-text.js
@@ -31,7 +31,7 @@ function subtreeText(virtualNode, context = {}) {
 	/**
 	 * Note: Strictly speaking if a child isn't named from content and it has no accessible name
 	 * accName says to ignore it. Browsers do this fairly consistently, but screen readers have
-	 * chosen to ignore this, but only for direct content, not for links.
+	 * chosen to ignore this, but only for direct content, not for labels / aria-labelledby.
 	 * That way in `a[href] > article > #text` the text is used for the accessible name,
 	 * See: https://github.com/dequelabs/axe-core/issues/1461
 	 */
--- a/lib/commons/text/subtree-text.js
+++ b/lib/commons/text/subtree-text.js
@@ -13,14 +13,32 @@ import getOwnedVirtual from '../aria/get-owned-virtual';
 function subtreeText(virtualNode, context = {}) {
 	const { alreadyProcessed } = accessibleTextVirtual;
 	context.startNode = context.startNode || virtualNode;
-	const { strict } = context;
+	const { strict, inControlContext, inLabelledByContext } = context;
 	if (
 		alreadyProcessed(virtualNode, context) ||
-		!namedFromContents(virtualNode, { strict })
+		virtualNode.props.nodeType !== 1
 	) {
 		return '';
 	}
 
+	if (
+		!namedFromContents(virtualNode, { strict }) &&
+		!context.subtreeDescendant
+	) {
+		return '';
+	}
+
+	/**
+	 * Note: Strictly speaking if a child isn't named from content and it has no accessible name
+	 * accName says to ignore it. Browsers do this fairly consistently, but screen readers have
+	 * chosen to ignore this, but only for direct content, not for links.
+	 * That way in `a[href] > article > #text` the text is used for the accessible name,
+	 * See: https://github.com/dequelabs/axe-core/issues/1461
+	 */
+	if (!strict) {
+		const subtreeDescendant = !inControlContext && !inLabelledByContext;
+		context = { subtreeDescendant, ...context };
+	}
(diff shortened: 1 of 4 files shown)


lib/commons/text/subtree-text.js
```diff
@@ -13,14 +13,32 @@ import getOwnedVirtual from '../aria/get-owned-virtual';
...
+		!namedFromContents(virtualNode, { strict }) &&
+		!context.subtreeDescendant
+	) {
+		return '';
+	}
+
+	/**
+	 * Note: Strictly speaking if a child isn't named from content and it has no accessible name
+	 * accName says to ignore it. Browsers do this fairly consistently, but screen readers have
+	 * chosen to ignore this, but only for direct content, not for links.
```
WilcoFiers, line 34:
```suggestion
	 * chosen to ignore this, but only for direct content, not for labels / aria-labelledby.
```

lib/commons/text/subtree-text.js
```diff
@@ -13,14 +13,32 @@ import getOwnedVirtual from '../aria/get-owned-virtual';
 function subtreeText(virtualNode, context = {}) {
 	const { alreadyProcessed } = accessibleTextVirtual;
 	context.startNode = context.startNode || virtualNode;
-	const { strict } = context;
+	const { strict, inControlContext, inLabelledByContext } = context;
 	if (
 		alreadyProcessed(virtualNode, context) ||
-		!namedFromContents(virtualNode, { strict })
+		virtualNode.props.nodeType !== 1
```
WilcoFiers, line 19: Even when subtreeDescendant is true, we need to return an empty string for things that aren't nodes.

test/commons/text/accessible-text.js
```diff
@@ -257,12 +257,12 @@ describe('text.accessibleTextVirtual', function() {
 		axe.testUtils.flatTreeSetup(fixture);
 
 		var target = axe.utils.querySelectorAll(axe._tree, '#t2label')[0];
-		// Chrome 72: This is This is a label of
-		// Firefox 62: This is ARIA Label
-		// Safari 12.0: This is This is a label of
+		// Chrome 86: This is This is a label of
```
straker, line 260: Are we marking what the browser's accessibility tree says or what screen readers with those browsers say?

WilcoFiers: the browser's name in the acc tree.

straker (changes requested):
We should add the example code you used as a test case for this change so we can make sure future changes affect it.

straker (approved):

goodwitch:
Hey @WilcoFiers is this really a common situation.  I was just chatting about this with Michael Fairchild...and he wonders how common that issue is? 

It seems like having a textbox wrapped by a link would be super rare (and also invalid if this were entirely HTML, but the mix of html/aria means it’s probably technically valid code and would pass 4.1.1)

What am I getting at? If screenreader or browser devs see this code reported in an issue, they might say “yeah, that’s not valid and such an edge case, I’ll put this at the lowest possible priority to fix”.

Issue #728, opened 2018-02-10
heading-order scope & related nodes bugs


WilcoFiers:
The heading-order rule makes the assumption that it's always used to test the entire page. That's incorrect, because the context object can have it test for instance only the header and footer. This can introduce bugs in two ways: 
1. An "in-between" heading is overlooked, creating a false positive (see below)
2. Issues can be overlooked when they cross the scope boundary.

Here's a use case:
```html
<!doctype html>
<html lang="en">
<title>O hai</title>
<header>
  <h1>Foo</h1>
</header>
<main>
  <h2>bar</h2>
</main>
<footer>
  <h3>Baz</h3>
</footer>

<script src="/axe.js"></script>
<script>
window.addEventListener('load' , function () {
  axe.run({ include: [['header'], ['footer']] }, {
      runOnly: { type: "rule", values: ["heading-order"] }
    }, function (err, res) {
      // Expect: Undefined
      // Actual: heading-order violation
      console.log(res.violations[0]);
    });
})
</script>
```

To solve this, instead of using an 'after' method, we should look for the *previous* heading on the page.

### Minor bug
I also noticed that the previous heading wasn't included as a related node. We should fix that.

marcysutton:
Good timing. I was about to make changes to this rule for Shadow DOM, but I hadn't fully looked into it yet. There are a number of checks in `navigation` that need to be updated–I've gotten up to `p-as-heading` so far. I'll take a look at this one next week.

marcysutton:
This appears to be more complicated than I first thought, and it warrants a discussion about how it should work.

WilcoFiers:
What's the concern Marcy?

dylanb:
Why not simply flag this as a page level rule?

WilcoFiers:
Page level rules can still have parts of the page scoped out. That doesn't solve the issue.

dylanb:
@WilcoFiers that is a bit of a conundrum...perhaps we should add an option for page level rules to be turned off on these scenarios? Seems like there is no really good behavior that works for all scenarios.

WilcoFiers:
I still don't see why we couldn't just have this rule crawl backward up the DOM and look for the previous visible heading. Seems like that would solve this pretty easily.

dylanb:
Because that does not work across iframe boundaries

padmavemulapati:
Verified with the latest axe-core-develop branch code (dated 10/12/2020) -
[...]

PR #2622, opened 2020-11-03
fix(heading-order): allow partial context to pass

straker:
These changes allow partial `heading-order` context to pass. It will also handle iframes if they are included in the context or not.

Closes issue: #728

## Reviewer checks


straker:
I have tried to debug this in IE11 for hours but can't figure out what's going on. It doesn't help that it's flakey for some reason. Best I can figure is that the iframe isn't running / returning axe results to the parent and so the iframe results are completely missing from the results object. Sometimes (rarely) the iframe times out and we get an error "Axe in frame timed out," but this doesn't happen every time the iframe results aren't in the results object.

A lot of the times when devtools are open the tests pass magically, or when I add console logs or debuggers. This makes me think it's a race condition, but I can't figure out where the problem is occurring in our integration `runner.js` script.

WilcoFiers (changes requested):

lib/checks/navigation/heading-order-evaluate.js
```diff
@@ -1,19 +1,50 @@
...
+import { isVisible } from '../../commons/dom';
+
+function getLevel(vNode) {
+	const ariaHeadingLevel = vNode.attr('aria-level');
+	const nodeName = vNode.props.nodeName;
 
 	if (ariaHeadingLevel !== null) {
-		this.data(parseInt(ariaHeadingLevel, 10));
-		return true;
+		return parseInt(ariaHeadingLevel, 10);
```
WilcoFiers, line 10: Should probably test the validity of this, so that `aria-level="-1"` doesn't cause axe to treat it as an iframe 👿

lib/checks/navigation/heading-order-evaluate.js
```diff
@@ -1,19 +1,50 @@
...
+	const selector = 'h1, h2, h3, h4, h5, h6, [role=heading], iframe, frame';
+	// TODO: es-modules_tree
+	const vNodes = querySelectorAllFilter(axe._tree[0], selector, vNode =>
+		isVisible(vNode.actualNode, true)
+	);
+
+	headingOrder = vNodes.map(vNode => {
+		// save the path so we can reconstruct the heading order
+		return {
+			ancestry: new DqElement(vNode.actualNode).ancestry,
```
WilcoFiers, line 41: Could you use `getAncestry()` here instead?

lib/checks/navigation/heading-order-after.js
```diff
@@ -1,15 +1,112 @@
...
+			};
+		});
+	});
+
+	// determine where all nested headingOrders fit into the
+	// top-level heading order. since we can encounter a nested
+	// iframe result before the parent result is processed, we'll have
+	// to do this loop a few times so can process the frame results in
+	// top-down order
+	while (nestedResults.length) {
```
WilcoFiers, line 64: I have no idea what is happening in this loop. Can we write this in a more descriptive way? Maybe put some stuff into more descriptive functions? This is not code I'd be able to maintain.

WilcoFiers (changes requested):

lib/checks/navigation/heading-order-after.js
```diff
@@ -1,15 +1,130 @@
...
+		}
+	}
+
+	// replace the ancestry path with information about the result
+	results.forEach(result => {
+		const path = result.node.ancestry.join(joinStr);
+		const heading = headingOrder.find(heading => {
+			return heading.ancestry === path;
+		});
+		const index = headingOrder.indexOf(heading);
```
WilcoFiers, line 106: Seems like this might work:

```suggestion
		const index = headingOrder.findIndex(heading => {
			return heading.ancestry === path;
		});
```

lib/checks/navigation/heading-order-after.js
```diff
@@ -1,15 +1,130 @@
...
+		headingOrder.splice(index, 1, ...result.data.headingOrder);
+	}
+
+	// replace each iframe in the top-level heading order with its
+	// results.
+	// since nested iframe results can appear before their parent
+	// iframe, we will just loop over the nested results and
+	// piece-meal replace each iframe in the top-level heading order
+	// with their results until we no longer have results to replace
+	while (nestedResults.length) {
```
WilcoFiers, line 84: We should probably test to make sure this doesn't get into an infinite loop. nestedResults.lenth is the same as it was last iteration, something went wrong. Probably best to throw an error and let axe-core's error handling deal with it instead of it crashing out completely.

WilcoFiers (changes requested):
Couple minor points

lib/checks/navigation/heading-order-after.js
straker: `findIndex` is not supported in IE11 and I didn't want to bring in a polyfill to use it

WilcoFiers: Fair enough!

WilcoFiers (approved):

PR #2623, opened 2020-11-03
docs: put aria-dialog-name in correct section

straker:
Not sure how, but the `aria-dialog-name` was added to the wrong section of the rule description file. It was added to the "WCAG 2.0 Level A & AA Rules" section instead of the "Best Practices Rules." Any builds would try to correct the issue, so making this its own pr to address it.

## Reviewer checks


WilcoFiers (approved):

dylanb:
Isn't this file supposed to be generated?

WilcoFiers:
@dylanb It is generated. I suspect the person who write the aria-dialog-name rule changed the tag in the github editor, and because of that the script that normally runs on commit didn't run.

Issue #2541, opened 2020-10-06
False positive: "Color-contrast" rule fails when using particular text-shadow values


iamrafan:
```color-contrast``` rule fails when an element has text-shadow values as mentioned in Case-1 in the example. 

Demo link: https://codepen.io/iamrafan/pen/abNevdV?editors=1100

<pre><code>axe-core version: 4.0.2
axe version: 4.6.1</code></pre>

padmavemulapati:
Verified with the latest axe-core develop branch code(v4.0.2 dated 12/11/2020)
with the code snippets: color-contrast - for text-shadows is passing for the below three cases.
1. `<div id="target" style="background-color: #666; color:#aaa; 
				text-shadow: 0 0 0.09em #000, 0 0 0.09em #000, 0 0 0.09em #000;">
				Hello world
				</div>`
2. `<div id="target" style="background-color: #aaa; color:#666; 
				text-shadow: 0 0 0.09em #000, 0 0 0.09em #000, 0 0 0.09em #000;">
				  Hello world
				</div>`
3.`<div id="target" style="background-color: #aaa; color:#666; 
				text-shadow: 0 0 0.09em #000, 0 0 0.09em #000, 0 0 0.09em #000;">
				Hello world
				</div>`
![image](https://user-images.githubusercontent.com/47409579/98974069-7ff49b80-253a-11eb-8ef0-c613bce12945.png)


and failing for the following two cases:
1.`<div id="parent" style="height: 40px; width: 30px; background-color: #000;">
					<div id="target" style="height: 20px; width: 15px; text-shadow: red 0 0 0.05em">foo
					</div></div>`
2.`<div style="height: 40px; width: 30px; background-color: #800000;">
						<div id="target" style="height: 20px; width: 15px; text-shadow: red 0 0 1em, green 0 0 0.5em">foo
						</div></div>`

![image](https://user-images.githubusercontent.com/47409579/98974135-913da800-253a-11eb-98d6-6d1b3196b7a0.png)

PR #2627, opened 2020-11-07
fix(color-contrast): allow small text shadows to serve as text outline

WilcoFiers:
Closes issue: #2541

## Reviewer checks


lib/commons/color/get-text-shadow-colors.js
```diff
@@ -76,23 +98,24 @@ function parseTextShadows(textShadow) {
...
 	return shadowColor;
 }
 
-function blurRadiusToAlpha(blurRadius) {
+function blurRadiusToAlpha(blurRadius, fontSize) {
 	// This formula is an estimate based on various tests.
 	// Different people test this differently, so opinions may vary.
-	return 3.7 / (blurRadius + 8);
+	const relativeBlur = blurRadius / fontSize;
+	return 0.185 / (relativeBlur + 0.4);
```
WilcoFiers, line 118: I used a 20px font when I got these numbers, so I divided the numbers by 20. Block 9 in the text-shadow integration tests handle this.

test/integration/rules/color-contrast/text-shadows.html
```diff
@@ -0,0 +1,551 @@
...
+			console.log(res);
+			/*
+      False positives:
+      - b5-r3-i1, b5-r3-i2
+      - b7-r3-i1, b7-r3-i2
+
+      False negatives:
+      - b3-r1-i1
+      - b4-r1-i1, b4-r2-i1
+      - b8-r1-i1, b8-r5-i1
```
WilcoFiers, line 540: Left this here just to have a record of it. All these test cases are on the edge of pass and fail. If we ever revisit them, these are the ones I think could be improved on, but I suspect I've put enough time into this it'll never come up.

doc/check-options.md
```diff
@@ -147,6 +147,7 @@ All checks allow these global options:
 | `boldValue`                                                 | `700`   | The minimum CSS `font-weight` value that designates bold text                                               |
 | `boldTextPt`                                                | `14`    | The minimum CSS `font-size` pt value that designates bold text as being large                               |
 | `largeTextPt`                                               | `18`    | The minimum CSS `font-size` pt value that designates text as being large                                    |
+| `shadowOutlineEmMax`                                        | `0.1`   | Maximum thickness of `text-shadow` in `em` at which it can server as an outline                             |
```
straker, line 150: I had to figure out why thickness meant based on the code. Rather than guessing it would be best to make it very clear what the option looks at and does.

```suggestion
| `shadowOutlineEmMax`                                        | `0.1`   | The maximum `blur-radius` value (in ems) of the CSS `text-shadow` property. `blur-radius` values greater than this value will be treated as a background color rather than an outline color.                             |
```

straker (changes requested):

test/integration/rules/color-contrast/text-shadows.html
straker: Can you put this comment in the code? Otherwise in a month or two we won't remember what this comment means.

straker (approved):

Issue #2578, opened 2020-10-21
Ignore elements from name rules that don't need one


WilcoFiers:
**Espected**: When an element that natively needs an accessible name is given a role that does not require a role, axe-core ignores does not call these out as issues.

**Actual**: In some instances, axe-core will fail these. For example:

```html
<img role="separator" />
<button role="gridcell" disabled>
  <!-- debatable if this needs it or not, ARIA says it does -->
</button>
```

This exception should probably only apply to elements that can not be focused. I think we can implement this by adding a matcher to the image-alt, svg-img-alt, button-name, link-name ... rules.

padmavemulapati:
Verified with the latest axe-core  develop code(v4.0.2)- 
For the following rules, axe-core ignoring the issue when needs an accessible name is given a role that does not require a role.
1.button-name
2. image-alt
3. input-button-name
4. input-img-alt
5. object-alt
for ex: button-name: test-snippet: '<button id="target" role="button"></button>'
![Uploading image.png…]()

PR #2629, opened 2020-11-09
fix(name-rules): ignore when explicit roles don't require a name

WilcoFiers:
Accessible name rules for native elements (image-alt, button-name, etc.) should be inapplicable if the role is changed to a role that does not require an accessible name, but only if that element is not focusable. This does not include link-name /area-alt because even with `tabindex="-1"`, these can be focused with click / touch.

Closes issue: #2578

## Reviewer checks

### DIFF
--- a/doc/standards-object.md
+++ b/doc/standards-object.md
@@ -72,8 +72,10 @@ The [`ariaRoles`](../lib/standards/aria-roles.js) object defines valid ARIA role
 - `requiredOwned` - array(optional). List of required owned roles.
 - `requiredAttrs` - array(optional). List of required attributes.
 - `allowedAttrs` - array(optional). List of allowed attributes (besides any required and global ARIA attributes).
+- `superclassRole` - array(optional). List of superclass roles.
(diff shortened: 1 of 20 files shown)


doc/standards-object.md
```diff
@@ -72,8 +72,10 @@ The [`ariaRoles`](../lib/standards/aria-roles.js) object defines valid ARIA role
 - `requiredOwned` - array(optional). List of required owned roles.
 - `requiredAttrs` - array(optional). List of required attributes.
 - `allowedAttrs` - array(optional). List of allowed attributes (besides any required and global ARIA attributes).
+- `superclassRole` - array(optional). List of superclass roles.
+- `accessibleNameDrequired` - boolean(optional. Default `false`). If elements with this role require an accessible name.
```
WilcoFiers, line 76:
```suggestion
- `accessibleNameRequired` - boolean(optional. Default `false`). If elements with this role require an accessible name.
```

straker (approved):

Issue #2479, opened 2020-08-24
Improve performance of color-contrast-matches


straker:
The color contrast matches function took 30s on a page with 30,000 `<p>` tags. We should investigate ways to improve its performance.

See https://github.com/dequelabs/axe-core/issues/2445.

michael-siek:
See: https://www.chase.com/content/dam/chaseonline/en/demos/cbo/launch.html slow due to queryselectorall for every element ancestor looking for arialabelledby id

straker:
Dev task

PR #2635, opened 2020-11-12
fix(color-contrast): greatly improve color-contrast-matches speed. add aria/get-accessible-ref

straker:
The color-contrast-matches function would try to query the entire DOM for every node with an `id` attribute, and all children of that node would also do the same check. This caused the matcher to take ~40s on a relatively small page with lots of IDs. Since we were already caching id references in `aria/is-accessible-ref`, I was able to leverage that same code to create a new function that returns an array of all id ref DOM nodes. That way we could just reference that cache in color-contrast-matches and reduce the time to about 1 second.

Here are the perf results in running only the color-contrast rule on https://www.chase.com/content/dam/chaseonline/en/demos/cbo/launch.html.

| Time Before (ms) | Time After (ms) |
| - | - |
| 42947.73 | 1271.18 |

**NOTE:** We are in a code freeze so cannot merge this until after 4.1 is released on Monday.

Closes issue: #2479 

## Reviewer checks


lib/commons/aria/is-accessible-ref.js
```diff
@@ -49,17 +16,10 @@ function isAccessibleRef(node) {
...
-			const { type } = standards.ariaAttrs[attr];
-			return idRefsRegex.test(type);
-		});
-
-		cacheIdRefs(root, refAttrs);
+		cacheIdRefs(root);
 	}
 
-	return cache.get('idRefs')[id] === true;
+	return !!cache.get('idRefs')[id];
```
WilcoFiers, line 22: Would be better for this to call `getAccessibleRef()` instead of reaching into its cache.

lib/commons/aria/get-accessible-refs.js
```diff
@@ -0,0 +1,25 @@
+import getRootNode from '../dom/get-root-node';
+import cache from '../../core/base/cache';
+import cacheIdRefs from './cache-idrefs';
```
WilcoFiers, line 3: Suttest putting that cachedIdRefs into this file. I don't think that's a function that should ever be available anywhere else.

lib/commons/aria/get-accessible-refs.js
```diff
@@ -0,0 +1,25 @@
...
+	root = root.documentElement || root; // account for shadow roots
+	const id = node.id;
+
+	// because axe.commons is not available in axe.utils, we can't do
+	// this caching when we build up the virtual tree
+	if (!cache.get('idRefs')) {
+		cacheIdRefs(root);
+	}
+
+	return cache.get('idRefs')[id] || [];
```
WilcoFiers, line 22: If light DOM has an ID, and shadow DOM has a ref this can return the light DOM ref from the shadow DOM ref because we're not keeping track of what root node the idref belonged to.

WilcoFiers (dismissed):

WilcoFiers (approved):
Looks good. I think there are a lot more places in axe-core that now we have a cache of ID references we could speed up. Probably best to open a tech debt issue for that one instead of pile it onto this PR. Here's a list though:

- checks
    - aria-required-parent-evaluate
    - explicit-evaluate
    - hidden-explicit-label-evaluate
    - multiple-labels
    - duplicate-id-evaluate
- matches
    - duplicate-id-active-matches
    - duplicate-id-misc-matches
    - scrollable-region-focusable-matches
- commons
    - text.labelVirtual
    - dom.findElmsInContext
    - dom.isVisible (usemap lookup)

To make that easier, we can probably simplify by having something like a `idrefsReverse` function, that takes a node (with an ID) and an attribute and looks up any element references the ID with that attribute. I'll leave it up to you if you want to do that in this PR or another.

```js
const virtualControls = idrefsReverse(ancestorNode, 'aria-labelledby')
	.map(control => getNodeFromTree(control));
```

Issue #1468, opened 2019-04-02
Axe-Core 3.2.0 added an ES6 Promise polyfill that overwrites native promise behavior


cscrosati:
After updating to 3.2.0/3.2.2 of axe-core, I noticed a lot of our QUnit test suite failed in interesting ways, where Promise resolution was involved.

After debugging, it turned out when loading axe-core, a promise polyfill was loaded that overrides the native promise behavior. Not only that, it overrides our own polyfill implementation when we use one ourselves.

Can the polyfill only be used if window.Promise doesn't exist? Right now, it overwrites what is in place.

WilcoFiers:
@cscrosati Thanks for reporting. That definitely wasn't the intent. @straker can you create a fix for this?

cscrosati:
BTW, thanks for fixing so promptly! :) Which official release should I expect this to be in?

PR #2637, opened 2020-11-12
core(cleanup): core/lib/imports: remove core-js and core-js-pure dependencies

jayaddison:
These polyfills, previously provided by `core-js` / `core-js-pure`, can also be fulfilled by existing bundled dependencies within `axe-core`.

In particular:

- `es6-promise` can provide `Promise`
- `typedarray` can provide `Uint32Array` (and was an implicit dependency, but is now added to `package.json` explicitly)
- `weakmap-polyfill` can provide `WeakMap`

This *might* provide an opportunity to remove `core-js` and `core-js-pure` completely from `axe-core`'s dependencies.

Unit tests pass **intermittently** locally, so additional verification would definitely help.  In particular, `UnhandledPromiseRejectionWarning` exceptions appear to be thrown unpredictably from the headless browser environment.

**Build Size Impact**

Before @ 3c2429b4032d29cdcfd3083e46e91854dccd733b
```sh
$ npm run-script build
...
./axe.min.js
Size:		419.95 Kb
Gzip Size:	108.92 Kb

Done.
```

After @ a9aebc21675611779b2a396ea50b7e07d36b9ec5 (this changeset)
```sh
$ npm run-script build
...
./axe.min.js
Size:		406.84 Kb
Gzip Size:	103.71 Kb

Done.
```

Relates to #1935

TODO:

- [ ] Investigate flaky tests due to unhandled promise rejections

## Reviewer checks

### DIFF
--- a/lib/core/imports/index.js
+++ b/lib/core/imports/index.js
@@ -9,10 +9,6 @@ import es6promise from 'es6-promise';
 import { Uint32Array } from 'typedarray';
 import 'weakmap-polyfill';
 
-if (!('WeakMap' in window)) {
-	window.WeakMap = WeakMap;
-}
-
 if (!('Promise' in window)) {
 	es6promise.polyfill();
 }
--- a/lib/core/imports/index.js
+++ b/lib/core/imports/index.js
@@ -9,10 +9,12 @@ import es6promise from 'es6-promise';
 import { Uint32Array } from 'typedarray';
 import 'weakmap-polyfill';
 
-es6promise.polyfill();
-
 if (!('WeakMap' in window)) {
 	window.WeakMap = WeakMap;
+}
+
+if (!('Promise' in window)) {
+	es6promise.polyfill();
 }
 
 if (!('Uint32Array' in window)) {
--- a/lib/core/imports/index.js
+++ b/lib/core/imports/index.js
@@ -5,30 +5,16 @@ import doT from '@deque/dot';
 import emojiRegexText from 'emoji-regex';
 import memoize from 'memoizee';
 
-import Promise from 'core-js-pure/features/promise';
-import Uint32Array from 'core-js-pure/features/typed-array/uint32-array';
-import WeakMap from 'core-js-pure/es/weak-map';
+import es6promise from 'es6-promise';
+import { Uint32Array } from 'typedarray';
+import 'weakmap-polyfill';
 
-/**
- * Polyfill `WeakMap`
- * Reference https://github.com/zloirock/core-js/
- */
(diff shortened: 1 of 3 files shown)


straker:
Sweet! Love seeing more file size reduction. I'll take a look at this more in-depth probably tomorrow. 

One thing to note though is the if statement we had around the `Promise` polyfill was to ensure that if someone else had loaded a promise library we didn't override it, which was the problem in https://github.com/dequelabs/axe-core/issues/1468. Unless `es6-promise` has changed how it loads the polyfill we'll probably still need to keep that check in as that was the fix in https://github.com/dequelabs/axe-core/pull/1470/files.

jayaddison:
Ah, great context - thanks @straker!  I hope I'm not incorrectly taking the code backwards given that `es6-promise` was in use previously; I didn't realize that.  Anyway, I'll restore that conditional check here.

jayaddison:
(and now I wonder whether `typedarray` is also potentially clobbering native `Uint32Array` implementations when present.  `weakmap-polyfill` claims to check for a native version before polyfilling, but I'm not sure about `typedarray` based on the source)

lib/core/imports/index.js
```diff
@@ -5,30 +5,18 @@ import doT from '@deque/dot';
...
+import { Uint32Array } from 'typedarray';
+import 'weakmap-polyfill';
 
-/**
- * Polyfill `WeakMap`
- * Reference https://github.com/zloirock/core-js/
- */
 if (!('WeakMap' in window)) {
 	window.WeakMap = WeakMap;
 }
```
straker, line 15: I believe this block is no longer need. It looks like weakmap-polyfill checks if `self.WeakMap` exists before doing anything and then sets it on the global `self` itself. https://github.com/polygonplanet/weakmap-polyfill/blob/master/weakmap-polyfill.js#L11-L28

```suggestion
```

straker (changes requested):

straker (approved):
Tested `typedarray` with the following code in node and it does not add the `Unit32Array` to the global namespace, so we should be safe to import it in.

```js
delete global.Uint32Array;

var typedarray = require("typedarray");
var Uint32ArrayPolyfill = typedarray.Uint32Array;

console.log(global.Uint32Array);  // undefined
console.log(Uint32Array);  // error
```

jayaddison:
Nice - thanks!  I wasn't able to reproduce the flaky tests I'd been experiencing locally; but if tests happen to fail more regularly (particularly related to `UnhandledPromiseRejectionWarning`), please do consider this for a revert.

Issue #2643, opened 2020-11-16
Util functions should have default arguments set


straker:
Functions like `text.sanitize` and `utils.tokenList` expect that the provide argument is a string. However, this forces us to pass an empty string ourselves to handle null attributes such as `tokenList(node.getAttribute('aria-labelledby') || '')`.

Instead, both these functions should just have a default argument of an empty string so we can remove all these `|| ''`.

```js
function tokenList(str = '') {
	return str
		.trim()
		.replace(/\s{2,}/g, ' ')
		.split(' ');
}
```

straker:
Dev task

PR #2645, opened 2020-11-16
refactor: set default return for text.sanitize and utils.tokenList

straker:
Now we don't have to pass an empty string for these functions. Also removed `use strict` as it's not needed (especially if we want to use default parameters as the code will error on compile if it's there).

Closes issue: #2643 

## Reviewer checks

### DIFF
--- a/lib/checks/generic/attr-non-space-content-evaluate.js
+++ b/lib/checks/generic/attr-non-space-content-evaluate.js
@@ -14,8 +14,8 @@ function attrNonSpaceContentEvaluate(node, options = {}, vNode) {
 		return false;
 	}
 
-	const attribute = vNode.attr(options.attribute) || '';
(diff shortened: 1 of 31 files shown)


lib/checks/generic/attr-non-space-content-evaluate.js
```diff
@@ -14,8 +14,8 @@ function attrNonSpaceContentEvaluate(node, options = {}, vNode) {
 		return false;
 	}
 
-	const attribute = vNode.attr(options.attribute) || '';
-	const attributeIsEmpty = !sanitize(attribute.trim());
+	const attribute = vNode.attr(options.attribute);
+	const attributeIsEmpty = !sanitize(attribute);
```
straker, line 18: sanitize already trims so no need to trim again.

WilcoFiers (approved):

PR #2648, opened 2020-11-17
tests: switch to karma

straker:
This looks like a lot of files but most are test files being updated to work with the new path structure. This updates our entire unit (and part of integration) test pipeline to use Karma. Doing this gains us a few benefits:

1. `npm test` now works in the terminal. I could never run `npm test` in the terminal as two css-preprocessor tests would consistently fail
1. allow us to not have to build any test `.html` files, removing those scripts from our gruntfile
1. this in turns gets us closer to removing grunt as our build tool now that we don't need it to build axe nor the test files
1. allows us to get debug information from tests. a [particular test in IE11](https://app.circleci.com/pipelines/github/dequelabs/axe-core/1385/workflows/513f1ab7-b9da-428b-a3cb-35e739758baa/jobs/24064) is super flakey only in CircleCI and has been failing a lot. However, since it was using selenium webdriver to load the tests, we couldn't get debug information out of the code thus making it nigh impossible to figure out what was going on. By using karma, we can now add debug information into the test file and karma will spit it out during the test
1. we can now add `.only` to a test and have `npm test` run just that single suite
1. the tests are faster now as we don't have to navigate to individual `.html` files. In particular, the IE11 tests went from 12mins average to 4mins.

## TODO:

* [x] As karma is not written to open HTML files directly, we will still need to use selenium webdriver to test all the `integration/full` files. Will work on that next.
* [x] Get IE11 working

## Reviewer checks


test/commons/color/get-background-color.js
```diff
@@ -743,13 +743,14 @@ describe('color.getBackgroundColor', function() {
 
 	it('should return the html canvas inherited from body bgColor when element content does not overlap with body', function() {
 		fixture.innerHTML =
-			'<div id="target" style="position: relative; top: 2px;">Text</div>';
+			'<div id="target" style="position: relative; top: 2px; height: 10px;">Text</div>';
```
straker, line 746: without the mocha styles on the page, this test would fail as the default div `bottom` rect was larger than the body `bottom` rect and would consider it overlapping. Just fixed to adjust the height of the div to be smaller than the body `bottom`.

test/commons/dom/url-props-from-attribute.js
```diff
@@ -55,19 +55,9 @@ describe('dom.urlPropsFromAttribute', function() {
...
-		 * 	- http://localhost:9876/test/commons
-		 *  - http://localhost:9876/test/commons/?grep=url...
-		 *
-		 * Hence only asserting the what we know to be static values
-		 */
-		assert.equal(actual.protocol, 'http:');
-		assert.equal(actual.hostname, 'localhost');
-		assert.equal(actual.port, '9876');
-		assert.equal(actual.pathname, '/test/commons/');
+		assert.equal(actual.protocol, location.protocol);
```
straker, line 58: Since we can now longer guess what the port / hostname will be, we should get them from the current location. This is actually more correct as when using Parallels to test this file my hostname would be `192.168.1.n` and would fail this test.

test/integration/api/external/index.js
```diff
@@ -93,8 +93,8 @@ describe('external API', function() {
 				flat(axe._tree[0])
 			);
 		});
-		it('must return undefined for nodes not in the axe._tree', function() {
-			assert.isUndefined(axe.utils.getNodeFromTree(randomElement()));
+		it('must return null for nodes not in the axe._tree', function() {
```
straker, line 96: This has always returned `null` so I'm not sure how it's been passing.

test/integration/virtual-rules/area-alt.js
```diff
@@ -1,4 +1,4 @@
-describe('area-alt', function() {
+describe('area-alt virtual-rule', function() {
```
straker, line 1: Without the additional `virtual-rule` this describe name would collide with describes from the `checks` directory and not output any describe blocks, making this a lot harder to know what tests this was from.

stephenmathieson:
This is amazing! Excellent work, Steve!

test/karma.conf.js
```diff
@@ -0,0 +1,77 @@
+var path = require('path');
+
+// allow running only certain directories
+var testDirs = [
```
straker, line 4: This code is just to make things easier to run directories, similar to opening up `localhost/test/checks`. This way you can pass `karma start test/karma.config.js testDirs=checks` and only run the check files (which is great as core files take the longest to run)

test/testutils.js
```diff
@@ -10,6 +10,17 @@ if (window.__AXE_EXTENSION__) {
 /*eslint indent: 0*/
 var testUtils = {};
 
+var fixture = document.createElement('div');
```
straker, line 13: Adding fixture to the body as we don't have the `runner.tmpl` anymore

test/integration/full/test-webdriver.js
```diff
@@ -0,0 +1,236 @@
+/*global window, Promise */
```
straker, line 1: This is more or less a copy paste of the grunt script by the same name, minus all the grunt things and with the ability to configure chrome and chromedriver

package.json
```diff
@@ -71,7 +70,6 @@
 		"test:rule-help-version": "mocha test/test-rule-help-version.js",
 		"test:node": "mocha test/node/*.js",
 		"version": "echo \"use 'npm run release' to bump axe-core version\" && exit 1",
-		"prepublishOnly": "grunt build && grunt file-exists",
```
stephenmathieson, line 72: We should still have a `prepublishOnly` (or `prepare`) script. Otherwise we'll have to manually run the build before we `npm publish`. Having this here reduces the chances of publishing a broken version of axe.

package.json
```diff
@@ -87,38 +85,28 @@
...
 		"grunt-contrib-concat": "^1.0.1",
-		"grunt-contrib-connect": "^2.0.0",
-		"grunt-contrib-copy": "^1.0.0",
 		"grunt-contrib-uglify": "^4.0.0",
 		"grunt-contrib-watch": "^1.1.0",
-		"grunt-parallel": "^0.5.1",
-		"grunt-run": "^0.8.1",
 		"html-entities": "^1.2.0",
 		"http-server": "^0.12.3",
-		"husky": "^3.0.0",
```
stephenmathieson, line 116: Why are you removing this? Husky is what manages our Git hooks (automatically running Prettier, etc).

stephenmathieson (changes requested):
It looks like a few thing were removed from `package.json` that shouldn't have been. We should either document these changes or add them back.

package.json
straker: the publish script already runs the `verify-release` script which checks if all files are there, so I figured we didn't need this one anymore https://github.com/dequelabs/axe-core/blob/develop/.circleci/config.yml#L176

package.json
straker: I wondered what that did. I did a search of all the packages and this wasn't used in a require anywhere.

straker: I think `lint-staged` runs our prettier. It doesn't look like Husky is used in our `package.json`

stephenmathieson: `husky` is what runs `lint-staged`. When installed, it sets up our Git hooks.

package.json
stephenmathieson: I see. I suppose we can safely remove this as long as we always remember to do `npm run build` first. 🤷

package.json
straker: We determined that we still need husky but it is not properly configured in our package.json (which explains how prettier isn't run on our commits sometimes).

.circleci/config.yml
```diff
@@ -75,7 +75,13 @@ jobs:
...
       - checkout
       - <<: *restore_dependency_cache_unix
-      - run: npm run test
+      - run: npm run build
+      - run: npm run test:unit -- --browsers Chrome,Firefox
+      - run:
+          command: npm run start
+          background: true
+      - run: npm run test:integration -- browser=chrome
+      - run: npm run test:integration -- browser=firefox
```
WilcoFiers, line 84: Can we change this so that it's consistent with test:unit?

```suggestion
      - run: npm run test:integration -- --browsers Chrome,Firefox
```

test/integration/rules/runner.js
```diff
@@ -28,126 +31,117 @@
...
-						describe(collection, function() {
-							var nodes;
-							before(function() {
-								if (typeof results[collection] === 'object') {
-									nodes = results[collection].nodes;
-								}
-							});
+	// don't run rules that are deprecated and disabled
+	var deprecated = rule.tags.indexOf('deprecated') !== -1;
+	(deprecated && !rule.enabled ? describe.skip : describe)(ruleId, function() {
```
WilcoFiers, line 44: I think we shouldaI don't think we should skip those.

WilcoFiers (dismissed):
Couple small things, looks excellent.

.circleci/config.yml
```diff
@@ -75,7 +75,13 @@ jobs:
     steps:
       - checkout
       - <<: *restore_dependency_cache_unix
-      - run: npm run test
+      - run: npm run build
+      - run: npm run test:unit -- --browsers Chrome,Firefox
+      - run:
+          command: npm run start
+          background: true
```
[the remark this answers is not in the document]
straker: We'll create a tech debt ticket once this is merged

WilcoFiers (approved):

Issue #2535, opened 2020-10-01
aria-required-children violation for seemingly valid use of group role in menu/menubar


JustSlone:
Hey folks, it looks like in version 4.0.0 the update to the aria-required-children rule fixing (I think) this issue: #2076 caused the following structure to be invalid: 

```html
<div role="menu">
  <ul role="group">
    <li role="menuitem">Inbox</li>
    <li role="menuitem">Archive</li>
    <li role="menuitem">Trash</li>
  </ul>
  <ul role="group">
    <li role="menuitem">Custom Folder 1</li>
    <li role="menuitem">Custom Folder 2</li>
    <li role="menuitem">Custom Folder 3</li>
  </ul>
  <ul role="group">
    <li role="menuitem">New Folder</li>
  </ul>
</div>
```
Repro: https://codepen.io/justslone-the-encoder/pen/KKzLLrz?editors=1111 
[gist of HTML for convenience](https://gistpreview.github.io/?66ccb25097b58e80ac4f2f5a053f5a04)

I'm not sure this was an intended consequence of the rule change made in version 4.0.0 as this structure seems valid and useful. Also this structure is allowed per the ARIA 1.2 specification

Expectation: Per [ARIA 1.2 menubar spec](https://w3c.github.io/aria/#menubar) and general examples on the use of the group role [(Example 2 here)](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_group_role) this should not be a violation
Actual: Errors are thrown for aria-required-children and aria-required-parent

Motivation: It seems this normal use of the group role should not be deemed invalid. Also since axe is used for validating accessibility conformance broadly, introducing this violation now will require sites to change away from a valid structure, only to change back again once 1.2 is final. 

<pre><code>
axe-core version: 4.0.0+
axe-webdriver, extension or other integration version: N/A
</code></pre>

padmavemulapati:
Verified with the latest axe-core-develop branch code , dated - 08/12/2020-
`aria-required-children` and `aria-required-parent` are showing correct results when added `role="group"`
code snippet for pass:`<div role="menu" id="target"><ul role="group"><li role="menuitem">Menuitem</li></ul></div>`
Pass-  condition screenshot:
![image](https://user-images.githubusercontent.com/47409579/101485940-ee821900-3981-11eb-990c-93cc42ec34eb.png)

code snippet for voilation:
1. role allows group and group does not have required child
[...]

PR #2661, opened 2020-11-24
fix(aria-required-children): allow group and rowgroup roles

straker:
This allows `group` and `rowgroup` on roles that allow it but also enforces that the descendants of the group have the required child as well.

Closes issue: #2535

## Reviewer checks


lib/checks/aria/aria-required-children-evaluate.js
```diff
@@ -103,7 +108,7 @@ function ariaRequiredChildrenEvaluate(node, options, virtualNode) {
     return true;
   }
 
-  const ownedRoles = getOwnedRoles(virtualNode);
+  const ownedRoles = getOwnedRoles(virtualNode, required);
```
WilcoFiers, line 111: Can you maybe just change like 107 to `if (required === null) {`. Right now required looks like a boolean. It's a little confusing.

WilcoFiers (changes requested):
Very small suggestion.

WilcoFiers (approved):

Issue #2658, opened 2020-11-24
Start integration server from the test-runner file


straker:
See https://github.com/dequelabs/axe-core/pull/2648#discussion_r528620458

I tried to do this but it caused a few problems. First, I could only get it to work with with `execa` but I was having a lot of trouble terminating the server once the tests ended. I got it working on a Mac but when we tested it with a PC the server wouldn't terminate.

Second, it seemed to cause IE11 to hang after about 47 page navigations. It worked locally on a PC machine but would just hang in CircleCI.

PR #2664, opened 2020-11-25
tests: start server on integration test

straker:
Closes issue: #2658 

## Reviewer checks


stephenmathieson (approved):
Looks like we're intentionally adding a race condition, but I'm not sure how else to handle this. LGTM but we should keep an eye on this.

straker:
Nope, this is still timing out after 47 tests... I have no idea what is going on but I don't know how else to do this in a single command so we don't have to start the server beforehand

stephenmathieson:
If you like, we can pair on it next week. Should I set something up?

straker:
IT WORKS! So this is the "best" way I've found to do this but it adds a lot of npm scripts. 

`start-server-and-test` does what we want, but it has a problem in that it can't pass command line arguments to the test command. That resulted in creating individual test scripts for each of the browsers. Unfortunately, it also [has a bug in Windows](https://github.com/bahmutov/start-server-and-test/issues/239) that doesn't allow it to work with npm commands, so I had to create yet another set of individual test scripts so we can run the script name instead of a command.

stephenmathieson (approved):

stephenmathieson:
Well done! Glad you figured it out!

WilcoFiers (approved):

PR #2668, opened 2020-11-27
[WIP] chore: fix npm run develop

WilcoFiers:
The karma integration broke our use of the develop branch. I think it is important to keep the develop command functioning the way it did so that we can debug tests in the browser. 

Here's what needs to be fixed:
- [x] Build axe-core on start
- [x] Watch for changes in lib to rebuild axe-core
- [x] Spin up a server to run tests
- [ ] Allow running all tests in the browser

Some of the tests broke, haven't figured out why yet. It also isn't possible to run the integration tests in the browser anymore.

## Reviewer checks

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,12 +1,9 @@
-/*eslint
-camelcase: ["error", {"properties": "never"}]
-*/
+/*eslint camelcase: ["error", {"properties": "never"}] */
 module.exports = function(grunt) {
-  'use strict';
-
   grunt.loadNpmTasks('grunt-babel');
   grunt.loadNpmTasks('grunt-contrib-clean');
   grunt.loadNpmTasks('grunt-contrib-concat');
+  grunt.loadNpmTasks('grunt-contrib-connect');
   grunt.loadNpmTasks('grunt-contrib-uglify');
   grunt.loadNpmTasks('grunt-contrib-watch');
   grunt.loadNpmTasks('grunt-bytesize');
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -102,6 +102,7 @@
     "grunt-bytesize": "^0.2.0",
     "grunt-contrib-clean": "^2.0.0",
     "grunt-contrib-concat": "^1.0.1",
+    "grunt-contrib-connect": "^3.0.0",
```
WilcoFiers, line 105: Not too keen on this solution, as we now have both http-server and grunt-connect. Should probably get rid of one of 'm. Maybe I should use `grunt-run` instead and have that start http-server?

Gruntfile.js
```diff
@@ -228,6 +234,7 @@ module.exports = function(grunt) {
     'aria-supported',
     'bytesize'
   ]);
+  grunt.registerTask('prepare', ['build', 'connect']);
   grunt.registerTask('default', ['build']);
-  grunt.registerTask('dev', ['watch']);
+  grunt.registerTask('dev', ['prepare', 'watch']);
```
WilcoFiers, line 239: Can probably get rid of `prepare`, since it's only 2 steps now.

stephenmathieson:
I don't really understand what's going on here (not your fault, Wilco; I'm not familiar with Grunt), but I think this might be a duplicate of #2664.

Also, can we just do `npm start` before running the integration tests?

PR #2670, opened 2020-11-30
tests: run ChromeHeadless for integration test 

straker:
Chrome is [failing to start a lot](https://app.circleci.com/pipelines/github/dequelabs/axe-core/1479/workflows/c9511638-cba6-4217-b2b5-2a5efa5b1fa2/jobs/24819) and with no apparent reason as this was all working fine when running it as Grunt. This is a copied/pasted file that worked before but now fails. Trying this to see if it resolves the issues as suggested in various sites.

Going to cancel and rerun this build a couple of times to see if it's working consistently.

## Reviewer checks

### DIFF
--- a/test/integration/full/test-webdriver.js
+++ b/test/integration/full/test-webdriver.js
@@ -134,7 +134,7 @@ function buildWebDriver(browser) {
     };
   }
 
-  // fix chrome DevToolsActivePort file doesn't exist
+  // fix chrome DevToolsActivePort file doesn't exist in CricleCI
   // @see https://stackoverflow.com/questions/50642308/webdriverexception-unknown-error-devtoolsactiveport-file-doesnt-exist-while-t
   if (browser === 'chrome') {
     capabilities = WebDriver.Capabilities.chrome();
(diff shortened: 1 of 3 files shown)


straker:
Alright. So after trying every combination of things suggested by the interwebs to fix this, the only thing that finally has worked consistently is to use chrome headless. I have stopped and restarted the job about 6 or 7 times now and it has yet to fail to start. I don't understand it but here we are.

test/integration/full/test-webdriver.js
```diff
@@ -4,7 +4,6 @@ var globby = require('globby');
 var WebDriver = require('selenium-webdriver');
 var chrome = require('selenium-webdriver/chrome');
 var chromedriver = require('chromedriver');
-var isCI = require('is-ci');
```
WilcoFiers, line 7: Looks like we can then drop this dependency.

WilcoFiers (changes requested):

WilcoFiers (approved):

Issue #2681, opened 2020-12-08
Runtime issue: TypeError: Cannot read property '0' of undefined


patrickhulce:
Originally filed at https://github.com/GoogleChrome/lighthouse/issues/11721#issuecomment-740267214

Run axe on https://www.rome2rio.com/s/Mumbai/Delhi

Supposedly related an interaction with color-contrast and `overflow: scroll`? From that original Lighthouse issue...

> I notice if I remove a overflow-x: scroll style attribute from two sections (Things to do, Places to stay) then the error goes away

```
TypeError: Cannot read property '0' of undefined
    at g (chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:33:117582)
    at chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:33:119639
    at Array.map (<anonymous>)
    at t.default (chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:33:119593)
    at t.default (chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:33:105942)
    at t.default (chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:33:101530)
    at t.default (chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:33:103754)
    at t.default (chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:33:102856)
    at Object.t.default (chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:33:15179)
    at c.run (chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:33:187710)
```

straker:
Thanks for the issue. It is indeed a separate problem and we'll be sure to get it fixed.

padmavemulapati:
@straker , is it similar to "axe-core Error: Cannot read property 'filter' of undefined" #2483 ?

straker:
@padmavemulapati kinda of. They throw an error in the same place / way, but the causes are different.

padmavemulapati:
Verified with the latest axe-core-develop codebase - 
Working fine with handling of text that overflows outside the parent:
TestScript:
`<div id="1" style="width: 250px; overflow: hidden">`
[...]

PR #2682, opened 2020-12-09
fix(color-contrast): account for text client rects that start outside the parent container

straker:
The element that was causing the problem had 12 client rects, all of them text rects that started on their own lines (so 12 lines of text). Some of those rects started outside the bounds of the element, so we tried to access part of a grid that didn't exist.

Fixed it so if any text rect wasn't fully contained within the parents bounding box, we just use the parents bounding box. Also moved the textElementStack tests to their own file as they were becoming so large.

Closes issue: #2681


lib/commons/dom/get-text-element-stack.js
```diff
@@ -38,14 +38,29 @@ function getTextElementStack(node) {
...
+      // @see https://github.com/dequelabs/axe-core/issues/2681
+      const outsideRectBounds = Array.from(rects).some(rect => {
+        const horizontalMidpoint = rect.left + rect.width / 2;
+        const verticalMidpoint = rect.top + rect.height / 2;
+
+        return (
+          horizontalMidpoint < nodeRect.left ||
+          horizontalMidpoint > nodeRect.left + nodeRect.width ||
+          verticalMidpoint < nodeRect.top ||
+          verticalMidpoint > nodeRect.top + nodeRect.height
```
WilcoFiers, line 60: I think this could be:

```suggestion
          horizontalMidpoint < nodeRect.left ||
          horizontalMidpoint > nodeRect.right ||
          verticalMidpoint < nodeRect.top ||
          verticalMidpoint > nodeRect.bottom
```

WilcoFiers (changes requested):
Very small change

WilcoFiers (approved):

PR #2691, opened 2020-12-15
feat(nested-interactive): new rule to flag nested interactive elements

straker:
Flag roles that have `childrenPresentational: true` and have focusable content.

Closes issue: #601


lib/rules/nested-interactive.json
```diff
@@ -0,0 +1,12 @@
+{
+  "id": "nested-interactive",
```
WilcoFiers, line 2: This looks like it can run on virtual nodes only. Can you put in some tests for that?

lib/rules/nested-interactive.json
```diff
@@ -0,0 +1,12 @@
...
+  "matches": "nested-interactive-matches",
+  "tags": ["cat.keyboard", "best-practice"],
```
WilcoFiers, line 4:
```suggestion
  "tags": ["cat.keyboard", "wcag2a", "wcag412"],
```

lib/rules/nested-interactive.json
```diff
@@ -0,0 +1,12 @@
...
+  "metadata": {
+    "description": "Ensure interactive controls are not nested",
+    "help": "Nested interactive controls are not announced by screen readers"
```
WilcoFiers, line 7: Help text has to be the short one:

```suggestion
    "description": "Nested interactive controls are not announced by screen readers",
    "help": "Ensure interactive controls are not nested"
```

test/rule-matches/nested-interactive-matches.js
```diff
@@ -0,0 +1,30 @@
...
+    rule = axe._audit.rules.find(function(rule) {
+      return rule.id === 'nested-interactive';
+    });
+  });
+
+  afterEach(function() {
+    fixture.innerHTML = '';
+  });
+
+  it('should match if element has children presentational', function() {
```
WilcoFiers, line 16: can you add a test that ensures this works for explicit roles too. Maybe a link with role=button?

WilcoFiers (dismissed):

WilcoFiers (approved):

PR #2692, opened 2020-12-15
chore: fix grunt develop

straker:
This fixes `npm run develop` by running tests after a file has changed. The karma file then runs only the tests that are associated with the file that was changed (instead of all tests). 

It does this by looking at the passed in file path and matching it against known tests. For example, a file ending with `-matches.js` should run a rule by the same name in the `rule-matches` directory. This isn't guaranteed to find a file that matches (for example, there are some rule-matches that aren't tested, or some aria checks that don't have the same name of the test file), but karma will alert you if there isn't a test by that name. So you can either fix it (like we should with the aria checks) or ignore it.

This also fixes a test failure in the api tests that would happen very sporadically. It turns out that the tree walker was adding comment nodes to the list of elements to use for `randomNodeInTree()` and that would throw off our APIs as we don't expect a comment node to be passed in (so `getNodeFromTree` would fail to return anything).

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -28,6 +28,11 @@ module.exports = function(grunt) {
     langs = [''];
   }
 
+  // run tests only for affected files instead of all tests
+  grunt.event.on('watch', function(action, filepath) {
+    grunt.config.set('watch.file', filepath);
+  });
+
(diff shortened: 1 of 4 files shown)


WilcoFiers (changes requested):
We should document this.

WilcoFiers (approved):
Approved, assuming docs follow shortly.

PR #2693, opened 2020-12-15
chore: split test:unit into different parts

straker:
This is just a minor quality of life change that helps make running `test:unit` a little bit nicer. Right now it runs all tests in a single run, but won't fail early if something in `commons` fails. This changes it so each directory is run individually so it will exit early if something fails.

Not only that, but trying to debug the tests was a bit rough as running all the tests in the browser was slow, especially in IE11 where https://github.com/dequelabs/axe-core/pull/2682 was failing but I couldn't reproduce it locally when running that specific test, but could not get all the tests to run properly in the browser. Splitting the tests out should help with debugging.

Lastly, the `test:debug` script just makes it easier to run karma in debug rather than remembering the commands. You can still pass `-- testDirs` to it so you can debug individual test dirs as well.


WilcoFiers (changes requested):
Can you update the contribute docs to explain how to use these commands?

WilcoFiers (approved):
Approved, assuming docs follow shortly

PR #2702, opened 2020-12-18
feat(role-text): add role-text rule

straker:
Add support for `role=text` only if the contents do not have focusable descendants. This is very similar to our new `nested-interactive` rule except that instead of focusing on interactive roles this focuses on making sure the `text` role is used in such a way that does not override the semantics of the children.

Now we could have made `role=text` also have `childrenPresentational: true`, but that means it would have thrown errors with the `nested-interactive` rule which didn't make much sense (it's not really an interactive component). This allows use to have its own rule page with an explanation on the proper use.

Also note that this allows `role=text` on any element that is allowed any role. This means you can do

```html
<mark role="text">Hello</mark>
```

And it will pass. I figured this was fine since these elements do allow any role to begin with.

Closes issue: #1597


doc/rule-descriptions.md
```diff
@@ -100,7 +102,6 @@ Rules that do not necessarily conform to WCAG success criterion but are industry
...
 | [meta-viewport-large](https://dequeuniversity.com/rules/axe/4.1/meta-viewport-large?application=RuleDescription)                                 | Ensures &lt;meta name=&quot;viewport&quot;&gt; can scale a significant amount                              | Minor              | cat.sensory-and-visual-cues, best-practice      | failure                    |
 | [meta-viewport](https://dequeuniversity.com/rules/axe/4.1/meta-viewport?application=RuleDescription)                                             | Ensures &lt;meta name=&quot;viewport&quot;&gt; does not disable text scaling and zooming                   | Critical           | cat.sensory-and-visual-cues, best-practice, ACT | failure                    |
-| [nested-interactive](https://dequeuniversity.com/rules/axe/4.1/nested-interactive?application=RuleDescription)                                   | Ensure interactive controls are not nested                                                                 | Serious            | cat.keyboard, best-practice                     | failure, needs&nbsp;review |
```
straker, line 103: Not sure how this ended up in the wrong place.

lib/rules/aria-text.json
```diff
@@ -0,0 +1,12 @@
+{
+  "id": "aria-text",
+  "selector": "[role=text]",
+  "tags": ["cat.aria", "best-practice"],
+  "metadata": {
+    "description": "Ensures \"role=text\" is used correctly",
+    "help": "\"role=text\" should only be used on elements with no focusable descendants"
```
WilcoFiers, line 7: Help test should be shorter than the description.

```suggestion
    "help": "Do not use \"role=text\" on elements with no focusable descendants"
```

lib/rules/aria-text.json
```diff
@@ -0,0 +1,12 @@
+{
+  "id": "aria-text",
+  "selector": "[role=text]",
+  "tags": ["cat.aria", "best-practice"],
+  "metadata": {
+    "description": "Ensures \"role=text\" is used correctly",
```
WilcoFiers, line 6: This should probably be a little more descriptive.

WilcoFiers (changes requested):

WilcoFiers (approved):

Issue #1575, opened 2019-05-16
Rule to warn about aria-label on static elements


cycomachead:
Expectation: 
I think a new rule to warn about aria-label on static elements would be welcome.

https://developer.paciellogroup.com/blog/2017/07/short-note-on-aria-label-aria-labelledby-and-aria-describedby/ as one example of the rules to follow.

Actual: 
New rule, so nothing happens yet. 

Motivation:
This tripped me up when starting out -- it's not obvious that aria-label will _not_ override the text in an element. It some cases it's easier to add aria-label than to add an additional element with sr-only, and I didn't find it intuitive as to why that'd be better. 

Additionally, at least VO (Mac / iOS) does respect the aria-label as the label that's read, so it's not a consistently implemented rule and can be missed in testing. 

I think an especially common case for this is with something like icon fonts.
`<i class="fa fa-cog" aria-label="Settings"></i>` seems like an intuitive thing to do.

WilcoFiers:
Good idea. Seems like a good best-practice rule. Are you interested in writing the rule?

cycomachead:
Sure, I'll try to give it a go soon!

straker:
Hey @cycomachead, just wanted to check in and see how it's going. Please let us know if you have any questions or would like some help with the rule.

straker:
https://github.com/dequelabs/axe-core/pull/2712#issuecomment-758867319

dylanb:
@straker will this be in the next release? Should we add it to the milestone?

padmavemulapati:
Verified with the latest axe-core code base,
below results for "aria-valid-attr-value" rule fails when `aria-label` and  `aria-labelledby` has all possible combinations

![image](https://user-images.githubusercontent.com/47409579/111941586-5b678c00-8af7-11eb-8c0a-ba5e78dc1895.png)
and 
error when generic elements use aria-label and aria-labelledy
![image](https://user-images.githubusercontent.com/47409579/111941786-d761d400-8af7-11eb-9c45-f162e2bc5db1.png)

PR #2712, opened 2020-12-22
feat(aria-label): new rule to flag incorrect use of aria-label, aria-labelledby, and aria-describeby

straker:
Closes issue: #1575


lib/rules/aria-label.json
```diff
@@ -0,0 +1,12 @@
+{
+  "id": "aria-label",
+  "selector": "[aria-label], [aria-labelledby], [aria-describedby]",
+  "tags": ["cat.aria", "best-practice"],
+  "metadata": {
+    "description": "Ensure aria-label(ledby) is only used on non-static elements",
+    "help": "Use aria-label, aria-labelledby, and aria-describedby on elements with non-static roles"
```
WilcoFiers, line 7: I'd like to see some test data on this. What is the problem we are solving with this rule?

WilcoFiers (changes requested):
We're going to need to run this rule by the JC. Can you do some testing for this to show what the problem is that comes from this? I suspect we're going to get complaints about this one.

straker:
I think this PR is not needed and we should instead use the [Prohibited States and Props from ARIA 1.2](https://www.w3.org/TR/wai-aria-1.2/#prohibitedattributes) as they cover this exact use case but in a more generic way.

Related discussions:
- https://github.com/w3c/aria/issues/756
- https://github.com/w3c/aria/pull/985

straker:
Note, this may not work with the `<i>` attribute as it doesn't have a role mapping. I will do some testing `aria-label` on different elements to see how screen readers handle it.

straker:
From Scott O'Hara on asking about HTML elements with no role mapping:

> no mapping = generic
> generic doesn't allow naming

straker:
Tested the following DOM in different browsers:

```html
<div aria-label="my label">foo</div>
<span aria-label="my label">foo</span>
<i aria-label="my label">foo</i>
<p aria-label="my label">foo</p>

<div aria-labelledby="label">foo</div>
<span aria-labelledby="label">foo</span>
<i aria-labelledby="label">foo</i>
<p aria-labelledby="label">foo</p>
```

No browser / screen reader combination read the `aria-label` or the `aria-labelledby` attributes. All names returned as "foo"

Issue #2649, opened 2020-11-18
locale specific axe.js not generating


padmavemulapati:
When trying to run `npm run build --lang=ja` its not generating `axe.ja.js` file.
Expectation: Should generate specified locale axe.js file

Actual: Generating browser default axe.js file only 
![image](https://user-images.githubusercontent.com/47409579/99562855-9db87980-29ee-11eb-9395-7b39fb9560c2.png)

Verified axe-core: 4.1.0

padmavemulapati:
Verified with the latest code base - axe-core, able to generate `axe.ja.js` for `npm run build -- --lang=ja`,
abs using this `axe.ja.js` analysed webdriverjs's analyse api and seeing the results in japanese language
![image](https://user-images.githubusercontent.com/47409579/103213748-a5940380-4934-11eb-851b-5d9325f83409.png)

padmavemulapati:
Again this issue is reproducing - 
tried to generate ` --lang=de` file but still getting `axe.js` instead `axe.de.js`

@straker , can you please confirm this from your end

clottman:
@padmavemulapati there's a difference in the command that works and the one that doesn't - the extra set of `--`. Those are necessary to actually pass through the option provided to the build script. 

This works for me: 
`npm run build -- --lang=ja`

This does not: 
`npm run build --lang=ja`

padmavemulapati:
@clottman , tried with your suggestion and its working .
`npm run build -- --lang=ja`
![image](https://user-images.githubusercontent.com/47409579/124438322-ab601b00-dd95-11eb-9f04-92a2407975d5.png)

PR #2713, opened 2020-12-22
docs: added information on building for locale with npm

SandersAaronD:
Added information on building for locale with npm. Grunt still works fine but the documentation didn't cover the correct way to pass arguments to grunt through npm. Since this is core npm behavior, it isn't something that can or should be changed, but should be documented so that it can be used for building.

Closes issue: #2649

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,10 @@ Axe can be built using your local language. To do so, a localization file must b
 
 `grunt build --lang=nl`
 
+or equivalently:
+
+`npm build -- --lang=nl`
+
 This will create a new build for axe, called `axe.<lang>.js` and `axe.<lang>.min.js`. If you want to build localized versions, simply pass in `--all-lang` instead.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -92,6 +92,10 @@ Axe can be built using your local language. To do so, a localization file must b
 
 `grunt build --lang=nl`
 
+or equivalently:
+
+`npm build -- --lang=nl`
```
straker, line 97: This seems to fail for me, but adding `run` seems to fix it.

```suggestion
`npm run build -- --lang=nl`
```

straker (changes requested):
Thanks for the doc change. Just a minor fix

README.md
straker: Better yet, we should probably not promote the use of the `grunt` step and just use the `npm run build` process for this.

SandersAaronD: "There should be one-- and preferably only one --obvious way to do it."

straker (approved):

straker:
Approved security and pr title.

Issue #1883, opened 2019-11-06
Allow runOnly parameter for axe.configure


alisan16:
Expectation: Allow setting runOnly parameter with axe.configure in addition to allowing the temporary setting on axe.run

Actual: Setting runOnly in axe.configure does not restrict rules being run to those in the runOnly category (i.e. tags)

Motivation: For our project, we want all rules with specific tags to be run (i.e. section508, wcag2a)

<pre><code>
axe-core version: 3.4.0
axe-webdriver, extension or other integration version: Using a custom driver for Ruby: https://github.com/department-of-veterans-affairs/sniffybara/tree/update-axe
</code></pre>

WilcoFiers:
Thank you for the suggestion. I think this is a good idea. Would you be interested in putting together a pull request?

WilcoFiers:
After talking this through, decided not to make this change. Axe.configure() is intended for use of custom rulesets, not on the fly manipulation of axe-core settings. Axe.run() is better suited for this.

PR #2715, opened 2020-12-22
feat(configure): allow configuring runOnly

straker:
Closes issue: #1883


stephenmathieson (changes requested):
Please update the TypeScript definition file to reflect this change.

WilcoFiers:
After talking this through, decided not to make this change. Axe.configure() is intended for use of custom rulesets, not on the fly manipulation of axe-core settings. Axe.run() is better suited for this.

Issue #1984, opened 2020-01-14
Resolve/ remove `isIE11` checks on various tests


jeeyyy:
Expectation: 
- Should not skip tests based on checking if navigator is IE.

Actual: 
- There are several tests which are skipped when run in IE11.
- The helper function to do this is `axe.testUtils.isIE11`

Motivation: 
- Given we have moved away from appveyor to circleci for running windows/ ie11 tests. It should be easier to debug failures in IE and make the necessary fixes.

PR #2716, opened 2020-12-22
tests: enable previously skipped tests in ie11

straker:
Closes issue: #1984


test/playground.html
```diff
@@ -9,7 +9,9 @@
     window.addEventListener('load', function() {
       axe.run(document, function(err, res) {
         console.log(res.violations);
-        res.incomplete.forEach(issue => console.log(issue));
+        res.incomplete.forEach(function(issue) {
```
straker, line 12: This was purposeful so the playground file could run in ie11.

straker:
It turns out that our ciricle ci windows machine cannot play any audio or videos. Trying to do so results in an error code of 4: MEDIA_ERR_SRC_NOT_SUPPORTED. As such, any of our preload code still requires being skipped in ie11 even though on a windows machine using Parallels the tests pass

WilcoFiers (approved):

Issue #2403, opened 2020-07-20
Add tests for generic rule-matches that are not part of any rules


straker:
Example: `layout-table-matches`.

straker:
We should be able to do this in a test by configuring axe with a new rule that uses the matches function.

PR #2718, opened 2020-12-23
tests: add missing tests for rule-matches

straker:
Added tests for all missing rule-matches.

Closes issue: #2403


test/rule-matches/inserted-into-focus-order-matches.js
```diff
@@ -0,0 +1,45 @@
+describe('inserted-into-focus-order-matches', function() {
```
WilcoFiers, line 1: I think this should also test for non-numerical tabindex values, like tabindex="abc"

WilcoFiers (changes requested):

WilcoFiers (approved):

Issue #2642, opened 2020-11-16
add axe.getRule(id)


straker:
Need a public way to return an individual rule so we don't have to keep doing `axe._audit.rules.find(ruleObj => ruleObj.id === ruleId)`

WilcoFiers:
Here's another use for this method: https://github.com/dequelabs/axe-core/pull/2648/files#diff-6cb05ef5f40343f1168c634a0aaa0c15f5e9cde40d7775c131ef70f263a7bc7aR34

padmavemulapati:
devTask

padmavemulapati:
Validated with the latest axe-core code base - 
after run build, and latest `axe.js` built, run `await axe.utils.getRule('region')` , below is the details we can see:
![image](https://user-images.githubusercontent.com/47409579/116100978-107afe80-a6cb-11eb-8a99-16652b2a9db9.png)

also verified for few more rules:, `bypass`
![image](https://user-images.githubusercontent.com/47409579/116101241-4cae5f00-a6cb-11eb-8628-49aba7a11073.png)

for `color-contrast`
![image](https://user-images.githubusercontent.com/47409579/116101449-71a2d200-a6cb-11eb-8672-40316a319345.png)

padmavemulapati:
Tested on axe-expert_v4.6.0.03923, we can see the `_audit` details by running - `utils.getRule('rule_id')`
![image](https://user-images.githubusercontent.com/47409579/230848253-e2ccaaa2-1e5a-43ab-9bb1-29ad3b32d560.png)

PR #2724, opened 2020-12-24
feat(utils.getRule): add function to get rule by id

straker:
Closes issue: #2642

### DIFF
--- a/lib/core/utils/index.js
+++ b/lib/core/utils/index.js
@@ -31,6 +31,7 @@ export { default as getFriendlyUriEnd } from './get-friendly-uri-end';
 export { default as getNodeAttributes } from './get-node-attributes';
 export { default as getNodeFromTree } from './get-node-from-tree';
 export { default as getRootNode } from './get-root-node';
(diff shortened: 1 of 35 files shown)


lib/core/utils/index.js
```diff
@@ -31,6 +31,7 @@ export { default as getFriendlyUriEnd } from './get-friendly-uri-end';
 export { default as getNodeAttributes } from './get-node-attributes';
 export { default as getNodeFromTree } from './get-node-from-tree';
 export { default as getRootNode } from './get-root-node';
+export { default as getRule } from './get-rule';
```
WilcoFiers, line 34: Why are you putting this on `.utils` instead of on `axe`? Not objecting, to it, but would like to know the reason for it.

WilcoFiers (changes requested):
We should talk about if this should be `axe.getRule()` or `axe.utils.getRule()`. Implementation looks fine.

WilcoFiers (approved):

Issue #2736, opened 2021-01-07
group should be an allowable requiredContext for menuitemcheckbox as of ARIA 1.2


dbjorge:
Expectation: As of [ARIA 1.2](https://www.w3.org/TR/wai-aria-1.2/#menuitemcheckbox) (specifically, PR https://github.com/w3c/aria/issues/737), `group` has become an acceptable required context role for the `menuitemcheckbox` role; axe-core should recognize this now that [it's picking up other ARIA 1.2 role relationship updates](https://github.com/dequelabs/axe-core/issues/2107).

Actual: `lib/standards/aria-roles.js` hasn't been updated to account for that yet, so axe 4.1.1 still reports an `aria-required-parent` failure for the button in this snippet ([codepen](https://codepen.io/dbjorge/full/WNGyWmd)):

```html
<div role="group">
  <button role="menuitemcheckbox">test element</button>
</div>
```

Motivation: false positive in aria-required-parent rule

<pre><code>
axe-core version: 4.1.1
axe-webdriver, extension or other integration version: accessibility insights for web 0.23.0

Browser and Assistive Technology versions: Microsoft Edge stable (Version 87.0.664.66 (Official build) (64-bit)), no AT in use
</code></pre>

straker:
We've been having some discussion of this in the PR that was opened for it. There's some issues this causes for screen reader so we're determining what should happen.

PR #2737, opened 2021-01-07
feat(aria-required-context): allow 'menuitem' and 'menuitemcheckbox' to be in a 'group'

smockle:
Allows `menuitem` and `menuitemcheckbox` to be in a `group`, per the [ARIA 1.2 spec](https://www.w3.org/TR/wai-aria-1.2/#menuitemcheckbox):

> In order to identify that they are related widgets, authors MUST ensure that menu item checkboxes are owned by an element with role menu or menubar, or by a role group which itself is owned by an element with role menu or menubar.

`group` was omitted from the [ARIA 1.1 spec](https://www.w3.org/TR/wai-aria-1.1/#menuitemcheckbox)—the change was tracked in https://github.com/w3c/aria/issues/737 (issue) and https://github.com/w3c/aria/pull/784 (PR).

Closes issue: #2736


straker:
Thanks for the pr. From reading the spec, they require that the `group` role still be contained within a `menu` role. As the pr is written, this would be valid but shouldn't be:

```html
<div role="group">
  <div role="menuitemcheckbox"></div>
</div>
```

We had to do something similar for our [aria-required-children rule](https://github.com/dequelabs/axe-core/pull/2661) where we would allow the `group` role but still enforce that the children of the `group` were the required children. We should probably do the same here for `aria-required-parent`.

However, before we do anything we should check with @WilcoFiers if we're missing something that would prevent us from doing this.

WilcoFiers:
@smockle It may be in a draft spec, but that doesn't mean it can be used without issues. We're going to need to do some tests on this before we can pull it in. I checked out the ARIA 1.2 menubar example, which shows this being used:
https://w3c.github.io/aria-practices/examples/menubar/menubar-editor.html

In VoiceOver, this creates an issue that the menu that wraps the groups is no longer announced as a menu, and is instead announced as a group. As far as issues go, that's not too bad, so I don't think that's a blocker, but we need to run it past a few more screen readers to make sure none of them choke on this. If they do, we should still add it to axe-core, but have it report as not being widely supported instead.

straker:
Alright, so I did some testing with different screen readers and browsers with the following code:

```html
<ul role="menu">
  <li role="group">
    <button role="menuitem">Group parent</button>
  </li>
  <li role="none">
    <button role="menuitem">None parent</button>
  </li>
</ul>

<ul role="menu">
  <li role="none">
    <button role="menuitem">None parent</button>
  </li>
  <li role="none">
    <button role="menuitem">None parent 2</button>
  </li>
</ul>
```

The second menu was to check what it sounded like without the `group` role.

Here's the results:
* Safari/VO - Reads the `li` with the `role=group` with "Bullet" as part of the name instead of skipping it like it does for the other `li`s with `role=none`. Though I had VO read it as a Menu correctly, so not sure what I did differently.
* Firefox/NVDA - The `group` role prevents the menu from listing the correct number of menuitems. Both the  menuitems in the first `ul` lists them as "1 of 1," meaning it doesn't see two items at all and treats them as separate lists
* IE11/JAWS - Doesn't seem to care that the `group` role is there and treats the selection of the first menuitem as part of the whole menu.
* Chrome/JAWS - Reads the `li` with the `role=group` as "Bullet" instead of skipping it like it does for the other `li`s with `role=none`. Reads the number of items in the menu correctly.

straker:
@WilcoFiers and I talked and everything looks fine. No major problems so we're good to move forward with this. @smockle We still need to add the code to `aria-required-parent-evaluate` that checks that the required parent role still exists past the `group` role.

I believe all we would need to do is add `group` to the https://github.com/dequelabs/axe-core/blob/develop/lib/checks/aria/aria-required-parent-evaluate.js#L24 check only if the role allows `group` as a required parent, so very similar to what we did in [aria-required-children](https://github.com/dequelabs/axe-core/blob/develop/lib/checks/aria/aria-required-children-evaluate.js#L25-L26).

Let me know if you have any questions or need any help.

smockle:
> We still need to add the code to `aria-required-parent-evaluate` that checks that the required parent role still exists past the `group` role.

Thanks, will do. 👍 I appreciate your investigative work related to this!

straker:
@smockle just checking in to see if you have any questions or would like us to take over the changes

smockle:
> @smockle just checking in to see if you have any questions or would like us to take over the changes

Thanks @straker, I’m sorry for the delay on this. I just added two tests in https://github.com/dequelabs/axe-core/pull/2737/commits/bcf15877b4eeb38b00c30f287e2bc150f0df65bd—do those look okay?

(The “should fail” test is failing because I haven’t updated https://github.com/dequelabs/axe-core/blob/develop/lib/checks/aria/aria-required-parent-evaluate.js#L24 yet—I wanted to make sure I could verify the change before making it.)

smockle:
> (The “should fail” test is failing because I haven’t updated https://github.com/dequelabs/axe-core/blob/develop/lib/checks/aria/aria-required-parent-evaluate.js#L24 yet — I wanted to make sure I could verify the change before making it.)

`getMissingContext` in `aria-required-parent-evaluate.js` has now been updated to check the role of a `group`’s parent. Tests should now pass.

straker:
Thanks for those updates! I think it's really close. Looking over the code I believe there's a very subtle bug. If the role does not allow `group` context but there is a `group` role as the parent, I believe the code as written will accept it and continue to check the next parent when it shouldn't.

That is, this should fail but doesn't look like it will with the current code:

```html
<div role="list">
  <div role="group">
    <div role="listitem">should fail</div>
  </div>
</div>
```

smockle:
> …Looking over the code I believe there's a very subtle bug. If the role does not allow `group` context but there is a `group` role as the parent, I believe the code as written will accept it and continue to check the next parent when it shouldn't.

Good catch, thanks! You’re exactly right: I added a test for this in https://github.com/dequelabs/axe-core/pull/2737/commits/06d615bac7ebf3235342601b46bbcdef5ecedf9b, and I’ll work on a fix. 👍

straker (approved):
LGTM. Well done! Thanks for these changes.

straker:
Security Review won't let me approve, but I reviewed for security.

Issue #2570, opened 2020-10-20
Add setup and teardown methods for axe-core


WilcoFiers:
Axe-core needs an official way on how commons like the accessible name computation can be used outside an axe.run() call. One way we might be able to do this is by creating a setup and teardown method. This might look something like this:

```js
axe.setup(document);
const node = document.querySelector('button')
const buttonName = axe.commons.text.accessibleNameVirtual(node)
console.log(buttonName);

axe.teardown();
```

PR #2738, opened 2021-01-07
feat(setup/teardown): add functions to setup and teardown axe-core internal data, deprecate axe._tree

straker:
New functions for setup up and teardown of axe-core internal data `axe._tree` and caches. As we're returning the tree now that is created, we are deprecating `axe._tree` in favor of the return from `axe.setup`

```js
const rootNode = axe.setup();
// use rootNode instead of axe._tree
```

Closes issue: #2570

### DIFF
--- a/lib/core/public/setup.js
+++ b/lib/core/public/setup.js
@@ -1,0 +1,12 @@
+import { getFlattenedTree, getSelectorData } from '../utils';
+
+/**
+ * Setup axe-core so axe.common functions can work properly.
+ * @param {Node} [node=document.documentElement] optional node. NOTE: passing in anything other than body or the documentElement may result in incomplete results.
+ */
+function setup(node) {
+  axe._tree = getFlattenedTree(node);
+  axe._selectorData = getSelectorData(axe._tree);
+}
+
+export default setup;
--- a/lib/core/public/setup.js
+++ b/lib/core/public/setup.js
@@ -5,8 +5,16 @@ import { getFlattenedTree, getSelectorData } from '../utils';
  * @param {Node} [node=document.documentElement] optional node. NOTE: passing in anything other than body or the documentElement may result in incomplete results.
  */
 function setup(node) {
+  if (axe._tree) {
+    throw new Error(
+      'Axe is already setup. Call `axe.teardown()` before calling `axe.setup` again.'
+    );
+  }
+
   axe._tree = getFlattenedTree(node);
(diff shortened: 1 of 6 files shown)


lib/core/public/setup.js
```diff
@@ -0,0 +1,12 @@
+import { getFlattenedTree, getSelectorData } from '../utils';
+
+/**
+ * Setup axe-core so axe.common functions can work properly.
+ * @param {Node} [node=document.documentElement] optional node. NOTE: passing in anything other than body or the documentElement may result in incomplete results.
+ */
+function setup(node) {
```
WilcoFiers, line 7: I think to avoid memory leaks, we should put in an assertion here that makes sure _tree and _selectorData aren't already defined. Have it error if you call setup multiple times in a row without calling teardown.

lib/core/public/teardown.js
```diff
@@ -0,0 +1,20 @@
+import cache from '../base/cache';
+
+/**
+ * Clean up axe-core tree and caches
+ */
+function teardown() {
```
WilcoFiers, line 6: As above, I think we should assert here that _tree and _selectorData are defined. If someone calls teardown if there's nothing to tear down, something has gone wrong and we should let 'm know.

WilcoFiers (changes requested):

lib/core/public/teardown.js
stephenmathieson: I strongly agree that `setup()` should prevent you from calling it multiple times. However, I don't see the harm in calling `teardown()` more than once/when not "setup". Consider the following:

```js
try {
  axe.setup()
  axe.foo()
} catch {
  // Oh well, we tried!
}
axe.teardown()
```

or:

```js
function doSomeAxeThing() {
  // Ensure we empty out any previous state.
  axe.teardown()
  // Do the axe thing.
  return axe.bar()
}
```

straker: Since we're just cleaning up I also see no harm in calling it multiple times. However, since it's not exactly clear that `axe.run` teardowns for you, I'll add a comment so it's understood.

stephenmathieson (approved):

lib/core/public/teardown.js
WilcoFiers: Good point Stephen.

WilcoFiers (approved):

Issue #2729, opened 2021-01-03
Error: Empty string passed to getElementById().


emersonveenstra:
I have a page that axe errors on, the console message says `Empty string passed to getElementById().` and the minified code matches https://github.com/dequelabs/axe-core/blob/develop/lib/rules/color-contrast-matches.js#L73

The page is not public, but I can provide a link if needed

Expectation: axe displays an accessibility report for a page

Actual: axe displays an error

<pre><code>
axe-core version: 4.0.2
axe-webdriver, extension or other integration version: 4.6.0

Browser and Assistive Technology versions
Firefox 85.0b4 (Dev Edition), no assistive technology used.

</code></pre>

WilcoFiers:
@emersonveenstra Thank you for reporting. That's an odd one. We'll be sure to get this fixed ASAP.

straker:
@emersonveenstra as the page is private, would you be able to verify if the latest changes fixes the problem? The latest develop version of axe can be found at https://www.npmjs.com/package/axe-core/v/next. If you add it to the page and run `axe.run()` hopefully the error won't show anymore.


```js
const script = document.createElement('script');
script.src = 'https://unpkg.com/axe-core@next';
document.body.appendChild(script);
```

```js
await axe.run();
```

emersonveenstra:
Works great, thanks!

straker:
Perfect, thanks for letting us know. Closing as the issue seems to be fixed in latest version.

PR #2739, opened 2021-01-07
fix(color-contrast-matches): do not pass empty string to getElementById

straker:
In Firefox, a warning is logged when passing an empty string to `document.getElementById`. This changes the color-contrast-matches code so we don't do that anymore for implicit `labels`.

image

Closes issue: #2729


WilcoFiers (changes requested):
Please add a test

WilcoFiers (approved):

Issue #2651, opened 2020-11-19
Should an aside in main fail axe?


WilcoFiers:
We keep getting feedback about failing this scenario:

```html
<main>
  <p>Lorem's ipsum</p>
  <aside>Some stuff</aside>
</main>
```

Axe-core goes by the ARIA standards on this, but lets do some actual testing to figure out if this creates problems.

WilcoFiers:
Did a little testing. Here's my code:

```html
  <main>
    <header>Header</header>
    <footer>Footer</footer>
    <aside>Aside</aside>
  </main>
```

- Chrome sees `main` and `complementary`
- Safari sees `main`, `banner` and `complementary`
- Firefox sees `main` and `complementary`

Not sure why Safari also treats considers the `heading` to be a landmark. If I put the `footer` outside the `main` it does pick it up. That seems like a bug in Safari.

scottaohara:
would agree that safari is incorrect here.

WilcoFiers:
I'm leaning towards creating an exception for `aside` in this rule. The `landmark-unique` rule will still call it out if someone uses multiple `aside` elements in the page, except if they give it a unique name, which I think also addresses the concern here.

padmavemulapati:
Verified with the latest axe-core develop branch code base(dated 12th jan 2021):
rule:`landmark-complementary-is-top-level` verified with the pass and fail criteria validated to allow `aside` inside `main`
Test snippet for fail:
`<nav><div role="complementary" id="target"></div></nav>
 ` <!--should return true if the complementary landmark is in main landmark-->`
  `<main><div role="complementary" id="target1"></div></main>`
  `<nav>`
    `<div role="complementary">`
     ` <p>This complementary landmark is in a main landmark</p>`
    `</div>`
    `</nav>`

![image](https://user-images.githubusercontent.com/47409579/104331156-c6538000-5514-11eb-8d0b-f68791fa5f98.png)

Test snippet for pass:
`<div role="main">`
      `<div role="complementary">`
       ` <p>This div has role complementary></p>`
      `</div>`
    `</div>`

![image](https://user-images.githubusercontent.com/47409579/104331475-1d595500-5515-11eb-8d53-5edeb1fbc213.png)

PR #2740, opened 2021-01-07
fix(landmark-complementary-is-top-level): allow aside inside main

straker:
Closes issue: #2651

### DIFF
--- a/lib/checks/keyboard/landmark-is-top-level-evaluate.js
+++ b/lib/checks/keyboard/landmark-is-top-level-evaluate.js
@@ -5,17 +5,22 @@ import { getComposedParent } from '../../commons/dom';
 function landmarkIsTopLevelEvaluate(node) {
   var landmarks = getAriaRolesByType('landmark');
   var parent = getComposedParent(node);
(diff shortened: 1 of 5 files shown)


lib/checks/keyboard/landmark-is-top-level-evaluate.js
```diff
@@ -5,17 +5,22 @@ import { getComposedParent } from '../../commons/dom';
 function landmarkIsTopLevelEvaluate(node) {
   var landmarks = getAriaRolesByType('landmark');
   var parent = getComposedParent(node);
+  var nodeRole = node.getAttribute('role') || implicitRole(node);
```
WilcoFiers, line 8: If we're going to mess with this, why not use aria.getRole?

WilcoFiers (approved):

lib/checks/keyboard/landmark-is-top-level-evaluate.js
straker: I wondered that myself but didn't know if there was a specific reason why we didn't before.

WilcoFiers (approved):

Issue #2734, opened 2021-01-07
Convert rules to use virtual node


straker:
- [x] aria/aria-allowed-attr
- [x] aria/aria-allowed-role
- [x] aria/aria-valid-attr
- [x] aria/aria-valid-attr-value
- [x] aria/aria-errormessage
- [x] aria/aria-required-attr
- [x] aria/fallbackrole
- [x] aria/invalidrole
- [x] aria/abstractrole
- [x] aria/unsupportedrole
- [x] keyboard/tabindex
- [x] language/has-lang
- [x] language/valid-lang
- [x] mobile/meta-viewport checks
- [x] navigation/meta-refresh
- [ ] scope-attr-valid rule (for html5scope check, if we don't have document assume true)

PR #2741, opened 2021-01-07
feat(virtual-node): add attrNames property which returns list of attribute names

straker:
The `getNodeAttributes` function does not support virtual nodes. Deprecating `getNodeAttributes` in favor of a property on the virtual node that returns an array of attribute names. From the list of names we can call `vNode.attr(name)` to get the value.

Part of issue #2734

### DIFF
--- a/test/core/base/virtual-node/abstract-virtual-node.js
+++ b/test/core/base/virtual-node/abstract-virtual-node.js
@@ -14,6 +14,17 @@ describe('AbstractVirtualNode', function() {
     assert.throws(fn);
   });
 
+  it('should throw an error when accessing attrNames', function() {
+    function fn() {
+      var abstractNode = new axe.AbstractVirtualNode();
+      if (abstractNode.attrNames.length) {
+        return;
+      }
+    }
+
+    assert.throws(fn);
+  });
+
   it('should throw an error when accessing hasClass', function() {
     function fn() {
       var abstractNode = new axe.AbstractVirtualNode();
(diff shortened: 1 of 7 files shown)


WilcoFiers (approved):

test/core/base/virtual-node/abstract-virtual-node.js
```diff
@@ -14,6 +14,17 @@ describe('AbstractVirtualNode', function() {
...
+  it('should throw an error when accessing attrNames', function() {
+    function fn() {
+      var abstractNode = new axe.AbstractVirtualNode();
+      if (abstractNode.attrNames.length) {
+        return;
+      }
+    }
+
+    assert.throws(fn);
+  });
```
stephenmathieson, line 26: This test seems to be testing that JavaScript property access works as expected. Is it any different than doing this:

```
function fn(){
  var a = {}
  a.b.c.d.e.f.g.h.length
  return 7
}

assert.throws(fn)
```

straker: Sorry I don't quite understand. Are you saying you'd prefer not having the `.length` part or that the test itself isn't useful?

stephenmathieson: Assuming I'm following the code correctly, the `attrNames` getter returns nothing (`undefined`), so doing `undefined.length` obviously throws. I think asserting that the exception raised by `get attrNames(){}` matches what we expect. For example:

```suggestion
  it('should throw an error when accessing attrNames', function() {
    function fn() {
      var abstractNode = new axe.AbstractVirtualNode();
      return abstractNode.attrNames;
    }

    assert.throws(fn, 'VirtualNode class must have an "attrNames" property');
  });
```

stephenmathieson: This is not a "big deal", nor a blocker. Feel free to merge since Wilco has already approved 👍

straker: Its good. I agree that with the `.length` part it may not through because of the reason we expect.

WilcoFiers (approved):

PR #2742, opened 2021-01-07
feat(utils/matches): support selectors level 4 :not and :is

straker:
Been wanting to do this for awhile now so decided to take some time to do it. We can now support the Selectors Level 4 [:is pseudo selector](https://drafts.csswg.org/selectors-4/#matches) (which allows us to simplify some heading selecting) and allows [:not to support more than one selector](https://drafts.csswg.org/selectors-4/#negation), which also helps us shorten and simplify selectors.

Since we use `utils/matches` for everything and we don't go through the DOM selector API, we don't have to worry about browser support since we match everything virtually.

### DIFF
--- a/test/core/utils/matches.js
+++ b/test/core/utils/matches.js
@@ -214,6 +214,13 @@ describe('utils.matches', function() {
         var virtualNode = queryFixture('<h1 id="target">foo</h1>');
         assert.isFalse(matches(virtualNode, 'h1:is([class])'));
       });
+
+      it('returns false if :is matches none of the selectors', function() {
+        var virtualNode = queryFixture('<h1 id="target">foo</h1>');
+        assert.isFalse(
+          matches(virtualNode, 'h1:is([class], span, #foo, .bar)')
+        );
+      });
     });
   });
 
--- a/test/core/utils/matches.js
+++ b/test/core/utils/matches.js
@@ -175,7 +175,7 @@ describe('utils.matches', function() {
 
       it('returns false if :not matches one of the selectors', function() {
         var virtualNode = queryFixture('<h1 id="target">foo</h1>');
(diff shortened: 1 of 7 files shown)


test/core/utils/matches.js
```diff
@@ -132,38 +132,87 @@ describe('utils.matches', function() {
...
+      });
+
+      it('returns false if :not matches element', function() {
+        var virtualNode = queryFixture('<h1 id="target">foo</h1>');
+        assert.isFalse(matches(virtualNode, 'h1:not([id])'));
+      });
+
+      it('returns false if :not matches one of the selectors', function() {
+        var virtualNode = queryFixture('<h1 id="target">foo</h1>');
+        assert.isFalse(matches(virtualNode, 'h1:not([role=heading], h1)'));
```
WilcoFiers, line 178: You want an h1 that's not an h1? xD How about:

```suggestion
        assert.isFalse(matches(virtualNode, 'h1:not([role=heading], [id])'));
```

test/core/utils/matches.js
```diff
@@ -132,38 +132,87 @@ describe('utils.matches', function() {
...
+        var virtualNode = queryFixture('<h1 id="target">foo</h1>');
+        assert.isTrue(matches(virtualNode, ':is([role=heading], h1)'));
+      });
+
+      it('returns true if :is matches complex selector', function() {
+        var virtualNode = queryFixture('<div><h1 id="target">foo</h1></div>');
+        assert.isTrue(matches(virtualNode, 'h1:is(div > #target)'));
+      });
+
+      it('returns false if :is does not match element', function() {
```
WilcoFiers, line 213: Can you add a test showing it can return false for multiple selectors, where all are false?

WilcoFiers (changes requested):
Minor test changes

WilcoFiers (approved):

PR #2744, opened 2021-01-08
test: Add tests for remote ACT repo

WilcoFiers:
First attempt at including ACT test cases into axe-core. This is designed to help us keep axe-core better in sync with ACT rules. This is just a first attempt, didn't want to complicate this too much by trying to solve for multiple things at once. Here is what isn't in here, that the ACT site says we can do:

- video & audio rules, axe-core only reports them as incomplete, they're not interesting
- ARIA required context role, axe-core ignores divs with no role, it should only ignore elements with role=none
-  "Visible label is part of accessible name", need to wait for content to load
- "role attribute has valid value", axe-core fails fallback roles


.gitmodules
```diff
@@ -0,0 +1,5 @@
+[submodule "act-rules-repo"]
```
straker, line 1: I question the need to make this a submodule. Working with submodules adds complexity that I'm not sure we want (e.g. you now have to have the `--recursive` flag to clone the repo). Is there a specific reason why we couldn't npm install the repo on the master branch and use the needed file from a path into node-modules?

straker (commented):
This is a really cool idea. One major problem I forsee is that debugging these tests are going to be difficult. Because the processor is working with a single file, the output is a single file of all test cases. This forces us to replace the current html content with the new content we need to test. If we need to get in the middle and debug a failing test case, we have no way to easily do that in the browser dev tools. 

It would be easier to debug if each act-rule entry in the `act-rules.json` file were it's own json file. This would allow us to create a 1:1 map of the test case to a single file we can debug (just like we do for our integration tests). This would also prevent the need of recreating the DOM each test as we could make the contents of the produced HTML file the contents of the test case.

.gitmodules
WilcoFiers: I haven't quite figure out what I want with this yet. I'm not sure these tests should run on every PR, since the test suite changes, a new ACT test case could become a blocker to PRs. On the other hand, if I'm updating a rule that has ACT test cases, we should automatically test that our changes didn't break the ACT implementation.

We'll discuss on Monday. I agree with you that if we decide to always run these, NPM is probably easier than gitmodules.

test/act-mapping/iamge-button-acc-name.json
```diff
@@ -0,0 +1,5 @@
+{
```
straker, line 1: This file has a typo in the name

test/act-mapping/preprocessor.js
```diff
@@ -0,0 +1,47 @@
...
+var testcases = testcaseContent.testcases;
+
+createActPreprocessor.$inject = ['logger'];
+// Simple Karma preprocessor, takes JSON, outputs a JS file that can run tests
+function createActPreprocessor(logger) {
+  var log = logger.create('preprocessor.act');
+
+  return function(actRuleJson, file, done) {
+    try {
+      log.debug('Processing "%s".', file.originalPath);
```
straker, line 17: This should update the name to be a `.js` file. Right now it saves it as a `.json` file and karma warns for all files saying it defaults to `.js` (thus why it still works).

> 01 02 2021 09:09:10.033:WARN [middleware:karma]: Unable to determine file type from the file extension, defaulting to js.
  To silence the warning specify a valid type for /Users/stevenlambert/deque/axe-core/test/act-mapping/form-field-acc-name.json in the configuration file.
  See http://karma-runner.github.io/latest/config/files.html

```js
file.path = file.originalPath.replace(/\.json$/, '.js');
```

test/act-mapping/karma.config.js
```diff
@@ -0,0 +1,36 @@
...
+    files: [
+      'axe.js',
+      'test/act-mapping/*.json',
+      {
+        pattern: 'node_modules/act-rules.github.io/test-assets/**/*',
+        included: false,
+        served: true
+      }
+    ],
+    browsers: ['ChromeHeadless'],
```
straker, line 23: Instead of changing the file path in the runner script we should use a proxy to serve the assets at the correct path

```json
proxies: {
      '/test-assets': '/base/node_modules/act-rules.github.io/test-assets',
    },
```

test/act-mapping/runner.js
```diff
@@ -0,0 +1,80 @@
...
+          assertResultsCorrect(testcase, result);
+          done();
+        })
+        .catch(done);
+    });
+  }
+
+  function exampleSetup(html) {
+    var parser = new DOMParser();
+    html = html.replace(
```
straker, line 48: Should remove this in favor of a proxy.

straker (dismissed):
When trying to debug the test cases in the browser, something redirects the page to github.com, making it impossible to debug using karma debug command:

```command
 npm run test:act -- --no-single-run --browsers=Chrome
```

(could add that as its own package command: `test:act:debug`)

straker (approved):

straker (approved):

test/act-mapping/autocomplete-valid.json
```diff
@@ -0,0 +1,5 @@
+{
```
soh-476, line 1: #2803

soh-476:
ok

PR #2745, opened 2021-01-11
docs: fix frame-title

straker:
This keeps happening every time we do inline suggestion commits to the rule/check metadata files. Since the doc doesn't get rebuilt, we have to rebuild it ourselves and repush. 

Would be nice if we had a github action or something that just rebuilt the file for us and added it to the branch or something so this stops happening.


stephenmathieson:
> Would be nice if we had a github action or something that just rebuilt the file for us and added it to the branch or something so this stops happening.

This doesn't seem super difficult. I think this would probably suffice:

```yml

name: Build docs thingy

on:
  push:
    branches:
      - develop

jobs:
  build:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
      - name: Cache node_modules
        uses: actions/cache@v2
        with:
          path: ./node_modules
          key: npm-cache-${{ runner.os }}-${{ hashFiles('./package-lock.json') }}
      - uses: actions/setup-node@v1
        with:
          node-version: 14 # I don't remember what version of Node we use here...
      - run: npm install
      - run: npm run docs-thingy
      - uses: stefanzweifel/git-auto-commit-action@v4.2.0
        with:
          commit_message: docs: Update thingy
          commit_options: --no-verify
          file_pattern: docs/thingy.md
```

WilcoFiers (approved):

PR #2753, opened 2021-01-14
docs: consistent dot notation in rule-development docs

jackdclark:
I was just reading the docs for writing custom rules & noticed an inconsistency in the check properties table


straker (approved):
Thanks for catching this.

straker:
Reviewed for security.

PR #2757, opened 2021-01-15
tests: use new axe.setup method

straker:
Wanted to start updating the tests to stop using `axe._tree` everywhere and use the new `axe.setup` method, starting by updating the testUtils code. However this was a bit of a rabbit hole as to remove `axe._tree` I needed the root node returned by `axe._setup` which was done in `fixtureSetup`. 

However, updating `fixtureSetup` cause code that relied on it returning the fixture element to stop working, so had to update those tests to not use fixture from `fixtureSetup` (most of the time they just needed to use `queryFixture` instead).

I also found that code would use `fixtureSetup` multiple times when really what it wanted was to setup the fixture, do some shadow work, then setup the tree. To resolve these issues I created a new function to just inject code into the fixture without setting up the tree.

I also noticed in a test that `axe._selectCache` wasn't reset as part of the teardown method so did that in teardown to resolve that problem.

Lastly, I wanted to consolidate some repeated setup code for later by not having every test have to reset the `fixture.innerHTML` itself or call clean up functions every time. So now the beforeEach code will setup the audit to the default state and clean up the fixture element. I also wanted to remove every call of `var fixture = document.querySelector('#fixture')` in every test so that it's more flexible in the future and uses pure functions, so created a new function to just return the fixture element in the testUtils.

Also, as part of this I want to stop using `flatTreeSetup` as `axe.setup` does the exact same thing.

### DIFF
--- a/test/testutils.js
+++ b/test/testutils.js
@@ -17,6 +17,14 @@ var originalAudit = axe._audit;
 var originalRules = axe._audit.rules;
 var originalCommons = (commons = axe.commons);
 
+// add fixture to the body if it's not already
+var fixture = document.getElementById('fixture');
+if (!fixture) {
+  fixture = document.createElement('div');
+  fixture.setAttribute('id', 'fixture');
+  document.body.insertBefore(fixture, document.body.firstChild);
+}
+
 /**
  * Create a check context for mocking/resetting data and relatedNodes in tests
  *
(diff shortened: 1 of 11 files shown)


test/testutils.js
```diff
@@ -96,12 +110,24 @@ testUtils.fixtureSetup = function(content) {
...
+ * Method for injecting content into a fixture and caching
+ * the flattened DOM tree (light and Shadow DOM together)
+ *
+ * @param {String|Node} content Stuff to go into the fixture (html or DOM node)
+ * @return HTMLElement
+ */
+testUtils.fixtureSetup = function(content) {
+  'use strict';
+  testUtils.injectIntoFixture(content);
+  axe.teardown();
```
straker, line 127: Test code that uses loops (like for `input` types) would do so by calling `querySelector` or `fixtureSetup` inside the loop without tearing down. Since the tree changed I figured it was acceptable to call teardown before setting up the tree.

WilcoFiers (approved):

Issue #2684, opened 2020-12-10
Support graphics roles in the aria-roles rule


TwoLaid:
Expectation: The `aria-roles` rule should pass on elements that use graphics roles such as `graphics-document` and `graphics-object`.

Actual: The rule fails when the website contains an element with `role="graphics-document"`

Motivation: As far as I can tell, these roles already have decent support, and are even used in some of the internal axe-core tests, such as https://github.com/dequelabs/axe-core/blob/f2f15377c917d0b6fc27d2181f64ebef26e5a9b0/test/integration/rules/svg-img-alt/svg-img-alt.html or the `svg-img-alt` rule.

<pre><code>
axe-core version: 3.5.5
axe-webdriver, extension or other integration version: n/a

Browser and Assistive Technology versions: macOS Catalina, VoiceOver and Safari

For Tooling issues:
- Node version: 12
- Platform:  macOS
</code></pre>

straker:
Thanks for the issue. We've been [meaning to get them added](https://github.com/dequelabs/axe-core/issues/1189). Would that be something you'd be interested in opening a pr for?

If so, I'd be happy to walk you through it. You'd just add a new file in [lib/standards](https://github.com/dequelabs/axe-core/tree/develop/lib/standards) called `graphicRoles` and then add each [graphic role](https://www.w3.org/TR/graphics-aria-1.0/) as documented by the [aria-roles doc](https://github.com/dequelabs/axe-core/blob/develop/doc/standards-object.md#aria-roles).

TwoLaid:
Yes, I'm interested in contributing! Thank you for the pointers.

WilcoFiers:
Issue resolved, closing the ticket.

Issue #1189, opened 2018-10-10
Add ARIA graphics module roles


WilcoFiers:
Add ARIA graphics module roles to the ARIA roles list as unsupported roles:
https://www.w3.org/TR/graphics-aria-1.0/

PR #2761, opened 2021-01-18
feat(standards): add graphics roles

TwoLaid:
Add support for roles of the WAI-ARIA Graphics Module: https://www.w3.org/TR/graphics-aria-1.0/#introduction

* Adds three new roles: `graphics-document`, `graphics-object`, `graphics-symbol`
* Adjusts existing standards test
* Extends Typescript types for the new roles
* Adds documentations for the new roles

Closes issue: https://github.com/dequelabs/axe-core/issues/2684 and https://github.com/dequelabs/axe-core/issues/1189.


axe.d.ts
```diff
@@ -162,7 +167,7 @@ declare namespace axe {
     unsupported?: boolean;
   }
   interface AriaRoles {
-    type: AriaRolesType | DpubRolesType;
```
straker, line 165: I don't think this is correct. The `type` should be the `type` property from the role, which is just `structure` for each one. So you shouldn't need a new type for this as `AriaRoleTypes` covers strcuture.

straker (changes requested):
Nice work! I'll let @WilcoFiers do a final review on this one to make sure we're not missing something.

axe.d.ts
```diff
@@ -162,7 +167,7 @@ declare namespace axe {
...
+    type: AriaRolesType | DpubRolesType | GraphicsRolesType;
```
straker, line 170: I don't think this is correct. The type should be the type property from the role, which is just structure for each one. So you shouldn't need a new type for this as AriaRoleTypes covers strcuture.

TwoLaid: Yes, you're absolutely right. I misunderstood that bit. Removing it now.

WilcoFiers (approved):

WilcoFiers:
@TwoLaid Thank you so much for the contribution! Much appreciated.

WilcoFiers:
Security app isn't responding. -- I reviewed this PR for security.

Issue #2687, opened 2020-12-14
iframe with focusable content does not have negative tabindex


WilcoFiers:
We should get an implementation of this rule going fairly soon:
https://github.com/act-rules/act-rules.github.io/pull/1515/files

## Violations

```html
<iframe tabindex="-1" srcdoc="<a href='/'>Home</a>"></iframe>
<iframe tabindex="-1" srcdoc="<button>Save</button>"></iframe>
```

## Passes:

```html
<iframe tabindex="-1" srcdoc="<h1>Hello world</h1>"></iframe>
<iframe tabindex="-1" srcdoc="<a href='/' tabindex='-1'>Home</a>"></iframe>
<iframe tabindex="-1" width="1" height="1" srcdoc="<a href='/'>Home</a>"></iframe>
```

padmavemulapati:
Verified with the latest axe-core- develop branch code base , I am unable to see the anything related to `frame-focusable` rule
![image](https://user-images.githubusercontent.com/47409579/107651353-6bdd4900-6ca5-11eb-8d11-b1b032b463eb.png)

straker:
@padmavemulapati It looks like you triggered the `frame-tested` rule. The rule only works when the iframes also have axe-core in them. Could you try adding axe-core to each of the iframes and testing again?

padmavemulapati:
Validated with the latest axe-core code base - seeing `frame-focusible-content` issue- when
``` index.html
<iframe title="frame" src="frame.html" tabindex="-1"></iframe>
<script src="/axe.js"></script>
```
```frame.html
<button>Save</button>
<script src="/axe.js"></script>
```

![image](https://user-images.githubusercontent.com/47409579/116051098-dcd1b180-a695-11eb-80d0-a929538c1532.png)

PR #2762, opened 2021-01-18
feat(frame-focusable): new rule to detect focusable content in iframes with tabindex=-1

straker:
Closes issue: #2687


lib/checks/keyboard/frame-focusable-after.js
```diff
@@ -0,0 +1,21 @@
...
+  results.forEach(result => {
+    const element = result.node._element;
+    const res = resultMap.get(element) || true;
+    resultMap.set(element, res && result.result);
+  });
+
+  const finalResults = [];
+  const entries = resultMap.entries();
+  for (const entry of entries) {
+    const [element, res] = entry;
```
WilcoFiers, line 12: Don't think we need to declare those other variables.

```suggestion
  for (const [element, res] of resultMap.entries()) {
```

lib/checks/keyboard/frame-focusable-after.js
```diff
@@ -0,0 +1,21 @@
+function frameFocusableAfter(results) {
```
WilcoFiers, line 1: I'm very confused by this function.

lib/core/utils/collect-results-from-frames.js
```diff
@@ -25,12 +25,19 @@ function collectResultsFromFrames(
 
   // Tell each axe running in each frame to collect results
   frames.forEach(frame => {
+    const tabindex = frame.node.getAttribute('tabindex');
+    const focusable = tabindex ? parseInt(tabindex) >= 0 : true;
```
WilcoFiers, line 29: This needs to account for invalid tabindex like `tabindex="foo"`.

lib/core/utils/collect-results-from-frames.js
```diff
@@ -25,12 +25,19 @@ function collectResultsFromFrames(
...
     var params = {
       options: options,
       command: command,
       parameter: parameter,
       context: {
         initiator: false,
+
+        // if any parent iframe is not focusable then the entire
+        // iframe stack is not focusable
+        focusable: context.focusable === false ? false : focusable,
```
WilcoFiers, line 40: Better to have the code explain itself then to add comments. We can solve this by changing the variable name.

```suggestion
        focusable: parentFrameContext.focusable === false ? false : focusable,
```

WilcoFiers (dismissed):
Couple comments, but I quite like the approach. Lets go with this.

PR #2764, opened 2021-01-18
feat(aria-allowed-attr): add ARIA 1.2 prohibited attrs check

straker:
Aira 1.2 added [Prohibited States and Props](https://www.w3.org/TR/wai-aria-1.2/#prohibitedattributes) which prevent certain ARIA attributes from being allowed on certain roles. Implementing this as a prerequisite of #1575

Related discussions:
- https://github.com/w3c/aria/issues/756
- https://github.com/w3c/aria/pull/985

In another PR, we can make it so in `implicitRole` any html element without a role returns `generic`, which prohibits `aria-label` and `aria-labelledby`, thus fixing #1575.

From Scott O'Hara on asking about HTML elements with no role mapping:

> no mapping = generic
> generic doesn't allow naming


WilcoFiers (approved):

veselints:
@straker why the latest official release of axe-core contains a rule that is part of the WAI-ARIA 1.2 recommendation? That is still a candidate. I believe the official releases should be based only on official specs and recommendations. I know that the 1.2 rec is an improvement of the 1.1 and soon or later the above should be part of axe-core, but the rules from a candidate should go to a beta/dev channel of the package and not in the official release.

straker:
@veselints axe-core contains best-practice rules and checks that aren't part of the ARIA spec (such as aria-dialog-name), as well as rules that inform the user of problems they might not be aware of (such as this one and presentation-role-conflict). We [test current screen reader support](https://github.com/dequelabs/axe-core/pull/2712#issuecomment-759732976) and will add new rules and future ARIA spec concept based on what screen readers do today.

PR #2766, opened 2021-01-19
fix(aria-allowed-attr): error when generic elements use aria-label and aria-labelledy

straker:
Tested the following DOM in different browsers:

```html
<div aria-label="my label">foo</div>
<span aria-label="my label">foo</span>
<i aria-label="my label">foo</i>
<p aria-label="my label">foo</p>

<div id="label">my label</div>
<div aria-labelledby="label">foo</div>
<span aria-labelledby="label">foo</span>
<i aria-labelledby="label">foo</i>
<p aria-labelledby="label">foo</p>
```

No browser / screen reader combination read the `aria-label` or the `aria-labelledby` attributes. All names returned as "foo"

**Note:** The "proper" way to handle this would be if axe-core returned `generic` instead of `null` for nodes without a role. However, we are not set up to do that and It would require going through all use cases where we check for `null` role and changing it for `generic`. I'm not sure we want to do that just yet, but we should at least add a tech debt ticket to do so if this is merged.

Closes issue: #1575


test/commons/standards/get-aria-roles-by-type.js
```diff
@@ -1,59 +1,32 @@
 describe('standards.getAriaRolesByType', function() {
   var getAriaRolesByType = axe.commons.standards.getAriaRolesByType;
 
-  before(function() {
-    axe._load({});
-  });
+  it('should return a list of role names by type', function() {
```
straker, line 4: I got tired of updating this test when adding new roles. This should allow us to not have it change so often.

lib/checks/aria/aria-prohibited-attr-evaluate.js
```diff
@@ -27,12 +27,9 @@ import standards from '../../standards';
...
 function ariaProhibitedAttrEvaluate(node, options, virtualNode) {
   const prohibited = [];
-  const role = getRole(virtualNode);
-
-  if (!role) {
-    return false;
-  }
 
+  // From Scott O'Hara on asking about HTML elements with no role mapping: no mapping = generic
+  const role = getRole(virtualNode) || 'generic';
```
WilcoFiers, line 32: I'm not so sure I agree with that suggestion of Scott, nor with the implication of it. I don't think we should call every use of aria-label on an element we don't have a role for as a failure of SC 4.1.2. That seems a bit of a stretch to me. A best practice might be more appropriate. I'm going to get some more Deque folks to weigh in on that one first.

test/checks/aria/aria-prohibited-attr.js
```diff
@@ -1,4 +1,4 @@
-describe.only('aria-prohibited-attr', function() {
+describe('aria-prohibited-attr', function() {
```
WilcoFiers, line 1: Huh, odd.

WilcoFiers (changes requested):

test/integration/rules/aria-allowed-attr/failures.html
```diff
@@ -20,4 +20,6 @@
 <div role="subscript" aria-labelledby="value" id="fail20">fail</div>
 <div role="superscript" aria-label="value" id="fail21">fail</div>
 <div role="superscript" aria-labelledby="value" id="fail22">fail</div>
+<div aria-label="value" id="fail23"></div>
+<div aria-labelledby="value" id="fail24"></div>
```
WilcoFiers, line 24: Add some incomplete examples too please.

lib/checks/aria/aria-prohibited-attr-evaluate.js
```diff
@@ -27,14 +28,12 @@ import standards from '../../standards';
...
-  const role = getRole(virtualNode);
-
-  if (!role) {
-    return false;
-  }
 
+  const role = getRole(virtualNode);
   const attrs = virtualNode.attrNames;
-  const prohibitedAttrs = standards.ariaRoles[role].prohibitedAttrs;
+  const prohibitedAttrs = role
```
WilcoFiers, line 34: Apologies for not thinking of this sooner, but I think this should check if this is a "null" attribute. Don't think this should complain about `<div aria-label=" " aria-labelledby="doesNotExist">`.

WilcoFiers (changes requested):

lib/checks/aria/aria-prohibited-attr-evaluate.js
```diff
@@ -23,32 +24,42 @@ import standards from '../../standards';
...
     this.data(prohibited);
+
+    // aria-label and aria-labelledy are not allowed on elements
+    // without roles, but if the element has text content then
+    // we will return undefined as a screen reader will read
+    // the text content
+    if (!role) {
+      if (sanitize(subtreeText(virtualNode)) !== '') {
+        return undefined;
+      }
```
WilcoFiers, line 60:
```suggestion
    if (!role && sanitize(subtreeText(virtualNode)) !== '') {
      return undefined;
```

WilcoFiers (changes requested):
Very minor tweak.

.circleci/config.yml
```diff
@@ -23,14 +23,14 @@ set_npm_auth: &set_npm_auth
 restore_dependency_cache_unix: &restore_dependency_cache_unix
   restore_cache:
     keys:
-      - v8-cache-unix-{{ checksum "package.json" }}
-      - v8-cache-unix-
+      - v9-cache-unix-{{ checksum "package-lock.json" }}
```
straker, line 26: In order to get the tests working again we had to update our version of chromedriver. However, just updating the `packge-lock.json` file didn't work as we checksum the `package.json` file. This also means security updates that just modify the package-lock file do not cause our builds to use the latest versions.

.circleci/config.yml
```diff
@@ -96,10 +96,6 @@ jobs:
           export PATH=/c/tools/selenium:$PATH
           echo $PATH
       - run: npm run test:integration:ie
-      # test examples
```
straker, line 99: Removed as we did in master as windows / chrome was broken.

WilcoFiers (approved):

Issue #2558, opened 2020-10-14
Allow aria-required on fieldset?


WilcoFiers:
This question came in through twitter:

> https://twitter.com/clcorde/status/1316048183299112967
> 
> Using aria-required attr on `<fieldset>`
> `<fieldset aria-required="true">` AXE is firing an error: aria attribute not allowed. I looked at the technical documentation and it seems to be supported for radiogroup role; is this contradictory?

The answer is that `fieldset` maps to the `group` role, which does not have aria-required. When I tested this out though, I couldn't get aria-required to work on fieldsets at all, not even by giving it role="radiogroup". I only tested with Safari+VO. Maybe I'm missing something, but it looks like there's a problem with the way axe-core does aria-required. We'll need to do some research. Here's the code I tried:

```html
<fielset role="radiogroup" aria-required="true">
	<legend>Favorite H. Jon Benjamin show</legend>
	<input type="radio" id="bob" /><label for="bob">Bob's burgers</label>
	<input type="radio" id="archer" /><label for="archer">Archer</label>
</fielset>
```

straker:
Tested this and only JAWS/IE and JAWS/Chrome report that the radio buttons are required. NVDA/Firefox, VO/Safari, and Android Voice Assistant/Chrome do not report required.

straker:
We can add an `unsupportedAttrs` property to the `aria-roles.js` file that flags certain combination of roles/attributes as unsupported due to lack of browser support. Make it's own check and add to the `aria-allowed-attr` rule.

straker:
Closing as this is just a single use case with a lot of special logic behind it

riteshjagga:
We need this too, reason being fieldset gives an ability to disable all the radio buttons using the disabled attribute.

For a group of radio buttons the **required** word should be announced one by the screen reader which is possible only if we allow `aria-required` on a `fieldset`. Adding `aria-required` on every radio button will announce it every time radio button is focused/selected which could cause confusion.

So, I think both `role="radiogroup"` and `aria-required` should be allowed.

straker:
@riteshjagga you should be able to add `role="radiogroup"` to the fieldset to be able to add the `aria-required` attribute. 

```html
<fieldset role="radiogroup" aria-required="true">
  <!-- Radio button 1 -->
  <!-- Radio button 2 -->
  <!-- Radio button 3 -->
</fieldset>
```

riteshjagga:
[...]

PR #2767, opened 2021-01-19
fix(aria-allowed-attr): flag attributes that are not supported with certain roles

straker:
I initially thought to make this it's own check, but the `aria-unsupported-attrs` check made the most sense in the end.

Closes issue: #2558


lib/checks/aria/aria-unsupported-attr-evaluate.js
```diff
@@ -1,7 +1,36 @@
...
+import { validateAttr, getRole } from '../../commons/aria';
 import standards from '../../standards';
 import matches from '../../commons/matches';
 
+/**
+ * Check that an element does not use any unsupported ARIA attributes for it's particular role.
+ * @param {VirtualNode} virtualNode
+ * @return {String[]}
+ */
+function getUnsupportedRoleAttrs(virtualNode) {
```
WilcoFiers, line 10: Kinda think this function should be on commons.aria instead.

test/integration/rules/aria-allowed-attr/failures.html
```diff
@@ -20,4 +20,5 @@
 <div role="subscript" aria-labelledby="value" id="fail20">fail</div>
 <div role="superscript" aria-label="value" id="fail21">fail</div>
 <div role="superscript" aria-labelledby="value" id="fail22">fail</div>
+<fieldset aria-required="false" id="fail23">fail</fieldset>
```
WilcoFiers, line 23: I think we're a little too hasty with this. This particular example will definitely not cause accessibility issues. aria-required=false does nothing. Even with aria-required=true... if there is some other indication that the element is required, I don't think we can call that out as a failure of WCAG TBH.

I feel like I should have spotted that a little sooner, but I think this is something that needs review at best, and I would argue only if it's aria-required=true, and the form controls in the fieldset don't have required or aria-required.

WilcoFiers (changes requested):

lib/standards/aria-roles.js
```diff
@@ -479,6 +479,7 @@ const ariaRoles = {
       'aria-expanded',
       'aria-orientation'
     ],
+    unsupportedAttrs: ['aria-required'],
```
WilcoFiers, line 482: Are you sure this doesn't work on any radiogroup? I only tested radiogroup on a fieldset with VO. I'd like to have more test data before we make this change.

lib/commons/aria/get-unsupported-role-attrs.js
```diff
@@ -0,0 +1,60 @@
...
+      ).length;
+
+      if (
+        role === 'radiogroup' &&
+        vNode.attr('aria-required') === 'true' &&
+        !requiredDescendants
+      ) {
+        return true;
+      }
+    }
```
WilcoFiers, line 44: This seems like it should be its own function somewhere, not embedded into this one. Sticking this here interrupts the reading flow. Maybe just put this below his function?

WilcoFiers (changes requested):

lib/commons/aria/get-unsupported-role-attrs.js
straker: This is in the same code style as https://github.com/dequelabs/axe-core/blob/develop/lib/checks/aria/aria-valid-attr-value-evaluate.js#L39. Should we add a tech debt ticket to update that one?

straker:
Closing as this is just a single use case with a lot of special logic behind it

Issue #1491, opened 2019-04-10
Consider using 'should' instead of 'must' in help text for best practices


mfairchild365:
I’d like to propose changing “must” to “should” in best practice rule help text. For example “Document must have one main landmark” would be “Document should have one main landmark”.

Motivation: This would better align with actual WCAG conformance requirements, the way Deque University refers to best practices, and RFC 2119.

WilcoFiers:
@mfairchild365 I think that's a good suggestion. I'm meeting with Melanie, Glenda and Sujasree to discuss aligning the descriptions of rules with those of the Deque Way. I'll take this on board as input.

dylanb:
No, this is a bad suggestion. If you want to meet the best practice you MUST do what it requires. The best practice itself is a SHOULD but if your decide to meet it then the requirements are clear.

melaniephilipp:
I was just going to file this same issue but discovered this open issue. As of this time, some axe best practices say must and some say should, so they aren't consistent within axe.  Also, in DQU and the DQWay we use the MUST, MAY, SHOULD terminology according to the RFC 2119 standard with respect to WCAG requirements. I realize the axe rules don't capitalize those terms and the analogy isn't perfect, but... the difference in use of "must" and "should" between axe and DQU and DQWay and Assure and WCAG is confusing to say the least.

WilcoFiers:
There are inconsistencies in these messages. In my opinion, the correct word to use in these messages is "must" for the reason Dylan said. If you want the issue to be resolved, you must resolve the thing.

melaniephilipp:
@goodwitch looping you in on this issue to get your thoughts.

goodwitch:
Sounds like we need to add this to the Jedi Council agenda.  @paulbohman needs to be included too.  See the DQU usages of MUST and SHOULD here https://dequeuniversity.com/checklist

WilcoFiers:
OK so RFC 2119 has nothing to do with this ticket. I agree with Dylan that if we treat these as requirements, they are a must. I dug into this with Glenda and Melanie this week, and what I found much more of a compelling reason to change this is to consider the [MuSCoW prioritisation method](https://en.wikipedia.org/wiki/MoSCoW_method). Best practices issues are not things that should block a release, whereas WCAG A and AA issues should do. Therefore it make sense for WCAG rules to not use the word "should" and for BPs to not use the word "must".

padmavemulapati:
[...]

PR #2770, opened 2021-01-20
fix(metadata): consistent use of 'must' and 'should'

WilcoFiers:
Ensure that best practices never use the word "must" in its description / help text, and that other rules don't use the word "should".

Closes issue: #1491

### DIFF
--- a/build/tasks/validate.js
+++ b/build/tasks/validate.js
@@ -235,20 +235,27 @@ function createSchemas() {
   return schemas;
 }
 
-function validateFiles(grunt, files, schema) {
+function validateFiles(grunt, files, schema, type) {
   var valid = true;
   files.forEach(function(f) {
     f.src.forEach(function(pathArg) {
       var file = grunt.file.readJSON(pathArg);
       file._path = pathArg;
       var result = revalidator(file, schema);
-
       if (!result.valid) {
         result.errors.forEach(function(err) {
           grunt.log.error(pathArg, err.property + ' ' + err.message);
         });
         valid = false;
-      } else {
+      }
+
+      const ruleIssues = type === 'rule' ? validateRule(file) : [];
+      if (ruleIssues.length > 0) {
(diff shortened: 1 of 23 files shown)


stephenmathieson:
Is this a "feature"? Seems like a `docs` or `fix` commit to me. 🤔

WilcoFiers:
@stephenmathieson Wouldn't say its docs... but yeah maybe it's more of a fix.

stephenmathieson (commented):
Change LGTM, but @straker should probably take a look since this touches `lib/rules`.

dylanb:
@WilcoFiers this change does not seem right to me. Saying `aria treeitem items should have an accessible name` is not as accurate as saying `aria treeitem items must have an accessible name`

Under what circumstances are these items allowed to not have an accessible name?

WilcoFiers:
@dylanb Yeah that one isn't too obvious. I ran this one by JC before we put it in. If a treeitem is interactive, it would need a name, but if it isn't, it's harder to argue that it does. ARIA says it does, which is why we put it as a best practice. But if you have a static treeitem, it's basically no different than a nested list. We don't require listitems to have an accessible name - ARIA doesn't either.

So that's why we landed on that one. I suppose we could've made it a WCAG rule and restrict it to focusable elements, but then again, we don't fail focusable listitems if they're empty either. This seemed like the most appropriate way to do it. It's maybe not as strict as it could we if we further refined it, but it also gets not false positives.

dylanb:
I am worried about all of them - that one was just an example.

I am concerned that these changes are not being made with usability in mind. We need to remember that 97%+ of the people reading these do not know the nuances that the JC would be concerned about.

I want to put the brakes on this and have a usability discussion about these first and foremost before we make any changes.

dylanb:
Discussed with @WilcoFiers : we are good to go on this

dylanb (approved):

PR #2782, opened 2021-01-27
chore: add parent frame context info

straker:
This is a small portion from #2762 that I'm hoping we can isolate to see if it's the thing causing ie11 to fail the frame-title stuff.


straker:
I've re-run the tests a few times now and haven't gotten any errors.

WilcoFiers (approved):

PR #2785, opened 2021-01-28
feat(frame-focusable-content): new rule to test iframes with tabindex=-1 do not have focusable content

straker:
Pulling out more of #2762 to see if it fixes the ie11 problem that only happens in circle.

**Note:** The `after` method is incomplete on purpose. This is just the minimalistic amount of code necessary in order to get the rule to run.


straker:
For whatever reason the thing that cases frame-test rule to fail is adding a nested iframe to this rules the integration test. It doesn't matter if the `after` method is empty or not. I have no idea why, so we are going to  just get this through for now and add a tech debt ticket to investigate it.

lib/core/utils/collect-results-from-frames.js
```diff
@@ -27,7 +27,12 @@ function collectResultsFromFrames(
   frames.forEach(frame => {
     const tabindex = parseInt(frame.node.getAttribute('tabindex'), 10);
     const focusable = isNaN(tabindex) || tabindex >= 0;
+
     const rect = frame.node.getBoundingClientRect();
+    let width = parseInt(frame.node.getAttribute('width'), 10);
+    let height = parseInt(frame.node.getAttribute('height'), 10);
+    width = isNaN(width) ? rect.width : width;
+    height = isNaN(height) ? rect.height : height;
```
straker, line 35: This prevents margin CSS from affecting the width/height (otherwise in Chrome a width and height of 1 would result in a rect size of 5x5).

test/rule-matches/frame-focusable-content-matches.js
```diff
@@ -0,0 +1,57 @@
+describe('frame-focusable-content-matches', function() {
+  'use strict';
+  var rule;
+
+  beforeEach(function() {
+    rule = axe.utils.getRule('frame-focusable-content');
+  });
+
+  it('returns false for the top-level context', function() {
+    var result = rule.matches(null, null, { initiator: true });
```
WilcoFiers, line 10: This test doesn't prove what you say it does. It could also be false because there's no with/height. I think we need to make sure all props exist in these tests.

WilcoFiers (dismissed):
Minor change request

WilcoFiers (approved):

PR #2789, opened 2021-02-04
fix: add noHtml to axe.configure

straker:


lib/core/utils/dq-element.js
```diff
@@ -39,8 +39,12 @@ function DqElement(element, options, spec) {
...
    * @type {String}
    */
-  this.source =
-    this.spec.source !== undefined ? this.spec.source : getSource(element);
+  // TODO: es-moduels_audit
+  this.source = axe._audit.noHtml
+    ? ''
+    : this.spec.source !== undefined
+    ? this.spec.source
+    : getSource(element);
```
WilcoFiers, line 47:
```suggestion
  if (axe._audit.noHtml) {
    this.source = ''
  } else if (this.spec.source !== undefined) {
    this.source = this.spec.source
  } else {
    this.source = getSource(element)
  }
```

lib/core/utils/dq-element.js
```diff
@@ -39,8 +39,12 @@ function DqElement(element, options, spec) {
    * The generated HTML source code of the element
    * @type {String}
    */
-  this.source =
-    this.spec.source !== undefined ? this.spec.source : getSource(element);
+  // TODO: es-moduels_audit
+  this.source = axe._audit.noHtml
+    ? ''
```
WilcoFiers, line 44: Should this be undefined?

WilcoFiers (changes requested):
I would like some integration tests that prove that 1. this does not send HTML over postMessage, and 2. it does not return html, either from the page level or from frames.

lib/core/utils/dq-element.js
```diff
@@ -39,8 +39,14 @@ function DqElement(element, options, spec) {
    * The generated HTML source code of the element
    * @type {String}
    */
-  this.source =
-    this.spec.source !== undefined ? this.spec.source : getSource(element);
+  // TODO: es-moduels_audit
```
stephenmathieson, line 42: Typo

```suggestion
  // TODO: es-modules_audit
```

lib/core/utils/dq-element.js
stephenmathieson: `undefined` or `null`? IMO `null` is preferred, since it's explicitly not set. NBD tho :)

WilcoFiers (approved):

WilcoFiers:
For the record; this does add a small feature, but it is a feature necessary to address a security fix that will need to be applied to all supported minor versions of axe-core. We are therefor tagging this as a "fix" instead of a "feat".

PR #2803, opened 2021-02-12
fix(utils): remove attributes from source string

straker:
Remove unwanted attributes from source string. Will add all the audit code around it in another pr so this one doesn't get too big.

For the record; this does add a small feature, but it is a feature necessary to address a security fix that will need to be applied to all supported minor versions of axe-core. We are therefor tagging this as a "fix" instead of a "feat".

### DIFF
--- a/lib/core/utils/filter-html-attrs.js
+++ b/lib/core/utils/filter-html-attrs.js
@@ -1,20 +1,137 @@
-function filterHtmlAttrs(htmlString, denylist = {}) {
-  const tagRegex = /<([^>]+?)>/g;
+// grab everything between an opening and closing bracket
+const tagStringRegex = /<([^>]+?)\/?>/g;
 
-  return htmlString.replace(tagRegex, (value, tagStr) => {
-    const [tagName, ...attributes] = tagStr.split(' ');
-    const name = tagName.toLowerCase();
+/* split a tag string by attributes:
+   (\S+)=       group 1: valued attribute name; look for non-whitespace characters followed by the "="" sign
+   ((["'])      start group 2: quoted attribute value; look for an opening quote (single or double)
+   (.*?)        start group 4: non-quoted attribute value; non-greedy match everything (.*?) (the 3rd group is ignored and only used for the back reference \3)
+   \3(?:\s|$)   end group 4: stop when we find a matching end quote and either a whitespace or the end of the string (allowing us to matching escaped quotes that match the opening quote)
+   |[^\s]+))    end of group 2; if no quote then match everything that is not a whitespace
+   |\b(\S+)\b   group 5: non-valued attribute name; look for any remaining non-matched strings and see if they are an attribute name without a value
 
-    if (!denylist[name]) {
-      return value;
+   tested with string: type="text" value="my value again" othervalue=foo ha='bar' ha='\"foo\"' again="\'thing\'" data-thing="foo" thing=bar bar=baz checked thing=bar thing="\"foo\"" thing='\'hello\''
+*/
+const attributeRegex = /(\S+)=((["'])(.*?)\3(?:\s|$)|[^\s]+)|\b(\S+)\b/g;
+
+const attrMap = {
+  for: 'htmlFor',
+  class: 'className'
+};
+
+/**
+ * Match an attribute to a matcher object.
+ * @param {String} nodeName - Name of the current node
+ * @param {Object[]} attributes - Array of node attributes and values
+ * @param {String} value - Value of the current attribute
+ * @param {any} matcher - How to match the value
+ */
+function matches(nodeName, attributes, value, matcher) {
+  switch (typeof matcher) {
+    case 'boolean':
+      return true;
+    case 'string':
+      return matcher === value;
+    case 'object':
+      if (Array.isArray(matcher)) {
+        return matcher.includes(value);
+      }
+
+      // matcher-like object, match all parts
+      let filter = true;
+
+      // match each property
+      if (matcher.nodeName) {
+        filter = filter && matcher.nodeName === nodeName;
+      }
+
+      if (matcher.attributes) {
+        Object.keys(matcher.attributes).forEach(attrName => {
+          const attr = attributes.find(attr => attr.name === attrName);
+          if (!attr) {
+            return false;
+          }
+
+          filter =
+            filter &&
+            matches(
+              nodeName,
+              attributes,
+              attr.value,
+              matcher.attributes[attrName]
+            );
+        });
+      }
+
+      return filter;
+  }
+}
+
+/**
+ * Filter attributes from an html string. This is not to prevent XSS attacks but instead is used to remove attributes which shouldn't appear in the output. As such, we can assume mostly normal attribute output and not attributes purposefully trying to break a parser
+ *
+ * Example:
+ * ```js
+ * // Remove attribute if present regardless of value
+ * axe.utils.filterHtmlAttrs('<div data-attr="foo">my div</div>', { 'data-attr': true });
+ *
+ * // Remove attribute if value matches
+ * axe.utils.filterHtmlAttrs('<div data-attr="foo">my div</div>', { 'data-attr': 'foo' });
+ *
+ * // Remove attribute if value matches list of values
+ * axe.utils.filterHtmlAttrs('<div data-attr="foo">my div</div>', { 'data-attr': ['foo', 'bar'] });
+ *
+ * // Remove attribute if tag matches matcher-like object
+ * axe.utils.filterHtmlAttrs('<div class="foo"><input type="text" class="foo"/></div>', { class: { nodeName: input } });
+ * ```
+ *
+ * @method getRootNode
+ * @memberof axe.utils
+ * @param {String} htmlString - HTML string to remove attributes from.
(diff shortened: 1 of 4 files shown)


lib/core/utils/filter-html-attrs.js
```diff
@@ -0,0 +1,138 @@
...
+ * axe.utils.filterHtmlAttrs('<div class="foo"><input type="text" class="foo"/></div>', { class: { nodeName: input } });
+ * ```
+ *
+ * @method getRootNode
+ * @memberof axe.utils
+ * @param {String} htmlString - HTML string to remove attributes from.
+ * @param {Object} filterAttrs - Attributes to remove and the qualifier of when to remove them.
+ * @returns {String}
+ */
+function filterHtmlAttrs(htmlString, filterAttrs) {
```
WilcoFiers, line 94: I think you should isolate the regex code into its own file / methods, abstract them away, so that the regex logic isn't mixed in with the actual filtering. I think this function should look something like this:

```js
function filterHtmlAttrs(htmlString, attrFilters) {
  let filteredString = htmlString
  const { elements } = parseTags(htmlString)
  elements.forEach(({ nodeName, attributes }) => {
    attributes.forEach((attribute) => {
      if (attributeMatchConfig(nodeName, attribute, attrFilters)) {
        filteredString.replace(attribute.raw, '');
      }
    })
  });

  return filteredString;
}
```

lib/core/utils/filter-html-attrs.js
```diff
@@ -0,0 +1,178 @@
...
+    const attributes = [];
+    while ((groups = attributeRegex.exec(attributeStr))) {
+      attributes.push({
+        // can be either a valued and non-valued attribute
+        name: (groups[1] || groups[5]).toLowerCase(),
+        // double, single, or non-quoted value
+        value: groups[2] || groups[3] || groups[4] || '',
+        raw: groups[0]
+      });
+    }
```
WilcoFiers, line 101: This seems like it should be:

```suggestion
    const attributes = parseAttributes(attributeStr);
```

lib/core/utils/filter-html-attrs.js
```diff
@@ -0,0 +1,178 @@
...
+
+    // no need to filter end tags
+    if (tagStr.charAt(0) === '/') {
+      continue;
+    }
+
+    const index =
+      tagStr.indexOf(' ') !== -1 ? tagStr.indexOf(' ') : tagStr.length;
+    const nodeName = tagStr.substring(0, index).toLowerCase();
+    const attributeStr = tagStr.substring(index + 1).trim();
```
WilcoFiers, line 89: Index of what?

```suggestion
    const nodeNameEnd =
      tagStr.indexOf(' ') !== -1 ? tagStr.indexOf(' ') : tagStr.length;
    const nodeName = tagStr.substring(0, nodeNameEnd).toLowerCase();
    const attributeStr = tagStr.substring(nodeNameEnd + 1).trim();
```

lib/core/utils/filter-html-attrs.js
```diff
@@ -0,0 +1,178 @@
...
+
+/**
+ * Filter out attributes from an an element.
+ * @param {Object} element - element to filter
+ * @param {Object} attrs - Attributes to filter
+ * @returns {Object} element with filtered attributes removed
+ */
+function filterAttributes({ nodeName, attributes, raw }, attrs) {
+  const filteredAttrs = attributes.filter(({ name, value }) => {
+    const matcher = attrs[attrMap[name] ? attrMap[name] : name];
```
WilcoFiers, line 121: This looks a little easier to read to me. Alternatively, maybe first update `name`.

```suggestion
    const matcher = attrs[attrMap[name] || name];
```

lib/core/utils/filter-html-attrs.js
```diff
@@ -0,0 +1,178 @@
...
+
+    if (!matcher) {
+      return true;
+    }
```
WilcoFiers, line 125: This seems like something your matches function should take care of instead.

lib/core/utils/filter-html-attrs.js
```diff
@@ -0,0 +1,178 @@
...
+    });
+  }
+
+  return elements;
+}
+
+/**
+ * Filter out attributes from an an element.
+ * @param {Object} element - element to filter
+ * @param {Object} attrs - Attributes to filter
```
WilcoFiers, line 116: This one variable has 3 different names in this file. `attrs`, `filterAttrs` and `matcher`. Please be consistent.

lib/core/utils/filter-html-attrs.js
```diff
@@ -0,0 +1,178 @@
...
+      // matcher-like object, match all parts
+      let filter = true;
+
+      // match each property
+      if (matcher.nodeName) {
+        filter = filter && matcher.nodeName === nodeName;
+      }
+
+      if (matcher.attributes) {
+        Object.keys(matcher.attributes).forEach(attrName => {
```
WilcoFiers, line 49: This looks like a `.every()`? Is that right? Can you use whatever array boolean method is appropriate for this?

lib/core/utils/filter-html-attrs.js
```diff
@@ -0,0 +1,178 @@
...
+      if (Array.isArray(matcher)) {
+        return matcher.includes(value);
+      }
+
+      // matcher-like object, match all parts
+      let filter = true;
+
+      // match each property
+      if (matcher.nodeName) {
+        filter = filter && matcher.nodeName === nodeName;
```
WilcoFiers, line 45: This looks a little more readable:

```suggestion
      if (matcher.nodeName && matcher.nodeName !== nodeName) {
        return false
```

test/core/utils/filter-html-attrs.js
```diff
@@ -0,0 +1,246 @@
...
+
+  it('should not filter attribute if does not match list of values', function() {
+    assert.equal(filterHtmlAttrs(html, { type: ['submit', 'password'] }), html);
+  });
+
+  it('should filter attribute if node matches nodeName matcher', function() {
+    assert.equal(
+      filterHtmlAttrs(html, {
+        type: {
+          nodeName: 'input'
```
WilcoFiers, line 35: Even though I know it isn't, this looks like a bug. I think we can fix that by changing the name.

```suggestion
          forElement: 'input'
```

test/core/utils/filter-html-attrs.js
```diff
@@ -0,0 +1,246 @@
...
+
+  it('should filter attribute if node matches attributes matcher (boolean)', function() {
+    assert.equal(
+      filterHtmlAttrs(html, {
+        type: {
+          attributes: {
+            value: true
+          }
+        }
+      }),
```
WilcoFiers, line 50: This definitely looks odd. You're doing an "and" relationship of two equal things by nesting. That seems pretty counter-intuitive. Is there any reason we can't use a simplified version of the matcher we already have? So it would be this:

```suggestion
      filterHtmlAttrs(html, {
        attributes: {
            type: true,
            value: true
          }
        }
      }),
```

I realise that makes some of these things a little more verbose, but I definitely think there's a thing to be said here for not introducing another matcher format to axe-core. Even if this implementation is a subset of the other one.

WilcoFiers (changes requested):

straker:
Went with using the node itself to apply css selector matching.

lib/core/utils/filter-html-attrs.js
```diff
@@ -0,0 +1,75 @@
...
+
+    cache.set(outerHTML, node);
+  }
+
+  // be sure to append text nodes as well
+  Array.from(element.childNodes).forEach(child => {
+    node.appendChild(filterHtmlAttrs(child, filterAttrs));
+  });
+
+  return node;
```
WilcoFiers, line 72: Why is this returning elements? Isn't doing a bit of string manipulation much faster than creating entirely new DOM structures?

test/core/utils/filter-html-attrs.js
```diff
@@ -0,0 +1,49 @@
...
+  it('should return string if no attributes are passed', function() {
+    assert.equal(filterHtmlAttrs(html), html);
+  });
+
+  it('should filter attribute if exists', function() {
+    assert.equal(filterHtmlAttrs(html, { type: true }).outerHTML, expected);
+  });
+
+  it('should filter attribute if matches CSS selector', function() {
+    assert.equal(filterHtmlAttrs(html, { type: 'input' }).outerHTML, expected);
```
WilcoFiers, line 24: Can you put in an actual CSS string here? This could just be nodeName match still.

WilcoFiers (dismissed):

lib/core/utils/filter-html-attrs.js
straker: If we're matching CSS selectors we cannot work on just strings. I wanted to remove attributes but IE11 wouldn't remove `value` from an input element so had to instead recreate the DOM from scratch.

test/core/utils/filter-html-attrs.js
straker: This is a CSS string? I'm matching the `type` attribute only if the element is an `input` node.

WilcoFiers (approved):

PR #2810, opened 2021-02-17
fix(types): Add noHtml option

### LINKED ISSUE #2793: Add `noHtml` to axe.d.ts
Forgot to add `noHtml` option to typescript options

dylanb:
was missed in the 4.1.2 release

Closes #2793


straker (changes requested):
Seems the indentation is off. I would have expected prettier to catch that but seems it did not.

WilcoFiers (approved):

Issue #2604, opened 2020-10-28
Rule idea: flag empty table headers for review


WilcoFiers:
Fairly similar to the empty-headings rule. When a table header has no accessible name, axe-core should report it as *needs review**. This applies both to native `th` elements and to ARIA `rowheader` and `columnheader` roles.

WilcoFiers:
Assigned to @ jlin95.

padmavemulapati:
Verified with the latest axe-core code base, 
seeing the new rule "empty-table-header" failure where when a table header has no accessible name
![image](https://user-images.githubusercontent.com/47409579/111937675-40444e80-8aee-11eb-9a75-4c99ae48f564.png)

PR #2811, opened 2021-02-17
feat(empty-table-header): new rule to flag empty table headers

jlin95:
<< Describe the changes >>
New rule created for https://github.com/dequelabs/axe-core/issues/2604

For initial review only!

Closes issue:


test/integration/rules/aria-empty-table-header/aria-empty-table-header.html
```diff
@@ -0,0 +1,21 @@
+<!DOCTYPE html>
```
straker, line 1: The document element shouldn't be part of the code. This code is injected into the body so all we really need is everything inside the body tag.

test/integration/rules/aria-empty-table-header/aria-empty-table-header.html
```diff
@@ -0,0 +1,21 @@
...
+<html lang="en">
+	<head>
+		<title>empty-table-header test</title>
+		<meta charset="utf8" />
+		<script src="/axe.js"></script>
+	</head>
+
+	<body>
+		<th id="pass1">Ok</th>
```
straker, line 10: This won't work and is probably why the tests are failing. Table elements (td, th, etc.) cannot be created outside of a table and the browser will just ignore them and turn them into text:

```
<text>Ok
  rowheader with name
  columnheader with name
  
  
    </text>
```

Need to wrap these elements inside a valid table.

lib/rules/aria-empty-table-header.json
```diff
@@ -0,0 +1,13 @@
...
+	"id": "aria-empty-table-header",
+	"selector": "th, [role=\"rowheader\"], [role=\"columnheader\"]",
+	"matches": "heading-matches",
+	"tags": ["wcag131"],
+	"metadata": {
+		"description": "Ensures table headers have discernible text",
+		"help": "Table header text must not be empty"
+	},
+	"all": [],
+	"any": ["has-accessible-text"],
```
straker, line 11: This is not a valid check name. Did you mean `has-text-content`?

lib/rules/aria-empty-table-header.json
```diff
@@ -0,0 +1,13 @@
+{
+	"id": "aria-empty-table-header",
+	"selector": "th, [role=\"rowheader\"], [role=\"columnheader\"]",
+	"matches": "heading-matches",
+	"tags": ["wcag131"],
+	"metadata": {
```
straker, line 6: The rule should mark empty header elements for needs review but the check produces failure results. Adding `reviewOnFail: true` as a property of the rule should fix it.

```suggestion
  "reviewOnFail": true,
	"metadata": {
```

straker (changes requested):
Thanks for putting this together. It's a great start and just needs a few things.

lib/rules/aria-empty-table-header.json
```diff
@@ -0,0 +1,14 @@
+{
+  "id": "aria-empty-table-header",
+  "selector": "th, [role=\"rowheader\"], [role=\"columnheader\"]",
+  "matches": "heading-matches",
```
straker, line 4: This matches is only for `heading` elements (h1-6) so the rule is failing to match the selector. I don't think you need a matches for this rule.

test/integration/rules/aria-empty-table-header/aria-empty-table-header.html
```diff
@@ -0,0 +1,49 @@
+<html lang="en">
+  <head>
+    <title>empty-table-header test</title>
+    <meta charset="utf8" />
+    <script src="/axe.js"></script>
+  </head>
+
+  <body>
+    <table id="pass1">
```
straker, line 9: The `id`s should be on the `th`/role elements, otherwise the selectors for the test come out to `table > tr > th`

test/integration/rules/aria-empty-table-header/aria-empty-table-header.json
```diff
@@ -0,0 +1,6 @@
+{
+	"description": "aria-empty-table-header tests",
+	"rule": "aria-empty-table-header",
+	"violations": [["fail1"], ["fail2"], ["fail3"]],
```
straker, line 4: Now that the rule no longer returns violations this should be `incomplete`. Also, the selectors need the `#` in front to match the id.

lib/rules/aria-empty-table-header.json
```diff
@@ -0,0 +1,14 @@
+{
+  "id": "aria-empty-table-header",
```
straker, line 2: Since this rule works on table elements and is not validating aria roles / attributes, it would probably be better to drop the `aria` from the name and just call the rule `empty-table-header`

straker (changes requested):
Thanks for the updates. Just a few more changes.

test/integration/rules/aria-empty-table-header/aria-empty-table-header.json
jlin95: So it should be "incomplete": [["#incomplete1"], etc", cmiiw?

straker: Sure, updating the ids would be nice too.

jlin95:
Out of curiosity, have been getting 
image while trying to run the axe.js build locally. 

I didn't make any changes to the Gruntfile/etc, just synced my fork with the develop branch...

test/integration/rules/empty-table-header/empty-table-header.json
```diff
@@ -0,0 +1,6 @@
+{
+  "description": "empty-table-header tests",
+  "rule": "empty-table-header",
+  "incomplete": [["incomplete1"], ["incomplete2"], ["incomplete3"]],
```
straker, line 4: Almost there! Just need to add the id selector `#` to each of these for the tests to pass.
```suggestion
  "incomplete": [["#incomplete1"], ["#incomplete2"], ["#incomplete3"]],
```

straker (changes requested):

straker:
> Out of curiosity, have been getting
> image while trying to run the axe.js build locally.
> 
> I didn't make any changes to the Gruntfile/etc, just synced my fork with the develop branch...

The develop script just constantly watches for changes to axe files and rebuilds the app or runs the tests. Looks like it's working but are you seeing something different?

jlin95:
> > Out of curiosity, have been getting
> > image while trying to run the axe.js build locally.
> > I didn't make any changes to the Gruntfile/etc, just synced my fork with the develop branch...
> 
> The develop script just constantly watches for changes to axe files and rebuilds the app or runs the tests. Looks like it's working but are you seeing something different?

Yep, basically previously I would navigate to localhost:9876/test/integrations and run all the tests at one shot from the browser, but now it seems that each rule/etc will execute the tests individually instead of all at one shot, please correct me if I understood it wrongly!

straker:
Oh yes, I believe that was before we switched to karma for our testing framework. Develop now just watches and re-runs the specific test file associated with the script. To run the full tests you do `npm test` or there is an entire script setup to run individual directory tests (e.g. `npm run test:unit:core` to run core tests).

WilcoFiers (approved):
Looks good. Thank you for the contribution!

WilcoFiers:
Reviewed for security.

PR #2816, opened 2021-02-23
fix(respondable): Avoid message duplication with messageId

WilcoFiers:
This change is a precursor to us adding custom message listeners to axe-core. Those would allow communication between windows without using window.postMessage, which can interfere with other scripts in a page, especially in environments like browser extensions.

This rewrite patches up a number of gaps as well. It gives every message a unique ID to avoid a message accidentally getting accepted twice. It now does a better job of logging invalid messages, and behavior between subscribers and responders has been made consistent.

This respondable implementation also prevents overriding the subscriber, which is why a lot of tests had to be tweaked.


lib/core/public/load.js
```diff
@@ -43,20 +44,21 @@ function runCommand(data, keepalive, callback) {
   }
 }
 
+if (window.top !== window) {
+  respondable.subscribe('axe.start', runCommand);
+  respondable.subscribe('axe.ping', (data, keepalive, respond) => {
```
WilcoFiers, line 49: Pulled these out to avoid registering them each time load is called.

dylanb (dismissed):

dylanb:
Approved but please look into the Windows failures and fix before a release

locales/pl.json
```diff
@@ -1,911 +1,911 @@
...
-  "failureSummaries": {
-    "any": {
-      "failureMessage": "Napraw następujące elementy: {{~it:value}}\n {{=value.split('\\n').join('\\n ')}}{{~}}"
-    },
-    "none": {
-      "failureMessage": "Napraw wszystkie poniższe: {{~it:value}}\n {{=value.split('\\n').join('\\n ')}}{{~}}"
-    }
-  },
-  "incompleteFallbackMessage": "AXE nie potrafił określić powodu. Czas użyć inspektora elementów!"
+	"lang": "pl",
```
straker, line 2: Looks like prettier was un-run on this file? seems whitespace turned into tabs

package-lock.json
```diff
@@ -1,13225 +1,8 @@
 {
   "name": "axe-core",
-  "version": "4.1.2",
-  "lockfileVersion": 2,
+  "version": "4.1.1",
```
straker, line 3: This version number is wrong. I think you need to merge latest from develop

sri-history.json
```diff
@@ -210,9 +210,5 @@
   "4.1.1": {
     "axe.js": "sha256-Tz6yShAcKxP8ce//YhbRkOLinUj+htEK2MjIi5yjY58=",
     "axe.min.js": "sha256-43geecotYZGPnRt6o9Uu7Pjl3GthxnDacXrzCGv11go="
-  },
```
straker, line 213: Don't want this removed.

test/core/utils/respondable.js
```diff
@@ -1,515 +1,561 @@
...
+  before(function() {
+    // assign axe a new uuid every time it's requested to trick
+    // the code that each respondable was called from a different
+    // context
+    Object.defineProperty(axe, '_uuid', {
+      get: function() {
+        return ++counter;
+      }
+    });
+  });
```
straker, line 47: Responsible no longer uses axe._uuid so this code isn't needed anymore

test/core/utils/respondable.js
```diff
@@ -1,515 +1,561 @@
...
-    var win = {
-      postMessage: function() {
-        success = true;
-      }
-    };
+  after(function() {
+    Object.defineProperty(axe, '_uuid', {
+      value: origAxeUUID
+    });
+  });
```
straker, line 80: Same with this code (not needed)

test/core/utils/respondable.js
```diff
@@ -1,515 +1,561 @@
...
+  describe('subscribe', function() {
+    it('is called with the same topic', function(done) {
+      var doneOnce = once(done);
+      frameSubscribe('greeting', function() {
+        doneOnce();
+      });
+      respondable(frameWin, 'greeting');
+      afterMessage(frameWin, function() {
+        doneOnce(new Error('subscribe was not called'));
+      });
```
straker, line 167: This test is a bit confusing because it doesn't assert anything. Would be a bit more understandable if you used sinon to assert called.

```suggestion
      var spy = sinon.spy();
      frameSubscribe('greeting', spy);
      respondable(frameWin, 'greeting');
      afterMessage(frameWin, function() {
        captureError(function() {
          assert.isTrue(spy.called);
          done();
        }, done)
      });
```

test/core/utils/respondable.js
```diff
@@ -1,515 +1,561 @@
...
-        _source: 'axeAPI.' + axe.version,
-        message: 'Help us Obi-Wan',
-        _axeuuid: 'otherAxe'
-      };
-
-      axe.utils.respondable(win, 'Death star', null, true, function(data) {
-        assert.equal(data, 'Help us Obi-Wan');
-        done();
+    it('is not called on a different topic', function(done) {
+      var doneOnce = once(done);
```
straker, line 171: Ditto here (and most of the following tests) for spies

test/core/utils/respondable.js
```diff
@@ -1,515 +1,561 @@
...
 
-      axe.utils.respondable(win, 'Death star', null, true, function() {
-        done(new Error('should not call callback'));
+  describe('respond', function() {
+    it('passes the response back', function(done) {
+      var receivedResponse;
+      frameSubscribe(
+        'greeting',
+        captureError(function(message, keepalive, respond) {
+          assert.isFalse(keepalive);
```
straker, line 300: Any reason to make this assert as part of this test?

test/core/utils/respondable.js
```diff
@@ -1,515 +1,561 @@
...
-
-      win = {
-        postMessage: function(message) {
-          var data = JSON.parse(message);
-          eventData.uuid = data.uuid + 'joker tricks!';
-          event.data = JSON.stringify(eventData);
-          document.dispatchEvent(event);
-        }
-      };
+    it('prohibits multiple response calls with keepalive: false', function(done) {
```
straker, line 318: This test was confusing since the assert shows keepalive is false. Maybe update to reflect where it's set?

```suggestion
    it('prohibits multiple response calls when respond sets keepalive to false', function(done) {
```

test/core/utils/respondable.js
```diff
@@ -1,515 +1,561 @@
...
 
-      setTimeout(function() {
-        done();
-      }, 100);
+      afterMessage(window, function() {
+        assert.deepEqual(receivedResponse, { responseNum: 1 });
+        afterMessage(window, function() {
+          assert.deepEqual(receivedResponse, { responseNum: 2 });
+          done();
+        });
```
straker, line 360: I feel like this can easily lead to a timing problem without the longer timeout delay like the prior test

```suggestion
        }, 100);
```

test/core/utils/respondable.js
```diff
@@ -1,515 +1,561 @@
...
+      );
+    });
 
-      axe.utils.respondable(win, 'Death star', null, true, function(data) {
-        assert.instanceOf(data, Error);
-        assert.equal(data.message, 'The exhaust port is open!');
-        window.evil = undefined;
-        done();
+    it('it errors if multiple callbacks are registered', function(done) {
+      frameSubscribe('greeting', function(message, _, respond) {
```
straker, line 440: The back and forth nature of these tests make them confusing to follow. You have to really understand ordering implied by the callback structure. It would be more helpful if the functions were separated from the rest of the code and named to imply order. Something like:

```
function firstResponse() { respond('2a', ... }
function secondResponse() { assert.equal(message, '2a'); }
function thirdResponse() { assert.equal(message, '3a'); }
```

test/core/utils/send-command-to-frame.js
```diff
@@ -13,6 +14,26 @@ describe('axe.utils.sendCommandToFrame', function() {
...
   };
 
+  it('should return results from frames', function(done) {
+    var frame = document.createElement('iframe');
+    frame.addEventListener('load', function() {
+      axe.utils.sendCommandToFrame(
+        frame,
+        params,
+        function() {
+          done();
```
straker, line 24: I feel like you should assert the result or something rather than just saying any callback is valid

straker (dismissed):

lib/core/utils/respondable/message-id.js
```diff
@@ -0,0 +1,22 @@
+import { v1 as createUuid } from '../uuid';
+
+const messageIds = [];
+
+export function createMessageId() {
+  const uuid = createUuid();
```
dbjorge, line 6: Ideally you want this to be hard to guess, in addition to being unique; v1 UUIDs only attempt to guarantee the latter and there isn't that much entropy between different UUIDs from the same source. We recommend combining a uuid with a nonce at least 32 bytes long that is generated using a cryptographically secure rng source (eg, [`window.crypto.getRandomValues`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues); note https://github.com/jsdom/jsdom/issues/1612 though).

dbjorge: (more important on channel ID creation than message ID creation but good to be consistent)

test/core/utils/respondable.js
WilcoFiers: It won't, this runs after the second message is received. There is no race condition here.

straker (approved):

dylanb (approved):

straker (approved):

PR #2824, opened 2021-03-02
docs: make the documentation on the rules options more precise

dylanb:
Make the documentation on the rules options more precise

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -215,9 +215,9 @@ axe.configure({
       - `options` - mixed(optional). This is the options structure that is passed to the evaluate function and is intended to be used to configure checks. It is the most common property that is intended to be overridden for existing checks.
       - `enabled` - boolean(optional, default `true`). This is used to indicate whether the check is on or off by default. Checks that are off are not evaluated, even when included in a rule. Overriding this is a common way to disable a particular check across multiple rules.
   - `rules` - Used to add rules to the existing set of rules, or to override the properties of existing rules
-    - The rules attribute is an Array of rule objects
-    - each rule object can contain the following attributes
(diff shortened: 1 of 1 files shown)


doc/API.md
```diff
@@ -215,9 +215,9 @@ axe.configure({
...
       - `enabled` - boolean(optional, default `true`). This is used to indicate whether the check is on or off by default. Checks that are off are not evaluated, even when included in a rule. Overriding this is a common way to disable a particular check across multiple rules.
   - `rules` - Used to add rules to the existing set of rules, or to override the properties of existing rules
-    - The rules attribute is an Array of rule objects
-    - each rule object can contain the following attributes
-      - `id` - string(required). This uniquely identifies the rule. If the rule already exists, it will be overridden with any of the attributes supplied. The attributes below that are marked required, are only required for new rules.
+    - The rules attribute is an object with key-value pairs, the key is the `id` of a rule and the value is a rule object for that rule
+    - each rule object can contain the following attributes. If the rule already exists, it will be overridden with any of the attributes supplied. The attributes below that are marked required, are only required when adding a new rule.
```
WilcoFiers, line 219:
```suggestion
    - The rules attribute is an Array of rule objects
    - each rule object can contain the following attributes. If the rule ID already exists, the rule will be overridden with any of the attributes supplied. The attributes below that are marked required, are only required when adding a new rule.
```

WilcoFiers (changes requested):
I don't know why it was ever designed this way, but axe.configure({ rules }) for sure is an array. You may have it confused with axe.run({ rules }), which is an object with IDs as keys.

doc/API.md
dylanb: facepalm

PR #2829, opened 2021-03-04
chore: prevent .only tests from being commited

straker:
We've had a few `.only` tests sneak into committed code. This should prevent that from happening again.


stephenmathieson (dismissed):
There's a [plugin for this](https://www.npmjs.com/package/eslint-plugin-mocha-no-only) so that we wouldn't have to maintain this funky looking code.

No need to bikeshed here tho 😄

straker:
I like that much better! Thanks for letting me know

stephenmathieson (approved):

PR #2838, opened 2021-03-13
Hidden input visible label

jfmacdonald:
Add tags, metadata, and none checks

Closes issue:


WilcoFiers:
After discussing this with a few folks, I'm not sure this rule is going to work out. I have concerns about its validity, and since this PR has gone untouched for weeks I'm going to close it.

Issue #2840, opened 2021-03-15
 require function is used in a way in which dependencies cannot be statically extracted


ezhylin:
**Expectation:** 
As a consumer of the library I would like to understand why I am getting the warning (see below) at build time using webpack and what should I do to get rid of it. 

**Actual:** 
during a build process, I see the following warning

<pre>
<code>
WARNING in ./node_modules/axe-core/axe.js 6275:19-26
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
 @ ./node_modules/@axe-core/react/dist/index.js
 @ ./js/src/library/a11y.js
 @ ./js/src/index.js
 @ multi core-js/stable ./js/src/index.js
</code>
</pre>

**Misc:**
<pre><code>
"@axe-core/react": "^4.1.0",
"webpack": "^4.46.0",
"webpack-cli": "^4.5.0",

For Tooling issues:
- Node version: v14.16.0
- Platform:  Mac
</code></pre>

straker:
Thanks for the issue. I believe you'll want to open this in https://github.com/dequelabs/axe-core-npm (where `@axe-core/react` is held). Also, would you be able to provide how you're using react-axe / webpack? Are you using it or `require`ing it in any special way?

ezhylin:
Thank you so much for the fast response, @straker ! 

Here is an example of how we require @axe-core/react in the app:

```js
// a11y.js
import axeCore from '@axe-core/react';

export const axe = async (React, ReactDOM, { enabled = false, timeout = 5000, ...config }) => {
    if (enabled) {
        await axeCore(React, ReactDOM, timeout, config).catch(console.error);
    }
}

// index.js
ReactDOM.render(
  <Provider store={store}>
    <App />
  </Provider>
  document.getElementById(id),
);

axe(React, ReactDOM, {
  enabled: !!process.env.MY_ENV_VAR,
});```

The webpack configuration is pretty simple. We use babel-loader to transpile JSX to JS and DefinePlugin to use environment variables. That's pretty much it.

istateside:
I don't think this is an issue in the @axe-core/react library. I believe this was introduced with [this line](https://github.com/dequelabs/axe-core/pull/2826/files#diff-fb259cc117394538f4c068d26b0e2bc427ffc8b8d4d5f8eeb58f4783b7b35891R27), added in the 4.1.3 release.

[...]

PR #2826, opened 2021-03-02
 chore(release): 4.1.3

straker:
Had to fix the tests again so added 1 extra commit.

PR #2843, opened 2021-03-16
fix(utils): fix warning thrown by Webpack

istateside:
https://github.com/dequelabs/axe-core/issues/2840#issuecomment-800588057

This PR updates [a line added in the recent 4.1.3 release](https://github.com/dequelabs/axe-core/pull/2826/files#diff-fb259cc117394538f4c068d26b0e2bc427ffc8b8d4d5f8eeb58f4783b7b35891R27). Webpack is throwing a warning when this library is included in a bundle, pointed at the given line.

I believe what's happening here is that Webpack sees the word "require", but doesn't understand that it's not being used to actually "require" something. Webpack doesn't know what to do with it, so it throws a warning.

The line is already wrapped in a try, so it should be fine to just drop the `&& require` in the condition. If `const nodeCrypto = require('crypto')` fails, the error will get caught and we can continue on.

Closes issue: #2840

### DIFF
--- a/lib/core/utils/uuid.js
+++ b/lib/core/utils/uuid.js
@@ -24,7 +24,7 @@ if (!_rng && _crypto && _crypto.getRandomValues) {
 }
 
 try {
-  if (!_rng && require) {
+  if (!_rng) {
     const nodeCrypto = require('crypto');
     _rng = () => nodeCrypto.randomBytes(16);
   }


dylanb (dismissed):

lib/core/utils/uuid.js
```diff
@@ -24,7 +24,7 @@ if (!_rng && _crypto && _crypto.getRandomValues) {
 }
 
 try {
-  if (!_rng && require) {
+  if (!_rng) {
```
stephenmathieson, line 27: This change worries me. How does this change effect other supported environments?

istateside: Before, if `require` wasn't defined, it would skip the conditional and continue on. Now, if `require` isn't defined, an error will be thrown by `require('crypto')` on the next line, which will be caught by the try/catch block. I don't think this changes the execution in a meaningful way

stephenmathieson: Thanks for the explanation.

dylanb (approved):

dylanb:
@WilcoFiers please merge if you are ok with this too

WilcoFiers (approved):
Thanks for the pull request. Yeah this makes sense.

PR #2847, opened 2021-03-22
chore: rerun rule-description

straker:
Another rule file update where the changes weren't rebuilt, probably due to accepting suggestions in the github interface but not rebuilding the code.


stephenmathieson (approved):
We should setup a GitHub action for this so we can stop worrying about it. I'm happy to do so if I get an OK from someone else...

Issue #2835, opened 2021-03-11
Runtime issue: Respondable target must be a frame in the current window


exp-tlawton:
This issue is happening in the axe-core chrome extension

Scenario:

1. Goes to url https://www.expedia.com/ 
2. Waits for page to fully render and scrolls through all available content
3. Attempts to run axe-core and gets the following error

https://user-images.githubusercontent.com/56890421/110852336-15910500-8267-11eb-9e7c-81c13c51794a.mp4

axe-core version: `4.1.3`
axe-extension version: `4.8.6`

```
Error: Respondable target must be a frame in the current window
    at rr (chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:56:95394)
    at ar (chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:56:95649)
    at ur (chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:56:96233)
    at cr (chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:56:98031)
    at chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:56:99806
    at chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:56:92014
    at Object.defer (chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:56:92048)
    at chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:56:99772
    at Array.forEach (<anonymous>)
    at mr (chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:56:99630)
```

straker:
Thanks for the issue. We'll look into getting this fixed.

exp-tlawton:
> Thanks for the issue. We'll look into getting this fixed.

Thanks @straker for the prompt response!

Best,

dylanb:
@WilcoFiers close?

padmavemulapati:
Validated with the latest code base of axe-core, not seeing any error when axe run on the above said test page
![image](https://user-images.githubusercontent.com/47409579/113700354-5eb46780-96f4-11eb-9c44-4ae5aef29318.png)

PR #2857, opened 2021-03-31
fix(respondable): work on iframes in shadow DOM

WilcoFiers:
Instead of checking window.frames, check frame.parent. Checked the HTML specs, apparently the spec authors thought this makes sense: https://github.com/WICG/webcomponents/issues/145

Closes issue: #2835


straker (approved):
Tested this with the help of the extension team and it looks like it's passing the gauntlet iframe test file. The original ticket that this closes no longer seems to throw the error so we weren't able to test it there.

Issue #2791, opened 2021-02-08
Region rule does not account for absolute URLs in skiplinks


WilcoFiers:
This course page creates 3 issues in the region rule: https://dequeuniversity.com/class/ux2/toc The ones on the "next" links are legit, those should be findable through landmarks. The skiplink however should have been excluded The URL "class/ux2/toc#content" is clearly a reference to another element on the same page. The rule should understand this.

padmavemulapati:
Valdiated with the latest axe-core dev branch code base,

earlier, it was finding `region` rule for `skiplinks` as well `next` links available in the page ` https://dequeuniversity.com/class/ux/toc`
![image](https://user-images.githubusercontent.com/47409579/151567243-6e7f0d2b-5c9d-4efe-96a7-7cd3d7aa6173.png)

As the suggested fix, `Region` rule does not account for `skiplinks`, with the latest code, not seeing failure for `skiplinks`
![image](https://user-images.githubusercontent.com/47409579/151567707-ec13a5b6-b7c3-41b7-8c0f-7a7050c81039.png)

PR #2875, opened 2021-04-15
fix(skip-link): work with absolute and relative paths

straker:
Closes issue: #2791


straker:
I approve for security but clicking the button in the check isn't working for me

lib/commons/dom/is-skip-link.js
```diff
@@ -13,22 +16,41 @@ const isInternalLinkRegex = /^\/?#[^/!]/;
...
   } else {
-    // define a skip link as any anchor element whose href starts with `#...`
-    // and which precedes the first anchor element whose href doesn't start
-    // with  `#...` (that is, a link to a page)
+    // define a skip link as any anchor element whose resolved href
+    // resolves to the current page and uses a fragment identifier (#)
+    // and which precedes the first anchor element whose resolved href
+    // does not resolve to the current page or that doesn't use a
+    // fragment identifier
+    const currentPage = window.location.origin + window.location.pathname;
```
WilcoFiers, line 28: I thought we couldn't count on window.location existing. Won't this throw in JSDOM?

lib/commons/dom/is-skip-link.js
```diff
@@ -13,22 +16,41 @@ const isInternalLinkRegex = /^\/?#[^/!]/;
...
     firstPageLink = querySelectorAll(
-      // TODO: es-module-_tree
       axe._tree,
-      'a:not([href^="#"]):not([href^="/#"]):not([href^="javascript"])'
-    )[0];
+      'a[href]:not([href^="javascript"])'
```
WilcoFiers, line 31: Should that be this: 

```suggestion
      'a[href]:not([href^="javascript:"])'
```

It's unlikely, but no reason to think someone can't just have `<a href="javascript/home.php">` or whatever.

lib/commons/dom/is-skip-link.js
```diff
@@ -13,22 +16,41 @@ const isInternalLinkRegex = /^\/?#[^/!]/;
...
     firstPageLink = cache.get('firstPageLink');
   } else {
-    // define a skip link as any anchor element whose href starts with `#...`
-    // and which precedes the first anchor element whose href doesn't start
-    // with  `#...` (that is, a link to a page)
+    // define a skip link as any anchor element whose resolved href
+    // resolves to the current page and uses a fragment identifier (#)
+    // and which precedes the first anchor element whose resolved href
+    // does not resolve to the current page or that doesn't use a
+    // fragment identifier
```
WilcoFiers, line 27: If you think an explanation like this is needed, create a function and put it in a docblock. Random blocks of comments like these are disruptive.

WilcoFiers (dismissed):
Couple small things.

lib/commons/dom/is-skip-link.js
straker: `window.origin` doesn't exist in IE11, so we can't use that, but `window.location` exists in IE11 and JSDOM. It's just that JSDOM doesn't have any data for the location properties so we can never determine if `path/of/url#id` is a skip link or not.

Boone6969 (approved):

lib/commons/dom/is-skip-link.js
WilcoFiers: OK, so I think what we do is just feature test. If we can do this check, we do it. If not (like in JSDOM) we fall back to the current behavior.

test/node/jsdom.js
```diff
@@ -58,4 +59,18 @@ describe('jsdom axe-core', function() {
...
   });
+
+  describe('isCurrentPageLink', function() {
+    it('should return null when url is not set', function() {
+      var dom = new jsdom.JSDOM(domStr);
+      var anchor = dom.window.document.getElementById('#skip');
+      assert.strictEqual(axe.commons.dom.isCurrentPageLink(anchor), null);
+    });
+
+    it('should return true when url is set', function() {
```
WilcoFiers, line 70: This doesn't check for "true"

lib/commons/dom/is-current-page-link.js
```diff
@@ -12,8 +12,9 @@ const angularRouterLinkRegex = /^#[!/]/;
  * @return {Boolean|null}
  */
 export default function isCurrentPageLink(anchor) {
-  // jsdom can have window.location.origin set to null
-  if (!window?.location.origin) {
+  // jsdom can have window.location.origin set to "null" (the string)
+  // if the url option is not set when parsing the dom string
+  if (window.location.origin === 'null') {
```
WilcoFiers, line 17: This doesn't look very robust. How about checking for a protocol?

```suggestion
  if (typeof window.location?.origin !== 'string' || window.location.origin.indexOf('://') === -1) {
```

WilcoFiers (changes requested):

WilcoFiers (approved):

PR #2877, opened 2021-04-19
WIP: feat: add file distribution for axe-monitor

wweeder:
work in progress: automate distribution of axe-core rules for axe-monitor. 

Closes issue:

### DIFF
--- a/S3/axe-core-package-gen.sh
+++ b/S3/axe-core-package-gen.sh
@@ -1,0 +1,26 @@
+tag=${git tag --format "%(refname:short) %(creatordate:short)" | tail -n1 | awk '{print$1}'};
+date=${git tag --format "%(refname:short) %(creatordate:short)" | tail -n1 | awk '{print$2}'};
+
+
+generate_version_files () {
+        mkdir $tag;
+	git checkout tags/$tag -b master; mkdir $tag;
+	npm install;
+	npm run build;
+	for axe.js $tag/; mv axe.min.js $tag/;
+        npm run build --lang=ja;
+        mv axe.js $tag/axe.ja.js; mv axe.min.js $tag/axe.ja.min.js;
+}
+
+create_release_json () {
+        echo "{" >> release.json; echo "    \"version\": \"$tag\"," >> release.json;
+	echo "    \"releaseDate\": \"$date\"," >> release.json;
+	echo "    \"releaseNotesUrl\": \"https://github.com/dequelabs/axe-core/releases/tag/$tag\"," >> release.json;
+	echo "    \"template\": \"standardTemplatesWcag21_ja.json\"" >> release.json;
+	echo "}" >> release.json
+	cp release.json release_ja.json;
+	mv release.json $tag/;
+	mv release_ja.json $tag/;
+}
+
+tar -zcvf axe-core-$tag.tar.gz $tag/;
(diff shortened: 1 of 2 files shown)


stephenmathieson (changes requested):
I don't fully understand what's happening here, but do not think we should have consumer-specific CI jobs here.

Why can't axe-monitor just `npm install axe-core` like everything else?

S3/axe-core-package-gen.sh
```diff
@@ -0,0 +1,26 @@
+tag=${git tag --format "%(refname:short) %(creatordate:short)" | tail -n1 | awk '{print$1}'};
+date=${git tag --format "%(refname:short) %(creatordate:short)" | tail -n1 | awk '{print$2}'};
+
+
+generate_version_files () {
+        mkdir $tag;
+	git checkout tags/$tag -b master; mkdir $tag;
```
stephenmathieson, line 7:
```suggestion
        mkdir $tag;
	git checkout tags/$tag -b master;
```

No need to do this twice.

S3/axe-core-package-gen.sh
```diff
@@ -0,0 +1,26 @@
...
+	npm install;
+	npm run build;
+	for axe.js $tag/; mv axe.min.js $tag/;
```
stephenmathieson, line 10: What is this loop for?

S3/axe-core-package-gen.sh
```diff
@@ -0,0 +1,26 @@
+tag=${git tag --format "%(refname:short) %(creatordate:short)" | tail -n1 | awk '{print$1}'};
+date=${git tag --format "%(refname:short) %(creatordate:short)" | tail -n1 | awk '{print$2}'};
+
+
+generate_version_files () {
+        mkdir $tag;
+	git checkout tags/$tag -b master; mkdir $tag;
+	npm install;
```
stephenmathieson, line 8: Why do this here? We already have deps installed/cached in CI.

S3/axe-core-package-gen.sh
```diff
@@ -0,0 +1,26 @@
+tag=${git tag --format "%(refname:short) %(creatordate:short)" | tail -n1 | awk '{print$1}'};
```
stephenmathieson, line 1: Please run `shellcheck` on this file and fix all of the errors it raises.

S3/axe-core-package-gen.sh
```diff
@@ -0,0 +1,26 @@
...
+        mkdir $tag;
+	git checkout tags/$tag -b master; mkdir $tag;
+	npm install;
+	npm run build;
+	for axe.js $tag/; mv axe.min.js $tag/;
+        npm run build --lang=ja;
+        mv axe.js $tag/axe.ja.js; mv axe.min.js $tag/axe.ja.min.js;
+}
+
+create_release_json () {
```
stephenmathieson, line 15: This function does not appear to be used.

S3/axe-core-package-gen.sh
```diff
@@ -0,0 +1,26 @@
+tag=${git tag --format "%(refname:short) %(creatordate:short)" | tail -n1 | awk '{print$1}'};
+date=${git tag --format "%(refname:short) %(creatordate:short)" | tail -n1 | awk '{print$2}'};
+
+
+generate_version_files () {
```
stephenmathieson, line 5: This function does not appear to be used.

stephenmathieson (changes requested):
Many small issues in the added shell script, plus I don't think we should be doing this (see prior review comment).

PR #2885, opened 2021-04-23
docs(frameMessenger): clarify advanced use cases

WilcoFiers:
Processing some feedback from @dbjorge. Thanks for the thorough review.

### DIFF
--- a/axe.d.ts
+++ b/axe.d.ts
@@ -324,7 +324,7 @@ declare namespace axe {
   };
   type Close = Function;
   type TopicHandler = (data: TopicData, responder: Responder) => void;
-  type ReplyHandler = (data: ReplyData, responder?: Responder) => void;
+  type ReplyHandler = (message: any | Error, keepalive: boolean, responder: Responder) => void;
   type Responder = (
     message: any | Error,
     keepalive?: boolean,
--- a/axe.d.ts
+++ b/axe.d.ts
@@ -318,17 +318,17 @@ declare namespace axe {
     open: (topicHandler: TopicHandler) => Close | void;
     post: (
       frameWindow: Window,
-      data: TopicData | ReplyData,
+      data: TopicData,
       replyHandler: ReplyHandler
-    ) => void;
+    ) => boolean | void;
   };
   type Close = Function;
-  type TopicHandler = (data: TopicData, responder?: Responder) => void;
+  type TopicHandler = (data: TopicData, responder: Responder) => void;
   type ReplyHandler = (data: ReplyData, responder?: Responder) => void;
   type Responder = (
-    message: any,
-    keepalive: boolean,
-    replyHandler: ReplyHandler
+    message: any | Error,
+    keepalive?: boolean,
+    replyHandler?: ReplyHandler
   ) => void;
   type TopicData = { topic: String } & ReplyData;
   type ReplyData = { channelId: String; message: any; keepAlive: Boolean };
--- a/axe.d.ts
+++ b/axe.d.ts
@@ -330,8 +330,8 @@ declare namespace axe {
     keepalive?: boolean,
     replyHandler?: ReplyHandler
   ) => void;
-  type TopicData = { topic: String } & ReplyData;
-  type ReplyData = { channelId: String; message: any; keepAlive: Boolean };
+  type TopicData = { topic: string } & ReplyData;
+  type ReplyData = { channelId: string; message: any; keepalive: boolean };
 }
 
 export = axe;
--- a/doc/frame-messenger.md
+++ b/doc/frame-messenger.md
@@ -45,7 +45,7 @@ The `open` function can `return` an optional `close` function. Axe-core will onl
 
 Currently, axe-core will only require `replyHandler` to be called once, so promises can also be used here. This may change in the future, so it is preferable to make it possible for `replyHandler` to be called multiple times. Some axe-core [plugins](plugins.md) may rely on this feature.
 
-A second frameMessenger feature available to plugins, but not used in axe-core by default is to reply to a reply. This works by passing `replyHandler` a `responder` callback as a second argument. This requires a different setup, in which callbacks as stored based on their `channelId` property.
+A second frameMessenger feature available to plugins, but not used in axe-core by default is to reply to a reply. This works by passing `replyHandler` a `responder` callback as a second argument. This requires a different setup, in which callbacks are stored based on their `channelId` property.
 
 ```js
 // store handlers based on channelId
--- a/doc/frame-messenger.md
+++ b/doc/frame-messenger.md
@@ -8,10 +8,13 @@ Tools like browser extensions and testing environments often have different chan
 axe.frameMessenger({
   // Called to initialize message handling
   open(topicHandler) {
+    // Map data from the bridge to
+    function subscriber(frameWin, data, response) {
+      // Data serializations / validation / etc. here
+      topicHandler(data, response);
+    }
     // Start listening for "axe-core" events
-    const unsubscribe = bridge.subscribe('axe-core', data => {
-      topicHandler(data);
-    });
+    const unsubscribe = bridge.subscribe('axe-core', subscriber);
     // Tell axe how to close the connection if it needs to
     return unsubscribe;
   },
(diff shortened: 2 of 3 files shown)


dylanb (approved):

doc/frame-messenger.md
```diff
@@ -8,10 +8,13 @@ Tools like browser extensions and testing environments often have different chan
 axe.frameMessenger({
   // Called to initialize message handling
   open(topicHandler) {
+    // Map data from the bridge to
```
dbjorge, line 11: missing end of sentence?

doc/frame-messenger.md
```diff
@@ -8,10 +8,13 @@ Tools like browser extensions and testing environments often have different chan
...
+    function subscriber(frameWin, data, response) {
+      // Data serializations / validation / etc. here
```
dbjorge, line 13: nit: this would probably be where an implementer would be *de*serializing data, not serializing it

doc/frame-messenger.md
```diff
@@ -38,6 +41,65 @@ The `open` function can `return` an optional cleanup function, which is called w
 
 ## axe.frameMessenger({ post })
 
-`post` is a function that dictates how axe-core communicates with frames. It is passed three arguments: `frameWindow`, which is the frames `contentWindow`, the `data` object, and a `replyHandler` that must be called when responses are received.
+`post` is a function that dictates how axe-core communicates with frames. It is passed three arguments: `frameWindow`, which is the frames `contentWindow`, the `data` object, and a `replyHandler` that must be called when responses are received. To inform axe-core that no message was sent, return `false`. This informs axe-core not to await for the ping to time out.
```
dbjorge, line 44: nit: `frames` -> `frame's`

doc/frame-messenger.md
```diff
@@ -38,6 +41,65 @@ The `open` function can `return` an optional cleanup function, which is called w
...
+  post(frameWindow, data, replyHandler) {
+    // Store the handler so it can be called later
+    channels[data.channelId] = replyHandler;
+    // Send a message to the frame
+    bridge.send(frameWindow, data);
+  },
+
+  open(topicHandler) {
+    function subscriber(frameWin, data, response) {
+      const { channelId, message, messageId } = data;
```
dbjorge, line 64: `messageId` isn't used, it can probably be omitted to simplify the example

doc/frame-messenger.md
```diff
@@ -38,6 +41,65 @@ The `open` function can `return` an optional cleanup function, which is called w
...
+axe.frameMessenger({
+  post(frameWindow, data, replyHandler) {
+    // Store the handler so it can be called later
+    channels[data.channelId] = replyHandler;
+    // Send a message to the frame
+    bridge.send(frameWindow, data);
+  },
+
+  open(topicHandler) {
+    function subscriber(frameWin, data, response) {
```
dbjorge, line 63: `response` isn't used, it can probably be omitted to simplify the example

doc/frame-messenger.md
```diff
@@ -38,6 +41,65 @@ The `open` function can `return` an optional cleanup function, which is called w
...
+      const responder = createResponder(frameWin, channelId);
+
+      // If there is a topic, pass it to the axe supplied topic-handler
+      if (data.topic) {
+        topicHandler(data, responder);
+
+        // If there is a replyHandler stored, invoke it
+      } else if (channels[channelId]) {
+        const replyHandler = channels[channelId];
+        replyHandler(message, keepalive, responder);
```
dbjorge, line 75: `keepalive` is never initialized

doc/frame-messenger.md
```diff
@@ -38,6 +41,65 @@ The `open` function can `return` an optional cleanup function, which is called w
...
+      // Create a callback to invoke on a reply.
+      const responder = createResponder(frameWin, channelId);
+
+      // If there is a topic, pass it to the axe supplied topic-handler
+      if (data.topic) {
+        topicHandler(data, responder);
+
+        // If there is a replyHandler stored, invoke it
+      } else if (channels[channelId]) {
+        const replyHandler = channels[channelId];
```
dbjorge, line 74: Does axe-core guarantee that it will only maintain one open topicHandler at a time? If so, it would be nice to document that in the `open` section; if not, I think this example will over-invoke replyHandlers that use `keepAlive`

axe.d.ts
```diff
@@ -318,17 +318,17 @@ declare namespace axe {
...
   type Responder = (
-    message: any,
-    keepalive: boolean,
-    replyHandler: ReplyHandler
+    message: any | Error,
+    keepalive?: boolean,
+    replyHandler?: ReplyHandler
   ) => void;
   type TopicData = { topic: String } & ReplyData;
   type ReplyData = { channelId: String; message: any; keepAlive: Boolean };
```
dbjorge, line 334: This uses a `keepAlive` property (uppercase A); based on the rest of the docs, L330 of the typings, and L29 of `respondable.js`, I think itshould probably be lowercase A

dbjorge, line 334: Should `keepAlive` be using the primitive `boolean` type here, or is `Boolean` really intended?

dbjorge (changes requested):
Thanks, this is a substantial improvement! A few more comments inline

doc/frame-messenger.md
```diff
@@ -38,6 +41,65 @@ The `open` function can `return` an optional cleanup function, which is called w
...
+    // Store the new reply handler, possibly replacing a previous one
+    //   to avoid receiving a message twice.
+    channels[channelId] = replyHandler;
+    // Send a message to the frame
+    bridge.send(frameWin, { channelId, message, keepalive });
+  };
+}
+```
+
+## Error handling & Timeouts
```
dbjorge, line 101: The typings for `Responder` got updated to suggest that it needs to be able to accept a message of type `Error`, but this section doesn't mention how to handle that case. It would be nice for it to talk about that explicitly, especially since it is an exception to the "axe will pass messages which are round-trippable through JSON.stringify" rule.

dbjorge: (if the intent is to pass it on as-is to `replyHandler`s as part of `ReplyData` `message`s, it would be good to update the type of the `message` property in `ReplyData` to match the `message` parameter in `Responder`)

axe.d.ts
```diff
@@ -318,17 +318,17 @@ declare namespace axe {
...
-      data: TopicData | ReplyData,
+      data: TopicData,
       replyHandler: ReplyHandler
-    ) => void;
+    ) => boolean | void;
   };
   type Close = Function;
-  type TopicHandler = (data: TopicData, responder?: Responder) => void;
+  type TopicHandler = (data: TopicData, responder: Responder) => void;
   type ReplyHandler = (data: ReplyData, responder?: Responder) => void;
```
dbjorge, line 327: The `ReplyHandler` type here (and corresponding docs) is still inconsistent with actual usage: https://github.com/dequelabs/axe-core/blob/b388eb3425817edc0742e25542be8d29f7aebdde/lib/core/utils/frame-messenger/message-handler.js#L67

Presuming you want to avoid breaking plugin compat, this should be 

```suggestion
  type ReplyHandler = (message: any | Error, keepalive: boolean, responder: Responder) => void;
```

For the sake of plugin compat, I think it's probably best to make the parameters non-optional; frameMessengers should always be sending them, even if axe-core's default reply handlers ignore them

dbjorge: (Given this change, it might be more clear to eliminate the `ReplyData` type and fold it into `TopicData`, since that's the only other spot it's used)

doc/frame-messenger.md
```diff
@@ -34,10 +37,71 @@ axe.frameMessenger({
...
+`post` is a function that dictates how axe-core communicates with frames. It is passed three arguments: `frameWindow`, which is the frame's `contentWindow`, the `data` object, and a `replyHandler` that must be called when responses are received. To inform axe-core that no message was sent, return `false`. This informs axe-core not to await for the ping to time out.
+
+Currently, axe-core will only require `replyHandler` to be called once, so promises can also be used here. This may change in the future, so it is preferable to make it possible for `replyHandler` to be called multiple times. Some axe-core [plugins](plugins.md) may rely on this feature.
+
+A second frameMessenger feature available to plugins, but not used in axe-core by default is to reply to a reply. This works by passing `replyHandler` a `responder` callback as a second argument. This requires a different setup, in which callbacks as stored based on their `channelId` property.
```
straker, line 48:
```suggestion
A second frameMessenger feature available to plugins, but not used in axe-core by default is to reply to a reply. This works by passing `replyHandler` a `responder` callback as a second argument. This requires a different setup, in which callbacks are stored based on their `channelId` property.
```

doc/frame-messenger.md
```diff
@@ -34,10 +37,71 @@ axe.frameMessenger({
...
+    bridge.send(frameWin, { channelId, message, keepalive });
+  };
+}
+```
+
+## Error handling & Timeouts
+
+If for some reason the frameMessenger fails to open, post, or close you should not throw an error. Axe-core will handle missing results by reporting on them in the `frame-tested` rule. It should not be possible for the `topicHandler` and `replyHandler` callbacks to throw an error. If this happens, please file an issue.
+
+Axe-core has a timeout mechanism built in, which pings frames to see if they respond before instructing them to run. There is no retry behavior in axe-core, which assumes that whatever channel is used is stable. If this isn't the case, this will need to be built into frameMessenger.
```
straker, line 105: This timeout is only applicable when using `respondable` for the messaging and not in frame messenger itself.

straker (changes requested):
@dbjorge should probably also give an approval this before it's merged

dbjorge:
Thanks for the update! It looks like there's still one unaddressed comment (the inconsistency between typings and usage on `ReplyHandler`), otherwise the updates look good to me

doc/frame-messenger.md
WilcoFiers: Not really. It's axe-core's engine that does the ping. Respondable doesn't have that. I think I'm going to leave this as is.

straker (approved):

PR #2887, opened 2021-04-23
chore: Convert ALL tabs to spaces

WilcoFiers:
Missed a couple of 'm.


stephenmathieson (commented):
Could add a `.json` extension to these files so that Prettier would do its thing 🤷‍♂️

WilcoFiers:
Wouldn't  the `*.{json,md,js}` do that @stephenmathieson?

stephenmathieson:
> Wouldn't the *.{json,md,js} do that @stephenmathieson?

No, the files start with `.`, so that glob won't match.

stephenmathieson:
First, do this:

```
git mv .huskyrc .huskyrc.json
git mv .prettierrc .prettierrc.json
```

Then update `package.json#scripts` to include:

```
"fmt": "prettier --write .*.json *.{json,md,js} ......."
```

The first glob (`.*.json`) will cover the `git mv`ed config files.

stephenmathieson (approved):

Issue #2727, opened 2020-12-31
Autocomplete username should allow email


ShaneHudson:
Expectation: Allow `<input type="email" autocomplete="username">` 

Actual: It says there is a serious impact and `Ensure the autocomplete attribute is correct and suitable for the form field`.

Motivation: It is very common for email addresses to be used as the login and this ensures it works correctly in password managers. There was a discussion around this on the spec and it appears to have been agreed to be allowed.

<pre><code>
axe-core version: 4.0.2
browser plugin: v4.6.0

Firefox 84.0.1
</code></pre>

WilcoFiers:
Thanks for reporting Shane. I had not yet thought of it this way. I see where you're coming from here, but I also think this could create some unnecessary confusion, which is what this rule (and the underlying WCAG criterion) is designed to catch. By communicating on the one hand that this is an email, and on the other that this is a username you sending mixed signals. That's not ideal. If you mean for this to be a email address, use that both in the label, type and autocomplete.

There is an issue there, for sure, but I'm not sure it's substantial enough for axe-core to fail it. I'll check with a few more folks, get their opinion on the matter.

WilcoFiers:
The HTML spec has been updated to allow autocomplete=username on type=email. I think that takes away most arguments against this. I'll do a little testing to make sure that works, but I think we can make that change now.

padmavemulapati:
Verified with the latest axe-core-develop branch code,
with proper usage of `autocomplete` attribute we are no longer to see the failure in `autocomplete-valid` rule.
code snippet:
<input type="email" autocomplete="username"> , which is fixed in the latest 4.2.1
also we can still  see failure when we pass invalid data for `autocomplete` attribute - 
`<input autocomplete="work mail" id="username" />` 
![image](https://user-images.githubusercontent.com/47409579/120331192-25bafc80-c30b-11eb-9c1d-303f05d03bb8.png)

![image](https://user-images.githubusercontent.com/47409579/120331515-7a5e7780-c30b-11eb-8e50-ac2507ce8b42.png)

where as failure we can see with the old versions 4.0.1
![image](https://user-images.githubusercontent.com/47409579/120331289-41be9e00-c30b-11eb-90c9-2cec5c7f1c30.png)

PR #2896, opened 2021-04-27
fix(autocomplete-appropriate): pass for autocomplete=username and type=email

clottman:
Per issue #2727, allow the `autocomplete-appropriate` rule to pass for an input like `<input type='email' autocomplete='username' >`

Also fixes spacing in the name of a test for autocomplete-valid. 

Closes issue: #2727


lib/checks/forms/autocomplete-appropriate-evaluate.js
```diff
@@ -12,6 +12,7 @@ function autocompleteAppropriateEvaluate(node, options, virtualNode) {
   const allowedTypesMap = {
     bday: ['text', 'search', 'date'],
     email: ['text', 'search', 'email'],
+    username: ['email'],
```
straker, line 15: I believe to fix the test we'll need this to also include `text` and probably `search`. Otherwise it's restricted to just `type=email`.

straker (changes requested):
Thanks for the pr! Added a suggestion that should hopefully help fix the tests.

lib/checks/forms/autocomplete-appropriate-evaluate.js
clottman: Ah! That makes sense, thanks for the hint. :)

straker:
Perfect. Sorry one last thing. Could you add an [integration test](https://github.com/dequelabs/axe-core/blob/develop/test/integration/rules/autocomplete-valid/autocomplete-valid.html) for the new combination?

straker (approved):
Thanks!

straker:
Reviewed for security

Issue #2897, opened 2021-04-28
axe-core throws a warn for treeitem with missing parent while parent is present


veselints:
Product: 
axe-core

Expectation: 
No warn should be present

Actual: 
Ensures elements with an ARIA role that require parent roles are contained by them. Required ARIA parents role not present: group, tree. The above issue is reported when testing the following markup:
```html
<div tabindex="0" role="tree">
	<li role="treeitem" aria-expanded="false">
		<span>foo</span>
	</li>
	<li role="treeitem" aria-expanded="true">
		<span>
		    bar
			<span></span>
		</span>
		<ul role="group" aria-hidden="false">
			<li role="treeitem">
				<span>ber</span>
			</li>
		</ul>
	</li>
</div>
```

Motivation: 
All `treeitem` elements have either `tree` or `group` as a direct parent.

axe-core version: 4.2.0

WilcoFiers:
@veselints Thanks for reporting. I think axe is tripping up over encountering `treeitem > group > treeitem`, and is expecting that to always be `tree > group > treeitem`. What you've done seems consistent with the ARIA practices, and from some testing I've done this seems to work reasonably well. I'll be sure to put in a fix for this.

padmavemulapati:
Verified with the latest code base - `axe-core-develop -code`, with 4.2.0
Warning seeing because of tripping around `treeitem` and `group`
Addressing `tree`  or `menu` in the parent will resolve the problem
Pass scenarios verified:
`<div role="list"> <div role="listitem"> <div role="group"> <div role="listitem" id="target"></div></div></div></div>`
`  <div role="menu"><div role="group"><p role="listitem" id="target">Nothing here.</p></div></div>`
Fail scenarios:
  `<div role="menu"><div role="menuitem"><div role="group"><div role="menuitem" id="target"></div></div></div></div>`
   ` <div role="list"><div role="listitem"><div role="none"><div role="listitem" id="target"></div></div></div></div>`
Pass results:
  
![image](https://user-images.githubusercontent.com/47409579/120314066-7543fd00-c2f8-11eb-93ba-0c3fa9400e31.png)

Fail results:
![Uploading image.png…]()

PR #2898, opened 2021-04-28
fix(required-parent): Allow *item > group > *item nesting

WilcoFiers:
Related to #2897. 

Need to look into menuitem before we merge this. `menuitem > group > menuitem` doesn't work nearly as well as treeitem and listitem seem to.

### DIFF
--- a/doc/check-options.md
+++ b/doc/check-options.md
@@ -7,6 +7,7 @@
   - [Global Options](#global-options)
   - [aria-allowed-role](#aria-allowed-role)
   - [aria-required-children](#aria-required-children)
+  - [aria-required-parent](#aria-required-parent)
   - [aria-roledescription](#aria-roledescription)
   - [color-contrast](#color-contrast)
   - [page-has-heading-one](#page-has-heading-one)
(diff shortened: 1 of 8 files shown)


test/integration/rules/preprocessor.js
```diff
@@ -20,7 +20,11 @@ var createIntegrationPreprocessor = function(logger) {
       // and add the test data to it
       var htmlpath = file.originalPath.replace(extRegex, '.html');
       var html = fs.readFileSync(htmlpath, 'utf-8');
-      var test = JSON.parse(content);
+      try {
+        var test = JSON.parse(content);
+      } catch (e) {
+        throw new Error('Unable to parse content of ' + file.originalPath)
```
WilcoFiers, line 26: Encountered this because I had a stray comma. Figured I'd tweak it a bit.

doc/check-options.md
```diff
@@ -107,6 +108,29 @@ All checks allow these global options:
   </tbody>
 </table>
 
+### aria-required-parent
+
+<table>
```
stephenmathieson, line 113: NBD but it's strange that we're not using Markdown to generate this table 🤷

straker (approved):

doc/check-options.md
straker: I remember it was because Markdown table syntax wasn't working so well with nested code blocks in the cells with multiline code blocks

Issue #2889, opened 2021-04-25
Allow access to axe standards


WilcoFiers:
Axe standards are a very useful tool. It would be great if there was a way to get the standards information out of axe-core. Something as simple is this could be sufficient.

```js
axe.getStandard()
```

padmavemulapati:
Verified with the latest axe-core-develop branch code base, with 4.2.0 version, 
Now it is allowing `getStandards` from `utils`
running a command from console, ` await axe.utils.getStandards('ariaAttrs');` 
will result,
![image](https://user-images.githubusercontent.com/47409579/120323186-dcff4580-c302-11eb-8494-c28558655f52.png)

PR #2903, opened 2021-04-28
fix(getStandards): Read standards from utils

WilcoFiers:
Should have made this info available before. Not sure why we overlooked that.

Closes issue #2889

### DIFF
--- a/lib/core/utils/get-standards.js
+++ b/lib/core/utils/get-standards.js
@@ -1,0 +1,6 @@
+import standards from '../../standards/index'
+import clone from './clone'
+
+export default function getStandards() {
+  return clone(standards);
+}
--- a/lib/core/utils/get-standards.js
+++ b/lib/core/utils/get-standards.js
@@ -1,4 +1,4 @@
-import standards from '../../standards/index'
+import standards from '../../standards'
 import clone from './clone'
 
 export default function getStandards() {
--- a/lib/core/utils/index.js
+++ b/lib/core/utils/index.js
@@ -38,6 +38,7 @@ export { default as getScrollState } from './get-scroll-state';
 export { default as getScroll } from './get-scroll';
 export { default as getShadowSelector } from './get-shadow-selector';
 export { default as getSelector, getSelectorData } from './get-selector';
+export { default as getStandards } from './get-standards';
 export { default as getStyleSheetFactory } from './get-stylesheet-factory';
 export { default as getXpath } from './get-xpath';
 export { default as getAncestry } from './get-ancestry';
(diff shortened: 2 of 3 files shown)


lib/core/utils/get-standards.js
```diff
@@ -0,0 +1,6 @@
+import standards from '../../standards/index'
```
straker, line 1:
```suggestion
import standards from '../../standards'
```

straker (dismissed):

lib/core/utils/get-standards.js
stephenmathieson: Is this worth requesting changes for? There's no functional difference between the two.

stephenmathieson (commented):
This function seems unnecessary. Why would someone use it rather than just doing `import standards from '../standards'`?

lib/core/utils/get-standards.js
straker: Guess it just code style at this point. No where else do we ask for `/index` and I thought when we added webpack we purposefully removed `index` as webpack handled it.

stephenmathieson: I see. IMO style decisions shouldn't result in requested changes. Instead we should have Prettier/ESLint handle that sort of thing for us, otherwise we end up painting the bike shed 14 different colors.

NBD tho 🤷‍♂️

straker: I'm not sure there is a prettier or eslint config option to remove `/index` from the end of things. Looks like if we want to enforce it we'll probably need to make our own eslint rule for it.

stephenmathieson: Elsewhere we've [just documented that we don't use "index" files](https://github.com/dequelabs/walnut/tree/develop/app#patterns--rules) (they're confusing- which index do I have open/hard to search for in editors/prone to nitpicks like this/etc):

> `index.*` files should not be used; `index.html` is the exception


If we feel strongly enough about not `import`ing files like this, we could document it (maybe in the contribution guidelines) rather than writing a custom lint rule for is.

WilcoFiers: I agree, indexes are messy. I'll update the PR, and I'm going to open another one to get rid of the index files.

straker (approved):

stephenmathieson:
Should this be considered a feature commit since it's adding something to the public API?

WilcoFiers:
@stephenmathieson If it was added to `axe.`, then yes it would have been a feature. Things in commons / utils are not treated as new features, since we need to be able to add to this as part of fixing bugs.

Issue #2899, opened 2021-04-28
accessible name calculation regression in 4.2


straker:
Axe 4.2 has a regression with the accessible name calculation. We're not sure how it happened since we didn't touch that code.

In 4.1.3, the following code has a space between the text content and the image `aria-label`

```html
<a href="../images/index.html">Images tool test page<img id="img18" aria-label="aria-label text" alt="logo" src="../images/Accessibility.jpg" width="50" height="50"></a>
```
```js
// 4.1.3
axe.commons.text.accessibleText($0)  // "Images tool test page aria-label text"
```

However in 4.2. the space is no longer there

```js
// 4.2.0
axe.commons.text.accessibleText($0)  // "Images tool test pagearia-label text"
```

PR #2914, opened 2021-05-03
fix(accessible name) Fix regression in accessible name computation

clottman:
This regression was introduced in [the commit to replace phrasing elements](https://github.com/dequelabs/axe-core/commit/6aab97f782304d3c43379d6f234673cbad6c0cdd), that replaced a hard-coded list of elements with a call to `getElementsByContentType('phrasing')`. 

In the hardcoded array, there are 44 elements that make up the `phrasingElements` array. In the new code using the function, there are 54 elements that are now considered phrasing elements. If the virtual node is a phrasing element as defined in that array, the new content does not get spaces added around it when appending to the virtual name. 

With this in mind, I think it's possible other regressions/behavior changes were caused by that commit but not caught by tests. I'm not sure what the correct behavior is - possibly to only include `img` elements in the things that get spaces appended around the accessible name, as I'm doing here, or maybe to do it for other elements that were added, too (going back to the hard-coded list). Or, possibly this was previously a bug when compared to the spec/actual AT behavior, and the change I made would be the wrong behavior?

Hopefully this PR helps expose the issue at least, even if there needs to be more discussion/investigation before it's merged. 

Closes issue: #2899

### DIFF
--- a/lib/standards/html-elms.js
+++ b/lib/standards/html-elms.js
@@ -350,14 +350,17 @@ const htmlElms = {
       },
       usemap: {
         matches: '[usemap]',
-        contentTypes: ['interactive', 'embedded', 'phrasing', 'flow']
+        contentTypes: ['interactive', 'embedded', 'flow']
       },
       default: {
         // Note: allow role presentation and none on image with no
         // alt as a way to prevent axe from flagging the image as
         // needing an alt
         allowedRoles: ['presentation', 'none'],
-        contentTypes: ['embedded', 'phrasing', 'flow']
+        // Note: spec change (do not count as phrasing), because browsers
+        // insert a space between an img's accessible name and other
+        // elements' accessible names
+        contentTypes: ['embedded', 'flow']
       }
     },
     // 5.10 img Element
--- a/test/commons/text/accessible-text.js
+++ b/test/commons/text/accessible-text.js
@@ -3208,6 +3208,17 @@ describe('text.accessibleTextVirtual', function() {
       assert.equal(accessibleText(target), 'Country of origin: United States');
     });
 
+    it('passes test 160', function() {
+      fixture.innerHTML =
+        '<a id="test" href="../images/index.html">Images tool test page<img id="img18" aria-label="aria-label text" alt="logo" src="../images/Accessibility.jpg" width="50" height="50"></a>';
+      axe.testUtils.flatTreeSetup(fixture);
+      var target = fixture.querySelector('#test');
+      assert.equal(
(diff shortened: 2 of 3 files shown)


straker (changes requested):
Amazing find! Wilco and I discussed this and we feel the "better" approach would be to remove `phrasing` [from the img spec](https://github.com/dequelabs/axe-core/blob/develop/lib/standards/html-elms.js#L351-L361). Chrome, Firefox, and Safari all add a space in the name so instead of special casing it we should conform the spec to what the browsers do.

When making the change, please [leave a comment](https://github.com/dequelabs/axe-core/blob/develop/lib/standards/html-elms.js#L452) explaining why we are making it different than the spec.

lib/standards/html-elms.js
```diff
@@ -350,14 +350,17 @@ const htmlElms = {
       },
       usemap: {
         matches: '[usemap]',
-        contentTypes: ['interactive', 'embedded', 'phrasing', 'flow']
+        contentTypes: ['interactive', 'embedded', 'flow']
```
clottman, line 353: I wasn't sure if the content type needed to be removed from both the `default` and `usemap` objects - I'm not sure how usemap is used.

straker: Yep, removed from both should be fine.

test/commons/text/accessible-text.js
```diff
@@ -3208,6 +3208,17 @@ describe('text.accessibleTextVirtual', function() {
       assert.equal(accessibleText(target), 'Country of origin: United States');
     });
 
+    it('passes test 160', function() {
```
straker, line 3211: Sorry missed this one last time. These tests are specifically acceptance tests taken from the [AccName 1.1 Testable Statements](https://www.w3.org/wiki/AccName_1.1_Testable_Statements#Name_test_case_539). Since this exact case doesn't appear to be covered by those tests, we'll want to add this test [above this entire section](https://github.com/dequelabs/axe-core/blob/70ecb6dce4fd8b5bda5b75eb2ce4696237cc4628/test/commons/text/accessible-text.js#L1426) as it's own test. That way we don't imply it's an acceptance test.

straker (changes requested):

straker (approved):
Great work!

dylanb:
security reviewed

dylanb:
@clottman you should join the axe Slack community https://join.slack.com/t/axecommunity/shared_invite/zt-pus38v38-qLP_I4CIExkgPwilI_Xkeg

Issue #2912, opened 2021-05-03
Deprecate autocomplete-appropriate check


WilcoFiers:
Research from ACT has shown that autocomplete-appropriate is no longer appropriate. No modern browser cares what the type os on an input element to determine whether or not to apply `autocomplete`. Because of that, I'd like to deprecate autocomplete-appropriate and remove it from the autocomplete-valid rule.

clottman:
@WilcoFiers are you looking for anything in particular in order to deprecate it, or would removing it from the codebase be what you're looking for? 

Either way I'd be happy to take a look at this.

straker:
@clottman Thanks for looking into it. 

To deprecate a check you'll just need to add `deprecated: true` to the checks `.json` metadata file and mark the commit as a `feat` (we only deprecate in feature PRs). We can't remove it from the rule for now, so it'll still run until we're ready to release the next feature version.

padmavemulapati:
Validated with latest code from develop branch
`autocomplete-appropriate` is not available and it is deprecated.
We can see with the latest change, for the couple of test snippets, where one is for `auto-complete` and the other is for `autocomplete-appropriate`(we can see both violations in old axe-core rule 4.1.1)
1. `<input autocomplete="badterm" id="fail-auto-complete" />`
2. `<input type="NUMBER" autocomplete="email" id="fail-auto-appropriate" />`

PR #2917, opened 2021-05-04
feat: deprecate autocomplete appropriate check

clottman:
Deprecates autocomplete-appropriate

Closes issue: #2912

### DIFF
--- a/lib/checks/forms/autocomplete-appropriate.json
+++ b/lib/checks/forms/autocomplete-appropriate.json
@@ -1,6 +1,7 @@
 {
   "id": "autocomplete-appropriate",
   "evaluate": "autocomplete-appropriate-evaluate",
+  "deprecated": true,
   "metadata": {
     "impact": "serious",
     "messages": {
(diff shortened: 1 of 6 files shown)


straker (changes requested):
I talked to Wilco and he mentioned that we should also remove the check from the rule metadata as well. Sorry about that.

clottman:
@straker I'm not sure what you mean by that, would you mind being more specific? 

I did notice looking at some other deprecated checks that the impact for the others was 'minor' - it makes sense to me that a deprecated check shouldn't be a severe impact, so I made that change. Can reverse it if that was the wrong call.

clottman:
Ah I think I figured it out! :D

straker:
Sorry missed the first question. Yes you removed it correctly. As for the impact change I don't think we should do that as the impact is based on user testing. The tests will also need to be updated with the removal of the check as those failures will no longer fail.

WilcoFiers (approved):

WilcoFiers:
Reviewed for security.

Issue #2926, opened 2021-05-11
aria-label on input elements reports incorrectly critical issue depending on its type


rastersysteme:
Product:
axe Extension

Actual:
Using `aria-label` on `input` elements with different used `type` attributes when no `label` is provided reports following error message:
> Elements must only use allowed ARIA attributes

![image](https://user-images.githubusercontent.com/1895414/117863535-2aecd480-b294-11eb-9e1d-36199e3e75ab.png)

Exemplary some input types:

```html
// critical issue
<input type="time" aria-label="some text" />
<input type="password" aria-label="some text" />
<input type="date" aria-label="some text" />

// no problem
<input type="number" aria-label="some text" />
<input type="search" aria-label="some text" />
<input type="text" aria-label="some text" />
```

Expectation:
No critical issue should be reported for **all** input types.

Motivation:
According to [WAI-ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/#aria-label) and its [practices](https://www.w3.org/TR/wai-aria-practices/#naming_with_aria-label) this behaviour is wrong and should be corrected.

<pre><code>
axe-core version: 4.2.0
axe DevTools - Web Accessibility Testing: 4.9.3

Platform: Mac
Browser: Chrome 90
</code></pre>

WilcoFiers:
Thank you for reporting this. I actually found this exact issue yesterday too and had it on my todo list to report. Axe seems to have a problem determining the implicit role of these elements. Looks like a fairly simple thing to fix.

WilcoFiers:
Forgot to reply, but there are actually two problems here. Firstly that axe-core is failing to identify some input types as `textbox` elements. The second is that global ARIA attributes are failing on elements with no implicit role. Another example of that is:

```html
<canvas aria-label="hello world"></canvas>
```

padmavemulapati:
Verified as part of https://github.com/dequelabs/axe-core/issues/2933 , not reproducing the issue now

PR #2929, opened 2021-05-12
fix: treat input with no role as textbox

WilcoFiers:
Have `<input type="password" aria-label="password">` pass aria-allowed-attr.

Related issue: #2926 -  **does not close it**.

### DIFF
--- a/lib/commons/standards/implicit-html-roles.js
+++ b/lib/commons/standards/implicit-html-roles.js
@@ -110,19 +110,8 @@ const implicitHtmlRoles = {
     }
 
     switch (vNode.props.type) {
-      case 'button':
-      case 'image':
-      case 'reset':
-      case 'submit':
-        return 'button';
       case 'checkbox':
         return 'checkbox';
-      case 'email':
-      case 'tel':
-      case 'text':
-      case 'url':
-      case '': // text is the default value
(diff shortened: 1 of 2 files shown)


lib/commons/standards/implicit-html-roles.js
```diff
@@ -131,6 +120,22 @@ const implicitHtmlRoles = {
...
+      case 'image':
+      case 'reset':
+      case 'submit':
+        return 'button';
+
+      case 'text':
+      case 'tel':
+      case 'url':
+      case 'email':
+      case '':
```
straker, line 134: According to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#htmlattrdeflist the `list` attribute is valid for more than what we have, so `input type="date" list="someId"` should be considered a combobox instead of a textbox. We should also add these to the case list to make it more complete

straker (changes requested):

lib/commons/standards/implicit-html-roles.js
WilcoFiers: It's not a combobox according to https://www.w3.org/TR/html-aria/#el-input-text-list

straker (approved):

Issue #2933, opened 2021-05-13
'aria-allowed-attr' fails instead of returning "needs review" when generic elements use 'aria-label' and 'aria-labelledby'


iamrafan:
```aria-allowed-attr``` fails a few use cases of generic elements using ```aria-label``` , ```aria-labelledby``` and returns ``` incomplete```  for a few other cases

Cases 1 to 6 in the examples below should return ```incomplete``` instead of a ```failure```

Demo link: https://codepen.io/iamrafan/full/abJdjYw

Product: axe-core

Expectation: Return all the usages of ARIA 1.2 prohibited states and properties on generic elements as ```needs review```

Motivation: Let the users review the issues instead of failing them

<pre><code>
axe-core version: axe-core 4.2.0-canary.43394bc
axe version: axe-coconut 4.8.6.11818
</code></pre>

**Comment by WilcoFiers** [comment]
@iamrafan Thank you for reporting the issue. I'm not sure how that slipped by us but I'm going to try to have a fix ready and released ASAP.

**Comment by iamrafan** [comment]
@WilcoFiers Thanks for taking this on, as a high priority

**Comment by WilcoFiers** [comment]
To be clear, there is a problem if you put `aria-label` on a div. screen readers are going to ignore it. That's what this update was designed to catch, but we overlooked the reach of this change for other elements.

**Comment by padmavemulapati** [comment]
Verified with the axe-core-develop branch code base with axe-core4.2.0
'aria-allowed-attr' failing  when generic elements use 'aria-label' and 'aria-labelledby', on the following elements

"audio,applet, canvas, dl, embed, iframe, input, label, meter, object, svg, video" with the below test snippet:

```
<div id="target" aria-label="hello world"></div>
  <div id="target" role="code" aria-label="hello world"></div>
  <span aria-label="value" id="one"></span>
<strong aria-label="value" id="two"></strong>
<kbd aria-label="value" id="three"></kbd>
<abbr aria-label="value" id="four"></abbr>
<custom-elm aria-label="value" id="five"></custom-elm>
<audio aria-label="value" id="pass1" controls></audio>
<iframe aria-label="value" id="pass2" src="/test/playground.html"></iframe>
<canvas aria-label="value" id="pass3"></canvas>
<dl aria-label="value" id="pass4"></dl>
<input aria-label="value" id="pass5"/>
<label aria-label="value" id="pass6"></label>
<meter aria-label="value" id="pass7"></meter>
[...]

PR #2935, opened 2021-05-17
fix(aria-allowed-attr): pass aria-label on some HTML elements

WilcoFiers:
Allow aria-label and aria-labelledby on the following elements, if those elements have no role.
- audio
- applet
- canvas
- dl
- embed
- iframe
- input
- label
- meter
- object
- svg
- video

Closes issue:  #2933, #2926

I did a bunch of testing with JAWS, NVDA and VoiceOver to come to this list, here are my test results:
https://codepen.io/wilcofiers/pen/WNpwMNj?editors=1000

"label" is the only one that isn't supported that I added anyway, because if you do `<label aria-label="foo" for="bar"></label>`, that label will be used in the accessible name of whatever form control has `id="bar"`. `dl` is on the list (unlike ul, and ol) because it doesn't have an implicit role.


doc/check-options.md
```diff
@@ -74,6 +75,43 @@ All checks allow these global options:
...
+  "applet",
+  "canvas",
+  "dl",
+  "embed",
+  "iframe",
+  "input",
+  "label",
+  "meter",
+  "object",
+  "span",
```
straker, line 105: I don't think `span` should be allowed

lib/checks/aria/aria-prohibited-attr-evaluate.js
```diff
@@ -26,42 +26,40 @@ import standards from '../../standards';
...
-  const role = getRole(virtualNode);
-  const attrs = virtualNode.attrNames;
-  const prohibitedAttrs = role
-    ? standards.ariaRoles[role].prohibitedAttrs
-    : ['aria-label', 'aria-labelledby'];
-
-  if (!prohibitedAttrs) {
+function ariaProhibitedAttrEvaluate(node, options = {}, virtualNode) {
+  const { elementsAllowedAriaLabel = [] } = options;
+  const attrBlacklist = getAttrBlacklist(virtualNode, elementsAllowedAriaLabel);
```
straker, line 31: We should avoid the term blacklist.

```suggestion
  const prohibitedAttrs = getProhibitedAttrs(virtualNode, elementsAllowedAriaLabel);
```

test/checks/aria/aria-prohibited-attr.js
```diff
@@ -64,4 +64,20 @@ describe('aria-prohibited-attr', function() {
     );
     assert.isFalse(checkEvaluate.apply(checkContext, params));
   });
+
+  it('should allow whitelisted nodes to have aria-label', function () {
```
straker, line 68: We should also avoid the term whitelist

```suggestion
  it('should allow passed nodes to have aria-label', function () {
```

test/checks/aria/aria-prohibited-attr.js
```diff
@@ -64,4 +64,20 @@ describe('aria-prohibited-attr', function() {
...
+    var params = checkSetup(
+      '<div id="target" aria-label="hello world"></div>',
+      { elementsAllowedAriaLabel: ['div'] }
+    );
+    assert.isFalse(checkEvaluate.apply(checkContext, params));
+  });
+  
+  it('should not allow whitelisted nodes with a prohibited role', function () {
```
straker, line 76:
```suggestion
  it('should not allow passed nodes with a prohibited role', function () {
```

lib/checks/aria/aria-prohibited-attr-evaluate.js
```diff
@@ -26,42 +26,40 @@ import standards from '../../standards';
...
+  if (!attrBlacklist) {
     return false;
   }
-
-  for (let i = 0; i < attrs.length; i++) {
-    const attrName = attrs[i];
+  
+  const prohibited = virtualNode.attrNames.filter(attrName => {
```
straker, line 36: Instead of looping over all attributes but only caring about the ones that are in the prohibited list, it would be faster to just look at the prohibited list.

(Sorry, multiline suggestion wasn't working for these lines)

```suggestion
  const prohibited = prohibitedAttrs.filter(attrName => {
```

lib/checks/aria/aria-prohibited-attr.json
```diff
@@ -1,11 +1,28 @@
...
+      "applet",
+      "canvas",
+      "dl",
+      "embed",
+      "iframe",
+      "input",
+      "label",
+      "meter",
+      "object",
+      "span",
```
straker, line 16: Ditto for `span` being allowed

straker (dismissed):

lib/checks/aria/aria-prohibited-attr-evaluate.js
WilcoFiers: ... fair enough.

doc/check-options.md
WilcoFiers, line 105:
```suggestion
```

straker (approved):

andrewleith:
I'm seeing the same issue on an `<i>` element.  Is this a bug as well or is it not allowed to put the an `aria-label` attribute on this tag?

Markup that is failing:
```
<i tabindex="0" class="remove icon" aria-label="clear field"></i>
```

Message from Axe:
> Elements must only use allowed ARIA attributes

straker:
@andrewleith `<i>` elements do not allow `aria-label`. Using `<i>` in this way is one of the common examples of why this rule was created since screen readers do not announce the `aria-label` on the element.

andrewleith:
Interesting.  Thanks for the info @straker :)
Are these rules written formally somewhere? It appears the markup I provided passes the NU validator but warns on the W3c one...

straker:
I'm not sure what you mean by formally, but [Leonie Watson wrote an excellent article](https://www.tpgi.com/short-note-on-aria-label-aria-labelledby-and-aria-describedby/) on this behavior, which was the basis of this rule.

melaniephilipp:
@straker @WilcoFiers Someone on the A11y Slack (external) channel today mentioned getting an error for aria-labelledby on a <video> element. Has this fix been pushed out to axe-core / axe browser extension yet?

doc/check-options.md
zelliott: @straker Are you able to link to the specific section of the ARIA spec that disallows `aria-label` on `span`? I found https://www.w3.org/TR/html-aria/, but I can't find anything that disallows it.

straker: So there is no spec documented reason for this (although it's coming). You can see where we got most of the concept in this pr https://github.com/dequelabs/axe-core/pull/2764. As it stands though, it's also a mix of a best practice since no screen reader supports it (https://github.com/dequelabs/axe-core/pull/2712#issuecomment-759732976).

Issue #2939, opened 2021-05-17
Run the frame-tested rule without respondable


WilcoFiers:
Currently frame-tested rule uses respondable to check if a frame responds when it is pinged, assuming that if a frame responds, it will be able to run axe-core. This isn't necessarily the case, for instance axe-core could time out in that frame and frame-tested would not be able to report it.

The way we can do this is by using an `after` method. If in addition to selecting all frames, we select the root of a page in this rule, we can check in the after method if every frame also has a root node. For example, if in the after method I get the following results, I can tell that `#frame1` was tested and has no iframes, but `#frame2` was not tested, since there are no results for it:

```JSON
[
  [["html"], ["#frame1"], ["#frame2"]]
  [["#frame1", "html"]
]
```

padmavemulapati:
Validated with the latest axe-core/develop branch code base,
'iframe' having `axe-core` running without responding 
with test script:
```
<!DOCTYPE html>
<html lang="en">
	<title>iframe</title>
	<body>
		<div id='mainConent' role='banner'>	
			<h2>I am an iframe with another iframe in it</h2>
			<h3>Example of an inaccessible List - 4 violations</h3>
			<div id="myList">  
				<dt>Coffee</dt>
				<dd>Black hot drink</dd>
				<dt>Milk</dt>
				<dd>White cold drink</dd>
			</div>	
			<br>
			<div id="myNestedIFrame">
				<iframe src="pageasnestediframe.html" name="nestedTestFrame" id="nestedTestFrame" height="200" width="300" title="I am a nested iframe with 2 violations">
          <script src="/axe.js"></script>
        </iframe>
			</div>	
		</div>
	</body>
    <script src="/axe.js"></script>
  <script>
    window.addEventListener('load', function() {
      axe.run(document, function(err, res) {
        console.log(res.violations);
        res.incomplete.forEach(function(issue) {
          console.log(issue);
        });
      });
    });
  </script>
</html>
```
![image](https://user-images.githubusercontent.com/47409579/124607937-ee041f00-de8b-11eb-90ce-6251bfdee3f0.png)

PR #2942, opened 2021-05-17
fix(frame-tested): run without respondable

straker:
Be able to run the `frame-tested` rule without using respondable.

Closes issue: #2939

### DIFF
--- a/lib/checks/media/frame-tested-after.js
+++ b/lib/checks/media/frame-tested-after.js
@@ -1,0 +1,29 @@
+const joinStr = ' > ';
+
+function frameTestedAfter(results) {
+  const iframes = {};
+
+  return results.filter(result => {
+    const frameResult =
+      result.node.ancestry[result.node.ancestry.length - 1] !== 'html';
+
+    if (frameResult) {
+      iframes[result.node.ancestry.join(joinStr)] = result;
+      return true;
+    }
+
+    // remove the `html` from the path to get the iframe path
+    const ancestry = result.node.ancestry
+      .slice(0, result.node.ancestry.length - 1)
+      .join(joinStr);
+
+    // pass for each iframe that has an html result
+    if (iframes[ancestry]) {
+      iframes[ancestry].result = true;
+    }
+
+    return false;
+  });
+}
+
+export default frameTestedAfter;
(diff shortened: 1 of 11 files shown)


WilcoFiers (dismissed):

lib/checks/media/frame-tested-after.js
```diff
@@ -0,0 +1,29 @@
...
+
+    if (frameResult) {
+      iframes[result.node.ancestry.join(joinStr)] = result;
+      return true;
+    }
+
+    // remove the `html` from the path to get the iframe path
+    const ancestry = result.node.ancestry
+      .slice(0, result.node.ancestry.length - 1)
+      .join(joinStr);
```
WilcoFiers, line 18: `ancestry` can be a nested array. We can't just join them.

lib/checks/media/frame-tested-after.js
```diff
@@ -0,0 +1,29 @@
+const joinStr = ' > ';
+
+function frameTestedAfter(results) {
+  const iframes = {};
+
+  return results.filter(result => {
```
WilcoFiers, line 6: Well I'll be damned! I did not know you could do a filter in `after`.

WilcoFiers (dismissed):

test/checks/media/frame-tested.js
```diff
@@ -1,52 +1,192 @@
...
+            ancestry: ['html > body > iframe#2', 'html']
+          }
+        }
+      ];
 
-      checkEvaluate.call(checkContext, iframe);
+      var afterResults = frameTestedAfter(results);
+      assert.lengthOf(afterResults, 2);
+      assert.isTrue(afterResults[0].result);
+      assert.isTrue(afterResults[1].result);
```
WilcoFiers, line 55: Probably better to also assert that these results are on the correct node.

test/checks/media/frame-tested.js
```diff
@@ -1,52 +1,192 @@
...
+          node: {
+            ancestry: [['html > body > custom-elm1', 'iframe#2'], 'html']
+          }
+        }
+      ];
+
+      var afterResults = frameTestedAfter(results);
+      assert.lengthOf(afterResults, 2);
+      assert.isTrue(afterResults[0].result);
+      assert.isTrue(afterResults[1].result);
```
WilcoFiers, line 129: Maybe add one that's failing?

WilcoFiers (changes requested):

WilcoFiers (approved):

Issue #2859, opened 2021-04-01
presentation-role-conflict should not error when element has no implicit role


straker:
The following code should not throw a presentation-role-conflict error as the element does not have an implicit role.

```html
<div role="none" tabindex="1"></div>
```

padmavemulapati:
Validated with latest develop branch code base,
observed `presentation-role-conflict` is appearing in `inapplicable` when tested with
`<div id="inapplicable1" role="presentation">Test</div> `
![image](https://user-images.githubusercontent.com/47409579/123645619-1ebcd680-d844-11eb-9802-857ed607606b.png)

PR #2948, opened 2021-05-19
fix(2859) presentation role conflict

clottman:
Adds a matcher to the presentation-role-conflict rule so that only elements that have an implicit role that could potentially conflict with an explicit role should be tested against the rule.

Since `div`s don't have an implicit role and thus should not be matched by the rule, two of the passing tests had to be removed.

Closes issue: #2859

### DIFF
--- a/lib/commons/aria/get-role.js
+++ b/lib/commons/aria/get-role.js
@@ -82,7 +82,7 @@ function getInheritedRole(vNode, explicitRoleOptions) {
   return getInheritedRole(vNode.parent, explicitRoleOptions);
 }
 
-function resolveImplicitRole(vNode, includeChromiumRoles, explicitRoleOptions) {
+function resolveImplicitRole(vNode, { includeChromiumRoles, ...explicitRoleOptions }) {
   const implicitRole = getImplicitRole(vNode, { includeChromiumRoles });
 
   if (!implicitRole) {
@@ -120,7 +120,7 @@ function hasConflictResolution(vNode) {
  */
 function resolveRole(
   node,
-  { noImplicit, includeChromiumRoles, ...explicitRoleOptions } = {}
+  { noImplicit, ...roleOptions } = {}
 ) {
   const vNode =
     node instanceof AbstractVirtuaNode ? node : getNodeFromTree(node);
@@ -133,7 +133,7 @@ function resolveRole(
   if (!explicitRole) {
     return noImplicit
       ? null
-      : resolveImplicitRole(vNode, includeChromiumRoles, explicitRoleOptions);
+      : resolveImplicitRole(vNode, roleOptions);
   }
 
   if (!['presentation', 'none'].includes(explicitRole)) {
@@ -145,7 +145,7 @@ function resolveRole(
     // has been set as the explicit role is not the true role
     return noImplicit
       ? null
-      : resolveImplicitRole(vNode, includeChromiumRoles, explicitRoleOptions);
+      : resolveImplicitRole(vNode, roleOptions);
   }
 
   // role presentation or none and no conflict resolution
--- a/lib/commons/aria/get-role.js
+++ b/lib/commons/aria/get-role.js
@@ -82,8 +82,10 @@ function getInheritedRole(vNode, explicitRoleOptions) {
   return getInheritedRole(vNode.parent, explicitRoleOptions);
 }
 
-function resolveImplicitRole(vNode, { includeChromiumRoles, ...explicitRoleOptions }) {
-  const implicitRole = getImplicitRole(vNode, { includeChromiumRoles });
+function resolveImplicitRole(vNode, { chromium, ...explicitRoleOptions }) {
+  const implicitRole = getImplicitRole(vNode, {
+    chromium
+  });
 
   if (!implicitRole) {
     return null;
@@ -118,22 +120,17 @@ function hasConflictResolution(vNode) {
  * @returns {string|null} Role or null
  * @deprecated noImplicit option is deprecated. Use aria.getExplicitRole instead.
  */
-function resolveRole(
-  node,
-  { noImplicit, ...roleOptions } = {}
-) {
+function resolveRole(node, { noImplicit, ...roleOptions } = {}) {
   const vNode =
     node instanceof AbstractVirtuaNode ? node : getNodeFromTree(node);
   if (vNode.props.nodeType !== 1) {
     return null;
   }
 
-  const explicitRole = getExplicitRole(vNode, explicitRoleOptions);
+  const explicitRole = getExplicitRole(vNode, roleOptions);
 
   if (!explicitRole) {
-    return noImplicit
-      ? null
-      : resolveImplicitRole(vNode, roleOptions);
+    return noImplicit ? null : resolveImplicitRole(vNode, roleOptions);
   }
 
   if (!['presentation', 'none'].includes(explicitRole)) {
@@ -143,9 +140,7 @@ function resolveRole(
   if (hasConflictResolution(vNode)) {
     // return null if there is a conflict resolution but no implicit
     // has been set as the explicit role is not the true role
-    return noImplicit
-      ? null
-      : resolveImplicitRole(vNode, roleOptions);
+    return noImplicit ? null : resolveImplicitRole(vNode, roleOptions);
   }
 
   // role presentation or none and no conflict resolution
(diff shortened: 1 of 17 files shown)


lib/rules/presentation-role-conflict-matches.js
```diff
@@ -0,0 +1,7 @@
+import { getImplicitRole } from '../commons/aria';
+
+function presentationRoleConflictMatches(node) {
```
WilcoFiers, line 3: Can you use the virtualNode here instead? We've been moving our rules over to exclusively use virtualNode.  Tests can run dramatically faster if the DOM isn't involved, and not all situations where axe-core is used needs to build a complete DOM tree.

lib/rules/presentation-role-conflict-matches.js
```diff
@@ -0,0 +1,7 @@
...
+  return getImplicitRole(node) !== null;
```
WilcoFiers, line 4: I think you're on the right track here, but we may need to exclude certain elements. This reminded me of #2935, where we had to account for elements that had a non-standard implicit role. The following for example probably should still be flagged by this rule:

```html
<svg role="none" aria-hidden="false">
  <title>Hello world</title>
</svg>
```

Kinda makes me wonder if I should have put that list of elements with non-standard roles somewhere that can be accessed outside the `aria-prohibited-attr` check. Maybe we need to add a toggle to `getImplicitRole()` which can return non-standard roles too? Something like:

```js
assert.isNull( getImplicitRole(svgElm) );
assert.equal( getImplicitRole(svgElm, { chromiumRoles: true ), 'SVGRoot') );
```

I think calling it `chromiumRoles` is fairly clear, and we'd have a "single source of truth" of what to return. Browsers have different role names for these elements. If we are going to do this, we'll need to make the `chromiumRoles` flag available through `getRole` as well, and we can use this in `aria-prohibited-attr`.

WilcoFiers (changes requested):

lib/rules/presentation-role-conflict-matches.js
clottman: I appreciate the callout here, I wasn't aware that certain elements could have a non-standard implicit role. Are those roles documented anywhere that you know of online, or was it something that mostly surfaced through testing?

WilcoFiers (changes requested):

lib/commons/aria/implicit-role.js
```diff
@@ -25,16 +38,14 @@ function implicitRole(node) {
     );
   }
 
-  // until we have proper implicit role lookups for svgs we will
-  // avoid giving them one
-  if (node && node.namespaceURI === 'http://www.w3.org/2000/svg') {
-    return null;
-  }
```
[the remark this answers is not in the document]
clottman: It looked like an unnecessary check - we don't have implicit role lookups for svg, so it will return null without this check based on the lines that follow. 

If we left it in, we'd have to add another check here for whether to use the chromium roles, and return the svg's implicit chromium role or else null. That's the same logic as is below.

lib/commons/aria/implicit-role.js
```diff
@@ -25,16 +38,14 @@ function implicitRole(node) {
...
-
   const nodeName = vNode.props.nodeName;
   const role = implicitHtmlRoles[nodeName];
 
   if (!role) {
+    if (chromiumRoles) {
```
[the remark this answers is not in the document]
clottman: Were you thinking something like this? 
```
  if (!role && chromiumRoles) {
    const chromiumRole = implicitChromiumRoles[nodeName];
    return chromiumRole || null;
  }

  if (typeof role === 'function') {
    return role(vNode);
  }

  return role || null;
}
```

lib/commons/aria/implicit-role.js
```diff
@@ -2,6 +2,19 @@ import implicitHtmlRoles from '../standards/implicit-html-roles';
 import { getNodeFromTree } from '../../core/utils';
 import AbstractVirtuaNode from '../../core/base/virtual-node/abstract-virtual-node';
 
+const implicitChromiumRoles = {
```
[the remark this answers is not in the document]
clottman: Would that be preferable to including it on standards/implicit-html-roles, which is currently what `getImplicitRole` uses?

lib/commons/aria/get-role.js
```diff
@@ -82,8 +82,8 @@ function getInheritedRole(vNode, explicitRoleOptions) {
   return getInheritedRole(vNode.parent, explicitRoleOptions);
 }
 
-function resolveImplicitRole(vNode, explicitRoleOptions) {
-  const implicitRole = getImplicitRole(vNode);
+function resolveImplicitRole(vNode, includeChromiumRoles, explicitRoleOptions) {
```
WilcoFiers, line 85: Why split this into two arguments? How about:

```suggestion
function resolveImplicitRole(vNode, { includeChromiumRoles, ...explicitRoleOptions }) {
```

lib/commons/aria/get-role.js
```diff
@@ -118,7 +118,10 @@ function hasConflictResolution(vNode) {
  * @returns {string|null} Role or null
  * @deprecated noImplicit option is deprecated. Use aria.getExplicitRole instead.
  */
-function resolveRole(node, { noImplicit, ...explicitRoleOptions } = {}) {
+function resolveRole(
+  node,
+  { noImplicit, includeChromiumRoles, ...explicitRoleOptions } = {}
```
WilcoFiers, line 123:
```suggestion
  { noImplicit, ...roleOptions } = {}
```

lib/commons/aria/get-role.js
```diff
@@ -128,7 +131,9 @@ function resolveRole(node, { noImplicit, ...explicitRoleOptions } = {}) {
   const explicitRole = getExplicitRole(vNode, explicitRoleOptions);
 
   if (!explicitRole) {
-    return noImplicit ? null : resolveImplicitRole(vNode, explicitRoleOptions);
+    return noImplicit
+      ? null
+      : resolveImplicitRole(vNode, includeChromiumRoles, explicitRoleOptions);
```
WilcoFiers, line 136:
```suggestion
      : resolveImplicitRole(vNode, roleOptions);
```

lib/commons/aria/get-role.js
```diff
@@ -138,7 +143,9 @@ function resolveRole(node, { noImplicit, ...explicitRoleOptions } = {}) {
   if (hasConflictResolution(vNode)) {
     // return null if there is a conflict resolution but no implicit
     // has been set as the explicit role is not the true role
-    return noImplicit ? null : resolveImplicitRole(vNode, explicitRoleOptions);
+    return noImplicit
+      ? null
+      : resolveImplicitRole(vNode, includeChromiumRoles, explicitRoleOptions);
```
WilcoFiers, line 148:
```suggestion
      : resolveImplicitRole(vNode, roleOptions);
```

lib/commons/aria/get-role.js
```diff
@@ -158,6 +165,7 @@ function resolveRole(node, { noImplicit, ...explicitRoleOptions } = {}) {
  * @param {boolean} options.abstracts  Allow role to be abstract
  * @param {boolean} options.dpub  Allow role to be any (valid) doc-* roles
  * @param {boolean} options.noPresentational return null if role is presentation or none
+ * @param {boolean} options.includeChromiumRoles Include implicit roles from chromium-based browsers in role result
```
WilcoFiers, line 168: This looks a little out of place with the rest. Just calling it `.options.chormium` seems like it would be more consistent with the other options here.

lib/standards/html-elms.js
```diff
@@ -601,7 +607,8 @@ const htmlElms = {
   },
   meter: {
     contentTypes: ['phrasing', 'flow'],
-    allowedRoles: false
+    allowedRoles: false,
+    chromiumRole: 'progressbar'
```
WilcoFiers, line 611: Should this be capitalised? If it isn't in Chromium then not, but it seems inconsistent.

lib/standards/html-elms.js
```diff
@@ -914,7 +923,8 @@ const htmlElms = {
         contentTypes: ['embedded', 'phrasing', 'flow']
       }
     },
-    allowedRoles: ['application']
+    allowedRoles: ['application'],
+    chromiumRole: 'video'
```
WilcoFiers, line 927: Same here. Why is it `Audio` but not `Video`?

lib/rules/has-implicit-chromium-role-matches.js
```diff
@@ -0,0 +1,7 @@
+import { getImplicitRole } from '../commons/aria';
+
+function hasImplicitChromiumRoleMatches(node, virtualNode) {
+  return getImplicitRole(virtualNode) !== null;
```
WilcoFiers, line 4: This still needs to use the new option. We'll need tests for it too.

WilcoFiers (changes requested):

lib/standards/html-elms.js
clottman: Yep, they are all inconsistent in Chrome(ium)

lib/standards/html-elms.js
clottman: Same answer - it's inconsistently reported in the browser.

WilcoFiers (approved):

WilcoFiers:
Reviewed for security.

Issue #2690, opened 2020-12-15
Is this Landmark Best Practice Issue accurate, or did axe make an overcall?


goodwitch:
Issue:  I ran axe on this page http://test.austinfree.net/ (as a Judge for Knowbility's AIR competition).  
The following Best Practice issue is being thrown:  'All page content must be contained by landmarks"
on the video player that is within an iframe that is within `<main>`.

Expectation: I was wondering if the iframe inside main was enough to count this player as being within a landmark, or if additional landmark need to be added within the iframe. 

<pre><code>
axe-core version:4.0.2
axe-extension: version 4.6.0

Browser and Assistive Technology versions:
Firefox: 83.0

For Tooling issues:
- Platform:  Mac 11.0.1
</code></pre>

WilcoFiers:
You are correct. That needs to be fixed. Looks like we didn't properly fix #2330 @straker. Sorry Glenda, we'll get it resolved ASAP.

svinkle:
@WilcoFiers Any update to this? We're running into the same issue with YouTube embeds elsewhere. Thanks.

padmavemulapati:
Validated on test page: `http://test.austinfree.net/`
able to reproduce using axe-expert, seeing `Best Practice` issue on `region` saying "All page content must be contained by landmarks"
![image](https://user-images.githubusercontent.com/47409579/125097909-358ce400-e0f4-11eb-97c5-c4caa13f3eaa.png)

and seeing the fix in the latest develop branch code- axe-core (9th July,2021)
not seeing any issue either `violation` or `Best-practice` on `region` within iframe that is within <main>.
![image](https://user-images.githubusercontent.com/47409579/125098368-94eaf400-e0f4-11eb-903f-d1f933230bcd.png)

WilcoFiers:
This could have been closed some time ago.

PR #2949, opened 2021-05-19
fix(region): contents in iframes should pass the region rule if the iframe itself is in a region

clottman:
If there's an iframe on the page, elements within that iframe should all be treated as being contained in a region if the iframe element was contained on the iframe.

Also adds lots of tests for this behavior. 

Closes issue: #2690


lib/rules/region.json
```diff
@@ -1,6 +1,7 @@
 {
   "id": "region",
   "selector": "body *",
+  "matches": "is-initiator-matches",
```
WilcoFiers, line 4: This is also going to cause valid region issues to be ignored. The trick to figure out is how to distinguish these three examples:

```html
<!-- failing iframe -->
<iframe srcdoc="<h1>hello world</h1>"></iframe>

<!-- passing iframe 1 -->
<main>
  <iframe srcdoc="<h1>hello world</h1>"></iframe>
</main>

<!-- passing iframe 2 -->
  <iframe srcdoc="<main>
    <h1>hello world</h1>
</main>"></iframe>
```

Axe-core currently fails passing iframe 1, your PR fixes that, but it does so by also passing the failed iframe. I think what we'd need to do is to pass information about which frames are in a region to a new after method, and then in the after method to pass all the content of iframes that are in a region. That'll need to be recursive too.

WilcoFiers (changes requested):

lib/checks/navigation/region-after.js
```diff
@@ -0,0 +1,24 @@
...
+      if (r.result) {
+        iframesInRegions.push(r.node.ancestry);
+      }
+      // pass every iframe element
+      r.result = true;
+    } else if (!r.result) {
+      // check if any node that failed is contained in an iframe that's in a region
+      // if it does, the node should pass
+      r.result = iframesInRegions.some(iframe => {
+        return iframe.every((value, index) => value === r.node.ancestry[index]);
```
WilcoFiers, line 16: I probably should have thought of this yesterday, but this isn't going to work for iframes nested in shadow DOM trees. There the `value` can be another array, not just a string.

lib/checks/navigation/region-evaluate.js
```diff
@@ -75,6 +75,11 @@ function findRegionlessElms(virtualNode, options) {
 
 function regionEvaluate(node, options, virtualNode) {
   let regionlessNodes = cache.get('regionlessNodes');
+
+  this.data({
```
WilcoFiers, line 79: This function is creating problems for us. We're relying on it to tell us if `iframe` elements are part of a landmark, and it isn't doing it accurately. Take the following:

```html
<iframe srcdoc="<p>Hello</p>"></iframe>
<div><iframe srcdoc="<p>Hello</p>"></iframe></div>
```

Because axe-core only fails the outer-most element in the first example the iframe will be failed, but in the second example, the iframe is passed, and the `div` is failed instead. We're probably going to need to special-case iframes here, check if they have an ancestor that is in `regionlessNodes`, and pass that information into `this.data()`. Since we're doing that, we then might as well revert the regionMatcher change. We'd use the data prop to tell if an iframe is in a region, rather than its result.

lib/checks/navigation/region-after.js
```diff
@@ -0,0 +1,24 @@
+function regionAfter(results) {
```
WilcoFiers, line 1: I think we need to create some unit tests for this after method. That'll be much easier than trying to pack all the logic of this into integration tests.

WilcoFiers (changes requested):

test/checks/navigation/region.js
```diff
@@ -290,14 +290,6 @@ describe('region', function() {
     assert.isTrue(checkEvaluate.apply(checkContext, checkArgs));
   });
 
-  it('treats iframe elements as regions', function() {
```
clottman, line 293: This is now done in the after method, so this test won't pass any more.

WilcoFiers (dismissed):

lib/core/utils/match-ancestry.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Check if two ancestries are identical
+ */
+function matchAncestry(ancestryA, ancestryB) {
```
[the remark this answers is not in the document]
clottman: From what I can tell, this should already be available through axe.utils because it's exported from core/utils/index.js, which core/core.js imports everything from and makes available as axe.utils. 

I can add tests for it though.

test/checks/navigation/region-after.js
```diff
@@ -0,0 +1,197 @@
...
+      {
+        data: { isIframe: false },
+        node: {
+          ancestry: ['html > body > iframe', 'html > body > p']
+        },
+        result: false
+      }
+    ]);
+
+    assert.deepEqual(result, [
```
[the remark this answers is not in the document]
clottman: Can do, I just copied what was in a nearby `after` test.

lib/checks/navigation/region-evaluate.js
clottman: With the updated code, the `p` in the iframe is what will fail in both of these cases.

lib/rules/region.json
clottman: For this case: 
```
<!-- inapplicable comment -->
<iframe srcdoc="<-- empty frame -->"></iframe>
``` 

We can't mark this as inapplicable because we don't know the iframe is empty until after the rule has already run.

lib/checks/navigation/region-after.js
```diff
@@ -0,0 +1,30 @@
+import { matchAncestry } from '../../core/utils';
+
+function regionAfter(results) {
+  results.forEach((r, index) => {
+    // this element failed the check
+    if (!r.result) {
+      // we need to see if it's actually contained in an iframe that passed the check
+      const ancestryMinusLast = r.node.ancestry.slice(0, -1);
```
WilcoFiers, line 8:
```suggestion
      const frameAncestry = r.node.ancestry.slice(0, -1);
```

lib/checks/navigation/region-after.js
```diff
@@ -0,0 +1,30 @@
...
+
+      // we only need to check up to index - 1 because the iframe this element might be contained in will have been found before this element
+      for (const earlierResult of results.slice(0, index)) {
```
WilcoFiers, line 11: This feels unnecessarily complex. I suggest you create a list of frames with `result:true`, and then try to match against that. You have no business running `matchAncestry` against anything other than iframes. You don't need to do it if ancestry has a length of 1 either.

That also lets you use clearer variable names. It took me a while to figure out what `earlierResult` was even for.

WilcoFiers (changes requested):
Some cleanup. Rest looks good.

lib/checks/navigation/region-after.js
```diff
@@ -0,0 +1,30 @@
+import { matchAncestry } from '../../core/utils';
+
+function regionAfter(results) {
+  results.forEach((r, index) => {
+    // this element failed the check
+    if (!r.result) {
```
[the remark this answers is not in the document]
clottman: We're inside a forEach here, so we don't want to `return` and we can't `break`.

lib/checks/navigation/region-after.js
clottman: That strategy won't work because we need to be able to evaluate nested iframes correctly - when an iframe is contained in another iframe that should pass, but the inner iframe does not pass on its own, we don't know until we've iterated sequentially through that the inner iframe should pass and thus be considered part of the set that should be matched against. 

Consider this (which is represented in the full integration tests as region-pass-nested-iframe.html): 
```
<main>
  <iframe id='outer' srcdoc="<iframe id='inner' srcdoc='<p>hello</p>'></iframe></iframe>
</main>
```

Here's the code I was working on, which has the problem described above, if you want to play around with it yourself.
```js
function regionAfter(results) {
  const passingFrameAncestries = results.filter(
    r => r.result && r.data.isIframe
  ).map(r => r.node.ancestry);

  results.forEach((r, index) => {
    // this element failed the check, and it's in an iframe
    if (!r.result && r.node.ancestry.length > 1) {
      // we need to see if it's actually contained in an iframe that passed the check
      const frameAncestry = r.node.ancestry.slice(0, -1);

      for (const frame of passingFrameAncestries) {
        if (matchAncestry(frameAncestry, frame)) {
          r.result = true;
          break;
        }
      }
    }
  });

  // iframe elements should always pass
  results.forEach(r => {
    if (!r.result && r.data.isIframe) {
      r.result = true;
    }
  });
  return results;
}
```

test/checks/navigation/region-after.js
```diff
@@ -0,0 +1,120 @@
...
+        },
+        result: false
+      }
+    ]);
+
+    assert.equal(results[0].result, true);
+    assert.equal(results[1].result, true);
+    assert.equal(results[2].result, true);
+  });
+});
```
WilcoFiers, line 120: There's a test missing from this, one that I strongly suspect fails.

```suggestion
});

  it('should pass content if an grandparent frame passes', function() {
    var results = checks.region.after([
      {
        data: { isIframe: true },
        node: {
          ancestry: ['html > body > iframe']
        },
        result: true
      },
      {
        data: { isIframe: true },
        node: {
          ancestry: ['html > body > iframe', 'html > body > iframe']
        },
        result: false
      },
      {
        data: { isIframe: false },
        node: {
          ancestry: [
            'html > body > iframe',
            'html > body > iframe',
            'html > body > p'
          ]
        },
        result: false
      }
    ]);

    assert.equal(results[0].result, true);
    assert.equal(results[1].result, true);
    assert.equal(results[2].result, true);
  });
```

WilcoFiers (changes requested):

test/checks/navigation/region-after.js
clottman: There's actually an integration test for this already - region-pass-nested-iframe.html. It's the reason I had to slightly modify your suggested code when I made region-after easier to understand.

WilcoFiers (approved):

PR #2951, opened 2021-05-20
fix(meta-viewport) Don't check zooming/scaling ability in iframe-embedded pages

### LINKED ISSUE #2210: Zooming and scaling must not be disabled on iframed page
Is it truly an accessibility issue if a page exists only for being iframed and it has user-scaling disabled? We are getting that issue, and do not have access to the source page, and I'm just trying to figure out how I would scale an iframe if I wanted to.

clottman:
Adds a matcher to the meta-viewport rules so that only the top-level/initiator page is checked to see if zooming and scaling is available appropriately.

Closes #2210

### DIFF
--- a/lib/checks/mobile/meta-viewport.json
+++ b/lib/checks/mobile/meta-viewport.json
@@ -1,7 +1,6 @@
 {
   "id": "meta-viewport",
   "evaluate": "meta-viewport-scale-evaluate",
-  "matches": "is-initiator-matches",
   "options": {
     "scaleMinimum": 2
   },
--- a/lib/checks/mobile/meta-viewport.json
+++ b/lib/checks/mobile/meta-viewport.json
@@ -1,6 +1,7 @@
 {
   "id": "meta-viewport",
   "evaluate": "meta-viewport-scale-evaluate",
+  "matches": "is-initiator-matches",
   "options": {
     "scaleMinimum": 2
   },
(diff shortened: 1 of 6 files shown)


lib/checks/mobile/meta-viewport.json
```diff
@@ -1,6 +1,7 @@
 {
   "id": "meta-viewport",
   "evaluate": "meta-viewport-scale-evaluate",
+  "matches": "is-initiator-matches",
```
WilcoFiers, line 4: Checks don't have matches methods. This isn't going to do anything.

```suggestion
```

WilcoFiers (changes requested):
Can you add some integration tests for these rules?

test/integration/rules/meta-viewport/meta-viewport-pass3.html
```diff
@@ -0,0 +1,21 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta id="mvp" name="viewport" content="maximum-scale=2;" />
+  </head>
+
+  <body>
+    <div id="mocha"></div>
+    <!-- iframe with zooming disabled should not fail the rule-->
+    <iframe
```
WilcoFiers, line 10: Unfortunately, IE doesn't support `srcdoc`. You'll have to add axe.js to the frame content, otherwise axe won't run there.

WilcoFiers (changes requested):

WilcoFiers (approved):

WilcoFiers:
I have reviewed this PR for security.

PR #2952, opened 2021-05-20
fix(aria-allowed-attr): allow aria-posinset and aria-setsize on row elements for treegrids

### LINKED ISSUE #2720: "aria-setsize" is not allowed in "treegrid"
When I use axe chrome extension in the official aria practice (https://www.w3.org/TR/wai-aria-practices-1.1/examples/treegrid/treegrid-1.html), there are some "critical" issues.

Expectation:
No issues are announced in axe

Actual: 
"aria-setsize" is not allowed in "treegrid"

Motivation:
This should not be announced as an "issue"

<pre><code>
axe-core version: 4.1.1
axe-webdriver, extension or other integration version: axe chrome extension v4.6.2

- Platform:  Mac, Chrome v87.0.4280.88
</code></pre>

clottman:
Adds aria-posinset and aria-setsize to the allowed aria attributes for row elements, in accordance with the [treegrid examples](https://www.w3.org/TR/wai-aria-practices-1.1/examples/treegrid/treegrid-1.html). 

See a live test case [on glitch](https://aria-treegrid-example.glitch.me/), which runs a local copy of axe that's been built with these changes. 

When I run this locally, the test I added isn't getting picked up and I'm not sure why - might need to pair on that. 

Closes #2720

### DIFF
--- a/lib/commons/aria/lookup-table.js
+++ b/lib/commons/aria/lookup-table.js
@@ -1637,9 +1637,7 @@ lookupTable.role = {
         'aria-level',
         'aria-selected',
         'aria-rowindex',
-        'aria-errormessage',
-        'aria-setsize',
-        'aria-posinset'
+        'aria-errormessage'
       ]
     },
     owned: {
--- a/lib/commons/aria/lookup-table.js
+++ b/lib/commons/aria/lookup-table.js
@@ -1637,7 +1637,9 @@ lookupTable.role = {
         'aria-level',
         'aria-selected',
         'aria-rowindex',
-        'aria-errormessage'
+        'aria-errormessage',
+        'aria-setsize',
+        'aria-posinset'
       ]
     },
     owned: {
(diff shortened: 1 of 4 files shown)


lib/commons/aria/lookup-table.js
```diff
@@ -1637,7 +1637,9 @@ lookupTable.role = {
         'aria-level',
         'aria-selected',
         'aria-rowindex',
-        'aria-errormessage'
+        'aria-errormessage',
+        'aria-setsize',
+        'aria-posinset'
```
WilcoFiers, line 1642: The lookup table has been deprecated in favor of using `lib/standards`. We don't need to update it anymore.  It'll just be removed in 5.0.

```suggestion
        'aria-errormessage'
```

WilcoFiers (changes requested):

WilcoFiers (approved):

WilcoFiers:
Reviewed for security.

Issue #2928, opened 2021-05-12
Runtime issue: Cannot read property 'data' of undefined


MikeHarris105:
<< I ran axe on the whole page of a geckoboard sharing url : https://share.geckoboard.com/dashboards/ZRJLCO4PAYMCXRTW >>

axe-core version: `4.2.0`
axe-extension version: `4.9.3`

```
TypeError: Cannot read property 'data' of undefined
    at Xl._l [as after] (chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:56:289593)
    at chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:56:324563
    at Array.forEach (<anonymous>)
    at ec.after (chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:56:324396)
    at chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:56:332384
    at Array.map (<anonymous>)
    at e.value (chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:56:332240)
    at chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:56:341091
    at chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:56:97118
    at chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:56:100990
```

WilcoFiers:
@MikeHarris105 Thank you for reporting. I'll pass this along to the appropriate people.

buckett:
We've also seen this issue when scanning an application.

WilcoFiers:
We're working on a fix for this. Should be out soon. Thank you for your patience.

WilcoFiers:
This issue is a duplicate. We believe this has been resolved.

Issue #2944, opened 2021-05-18
Runtime issue: Cannot read property 'data' of undefined


vavrapavel:
Ran a full scan on instance of an internally hosted enterprise application Polarion (GUI done via GWT): https://polarion.plm.automation.siemens.com/products/polarion-alm

axe-core version: `4.2.0`
axe-extension version: `4.9.3`

```
TypeError: Cannot read property 'data' of undefined
    at Xl._l [as after] (chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:56:289593)
    at chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:56:324563
    at Array.forEach (<anonymous>)
    at ec.after (chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:56:324396)
    at chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:56:332384
    at Array.map (<anonymous>)
    at e.value (chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:56:332240)
    at chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:56:341091
    at chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:56:97118
    at chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/vendor.bundle.js:56:100990
```

straker:
Thanks for the issue. We'll look into it.

srayos:
Good morning! I am an engineer with Dell Technologies and our teams are experiencing this error, too. Can you please provide us with a status update? Is there any work-around at this point? Thank you!

gdkraus:
We are experiencing the same issue with some of our internally hosted apps, but not all of our apps.

gdkraus:
If it helps in diagnosing this, for us it seems to happen with apps with iframes, but not all apps with iframes. If I load two of our internal apps which have the same structure for the content in the parent frame but the single iframe in the app is different, if the iframe has a lot of contents the error occurs. If the iframe doesn't have much content it doesn't happen. When I get a chance I'll try to debug it further for the root cause in our apps.

Note, all of this definitely started happening with one of the latest axe devtools releases as we have never seen this error before.

gdkraus:
When the error occurs, in the console I get the following message.

[...]

PR #2965, opened 2021-05-29
fix(heading-order): Crash on page with iframes but no headings

WilcoFiers:
This PR is a rewrite of heading-order-after. It fixes a number of issues:

- When a window has frames, but no headings the headings should still be placed in the heading order. (This caused the errors of #2928 and #2944)
- The first result no longer automatically passes. It can fail if a heading before it was excluded, and there is a level-skip between them.
- Ancestor selectors in shadow DOM trees should be correctly recognised
- If for some reason the heading can't be found in the heading map, for example because of a custom selector, the rule no longer errors.

Closes issue: #2928 and #2944


gdkraus:
@WilcoFiers, in reference to #2944 the description of what this change solves does not describe our situation. In our app, we have an iframe that has multiple headings in it.

gdkraus:
@WilcoFiers, or do you mean if the outermost frame contains no headings, but contains iframes, that is when the problem occurs?

gdkraus:
I manually edited our parent frame to have a single heading and ran axe DevTools again and it worked.

straker (approved):

PR #2970, opened 2021-06-02
chore: don't tag releases

straker:
I believe this was causing our changelogs to always be incorrect and require manual editing. When we run `standard-version`, it automatically creates a tag for the next version. However that tag is never pushed to the remote and we instead have our CI script create a tag for us after the PR has been merged. This causes the tag on the local machine to be out-of-sync with the tag in the remote.

By skipping the tagging step of standard-version and forcing the release script to fetch all tags before creating a release, this should allow standard-version to find the correct tagged commit and create the changelog properly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,9 @@
   "standard-version": {
     "scripts": {
       "postbump": "npm ci && npm run sri-update"
+    },
+    "skip": {
+      "tag": true
     }
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -87,7 +90,7 @@
     "test:rule-help-version": "mocha test/test-rule-help-version.js",
     "test:node": "mocha test/node/*.js",
     "version": "echo \"use 'npm run release' to bump axe-core version\" && exit 1",
-    "release": "standard-version -a",
+    "release": "git fetch origin --tags --force && standard-version -a",
```
stephenmathieson, line 93: Why `--force`?

straker: https://stackoverflow.com/a/55268620/2124254

straker: For now anyone whose done the release script will have incorrect tags, so best we solve that problem too

stephenmathieson (approved):

Issue #2966, opened 2021-05-30
Contrast issues of button with pseudo element styles


davodavodavo3:
Product: Axe Extension, LH

**Expectation:** 
LH/Axe should recognize that there is a :before pseudo element with background color of #000000 and in this case the button text contrast is enough 21.

**Actual:** 
Background and foreground colors do not have a sufficient contrast ratio.
Element has insufficient color contrast of 1.07

LH/Axe are checking directly the button styles where background color is transparent, and they says that contrast of white text is not enough.
![image](https://user-images.githubusercontent.com/10462563/120120603-4bcb8a00-c1af-11eb-9f68-dceec8a46293.png)

Motivation: Because it's showing wrong information

Browser and Assistive Technology versions
Affected Channels: DevTools
Axe version: 4.1.2
Lighthouse version: 7.2.0
Chrome version: 91.0.4472.77
Node.js version:
Operating System: Windows 10

davodavodavo3:
Steps to reproduce
Run LH Accessibility audit on https://theskinexperts.co.uk
Check Contrast issues related to the black button with white text

WilcoFiers:
Unfortunately, there are no APIs to get at computed properties of pseudo elements. That makes testing them challenging. Axe-core has some code that tried to figure out the impact pseudo elements can have on color. We'll have to check why in this particular scenario it does not.

straker:
Looks like we [start the pseudo check at the parent element](https://github.com/dequelabs/axe-core/blob/develop/lib/checks/color/color-contrast-evaluate.js#L102-L115) of the node and never look at the node itself.

davodavodavo3:
Great, thank you guys. Do I need to inform Google LH about this so they put some notification there about that color contrast can be not correct?

padmavemulapati:
Validated with latest develop branch code base,
color-contrast is checking now for pseudo elements (node level check not just if any of the element's parents have pseudo elements)
![image](https://user-images.githubusercontent.com/47409579/124585351-d2d8e580-de72-11eb-8798-0b1f3aeb2773.png)

PR #2980, opened 2021-06-04
fix(color-contrast): check for pseudo elements on element itself, not just parents

clottman:
We should return elements as Needs Review/incomplete if there is a pseudo element on the element, not just if any of the element's parents have pseudo elements. This PR makes that the case. 

Closes issue: #2966


WilcoFiers (approved):

WilcoFiers:
Reviewed for security

padmavemulapati:
Validated with latest develop branch code base,
color-contrast is checking now for pseudo elements (node level check not just if any of the element's parents have pseudo elements)
image

PR #2985, opened 2021-06-10
fix(color-contrast): add special case for new sr-only technique

### LINKED ISSUE #2962: color-contrast false positive on element styled as screen-reader-only
Product: axe-core

Expectation: Elements using the following sr-only CSS should be detected as inapplicable to the color-contrast rule:

```css
.sr-only {
  position: absolute;
  width: 1px;
  height: 1px;
  margin: -1px;
  padding: 0px;
  border: 0px;
  overflow: hidden;
}
```

[Gist containing a full self-contained repro](https://gist.github.com/dbjorge/9cbc74299092bbba4e33e3d2d47766a0)

Actual: Elements with that `sr-only` styling can fail the `color-contrast` rule.

Motivation: An Accessibility Insights for Web user reported this as a false positive.

Notes:

As a workaround, we've recommended that the user add a `clip: rect(0,0,0,0);` property to their `sr-only` class, similar to Bootstrap's `sr-only` implementation. This causes axe-core to understand that the element should be ineligible for the `color-contrast` test.

However, the item is still invisible without the `clip` property (the combination of other properties is sufficient), so it would be ideal for axe-core to understand that.

This technically caused a false positive, but one that has a straightforward workaround and which is straightforward for a user to "fix" by adding noop color CSS according to the color-contrast instructions, so I'd consider this to be lower priority than most other false positive issues.

<pre><code>
axe-core version: 4.2.1
axe-webdriver, extension or other integration version: Accessibility Insights for Web 2.26.1

Browser and Assistive Technology versions: Microsoft Edge Version 90.0.818.66 (Official build) (64-bit), no AT

For Tooling issues:
- Node version: n/a
- Platform:  n/a
</code></pre>

clottman:
The minimum CSS that makes the element from the example visually hidden is a combination of: 
```css
overflow: hidden;
position: absolute;
height: 1px;
```

So if the user is doing that, they are probably trying to hide the element visually and axe-core should recognize that.

Note that the repro case will not fail in Firefox even without this code. 
 
Closes #2962

### DIFF
--- a/lib/commons/dom/is-visible.js
+++ b/lib/commons/dom/is-visible.js
@@ -163,7 +163,7 @@ function isVisible(el, screenReader, recursed) {
       style.getPropertyValue('opacity') === '0' ||
       (getScroll(el) && elHeight === 0) ||
       (style.getPropertyValue('position') === 'absolute' &&
-        elHeight <= 1 &&
+        elHeight < 2 &&
         style.getPropertyValue('overflow') === 'hidden'))
   ) {
     return false;
--- a/lib/commons/dom/is-visible.js
+++ b/lib/commons/dom/is-visible.js
@@ -157,14 +157,20 @@ function isVisible(el, screenReader, recursed) {
 
   // hidden from visual users
   const elHeight = parseInt(style.getPropertyValue('height'));
+
+  // ways to hide content visually
+  const scrollableWithZeroHeight = getScroll(el) && elHeight === 0;
+  const posAbsoluteOverflowHiddenAndSmall =
+    style.getPropertyValue('position') === 'absolute' &&
+    elHeight < 2 &&
+    style.getPropertyValue('overflow') === 'hidden';
+
   if (
     !screenReader &&
     (isClipped(style) ||
       style.getPropertyValue('opacity') === '0' ||
-      (getScroll(el) && elHeight === 0) ||
-      (style.getPropertyValue('position') === 'absolute' &&
-        elHeight < 2 &&
-        style.getPropertyValue('overflow') === 'hidden'))
+      scrollableWithZeroHeight ||
+      posAbsoluteOverflowHiddenAndSmall)
   ) {
     return false;
   }
--- a/lib/commons/dom/is-visible.js
+++ b/lib/commons/dom/is-visible.js
@@ -156,11 +156,15 @@ function isVisible(el, screenReader, recursed) {
   }
 
   // hidden from visual users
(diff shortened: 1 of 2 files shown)


WilcoFiers (dismissed):
Can you add and/or update a test to go with this?

lib/commons/dom/is-visible.js
```diff
@@ -156,11 +156,15 @@ function isVisible(el, screenReader, recursed) {
   }
 
   // hidden from visual users
+  const elHeight = parseInt(style.getPropertyValue('height'));
   if (
     !screenReader &&
     (isClipped(style) ||
       style.getPropertyValue('opacity') === '0' ||
-      (getScroll(el) && parseInt(style.getPropertyValue('height')) === 0))
+      (getScroll(el) && elHeight === 0) ||
```
WilcoFiers, line 164: Is there a reason you left this at 0?

lib/commons/dom/is-visible.js
```diff
@@ -156,11 +156,15 @@ function isVisible(el, screenReader, recursed) {
...
+      (style.getPropertyValue('position') === 'absolute' &&
+        elHeight <= 1 &&
+        style.getPropertyValue('overflow') === 'hidden'))
```
WilcoFiers, line 167: 6 line if statement's a bit much. Can you isolate some of the logic on this?

lib/commons/dom/is-visible.js
```diff
@@ -156,11 +156,15 @@ function isVisible(el, screenReader, recursed) {
...
   // hidden from visual users
+  const elHeight = parseInt(style.getPropertyValue('height'));
   if (
     !screenReader &&
     (isClipped(style) ||
       style.getPropertyValue('opacity') === '0' ||
-      (getScroll(el) && parseInt(style.getPropertyValue('height')) === 0))
+      (getScroll(el) && elHeight === 0) ||
+      (style.getPropertyValue('position') === 'absolute' &&
+        elHeight <= 1 &&
```
WilcoFiers, line 166: I would suggest we make this `< 2` to allow for some rounding above 1.
```suggestion
        elHeight < 2 &&
```

WilcoFiers (changes requested):

lib/commons/dom/is-visible.js
clottman: Yes, turns out it wasn't relevant to the case I was trying to address so I left it alone. Would you rather it be bumped up too?

WilcoFiers (approved):

PR #2989, opened 2021-06-11
fix(aria-roles): Mark as needs review if both none and presentation are used on element with no implicit role

### LINKED ISSUE #2909: Review fallback issues for role none/presentation on elements no implicit role
The following shouldn't be failed. It'd be better to mark it for review instead. It **might** be an issue if the div is between a `menu` and `menuitem`, but pretty much any other case this isn't going to matter.

```html
<div role="none presentation">
<div role="presentation none">
```

clottman:
Closes issue #2909.

### DIFF
--- a/test/checks/aria/fallbackrole.js
+++ b/test/checks/aria/fallbackrole.js
@@ -13,28 +13,55 @@ describe('fallbackrole', function() {
       '<div id="target" role="button foobar">Foo</div>'
     );
     assert.isTrue(
-      checks.fallbackrole.evaluate(virtualNode.actualNode, 'radio', virtualNode)
+      checks.fallbackrole.evaluate(virtualNode.actualNode, null, virtualNode)
     );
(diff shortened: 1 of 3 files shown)


test/checks/aria/fallbackrole.js
```diff
@@ -13,28 +13,55 @@ describe('fallbackrole', function() {
       '<div id="target" role="button foobar">Foo</div>'
     );
     assert.isTrue(
-      checks.fallbackrole.evaluate(virtualNode.actualNode, 'radio', virtualNode)
+      checks.fallbackrole.evaluate(virtualNode.actualNode, null, virtualNode)
```
clottman, line 16: the options param isn't actually used by the check, so it doesn't need to be in the test (unless there's a reason for it that I'm missing)

WilcoFiers (approved):

PR #2990, opened 2021-06-11
fix(docs): update readme to only include correct localization build flag command

### LINKED ISSUE #2649: locale specific axe.js not generating

clottman:
Only one of these commands actually works; they are not equivalent. 

Closes #2649

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,11 +95,7 @@ The [axe-core API](doc/API.md) package consists of:
 
 Axe can be built using your local language. To do so, a localization file must be added to the `./locales` directory. This file must have be named in the following manner: `<langcode>.json`. To build axe using this locale, instead of the default, run axe with the `--lang` flag, like so:
 
-`grunt build --lang=nl`
-
-or equivalently:
-
-`npm run build -- --lang=nl`
+`grunt build -- --lang=nl`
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -95,11 +95,7 @@ The [axe-core API](doc/API.md) package consists of:
 
 Axe can be built using your local language. To do so, a localization file must be added to the `./locales` directory. This file must have be named in the following manner: `<langcode>.json`. To build axe using this locale, instead of the default, run axe with the `--lang` flag, like so:
 
-`grunt build --lang=nl`
-
-or equivalently:
-
-`npm run build -- --lang=nl`
+`grunt build -- --lang=nl`
```
straker, line 98: We should keep the `npm` command in the docs since `grunt` only works if you have it installed globally. In fact, we should probably update the docs to either use the `npm` command or use `npx grunt` so these grunt commands work (if there's not an npm script available).

straker (changes requested):

README.md
clottman: Oh you know, I thought the line I removed was `npm run build --lang=nl`, not using grunt. The issue it seems users were having is trying to run `npm run build` without having those two extra dashes.

clottman:
This wasn't the change I thought I was making, so it's not accurate.

PR #2994, opened 2021-06-14
feat: Support multiple languages at once in builds

### LINKED ISSUE #2714: Unused and inconsistent code in build
Expectation: Axe either accepts multiple locale arguments for building, or does not have code in the gruntfile to parse a list of multiple locale arguments.

Actual: Axe build fails if you provide multiple locales in a delimited list, but the gruntfile parses them out as if it were intended to accept them.

Motivation:
Right now the gruntfile appears to think that you can and should take multiple language arguments in comma and/or semicolon separated lists, but the aria-supported file chokes on this really badly because it doesn't perform the "parse in the list" step.

I'd be inclined to fix this but I'm unclear on what resolution would be desired; fully support passing in multiple languages in (something)-delimited lists, or remove the unused code that parses them out from the gruntfile?

clottman:
Most of the code for this was already in place; only the aria-supported task was failing when you tried to enter multiple languages to build for. It doesn't seem like that task actually does anything different depending on the language, so this code uses the first language of the languages that were entered (if any) to find the right axe.**.js file to use to generate aria-supported.md. 

Closes issue #2714

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -110,6 +110,9 @@ module.exports = function(grunt) {
       data: {
         entry: 'lib/commons/aria/index.js',
         destFile: 'doc/aria-supported.md',
+        options: {
+          langs: langs
+        },
         listType: 'unsupported' // Possible values for listType: 'supported', 'unsupported', 'all'
       }
     },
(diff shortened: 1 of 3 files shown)


straker (approved):
Tested it and the command produces multiple axe scripts (unminified and minified) with the correct langs in the files.

Nice work.

Issue #2937, opened 2021-05-17
New utils.getFrameContexts method


WilcoFiers:
Required for #2936

This `axe.utils.getFrameContexts()` method takes a context, and returns an array with a `frameSelector` and `frameContext` property. This should work something like this:

```js
const context = {
  include: [
    ["#frame1", "header"],
    ["#frame2,#iframe3", "footer"]
  ]
}
assert.deepEqual(
  axe.utils.getFrameContexts(context),
  [{
    frameSelector: "#frame1",
    frameContext: { include: "header" }
  }, {
    frameSelector: "#frame2",
    frameContext: { include: "footer" }
  }, {
    frameSelector: "#frame3",
    frameContext: { include: "footer" }
  }]
);
```

This function must return one item per frame that is in context, providing a unique selector for each frame. This might be best done using the ancestry selector, which can be generated very quickly. This method can not rely on the virtual tree existing.

padmavemulapati:
Validated with the latest develop branch code base, 
observed the content fetched from the source for. iframe's context when tried to retrieve from 
`await axe.utils.getFrameContexts('#testFrame');` where `#testFrame` is the iframe selector of a test page 
`http://qateam.dequecloud.com/attest/api/test.html`

![image](https://user-images.githubusercontent.com/47409579/123617524-f7a2dc80-d824-11eb-952d-b3fcf783bb1c.png)

PR #2995, opened 2021-06-15
feat(utils): add getFrameContexts method

WilcoFiers:
Add a utility which is passed a context object, and returns the context for each iframe.

Closes issue: #2937

### DIFF
--- a/lib/core/base/context.js
+++ b/lib/core/base/context.js
@@ -42,29 +42,51 @@ function pushUniqueFrame(collection, frame) {
 function pushUniqueFrameSelector(context, type, selectorArray) {
   context.frames = context.frames || [];
 
-  var result, frame;
-  var frames = document.querySelectorAll(selectorArray.shift());
+  const frameSelector = selectorArray.shift();
+  const frames = document.querySelectorAll(frameSelector);
 
-  frameloop: for (var i = 0, l = frames.length; i < l; i++) {
-    frame = frames[i];
-    for (var j = 0, l2 = context.frames.length; j < l2; j++) {
-      if (context.frames[j].node === frame) {
-        context.frames[j][type].push(selectorArray);
-        break frameloop;
+  frames.forEach(frame => {
+    context.frames.forEach(contextFrame => {
+      if (contextFrame.node === frame) {
+        contextFrame[type].push(selectorArray);
       }
-    }
-    result = {
-      node: frame,
-      include: [],
-      exclude: []
-    };
+    });
 
-    if (selectorArray) {
-      result[type].push(selectorArray);
-    }
+    if (!context.frames.find(result => result.node === frame)) {
+      const result = {
+        node: frame,
+        include: [],
+        exclude: []
+      };
 
-    context.frames.push(result);
-  }
+      if (selectorArray) {
+        result[type].push(selectorArray);
+      }
+
+      context.frames.push(result);
+    }
+  });
+
+  // frameloop: for (var i = 0, l = frames.length; i < l; i++) {
+  //   frame = frames[i];
+  //   for (var j = 0, l2 = context.frames.length; j < l2; j++) {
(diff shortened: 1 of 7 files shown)


test/core/utils/get-frame-context.js
```diff
@@ -0,0 +1,182 @@
...
+      include: [['main']],
+      exclude: []
+    })
+    assert.deepEqual(contexts[2], {
+      include: [],
+      exclude: [['footer']]
+    })
+  });
+  
+  it.skip('mixes contexts if the frame is selected twice', function () {
```
WilcoFiers, line 81: There's a bug the presents this from passing. Need to fix that, but maybe in a separate PR.

lib/core/utils/get-frame-contexts.js
```diff
@@ -0,0 +1,11 @@
+import Context from '../base/context'
+import getAncestry from './get-ancestry'
+
+export default function getFrameContexts(context) {
+  const { frames } = new Context(context);
```
WilcoFiers, line 5: This creates `axe._tree`, which is going to slow stuff down. Can we avoid that? And if not, how do we avoid either a tree that is outdated by the time axe.runPartial is called, or generating the tree twice.

straker: Do we want to fix this in this pr or another one?

lib/core/base/context.js
```diff
@@ -42,29 +42,51 @@ function pushUniqueFrame(collection, frame) {
...
+  //     include: [],
+  //     exclude: []
+  //   };
+
+  //   if (selectorArray) {
+  //     result[type].push(selectorArray);
+  //   }
+
+  //   context.frames.push(result);
+  // }
```
WilcoFiers, line 89: I assume you didn't mean to leave that in.
```suggestion
```

lib/core/public/load.js
```diff
@@ -13,6 +13,7 @@ function runCommand(data, keepalive, callback) {
   };
 
   var context = (data && data.context) || {};
+  console.log(data.context);
```
WilcoFiers, line 16:
```suggestion
```

lib/core/utils/get-frame-contexts.js
WilcoFiers: Nope, we can let Context build axe._tree. I've accounted for it in the updated proposal.

test/core/utils/get-frame-context.js
```diff
@@ -0,0 +1,280 @@
...
+
+  it('skips frames excluded by a parent', function() {
+    fixture.innerHTML = '<iframe></iframe>';
+    var frameContexts = getFrameContexts({
+      exclude: [['#fixture']]
+    });
+    assert.lengthOf(frameContexts, 0);
+  });
+
+  it('normalizes the context', function() {
```
WilcoFiers, line 164: Can you check the selector of these frames, make sure we get the right one?

test/core/utils/get-frame-context.js
```diff
@@ -0,0 +1,280 @@
...
+    });
+    assert.lengthOf(contexts, 1);
+    assert.deepEqual(contexts[0], {
+      initiator: false,
+      include: [],
+      exclude: []
+    });
+  });
+
+  it('accepts elements', function() {
```
WilcoFiers, line 199: Same here, check the selector.

test/testframe1.html
```diff
@@ -0,0 +1,6 @@
+<!DOCTYPE html>
```
WilcoFiers, line 1: Delete please.

WilcoFiers (commented):
LGTM

michael-siek (approved):

WilcoFiers:
For the record: I passed this one off to Steve, who finished it up. Steve checked my work, and I checked his additions. Michael reviewed this for security.

Issue #2960, opened 2021-05-25
Restore autocomplete-valid act-mapping test


clottman:
Related to https://github.com/dequelabs/axe-core/pull/2917
When autocomplete-appropriate was changed, the act-mapping test for that had to be removed because axe-core got ahead of the ACT tests. Once ACT has been updated, we should put back the ACT test for autocomplete-appropriate.

WilcoFiers:
This can be resolved when https://github.com/act-rules/act-rules.github.io/pull/1583 is merged.

straker:
Closed via #2997

PR #2997, opened 2021-06-15
tests: put back autocomplete-valid ACT tests

straker:
https://github.com/act-rules/act-rules.github.io/pull/1583 was merged.

Closes issue: #2960

Awaiting https://github.com/act-rules/act-rules.github.io/pull/1653


straker:
Had to not use the CircleCi cache in order to get the latest version of act rules since updates don't modify the package-lock.

.circleci/config.yml
```diff
@@ -113,7 +113,8 @@ jobs:
     <<: *unix_box
     steps:
       - checkout
-      - <<: *restore_dependency_cache_unix
+      # Don't restore npm cache since we always need latest #master branch
+      - run: npm install
```
WilcoFiers, line 117: I'm not so sure this is the right thing to do here Steve.

WilcoFiers, line 117: What about doing it this way instead?

```suggestion
      - <<: *restore_dependency_cache_unix
      # Always run on the latest master branch, regardless of cache
      - run: npm install act-rules/act-rules.github.io#master
```

WilcoFiers (changes requested):

WilcoFiers (approved):

Issue #2815, opened 2021-02-19
Color contrast foreground color should account for text-shadow


indiamos:
Expectation: Tools should assess color contrast correctly and not return false issues

Actual: Text is incorrectly flagged for color contrast issues

Motivation: We'd like to be able to trust our tools and not spend time chasing down fake issues.

<pre><code>
axe-core version: 4.1.2
axe-webdriver, extension or other integration version: axe Chrome extension v4.7.2

Browser and Assistive Technology versions
- Platform:  MacOS 10.15.7
- Browser:  Chrome 88.0.4324.150 (Official Build) (x86_64)
</code></pre>

----------

When I check the page https://studio.onepeloton.com/account/classes using either axe Expert or the free axe extension, I get a “serious” contrast error on the label of this select element. The “Fix the following” instructions seem to think this element uses white text on a white background.

<img width="1156" alt="Axe contrast reading" src="https://user-images.githubusercontent.com/146094/108545216-86bb4900-72b5-11eb-8a52-534f24d4eadf.png">

If I check the same element using Colour Contrast Analyser (CCA) Version 3.1.1, there is no contrast issue.

<img width="1157" alt="CCA contrast reading" src="https://user-images.githubusercontent.com/146094/108545241-8de25700-72b5-11eb-9b94-9627fa3b9d81.png">

Lighthouse 6.4.0, which is still using Axe 3.5.5, also reports no contrast issues on this page.

straker:
Thanks for the issue. That link is behind a login. Is there another page on the site that I could look at?

indiamos:
Ah, sorry! The exact same issue happens on this page: https://studio.onepeloton.com/shop

straker:
Thanks, that one works and can replicate the issue. I'll look into it to see what's the issue.

straker:
Well look into I found the issue. The problem is that the color of the element is completely transparent (`rgba(0, 0, 0, 0)`) and the text color is actually coming from the text-shadow (`rgb(24 26 29) 0px 0px 0px`). This is the first time I've seen this done so we'll need to figure out what to do.

indiamos:
Ugh, how revolting; I hope it's also the _last_ time you see this done. While you work on that, I'll try to figure out _why_ we're doing it and whether we could just not…

Thanks!

straker:
[...]

PR #3001, opened 2021-06-15
Use text shadow as foreground color if foreground is transparent

clottman:
If the foreground color is transparent, use the text shadows as the effective foreground color for computing color contrast. 

This has some limitations because of how we're computing get-text-shadow-colors - if the offset (either x or y) is greater than the blur, we say the effective color of the text shadow for contributing to contrast is nothing/transparent. But, testing in the browser when the `color` of the text is transparent, the text appears to be the color of the text shadow even when a small offset is used. 

For example try: 
```
<div id="target" style="color: rgba(0,0,0,0); text-shadow: 1px 1px 0 rgb(32, 32, 64);">This is my text</div>
```

According to axe-core, the text of this element would be transparent, but you can see [in this codepen](https://codepen.io/clottman1/pen/WNpLPLO) that the text is readable. I'm not sure how we want to handle that.

Closes issue: #2815


lib/commons/color/get-foreground-color.js
```diff
@@ -63,6 +64,11 @@ function getForegroundColor(node, _, bgColor) {
     return null;
   }
 
+  if (fgColor.alpha === 0) {
+    const textShadowColors = getTextShadowColors(node, { minRatio: 0 });
```
WilcoFiers, line 68: This might hurt performance a bit. Can you work with @straker to figure out if this needs any performance optimisations?

lib/commons/color/get-foreground-color.js
```diff
@@ -63,6 +64,11 @@ function getForegroundColor(node, _, bgColor) {
     return null;
   }
 
+  if (fgColor.alpha === 0) {
```
WilcoFiers, line 67: I think we should probably blend text-shadow any time the alpha is < 1.

lib/commons/color/get-foreground-color.js
WilcoFiers, line 68: How is this effected by offset and blur? I would like to see more tests on that. If blur is 0, even a small offset will result in the shadow sitting next to the text, instead of blending with it. The greater the blur, the less that is the case, but then the effect on the text color is greatly reduced.

WilcoFiers (changes requested):

lib/commons/color/get-foreground-color.js
clottman: If blur is 0, and there is an offset greater than zero, currently `getTextShadowColors` treats that as not being a shadow that contributes to the text color at all.

clottman: He didn't think it would, since it's only happening when fgColor has an alpha of < 1, which is more rare.

lib/commons/color/get-foreground-color.js
clottman: We're currently blending the text-shadow into the background color most of the time if it's a thin shadow. Should we be blending it with the foreground color too, or only doing one or the other?

WilcoFiers: It can effect both, so I think it should be both.

lib/commons/color/get-foreground-color.js
WilcoFiers: We'll just need a test for that to make sure.

WilcoFiers (changes requested):

lib/commons/color/get-foreground-color.js
clottman: Added here. https://github.com/dequelabs/axe-core/pull/3001/commits/d3bf005a3c238c9b7e10c344b4d901d36dc98d2f

lib/commons/color/get-foreground-color.js
```diff
@@ -64,9 +64,9 @@ function getForegroundColor(node, _, bgColor) {
     return null;
   }
 
-  if (fgColor.alpha === 0) {
+  if (fgColor.alpha < 1) {
     const textShadowColors = getTextShadowColors(node, { minRatio: 0 });
-    return [...textShadowColors, bgColor].reduce(flattenColors);
+    return [fgColor, ...textShadowColors, bgColor].reduce(flattenColors);
```
WilcoFiers, line 69: Can you add a test that shows these three colors blend correctly?

WilcoFiers (dismissed):

lib/checks/color/color-contrast-evaluate.js
```diff
@@ -150,7 +156,8 @@ function colorContrastEvaluate(node, options, virtualNode) {
     fontSize: `${((fontSize * 72) / 96).toFixed(1)}pt (${fontSize}px)`,
     fontWeight: bold ? 'bold' : 'normal',
     messageKey: missing,
-    expectedContrastRatio: expected + ':1'
+    expectedContrastRatio: expected + ':1',
+    shadowColor: shadowColors.map(c => c.toHexString()).join(', ')
```
WilcoFiers, line 160: I think we should put the flattened shadow color in here. I thought I left that comment before, but can't seem to find it.

test/commons/color/get-text-shadow-colors.js
```diff
@@ -152,4 +168,28 @@ describe('axe.commons.color.getTextShadowColors', function() {
     assert.lengthOf(shadowColors, 1);
     assert.equal(shadowColors[0].blue, 255);
   });
+
+  it('returns a transparent shadow when x offset is greater than blur', function() {
+    fixture.innerHTML =
+      '<span style="text-shadow: ' + '1px 0 0 #F00' + '">Hello world</span>';
```
WilcoFiers, line 174: Prettier didn't know what you were on about. So lets keep it at:

```suggestion
      '<span style="text-shadow: 1px 0 0 #F00">Hello world</span>';
```

test/commons/color/get-text-shadow-colors.js
```diff
@@ -152,4 +168,28 @@ describe('axe.commons.color.getTextShadowColors', function() {
...
+    assert.lengthOf(shadowColors, 1);
+    assert.equal(shadowColors[0].red, 0);
+    assert.equal(shadowColors[0].green, 0);
+    assert.equal(shadowColors[0].blue, 0);
+    assert.equal(shadowColors[0].alpha, 0);
+  });
+
+  it('returns a transparent shadow when y offset is greater than blur', function() {
+    fixture.innerHTML =
+      '<span style="text-shadow: ' + '0 1px 0 #F00' + '">Hello world</span>';
```
WilcoFiers, line 186:
```suggestion
      '<span style="text-shadow: 0 1px 0 #F00">Hello world</span>';
```

WilcoFiers (changes requested):

lib/checks/color/color-contrast-evaluate.js
clottman: Made that change, but this will result in some strange seeming messages - 
```
  <div
    id="target"
    style="background-color: #aaa; color:#666; text-shadow: 1px 1px 0.01em #000"
  >
    Hello world
  </div>
  ```
  will report that the text shadow color is #aaa when the user would reasonably assume that the color should be #000. Do we need to update the failure messages somehow?


  ```
  Fix any of the following:\n  Element has insufficient color contrast of 2.47 between the foreground and shadow color (foreground color: #666666, text-shadow color: #aaaaaa, font size: 12.0pt (16px), font weight: normal). Expected contrast ratio of 4.5:1"
  ```

WilcoFiers (approved):

PR #3006, opened 2021-06-16
feat(runPartial): Test without frame communication

### LINKED ISSUE #2936: New runPartial and finishRun methods
Since axe-core's use of window.postMessage() can interfere with the operations, we need a way to run axe-core that does not require window.postMessage. Browser extensions can use `axe.frameMessenger()` for this, which was released in axe-core 4.2. However using frameMessenger in APIs usually isn't possible, which is why in axe-core 4.3 we're going to introduce an alternative method of running axe-core.

With axe.runPartial() and axe.finishRun it will be possible to split an axe-core run into two phases. In the first phase, all frames are tested separately, and the results are collected in the correct order in the script driving the browser. Then in the second phase, those results are combined into an axe-core report. This will look something like this:

```js
// Recursively wait for the frame to load, inject axe, then find nested frames
// and return each frame (including top)
const frames = await setupFrames(page, context);
const frameResults = await asyncMap(frames, async ({ frame, frameContext }) => {
  const initiator = frame === page.mainFrame(); 
  return frame.execute(() =>
     axe.runPartial({ ...frameContext, initiator }, options)
)})
const axeReport = await blankPage.execute(async () => {
  await frame.execute(axeSource);
  return axe.finishRun(frameResults, options)
})
```

WilcoFiers:
This adds two new methods to the `axe` object. `axe.runPartial` runs axe in the current window, and collects information about its iframes. `axe.finishRun` takes information from multiple `axe.runPartial` calls from different windows, builds a report from it.

Closes #2936


lib/core/public/finish-run.js
```diff
@@ -0,0 +1,40 @@
+import { getReporter } from './reporter';
+import { mergeResults, publishMetaData, finalizeRuleResult, DqElement } from '../utils';
+
+export default function finishRun(data, options = {}) {
```
straker, line 4: It wasn't clear what `data` was in this function, had to read the tests to understand it was an array of `runPartial` results.

test/integration/full/run-partial/after-method.js
```diff
@@ -0,0 +1,60 @@
...
+  
+  it('can pass with a complex "after" method', function (done) {
+    var options = { runOnly: ruleName };
+    var context = { exclude: [] }
+    Promise.all(
+      runPartialRecursive(clone(context), options)
+    ).then(function (partialResults) {
+      return Promise.all([
+        axe.finishRun(partialResults, options),
+        axe.run(clone(context), options)
```
straker, line 19: Was confused why you were running `axe.run` here. Looking further into the test I see you're making sure they are the same result. Might be better if this sort of test was first defined like:

"it(should give the same results as `axe.run`)"

straker (commented):
So far looking good

axe.d.ts
```diff
@@ -278,6 +278,26 @@ declare namespace axe {
...
 
+  /**
+   * TODO
+   * @param context
+   * @param options
+   */
+  function runPartial(
+    context: ElementContext,
+    options: RunOptions
+  ): Promise<unknown>;
```
WilcoFiers, line 289: Techdebt ticket here: https://github.com/dequelabs/axe-core/issues/3059

test/testutils.js
```diff
@@ -428,32 +428,34 @@ testUtils.isIE11 = (function isIE11(navigator) {
...
-
-  // remove all attributes from fixture (otherwise a leftover
-  // style attribute would cause avoid-inline-spacing integration
-  // test to fail with [#fixture] being included in the results)
-  var attrs = fixture.attributes;
-  for (var i = 0; i < attrs.length; i++) {
-    var attrName = attrs[i].name;
-    if (attrName !== 'id') {
-      fixture.removeAttribute(attrs[i].name);
+if (typeof beforeEach !== 'undefined' && typeof afterEach !== 'undefined') {
```
WilcoFiers, line 431: I put this in so that we can use testutils outside a test run. I was using it in playground.

straker (approved):

Issue #2925, opened 2021-05-08
Color-contrast rule does not correctly report on text shadow colors


WilcoFiers:
When a text has a text-shadow, axe-core will either report the contrast between the shadow and the text color, or the shadow and the background color. The resulting message should make it clear that this is the case. This may involve a few changes:

1. Axe-core needs to report the `shadowColor` property as part of its `data`, if there is a `shadowColor` worth reporting.
2. The `failed` message needs to be split up. We need to report which of the three contrast measurements axe-core used in that messages. We'd have the following message keys:
    - `fgOnBgColor": "Element has insufficient color contrast of ${data.contrastRatio} (foreground color: ${data.fgColor}, background color: ${data.bgColor}, font size: ${data.fontSize}, font weight: ${data.fontWeight}). Expected contrast ratio of ${data.expectedContrastRatio}",`
    - `fgOnShadowColor": "Element has insufficient color contrast of ${data.contrastRatio} (foreground color: ${data.fgColor}, text-shadow color: ${data.shadowColor}, font size: ${data.fontSize}, font weight: ${data.fontWeight}). Expected contrast ratio of ${data.expectedContrastRatio}",`
    - `shadowOnBgColor": "Element has insufficient color contrast of ${data.contrastRatio} (text-shadow color: ${data.shadowColor}, background color: ${data.bgColor}, font size: ${data.fontSize}, font weight: ${data.fontWeight}). Expected contrast ratio of ${data.expectedContrastRatio}",`

Reporting the issue like this would make it clear why in some cases, even if the fg / bg color is sufficient, the rule can still fail.

PR #3010, opened 2021-06-16
Use more specific failure messages for color contrast failures

clottman:
Builds off the other color contrast work that's been happening, so base of the PR is not `develop`
Closes issue: #2925

### DIFF
--- a/lib/checks/color/color-contrast.json
+++ b/lib/checks/color/color-contrast.json
@@ -23,8 +23,8 @@
       "pass": "Element has sufficient color contrast of ${data.contrastRatio}",
       "fail": {
         "default": "Element has insufficient color contrast of ${data.contrastRatio} (foreground color: ${data.fgColor}, background color: ${data.bgColor}, font size: ${data.fontSize}, font weight: ${data.fontWeight}). Expected contrast ratio of ${data.expectedContrastRatio}",
-        "fgOnShadowColor": "Element has insufficient color contrast of ${data.contrastRatio} (foreground color: ${data.fgColor}, text-shadow color: ${data.shadowColor}, font size: ${data.fontSize}, font weight: ${data.fontWeight}). Expected contrast ratio of ${data.expectedContrastRatio}",
-        "shadowOnBgColor": "Element has insufficient color contrast of ${data.contrastRatio} (text-shadow color: ${data.shadowColor}, background color: ${data.bgColor}, font size: ${data.fontSize}, font weight: ${data.fontWeight}). Expected contrast ratio of ${data.expectedContrastRatio}"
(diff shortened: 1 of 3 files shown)


lib/checks/color/color-contrast.json
```diff
@@ -21,7 +21,11 @@
     "impact": "serious",
     "messages": {
       "pass": "Element has sufficient color contrast of ${data.contrastRatio}",
-      "fail": "Element has insufficient color contrast of ${data.contrastRatio} (foreground color: ${data.fgColor}, background color: ${data.bgColor}, font size: ${data.fontSize}, font weight: ${data.fontWeight}). Expected contrast ratio of ${data.expectedContrastRatio}",
+      "fail": {
+        "fgOnBgColor": "Element has insufficient color contrast of ${data.contrastRatio} (foreground color: ${data.fgColor}, background color: ${data.bgColor}, font size: ${data.fontSize}, font weight: ${data.fontWeight}). Expected contrast ratio of ${data.expectedContrastRatio}",
+        "fgOnShadowColor": "Element has insufficient color contrast of ${data.contrastRatio} (foreground color: ${data.fgColor}, text-shadow color: ${data.shadowColor}, font size: ${data.fontSize}, font weight: ${data.fontWeight}). Expected contrast ratio of ${data.expectedContrastRatio}",
```
straker, line 26: These are very subtle differences in the text, we might want to improve the message itself (not just the numbers we display) to tell the user that shadow text is being applied rather than having the user glean that from the text inside the parenthesis.

lib/checks/color/color-contrast.json
```diff
@@ -21,7 +21,11 @@
     "impact": "serious",
     "messages": {
       "pass": "Element has sufficient color contrast of ${data.contrastRatio}",
-      "fail": "Element has insufficient color contrast of ${data.contrastRatio} (foreground color: ${data.fgColor}, background color: ${data.bgColor}, font size: ${data.fontSize}, font weight: ${data.fontWeight}). Expected contrast ratio of ${data.expectedContrastRatio}",
+      "fail": {
+        "fgOnBgColor": "Element has insufficient color contrast of ${data.contrastRatio} (foreground color: ${data.fgColor}, background color: ${data.bgColor}, font size: ${data.fontSize}, font weight: ${data.fontWeight}). Expected contrast ratio of ${data.expectedContrastRatio}",
```
straker, line 25: Any time there's an object with multiple choices for the message, one should always be labeled `default`. In this case, the `fgOnBgColor` should be `default` and then you don't need to assign it a value for the message text as it can be assumed this case when `null`.

straker (changes requested):

straker (approved):

PR #3015, opened 2021-06-17
docs: clarify axe.run options paramter does not modify rules or checks

straker:
The docs implied that `axe.run` could change rules or check metadata (like passing a check option) when it doesn't. Changed it to not imply that.

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -307,7 +307,7 @@ axe.run(context, options, (err, results) => {
 #### Parameters axe.run
 
 - [`context`](#context-parameter): (optional) Defines the scope of the analysis - the part of the DOM that you would like to analyze. This will typically be the `document` or a specific selector such as class name, ID, selector, etc.
-- [`options`](#options-parameter): (optional) Set of options which temporarily changes how axe runs. This contrasts with `axe.configure`, which is more permanent.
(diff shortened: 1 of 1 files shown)


doc/API.md
```diff
@@ -307,7 +307,7 @@ axe.run(context, options, (err, results) => {
 #### Parameters axe.run
 
 - [`context`](#context-parameter): (optional) Defines the scope of the analysis - the part of the DOM that you would like to analyze. This will typically be the `document` or a specific selector such as class name, ID, selector, etc.
-- [`options`](#options-parameter): (optional) Set of options passed into rules or checks, temporarily modifying them. This contrasts with `axe.configure`, which is more permanent.
+- [`options`](#options-parameter): (optional) Set of options which temporarily changes how axe runs. This contrasts with `axe.configure`, which is more permanent.
```
clottman, line 310: How about this?
```suggestion
- [`options`](#options-parameter): (optional) Set of options that change how `axe.run` works, including what rules will run. To pass options to specific checks, use `axe.configure`.
```

straker: I like it.

clottman (approved):

stephenmathieson (approved):

Issue #2781, opened 2021-01-26
Pseudo element contrast failure too invasive?


cchaos:
👋  

I've run into a situation where `:nth-child` style pseudo selectors are wrongly being identified as pseudo **elements** and therefore, can't assume contrast acceptance. I've tracked down to this section here, but it does seem to only be checking for `:before` and `:after`s. Maybe something in the function being called?

https://github.com/dequelabs/axe-core/blob/f2f15377c917d0b6fc27d2181f64ebef26e5a9b0/lib/checks/color/color-contrast-evaluate.js#L103-L116

Anyhow, here's also a screenshot showing the run through using the Chrome plugin and the result:
<img width="1452" alt="Screen Shot 2021-01-26 at 14 25 39 PM" src="https://user-images.githubusercontent.com/549577/105895360-6e704b00-5fe3-11eb-9b8c-b44c78bbd2a7.png">

myasonik:
In case it helps someone investigate, I found this change happened in this PR: https://github.com/dequelabs/axe-core/pull/2613

Which was released in 4.1.0 and then got rolled out to the browser plugins with 4.1.1

straker:
Interesting. Is there a live site I could look at? It looks like at some point in that elements parent chain a `:before` or `:after` is declared. Looking at the related node it belongs to the `.euiSideNavItem_items` class

myasonik:
The example above is from https://eui.elastic.co

straker:
The before-mentioned class has this css on it which is causing the message:

```css
.euiSideNavItem__items:after {
    position: absolute;
    content: '';
    top: 0;
    bottom: 12px;
    width: 1px;
    background: #D3DAE6;
    left: 0;
}
```

since the content is `''` it is being flagged as having a pseudo element with background color.

cchaos:
🤦  Good catch on there actually being a pseudo element. The element just also happened to have a parent selector that uses a pseudo selector.

The pseudo element is the tiny little border the connects our text to an indent line which is hidden until the previous item opens.

<img width="334" alt="Screen Shot 2021-01-26 at 18 24 17 PM" src="https://user-images.githubusercontent.com/549577/105919522-c8810880-6003-11eb-9535-07169b023653.png">

So then I guess the question is, is the current "failure" too invasive? This is a trick we use a lot in CSS to position, primarily graphical (non-critical) elements and the errors are then overwhelming. Is it possible to ignore pseudo elements when their content is an empty string?

[...]

PR #3016, opened 2021-06-17
feat(color-contrast) ignore hidden pseudo elements, add option to ignore pseudo elements entirely

clottman:
Two things: 
 - pseudo elements with visibility:hidden or display:none aren't counted as potentially contributing to color contrast and thus requiring we return undefined
 - adds an option to `ignorePseudo` which bypasses the determination of if there are pseudo elements contributing to color contrast that would mean we should return undefined

Closes issue: #2781


doc/check-options.md
```diff
@@ -205,6 +205,7 @@ All checks allow these global options:
 | ----------------------------------------------------------- | :------ | :------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `ignoreUnicode`                                             | `true`  | Do not check the color contrast of Unicode characters                                                                                                                                        |
 | `ignoreLength`                                              | `false` | Do not check the color contrast of short text content                                                                                                                                        |
+| `ignorePseudo`                                              | `false` | Do not check the color contrast of elements with pseudo-element styles                                                                                                                       |
```
straker, line 208: Since we don't really check them for contrast in the first place.

```suggestion
| `ignorePseudo`                                              | `false` | Do not mark pseudo elements as Needs Review                                                                                                                        |
```

lib/checks/color/color-contrast-evaluate.js
```diff
@@ -16,17 +16,19 @@ import {
 } from '../../commons/color';
 import { memoize } from '../../core/utils';
 
-const hasPsuedoElement = memoize(function hasPsuedoElement(node, pseudo) {
+const hasPseudoElement = memoize(function hasPseudoElement(node, pseudo) {
```
straker, line 19: Good catch. I can never spell this word correctly the first time...

straker (dismissed):

straker (approved):

Issue #2849, opened 2021-03-25
Fix flakey respondable test


straker:
This test fails a lot and is fixed by a rerun https://app.circleci.com/pipelines/github/dequelabs/axe-core/1939/workflows/554f50b9-a8c3-4141-b489-49366390836d/jobs/28872.

straker:
https://app.circleci.com/pipelines/github/dequelabs/axe-core/2626/workflows/97788640-c92f-4cfe-847e-9f27c176d1b4/jobs/34636

straker:
https://app.circleci.com/pipelines/github/dequelabs/axe-core/2626/workflows/d546c64c-157c-48e1-9338-0155bd879408/jobs/34643

PR #3021, opened 2021-06-18
test(respondable): add retries

WilcoFiers:
Attempt to stop this test from failing half the time.

Related #2849


straker:
I'm not sure this is a true fix for the problem. I'd be OK fixing this using `retries` as a temporary fix, but It'd be better to figure out why it's so flaky as it has failed 3 times in a row before on restarts.

straker (approved):

PR #3022, opened 2021-06-18
fix(accText): ignore text in embedded content elements

### LINKED ISSUE #3017: axe returns fallback text of `video` elements as the accessible name
Product: axe-core
Expectation: The accessible name calculation should not return the fallback text for `<video>` elements.
Actual: Accessible name calculation includes fallback text
Motivation: Screen readers to not appear to have this as the accessible name, nor do the accessibility trees of different browsers that I checked.

<pre><code>axe-core version: 4.2.2
axe-webdriver, extension or other integration version: N/A

(see included test case)
</code></pre>

I've setup a [codepen with different scenarios](https://codepen.io/scurker/full/GRWeZWe) of alternate text for `<video>` elements.

```js
await import('https://unpkg.com/axe-core')
axe.utils.getFlattenedTree()
$$('video').map(axe.commons.text.accessibleText)
```

The above returns the following:

```js
[
  "", 
  "this is a video", 
  "this is a label for the video", 
  "", 
  "Your browser is not supported"
]
```

The last video in the above test case is one with a fallback:

```html
<video>
  <source src="..." />
  Your browser is not supported
</video>
```

I don't think the accessible name calculation should be included except in instances where the `video` element is not supported, but I have not tested that specific use case. It's possible that this happens for `audio` elements as well.

According to the [html spec](https://html.spec.whatwg.org/multipage/media.html#the-video-element) the content of the elements is intended _not_ to address accessibility concerns.

WilcoFiers:
Content inside embedded elements like `video`, `canvas`, `object`, etc. should be ignored in computing the accessible name.

Closes  issue #3017

### DIFF
--- a/lib/standards/html-elms.js
+++ b/lib/standards/html-elms.js
@@ -674,7 +674,7 @@ const htmlElms = {
     noAriaAttrs: true
   },
   picture: {
-    contentTypes: ['embedded', 'phrasing', 'flow'],
+    contentTypes: ['phrasing', 'flow'],
     allowedRoles: false,
     noAriaAttrs: true
   },
--- a/lib/standards/html-elms.js
+++ b/lib/standards/html-elms.js
@@ -674,7 +674,7 @@ const htmlElms = {
     noAriaAttrs: true
   },
   picture: {
-    // In HTML, this is considered "embedded". For axe-core we don't want this since its children will not be hidden
+    // Note: spec change (do not count as embedded), because browsers do not hide text inside the picture element
     contentTypes: ['phrasing', 'flow'],
     allowedRoles: false,
     noAriaAttrs: true
--- a/lib/standards/html-elms.js
+++ b/lib/standards/html-elms.js
@@ -674,6 +674,7 @@ const htmlElms = {
     noAriaAttrs: true
   },
   picture: {
+    // In HTML, this is considered "embedded". For axe-core we don't want this since its children will not be hidden
     contentTypes: ['phrasing', 'flow'],
     allowedRoles: false,
     noAriaAttrs: true
--- a/test/checks/shared/has-visible-text.js
+++ b/test/checks/shared/has-visible-text.js
@@ -13,7 +13,7 @@ describe('has-visible-text', function() {
   });
 
   it('should return false if there is no visible text', function() {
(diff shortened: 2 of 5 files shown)


lib/standards/html-elms.js
```diff
@@ -674,7 +674,7 @@ const htmlElms = {
     noAriaAttrs: true
   },
   picture: {
-    contentTypes: ['embedded', 'phrasing', 'flow'],
+    contentTypes: ['phrasing', 'flow'],
```
WilcoFiers, line 677: This was a mistake. `figure` is not am embedded element. The `img` inside it is. You can put text inside a picture along side the img and the text will be rendered. It also isn't described as embedded in the HTML spec:
https://html.spec.whatwg.org/multipage/grouping-content.html#the-figure-element

test/checks/shared/has-visible-text.js
```diff
@@ -13,7 +13,7 @@ describe('has-visible-text', function() {
   });
 
   it('should return false if there is no visible text', function() {
-    var params = checkSetup('<object id="target"></object>');
+    var params = checkSetup('<p id="target"></p>');
```
WilcoFiers, line 16: Don't know why we did that. Clearly that's wrong.

lib/standards/html-elms.js
WilcoFiers, line 677:
```suggestion
    // In HTML, this is considered "embedded". For axe-core we don't want this since its children will not be hidden
    contentTypes: ['phrasing', 'flow'],
```

lib/standards/html-elms.js
```diff
@@ -674,7 +674,8 @@ const htmlElms = {
     noAriaAttrs: true
   },
   picture: {
-    contentTypes: ['embedded', 'phrasing', 'flow'],
+    // In HTML, this is considered "embedded". For axe-core we don't want this since its children will not be hidden
```
straker, line 677: This matches the comment style from the rest of the doc and makes it a bit more clear why axe-core does not want children to be hidden

```suggestion
    // Note: spec change (do not count as embedded), because browsers do not hide text inside the picture element
```

straker (changes requested):

dylanb (approved):

Issue #2971, opened 2021-06-02
Heading-order may not correctly calculate level of heading element with aria-level


straker:
ARIA allows `<h1-6>` elements to have the `aria-level` attributes. This causes a potential bug in our [heading-order check](https://github.com/dequelabs/axe-core/blob/develop/lib/checks/navigation/heading-order-evaluate.js#L6-L18) as we only look at `aria-level` if the element has a `role=heading` attribute. Meaning that `<h1 aria-level="6">Goodbye</h1>` is calculated as a level 1 heading instead of a level 6 heading.

However, this doesn't work in all browsers. In modern browsers the `aria-level` is used to determine the heading level, but in IE 11 / JAWS the `aria-level` is ignored:

* FireFox / NVDA - "Heading level 6, Goodbye"
* Chrome / JAWS - "Heading level 6, Goodbye"
* IE 11 / JAWS - "Heading level 1, Goodbye"
* Safari / VO - "Heading level 6, Goodbye"
* Android Chrome / Voice Assistant - "Goodbye, Heading 6"

straker:
Since everyone else is using the `aria-level` to calculate the level, we should update our code to do that.

padmavemulapati:
Validated with the latest develop branch code base,
1.When we pass `aria-level="6" even it tagged in `h1` , and no `role` attribute is passing . It is announcing as heading 6.

![image](https://user-images.githubusercontent.com/47409579/128517080-06f5fc02-fa6f-41d3-bc19-e213236e6467.png)
2. we only look at aria-level if the element has a role=heading attribute. Meaning that `<h2 aria-level="5" and role=heading>hello</h2> ` is calculated as a level 5 heading instead of a level 2 heading.
![image](https://user-images.githubusercontent.com/47409579/128517372-508d1c81-769d-4b55-9c54-c57a6f99a82e.png)

straker:
No docs

PR #3028, opened 2021-06-22
fix(heading-order): use aria-level on headings in addition to role=header elements

clottman:
Allow aria-level on hn elements to override the semantic level of the element. 

Closes issue: #2971

### DIFF
--- a/lib/checks/navigation/heading-order-evaluate.js
+++ b/lib/checks/navigation/heading-order-evaluate.js
@@ -4,17 +4,18 @@ import { isVisible } from '../../commons/dom';
 
 function getLevel(vNode) {
   const role = vNode.attr('role');
-  if (role && role.includes('heading')) {
-    const ariaHeadingLevel = vNode.attr('aria-level');
-    const level = parseInt(ariaHeadingLevel, 10);
+  const headingRole = role && role.includes('heading');
+  const ariaHeadingLevel = vNode.attr('aria-level');
+  const level = parseInt(ariaHeadingLevel, 10);
 
-    // default aria-level for a heading is 2 if it is
-    // not set or set to an incorrect value
-    // @see https://www.w3.org/TR/wai-aria-1.1/#heading
-    if (isNaN(level) || level < 1 || level > 6) {
-      return 2;
-    }
+  // default aria-level for a heading is 2 if it is
+  // not set or set to an incorrect value
+  // @see https://www.w3.org/TR/wai-aria-1.1/#heading
+  if ((headingRole && isNaN(level)) || level < 1 || level > 6) {
+    return 2;
+  }
 
+  if (level) {
     return level;
   }
 
--- a/lib/checks/navigation/heading-order-evaluate.js
+++ b/lib/checks/navigation/heading-order-evaluate.js
@@ -1,27 +1,33 @@
 import cache from '../../core/base/cache';
 import { querySelectorAllFilter, getAncestry } from '../../core/utils';
 import { isVisible } from '../../commons/dom';
+import { getRole } from '../../commons/aria';
 
 function getLevel(vNode) {
-  const role = vNode.attr('role');
+  const role = getRole(vNode);
   const headingRole = role && role.includes('heading');
   const ariaHeadingLevel = vNode.attr('aria-level');
-  const level = parseInt(ariaHeadingLevel, 10);
+  const ariaLevel = parseInt(ariaHeadingLevel, 10);
+
+  const [, headingLevel] = vNode.props.nodeName.match(/h(\d)/) || [];
+
+  if (!headingRole) {
+    return -1;
+  }
+
+  if (headingLevel && !ariaHeadingLevel) {
+    return parseInt(headingLevel, 10);
+  }
 
   // default aria-level for a heading is 2 if it is
   // not set or set to an incorrect value
   // @see https://www.w3.org/TR/wai-aria-1.1/#heading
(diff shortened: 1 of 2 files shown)


lib/checks/navigation/heading-order-evaluate.js
```diff
@@ -4,17 +4,18 @@ import { isVisible } from '../../commons/dom';
...
-      return 2;
-    }
+  const headingRole = role && role.includes('heading');
+  const ariaHeadingLevel = vNode.attr('aria-level');
+  const level = parseInt(ariaHeadingLevel, 10);
+
+  // default aria-level for a heading is 2 if it is
+  // not set or set to an incorrect value
+  // @see https://www.w3.org/TR/wai-aria-1.1/#heading
+  if ((headingRole && isNaN(level)) || level < 1 || level > 6) {
```
WilcoFiers, line 14: This is going to fail on `<h1 role="heading">`. Not sure if that's a new problem, but it's a problem.

WilcoFiers (dismissed):

lib/checks/navigation/heading-order-evaluate.js
```diff
@@ -4,17 +4,18 @@ import { isVisible } from '../../commons/dom';
...
+    return 2;
+  }
 
+  if (level) {
```
straker, line 18: This will return the `aria-level` for non-header elements like `<iframe aria-level="2"></iframe>`, when we actually need the iframe to be level=-1

straker (dismissed):

lib/checks/navigation/heading-order-evaluate.js
clottman: The rule doesn't run on iframes, as far as I can tell.

straker: The rule doesn't, but we use a selector inside the evaluate to look for iframes and assign them a level of -1 so in the after we can reconstruct heading order correctly for all iframes on the page. https://github.com/dequelabs/axe-core/blob/0d6604b74118c11546e376987323863ef1751e78/lib/checks/navigation/heading-order-evaluate.js#L40-L46

lib/checks/navigation/heading-order-evaluate.js
```diff
@@ -49,7 +60,7 @@ function headingOrderEvaluate() {
       level: getLevel(vNode)
     };
   });
-
+  console.log(headingOrder);
```
WilcoFiers, line 63:
```suggestion
```

WilcoFiers (dismissed):

lib/checks/navigation/heading-order-evaluate.js
```diff
@@ -1,26 +1,37 @@
...
+    return parseInt(headingLevel, 10);
   }
 
-  const headingLevel = vNode.props.nodeName.match(/h(\d)/);
-  if (headingLevel) {
-    return parseInt(headingLevel[1], 10);
+  // default aria-level for a heading is 2 if it is
+  // not set or set to an incorrect value
+  // @see https://www.w3.org/TR/wai-aria-1.1/#heading
+  // note that NVDA and VO allow any positive level
```
straker, line 25:
```suggestion
  /*
   * default aria-level for a role=heading is 2 if it is
   * not set or set to an incorrect value. 
   * default aria-level for a heading element is the 
   * heading level.
   * note that NVDA and VO allow any positive level
   * @see https://www.w3.org/TR/wai-aria-1.1/#heading
   * @see https://codepen.io/straker/pen/jOBjNNe
   */
```

test/checks/navigation/heading-order.js
```diff
@@ -79,6 +79,91 @@ describe('heading-order', function() {
...
+      ]
+    });
+  });
+
+  it('should ignore aria-level on iframe when not used with role=heading', function() {
+    var vNode = queryFixture('<iframe aria-level="2"></iframe>');
+    axe.testUtils
+      .getCheckEvaluate('heading-order')
+      .call(checkContext, null, {}, vNode, { initiator: true });
+    //assert.equal(checkContext._data.headingOrder[0].ancestry[0], 'html > body > div:nth-child(1) > iframe:nth-child(1)');
```
straker, line 110:
```suggestion
```

straker (changes requested):
I believe this is the last thing. Everything else looks good. Good work.

straker (approved):

lib/checks/navigation/heading-order-evaluate.js
```diff
@@ -1,26 +1,42 @@
...
+  /*
+   * default aria-level for a role=heading is 2 if it is
+   * not set or set to an incorrect value. 
+   * default aria-level for a heading element is the 
+   * heading level.
+   * note that NVDA and VO allow any positive level
+   * @see https://www.w3.org/TR/wai-aria-1.1/#heading
+   * @see https://codepen.io/straker/pen/jOBjNNe
+   */
+  if (isNaN(ariaLevel) || ariaLevel < 1 || ariaLevel > 6) {
```
straker, line 31: Alright. Wilco and I talked and we came to the conclusion that we want to change how we're handling positive values. Since `aria-level` > 6 is handled different by different screen readers, we're going to flag it as part of a new check for aria-valid-attr-value and call out its lack of support.

But for heading-order we're just going to accept what the user gave us so there isn't two different failures caused by the same problem.

```suggestion
  if (isNaN(ariaLevel) || ariaLevel < 1) {
```

straker (dismissed):

straker (approved):

Issue #2505, opened 2020-09-02
Permit ARIA 1.2 combobox (without textbox child)


stevenyxu:
Expectation: A compliant ARIA 1.2 (draft) combobox should not raise axe violations. 

Actual: ARIA 1.2 combobox raises aria-required-children violation. Demo: https://w3c.github.io/aria-practices/examples/combobox/combobox-select-only.html. Currently raises 1 violation, aria-required-children.

Motivation: ARIA 1.2 is redefining combobox ([draft spec](https://www.w3.org/TR/wai-aria-1.2/#combobox)) to no longer require a textbox child because of [various issues with the 1.1 combobox](https://github.com/w3c/aria/wiki/Resolving-ARIA-1.1-Combobox-Issues). Aside from resolving long-standing labelling and SR support issues, this is particularly useful for web applications that want to emulate a native select or provide a rich selection widget (e.g. a form select with a popup that includes multiselection, options filter, select all/none). The current common practice is to use a popup button as the trigger rather than a combobox, but this is problematic in a form setting because button has problems with `aria-required` (prohibited by spec, flagged by axe, though supported in practice) and `aria-invalid` (permitted by spec, but not supported in practice). I've tested the new pattern with combobox trigger on JAWS, NVDA, ChromeVox on ChromeOS, VoiceOver, iOS VoiceOver, and Android TalkBack, and they all correctly announce current value, requiredness, and invalidness. In light of the apparent widespread support and that it's in a draft spec, it would be nice to permit it in axe. A couple of projects I'm involved with including [Angular Material](https://github.com/angular/components/pull/20082) have already cut over to the new pattern because of the clear a11y wins, and we're currently suppressing (in tooling) and manually ignoring (when using axe extension) the axe violation.

<pre><code>
axe-core version: 3.5.5
axe-webdriver, extension or other integration version: 4.5.3

Browser and Assistive Technology versions

For Tooling issues:
- Platform:  Tested on Chrome 84 OS X, reproducible running axe-core as a standalone library on node v10.
</code></pre>

straker:
Thanks for the issue. We've been discussing this very pattern over in #2478. The TL;DR is that we want to support the pattern, but we need to figure out what to do with the old 1.1 pattern first.

stevenyxu:
[...]

PR #3033, opened 2021-06-23
fix(combobox): support aria 1.2 pattern for comboboxes

clottman:
Closes issue: #2505


lib/standards/aria-roles.js
```diff
@@ -117,8 +117,8 @@ const ariaRoles = {
   },
   combobox: {
     type: 'composite',
-    requiredOwned: ['listbox', 'tree', 'grid', 'dialog', 'textbox'],
-    requiredAttrs: ['aria-expanded'],
+    requiredOwned: ['listbox', 'tree', 'grid', 'dialog'],
```
straker, line 120: From https://github.com/dequelabs/axe-core/issues/2505#issuecomment-788703942

> It seems like if axe removed the aria-required-children for combobox entirely

```suggestion
```

lib/standards/aria-roles.js
```diff
@@ -117,8 +117,8 @@ const ariaRoles = {
...
+    requiredAttrs: ['aria-expanded', 'aria-controls', 'aria-owns'],
```
straker, line 121: This implies that both are required when in reality it's just one or the other. That's not something we can support, so maybe we just list the 1.2 spec required attrs and then special case `aria-owns` to support the 1.0/1 pattern in aria-required-attrs-evaluate.

Also add `aria-owns` to the `allowedAttrs` array and remove `aria-controls` from it (since it's now in the required array).

```suggestion
    requiredAttrs: ['aria-expanded', 'aria-controls'],
```

lib/checks/aria/aria-required-children-evaluate.js
```diff
@@ -42,17 +42,6 @@ function missingRequiredChildren(virtualNode, role, required, ownedRoles) {
 
   // combobox exceptions
   if (isCombobox) {
```
straker, line 44: All of this special case should probably be removed. We don't want to check for required owned elements.

lib/checks/aria/aria-required-attr-evaluate.js
```diff
@@ -57,7 +56,26 @@ function ariaRequiredAttrEvaluate(node, options = {}, virtualNode) {
   }
 
   if (missing.length) {
-    this.data(missing);
+    const isCombobox = role === 'combobox';
+    const messageKey = null;
+    if (isCombobox) {
```
straker, line 61: We probably don't need any of the new messages and could just leave it to saying we require `aria-controls` by default. If the user uses `aria-owns` instead we pass the check.

lib/checks/aria/aria-required-attr-evaluate.js
```diff
@@ -56,11 +55,19 @@ function ariaRequiredAttrEvaluate(node, options = {}, virtualNode) {
     }
   }
 
+  if (
```
straker, line 58: We should add a comment that gives a reason for this special case and point to the comment from Sarah in the original issue.

test/integration/rules/aria-required-attr/required-attr.html
```diff
@@ -28,7 +28,14 @@
   ok
 </div>
 <div role="heading" id="pass4" aria-level="1">ok</div>
-<div role="combobox" id="pass5" aria-expanded="true">ok</div>
+<div
```
straker, line 31: Should add another comobox that uses the `aria-owned` pattern as well

straker (changes requested):

straker (approved):

Issue #2671, opened 2020-12-01
Give SerialVirtualNode a smart nodeType default


WilcoFiers:
This is just a small quality-of-life improvement. Instead of having SerialVirtualNode default to nodeType 1, it would be good if it could determine the nodeType based on what it is passed:

```js
new SerialVirtualNode({ nodeName: '#cdata-section' }).props.nodeType // 2
new SerialVirtualNode({ nodeName: '#text' }).props.nodeType // 3
new SerialVirtualNode({ nodeName: '#comment' }).props.nodeType // 8
new SerialVirtualNode({ nodeName: '#document' }).props.nodeType // 9
new SerialVirtualNode({ nodeName: '#document-fragment' }).props.nodeType // 11
```

The other way around would also be nice:

```js
new SerialVirtualNode({ nodeType: 2 }).props.nodeName // '#cdata-section'
new SerialVirtualNode({ nodeType: 3 }).props.nodeName // '#text'
new SerialVirtualNode({ nodeType: 8 }).props.nodeName // '#comment'
new SerialVirtualNode({ nodeType: 9 }).props.nodeName // '#document'
new SerialVirtualNode({ nodeType: 11 }).props.nodeName // '#document-fragment'
```

These are the only 5 cases for which this can work.

hchiam:
I'd like to try setting this up. I might simply have `normaliseProps` in serial-virtual-node.js call a couple helper functions

padmavemulapati:
devTask

PR #3036, opened 2021-06-24
chore(serial-virtual-node): add node type and node value defaults

hchiam:
From nodeName to nodeType. 
From nodeType to nodeName. 

Added NodeNamesToTypesEnum inside SerialVirtualNode. 

Closes issue: #2671


doc/rule-descriptions.md
```diff
@@ -24,7 +24,7 @@
...
-| [aria-roles](https://dequeuniversity.com/rules/axe/4.2/aria-roles?application=RuleDescription)                                   | Ensures all elements with a role attribute use a valid value                                                  | Serious, Critical | cat.aria, wcag2a, wcag412                                                                  | failure                    |                                                                                                        |
+| [aria-roles](https://dequeuniversity.com/rules/axe/4.2/aria-roles?application=RuleDescription)                                   | Ensures all elements with a role attribute use a valid value                                                  | Serious, Critical | cat.aria, wcag2a, wcag412                                                                  | failure, needs&nbsp;review |                                                                                                        |
```
hchiam, line 27: not sure if needed but this was an auto-generated change

lib/core/base/virtual-node/serial-virtual-node.js
```diff
@@ -44,14 +44,48 @@ class SerialVirtualNode extends AbstractVirtualNode {
...
+/**
+ * Map nodeName to nodeType
+ */
+const NodeNamesToTypesEnum = Object.freeze({
+  '#cdata-section': 2,
+  '#text': 3,
+  '#comment': 8,
+  '#document': 9,
+  '#document-fragment': 1
+});
```
hchiam, line 56: Is the comment necessary/sufficient? I'm also open to renaming, moving, or removing this.

hchiam (commented):
ready for initial review

lib/core/base/virtual-node/serial-virtual-node.js
straker: No, the comment isn't needed. I would say the `Objet.freeze` isn't need as well and you could remove that part of it (we only freeze publicly available things we don't want modified). Also we lowercase our consts (and you can drop the `enum`)

```js
nodeNamesToTypes
```

lib/core/base/virtual-node/serial-virtual-node.js
```diff
@@ -44,14 +44,48 @@ class SerialVirtualNode extends AbstractVirtualNode {
...
+
 /**
  * Convert between serialised props and DOM-like properties
  * @param {Object} serialNode
  * @return {Object} normalProperties
  */
 function normaliseProps(serialNode) {
   let { nodeName } = serialNode;
-  const { nodeType = 1 } = serialNode;
+  if (serialNode.nodeType === 2) {
```
straker, line 65: This could be simplified to look at the object and return the value. If you reverse the values and keys of the object you can also do the same for nodeName.

```js

```js
const nodeTypeToName = {};
const keys = Object.keys(nodeNamesToTypes);
Object.values(nodeNamesToTypes).forEach((value, index) => {
  nodeTypeToName[value] = keys[index];
});

const nodeType = serialNode.nodeType ?? nodeNamesToTypes[serialNode.nodeName] ?? 1;
const nodeName = serialNode.nodeName ?? nodeTypeToName[serialNode.nodeType];
```

lib/core/base/virtual-node/serial-virtual-node.js
```diff
@@ -60,6 +94,12 @@ function normaliseProps(serialNode) {
     typeof nodeName === 'string',
     `nodeName has to be a string, got '${nodeName}'`
   );
+  assert(
+    NodeNamesToTypesEnum[nodeName] === nodeType ||
+      !(nodeName in NodeNamesToTypesEnum) ||
+      !(nodeType in [2, 3, 8, 9, 11]),
+    `nodeName and nodeType have to match`
+  );
```
straker, line 102: I don't think this assert is needed as we can let the previous asserts catch invalid types of nodeType and nodeName. We don't want to test too much into valid values.

```suggestion
```

test/core/base/virtual-node/serial-virtual-node.js
```diff
@@ -77,6 +77,21 @@ describe('SerialVirtualNode', function() {
...
+      var vNode3 = new SerialVirtualNode({ nodeName: '#text' });
+      assert.equal(vNode3.props.nodeType, 3);
+      var vNode8 = new SerialVirtualNode({ nodeName: '#comment' });
+      assert.equal(vNode8.props.nodeType, 8);
+      var vNode9 = new SerialVirtualNode({ nodeName: '#document' });
+      assert.equal(vNode9.props.nodeType, 9);
+      var vNode11 = new SerialVirtualNode({ nodeName: '#document-fragment' });
+      assert.equal(vNode11.props.nodeType, 11);
+    });
+
```
straker, line 94: Please also add a test for the other way around (default to correct nodeName for certain nodeTypes).

doc/rule-descriptions.md
straker: This happens sometimes when the file is out of date from develop. We can just include it as part of this commit it's fine.

straker (changes requested):
Thanks for taking this on! Have a few suggestions.

hchiam (commented):
tests passing again, ready for review 🤞

straker (approved):
Great work. Thanks for taking this on

WilcoFiers:
Reviewed for security

Issue #3037, opened 2021-06-24
color-contrast-matches isClipped function catches too much


straker:
On https://www.drupal.nl/, none of the anchor elements in the top-nav are flagged by color-contrast (even though they should all fail). Turns out the parent `nav` element is marked as hidden by our `is-visible` function in the color-contrast-matcher because it uses a CSS clip but doesn't do it to try to hide the element from screen readers.

```css
.js .nav-collapse {
    clip: rect(0 0 0 0);
    max-height: none;
    position: relative;
    display: block;
    overflow: hidden;
    zoom: 1;
}
```

clottman:
Clip only applies to absolutely positioned elements (`position: absolute` or `position: fixed`), and in this case the nav element is relatively positioned, so the clip attribute isn't actually doing anything. Note, clip-path can apply to any element regardless of positioning. 

So, I propose that our is-visible function, when checking isClipped, should also check positioning when an element `isClipped` because of the `clip` property.

padmavemulapati:
Clipped elements which are styled as `position:absolute` and `position:fixed` not visible
Validated with the latest develop-branch code base
![image](https://user-images.githubusercontent.com/47409579/128515846-849ab8e0-f280-422d-abfa-1d4efe687d6b.png)

straker:
No docs

PR #3038, opened 2021-06-24
fix(color-contrast-matches): only absolutely positioned elements using clip should be not visible

clottman:
Clip is only applied on absolutely positioned elements (`position: fixed` or `position: absolute`), so if an element is using `clip` but doesn't have that positioning, it will still be visible.

Closes issue: #3037

### DIFF
--- a/test/commons/dom/is-visible.js
+++ b/test/commons/dom/is-visible.js
@@ -158,6 +158,20 @@ describe('dom.isVisible', function() {
       assert.isFalse(axe.commons.dom.isVisible(el));
     });
 
+    it('should detect when clip is not applied because of positioning', function() {
+      var el,
+        clip =
+          'clip: rect(1px 1px 1px 1px);' +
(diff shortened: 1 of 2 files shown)


test/commons/dom/is-visible.js
```diff
@@ -174,6 +188,21 @@ describe('dom.isVisible', function() {
       assert.isFalse(axe.commons.dom.isVisible(el));
     });
 
+    it('should detect when clip is not applied because of positioning on parent', function() {
```
WilcoFiers, line 191: Why are you repeating tests?

WilcoFiers (changes requested):

test/commons/dom/is-visible.js
clottman: I'm not - this is when clip is not applied because of positioning on the parent, the previous test is positioning on the element itself. Down below, it's testing when the screenreader option is set using these same two cases.

WilcoFiers (approved):

PR #3047, opened 2021-06-28
chore(getFrameContexts): include all context info

WilcoFiers:
Turns out I need all context properties to be returned from getFrameContexts as part of #3006. Problem was that context.frames doesn't actually have all the necessary info. So I moved the logic that created that info from `collectResultsFromFrames` to a new `createFrameContext` method, which is used by Context.

That way it is available immediately from Context, which makes it much easier to return from getFrameContexts.


lib/core/utils/get-bounding-size.js
```diff
@@ -0,0 +1,11 @@
+export default function getBoundingSize(domNode) {
+  let width = parseInt(domNode.getAttribute('width'), 10);
+  let height = parseInt(domNode.getAttribute('height'), 10);
+
+  if (isNaN(width) || isNaN(height)) {
+    const rect = domNode.getBoundingClientRect();
+    width = isNaN(width) ? rect.width : width;
```
straker, line 7: This will accept any element with `width` or `height` attribute as the true size, even if the element itself doesn't apply them. Since this is a util function we'll need to account for non-frame elements.

```js
fixture.innerHTML = '<div width="100px"></div>';
var node = fixture.querySelector('div');
axe.utils.getBoundingSize(node); // { width: 100, height: 0 }
```

straker (changes requested):

lib/core/utils/get-bounding-size.js
WilcoFiers: I didn't put it on axe.utils. I don't think that matters.

straker: You did https://github.com/dequelabs/axe-core/pull/3047/files#diff-b5e51e5519d1527b32e31b86be8881c77bc4aed33ebcb09361127b17376ff6f7R30

straker (approved):

PR #3050, opened 2021-06-28
chore: allow aria-valid-attr-value to run virtually

straker:
Part of issue #2734


lib/checks/aria/aria-errormessage-evaluate.js
```diff
@@ -24,33 +24,38 @@ import { tokenList } from '../../core/utils';
...
     if (attr.trim() === '') {
       return standards.ariaAttrs['aria-errormessage'].allowEmpty;
     }
-    const idref = attr && doc.getElementById(attr);
+    let idref;
+
+    try {
+      idref = attr && idrefs(virtualNode, 'aria-errormessage')[0];
+    } catch (e) {
+      return undefined;
```
WilcoFiers, line 50: Think you need to set messageKey here.

lib/checks/aria/aria-errormessage.json
```diff
@@ -11,7 +11,8 @@
       },
       "incomplete": {
         "singular": "ensure aria-errormessage value `${data.values}` references an existing element",
-        "plural": "ensure aria-errormessage values `${data.values}` reference existing elements"
+        "plural": "ensure aria-errormessage values `${data.values}` reference existing elements",
+        "idrefs": "Unable to determine if aria-errormessage element exists on the page: ${data.values}"
```
WilcoFiers, line 15:
```suggestion
        "idrefs": "unable to determine if aria-errormessage element exists on the page: ${data.values}"
```

lib/commons/aria/validate-attr-value.js
```diff
@@ -44,11 +47,15 @@ function validateAttrValue(node, attr) {
       }, list.length !== 0);
 
     case 'idref':
-      return !!(value && doc.getElementById(value));
+      try {
+        const doc = getRootNode(vNode.actualNode);
+        return !!(value && doc.getElementById(value));
+      } catch (e) {
+        throw new TypeError('Cannot resolve id references for non-DOM nodes');
```
WilcoFiers, line 54: This can happen with unmounted nodes too. This seems like a more appropriate message:

```suggestion
        throw new TypeError('Cannot resolve id references for partial DOM');
```

test/checks/aria/errormessage.js
```diff
@@ -193,4 +190,22 @@ describe('aria-errormessage', function() {
...
+          'aria-invalid': 'true',
+          'aria-errormessage': 'test'
+        }
+      });
+      assert.isUndefined(
+        axe.testUtils
+          .getCheckEvaluate('aria-errormessage')
+          .call(checkContext, null, null, vNode)
+      );
+      assert.deepEqual(checkContext._data, ['test']);
```
WilcoFiers, line 208: You'll need to test for messsageKey here.

lib/checks/aria/aria-valid-attr-value-evaluate.js
```diff
@@ -89,17 +88,22 @@ function ariaValidAttrValueEvaluate(node, options) {
...
+      if (
+        !skipAttrs.includes(attrName) &&
+        options.indexOf(attrName) === -1 &&
+        aria.test(attrName) &&
+        (preChecks[attrName] ? preChecks[attrName]() : true) &&
+        !validateAttrValue(virtualNode, attrName)
+      ) {
+        invalid.push(`${attrName}="${attrValue}"`);
+      }
+    } catch (e) {
```
WilcoFiers, line 104: This block will catch a lot more than just idref issues.

WilcoFiers (changes requested):

lib/checks/aria/aria-valid-attr-value-evaluate.js
```diff
@@ -87,25 +85,36 @@ function ariaValidAttrValueEvaluate(node, options, virtualNode) {
     }
   };
 
-  for (let i = 0, l = attrs.length; i < l; i++) {
-    const attrName = attrs[i];
+  virtualNode.attrNames.forEach(attrName => {
+    // skip any attributes handled elsewhere (we don't want use
+    // Array.filter first as that would mean looping over the
+    // attributes twice. on a page with tens of thousands of
+    // nodes that's a lot of overhead cost in performance)
```
WilcoFiers, line 92: This really isn't necessary.

```suggestion
```

WilcoFiers (dismissed):

lib/checks/aria/aria-valid-attr-value-evaluate.js
straker: I felt it was since I was refactoring the code to use a filter then a foreach but realized it would cause a performance hit. Didn't want someone to make that same mistake in the future

WilcoFiers (approved):

Issue #2823, opened 2021-03-01
Role="main" and <main> interaction with aria-hidden="true" creates false positive


MatthewSbar:
Expectation: The test "Ensures the document has at most one main landmark" should respect `aria-hidden` when detecting `<main>` elements

Actual: Both `<main>` and `<div role="main">` are counted as main landmarks despite `<main>` being a child of an element with `aria-hidden="true"`

Motivation: This creates a double standard for testing, particularly with modals. If a page normally has a `<main>` element, and opening a modal adds `aria-hidden` to that element's parent, the page is (correctly) seen as having no `main` landmark, however, if a `role="main"` is added within the modal, the `aria-hidden` `<main>` element is counted towards the maximum of one. In other words, the test to see if there is any `main` landmark respects `aria-hidden`, but the test to see if thee is no more than one does not.

Here's an example page that demonstrates the issue:

```
<html lang="en">
  <body>
    <title>Fake page</title>
    <div aria-hidden="true">
      <h1>Hidden page heading</h1>
      <main>hidden main</main>
    </div>
    <div role="dialog" aria-label="fake dialog">
      <h1>Fake dialog heading</h1>
      <div role="main">Fake dialog main</div> <!-- This line is causing a failure -->
    </div>
  </body>
</html>
```
 
In the above example, removing `role-"main"` will mean the page has 0 `main` landmarks, including it as is means the page will have >1 `main` landmarks.
 
<pre><code>
axe-core version: 4.1.2
axe-webdriver, extension or other integration version: 4.7.2

Browser and Assistive Technology versions
</code></pre>

straker:
Good catch.

Dragoneric:
Good finding :-) 
Example: Cookie-Disclaimer on Websites are often used in the shown HTML-syntax

WilcoFiers:
Agreed, this needs to be changed. Thanks for reporting. In addition, when fixing this I think we need to make sure we continue to pass the following:

```html
<html lang="en">
  <body>
    <title>Fake page</title>
    <div aria-hidden="true">
      <h1>Hidden page heading</h1>
      <main>hidden main</main>
    </div>
    <div role="dialog" aria-label="fake dialog">
      <h1>Fake dialog heading</h1>
      <div>Fake dialog main</div> <!-- This line is causing a failure -->
    </div>
  </body>
</html>
```

Dragoneric:
[...]

PR #3051, opened 2021-06-28
fix(page-no-duplicate): don't count elements hidden from screenreaders as potential duplicates

clottman:
Closes issue: #2823


lib/checks/generic/page-no-duplicate-evaluate.js
```diff
@@ -18,7 +18,7 @@ function pageNoDuplicateEvaluate(node, options, virtualNode) {
   cache.set(key, true);
 
   let elms = querySelectorAllFilter(axe._tree[0], options.selector, elm =>
-    isVisible(elm.actualNode)
+    isVisible(elm.actualNode, true)
```
WilcoFiers, line 21: I think this is now going to fail the scenario where there is a `main` with aria-hidden="true". I don't think that should fail.

WilcoFiers (changes requested):

WilcoFiers (approved):

PR #3054, opened 2021-06-29
fix: Remove unnecessary files from npm package

stephenmathieson:
We just had to manually publish `@next`, as [a `develop` build failed](https://app.circleci.com/pipelines/github/dequelabs/axe-core/2462/workflows/19d5c98e-c7e4-4a85-87f1-3503179ac326/jobs/33275). When doing this, I noticed how many things are included in the tarball we send to npm. This patch drastically reduces the size of the tarball and the number of files it contains.

Previously:

```
npm notice === Tarball Details ===
npm notice name:          axe-core
npm notice package size:  811.9 kB
npm notice unpacked size: 3.5 MB
npm notice total files:   689
```

Now:

```
npm notice === Tarball Details ===
npm notice name:          axe-core
npm notice package size:  314.3 kB
npm notice unpacked size: 1.5 MB
npm notice total files:   7
```

This will hopefully reduce the chance of another failed upload.


straker (approved):
I believe these are all the files we need, but @WilcoFiers should also approve so we don't mess it up.

package.json
```diff
@@ -50,6 +50,12 @@
   ],
   "main": "axe.js",
   "typings": "axe.d.ts",
+  "files": [
+    "axe.js",
+    "axe.min.js",
+    "axe.d.ts",
+    "locales/"
```
WilcoFiers, line 57:
```suggestion
    "LICENSE",
    "locales/"
```

WilcoFiers (dismissed):
Not sure why this is better over updating `npmignore`. At the very least let's not use both.

package.json
stephenmathieson: npm includes this itself.

stephenmathieson:
> Not sure why this is better over updating npmignore. At the very least let's not use both.

Good catch. I'll remove the `.npmignore`. IMO being explicit about what's being published is much better than implicitly including everything other than what's listed in that file.

package.json
WilcoFiers: Does it? OK sure cool.

WilcoFiers (approved):

Issue #2961, opened 2021-05-25
aria-hidden-focus does not account for inputs being inside a disabled <fieldset>


thomasbickley-bjss:
Product: axe Extension

Expectation: An input that is inside a `<fieldset>` that has been disabled should not be treated as focusable by the aria-hidden-focus rule

Actual: An input that is inside a `<fieldset>` that has been disabled is still treated as focusable by the aria-hidden-focus rule

Motivation: The aria-hidden-focus rule returns a false positive when aria-hidden is used with a disabled fieldset

Example:
The aria-hidden-focus rule flags the following as a serious issue, despite the input not actually being focusable. 
```
<fieldset disabled="disabled" aria-hidden="true">
	<input />
</fieldset>
```

<pre><code>axe-core version: 4.2.0
axe extension version: 4.9.3
</code></pre>

WilcoFiers:
Thank you for reporting the issue. We'll be sure to fix it. I think the problem is in dom.focusDisabled, which do
https://github.com/dequelabs/axe-core/blob/develop/lib/commons/dom/focus-disabled.js#L10

To fix this we'll need to have i check walk up the tree to see if there's a disabled fieldset. Should be fairly straightforward. The one thing to be mindful of in fixing this is to check how this interacts with shadow DOM. I recall that `disable` only works on descendants in the same DOM tree. We should test it to make sure.

padmavemulapati:
Validated with the latest develop branch code base,
observed that, 
1. When given content made unfocusable through disabled fieldset`<fieldset id="target" disabled aria-hidden="true"><input /></fieldset>` test will pass
2. Fail when content is in the legend of a disabled fieldset
` <fieldset id="target" disabled aria-hidden="true"><legend><input /></legend></fieldset>`
3.also fail when content is in an aria-hidden but not disabled fieldset
`<fieldset id="target" aria-hidden="true"><input /></fieldset>`
![image](https://user-images.githubusercontent.com/47409579/128302122-ec206098-b767-4d82-b64e-aa77d21ffc79.png)

straker:
No docs

thomasbickley-bjss:
@straker I don't understand what you mean. What docs are missing here?

straker:
Sorry, that was a paperwork style comment. Every issue needs testing and then might need docs updated. Was just saying that this issue doesn't need to update or add any docs.

thomasbickley-bjss:
Was closing it accidental then?

[...]

PR #3056, opened 2021-06-30
fix(aria-hidden-focusable): disabled aria-hidden fieldset should not have focusable children

clottman:
Closes issue: #2961

### DIFF
--- a/lib/commons/dom/focus-disabled.js
+++ b/lib/commons/dom/focus-disabled.js
@@ -11,6 +11,7 @@ function focusDisabled(el) {
   const vNode = el instanceof AbstractVirtualNode ? el : getNodeFromTree(el);
 
   if (vNode.hasAttr('disabled')) {
+    vNode._isDisabled = true;
     return true;
   }
 
@@ -21,9 +22,10 @@ function focusDisabled(el) {
     if (parentNode.props.nodeName === 'legend') {
       break;
     }
+
     if (
       parentNode.props.nodeName === 'fieldset' &&
-      parentNode.hasAttr('disabled') &&
+      (parentNode._isDisabled || parentNode.hasAttr('disabled')) &&
       parentNode.shadowId === vNode.shadowId
     ) {
       return true;
@@ -35,10 +37,12 @@ function focusDisabled(el) {
     // if the virtual node does not have an actual node, treat it
     // as not hidden
     if (!vNode.actualNode) {
+      vNode._isDisabled = false;
       return false;
     }
-
-    return isHiddenWithCSS(vNode.actualNode);
+    const isDisabled = isHiddenWithCSS(vNode.actualNode);
+    vNode._isDisabled = isDisabled;
+    return isDisabled;
   }
 
   return false;
--- a/lib/commons/dom/focus-disabled.js
+++ b/lib/commons/dom/focus-disabled.js
@@ -38,7 +38,7 @@ function focusDisabled(el) {
       parentNode.hasAttr('disabled') &&
       parentNode.shadowId === vNode.shadowId
     ) {
-      vNode._inDisabledFieldset = true;
+      fieldsetDisabled = true;
     }
     parentNode = parentNode.parent;
   }
--- a/lib/commons/dom/focus-disabled.js
+++ b/lib/commons/dom/focus-disabled.js
@@ -24,7 +24,8 @@ function focusDisabled(el) {
     }
     if (
       parentNode.props.nodeName === 'fieldset' &&
-      parentNode.hasAttr('disabled')
+      parentNode.hasAttr('disabled') &&
+      parentNode.shadowId === vNode.shadowId
     ) {
       return true;
     }
--- a/lib/commons/dom/focus-disabled.js
+++ b/lib/commons/dom/focus-disabled.js
@@ -19,24 +19,25 @@ function focusDisabled(el) {
   let parentNode = vNode.parent;
   const ancestors = [];
   let fieldsetDisabled = false;
-  while (parentNode && !fieldsetDisabled) {
+  while (
+    parentNode &&
+    parentNode.shadowId === vNode.shadowId &&
+    !fieldsetDisabled
+  ) {
     ancestors.push(parentNode);
     if (parentNode.props.nodeName === 'legend') {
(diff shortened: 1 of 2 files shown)


lib/commons/dom/focus-disabled.js
```diff
@@ -14,6 +14,23 @@ function focusDisabled(el) {
...
+    // https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset
+    if (parentNode.props.nodeName === 'legend') {
+      parentNode = null;
+      continue;
+    }
+    if (
+      parentNode.props.nodeName === 'fieldset' &&
+      parentNode.hasAttr('disabled')
+    ) {
+      return true;
```
straker, line 29: This will return true even if the fieldset is part of a different shadow tree. However form fields inside a shadow root do not apply `disabled` from a parent fieldset.

```html
<fieldset disabled>
  <legend><label>This is strange<input name="hello"></label></legend>

  <label>this should be disabled <input></label>

  <div id="shadow"></div>
</fieldset>

<script src="/axe.js"></script>
<script>
  const shadow = document.querySelector('#shadow');
  const shadowRoot = shadow.attachShadow({mode: 'open'});
  shadowRoot.innerHTML = '<label>Shadow input <input></label>';
</script>
```

VirtualNodes have a `shadowId` property on them, so if the two are equal then we can return `true`, otherwise it won't apply.

straker (dismissed):

lib/commons/dom/focus-disabled.js
```diff
@@ -14,6 +14,24 @@ function focusDisabled(el) {
     return true;
   }
 
+  let parentNode = vNode.parent;
+  while (parentNode) {
+    // if a form element is in a legend, that element will not be disabled even if the fieldset is
+    // https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset
+    if (parentNode.props.nodeName === 'legend') {
+      parentNode = null;
+      continue;
```
straker, line 23: This could just be a `break` rather than setting `parentNode` to null to quit the loop in the next iteration and then continuing.

```suggestion
      break;
```

straker (changes requested):

test/checks/keyboard/focusable-disabled.js
```diff
@@ -41,6 +41,50 @@ describe('focusable-disabled', function() {
     assert.isTrue(actual);
   });
 
+  it('returns true when content made unfocusable through disabled fieldset', function() {
+    var params = checkSetup(
+      '<fieldset disabled aria-hidden="true"><input id="target" /></fieldset>'
```
straker, line 46: These tests are technically running against the wrong node. The evaluate function should be run against the element that has the `aria-hidden` on it. Each of these tests (except for the shadow root one) runs against an input element. This can potentially cause problems later on if something about how we get tabbableElements stops returning the node itself.

lib/commons/dom/focus-disabled.js
```diff
@@ -14,6 +14,23 @@ function focusDisabled(el) {
     return true;
   }
 
+  let parentNode = vNode.parent;
+  while (parentNode) {
```
straker, line 18: Looking at this more closely this will cause a pretty big performance problem. Since each child element of the `aria-hidden` element will use this function, each of them will walk up the entire parent tree of the page. If the aria-hidden is placed near the top of the tree, that is a lot of checking of nodes already checked.

We should cache on the VirtualNode information about this lookup so each element only needs to go up a few nodes until it finds an element that has already gone through this lookup chain.

We'll want to do something similar to [is-hidden](https://github.com/dequelabs/axe-core/blob/develop/lib/core/utils/is-hidden.js) and cache the results of the recursive lookup onto each node we look at so we only ever have to look at a parent tree once.

straker (changes requested):

straker (changes requested):
The cache on `focusDisabled` isn't quite working. The cache won't break early from looking at the entire parent tree looking for a fieldset. What you'll need to do is look at at `_isDisabled` for every parent element and if it isn't undefined return the value. You'll also then want to cache that value on the node itself so a child of the node will only have to go up 1 parent to find the result.

lib/commons/dom/focus-disabled.js
```diff
@@ -14,13 +14,49 @@ function focusDisabled(el) {
     return true;
   }
 
+  // if a form element is in a legend, that element will not be disabled even if the fieldset is
+  // https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset
+  let parentNode = vNode.parent;
+  const ancestors = [];
+  let fieldsetDisabled = false;
+  while (parentNode && !fieldsetDisabled) {
```
straker, line 22: If we know we're not in the same shadow tree anymore we can stop looking at the parent tree completely as well. This also removes the need to check for this in the if statements as well.

```suggestion
  while (parentNode && parentNode.shadowId === vNode.shadowId && !fieldsetDisabled) {
```

straker: Also, we'll want to break once we know whether or not there is a parent fieldset disabled. Right now this only breaks if the value is true, but we want to short circuit the entire tree as soon as any node has the cached value regardless of what the value is.

straker (changes requested):

straker (changes requested):

lib/commons/dom/focus-disabled.js
```diff
@@ -19,24 +19,25 @@ function focusDisabled(el) {
   let parentNode = vNode.parent;
   const ancestors = [];
   let fieldsetDisabled = false;
-  while (parentNode && !fieldsetDisabled) {
+  while (
```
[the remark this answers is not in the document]
clottman: I'm doing that with the `break` [on line 35](https://github.com/dequelabs/axe-core/pull/3056/files#diff-9926a4ccb8ed5d395649db684d3ad79eda8b54183ce202cae649e59dd999b797R35)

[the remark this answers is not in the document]
straker: Gotcha. Sorry, didn't see that break.

straker (approved):

Issue #3029, opened 2021-06-22
Cannot select shadow dom elements for `include` option of axe.run


straker:
Axe-core currently is unable to `include` elements that reside inside a shadow DOM. Trying to do so results in an error.

```html
<div id="shadow"></div>'

<script>
(async () => {
  var shadow = document.querySelector('#shadow');
  var shadowRoot = shadow.attachShadow({ mode: 'open' });
  shadowRoot.innerHTML = '<main><iframe id="f1"></iframe></main>';

  await axe.run({ include: ['main'] });  // Error: No elements found for include in page Context
})();
</script>
```

WilcoFiers:
I'm going to close this issue. We have never seen a use case for this. If anyone does encounter issues with this, feel free to open a new issue.

PR #3063, opened 2021-07-06
feat(axe.run) include shadow elements

clottman:
This seems like it should be a straightforward change from document.querySelectorAll to the axe querySelectorAll but it breaks a bunch of stuff. 

Closes issue: #3029

### DIFF
--- a/lib/core/base/context.js
+++ b/lib/core/base/context.js
@@ -46,7 +46,9 @@ function pushUniqueFrameSelector(context, type, selectorArray) {
   context.frames = context.frames || [];
 
   const frameSelector = selectorArray.shift();
-  const frames = querySelectorAll(context.flatTree, frameSelector);
+  const frames = querySelectorAll(context.flatTree, frameSelector).map(
+    vNode => vNode.actualNode
+  );
 
   Array.from(frames).forEach(frame => {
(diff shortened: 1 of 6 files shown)


lib/core/base/context.js
```diff
@@ -45,7 +46,9 @@ function pushUniqueFrameSelector(context, type, selectorArray) {
   context.frames = context.frames || [];
 
   const frameSelector = selectorArray.shift();
-  const frames = document.querySelectorAll(frameSelector);
+  const frames = querySelectorAll(context.flatTree, frameSelector).map(
```
WilcoFiers, line 49: This introduced quite a few problems:

1. Existing uses of include/exclude are going to find more nodes than it does today, because they are now applied over all trees.
2. Real CSS selectors will find the wrong element. `some-component > :first-child` will get the first child in the flat tree, instead of the first slotted element.
3. Selectors currently supported won't be. utils.querySelectorAll only supports a subset of CSS
4. IDs selection becomes very quirky, since they're no longer used

This has the potential to break a fair amount of existing scripts running axe-core today. I'm not in favor of this change.

WilcoFiers (changes requested):

PR #3067, opened 2021-07-07
chore: allow aria-allowed-role to run virtually

straker:
Converting `is-visible` was a bit of a pain...

Part of issue #2734


lib/commons/aria/get-element-unallowed-roles.js
```diff
@@ -25,26 +31,18 @@ const dpubRoles = [
...
   }
 
-  if (node.hasAttribute('role')) {
-    const nodeRoles = tokenList(node.getAttribute('role').toLowerCase());
+  if (vNode.hasAttr('role')) {
+    const nodeRoles = tokenList(vNode.attr('role').toLowerCase());
     roles = roles.concat(nodeRoles);
   }
 
-  if (node.hasAttributeNS('http://www.idpf.org/2007/ops', 'type')) {
```
straker, line 40: We determined that we no longer need to support this so we're good to move out of it.

lib/commons/dom/is-visible.js
```diff
@@ -112,50 +113,74 @@ function isVisible(el, screenReader, recursed) {
     );
   }
 
-  const vNode = getNodeFromTree(el);
+  const vNode = el instanceof AbstractVirtualNode ? el : getNodeFromTree(el);
```
straker, line 116: The general strategy is check things that virtual node has access to, followed by things only a true DOM node has access to. A detatched virtual node is visible.

test/commons/dom/is-visible.js
```diff
@@ -5,25 +5,24 @@ describe('dom.isVisible', function() {
...
   describe('default usage', function() {
     // Firefox returns `null` if accessed inside a hidden iframe
     it('should return false if computedStyle return null for whatever reason', function() {
-      var orig = window.getComputedStyle;
-      window.getComputedStyle = function() {
-        return null;
-      };
-      assert.isFalse(axe.commons.dom.isVisible(fakeNode));
-      window.getComputedStyle = orig;
+      computedStyleStub = sinon.stub(window, 'getComputedStyle').returns(null);
```
straker, line 23: If this test failed it would not reset the override and so all future tests would fail.

lib/core/utils/is-html-element.js
```diff
@@ -1,137 +1,20 @@
...
 /**
  * Verifies that if a given html tag is valid
  * @method isHtmlElement
  * @memberof axe.utils
- * @param htmlTag htmlTag to check if valid
+ * @param {HTMLElement|VirtualNode} node node to check if valid
  * @return {Boolean} true/ false
  */
 function isHtmlElement(node) {
+  const nodeName = node.props?.nodeName ?? node.nodeName.toLowerCase();
```
WilcoFiers, line 11: <3

lib/core/utils/is-html-element.js
```diff
@@ -1,137 +1,20 @@
...
+
   if (node.namespaceURI === 'http://www.w3.org/2000/svg') {
     return false;
   }
 
-  return htmlTags.includes(node.nodeName.toLowerCase());
+  return Object.keys(standards.htmlElms).includes(nodeName);
```
WilcoFiers, line 17: Don't think you need to create an array here:

```suggestion
  return !!standards.htmlElms[nodeName];
```

The other thing about this, if someone for some reason wanted to unset an HTML element with `axe.configure({ standards })`, they they wouldn't be able to, since this still returns true, even if the prop on `htmlElms` is undefined or null.

lib/commons/aria/get-element-unallowed-roles.js
```diff
@@ -54,21 +52,22 @@ function getRoleSegments(node) {
 /**
  * gets all unallowed roles for a given node
  * @method getElementUnallowedRoles
- * @param {Object} node HTMLElement to validate
- * @param {String} tagName tag name of a node
+ * @param {Object|VirtualNode} node HTMLElement to validate
```
WilcoFiers, line 55:
```suggestion
 * @param {Element|VirtualNode} node HTMLElement to validate
```

test/checks/aria/aria-allowed-role.js
```diff
@@ -1,421 +1,333 @@
...
     var actual = axe.testUtils
       .getCheckEvaluate('aria-allowed-role')
-      .call(checkContext, target, options);
+      .call(checkContext, null, options, vNode);
     var expected = false;
     assert.equal(actual, expected);
     assert.deepEqual(checkContext._data, ['row']);
   });
 
-  it('returns true when A has namespace as svg', function() {
```
WilcoFiers, line 45: Where did this go?

test/checks/aria/aria-allowed-role.js
```diff
@@ -1,421 +1,333 @@
...
     );
     assert.deepEqual(checkContext._data, ['none']);
   });
 
   it('returns true when input of type image and no role', function() {
-    var node = document.createElement('img');
-    node.setAttribute('type', 'image');
-    fixture.appendChild(node);
-    flatTreeSetup(fixture);
+    var vNode = queryFixture('<img id="target" type="image"/>');
```
WilcoFiers, line 139:
```suggestion
    var vNode = queryFixture('<input id="target" type="image"/>');
```

test/commons/aria/get-element-unallowed-roles.js
```diff
@@ -66,16 +67,28 @@ describe('aria.getElementUnallowedRoles', function() {
...
     var node = document.createElement('tr');
     node.setAttribute('role', 'row');
     flatTreeSetup(node);
-    var actual = axe.commons.aria.getElementUnallowedRoles(node, false);
+    var actual = getElementUnallowedRoles(node, false);
     assert.isNotEmpty(actual);
     assert.include(actual, 'row');
   });
+
+  it('works with SerialVirtualNode', function() {
```
WilcoFiers, line 83: Can you add at least one additional test showing it doesn't always return empty?

test/rule-matches/aria-allowed-role-matches.js
```diff
@@ -1,47 +1,27 @@
...
+  var queryFixture = axe.testUtils.queryFixture;
   var rule;
 
   beforeEach(function() {
     rule = axe.utils.getRule('aria-allowed-role');
   });
 
-  afterEach(function() {
-    fixture.innerHTML = '';
-  });
```
WilcoFiers, line 14: Why did you remove this?

test/integration/virtual-rules/aria-allowed-role.js
```diff
@@ -0,0 +1,106 @@
...
+        role: 'link'
+      }
+    });
+
+    assert.lengthOf(results.passes, 1);
+    assert.lengthOf(results.violations, 0);
+    assert.lengthOf(results.incomplete, 0);
+  });
+
+  it('should incomplete for hidden element', function() {
```
WilcoFiers, line 70: Why is this incomplete?

WilcoFiers (changes requested):

WilcoFiers:
Also, change the PR title

test/checks/aria/aria-allowed-role.js
straker: https://github.com/dequelabs/axe-core/pull/3067#discussion_r665746595

test/rule-matches/aria-allowed-role-matches.js
straker: Our testutils file does this for us https://github.com/dequelabs/axe-core/blob/develop/test/testutils.js#L440-L442

test/integration/virtual-rules/aria-allowed-role.js
straker: Because hidden elements with unallowed roles return undefined https://github.com/dequelabs/axe-core/blob/develop/lib/checks/aria/aria-allowed-role-evaluate.js#L33-L35

WilcoFiers (approved):

PR #3068, opened 2021-07-08
fix(utils): unify selecting nodes in shadow tree with shadowSelect()

WilcoFiers:
Adds axe.utils.shadowSelect to select elements within a shadow tree. This method was created to be used along side `axe.utils.getFrameContexts()`.

### DIFF
--- a/lib/core/utils/shadow-select.js
+++ b/lib/core/utils/shadow-select.js
@@ -1,0 +1,26 @@
+/**
+ * Find the first element to match a selector.
+ * Use an array of selectors to reach into shadow DOM trees
+ * 
+ * @param {string|string[]} selector String or array of strings with a CSS selector
+ * @param {Document} doc Optional document node
+ * @returns {Element|Null}
+ */
+export default function shadowSelect(selectors) {
+  // Spread to avoid mutating the input
+  const selectorArr = Array.isArray(selectors) ? [...selectors] : [selectors];
+  return selectRecursive(selectorArr, document)
+}
+
+/* Find an element in shadow or light DOM trees, using an axe selector */
+function selectRecursive(selectors, doc) {
+  const selectorStr = selectors.shift();
+  const elm = selectorStr ? doc.querySelector(selectorStr) : null;
+  if (selectors.length === 0) {
+    return elm;
+  }
+  if (!elm?.shadowRoot) {
+    return null;
+  }
+  return selectRecursive(selectors, elm.shadowRoot);
+}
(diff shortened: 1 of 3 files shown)


test/core/utils/shadow-select.js
```diff
@@ -0,0 +1,92 @@
...
+    it('returns null if the node does not exist in the shadow tree', function () {
+      var shadowRoot = appendShadowTree(fixture, 'div')
+      shadowRoot.innerHTML = '<b class="hello"></b>';
+      assert.isNull(shadowSelect(['#fixture > div', '.goodbye']));
+    });
+
+    it('returns null if an intermediate node is not a shadow root', function () {
+      var shadowRoot = appendShadowTree(fixture, 'article');
+      shadowRoot.innerHTML = '<section></section>';
+      assert.isNull(shadowSelect(['#fixture > article', 'section', 'p']));
```
straker, line 57: Since the `p` tag doesn't exist in the fixture, even if the `section` was a shadow root this would return `null`. 

```suggestion
      shadowRoot.innerHTML = '<section><p class="hello"></p></section>';
      assert.isNull(shadowSelect(['#fixture > article', 'section', 'p']));
```

lib/core/utils/shadow-select.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * Find the first element to match a selector.
+ * Use an array of selectors to reach into shadow DOM trees
+ * 
+ * @param {string|string[]} selector String or array of strings with a CSS selector
+ * @param {Document} doc Optional document node
+ * @returns {Element|Null}
+ */
+export default function shadowSelect(selectors) {
```
straker, line 9: So this function has a very specific and singular purpose: to select frame elements returned from `getFrameContexts`. The frames should be selectable without the virtual tree and using the selector `getFrameContexts` returns. I think we should rename this function to be very specific as to when / why to use it (and that it works specifically in conjunction with `getFrameContexts`).`getFrameInContext` perhaps?

Also, this function cannot be used by highlighter or directly with axe result selectors since those can include iframe elements which this does not handle (e.g. `["iframe", ["#shadow", "button"]]`).

```suggestion
export default function getFrameInContext(selectors) {
```

straker (changes requested):

lib/core/utils/shadow-select.js
WilcoFiers: There is nothing in this function that makes it just about frames. This is not a good name.

straker (approved):

Issue #3045, opened 2021-06-28
axe.getFrameContexts mutates its input


WilcoFiers:
This shouldn't fail:

```js
const context = { exclude: [['iframe', 'foo']] }
axe.utils.getFrameContexts(context);
assert.deepEqual(context, { exclude: [['iframe', 'foo']] });
```

WilcoFiers:
This isn't going to need QA. Closing it.

padmavemulapati:
Validated with the latest develop branch code base, 
on passing the `axe.utils.getFrameContexts({exclude: [['iframe','#testFrame']] });` 
for a test file
"http://qateam.dequecloud.com/attest/api/test.html" , and `id` of the `iframe` being `#testFrame`
and it excluding the frame contexts, below is the screenshot,
![image](https://user-images.githubusercontent.com/47409579/128203601-811fbd7c-4bd1-42c3-a0a3-08f0d9a475f7.png)

straker:
No docs

PR #3076, opened 2021-07-12
fix(core): stop mutating Context's input

clottman:
getFrameContexts was mutating its input because Context mutates its input. This change makes it so Context no longer mutates its input. 

Closes issue: #3045

### DIFF
--- a/lib/core/base/context.js
+++ b/lib/core/base/context.js
@@ -236,12 +236,13 @@ export default function Context(spec, flatTree) {
   this.focusable = typeof spec?.focusable === 'boolean' ? spec.focusable : true;
   this.size = typeof spec?.size === 'object' ? spec.size : {};
 
-  spec = normalizeContext(spec);
-
+  const normalizedSpec = Object.assign({}, normalizeContext(spec));
   // cache the flattened tree
-  this.flatTree = flatTree ?? getFlattenedTree(getRootNode(spec));
-  this.exclude = spec.exclude;
-  this.include = spec.include;
+  this.flatTree = flatTree ?? getFlattenedTree(getRootNode(normalizedSpec));
+  this.exclude = normalizedSpec.exclude.map(x => x);
+  this.include = Array.isArray(normalizedSpec.include)
+    ? normalizedSpec.include.map(x => x)
+    : normalizedSpec.include;
 
   this.include = parseSelectorArray(this, 'include');
   this.exclude = parseSelectorArray(this, 'exclude');
--- a/lib/core/base/context.js
+++ b/lib/core/base/context.js
@@ -88,12 +88,19 @@ function normalizeContext(context) {
       context.hasOwnProperty('include') ||
       context.hasOwnProperty('exclude')
     ) {
+      // const contextDupe = JSON.parse(JSON.stringify(context));
+      let exclude = context.exclude ? context.exclude : [];
+      if (exclude !== []) {
+        exclude = Array.isArray(context.exclude)
+          ? JSON.parse(JSON.stringify(context.exclude))
+          : exclude;
+      }
       return {
         include:
           context.include && +context.include.length
-            ? context.include
+            ? Array.from(context.include)
             : [document],
-        exclude: context.exclude || []
+        exclude: exclude
       };
     }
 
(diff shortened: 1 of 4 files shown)


lib/core/base/context.js
```diff
@@ -91,9 +92,9 @@ function normalizeContext(context) {
       return {
         include:
           context.include && +context.include.length
-            ? context.include
+            ? Array.from(context.include)
             : [document],
-        exclude: context.exclude || []
+        exclude: context.exclude ? clone(context.exclude) : []
```
WilcoFiers, line 97: This isn't going to work. Exclude can contain DOM nodes. We can't clone DOM elements.

WilcoFiers (changes requested):

lib/core/base/context.js
stephenmathieson: Looks like DOM nodes aren't being cloned.

lib/core/utils/clone.js
```diff
@@ -8,6 +8,9 @@ function clone(obj) {
   var index,
     length,
     out = obj;
+  if (obj instanceof window.Node) {
```
stephenmathieson, line 11: It might seem obvious, but a quick comment here might be helpful, as this code path does not "clone" anything.

```suggestion
  // DOM nodes cannot be cloned.
  if (obj instanceof window.Node) {
```

straker (changes requested):
I think we should change the approach. Instead of trying to remember that we should clone every property of the context that may be an object (one missed here https://github.com/dequelabs/axe-core/pull/3076/files#diff-c514689a0837229f846982a6d67279f1ea867143954633477554df41dbca15e4L102), we should just clone the context object upfront as the first thing of `Context` function.

```js
export function Context(spec, flatTree) {
  spec = clone(spec);
}
```

Then we should also test is that the context returned from `new Context` doesn't strict equal any parts of the old context. Even if the context function itself doesn't mutate the object, that context is used in a ton of places so we don't want to share any memory with the original spec.

```js
// test/context.js
it('should not share memory with complex object', function() {
  var spec = { 
    include: [['#header'], ['a'], ['#frame1', '#frame2', '#fix']],
    exclude: [['iframe', '#foo']] ,
    size: { width: 100, height: 100 }
  };
  var context = new Context(spec);
  assert.notStrictEqual(spec.include, context.include);
  spec.include.forEach(function(_, index) {
    assert.notStrictEqual(spec.include[index], context.include[index]);
  });
  assert.notStrictEqual(spec.exclude, context.exclude);
  spec.exclude.forEach(function(_, index) {
    assert.notStrictEqual(spec.exclude[index], context.exclude[index]);
  });
  assert.notStrictEqual(spec.size, context.size);
});

it('should not share memory with simple array', function() {
  var spec = ['#f1']
  var context = new Context(spec);
  assert.notStrictEqual(spec.include, context.include);
});
```

lib/core/utils/clone.js
straker: In node `window.Node` doesn't exist so that's why the build is failing.

```suggestion
  if (window?.Node && obj instanceof window.Node) {
```

clottman:
@straker because of the dom node cloning issue we won't be able to ensure that the node passed in to context (if one was) is not the same node as Context is using. In fact it always will be the same node.

straker:
Correct. For the array include/exclude we should ensure no memory sharing but that can't be helped for the DOM node.

straker (approved):

lib/core/utils/clone.js
```diff
@@ -8,6 +8,10 @@ function clone(obj) {
   var index,
     length,
     out = obj;
+  // DOM nodes cannot be cloned.
+  if (window?.Node && obj instanceof window.Node) {
```
straker, line 12: In IE11 this fails as a node can also be an `HTMLCollection`.

```suggestion
  if (
    (window?.Node && obj instanceof window.Node) || 
    (window?.HTMLCollection && obj instanceof window.HTMLCollection)
  ) {
```

straker (changes requested):

straker (approved):

Issue #3082, opened 2021-07-13
role="presentation" not supported on SVG element


jhensley:
Product: axe-core

Expectation: Allow `role="presentation"` on `<svg>` elements

Actual: `role="presentation"` not seen as a valid role on `<svg>` elements

Motivation: Based on the WAI specs, any role should be valid on an `<svg>` element

`<svg role="presentation" alt="foo"></svg>`
axe-core version: 2.3.0

[![Edit axe-core react (forked)](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/axe-core-react-forked-vjpho?fontsize=14&hidenavigation=1&theme=dark)

straker:
Thanks for the issue. Our code is out-of-date with the latest version of the HTML in ARIA spec. [Back when we first took from it](https://github.com/w3c/html-aria/blob/d29561f67c3a6cf702807a0241c6171282c08b83/index.html#L2380-L2394) SVG was only allowed `application`, `document`, and `img`. [The latest spec allows all roles](https://www.w3.org/TR/html-aria/#el-svg)

jhensley:
Makes sense, I only discovered it when upgrading to axe-core 4.3.0; guessing some other check (perhaps this: https://github.com/dequelabs/axe-core/pull/2989) is picking it up or something.

PR #3086, opened 2021-07-15
fix(aria-allowed-role): allow all roles on SVGs

clottman:
The current ARIA spec allows any role on an SVG element. 

Closes issue: #3082

### DIFF
--- a/lib/standards/html-elms.js
+++ b/lib/standards/html-elms.js
@@ -822,7 +822,7 @@ const htmlElms = {
   },
   svg: {
     contentTypes: ['embedded', 'phrasing', 'flow'],
-    allowedRoles: ['application', 'document', 'img'],
+    allowedRoles: true,
     chromiumRole: 'SVGRoot',
     namingMethods: ['svgTitleText']
   },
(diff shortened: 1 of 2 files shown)


lib/standards/html-elms.js
```diff
@@ -822,7 +822,7 @@ const htmlElms = {
   },
   svg: {
     contentTypes: ['embedded', 'phrasing', 'flow'],
-    allowedRoles: ['application', 'document', 'img'],
+    allowedRoles: true,
```
WilcoFiers, line 825: Is this the only change we need to make for an updated ARIA in HTML?

clottman: That I'm not sure about, I just saw the ticket for svgs in particular being updated.

straker (changes requested):
Some other things that need updating:

* [b](https://www.w3.org/TR/html-aria/#el-b): `allowedRoles: true`
* [nav](https://www.w3.org/TR/html-aria/#el-nav):  add roles to `allowedRole`: menu, menubar, tablist
* [progress](https://www.w3.org/TR/html-aria/#el-progress): `allowedRoles: false`

WilcoFiers:
This is getting addressed by #3117

PR #3088, opened 2021-07-20
fix(run,finishRun): don't mutate options, set default reporter to v1

straker:
This fixes an issue where not passing an `options` object to `axe.finishRun` would result in axe not using the `v1` reporter even though the [v1 reporter is the default for a axe.run](https://github.com/dequelabs/axe-core/blob/develop/lib/core/public/run/normalize-run-params.js#L39). This caused the results of the `axe.finishRun` to not include `failureSummary`.

```js
const results = await axe.runPartial()
await axe.finishRun([results])  // v1 reporter not used
```

The better question would be why we [set the v2 reporter as default](https://github.com/dequelabs/axe-core/blob/develop/lib/core/core.js#L95) when axe.run uses the v1 by default.

### DIFF
--- a/lib/core/public/finish-run.js
+++ b/lib/core/public/finish-run.js
@@ -9,6 +9,9 @@ import {
 
 export default function finishRun(partialResults, options = {}) {
   options = clone(options);
+
+  // normalize the runOnly option for the output of reporters toolOptions
+  axe._audit.normalizeOptions(options);
   options.reporter = options.reporter ?? axe._audit?.reporter ?? 'v1';
 
   setFrameSpec(partialResults);
--- a/lib/core/public/finish-run.js
+++ b/lib/core/public/finish-run.js
@@ -3,10 +3,14 @@ import {
   mergeResults,
   publishMetaData,
   finalizeRuleResult,
-  DqElement
+  DqElement,
+  clone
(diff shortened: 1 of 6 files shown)


straker:
I'm not sure what we should do at this point. The test is failing because we need to mutate the `options` object to pass the normalized run options from `partialRun` to `finishRun` in order for the results of an `axe.run` and `axe.finishRun` to be the same.

The only thing I can think of is that we don't use the modified options object in the reporter but instead try to use the original options object the user uses for `toolOptions`. Then it will show exactly what the user used to run the results rather than our modified and normalized options.  

```js
const runResults = await axe.run({ runOnly: 'region' });
console.log(runResults.toolOptions.runOnly) //  { type: 'rule', values: ['region'] }

const partialResults = await axe.runPartial({ runOnly: 'region' });
const partialRunResults = await axe.finishRun({ runOnly: 'region' });
console.log(partialRunResults.toolOptions.runOnly) //  'region'

```

WilcoFiers (changes requested):
LGTM, but can you also get rid of the use of `clone` in testUtils.runPartialRecursive. We're not going to need it there anymore now that we have this in.

WilcoFiers (approved):

Issue #3093, opened 2021-07-22
Be less conservative with pseudo elements and contrast


WilcoFiers:
Marking any text near a pseudo element with a background for review has introduced a lot of noise in the axe-core results. I think in part this is because we're not considering the size of the pseudo elements. If a pseudo element is a 10x10 pixel element, if it's behind a 400x100 pixel sized paragraph, the impact on it isn't going to be significant.

I would like us to add a threshold to the color-contrast rule. If we can work out the size of the pseudo element, and it is no more than 50% the area of the of the text it is near, the pseudo element should be ignored.

PR #3097, opened 2021-07-26
fix(color-contrast): check for size before ignoring pseudo elements

WilcoFiers:
This PR adds a `pseudoSizeThreshold` options, which by default will ignore any pseudo element who's size is less than 25% the area of the element who's text is being tested. This should significantly reduce the number of `incomplete` results axe-core reports on color-contrast.

Closes issue: https://github.com/dequelabs/axe-core/issues/3093


straker (dismissed):
Checking width/height isn't sufficient. Width and height can be defined with `left`, `right`, `top`, and `bottom`, as well as percentage values of `width` and `height`.

```css
::before {
  content: '';
  position: absolute;
  left: 0;
  right: 0;
  top: 0;
  bottom: 0;
}

::before {
  content: '';
  position: absolute;
  left: 0;
  width: 100%;
  height: 50%;
}
```

To further complicate height, a percentage value depends on there being a defined height somewhere in the parent chain for it to even work. So you'd have to figure that out and then calculate all percentages down from there.

Also, we need to take into account scale transforms or anything else that can affect the size

lib/checks/color/color-contrast-evaluate.js
```diff
@@ -188,4 +148,52 @@ function colorContrastEvaluate(node, options, virtualNode) {
...
+  if (
+    getOwnBackgroundColor(style).alpha === 0 &&
+    matchPseudoStyle('background-image', 'none')
+  ) {
+    return 0; // There is no background
+  }
+
+  // Find the size of the pseudo element;
+  const pseudoWidth = parseInt(style.getPropertyValue('width'));
+  const pseudoHeight = parseInt(style.getPropertyValue('height'));
```
WilcoFiers, line 190: TODO: Check the unit here. If this isn't 'px', return Invinity. IE doesn't normalise units. They'll just have to review them.

lib/checks/color/color-contrast-evaluate.js
```diff
@@ -188,4 +148,64 @@ function colorContrastEvaluate(node, options, virtualNode) {
...
+    getOwnBackgroundColor(style).alpha === 0 &&
+    matchPseudoStyle('background-image', 'none')
+  ) {
+    return 0; // There is no background
+  }
+
+  // Find the size of the pseudo element;
+  const pseudoWidth = parseUnit(style.getPropertyValue('width'));
+  const pseudoHeight = parseUnit(style.getPropertyValue('height'));
+  if (pseudoWidth.unit !== 'px' || pseudoWidth.unit !== 'px') {
```
straker, line 191:
```suggestion
  if (pseudoWidth.unit !== 'px' || pseudoHeight.unit !== 'px') {
```

lib/checks/color/color-contrast-evaluate.js
```diff
@@ -188,4 +148,64 @@ function colorContrastEvaluate(node, options, virtualNode) {
...
+  if (ignorePseudo) {
+    return;
+  }
+  const rect = vNode.boundingClientRect;
+  const minimumSize = rect.width * rect.height * pseudoSizeThreshold;
+  do {
+    const beforeSize = getPseudoElementArea(vNode.actualNode, ':before')
+    const afterSize = getPseudoElementArea(vNode.actualNode, ':after')
+    if (beforeSize + afterSize > minimumSize) {
+      return vNode // Combined area of before and after exceeds the minimum size
```
straker, line 164: Could you rather comment on _why_ you decided to do them both combined rather than separate checks against the minimumSize? That would be more helpful when trying to debug this later

lib/checks/color/color-contrast-evaluate.js
```diff
@@ -188,4 +148,64 @@ function colorContrastEvaluate(node, options, virtualNode) {
...
+function findPseudoElement(vNode, {
+  pseudoSizeThreshold = 0.25,
+  ignorePseudo = false
+}) {
+  if (ignorePseudo) {
+    return;
+  }
+  const rect = vNode.boundingClientRect;
+  const minimumSize = rect.width * rect.height * pseudoSizeThreshold;
+  do {
```
straker, line 160: This loop is going to be slow. Before we memozied this part of the loop so looking up the parent chain wasn't so bad for every node as shared ancestors would only run through it once and then get cached. However the memoization has been removed so each element will be run through this check for every descendant.

straker (changes requested):

lib/checks/color/color-contrast-evaluate.js
```diff
@@ -188,4 +148,64 @@ function colorContrastEvaluate(node, options, virtualNode) {
...
+    const beforeSize = getPseudoElementArea(vNode.actualNode, ':before')
+    const afterSize = getPseudoElementArea(vNode.actualNode, ':after')
+    if (beforeSize + afterSize > minimumSize) {
+      return vNode // Combined area of before and after exceeds the minimum size
+    }
+  } while (vNode = vNode.parent);
+}
+
+const getPseudoElementArea = memoize(function hasPseudoElement(node, pseudo) {
```
WilcoFiers, line 169:
```suggestion
const getPseudoElementArea = memoize(function getPseudoElementArea(node, pseudo) {
```

lib/checks/color/color-contrast-evaluate.js
```diff
@@ -188,4 +148,67 @@ function colorContrastEvaluate(node, options, virtualNode) {
...
+  ) {
+    return 0; // There is no background
+  }
+
+  // Find the size of the pseudo element;
+  const pseudoWidth = parseUnit(style.getPropertyValue('width'));
+  const pseudoHeight = parseUnit(style.getPropertyValue('height'));
+  if (pseudoWidth.unit !== 'px' || pseudoHeight.unit !== 'px') {
+    // IE doesn't normalize to px. Infinity gets everything to undefined
+    return (pseudoWidth === 0 || pseudoHeight === 0
```
straker, line 193:
```suggestion
    return (pseudoWidth.value === 0 || pseudoHeight.value === 0
```

straker (approved):

Issue #3081, opened 2021-07-13
aria-supported script should use standards object


straker:
The aria-supported grunt task is out-of-date and uses the deprecated lookup table instead of the standards object

PR #3104, opened 2021-07-28
chore: use standards in aria-supported script

straker:
Currently nothing is unsupported in our standards object though, but I tested it with `aria-busy` and it worked.

Closes issue: #3081

### DIFF
--- a/build/tasks/aria-supported.js
+++ b/build/tasks/aria-supported.js
@@ -39,14 +39,14 @@ module.exports = function(grunt) {
 
       const { diff: rolesTable, notes: rolesFootnotes } = getDiff(
         roles,
-        axe.commons.aria.lookupTable.role,
+        axe.utils.getStandards().ariaRoles,
         listType
       );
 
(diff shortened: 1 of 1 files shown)


stephenmathieson (commented):
Any reason to invoke this function twice? Why not do:

```js
const {ariaRoles, ariaAttrs} = axe.utils.getStandards()
```

build/tasks/aria-supported.js
```diff
@@ -39,14 +39,14 @@ module.exports = function(grunt) {
 
       const { diff: rolesTable, notes: rolesFootnotes } = getDiff(
         roles,
-        axe.commons.aria.lookupTable.role,
+        axe.utils.getStandards().ariaRoles,
```
WilcoFiers, line 42: I agree with Stephen it's better to invoke getStandards() only once.

WilcoFiers (dismissed):

stephenmathieson (approved):

PR #3117, opened 2021-08-09
fix(aria-allowed-role): Update allowed roles based on ARIA spec updates

timogasda:
Updating some of the allowed role rules to reflect the latest changes to the ARIA spec.

* `<b>` now allows any roles
* `<nav>` now also allows `menu`, `menubar`, `tablist`
* `<svg>` now allows any roles

Based on ARIA spec: https://www.w3.org/TR/html-aria/

Closes issue: https://github.com/dequelabs/axe-core/issues/3082

### DIFF
--- a/lib/standards/html-elms.js
+++ b/lib/standards/html-elms.js
@@ -692,7 +692,7 @@ const htmlElms = {
   },
   progress: {
     contentTypes: ['phrasing', 'flow'],
-    allowedRoles: true,
+    allowedRoles: false,
     implicitAttrs: {
       'aria-valuemax': '100',
       'aria-valuemin': '0',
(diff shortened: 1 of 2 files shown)


WilcoFiers:
@timogasda Thank you for the PR! We were actually working on this too, but got a little side tracked. @straker will be reviewing this.

straker:
Nice! Could you also add `allowedRoles: false` to the [progress](https://www.w3.org/TR/html-aria/#el-progress) element? I think that's the last one for the update.

timogasda:
@straker Of course, done!

straker (approved):
Thanks for doing that. Great job.

straker:
There's currently an issue with our CircleCI config that doesn't allow the test scripts to run from outside contributors. The only work around we've found so far is for us to create a new branch and then cherry-pick your changes into it and open a new pr. I'm going to do that and then close this pr. Sorry about that.

PR #3127, opened 2021-08-19
fix(core): assign axe to global rather than window [WIP]

WilcoFiers:
This PR allows for axe to be assigned to a global object, and not have that be the actual window object.

Closes issue:


stephenmathieson (commented):
Should this be [`globalThis`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis) instead?

WilcoFiers:
Was thinking about that @stephenmathieson.

Issue #3122, opened 2021-08-18
Add ARIA 1.3 annotation roles


straker:
[ARIA 1.3 annotations](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Annotations#aria_annotations_features) adds a few new roles. We should add the roles to `standards/aria-roles` so users can use them and not have axe-core fail them as unknown roles. Using the roles doesn't cause any problems and will only add information for browsers that support them.

```html
<div role="comment"></div>
```

**Note:** This ticket _does not include_ adding `aria-description` attribute.

PR #3138, opened 2021-08-27
feat: add new ARIA roles

Zidious:
[ARIA 1.3 Annotations](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Annotations#aria_annotations_features) Added new roles: mark, suggestion, and comment.

Closes issue: https://github.com/dequelabs/axe-core/issues/3122


lib/standards/aria-roles.js
```diff
@@ -674,6 +684,12 @@ const ariaRoles = {
     nameFromContent: true,
     childrenPresentational: true
   },
+  suggestion: {
+    type: 'landmark',
+    requiredAttrs: ['insertion', 'deletion'],
+    superclassRole: ['section'],
+    prohibitedAttrs: ['aria-lebel', 'aria-labelledby']
```
straker, line 691: Typo

```suggestion
    prohibitedAttrs: ['aria-label', 'aria-labelledby']
```

lib/standards/aria-roles.js
```diff
@@ -674,6 +684,12 @@ const ariaRoles = {
     nameFromContent: true,
     childrenPresentational: true
   },
+  suggestion: {
+    type: 'landmark',
+    requiredAttrs: ['insertion', 'deletion'],
```
straker, line 689: This is [required owned roles not attributes](https://w3c.github.io/aria/#suggestion)

```suggestion
    requiredOwned: ['insertion', 'deletion'],
```

straker (changes requested):
Great start! Since some of these roles have attributes, owned elements, or prohibited attributes, we'll need to add a few more tests to make sure they are all covered

* [test/integration/rules/aria-allowed-attr/failure](https://github.com/dequelabs/axe-core/blob/develop/test/integration/rules/aria-allowed-attr/failures.html) - add elements to ensure `mark` and `suggestion` do not allow `aria-label` and `aria-labelledby`
* [test/integration/rules/aria-allowed-attr/passes](https://github.com/dequelabs/axe-core/blob/develop/test/integration/rules/aria-allowed-attr/passes.html) - add element for `comment` to allow `aria-level`, `aria-posinset`, `aria-setsize`
* [test/integration/rules/aria-required-children](https://github.com/dequelabs/axe-core/blob/develop/test/integration/rules/aria-required-children/aria-required-children.html) - add pass and fail tests for `suggestion` having `insertion` and `deletion` children/`aria-owns`

test/integration/rules/aria-required-children/aria-required-children.html
```diff
@@ -55,3 +55,12 @@
     <li role="menuitem" id="ignore14">>Trash</li>
   </ul>
 </div>
+
+<div role="suggestion" id="pass8">
+  <span role="deletion">option</span>
+  <span role="insertion">option</span>
+</div>
+
+<div role="suggestion" id="fail8">
```
straker, line 64: Good start. So with our integration tests we want to be really thorough and test all possible states, so we'll want 3 more tests added. One for `suggestion` without any children, and one for `suggestion` with just an `insertion` child, and one for `suggestion` with just a `deletion` child.

test/integration/rules/aria-allowed-attr/passes.html
```diff
@@ -1941,3 +1941,7 @@
     </tr>
   </tbody>
 </table>
+
+<div role="comment" aria-level="1" id="pass91">ok</div>
+<div role="comment" aria-posinset="1" id="pass92">ok</div>
+<div role="comment" aria-setsize="16" id="pass93">ok</div>
```
straker, line 1947: For this particular test we can do this in just a single element since if any one of the attributes is not allow the element will fail.

```suggestion
<div 
  id="pass91"
  role="comment" 
  aria-level="1" 
  aria-posinset="1" 
  aria-setsize="16" 
>ok</div>
```

test/integration/rules/aria-allowed-attr/passes.json
```diff
@@ -93,6 +93,9 @@
     ["#pass88"],
     ["#pass89"],
     ["#pass90"],
-    ["#treegrid"]
+    ["#treegrid"],
+    ["#pass91"],
+    ["#pass92"],
+    ["#pass93"]
```
straker, line 99:
```suggestion
    ["#treegrid"],
    ["#pass91"]
```

straker (dismissed):

test/integration/rules/aria-required-children/aria-required-children.html
Zidious: Unable to test for one child due to `requiredOwned` (as discussed). However [ARIA Suggestion Docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Suggestion_role) say that both need to be present, playground throws no errors if using one child

straker (commented):
LGTM. Will wait for @WilcoFiers final approve.

@WilcoFiers the main question we have is how to handle the `role=suggestion` element requiring that it owns both a `role=insertion` and `role=deletion` element. Currently `aria-required-children` will pass if at least one role is owned, but not enforce all roles.

WilcoFiers:
@straker Suggestion doesn't require both roles. MDN is wrong about that:

> Authors MUST ensure that a suggestion contains either one insertion child or one deletion child or ensure that it contains two children where one is an insertion and the other is a deletion. Authors MUST ensure a suggestion does not contain any other children.
> https://w3c.github.io/aria/#suggestion

Either way, it's so early days on this spec, I don't think we should be too harsh on testing just yet.

WilcoFiers (dismissed):

lib/standards/aria-roles.js
```diff
@@ -391,6 +396,11 @@ const ariaRoles = {
     accessibleNameRequired: true,
     childrenPresentational: true
   },
+  mark: {
+    type: 'landmark',
```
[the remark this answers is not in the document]
Zidious: Done

lib/standards/aria-roles.js
```diff
@@ -674,6 +684,12 @@ const ariaRoles = {
     nameFromContent: true,
     childrenPresentational: true
   },
+  suggestion: {
+    type: 'landmark',
```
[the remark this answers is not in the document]
Zidious: Done

lib/standards/aria-roles.js
```diff
@@ -147,6 +147,11 @@ const ariaRoles = {
     allowedAttrs: ['aria-expanded'],
     superclassRole: ['landmark']
   },
+  comment: {
+    type: 'landmark',
+    allowedAttrs: ['aria-level', 'aria-posinset', 'aria-setsize'],
+    superclassRole: ['landmark']
```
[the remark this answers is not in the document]
Zidious: Done

WilcoFiers (approved):

Issue #3108, opened 2021-08-03
Consider allowing header[role=banner]


WilcoFiers:
```html
  <main>
    <header role="banner">
      Hello world
    </header>
  </main>
```

This fails `aria-allowed-role`, in addition to failing `landmark-banner-is-top-level`. It's probably better for axe-core to only fail it under the top-level rule. It's a little odd to allow role=banner on a `header` if it's at the top, but to fail it if it's inside another landmark.

The same problem exists for other landmarks. I think the fix is to add `banner` to the header element in standards/html-elms.js.

WilcoFiers:
Actually it'd be pretty cool if axe-core explicitly reported on the top-level rule to remove the role attribute.

padmavemulapati:
Validated with the latest develop branch code base, now its showing only one rule failing for the code snippet
```
<main>
    <header role="banner">
      Hello world
    </header>
  </main>
```
when header is having `role=banner` 
![image](https://user-images.githubusercontent.com/47409579/187240247-10ec140e-3a2a-4a8e-867b-2cac795e9ab7.png)

Previouly both `aria-allowed-role` and `landmark` rules failing.
<img width="1461" alt="image" src="https://user-images.githubusercontent.com/47409579/187240439-9cec7219-0c2b-4259-9aa6-6dd2cc15eacc.png">

PR #3142, opened 2021-08-31
fix(aria-allowed-role): landmark roles banner on header and contentinfo on footer to only report on top-level rule

Zidious:
landmark roles `banner` on header and `contentinfo` on footer to report only top-level rule and fail `aria-allowed-role` when `allowImplicit: false`.

Clean up `get-unallowed-roles` to fail aria-role and not all roles.

Closes Issue: https://github.com/dequelabs/axe-core/issues/3108 & https://github.com/dequelabs/axe-core/issues/3213


WilcoFiers (changes requested):
I'd like to see this addressed:
https://github.com/dequelabs/axe-core/issues/3108#issuecomment-892010812

Plus, my comment mentions that there are other elements for which this issue exists. This PR should include all of them IMO. As it is, this seems incomplete to me.

WilcoFiers (dismissed):

lib/commons/aria/get-element-unallowed-roles.js
```diff
@@ -71,23 +71,14 @@ function getElementUnallowedRoles(node, allowImplicit = true) {
...
-    ) {
-      return true;
+    const roleMatches =
+      role === implicitRole ||
+      // allow landmark roles to use their implicit role inside of another landmark
+      // @see https://github.com/dequelabs/axe-core/pull/3142
+      landmarkRoles[nodeName] === role;
+
+    if (roleMatches) {
+      return !allowImplicit;
```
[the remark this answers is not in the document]
Zidious: good catch, added tests for this case too.

lib/commons/aria/is-aria-role-allowed-on-element.js
```diff
@@ -15,17 +15,17 @@ function isAriaRoleAllowedOnElement(node, role) {
...
-
   const spec = getElementSpec(vNode);
 
   if (Array.isArray(spec.allowedRoles)) {
     return spec.allowedRoles.includes(role);
   }
 
+  // always allow the explicit role to match the implicit role
+  if (role === implicitRole) {
+    console.log('Is this hit in any case?');
```
WilcoFiers, line 26:
```suggestion
```

Zidious: whoops, removed.

lib/commons/aria/is-aria-role-allowed-on-element.js
```diff
@@ -15,17 +15,17 @@ function isAriaRoleAllowedOnElement(node, role) {
...
-    return true;
-  }
-
   const spec = getElementSpec(vNode);
 
   if (Array.isArray(spec.allowedRoles)) {
     return spec.allowedRoles.includes(role);
   }
 
+  // By default, ARIA in HTLM does not allow implicit roles to be the same as explicit ones
```
WilcoFiers, line 24:
```suggestion
  // By default, ARIA in HTML does not allow implicit roles to be the same as explicit ones
```

lib/commons/aria/get-element-unallowed-roles.js
```diff
@@ -22,6 +17,11 @@ const dpubRoles = [
   'doc-noteref'
 ];
 
+const landmarkRoles = {
```
dylanb, line 20: nit: aren't there other landmark roles? What is special about these two?

lib/commons/aria/get-element-unallowed-roles.js
```diff
@@ -59,50 +56,32 @@ function getRoleSegments(vNode) {
...
-
   // by pass custom elements
   if (!isHtmlElement(vNode)) {
     return [];
   }
-
+  // allow landmark roles to use their implicit role inside another landmark
+  // @see https://github.com/dequelabs/axe-core/pull/3142
+  const { nodeName } = vNode.props;
+  const implicitRole = getImplicitRole(vNode) || landmarkRoles[nodeName];
```
dylanb, line 66: is nodeName always lower case (XHTML)?

test/commons/aria/get-element-unallowed-roles.js
```diff
@@ -62,25 +62,126 @@ describe('aria.getElementUnallowedRoles', function() {
...
+
+  it('returns empty on role=button on div elm when role is not implicit and allowImplicit: false', function() {
+    var node = document.createElement('div');
+    node.setAttribute('role', 'button');
+    flatTreeSetup(node);
+    var actual = getElementUnallowedRoles(node, false);
     assert.isEmpty(actual);
   });
 
-  it('returns false with implicit role of row for TR when allowImplicit is set to false via options', function() {
```
dylanb, line 74: Why did we get rid of this case?

lib/commons/aria/get-element-unallowed-roles.js
WilcoFiers: These are the only two elements that change role depending on if they are in a landmark. If you do `main > header`, the header element is not a `banner`. If you just do `body > header` it is.

This causes the odd behavior that `body > header[role=banner]` passes, but `body > main > header[role=banner]` fails. That last case should fail for being a nested landmark (which it does), but not for having a role on an element that doesn't allow it.

lib/commons/aria/get-element-unallowed-roles.js
WilcoFiers: It is. We cast nodeName to lower case before assigning it to vNode.props. It's more consistent that way.

test/commons/aria/get-element-unallowed-roles.js
WilcoFiers: Agreed, this can probably be put back

lib/commons/aria/get-element-unallowed-roles.js
dylanb: yeah, I was commenting on the name of the variable, makes it seem like it is all the landmark roles

dylanb (changes requested):
put the `<tr>` test case back

dylanb (approved):

Issue #2918, opened 2021-05-05
Test axe-core against ARIA practices


WilcoFiers:
We have from time to time had false positives because we've overlooked different ARIA patterns. Axe-core would be better if we proactively tested it against the ARIA practices. We could perhaps do this in a similar way to how we test ACT rules.

Here is the repo we'd have to check out https://github.com/w3c/aria-practices

Malvoz:
This issue should be closed per https://github.com/dequelabs/axe-core/pull/3144.

WilcoFiers:
Yes it should be. Thanks @Malvoz.

PR #3144, opened 2021-09-02
chore: Test against ARIA practices

WilcoFiers:
One of the more common reasons for reported false positives comes from axe-core failing something recommended in the ARIA practices. This PR adds a test script will have axe-core running against the ARIA practices.

- [x] Set up basic test script
- [x] Open issues for problems discovered in the test:apg
- [x] Lock the APG dependency to a fix commit for stable testing
- [x] Set up CI, including a nightly test where we're running against the latest on #main

Closes issue: #2918


WilcoFiers:
PR is blocked until the aria-practices removes husky from its postinstall script. I opened a PR for that: https://github.com/w3c/aria-practices/pull/2033

.circleci/config.yml
```diff
@@ -354,6 +379,9 @@ workflows:
                 - develop
     jobs:
       - dependencies_unix
-      - test_nightly:
+      - test_nightly_chrome:
```
stephenmathieson, line 382: How will we know if/when this fails? Is there someone responsible for checking CI logs daily?

test/aria-practices/apg.spec.js
```diff
@@ -0,0 +1,85 @@
...
+const express = require('express');
+const chromedriver = require('chromedriver');
+const AxeBuilder = require('@axe-core/webdriverjs');
+const testListen = require('test-listen');
+const { getWebdriver, connectToChromeDriver } = require('./run-server');
+const { assert } = require('chai');
+const globby = require('globby');
+
+describe('aria-practices', function () {
+  const apgPath = path.resolve(__dirname, '../../node_modules/aria-practices/');
```
stephenmathieson, line 13:
```suggestion
  const apgPath = require.resolve('aria-practices')
```

test/aria-practices/apg.spec.js
```diff
@@ -0,0 +1,85 @@
...
+      
+      const builder = new AxeBuilder(driver, axeSource);
+      builder.disableRules([
+        ...disabledRules['*'],
+        ...(disabledRules[filePath] || []),
+      ]);
+      
+      const { violations } = await builder.analyze();
+      const issues = violations.map(({ id, nodes }) => ({ id, issues: nodes.length }))
+      assert.deepEqual(issues, []);
```
stephenmathieson, line 82:
```suggestion
      assert.lengthOf(issues, 0);
```

.circleci/config.yml
WilcoFiers: Yeah, those get kicked out to our #attest-build channel on Slack.

test/aria-practices/apg.spec.js
WilcoFiers: Gd point. I originally started with those in another dir, but this is better.

WilcoFiers: Never mind. I remember why. aria-practices doesn't have a package.json. Node complains about that.

.circleci/config.yml
```diff
@@ -5,7 +5,7 @@ defaults: &defaults
 
 unix_box: &unix_box
   docker:
-    - image: circleci/node:10-browsers
+    - image: circleci/node:16-browsers
```
WilcoFiers, line 8: So that we get access to Chrome 95

test/aria-practices/apg.spec.js
```diff
@@ -0,0 +1,88 @@
...
+  this.retries(3);
+
+  before(async () => {
+    const axePath = require.resolve('../../axe.js');
+    axeSource = fs.readFileSync(axePath, 'utf8');
+    chromedriver.start([`--port=${port}`]);
+    await new Promise(r => setTimeout(r, 500));
+    await connectToChromeDriver(port);
+
+    const app = express();
```
straker, line 29: Are any of the test files using absolute paths? If not I think our `npm run start` command should server node_modules directly as well, so we might not need express for this.

package.json
```diff
@@ -106,16 +107,18 @@
     "fmt": "prettier --write .*.json *.{json,md,js} **/*.ts './{.circleci,.github,build,doc,lib,locales,test,typings}/**/*.{json,md,js,ts,html}'"
   },
   "devDependencies": {
+    "@axe-core/webdriverjs": "^4.2.2",
```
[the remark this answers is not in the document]
WilcoFiers: Don't think so. It's a dev dependency.

straker (changes requested):

straker (approved):

Issue #3130, opened 2021-08-23
p-as-heading rule should account for text length


WilcoFiers:
Axe-core's p-as-heading rule is reporting an incorrect issue on the Azure page:
https://azure.microsoft.com/en-us/services/active-directory/external-identities/#customers

This problem can be avoided by checking the length of the possible heading, compared to the length of the supposed paragraph. I would say that if the heading is longer than the paragraph, we pass it. If it's less than twice as long, we set it for review, and only if the suspected fake heading is at least 200% shorter than the paragraph do we allow it to be failed.

The way to do this would be in the `p-as-heading-evaluate.js` file, to grab textContent of both nodes and compare the length. We should probably make these configurable through a check option as well.

WilcoFiers:
Not the easiest "good first issue", but doable for anyone comfortable with JS, and willing to dig a little.

ilyaryabchinski:
May I take work on this one?

Zidious:
> May I take work on this one?

Sorry! I am currently working on this ticket.

ilyaryabchinski:
No worries!

dan-tripp:
I was looking at this issue too.  I'm happy to leave it to @Zidious - but @WilcoFiers , I don't suppose you have any more "good first issues" that I could work on?   I don't see any more which don't have an assignee or a PR.

padmavemulapati:
Validated with the latest axe-core develop branch code base, 
as per the fix validations went like:
If the heading is longer than the paragraph, it will be  pass . 
If it's less than twice as long, it will go for review, 
and only if the suspected heading is at least 200% shorter than the paragraph it will fail.


![image](https://user-images.githubusercontent.com/47409579/137679728-e7da44aa-6123-43aa-80a6-afe82242777c.png)

PR #3145, opened 2021-09-03
fix(p-as-heading): `p-as-heading` rule to account for `textContent` length

Zidious:
Ref: #3130


straker:
@WilcoFiers we had questions on this requirement:

> I would say that if the heading is longer than the paragraph, we pass it. If it's less than twice as long, we set it for review, and only if the suspected fake heading is at least 200% shorter than the paragraph do we allow it to be failed.

There's a gap in the algorithm, like what happens if it's more than twice as long but not greater than the length. Could you give a bit more clarification? Here's what I have reading that description:

```js
const headingLength = headingEl.textContent.length;
const paragraphLength = pEl.textContext.length;

if (headingLength > paragraphLength) {
  return true;
}
else if (headingLength < paragraphLength / 2) {
  return false;
} 
else {
  return undefined;
}

**Comment by WilcoFiers** [comment]
@straker I think your dummy code is pretty much how I think of it. We should probably put in options to make these values configurable. Beyond that, these seem like reasonable heuristics for this rule.

**Comment by Zidious** [comment]
> @straker I think your dummy code is pretty much how I think of it. We should probably put in options to make these values configurable. Beyond that, these seem like reasonable heuristics for this rule.

@WilcoFiers Could you provide a bit more info on configurable by options?

File: lib/checks/navigation/p-as-heading-evaluate.js [hunk 1]
```diff
@@ -79,6 +79,18 @@ function pAsHeadingEvaluate(node, options, virtualNode) {
...
 
+  if (node && nextSibling) {
+    const headingLength = node.textContent.trim().length;
+    const paragraphLength = nextSibling.textContent.trim().length;
+    if (headingLength > paragraphLength) {
+      return true;
+    } else if (headingLength < paragraphLength / 2) {
+      return false;
+    } else {
+      undefined;
```
WilcoFiers, line 90: Did you mean to return here? ;)

You'll also need to add an `incomplete` message to the rule metadata.

test/checks/navigation/p-as-heading.js
```diff
@@ -141,6 +141,26 @@ describe('p-as-heading', function() {
     );
   });
 
+  it('returns true if the heading elm length is greater than the paragraph elm', function() {
+    var params = checkSetup(
+      '<p id="target">elm1elm1</p>' + '<p>elm2</p>',
```
WilcoFiers, line 146:
```suggestion
      '<p id="target"><b>elm1elm1</b></p>' + '<p>elm2</p>',
```

Otherwise this isn't proving the case.

test/checks/navigation/p-as-heading.js
```diff
@@ -141,6 +141,26 @@ describe('p-as-heading', function() {
     );
   });
 
+  it('returns true if the heading elm length is greater than the paragraph elm', function() {
```
WilcoFiers, line 144: You're missing a test for the `undefined` case.

WilcoFiers (changes requested):

lib/checks/navigation/p-as-heading-evaluate.js
```diff
@@ -79,6 +79,24 @@ function pAsHeadingEvaluate(node, options, virtualNode) {
...
+    const paragraphLength =
+      nextSibling.textContent.trim().length * optionsParagraphLength;
+
+    if (headingLength > paragraphLength) {
+      return true;
+    } else if (headingLength < paragraphLength / 2) {
+      return false;
+    } else if (headingLength < paragraphLength) {
+      return undefined;
+    }
```
straker, line 97: Should use numbers instead of arrays. example (naming TBD)

```js
passLength: 1,
failLength: 0.5

headingLength = node.textContent.trim().length
paragraphLength = nextSibling.textContent.trim().length

if (headingLength > paragraphLength * passLength){
  return true
}
else if (headingLength < paragraphLength * failLength) {
  return false
}
```

lib/checks/navigation/p-as-heading.json
```diff
@@ -18,13 +18,16 @@
       {
         "size": 1.4
       }
-    ]
+    ],
+    "passLength": 1,
+    "failLength": 0.5
```
WilcoFiers, line 23: Options need documentation.

WilcoFiers (changes requested):
One comment, rest looks good.

doc/check-options.md
```diff
@@ -415,8 +415,25 @@ h6:not([role]),
...
-      <td align="left">Common CSS values used to display `p` elements as `h1-h6` elements determining if a `p` element is being improperly repurposed</td>
     </tr>
+    <tr>
+     <td>
+        <code>passLength</code>
+      </td>
+      <td align="left">
+        <pre lang=js><code>"passLength": 1</code></pre>
+        </td>
+      <td align="left">Valid length used to check if a paragraph element is not being used as a heading</td>
```
WilcoFiers, line 426:
```suggestion
      <td align="left">Relative length, if the the candidate heading is X times or greater the length of the candidate paragraph, it will pass.</td>
```

doc/check-options.md
```diff
@@ -415,8 +415,25 @@ h6:not([role]),
...
+      </tr>
+       <tr>
+     <td>
+        <code>faiLength</code>
+      </td>
+      <td align="left">
+        <pre lang=js><code>"failLength": 0.5</code></pre>
+        </td>
+      <td align="left">Invalid length used to check if a paragraph element is being used as a heading</td>
```
WilcoFiers, line 435:
```suggestion
      <td align="left">Relative length, if the the candidate heading is X times or less the length of the candidate paragraph, it can fail.</td>
```

WilcoFiers (changes requested):

WilcoFiers (approved):

PR #3154, opened 2021-09-08
fix(nativeSelectValue): update selected value on change

kevin940726:
`nativeSelectValue` doesn't return the selected value after manual selection. This PR fixes that by using the [`selected`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/option#attr-selected) on `<option>` to get the up-to-date value.

### DIFF
--- a/lib/commons/text/form-control-value.js
+++ b/lib/commons/text/form-control-value.js
@@ -106,7 +106,7 @@ function nativeSelectValue(node) {
   }
 
   const options = querySelectorAll(vNode, 'option');
-  const selectedOptions = options.filter(option => option.actualNode.selected);
+  const selectedOptions = options.filter(option => option.props.selected);
 
   // browser automatically selects the first option
   if (!selectedOptions.length) {
--- a/lib/commons/text/form-control-value.js
+++ b/lib/commons/text/form-control-value.js
@@ -106,7 +106,7 @@ function nativeSelectValue(node) {
   }
 
   const options = querySelectorAll(vNode, 'option');
-  const selectedOptions = options.filter(option => option.hasAttr('selected'));
+  const selectedOptions = options.filter(option => option.actualNode.selected);
 
   // browser automatically selects the first option
   if (!selectedOptions.length) {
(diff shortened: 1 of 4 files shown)


lib/commons/text/form-control-value.js
```diff
@@ -106,7 +106,7 @@ function nativeSelectValue(node) {
   }
 
   const options = querySelectorAll(vNode, 'option');
-  const selectedOptions = options.filter(option => option.hasAttr('selected'));
+  const selectedOptions = options.filter(option => option.actualNode.selected);
```
stephenmathieson, line 109: I believe this means dependent checks can no longer operate on virtual nodes, which is something we've been pushing for lately. We may need to find a different way to solve this issue, as I don't think we want to cause a regression while fixing a bug.

straker: What we'll probably need to do is add `selected` as a [prop](https://github.com/dequelabs/axe-core/blob/develop/lib/core/base/virtual-node/virtual-node.js#L57-L80) of the virtual node class so we can access it that way

```js
const selectedOptions = options.filter(option => option.props.selected);
```

kevin940726: Oh I didn't know that 😅. I pushed a change in [`3a71bff` (#3154)](https://github.com/dequelabs/axe-core/pull/3154/commits/3a71bffa0b084356d16bc49a8f7acefebb64b7fa) to use `props.selected` instead. Not sure if my changes make sense though.

straker (changes requested):
Awesome. That looks better. One last thing, could you add a test to https://github.com/dequelabs/axe-core/blob/develop/test/core/base/virtual-node/virtual-node.js? I know we don't have tests for the node specific props (`multiple`, `nodeValue`, `value`, etc.), but I think that might be a failing on our part.

The test can be as simple "node with `selected` property is reflected in `props`" or something, just so we don't delete it one day without knowing.

test/core/base/virtual-node/virtual-node.js
```diff
@@ -97,6 +107,32 @@ describe('VirtualNode', function() {
       });
     });
 
+    describe('props', function() {
```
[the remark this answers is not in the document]
kevin940726: Oops! deleted!

straker (approved):
Great work! Thanks for fixing this.

straker:
Reviewed for security

Issue #3149, opened 2021-09-07
Bug: reports incorrect error on aria-errormessage (ARIA 1.2) 


Zidious:
Ref: https://github.com/dequelabs/axe-core-npm/issues/328

[ARIA 1.2](https://www.w3.org/TR/wai-aria-1.2/#aria-errormessage)

axe-core not reporting that content is hidden when `aria-errormessage` is referenced by a `hidden` element.

dylanb:
Is the error incorrect or is it just missing the information about the content being hidden?
What about `aria-labelledby` as a way to announce the error message in addition to the ones we currently check for?

PR #3156, opened 2021-09-13
fix: check for hidden elements on `aria-errormessage`          

Zidious:
Check if hidden elements are present on `aria-errormessage` via `isVisible()` and add tests.

Closes issue: https://github.com/dequelabs/axe-core/issues/3149

### DIFF
--- a/lib/checks/aria/aria-errormessage-evaluate.js
+++ b/lib/checks/aria/aria-errormessage-evaluate.js
@@ -55,7 +55,7 @@ function ariaErrormessageEvaluate(node, options, virtualNode) {
     }
 
     if (idref) {
-      if (!isVisible(idref)) {
+      if (!isVisible(idref, true)) {
         this.data({
           messageKey: 'hidden',
           values: tokenList(attr)
--- a/lib/checks/aria/aria-errormessage-evaluate.js
+++ b/lib/checks/aria/aria-errormessage-evaluate.js
@@ -1,7 +1,7 @@
 import standards from '../../standards';
 import { idrefs } from '../../commons/dom';
 import { tokenList } from '../../core/utils';
-
+import { isVisible } from '../../commons/dom';
 /**
  * Check if `aria-errormessage` references an element that also uses a technique to announce the message (aria-live, aria-describedby, etc.).
  *
@@ -55,6 +55,13 @@ function ariaErrormessageEvaluate(node, options, virtualNode) {
     }
 
     if (idref) {
+      if (!isVisible(idref)) {
+        this.data({
+          messageKey: 'hidden',
+          values: tokenList(attr)
+        });
+        return false;
+      }
       return (
         idref.getAttribute('role') === 'alert' ||
         idref.getAttribute('aria-live') === 'assertive' ||
@@ -62,6 +69,7 @@ function ariaErrormessageEvaluate(node, options, virtualNode) {
         tokenList(virtualNode.attr('aria-describedby')).indexOf(attr) > -1
       );
     }
+
     return;
   }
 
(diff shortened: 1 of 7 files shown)


lib/checks/aria/aria-errormessage-evaluate.js
```diff
@@ -55,13 +55,21 @@ function ariaErrormessageEvaluate(node, options, virtualNode) {
     }
 
     if (idref) {
+      if (!isVisible(idref)) {
```
WilcoFiers, line 58: You'll want to check that it's visible to screen readers, not that it's on screen:

```suggestion
      if (!isVisible(idref, true)) {
```

Please add tests checking that off screen messages don't error, and that a message with aria-hidden on it does.

WilcoFiers (changes requested):
I feel like adding at least one integration test here would be a good idea.

test/checks/aria/errormessage.js
```diff
@@ -146,7 +146,7 @@ describe('aria-errormessage', function() {
 
   it('should return false when hidden attribute is used', function() {
     var vNode = queryFixture(
-      '<input type="text" id="target" aria-invalid="true" aria-errormessage="id-message-1"></div>' +
+      '<div><input type="text" id="target" aria-invalid="true" aria-errormessage="id-message-1"></div>' +
```
WilcoFiers, line 149: Why do you need the div?

test/integration/rules/aria-valid-attr-value/aria-valid-attr-value.html
```diff
@@ -51,6 +51,46 @@ <h2>Violations</h2>
   <div aria-posinset="-22" id="violation42">hi</div>
 
   <div aria-setsize="-22" id="violation43">hi</div>
+
```
WilcoFiers, line 54: Can you add an example with a hidden error, that passes because it has aria-invalid="false"?

WilcoFiers (changes requested):
Very minor stuff.

test/checks/aria/errormessage.js
Zidious: Good find

test/integration/rules/aria-valid-attr-value/aria-valid-attr-value.html
```diff
@@ -276,6 +316,35 @@ <h2>Possible False Positives</h2>
...
+      aria-errormessage="pass188"
+    />
+  </div>
+
+  <div>
+    <input
+      type="text"
+      id="pass190"
+      aria-invalid="false"
+      aria-errormessage="pass68"
```
WilcoFiers, line 344: That's not a hidden element.

test/checks/aria/errormessage.js
```diff
@@ -144,6 +144,95 @@ describe('aria-errormessage', function() {
...
+    var vNode = queryFixture(
+      '<input type="text" id="target" aria-invalid="true" aria-errormessage="id-message-1">' +
+        '<div id="id-message-1" hidden>Error message 1</div>'
+    );
+    assert.isFalse(
+      axe.testUtils
+        .getCheckEvaluate('aria-errormessage')
+        .call(checkContext, null, null, vNode)
+    );
+    console.log(checkContext._data);
```
WilcoFiers, line 157: This

WilcoFiers (changes requested):

test/integration/rules/aria-valid-attr-value/aria-valid-attr-value.html
Zidious: Ah, I see it, good catch.

WilcoFiers (approved):

PR #3159, opened 2021-09-15
chore: hidden and aria-hidden for nested-interactive

oscarmarina:
The standard technique for hiding content for all users has been the use CSS display:none. Now, both ARIA and HTML5 also provide a semantic indication of content state that indicates content is hidden or should not be available to users

- [HTML5 Accessibility Chops: hidden and aria-hidden - Steve Faulkner](https://www.tpgi.com/html5-accessibility-chops-hidden-and-aria-hidden/)
- [The state of hidden content support in 2016 - Steve Faulkner](https://www.tpgi.com/the-state-of-hidden-content-support-in-2016/)

In nested interactive controls they would have the same behavior if hidden pass the test; aria-hidden=true test would behave in the same way

### DIFF
--- a/test/integration/rules/nested-interactive/nested-interactive.html
+++ b/test/integration/rules/nested-interactive/nested-interactive.html
@@ -3,7 +3,7 @@
 <div role="tab" id="pass3">pass</div>
 <div role="checkbox" id="pass4">pass</div>
 <div role="radio" id="pass5"><span>pass</span></div>
-<div role="button" id="pass6"><button aria-hidden="true" tabindex="-1">pass</button></div>
+<div role="button" id="pass7"><button aria-hidden="true" tabindex="-1">pass</button></div>
 
 <button id="fail1"><span tabindex="0">fail</span></button>
 <div role="button" id="fail2"><input /></div>
--- a/test/integration/rules/nested-interactive/nested-interactive.html
+++ b/test/integration/rules/nested-interactive/nested-interactive.html
@@ -3,12 +3,15 @@
 <div role="tab" id="pass3">pass</div>
 <div role="checkbox" id="pass4">pass</div>
 <div role="radio" id="pass5"><span>pass</span></div>
+<div role="button" id="pass6"><button aria-hidden="true" tabindex="-1">pass</button></div>
 
 <button id="fail1"><span tabindex="0">fail</span></button>
 <div role="button" id="fail2"><input /></div>
 <div role="tab" id="fail3"><button id="pass6">fail</button></div>
 <div role="checkbox" id="fail4"><a href="foo.html">fail</a></div>
 <div role="radio" id="fail5"><span tabindex="0">fail</span></div>
+<div role="button" id="fail6"><button aria-hidden="true">fail</button></div>
+<div role="button" id="fail7"><button tabindex="-1">fail</button></div>
 
 <a id="ignored1" href="foo.html">ignored</a>
 <span id="ignored2">ignored</span>
--- a/test/integration/virtual-rules/nested-interactive.js
+++ b/test/integration/virtual-rules/nested-interactive.js
@@ -94,6 +94,30 @@ describe('nested-interactive virtual-rule', function() {
     assert.lengthOf(results.incomplete, 0);
   });
 
+  it('should pass for element with interactive control with aria-hidden="true"', function() {
+    var node = new axe.SerialVirtualNode({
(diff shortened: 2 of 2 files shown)


straker:
Thanks for the pr. Unfortunately I don't think this works as expected. Using `display: none` removes the element both visually and from the tab order, while `aria-hidden` does neither of those. Testing this out in different screen reader / browser combinations:

```html
<div role="button"><button aria-label="true">My Button</button></div>
```

- IE / JAWS - Can tab to the button and is read as "Unlabeled 0 button"
- Edge / JAWS - Can tab to the button and is read as "My Button, button", activating the button moves focus off of it
- FireFox / NVDA - Can tab to the button and nothing is read out, sometimes just "blank"
- Safari / VO - Can tab to the button and is read as "button" (no label). Furthermore you can press "down" to enter the text and "My Button" is read.

oscarmarina:
I agree the correct combination would be `aria-hidden="true"` and `tabindex="-1"` as explainend in
 [accessibilityinsights.io → aria-hidden-focus](https://accessibilityinsights.io/info-examples/web/aria-hidden-focus/)

- Element with aria-hidden="true" contains only focusable elements that are disabled or not tabbable


For me this is a variant of the same thing but I don't know if you shared this opinion.

straker:
So talking with Wilco we determined that since the example fails `aria-command-name` and `aria-hidden-focus`, it doesn't also need to fail `nested-interactive`. Furthermore, since the description of the rule is `Nested interactive controls are not announced by screen readers`, adding `aria-hidden=true` passes this description as the button is no longer announced to a screen reader.

test/integration/rules/nested-interactive/nested-interactive.html
```diff
@@ -3,12 +3,15 @@
 <div role="tab" id="pass3">pass</div>
 <div role="checkbox" id="pass4">pass</div>
 <div role="radio" id="pass5"><span>pass</span></div>
+<div role="button" id="pass6"><button aria-hidden="true" tabindex="-1">pass</button></div>
```
straker, line 6: There's already an `id="pass6"` on line 10.

```suggestion
<div role="button" id="pass7"><button aria-hidden="true" tabindex="-1">pass</button></div>
```

straker (changes requested):
You'll also need to edit https://github.com/dequelabs/axe-core/blob/develop/test/integration/rules/nested-interactive/nested-interactive.json and add the new `ids` to the list of fail / passes.

scottaohara:
Tested the following snippet with a colleague the other day - with Dragon both checkboxes in the following markup snippet were still exposed;
```html
<div role=checkbox aria-checked=true tabindex=0 aria-label=foo>
  <input type=checkbox aria-hidden=true tabindex=-1> <!-- visually hidden -->
</div>
```
Using VoiceOver and the down arrow key alone to read line-by-line, the visually hidden checkbox is still reached, causing a focus stop without an announcement.

While screen readers are trying to error correct for author blunders like these, it's not bullet proof across all other AT - more of which still needs testing.   I'd request that this be taken into consideration before passing stuff that can still pose issues, and would not discourage authors from poor coding practices.

Thanks

WilcoFiers:
Apologies for the slowed response. I've been a little reluctant to dive into this. Scott is right in that this pattern just doesn't work very well, and that you probably just shouldn't use it.

I think what needs to happen is that we create a new failure message, specifically to explain that using negative tabindex can still cause problems. I'll create an issue for this.

straker:
I believe this pr is replaced by https://github.com/dequelabs/axe-core/pull/3194, since we don't want tabindex=-1 to pass but now warn of the problem.

straker:
Going to close this issue. Negative tabindex will now produce a warning stating that it won't prevent assistive technologies from focusing the element. In terms of the number of rule failures, I believe all 3 should be reported for the following reasons:

If the user starts with this HTML:

```html
<div role="button"><button aria-hidden="true">My Button</button></div>
```

We report 3 failures: `aria-hidden-focus`, `button-name`, and `nested-interactive`. So the user then does this:

```html
<div role="button" aria-label="My Button"><button tabindex="-1" aria-hidden="true">My Button</button></div>
```

Which should correctly produce a `nested-interactive` failure with the message warning of `tabindex=-1`. If we were to prevent `nested-interactive`, we'd first have to know what other rules failed, which we can't do. So it's better that all 3 fail the first time as they are all 3 different problems on the structure.

Issue #3151, opened 2021-09-07
 Bug: does not report error incorrect use of aria attributes on element with role="row" (ARIA 1.2)


Zidious:
Ref: https://github.com/dequelabs/axe-core-npm/issues/317

[ARIA 1.2: Row](https://www.w3.org/TR/wai-aria-1.2/#row)

straker:
This is similar to #2794

Issue #2794, opened 2021-02-09
grid vs. treegrid allowed row attributes


smhigley:
Expectation: `aria-posinset`, `aria-setsize`, `aria-expanded`, and `aria-level` should all be allowed on `role="row"`, but only when descending from a `treegrid`.

Actual: `aria-posinset` and `aria-setsize` always cause an allowed attribute error on `row`, while `aria-expanded` and `aria-level` never do.

You can see the false positives on the aria-practices example on this page: https://w3c.github.io/aria-practices/examples/treegrid/treegrid-1.html. If you change the parent role to `grid` then run the test again, you can see that `aria-expanded` and `aria-level` are not flagged.

Motivation:
Valid treegrid attributes are getting errors flagged, resulting in a stream of bad bugs, and some easily-caught errors like `aria-expanded` or `aria-level` on grids and tables are not flagged.

<pre><code>
axe-core version: 4.1.1

straker:
I'm not going to lie, I'm having a lot of difficulty navigating that treegrid. I'm trying to test screen reader support but any type of navigation I'm trying isn't consistent and makes it hard to use. For example, in IE11/JAWS I tab to the start of the tree grid, which reads the first row, tells me it's expanded, and that it's 1, 1 of 1 (which I assume are the `aira-postinset` and `aria-setsize` being read out). However, when I press down arrow it goes to the next row (expected), but neither size nor number in list is read out. Pressing down to move to the next row does not do that an instead it moves screen reader focus to the next column, and then I'm broke.

smhigley:
JAWS and other screen readers will read level, expand/collapse, row, and column information conditionally based on the type of interaction you're doing. I'd also recommend using Chrome over IE11 for treegrids :).

Does the specific screen reader UX matter for this issue? The `posinset` and `setsize` attributes are necessary on a treegrid that doesn't render all its rows, and I can't think of a reason not to warn about the use of those attributes + expanded/level on a grid or table row.

straker:
Just looking at what each screen reader does with the attributes is all.

smhigley:
@straker I just wanted to follow up -- is there any other info you'd need for this issue? Do you have a sense of when it might be fixed, or if an external PR would be accepted if one were made? I'm very aware that treegrid support in general is all over the place, but I can't think of a reason to allow a treegrid role but flag specific valid attributes on them.

straker:
[...]

PR #3160, opened 2021-09-15
fix(aria-allowed-attr): check for invalid `aria-attributes` for `role="row"` 

Zidious:
`aria-posinset`, `aria-setsize`, `aria-expanded`, and `aria-level` must not be applied to a row that descends from a table or grid

closes issues: https://github.com/dequelabs/axe-core/issues/3151 & https://github.com/dequelabs/axe-core/issues/2794


lib/checks/aria/aria-allowed-attr-evaluate.js
```diff
@@ -31,13 +32,42 @@ function ariaAllowedAttrEvaluate(node, options, virtualNode) {
...
 
+  // TODO: look into memoizing getRole and removing this local cache
+  let tableMap = cache.get('aria-allowed-attr-table');
+  if (!tableMap) {
+    tableMap = new WeakMap();
+    cache.set('aria-allowed-attr-table', tableMap);
+  }
+
+  // check if the parent exists otherwise a TypeError will occur (virtual-nodes specifically)
+  if (virtualNode.parent) {
```
straker, line 49: We'll only want to do this parent lookup if we are on a `row` element which has one of the attributes we care about. Easiest way would be to mimic what we've done in https://github.com/dequelabs/axe-core/blob/develop/lib/checks/aria/aria-valid-attr-value-evaluate.js#L36 with our prechecks inside of the attribute name for loop on line 74. Something like:

```suggestion
  if (virtualNode.parent && role === 'row') {
```

```js
function validateRowAttrs() {
  /* code to check parent */
}

const preChecks = {
  'aria-postinset': validateRowAttrs
  'aria-setsize': validateRowAttrs,
  etc.
}

if (allowed) {
    for (let i = 0; i < attrs.length; i++) {
      const attrName = attrs[i];
      if (validateAttr(attrName) && !allowed.includes(attrName) && (preChecks[attrName] ? preChecks[attrName](validValue) : true)) {
```

straker (dismissed):

lib/checks/aria/aria-allowed-attr-evaluate.js
Zidious: Maybe I am not understanding, but can this be done in the same if statement? As we create our own data shape: 
```JS
 this.data({
        messageKey: 'table',
        values: validValue,
        tableRole
      });
```

Zidious: After deep diving into both files, I was able to mimic similar to what you suggested. Please let me know your thoughts on the newly updated solution!

lib/checks/aria/aria-allowed-attr-evaluate.js
```diff
@@ -27,31 +33,80 @@ import { getRole, allowedAttr, validateAttr } from '../../commons/aria';
...
   let allowed = allowedAttr(role);
-
   // @deprecated: allowed attr options to pass more attrs.
   // configure the standards spec instead
   if (Array.isArray(options[role])) {
     allowed = uniqueArray(options[role].concat(allowed));
   }
 
+  let tableRole;
+  function validateRowAttrs(validValue) {
```
WilcoFiers, line 47: I kinda feel like this should be hoisted out of this parent funtion.

lib/checks/aria/aria-allowed-attr-evaluate.js
```diff
@@ -27,31 +33,80 @@ import { getRole, allowedAttr, validateAttr } from '../../commons/aria';
...
+    // check if the parent exists otherwise a TypeError will occur (virtual-nodes specifically)
+    if (virtualNode.parent && role === 'row' && validValue) {
+      // TODO: look into memoizing getRole and removing this local cache
```
WilcoFiers, line 50: Definitely don't want to merge this with a TODO still unaddressed.

lib/checks/aria/aria-allowed-attr-evaluate.js
```diff
@@ -27,31 +33,80 @@ import { getRole, allowedAttr, validateAttr } from '../../commons/aria';
...
+      ) {
+        rowAttrInvalid.push(attrName + '="' + virtualNode.attr(attrName) + '"');
+      }
       if (validateAttr(attrName) && !allowed.includes(attrName)) {
         invalid.push(attrName + '="' + virtualNode.attr(attrName) + '"');
       }
     }
   }
 
+  if (rowAttrInvalid.length) {
```
WilcoFiers, line 98: I don't think we should ignore the other `invalid` array like that. Not sure if that means tweaking this, or adding a check.

lib/checks/aria/aria-allowed-attr-evaluate.js
```diff
@@ -27,31 +33,80 @@ import { getRole, allowedAttr, validateAttr } from '../../commons/aria';
...
+
   if (allowed) {
+    let validValue;
     for (let i = 0; i < attrs.length; i++) {
       const attrName = attrs[i];
+      validValue = validateAttrValue(virtualNode, attrName);
+      if (
+        preChecks[attrName]
+          ? preChecks[attrName](validValue)
+          : false && validateAttr(attrName) && allowed.includes(attrName)
```
WilcoFiers, line 88: Can we clean this up? The logic on this is difficult to follow.

lib/checks/aria/aria-allowed-attr-evaluate.js
```diff
@@ -27,31 +33,80 @@ import { getRole, allowedAttr, validateAttr } from '../../commons/aria';
...
+        tableMap.set(table, tableRole);
+      }
+
+      if (['table', 'grid'].includes(tableRole) && role === 'row') {
+        return true;
+      }
+    }
+  }
+
+  const preChecks = {
```
WilcoFiers, line 73: I think this list should be configurable in some way. Maybe as options on the check?

lib/checks/aria/aria-allowed-attr-evaluate.js
```diff
@@ -27,31 +33,80 @@ import { getRole, allowedAttr, validateAttr } from '../../commons/aria';
...
+      if (!tableMap) {
+        tableMap = new WeakMap();
+        cache.set('aria-allowed-attr-table', tableMap);
+      }
+
+      const table = closest(
+        virtualNode,
+        'table, [role="treegrid"], [role="table"], [role="grid"]'
+      );
+      tableRole = tableMap.get(table);
```
WilcoFiers, line 61: Role computation isn't all that expensive. Does caching it this actually make a difference?

WilcoFiers (changes requested):

lib/checks/aria/aria-allowed-attr-evaluate.js
straker: Since we're potentially doing role computation on the same table element, I figured it would be slow enough on a table with thousands of rows in the same table (like https://web.archive.org/web/20190613132353/https://giveawaylisting.com/). This is also why I wanted the TODO left in there as a note that we should look at memoizing the `getRole` function since we can call it for the same element multiple times, but I need to do a performance audit on it to find out.

lib/checks/aria/aria-allowed-attr-evaluate.js
Zidious: I have done some clean up using the existing `invalid` array and still checks the aria-attributes

lib/checks/aria/aria-allowed-attr-evaluate.js
Zidious: `tableRole` doesn't need to be outside of `validateRowAttrs` after doing some clean up. However, I think we can't hoist that block completely out of the function as we use that variable for setting the cache and `closest()` which we only do if the `virtualNode` has a parent and the role is a row.

lib/checks/aria/aria-allowed-attr-evaluate.js
Zidious: I have added a options check with the latest commit `ariaAttrs` which are configurable

lib/checks/aria/aria-allowed-attr-evaluate.js
```diff
@@ -27,21 +28,56 @@ import { getRole, allowedAttr, validateAttr } from '../../commons/aria';
...
   const attrs = virtualNode.attrNames;
   let allowed = allowedAttr(role);
-
   // @deprecated: allowed attr options to pass more attrs.
   // configure the standards spec instead
   if (Array.isArray(options[role])) {
     allowed = uniqueArray(options[role].concat(allowed));
   }
 
+  // TODO: look into memoizing getRole and removing this local cache
```
WilcoFiers, line 40: Why is there a todo in yer PR?

lib/checks/aria/aria-allowed-attr-evaluate.js
```diff
@@ -27,21 +28,56 @@ import { getRole, allowedAttr, validateAttr } from '../../commons/aria';
...
+  );
+
   if (allowed) {
     for (let i = 0; i < attrs.length; i++) {
       const attrName = attrs[i];
-      if (validateAttr(attrName) && !allowed.includes(attrName)) {
+      if (
+        validateAttr(attrName) &&
+        attrName in preChecks &&
+        preChecks[attrName]()
```
WilcoFiers, line 77: Can probably just do this:

```suggestion
        preChecks[attrName]?.()
```

lib/checks/aria/aria-allowed-attr-evaluate.js
```diff
@@ -27,21 +28,56 @@ import { getRole, allowedAttr, validateAttr } from '../../commons/aria';
...
+      if (['table', 'grid'].includes(tableRole) && role === 'row') {
+        return true;
+      }
+    }
+  }
+  // Allows options to be mapped to object e.g. {'aria-level' : validateRowAttrs}
+  const preChecks = Object.assign(
+    {},
+    ...options.ariaAttr.map(key => ({ [key]: validateRowAttrs }))
+  );
```
WilcoFiers, line 69: I would suggest this:

```suggestion
  const ariaAttr = Array.isArray(options.ariaAttr) ? options.ariaAttr : []
  const preChecks = Object.fromEntries(ariaAttr.map(key => (
    [key, validateRowAttrs]
  ));
```

lib/checks/aria/aria-allowed-attr.json
```diff
@@ -1,6 +1,9 @@
 {
   "id": "aria-allowed-attr",
   "evaluate": "aria-allowed-attr-evaluate",
+  "options": {
```
WilcoFiers, line 4: We need to document our check options in `doc/check-options.md`.

lib/checks/aria/aria-allowed-attr.json
```diff
@@ -1,6 +1,9 @@
...
+    "ariaAttr": ["aria-posinset", "aria-setsize", "aria-expanded", "aria-level"]
```
WilcoFiers, line 5: This needs a better name. How about:

```suggestion
    "invalidRowAttrs": ["aria-posinset", "aria-setsize", "aria-expanded", "aria-level"]
```

WilcoFiers (dismissed):

lib/checks/aria/aria-allowed-attr.json
Zidious: I agree

lib/checks/aria/aria-allowed-attr-evaluate.js
WilcoFiers: I'll leave that for Steve to decide, but if we don't, we should have an issue open for it instead of leaving it as a "todo" in the code.

straker: I'm good with an issue instead of a TODO. I'll add this spot in the code in the issue so we know to remove the cache if we find we should memoize the function.

straker: https://github.com/dequelabs/axe-core/issues/3170

WilcoFiers (approved):

PR #3161, opened 2021-09-22
fix(typescript): allow passing a NodeList to ElementContext

kaelig:
As per the spec, the context parameter can be passed one of the following:

> A `NodeList` such as returned by `document.querySelectorAll`.

Source: https://www.deque.com/axe/core-documentation/api-documentation/#context-parameter


straker (approved):
Thanks for the pr

straker:
Reviewed for security

Issue #2934, opened 2021-05-15
"role=text": no focusable descendants


tujoworker:
I really love that "role=text" made it in to axe-core. So I would like to thank everyone involved at first 🙏

PR: https://github.com/dequelabs/axe-core/pull/2702

**Product:** axe-core

**Expectation:** Allow to have "tabindex=-1" as a descendants of "role=text"

**Actual**: Ensures "role=text" is used on elements with no focusable descendants

**Motivation:** I use "tabindex=-1" to let JavaScript set the focus to a descendant to make a copy to the clipboard. I use "role=text" to let VoiceOver (and others) read the content without destruction.

Simple example (fails):
```html
<p>
  some text
  <span role="text">
    more text
    <span tabindex="-1">JavaScript is able to focus this</span>
  </span>
</p>
```

Second example (fails):
```html
<p>
  some text
  <span role="text">
    more text
    <span tabindex="-1" aria-hidden="true" class="visually-hidden">JavaScript is able to focus this</span>
  </span>
</p>
```

**axe-core version:** 4.2

WilcoFiers:
Interesting use case. Can't say I've ever seen it used that way. We'll have a look. I'm not sure about the case where it is visible. A screen reader user using touch screen can still focus elements with tabindex="-1", that's kind of an edge case though. And it wouldn't be relevant if the element is off screen.

WilcoFiers:
I think we need to change this rule to not consider elements with negative `tabindex` to be focusable. It is technically true, but difficult to explain, and a very unusual edge case to handle.

tujoworker:
This would be wonderful 🙏🏻✨

dan-tripp:
I would like to fix this issue.  It seems to me that there are two ways to fix it:
1) change only the "aria-text" rule to not consider elements with negative tabindex to be focusable, or 
2) change all axe-core code to not consider elements with negative tabindex to be focusable.  

#2 would have more side effects, such as in lib/commons/aria/get-role.js: the hasConflictResolution(vNode) function.  Any guidance, @WilcoFiers ?

WilcoFiers:
@dan-tripp I think the simplest solve would be to test the tabindex in the `focusable-element-evaluate` check, similar to what we do in `focusable-element-evaluate.js`

```js
const isFocusable = virtualNode.isFocusable;
let tabIndex = parseInt(virtualNode.attr('tabindex'), 10);
tabIndex = !isNaN(tabIndex) ? tabIndex : -1

[...]

Issue #3163, opened 2021-09-23
Add custom message for negative tabindex in nested-interactive


WilcoFiers:
See #3159

Even though they seem like they should, things like this just don't work well:

```html
  <div role="button" id="pass7" tabindex="0">
    <button aria-hidden="true" tabindex="-1">hidden</button> text
  </div>
  <div role=checkbox aria-checked=true tabindex=0 aria-label=foo>
    <input type=checkbox aria-hidden=true tabindex=-1> <!-- visually hidden -->
  </div>
```

The button / input inside will still be packed up by various AT, that are trying very hard to ensure they don't miss things. This is by no means obvious, and so it would be good to add a specific message key to the no-focusable-content check that reports that using aria-hidden and negative tabindex is not a reliable way of hiding interactive elements. We should have a second message for role=none|presentation and negative tabindex too, just so we don't try to combine too many things.

See ariaErrormessageEvaluate for an example of how to use `messageKey`.

dan-tripp:
I will try to implement this.  Wish me luck.

straker:
Awesome. Let me know if you have any questions.

DanielTripp:
Should this issue be closed?  I think I fully addressed it with https://github.com/dequelabs/axe-core/pull/3194

WilcoFiers:
Yup, a little behind on my grooming work. Thanks Daniel!

PR #3165, opened 2021-09-24
fix(nested-interactive/aria-text): allow "tabindex=-1" on elements with no role

dan-tripp:
Closes issue: #2934


dan-tripp:
"ci/circleci: test_win — Your tests failed on CircleCI" - I can't tell if that test failed because of my changes, or if it was going to fail anyway.  code owners - let me know if I should do anything.

bastilimbach:
@dan-tripp Thanks for fixing this!
I came across a different case where tabindex="-1" should be ignored: `presentation-role-conflict`

**Fails**:
```javascript
<button>
  Test
  <svg class="icon" tabindex="-1" role="presentation">...</svg>
</button>
```

**Works**:
```javascript
<button>
  Test
  <svg class="icon" role="presentation">...</svg>
</button>
```

**Codepen**: 
https://codepen.io/bastilimbach/full/KKqEVGG

Is this something you are willing to include in this PR? I can also create a separate issue.

dan-tripp:
@bastilimbach I'd be happy to, and that change seems sensible to me.  However, I'm new here, so I think that I had better leave it to the core team to decide what the desired behaviour ought to be.  If it's not too much trouble, could you create a new issue and tag me?  Then I'll try to make a separate PR for your new issue, if it gets some approval from the core team.

straker:
@bastilimbach so with `presentation-role-conflict` that is a true error. The algorithm takes into account any value of tabindex, so testing a similar example all browsers and screen readers ignore the `role=presentation` on the element. 

```html
<ul>
  <li tabindex="-1" role="presentation">World</li>
</ul>
```

Chrome - Treats the li element as a `listitem` role
Firefox - Treats the li element as a `listitem` role
Safari / VO - Announces "group" (instead of no role)

The reason is that because of the tabindex the element can be focused programmatically which means that screen readers must announce a role for the element (it can't be presentational).

straker:
Thanks for taking on this pr. I've been trying to figure out if there were any problems with allowing `tabindex=-1` on the descendants of `role=text`. What I've decided is that we should hold off on doing this. 

The reason for that is that this change directly affects the `nested-interactive` rule, allowing `tabindex=-1` on descendants of interactive elements to pass the rule. [It was pointed out](https://github.com/dequelabs/axe-core/pull/3159#issuecomment-921770698) that doing this would pose problems for screen readers and so we decided to [create new messages to point that out](https://github.com/dequelabs/axe-core/issues/3163). So before we do this we first need to resolve that issue.

If this change only affected the `aria-text` rule I would be fine with it, but since it also affects the `nested-interactive` rule we should hold off.

dan-tripp:
@straker  Many thanks for your comments.  I'll look into that new issue (#3163).

straker:
@dan-tripp I believe this pr is replaced by https://github.com/dequelabs/axe-core/pull/3194, since we don't want tabindex=-1 to pass but now warn of the problem.

straker (changes requested):
Alright. Discussed with @WilcoFiers after thoroughly testing things. Testing both `role="text` and interactive roles (.e.g `role="button"`), we've determined that a tabindex *can be allowed* under certain conditions. Basically if the element with the tabindex is not a widget role, then tabindex on the element is fine. Otherwise we should still flag it as we do now.

So something like this:

```js
// no-focusable-content-evaluate.js
import { getRole, getRoleType } from '../../commons/aria';

function getFocusableDescendants(vNode) {
  if (!vNode.children) {
    if (vNode.props.nodeType === 1) {
      throw new Error('Cannot determine children');
    }

    return [];
  }

  const retVal = [];
  vNode.children.forEach(child => {
    const role = getRole(child);

    if(getRoleType(role) === 'widget' && isFocusable(child)) {
      retVal.push(child);
    } else {
      retVal.push(...getFocusableDescendants(child));
    }
  });
  return retVal;
}
```

dan-tripp:
Right on, I'll try to implement that this week.

dan-tripp:
Sorry I haven't made progress on this.  I'm having a strange git problem and no time to deal with it.  I've updated some tests.  I think I'm close.

WilcoFiers:
@dan-tripp No rush.

dan-tripp:
My latest are ready for review.  Let me know what you think.  And thank you for your patience.

lib/checks/keyboard/no-focusable-content-evaluate.js
```diff
@@ -40,7 +41,8 @@ function getFocusableDescendants(vNode) {
 
   const retVal = [];
   vNode.children.forEach(child => {
-    if (isFocusable(child)) {
+    const role = getRole(child);
+    if(getRoleType(role) === 'widget' && isFocusable(child)) {
```
WilcoFiers, line 45:
```suggestion
    if (getRoleType(role) === 'widget' && isFocusable(child)) {
```

test/checks/keyboard/no-focusable-content.js
```diff
@@ -28,17 +28,22 @@ describe('no-focusable-content tests', function() {
...
     );
 
-    assert.isFalse(noFocusableContent.apply(checkContext, params));
-    assert.deepEqual(checkContext._data, null);
-    assert.deepEqual(checkContext._relatedNodes, [params[2].children[0]]);
+    assert.isTrue(noFocusableContent.apply(checkContext, params));
   });
 
-  it('should return false if element has natively focusable content', function() {
+  it('should return true if element has content which has negative tabindex and non-widget role', function() {
```
WilcoFiers, line 39: Can you add an example that uses span, tabindex and some widget role that shows those failing?

test/checks/keyboard/no-focusable-content.js
```diff
@@ -70,4 +75,33 @@ describe('no-focusable-content tests', function() {
     assert.deepEqual(checkContext._data, { messageKey: 'notHidden' });
     assert.deepEqual(checkContext._relatedNodes, [params[2].children[0]]);
   });
+
+  it('should return true if element has content which is natively focusable and has a widget role but is disabled', function() {
+    var vNode = queryFixture(
+      '<button id="target"><a href="foo.html" disabled>Hello</a></button>'
```
WilcoFiers, line 81: Hmm, that shouldn't have worked. Links don't have a disabled attribute. I opened a separate issue for that (#3315). No need to fix that in this PR. But we should change the example. I suggest:

```suggestion
      '<button id="target"><input value="hello" disabled></button>'
```

WilcoFiers (changes requested):
Close, couple more minor things

dan-tripp:
Okay, I think this PR is ready for review again.  Thank you for all your suggested changes, @WilcoFiers - I think that I'm getting the hang of this.

WilcoFiers:
@dan-tripp Thank you so much for this contribution! Great work, really appreciate it.

dan-tripp:
Great!  Glad I can help.

PR #3168, opened 2021-09-27
fix(is-visible): do not error if window.Node does not exist

straker:
Allows `isVisible` to be run if `window.Node` is not defined.

### DIFF
--- a/test/test-virtual-rules.js
+++ b/test/test-virtual-rules.js
@@ -1,0 +1,23 @@
+var path = require('path');
+var assert = require('chai').assert;
+var glob = require('glob');
+var axe = require(path.join(__dirname, '../axe'));
+
+var files = glob.sync(path.join(__dirname, 'integration/virtual-rules/*.js'));
+
+before(function() {
+  global.axe = axe;
+  global.assert = assert;
+});
+
+after(function() {
+  delete global.axe;
+  delete global.assert;
+});
+
+describe('virtual-rule node tests', function() {
+  files.forEach(function(file) {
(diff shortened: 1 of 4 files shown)


test/test-virtual-rules.js
```diff
@@ -0,0 +1,23 @@
+var path = require('path');
+var assert = require('chai').assert;
+var glob = require('glob');
+var axe = require(path.join(__dirname, '../axe'));
```
stephenmathieson, line 4: Isn't this the same thing?

```suggestion
var axe = require('..');
```

straker: `axe` isn't a folder so it's the `axe.js` file we're after

WilcoFiers, line 4:
```suggestion
var axe = require(../axe');
```

WilcoFiers (changes requested):

test/test-virtual-rules.js
stephenmathieson: [`axe.js` is set to `main` in `package.json`](https://github.com/dequelabs/axe-core/blob/develop/package.json#L51).  You don't need anything other than `..`. I'm 👍 for removing the `__dirname` though.

straker: Ah. I didn't know that. I knew it looked for an `index` file when using a directory, but I didn't know a relative file path would look at `pacakge.json` as well (I only thought that was for module names).

WilcoFiers (approved):

Issue #3174, opened 2021-10-01
Possible false positive for label-title-only, caused by aria-describedby


fymmot:
### Product
axe-core

### Description: 

I think I may have found a false positive bug for the label-title-only rule when using a combination of `aria-labelledby`, `aria-describedby` and `aria-hidden` on the labelling element. 

Reproduced and explained here: https://codepen.io/fymmot/full/XWgQJLw

(My purpose for using aria-labelledby and aria-hidden in this way is to avoid duplicate announcements of the label texts for screen reader users. In our particular use case, we are using multiple elements to construct the label, which causes multiple swipes or down arrow interactions to move past it for the user.)

### Expectation: 

I expect that I can label an `<input type="radio">` using `aria-labelledby` and having `aria-hidden` set on the labelling element.

I also expect that I can use `aria-describedby` on the same `<input>`, without it triggering a `label-title-only` error in aXe.

### Actual: 

When using a combination of `aria-labelledby`, `aria-hidden` and `aria-describedby`, aXe throws a `label-title-only` error.

I also get the same error if using `hidden` instead of `aria-hidden`.

I can remove the error by removing the `aria-describedby` attribute but keeping `aria-labelledby`. 

I can also fix the error by removing `aria-hidden` or `hidden` from the labelling element. 

### Motivation: 

This behavior does not seem logical, or intended. I don't think the presence of aria-describedby should affect the label-title-only test, as long as the aria-labelledby is valid.

If the combination of aria-labelledby and aria-hidden is discouraged or disallowed, it should throw an error regardless of aria-describedby, which it does not

<pre><code>
axe-core version: 4.3.3
axe devtools extension (MacOS Chrome) version: 4.17.1
</code></pre>

straker:
Thanks for the issue. The reason this happens is that the check looks to see if the input does not have a label but uses `title` or `aria-describedby`. That's why the first case get's flagged (as the label is hidden) but the rest dont.

However the label is read by screen readers even if it is hidden. That's why the `label` rule isn't reporting a violation for the inputs. Confirming with @WilcoFiers to see if we need to check the visible state of the label for this rule.

straker:
We determined that this is indeed a false positive.

macjohnny:
a fix is implemented in https://github.com/dequelabs/axe-core/pull/3183

nolandanley:
[...]

PR #3183, opened 2021-10-05
fix(label-title-only): allow hidden labels

macjohnny:
take into account labels with `aria-hidden="true"` when evaluating the `label-title-only` rule.

Closes issue:
#3174


straker:
Alright, talked to Wilco and we decided that `aria-hidden` labels should pass, but visually hidden labels should not. That means we can just remove the `true` argument in the `visualVirtual` call in [labelVirtual](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/aria/label-virtual.js#L24).

```js
return vNode ? visibleVirtual(vNode) : '';
```

In conjunction with that change, we should update the description to the [label-title-only rule](https://github.com/dequelabs/axe-core/blob/develop/lib/rules/label-title-only.json#L7) to state what the rule is really looking for: that the element has a visual label.

```json
    "description": "Ensures that every form element has a visible label and is not solely labeled using hidden labels, or the title or aria-describedby attributes",
```

macjohnny:
@straker updated the PR according to your suggestions, please take a look :)

PR #3184, opened 2021-10-05
docs: Update rule-proposal to add link to tag assignment

dylanb:
Link to clarification of how tags are assigned.

### DIFF
--- a/doc/rule-proposal.md
+++ b/doc/rule-proposal.md
@@ -28,7 +28,7 @@ Example: "Elements must only use allowed ARIA attributes"
 
 #### Tags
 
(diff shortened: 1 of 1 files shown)


doc/rule-proposal.md
```diff
@@ -28,7 +28,7 @@ Example: "Elements must only use allowed ARIA attributes"
 
 #### Tags
 
-Indicate which tags the rule should use.
+Indicate which tags the rule should use. See the documentation on [how we assign tags](https://github.com/dequelabs/axe-core/blob/develop/doc/API.md#axe-core-tags).
```
stephenmathieson, line 31:
```suggestion
Indicate which tags the rule should use. More information available in the [how we assign tags documentation](https://github.com/dequelabs/axe-core/blob/develop/doc/API.md#axe-core-tags).
```

Issue #3166, opened 2021-09-24
TypeError: Cannot read properties of undefined (reading 'filter')


patrickhulce:
Product: axe-core

Expectation: Complete color-contrast audit without erroring

Actual: Fails with an exception

```
TypeError: Cannot read properties of undefined (reading 'filter')
    at Aa (https://cdnjs.cloudflare.com/ajax/libs/axe-core/4.3.3/axe.min.js:12:111917)
    at Ca (https://cdnjs.cloudflare.com/ajax/libs/axe-core/4.3.3/axe.min.js:12:112362)
    at _s (https://cdnjs.cloudflare.com/ajax/libs/axe-core/4.3.3/axe.min.js:12:284518)
    at Os (https://cdnjs.cloudflare.com/ajax/libs/axe-core/4.3.3/axe.min.js:12:284648)
    at Ps (https://cdnjs.cloudflare.com/ajax/libs/axe-core/4.3.3/axe.min.js:12:285349)
    at Ls (https://cdnjs.cloudflare.com/ajax/libs/axe-core/4.3.3/axe.min.js:12:287359)
    at Object.color-contrast-evaluate (https://cdnjs.cloudflare.com/ajax/libs/axe-core/4.3.3/axe.min.js:12:299707)
    at mu.run (https://cdnjs.cloudflare.com/ajax/libs/axe-core/4.3.3/axe.min.js:12:323457)
    at https://cdnjs.cloudflare.com/ajax/libs/axe-core/4.3.3/axe.min.js:12:326089
    at c (https://cdnjs.cloudflare.com/ajax/libs/axe-core/4.3.3/axe.min.js:12:97607)
```

which resolves to these lines

https://github.com/dequelabs/axe-core/blob/1ee88cb4bb557560f10eab136464c321d4dee81e/lib/commons/dom/get-rect-stack.js#L478-L479

<details>
<summary>Reproduction Details</summary>

https://travisdowns.github.io/misc/sorting.html

```
<html lang="en"><head>
    <meta http-equiv="X-UA-Compatible" content="IE=edge">
    <title>Lighthous Accessibility Failure</title>

    <!-- <link rel="stylesheet" href="/assets/css/light.css"> -->
</head>
  <body>

    <style>
      pre {
        overflow-x: auto;
      }
    </style>

<pre><span>
[...]

PR #3186, opened 2021-10-05
fix(color-contrast): account for elements that do not fill entire bounding size

straker:
Closes issue: #3166

### DIFF
--- a/lib/commons/dom/get-rect-stack.js
+++ b/lib/commons/dom/get-rect-stack.js
@@ -360,6 +360,8 @@ function addNodeToGrid(grid, vNode) {
     const endRow = ((y + rect.height) / gridSize) | 0;
     const endCol = ((x + rect.width) / gridSize) | 0;
 
+    grid._numCols = Math.max(grid._numCols ?? 0, endCol);
+
     for (let row = startRow; row <= endRow; row++) {
       grid.cells[row] = grid.cells[row] || [];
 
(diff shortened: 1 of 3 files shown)


lib/commons/dom/get-rect-stack.js
```diff
@@ -360,6 +360,8 @@ function addNodeToGrid(grid, vNode) {
     const endRow = ((y + rect.height) / gridSize) | 0;
     const endCol = ((x + rect.width) / gridSize) | 0;
 
+    grid._numCols = Math.max(grid._numCols ?? 0, endCol);
```
WilcoFiers, line 363: This feels a little off. If `_numCols` needs to be different, it should be fixed where that value is set. If it doesn't need to be changed, this should probably just be a local variable.

lib/commons/dom/get-rect-stack.js
```diff
@@ -475,21 +477,34 @@ export function getRectStack(grid, rect, recursed = false) {
...
-        y >= rectY
-      );
-    });
-  });
+
+  // we're making an assumption that there cannot be an element in the
+  // grid which escapes the grid bounds. For example, if the grid is 4x4 there
+  // can't be an element whose midpoint is at column 5. If this happens this
+  // means there's an error in our grid logic that needs to be fixed
+  if (row > grid.cells.length || col > grid._numCols) {
```
WilcoFiers, line 485: I don't think we should be checking private variables here. They're private for a reason.

lib/commons/dom/get-rect-stack.js
```diff
@@ -475,21 +477,34 @@ export function getRectStack(grid, rect, recursed = false) {
...
+    throw new Error('Element midpoint exceeds the grid bounds');
```
WilcoFiers, line 486: I don't follow why you would want this to throw. Crashing axe-core doesn't seem like the right solution here.

WilcoFiers (changes requested):

lib/commons/dom/get-rect-stack.js
straker: Because I want to know when our custom grid logic fails, and it shows there's an issue that needs to be fixed. Without throwing, color contrast would hide issues that should be fixed.

Issue #3190, opened 2021-10-06
AAA rules should not be mixed with other result types


WilcoFiers:
Axe-core has one AAA rule that is also a level A rule, and one rule that's both AAA and a best-practice. Our API.md doc says that this shouldn't be the case, and the fact that we did has caused some confusion.

PR #3191, opened 2021-10-06
fix: Separate Level AAA rules from A and best-practices

WilcoFiers:
Per Dylan's suggestion, and according to our own API docs the AAA tag should never have been used on rules tagged at a different level, or as best practice. This PR puts ensures AAA rules are now treated as its own "type" or rule.

Closes issue: #3190


doc/API.md
```diff
@@ -80,8 +80,10 @@ The `experimental`, `ACT` and `section508` tags are only added to some rules. Ea
 | ---------------- | ---------------------------------------------------- |
 | `wcag2a`         | WCAG 2.0 Level A                                     |
 | `wcag2aa`        | WCAG 2.0 Level AA                                    |
+| `wcag2aaa`       | WCAG 2.0 Level AAA                                   |
 | `wcag21a`        | WCAG 2.1 Level A                                     |
 | `wcag21aa`       | WCAG 2.1 Level AA                                    |
+| `wcag21aaa`      | WCAG 2.1 Level AAA                                   |
```
straker, line 86: We don't have any wcag21aaa rules. Probably shouldn't list it as a valid tag until we have at least one as trying to use it will fail:

```js
await axe.run({ runOnly: ['wcag21aaa']}) //Could not find tags `wcag21aaa`
```

straker (changes requested):

Issue #3177, opened 2021-10-01
ARIA in HTML allowances changes for br and wbr


scottaohara:
Product: axe-core

Expectation: `wbr` and `br` elements only allow the `none` and `presentation` roles. And only allow the `aria-hidden` attribute.

Actual: `br` presently allows any global `aria-` attribute.  `wbr` presently allows for any role and global `aria-*` attribute, or `aria-*` attribute applicable to the assigned explicit role.

role test page for [`wbr`](https://thepaciellogroup.github.io/aria-test-builder/dist/testcases/wbr-test.html)

Motivation: ARIA in HTML will be changing the allowances for `br` and `wbr` elements.  These elements will only allow the `none` and `presentation` roles, and the `aria-hidden` attribute.  Other roles and global `aria-` attributes do not make sense to be used on these elements, particularly due to their inability to be effectively modified by authors.

<pre><code>axe-core version: 4.3.3
chromium 94</code></pre>

Edit: indicating tasks requested in this issue
- [x] Done: update role allowances for `br` and `wbr`
- [ ] TODO: ensure that `aria-hidden` is the only allowed `aria-*` attribute for `br` and `wbr` elements

straker:
Thanks for letting us know. Should be an easy fix by updating the [html-elms file](https://github.com/dequelabs/axe-core/blob/develop/lib/standards/html-elms.js#L939) and changing the `allowedRoles` to be `['none', 'presentation']`.

As for the `aria-hidden` part, we'll have to modify the [aria-allowed-attr-evaluate script](https://github.com/dequelabs/axe-core/blob/develop/lib/checks/aria/aria-allowed-attr-evaluate.js#L70) to prevent `aria-hidden` specifically on these elements since it's a global attribute. But I'd like @WilcoFiers opinion before we do that first.

scottaohara:
Thank you @straker.  I'd be happy to make the PR for the updated html-elms file.  Appreciate you pointing me there.

Re: your second line - i think you have a mistype or my comment was misunderstood:
>we'll have to modify the aria-allowed-attr-evaluate script to prevent aria-hidden specifically on these elements

we do not want to prevent `aria-hidden` on these elements, but rather we want to allow `aria-hidden` and prevent other global attributes since they don't make sense on these elements and since they'd be in opposition to the allowed roles.

PR #3192, opened 2021-10-06
fix(aria-allowed-role): updates the allowed roles for the wbr element to none and presentation

scottaohara:
Changes the allowed roles of the `wbr` element from "all roles", to `none` and `presentation` only.

related to issue #3177 (but does not entirely close it)

### DIFF
--- a/lib/standards/html-elms.js
+++ b/lib/standards/html-elms.js
@@ -936,7 +936,7 @@ const htmlElms = {
   },
   wbr: {
     contentTypes: ['phrasing', 'flow'],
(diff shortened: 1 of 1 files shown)


straker:
Thanks for the pr. Should we also update the `b` element to restrict the roles as well in this pr?

scottaohara:
Hi @straker. Nope this is good to go. We are marching the br element here, not the b element. So aside from the aria-hidden we spoke of in the issue related to this pr, this update is all that’s needed for updating the allowed roles.

Issue #3143, opened 2021-09-01
`aria-allowed-role` rule false positive for `<img>` with no alt but with accessible name


kospl:
Product: axe-core

Expectation: `<img>` with role `button` and accessible name is a valid element.

Actual: axe core 4.3.3 reports it as a violation of `aria-allowed-role` rule.

Motivation: as described in https://www.w3.org/TR/html-aria/#el-img-no-alt `<img>` with no alt but WITH accessible name may have same roles as https://www.w3.org/TR/html-aria/#el-img including `button`.

<pre><code>
axe-core version: 4.3.3

Browser and Assistive Technology versions: Chrome/92.0.4515.159

For Tooling issues:
- Platform:  Windows
</code></pre>

Repro minimal example:

    <img role="button" aria-label="Search">
<br>

    axe.run({
      runOnly: ['aria-allowed-role']
    }, function (err, results) {
      console.log(results.violations);
    });

Result (axe-core 4.3.3):

![image](https://user-images.githubusercontent.com/26386660/131654911-f0f0be1f-44b1-4532-b379-890b39f6f067.png)

straker:
So an image without an `alt` attribute is considered as `role=presentation` and is not allowed to have any role. See https://www.w3.org/TR/html-aria/#el-img-empty-alt

kospl:
Related code block for `noAlt` [removed about year ago](https://github.com/dequelabs/axe-core/commit/292866526fcddecc8236e2a9a8d796f3f80932b4#diff-71c6e1ff51ae19a73e38e88147c50e670e87b1293961bff4b9f6a21fe3b56609), not sure why, but I think it has something to do with this false positive.

padmavemulapati:
Validated with the latest axe-core dev branch code base, 
Test snippet:
```
<img role="button" aria-label="Search">
  <img role="button" alt="foo" /> 
<img role="button" aria-label="foo" /> 
```
![image](https://user-images.githubusercontent.com/47409579/146204428-6c60b524-67d6-42ef-bec6-078d3700124b.png)

Not finding any failures.

kospl:
Yeah, some work have been done, but still fix is absent in all axe tools other than dev, that's why this issue is still open.

WilcoFiers:
Nope, sorry. Our policy is to close the issue after QA has confirmed the issue is resolved in the develop branch. @kospl, this fix will be in the next released, which its currently looking will be the 4.4 release on January 31st.

PR #3224, opened 2021-10-19
fix(aria-allowed-role): allow title, aria-label and aria-labelledby on a img element with a supported role

Zidious:
Currently when an `img` element has a `title`, `aria-label` or `aria-labelledby` and a supported role i.e. button we report it as a violation, this appears to be a false positive as we do not report it as a violation when an `alt` tag is added.

```html
<img role="button" alt="foo" /> - pass
<img role="button" aria-label="foo" /> - fail

<div id="bar">hello world</div>
<img role="button" aria-labelledby="bar" /> - fail
```

we have added a new matcher to match an element that has a non-empty accessible name. By doing so gives us the ability to add a matches property on a new variant `noAlt` that will check if the `img` element text is accessible (in this case, title, aria-label and, aria-labelledby) if an `alt` tag is not present.

W3C resources provided by the by the author in https://github.com/dequelabs/axe-core/issues/3143: 
https://www.w3.org/TR/html-aria/#el-img-no-alt
https://www.w3.org/TR/html-aam-1.0/#img-element-accessible-name-computation
https://www.w3.org/TR/html-aria/#el-img

Closes issue: https://github.com/dequelabs/axe-core/issues/3143


WilcoFiers (dismissed):

lib/standards/html-elms.js
```diff
@@ -329,6 +329,31 @@ const htmlElms = {
   },
   img: {
     variant: {
+      noAlt: {
+        matches: {
+          hasAccessibleName: true,
+          attributes: {
+            alt: null
+          }
+        },
```
WilcoFiers, line 338: I think this should let us get rid of the second variant.

```suggestion
        matches: [{
          // Because <img role="none" alt="foo" /> has no accessible name:
          attributes: {
            alt: '/.+/'
          }
        }, {
          hasAccessibleName: true,
        }],
```

WilcoFiers (dismissed):

WilcoFiers (changes requested):

WilcoFiers (approved):

Issue #3187, opened 2021-10-05
Color-contrast does not take into account mix-blend-mode


straker:
The [mix-blend-mode CSS property](https://developer.mozilla.org/en-US/docs/Web/CSS/blend-mode) specifies how colors are blended together. By default the `mix-blend-mode` is set to `normal`, which is mostly how our `flattenColor` algorithm works. But by changing the mix-blend-mode the background color can be completely different.

For example, taking this HTML, the output is faint green color.

```html
<div style="background-color: rgba(255, 255, 255, 1.0);">
  <div style="background-color: rgba(0, 128, 0, 0.25);">Hello World</div>
</div>
```

![image](https://user-images.githubusercontent.com/2433219/136101592-f586b612-881b-469f-9de9-33fd10eb03a7.png)

But if we change the blend mode we get a completely different color, in this case a faint pink color.

```html
<div style="background-color: rgba(255, 255, 255, 1.0);">
  <div style="background-color: rgba(0, 128, 0, 0.25);mix-blend-mode: exclusion;">Hello World</div>
</div>
```

![image](https://user-images.githubusercontent.com/2433219/136101850-1682d32f-7385-48e3-ada6-e3614b32e921.png)

Zidious:
Note: mix blend mode also needs to take into account text colour when blending.

PR #3226, opened 2021-10-20
feat(color-contrast): add support for CSS mix-blend-mode

Zidious:
`mix-blend-mode` functions:

- [x] unit tests
- [x] integration tests

This PR focuses on blending BG colours. An example of mix-blend-mode in action, consider the following: 

```html
<div id="A" style="background-color: rgba(255, 255, 255, 1.0);">
  <div id="B" style="background-color: rgba(0, 128, 0, 0.25);">
    <div id="C" style="background-color: rgba(255, 0, 0, 0.5); mix-blend-mode: difference;">
      <div id="D" style="background-color: rgba(0, 0, 255, 0.25);">Test Element</div>
    </div>
  </div>
</div>
```

When mix-blend-mode is applied onto a div it is considered an [isolated group](https://drafts.fxtf.org/compositing-1/#isolatedgroups) and therefore its own [stacking context](https://drafts.fxtf.org/compositing-1/#csscompositingrules_CSS). 

The stack of the above snippet would look like the following: 

```
Root Group
  - #A
  - #B
  - isolated group
    - #C 
    - #D
```

When blending we do bottom up: 

```
rootGroup = normal(B,normal(A, rootBackDrop))
isolatedGroup = normal(D, C)
difference(rootGroup, isolatedGroup) // outputs the final colour
```

For reference: [compositing BG colours w3c assistance](https://github.com/w3c/fxtf-drafts/issues/440)

Resources: [Compositing and Blending](https://drafts.fxtf.org/compositing-1/#blending) - Note: further research I would like to investigate how to go about blending nested mix-blend-modes when in the same div block.

Closes issue: https://github.com/dequelabs/axe-core/issues/3187


test/commons/color/flatten-colors.js
```diff
@@ -38,9 +38,251 @@ describe('color.flattenColors', function() {
     assert.equal(flat5.alpha, 1);
 
     var flat6 = axe.commons.color.flattenColors(quarterLightGreen, halfRed);
-    assert.equal(flat6.red, 96);
```
dylanb, line 41: why did this change?

test/integration/full/contrast/blending.js
```diff
@@ -1,18 +1,173 @@
...
   var resultElms = [];
   var expected = [
+    // normal
     'rgb(223, 112, 96)',
     'rgb(255, 128, 128)',
     'rgb(191, 223, 191)',
     'rgb(125, 38, 54)',
     'rgb(179, 38, 0)',
     'rgb(179, 0, 77)',
-    'rgb(143, 192, 80)',
```
dylanb, line 11: why this change? Seems like a rounding difference of some sort.

dylanb (commented):
two questions, see comments

test/commons/color/flatten-colors.js
Zidious: The way were were actually blending the colours was not quite accurate. When you start adding alpha values to colours the way they are calculated is different. Opacity adds another step whereby the pre-multiplied computed new RGB values need to be divided by the new computed alpha, we were missing this step: 

> To extract the color component of a pre-multiplied value, the formula is reversed: Co = co / αo
source: https://www.w3.org/TR/compositing-1/#simplealphacompositing
helpful source: https://github.com/w3c/fxtf-drafts/issues/440#issuecomment-956418953

test/integration/full/contrast/blending.js
Zidious: I believe this may of impacted by https://github.com/dequelabs/axe-core/pull/3226#discussion_r752386049 and the threshold of rounding was impacted

dylanb (approved):

PR #3249, opened 2021-11-02
fix(respondable): avoid crashes if the data in `window.postMessage` is `null`

esanzgar:
axe-core crashes if another process sends a `window.postMessage` with
data equals to `null`.

In the previous version, because `null` has a type of `'object'`, it
crashed when trying to access the `channelId`.

This change detects first if the data is not `null`.

### DIFF
--- a/lib/core/utils/frame-messenger/message-handler.js
+++ b/lib/core/utils/frame-messenger/message-handler.js
@@ -23,29 +23,34 @@ export function messageHandler(
   { origin, data: dataString, source: win },
   topicHandler
 ) {
-  const data = parseMessage(dataString) || {};
-  const { channelId, message, messageId } = data;
-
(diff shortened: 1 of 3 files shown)


straker:
Thanks for the pr. Could you add a test for this case in the respondable test? Probably [after the repeater test](https://github.com/dequelabs/axe-core/blob/develop/test/core/utils/respondable.js#L531)

straker (dismissed):
Thanks for the pr. Could you add a test for this case in the respondable test? Probably [after the repeater test](https://github.com/dequelabs/axe-core/blob/develop/test/core/utils/respondable.js#L531)

esanzgar:
I added the test. I am not very familiar with the internals with axe-core, so I am afraid you will need to take it from there.

WilcoFiers (approved):

straker:
Reviewed for security

PR #3254, opened 2021-11-03
remove dpub role allowances for li element

### LINKED ISSUE #3253: update dpub role allowances for li element
[ARIA in HTML](https://github.com/w3c/html-aria/pull/369) is removing the allowances for [`doc-biblioentry`](https://w3c.github.io/dpub-aria/#doc-biblioentry) and [`doc-endnote`](https://w3c.github.io/dpub-aria/#doc-endnote) on `<li>` elements.  

These are not technically valid children of `ul`, `ol` or `role=list` elements per the ARIA spec.  Due to this, these roles are being deprecated in DPUB ARIA 1.1. Authors should be using standard list elements, or `role=list` and `role=listitem` elements instead.

Edited issue description to provide more clarity / per the closure of the original PR.

scottaohara:
Updates `<li>` element to no longer allow the dpub roles [`doc-biblioentry`](https://w3c.github.io/dpub-aria/#doc-biblioentry) and [`doc-endnote`](https://w3c.github.io/dpub-aria/#doc-endnote) which are being deprecated in DPUB ARIA 1.1 due to better aligning with ARIA's clarified child element (role) allowances of a `list`.

Closes #3253

### DIFF
--- a/lib/standards/html-elms.js
+++ b/lib/standards/html-elms.js
@@ -547,9 +547,7 @@ const htmlElms = {
       'radio',
       'separator',
       'tab',
-      'treeitem',
-      'doc-biblioentry',
-      'doc-endnote'
+      'treeitem'
     ],
     implicitAttrs: {
       'aria-setsize': '1',


straker:
Thanks for the pr. I'm not sure we can take this yet as it would start to potentially break users where it didn't before. Since they are only being deprecated, they'll probably still need to be allowed, but maybe we add some sort of deprecation warning for the user?

 I'll consult with @WilcoFiers and we'll figure out what to do.

scottaohara:
Thanks @straker, and yes that makes sense.

straker:
Do you happen to know of any cases where having `doc-biblioentry` or `doc-endnote` cause problems if they are used? If not we probably will continue to leave them as allowed even if they are deprecated.

scottaohara:
good question.  theoretically i do not think there will be issues, but I will verify.

What I do readily know is that if used in a web context, the role is ignored and the general list item semantics are continued to be exposed.  So, to the point of your last comment, no harm no foul.

But that raises a question I have about why axe wouldn't mention anything (and apologies if that's not what you're saying, but that's how i'm interpreting 'leave them as allowed') about a role that is deprecated, per it not having a user impact.  For contrast, `<p role=nonsense>...</p>` is flagged as a critical accessibility error by axe-core.   

But in reality, the paragraph is still properly exposed as a paragraph, because browsers ignore the invalid role. 

Being an invalid role, ARIA states:
>If the role attribute contains no tokens matching the name of a non-abstract WAI-ARIA role, the user agent MUST treat the element as if no role had been provided.

My point here being that i think it'd still be good to indicate these roles as being deprecated, even if not called out as a 'failure', especially in regards to critical flags being raised for invalid roles which also don't, at least in [quickly testing with JAWS, NVDA and Narrator](https://codepen.io/scottohara/pen/gOxvwJN) don't seem to have user impact either.   *Note:* I'm not actually requesting this flag be removed by axe at all.  I just am pointing out that it may not, or may not always belong in the 'critical' category.

straker:
The questions was just to verify that there wasn't any problems with screen readers if the roles were left as allowed. As for the messages, we currently don't have any system in place to warn users of deprecated things, especially ARIA standards. So we'll need to think about how we want to handle this.

scottaohara:
thank you @straker

straker:
I'm going to close the pr since we won't be removing them and repost the discussion about deprecation messages in the original issue

PR #3265, opened 2021-11-08
fix(aria-allowed-role): remove role allowances for embed, iframe and object

### LINKED ISSUE #3264: Remove ARIA role allowances for iframe, object, embed
ARIA in HTML is updating the `iframe`, `object` and `embed` elements to no longer allow any explicit ARIA roles, as specifying these roles have been found to cause unexpected issue with AT/browsers - https://github.com/w3c/html-aria/pull/370

scottaohara:
closes #3264

remove role allowances for `iframe`, `embed` and `object` per https://github.com/w3c/html-aria/pull/370

### DIFF
--- a/lib/standards/html-elms.js
+++ b/lib/standards/html-elms.js
@@ -219,7 +219,7 @@ const htmlElms = {
   },
   embed: {
     contentTypes: ['interactive', 'embedded', 'phrasing', 'flow'],
-    allowedRoles: ['application', 'document', 'img', 'presentation', 'none'],
+    allowedRoles: false,
     chromiumRole: 'EmbeddedObject'
   },
   fieldset: {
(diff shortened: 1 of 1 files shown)


straker:
Thanks for the pr. Because this is removing roles we'll probably need to make this at least a `feat`, but it also might be a breaking change? @WilcoFiers will need to decide how we want to pull this in.

scottaohara:
thanks @straker.  In the future, if you would rather I just create issues and not PRs, I'm happy to do that. The intent is to help with the work, not create extra. So if I'm doing that, please let me know and I'll stop :)

WilcoFiers (changes requested):
I left a review on the ARIA in HTML PR explaining why I disagree with these changes. Lets discuss there.

straker:
@scottaohara I think any changes in ARIA that add new roles are good for prs. Changes that remove roles should probably be issues since we'll want to make sure we don't break users before we're ready.

WilcoFiers:
I'm going to close this, since the proposal is getting reworked. We can reopen the PR, or create a new one once there's an updated proposal. Again, these PRs are really appreciated Scott!

Issue #3272, opened 2021-11-10
Allow axe-core to skip frame pings with pingWaitTime: 0


WilcoFiers:
Before telling a frames to test, axe-core sends a ping, to check if the frame has a compatible version of axe-core in it. This is to avoid a scenario where it waits a whole minute for a frame to send back results, only to find that that didn't have axe-core in it at all.

In scenarios where we can guarantee this unnecessary waiting doesn't happen, such as browser extensions, we don't need to send this ping message. A simple solve for this could be to set allow `pingWaitTime: 0`, to just skip the ping step, and go right to the run command.

PR #3273, opened 2021-11-11
feat(options): make axe.ping configurable with pingWaitTime

WilcoFiers:
- Document the existence of pingWaitTime
- Have `pingWaitTime: 0` skip the ping step altogether

Closes issue: #3272

### DIFF
--- a/axe.d.ts
+++ b/axe.d.ts
@@ -95,6 +95,7 @@ declare namespace axe {
     frameWaitTime?: number;
     preload?: boolean;
     performanceTimer?: boolean;
+    pingWaitTime?: number;
   }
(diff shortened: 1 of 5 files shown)


straker (approved):

scurker:
Just so I follow:

Setting `pingWaitTime` to `0` means that only an `axe.run` is sent? Is this making the assumption that we're re-implementing a ping in the axe extension to first validate that we can communicate with a frame?

My concern lies that if only a single call is made, we don't want to wait `60000ms` to timeout if the frame is unable to communicate. We can implement our own ping method in the extension that encompasses both, but just FYI.

Issue #3266, opened 2021-11-08
Request updates to allowed roles for area without href


scottaohara:
[ARIA in HTML](https://github.com/w3c/html-aria/pull/360) is looking to update the allowed roles for `area` without `href` from **no role** to instead allow `link` and `button`, only.

WilcoFiers:
That makes a lot of sense. Any particular reason you wouldn't also allow button on `area[href]` though?

scottaohara:
thanks Wilco.

That idea did come up when discussing this change (both in the ARIA WG when they had done the initial review during CR, and when we were discussing after I looked into what other roles might have made sense here).

Essentially, keeping the status quo boils back down to a version of the links vs buttons debate / user expectations of role announcement.  If authors need an `area` to be a `role=button`, then they would now have this opportunity by not adding, or by removing, the `href`.  

If need be, we can discuss this further in a new issue in ARIA in HTML.  

Related issue: https://github.com/w3c/html-aria/issues/187

WilcoFiers:
Tanks for clarifying. Sounds good.

padmavemulapati:
Validated with the develop branch axe-core code base,
`aria-allowed-role` rule allowing the roles for `button` and `'link` `area` without `href`

![image](https://user-images.githubusercontent.com/47409579/141989802-f2c6a355-dbfa-4541-96c4-b454cd12cfda.png)

PR #3275, opened 2021-11-12
fix(allowed-role): area without href can have a button or link role

WilcoFiers:
As per a recent update in ARIA in HTML

Closes issue: #3266


test/commons/aria/is-aria-role-allowed-on-element.js
```diff
@@ -223,12 +223,4 @@ describe('aria.isAriaRoleAllowedOnElement', function() {
     var actual = axe.commons.aria.isAriaRoleAllowedOnElement(node, 'link');
     assert.isTrue(actual);
   });
-
-  it('returns false if elements implicit role does not match the role', function() {
```
WilcoFiers, line 227: I don't think this test adds any value. It doesn't return false for just this reason, it returns false because none of the things we test for that could make it true are. So I decided to remove it rather than put in some other element.

straker (approved):

Issue #2759, opened 2021-01-18
Deprecation of `doc-biblioentry` and `doc-endnotes` roles


Seirdy:
DPUB-ARIA 1.1 [deprecates](https://github.com/w3c/dpub-aria/pull/22) the `doc-biblioentry` and `doc-endnote` roles. Using `<li>` or list/listitem is the preferred way to handle these endnotes.

Perhaps Axe could reduce the severity of its "missing children" checks wrt missing `doc-biblioentry` and `doc-endnote` roles, or eliminate them entirely.

WilcoFiers:
I'm aware of the change in DPUB ARIA, but I don't really understand what you are suggesting to do. Can you provide an example of something that axe-core is currently failing, that you would like to see the severity changed on?

Seirdy:
Currently failing:
- URL: https://seirdy.one/2021/01/12/password-strength.html
- Error: "Certain ARIA roles must contain particular children"

Apparently `doc-endnote` is missing.

The error is triggered at:

``` html
<section class="footnotes" role="doc-endnotes" aria-labelledby="note-hd">
```

Upon further checking, I actually don't see any child `<li>` element that *isn't* labelled `doc-endnote`; this is starting to look like a bug. I'll open another issue if this is the case.

That being said, since `doc-endnote` is deprecated, axe shouldn't raise this error when `doc-endnote` isn't found as a child of `doc-endnotes`.

Seirdy:
Full context:

``` html
<section class="footnotes" role="doc-endnotes" aria-labelledby="note-hd">
	<h2 id="note-hd">References and endnotes</h2>
	<ol>
		<li id="fn:1" role="doc-endnote">
			<p>James Massey (1994). &ldquo;Guessing and entropy&rdquo; (PDF). Proceedings of 1994 IEEE
			International Symposium on Information Theory. IEEE. p. 204. <a href="#fnref:1" class="footnote-backref" role="doc-backlink">&#x21a9;&#xfe0e;</a></p>
		</li>
		<li id="fn:2" role="doc-endnote">
			<p>Assis, A. K. T.; Neves, M. C. D. (3 July 1995). &ldquo;History of the 2.7 K
			Temperature Prior to Penzias and Wilson&rdquo; <a href="#fnref:2" class="footnote-backref" role="doc-backlink">&#x21a9;&#xfe0e;</a></p>
		</li>
		<li id="fn:3" role="doc-endnote">
[...]

PR #3280, opened 2021-11-12
feat(dpub): upgrade to DPUB 1.1 and report deprecated roles

WilcoFiers:
This PR updates axe-core to use DPUB ARIA 1.1, which has a few changes:

- DPUB roles no longer have required context
- `doc-biblioentry` and `doc-endnotes` is now failed for being deprecated

Closes issue: #2759


lib/checks/aria/deprecatedrole.json
```diff
@@ -0,0 +1,11 @@
+{
+  "id": "deprecatedrole",
+  "evaluate": "deprecatedrole-evaluate",
+  "metadata": {
+    "impact": "minor",
+    "messages": {
+      "pass": "ARIA role is deprecated",
+      "fail": "The role used deprecated: ${data.values}"
```
straker, line 8:
```suggestion
      "fail": "The role used is deprecated: ${data.values}"
```

lib/checks/aria/deprecatedrole.json
```diff
@@ -0,0 +1,11 @@
+{
+  "id": "deprecatedrole",
+  "evaluate": "deprecatedrole-evaluate",
+  "metadata": {
+    "impact": "minor",
+    "messages": {
+      "pass": "ARIA role is deprecated",
```
straker, line 7:
```suggestion
      "pass": "ARIA role is not deprecated",
```

lib/checks/aria/deprecatedrole-evaluate.js
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Check that an elements semantic role is deprecated.
+ *
+ * Deprecated roles are taken from the `ariaRoles` standards object from the roles `deprecated` property.
+ *
+ * @memberof checks
+ * @return {Boolean} True if the elements semantic role is deprecated. False otherwise.
+ */
+export default function deprecatedroleEvaluate(node, options, virtualNode) {
+  const role = getRole(virtualNode, { dpub: true });
```
straker, line 13: Should we look at at each role segment instead of the semantic role?

```js
const roles = tokenList(virtualNode.attr('role'));
roles.forEach(role => {
  const roleDefinition = standards.ariaRoles[role];
  // ... 
});
```

lib/checks/aria/deprecatedrole.json
```diff
@@ -0,0 +1,11 @@
+{
+  "id": "deprecatedrole",
+  "evaluate": "deprecatedrole-evaluate",
+  "metadata": {
+    "impact": "minor",
```
straker, line 5: Should we make this somehow return incomplete rather than a failure? Merging this will start failing for something that passed for a non-major release.

straker (changes requested):
Also closes #3253

lib/checks/aria/deprecatedrole.json
WilcoFiers: That's why it's a `feat`. New failures can come up from minor releases.

straker (approved):

danielweck:
This PR introduced a bug, see issue: https://github.com/dequelabs/axe-core/issues/3372

Issue #3291, opened 2021-11-16
commons/dom/isVisible inconsistently handles undefined VNode


thuey:
Product: axe-core

Expectation: Gracefully handle when a VNode is undefined

Actual: An error is thrown

Motivation: Gracefully checking if the VNode is defined would be more consistent with surrounding code and would avoid errors.

Line 154 should handle the case where `vNode` is undefined, like the case is handled on 160

https://github.com/dequelabs/axe-core/blob/5908f0d644c20e7091329bd8bbeb191837d27feb/lib/commons/dom/is-visible.js#L154-L162

<pre><code>
axe-core version: 4.3.5

Browser and Assistive Technology versions
- Chrome Version 95.0.4638.69
</code></pre>

thuey:
After further investigation, this scenario should never have been hit. The problem was I was passing in a disconnected DOM node.

PR #3292, opened 2021-11-16
fix(commons/dom): handle case where vNode is undefined in isVisible

thuey:
Check to make sure `vNode` is defined before trying to access `vNode.parent`. This is inconsistent with what the code is already doing.

Closes issue: #3291

### DIFF
--- a/lib/commons/dom/is-visible.js
+++ b/lib/commons/dom/is-visible.js
@@ -151,7 +151,7 @@ function isVisible(el, screenReader, recursed) {
 
   // detatched virtual node
   if (!el) {
-    const parent = vNode.parent;
+    const parent = vNode && vNode.parent;
     let visible = true;
     if (parent) {
       visible = isVisible(parent, screenReader, true);


stephenmathieson (changes requested):
Needs a test.

lib/commons/dom/is-visible.js
```diff
@@ -151,7 +151,7 @@ function isVisible(el, screenReader, recursed) {
 
   // detatched virtual node
   if (!el) {
-    const parent = vNode.parent;
+    const parent = vNode && vNode.parent;
```
thuey, line 154: @WilcoFiers I am hitting a case where `vNode` is undefined when `el` was a document fragment (but `el` also got set to `undefined` on line 138 because `el.host` was undefined). So, basically, both `el` and `vNode` are undefined by the time you get to line 154. Based on the comment `// detatched virtual node`, it seems like the assumption here is that this block of code will only run for virtual nodes (which the document fragment is not).

This document fragment appears to have come from a user-agent shadow DOM when we are checking `isVisible` on the SVG that is inside the user-agent shadow DOM:
image

The SVG referenced here is a symbol:
image


Although checking if `vNode` is defined or not would stop the error from happening, maybe we shouldn't get into this situation to begin with. Should SVG element that is in this user-agent shadow DOM be filtered out somehow?

thuey:
After further investigation, this scenario should never have been hit. The problem was I was passing in a disconnected DOM node.

Issue #3309, opened 2021-11-24
Color-contrast inconsistency of bgOverlap message


WilcoFiers:
The code to detect overlap only works when an element is scrolled into view. This isn't at all necessary, and is a hold-over from before axe-core was using a grid to test color-contrast. You can tell by looking at something like the following, which reports `#target` as overlapped when scrolled into view, but ignores the overlapping element when it's not in view.

```html
<div style="height: 1em; overflow: auto; background: cyan;">
  <div style="background: rgba(0,255,255,0.6); margin-bottom: -1em; height: 1em; position:relative;"></div>
  <div id="target">foo</div>
</div>
```

I haven't seen this produce false positives, but in theory it can. The more problematic part is that it changes the result, based on where in the page the scroll happens.

padmavemulapati:
Verified with the latest dev branch code,
Seeing no failures on `color-contrast`
when `target` as id and overlapped it ignores the overlapping element and also for inline elements with position absolute, these two cases show it as `incomplete`
otherwise:
non-opaque is obscuring it, scrolled out of view, and nested inline elements in the middle of a text show in `scrollable-region-focusable` error not `color-contrast`
![image](https://user-images.githubusercontent.com/47409579/147907560-902ff468-28da-4620-914e-2d2dd959f20a.png)

PR #3310, opened 2021-11-24
fix(color-contrast): inconsistency of bgOverlap message based on scroll

WilcoFiers:
bgOverlap was not computed correctly. What it was doing is use elementsFromPoint on the top-right pixel of a field. It got the tests to pass, but doing that doesn't actually tell us the difference between an element that overlaps, and a non-text element that simply sits inline with the text of an element. Because axe-core no longer scrolls the page when running color-contrast, our use of elementsFromPoint caused inconsistencies.

The fix here is to check if an element that appears visually on top of the element with text is actually an inline descendant of it. That gets rid of elementsFromPoint, solving the inconsistency problem, and it means we can actually properly tell what elements might overlap, and which ones are in the same flow as the text.

Closes issue: #3309

### DIFF
--- a/lib/commons/color/get-background-stack.js
+++ b/lib/commons/color/get-background-stack.js
@@ -2,39 +2,28 @@ import filteredRectStack from './filtered-rect-stack';
 import elementHasImage from './element-has-image';
 import getOwnBackgroundColor from './get-own-background-color';
 import incompleteData from './incomplete-data';
-import shadowElementsFromPoint from '../dom/shadow-elements-from-point';
 import reduceToElementsBelowFloating from '../dom/reduce-to-elements-below-floating';
 
 /**
- * Determines overlap of node's content with a bgNode. Used for inline elements
+ * Determine if element B is an inline descendant of A
  * @private
- * @param {Element} targetElement
- * @param {Element} bgNode
+ * @param {Element} node
+ * @param {Element} descendant
  * @return {Boolean}
  */
-function contentOverlapping(targetElement, bgNode) {
-  // get content box of target element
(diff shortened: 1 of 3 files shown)


test/commons/color/get-background-color.js
```diff
@@ -1046,7 +1080,7 @@ describe('color.getBackgroundColor', function() {
 
     it('returns the html background when body does not cover the element', function() {
       fixture.innerHTML =
-        '<div id="target" style="position: absolute; top: 1000px;"><label>elm<input></label></div>';
+        '<div id="target" style="position: absolute; top: 1000px;">elm<input></div>';
```
WilcoFiers, line 1083: Element would never target the parent of en element with text.

lib/commons/color/get-background-stack.js
```diff
@@ -2,59 +2,45 @@ import filteredRectStack from './filtered-rect-stack';
...
-        // remove elements not contributing to the background
-        elmStack.splice(i, 1);
-      }
+  // Reverse order, so that we can safely splice
+  for (let i = elmIndex - 1; i >= 0; i--) {
+    if (!isInlineDescendant(originalElm, elmStack[i])) {
+      return true;
     }
+    // Ignore inline descendants, for example:
+    // <p>text <img></p>; We don't care about the <img> element,
```
dylanb, line 40: What happens if `` is positioned somehow, acting like a background?

WilcoFiers: position: absolute / sticky force an element into display:block. I'll put in a test for it though.

PR #3320, opened 2021-12-03
adds role=radio to img alt=foo allowances

scottaohara:
A `role=radio` was meant to be an allowance for an `img alt=foo`, but it had mistakenly been left out of the spec.
ARIA in HTML is fixing this oversight, https://github.com/w3c/html-aria/pull/381, so this PR is to match this added allowance.

### DIFF
--- a/lib/standards/html-elms.js
+++ b/lib/standards/html-elms.js
@@ -358,7 +358,7 @@ const htmlElms = {
           'menuitemradio',
           'option',
           'progressbar',
-	  'radio',
+          'radio',
           'scrollbar',
           'separator',
           'slider',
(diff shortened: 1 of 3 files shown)


WilcoFiers (changes requested):
Thanks Scott. Would you be up for adding a passing integration test for this?

https://github.com/dequelabs/axe-core/tree/develop/test/integration/rules/aria-allowed-attr

scottaohara:
@WilcoFiers would this be a new type of integration test then?  it looks like what's in that folder now is divs with allowed roles and attributes.  but this change is for the img element role allowances.  maybe i'm missing something here.

WilcoFiers:
@scottaohara You are right. Apologies. I added the wrong link. This is about what roles are allowed, not what attributes:
https://github.com/dequelabs/axe-core/tree/develop/test/integration/rules/aria-allowed-role

WilcoFiers (approved):
LGTM. Thank you again @scottaohara!

PR #3327, opened 2021-12-13
perf(gather): Faster shadowDOM testing

WilcoFiers:
Testing if deeply nested shadow DOM nodes are in context can be a very slow operation, especially on sites that make extreme use of shadow DOM (think 30+ layers of nested shadow trees). For a very large part, this is because calling  `containsShadowChild` thousands of times recursively is very slow.

This PR avoids calling `isNodeInContext` when it isn't necessary (when there are no excluded nodes), and makes `containsShadowChild` significantly faster by walking down to the root, rather than walking the entire tree.

### DIFF
--- a/lib/core/utils/contains.js
+++ b/lib/core/utils/contains.js
@@ -13,9 +13,10 @@ function contains(vNode, otherVNode) {
     if (vNode.shadowId === otherVNode.shadowId) {
       return true;
     }
-    return !!vNode.children.find(child => {
-      return containsShadowChild(child, otherVNode);
-    });
+    if (!otherVNode.parent) {
+      return false;
+    }
+    return containsShadowChild(vNode, otherVNode.parent);
   }
 
   if (vNode.shadowId || otherVNode.shadowId) {
@@ -31,7 +32,7 @@ function contains(vNode, otherVNode) {
       vNode.actualNode.compareDocumentPosition(otherVNode.actualNode) & 16
     );
   } else {
-    // fallback for virtualNode only contexts (e.g. linting)
+    // fallback for virtualNode only contexts
     // @see https://github.com/Financial-Times/polyfill-service/pull/183/files
     do {
       if (otherVNode === vNode) {
--- a/lib/core/utils/contains.js
+++ b/lib/core/utils/contains.js
@@ -6,42 +6,32 @@
  * @param  {VirtualNode} otherVNode The vNode to test is contained by `vNode`
  * @return {Boolean}           Whether `vNode` contains `otherVNode`
  */
-function contains(vNode, otherVNode) {
+export default function contains(vNode, otherVNode) {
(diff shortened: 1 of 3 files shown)


test/integration/virtual-rules/nested-interactive.js
```diff
@@ -109,7 +109,7 @@ describe('nested-interactive virtual-rule', function() {
 
     var results = axe.runVirtualRule('nested-interactive', node);
 
-    assert.lengthOf(results.passes, 0);
+    assert.lengthOf(results.passes, 1);
```
WilcoFiers, line 112: I have no explanation for why this test wasn't erroring before... There are two applicable nodes in this tree, one passes, the other fails.

lib/core/utils/select.js
```diff
@@ -85,3 +81,19 @@ function select(selector, context) {
...
 export default select;
+
+/**
+ * Return a filter method to test if a node is in context; or
+ *  null if the node is always included.
+ * @private
+ * @param {Context}
+ * @return {Function|null}
+ */
+function getContextFilter(context) {
```
WilcoFiers, line 92: I have no idea what tests to write for this. I'm not sure it's necessary, since PR doesn't change output. It just makes some things faster. If you feel tests are necessary, please let me know how.

lib/core/utils/contains.js
```diff
@@ -6,41 +6,32 @@
...
+export default function contains(vNode, otherVNode) {
   /*eslint no-bitwise: 0*/
-
-  function containsShadowChild(vNode, otherVNode) {
-    if (vNode.shadowId === otherVNode.shadowId) {
-      return true;
-    }
-    return !!vNode.children.find(child => {
-      return containsShadowChild(child, otherVNode);
-    });
```
WilcoFiers, line 18: This was the biggest problem. Searching all descendants of vNode instead of just looking at the ancestors of ovetherVNode.

dylanb (approved):

test/integration/virtual-rules/nested-interactive.js
dylanb: this blew my mind but I gave up trying to understand why

lib/core/utils/select.js
dylanb: You could create a very deep shadow tree that times out the tests with the old code and doesn't now?

test/integration/virtual-rules/nested-interactive.js
WilcoFiers: Yeah, likewise.

lib/core/utils/select.js
WilcoFiers: I suppose so... Didn't try that. I'll play around with that for a bit, see if that works.

michael-siek:
Merging https://github.com/dequelabs/axe-core/issues/3332

PR #3335, opened 2022-01-06
ci: always install chromedriver version that matches chrome version

straker:
We've continually had problems with chromedriver version. When we had it as `@latest` we ran into a problem where CircleCI hadn't updated to the latest Chrome version so the test failed. When we pin we run into problems with always using that version of chromedriver even if the [version of Chrome is newer](https://app.circleci.com/pipelines/github/dequelabs/axe-core/3246/workflows/5fe071cd-f02f-4688-8106-ac1d293fa11f/jobs/39937).

This pr creates a CircleCI script that will get the current version of Chrome installed and then install the matching chromedriver version. Tested it and it works for both [normal test](https://app.circleci.com/pipelines/github/dequelabs/axe-core/3249/workflows/876574e9-d252-4a59-aaa7-a4b05d16d400/jobs/39962) and [nightly](https://app.circleci.com/pipelines/github/dequelabs/axe-core/3251/workflows/3d04e65c-e102-4704-a846-5fa65f0333e4/jobs/39969) (after Chrome Beta is installed). This also means we can just pin to the latest version of chromedriver for local dev.


package.json
```diff
@@ -105,7 +105,8 @@
     "next-release": "standard-version --scripts.prebump=./build/next-version.js --skip.commit=true --skip.tag=true",
     "sri-update": "grunt build && node build/sri-update && git add sri-history.json",
     "sri-validate": "node build/sri-update --validate",
-    "fmt": "prettier --write .*.json *.{json,md,js} **/*.ts './{.circleci,.github,build,doc,lib,locales,test,typings}/**/*.{json,md,js,ts,html}'"
+    "fmt": "prettier --write .*.json *.{json,md,js} **/*.ts './{.circleci,.github,build,doc,lib,locales,test,typings}/**/*.{json,md,js,ts,html}'",
+    "postinstall": "build/postinstall.sh"
```
stephenmathieson, line 109: We cannot use `postinstall` for this. The `postinstall` script happens on users' machines (post installation).

We could use `prepare` instead.

```suggestion
    "prepare": "build/postinstall.sh"
```

stephenmathieson (changes requested):

stephenmathieson:
If this is fixing a CI-only problem, why don't we just have CircleCI run the script (rather than coupling it to `npm install`)?

build/postinstall.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
stephenmathieson, line 1: What about Windows?

stephenmathieson (changes requested):

straker:
The way I see it this could also be a contributor problem as we are pinning to a version of chromedriver that may not work with the users installed version of chrome.

build/postinstall.sh
straker: Currently failing, but trying to see if this even works at all.

package.json
straker: I thought prepare was only run on `publish` not install.

build/postinstall.sh
stephenmathieson: It doesn't. Windows doesn't support bash.

straker:
Going to close. Doesn't look like Windows provides a way to do this. Will look for an alternative solution.

package.json
stephenmathieson: From https://docs.npmjs.com/cli/v8/using-npm/scripts, the `prepare` script is run:

- Runs any time before the package is packed, i.e. during npm publish and npm pack
- Runs BEFORE the package is packed
- Runs BEFORE the package is published
- Runs on local npm install without any arguments

.circleci/config.yml
```diff
@@ -36,6 +36,20 @@ restore_dependency_cache_win: &restore_dependency_cache_win
       - v9-cache-win-{{ checksum "package-lock.json" }}
       - v9-cache-win-
 
+# install the version of chromedriver that matches the currently installed version of Chrome (taken from CircleCI chromedriver script)
+install_chromedriver: &install_chromedriver
```
stephenmathieson, line 40: Might be nice to make it explicit that this cannot be used on Windows.

```suggestion
install_linux_chromedriver: &install_linux_chromedriver
```

stephenmathieson (dismissed):

stephenmathieson (approved):

Issue #3334, opened 2022-01-03
Wrong typing for include-exclude object


matyasf:
Product: axe-core, issue type: Typing bug

Expectation: include-exclude object accepts DOM node instances
Actual: it only accepts strings

See the docs for this: https://github.com/dequelabs/axe-core/blob/develop/doc/API.md#include-exclude-object a DOM node is a valid value.
Compare this to its typing: https://github.com/dequelabs/axe-core/blob/develop/axe.d.ts#L54

<pre><code>
axe-core version: 4.3.5
</code></pre>

straker:
Thanks for the issue. Have a pr to fix this.

PR #3338, opened 2022-01-07
fix(axe.d.ts): allow Node for include/exclude object

straker:
These all work correctly:

```js
const div = document.querySelector('div');
await axe.run({ include: div });
await axe.run({ include: [div] });
await axe.run({ include: doument.querySelectorAll('div') });
```

Closes issue: #3334

### DIFF
--- a/axe.d.ts
+++ b/axe.d.ts
@@ -50,8 +50,8 @@ declare namespace axe {
   type CrossFrameSelector = CrossTreeSelector[];
 
   type ContextObject = {
-    include?: BaseSelector | Array<BaseSelector | BaseSelector[]>;
-    exclude?: BaseSelector | Array<BaseSelector | BaseSelector[]>;
+    include?: Node | BaseSelector | Array<Node | BaseSelector | BaseSelector[]>;
+    exclude?: Node | BaseSelector | Array<Node | BaseSelector | BaseSelector[]>;
   };
 
   type RunCallback = (error: Error, results: AxeResults) => void;
--- a/typings/axe-core/axe-core-tests.ts
+++ b/typings/axe-core/axe-core-tests.ts
@@ -1,7 +1,7 @@
 import * as axe from '../../axe';
 
 var context: any = document;
-var $fixture = document.querySelectorAll('div');
+var $fixture = [document];
 var options = { iframes: false, selectors: false, elementRef: false };
 
 axe.run(context, {}, (error: Error, results: axe.AxeResults) => {
--- a/typings/axe-core/axe-core-tests.ts
+++ b/typings/axe-core/axe-core-tests.ts
@@ -1,7 +1,7 @@
 import * as axe from '../../axe';
 
 var context: any = document;
-var $fixture: any = {};
+var $fixture: NodeList = document.querySelectorAll('div');
 var options = { iframes: false, selectors: false, elementRef: false };
 
 axe.run(context, {}, (error: Error, results: axe.AxeResults) => {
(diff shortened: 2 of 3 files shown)


typings/axe-core/axe-core-tests.ts
```diff
@@ -1,7 +1,7 @@
 import * as axe from '../../axe';
 
 var context: any = document;
-var $fixture: any = {};
+var $fixture: NodeList = document.querySelectorAll('div');
```
stephenmathieson, line 4: `NodeList` is not quite right here. QSA returns `NodeList<HTMLtypeElement>`, where `type` is `Div` in this case, so this should be:

```suggestion
var $fixture: NodeList<HTMLDivElement> = document.querySelectorAll('div');
```

However, declaring any type here is redundant, as the compiler knows what QSA returns. IMO this should just be:

```suggestion
var $fixture = document.querySelectorAll('div');
```

stephenmathieson (approved):

stephenmathieson (approved):

typings/axe-core/axe-core-tests.ts
```diff
@@ -1,7 +1,7 @@
 import * as axe from '../../axe';
 
 var context: any = document;
-var $fixture: any = {};
+var $fixture = document.querySelectorAll('div');
```
WilcoFiers, line 4: I'm a little surprised this works, since QSA returns a NodeList rather than an array. Should we explicitly mention NodeList?

straker: Looked into this. It looks like it doesn't matter what exactly is passed in for the include/exclude object so long as it is array-like. Once we get the context object we immediately call [parseSelectorArray](https://github.com/dequelabs/axe-core/blob/develop/lib/core/base/context.js#L248), which loops over the context include/object object using a [for loop that looks at the length of the object](https://github.com/dequelabs/axe-core/blob/develop/lib/core/base/context.js#L133).

I'm OK if we don't mention this particular case for now. We should loop back to this and determine if it's something we explicitly want to support, and if so add an official test to our code to ensure it continues to work.

stephenmathieson: In that case, wouldn't `ArrayLike<Node>` be the best option?

straker: If we decide to officially support it then ya, that should be the correct type. For now I'm going to change the ts test to not use querySelectorAll and will update it back in another pr if we decide to support it.

axe.d.ts
```diff
@@ -50,8 +50,8 @@ declare namespace axe {
   type CrossFrameSelector = CrossTreeSelector[];
 
   type ContextObject = {
-    include?: BaseSelector | Array<BaseSelector | BaseSelector[]>;
-    exclude?: BaseSelector | Array<BaseSelector | BaseSelector[]>;
+    include?: Node | BaseSelector | Array<Node | BaseSelector | BaseSelector[]>;
+    exclude?: Node | BaseSelector | Array<Node | BaseSelector | BaseSelector[]>;
```
stephenmathieson, line 54: NBD, but it sounds like this should be:

```suggestion
    include?: Node | BaseSelector | ArrayLike<Node> | Array<BaseSelector | BaseSelector[]>;
    exclude?: Node | BaseSelector | ArrayLike<Node> | Array<BaseSelector | BaseSelector[]>;
```

WilcoFiers (approved):

Issue #3241, opened 2021-10-28
aria-allowed-attr possible regression?


thepassle:
Product: axe-core

Expectation: element with both `aria-controls` and `aria-expanded` passes

Actual: element with both `aria-controls` and `aria-expanded` fails

Motivation: I believe it might have been an accidental regression via https://github.com/dequelabs/axe-core/blob/develop/CHANGELOG.md#:~:text=aria-allowed-attr%3A%20check%20for%20invalid

<pre><code>
axe-core version: 4.3.4
</code></pre>

We recently found a test started failing that uses axe-core. We previously had `axe-core@4.3.3` installed, where the test passes, the test has started failing since `axe-core@4.3.4`, which leads me to believe it may have been an accidental regression via this change: https://github.com/dequelabs/axe-core/blob/develop/CHANGELOG.md#:~:text=aria-allowed-attr%3A%20check%20for%20invalid

For context, we have an element with both an `aria-controls` and `aria-expanded` attribute, e.g.:
```html
<my-el aria-controls="menu" aria-expanded="false"></my-el>
```
As far as my accessibility knowledge goes, this should be valid (but I'm happy to be corrected if I'm wrong). I also checked with a couple of colleagues, who also expected this to be valid.

WilcoFiers:
We've released 4.3.5, which has disabled this test. It wasn't meant to go out in a patch, we had it planned for the 4.4.0 release. We're going to take a closer look at how this can affect custom elements. There may be ways in which this can be alright, but Steve's right that these attributes don't work just by themselves. Either you reflect them onto another element, or you make that particular component fully accessible (providing role and focus).

svanherk:
If it helps, we were also hit by this problem and use these attributes on custom elements.  We do what you described above - our custom button that opens a dropdown is written like this:
```
<custom-button aria-expanded="true"></custom-button>
```
And turns into this:
```
<custom-button aria-expanded="true">
  #shadow-root
    <button aria-expanded="true"></button>
</custom-button>
```
As you said, we really shouldn't have been doing this but it was interpreted correctly by the screenreader. For our newer components we've been using `data-aria-expanded` or just `expanded` on the custom element instead.  We have a number of older components that will need to be updated and fixed throughout our application, so we'll be blocked from going to `4.4.0` until we do that.

straker:
@svanherk Thanks for letting us know. We'll try to come up with a better solution for 4.4.0 taking that into account.

WilcoFiers:
[...]

PR #3342, opened 2022-01-10
feat(aria-allowed-attr): report violation for non-global ARIA attrs on element without role

straker:
This change reintroduces the change to [report violations on elements without a role for non-global ARIA attrs](https://github.com/dequelabs/axe-core/pull/3102), which was [reverted](https://github.com/dequelabs/axe-core/pull/3243) when we accidentally released it.

This also introduces a fix for custom elements where we will report Needs Review for any custom element that uses a non-global ARIA attribute (rather than a failure).

Closes issue: https://github.com/dequelabs/axe-core/issues/3241


straker:
The error from aria practices seems to be a legit issue on aria practices? The violation reported on the `button_idl.html` page is from an anchor element which uses `aria-pressed`, which shouldn't be allowed.

This is the page: https://www.w3.org/TR/wai-aria-practices-1.2/examples/button/button_idl.html

And the error:

```js
{
    "any": [
        {
            "id": "aria-allowed-attr",
            "data": [
                "aria-pressed=\"false\""
            ],
            "relatedNodes": [],
            "impact": "critical",
            "message": "ARIA attribute is not allowed: aria-pressed=\"false\""
        }
    ],
    "all": [],
    "none": [],
    "impact": "critical",
    "html": "<a tabindex=\"0\" id=\"toggle\" aria-pressed=\"false\">\n          Mute\n          <svg aria-hidden=\"true\" focusable=\"false\">\n            <use xlink:href=\"#icon-sound\"></use>\n          </svg>\n        </a>",
    "target": [
        "#toggle"
    ],
    "failureSummary": "Fix any of the following:\n  ARIA attribute is not allowed: aria-pressed=\"false\""
}
```

lib/checks/aria/aria-allowed-attr.json
```diff
@@ -16,7 +16,8 @@
       "fail": {
         "singular": "ARIA attribute is not allowed: ${data.values}",
         "plural": "ARIA attributes are not allowed: ${data.values}"
-      }
+      },
+      "incomplete": "ARIA attribute is not allowed on custom element: ${data.values}"
```
WilcoFiers, line 20: How about:

```suggestion
      "incomplete": "Check that there is no problem the ARIA attribute is ignored on this element: ${data.values}"
```

WilcoFiers (changes requested):

lib/checks/aria/aria-allowed-attr-evaluate.js
```diff
@@ -82,7 +82,7 @@ function ariaAllowedAttrEvaluate(node, options, virtualNode) {
 
   if (invalid.length) {
     this.data(invalid);
-    return false;
+    return isHtmlElement(virtualNode) ? false : undefined;
```
WilcoFiers, line 85: I think this exception should only be used:
1. If the element is not an HTML element
2. If the element is not focusable
3. If the element has no role

lib/checks/aria/aria-allowed-attr.json
```diff
@@ -17,7 +17,7 @@
         "singular": "ARIA attribute is not allowed: ${data.values}",
         "plural": "ARIA attributes are not allowed: ${data.values}"
       },
-      "incomplete": "ARIA attribute is not allowed on custom element: ${data.values}"
+      "incomplete": "Check that there is no problem the ARIA attribute is ignored on this element: ${data.values}"
```
WilcoFiers, line 20:
```suggestion
      "incomplete": "Check that there is no problem if the ARIA attribute is ignored on this element: ${data.values}"
```

WilcoFiers (approved):

PR #3343, opened 2022-01-10
update role allowances for aria-expanded [WIP]

### LINKED ISSUE #3339: roles that allow aria-expanded and aria-roles.js question
ARIA 1.2 has fixed an error where ARIA 1.1 allowed `aria-expanded` on a number of elements that didn't make sense.  e.g., `role=alert aria-expanded=...`.  [ARIA issue where this topic was discussed](https://github.com/w3c/aria/issues/681) - fixed in 2019, so has been part of the initial and current ARIA 1.2 CR draft.

I noticed in the aria-roles.js file that `aria-expanded` is still marked as an allowed attribute on these roles, which at first made sense to me because the file notes on the first line that it is referencing ARIA 1.1... so since they were allowed then, if axe is still referencing ARIA 1.1 then they would still be allowed, though their use by authors is suspect / probably incorrect.   

However, upon further review of the file, there are a number of roles that were introduces in ARIA 1.2 that are also included... which again also makes sense as many of them have been implemented by browsers... but this then made me wonder if it would be a good idea for axe to also correct the `aria-expanded` allowances, even though the correction is tied to ARIA 1.2.  

Regardless of when axe might want to pull in the correction, I'd be happy to create a PR to remove the `aria-expanded` allowances from the necessary roles in this file.

scottaohara:
All roles reviewed to largely remove (but in a few cases add) `aria-expanded` as a supported attribute.  

This matches both work done in ARIA 1.2 to correct for roles that should not have allowed the attribute, while also incorporating some ARIA 1.2 updates where some roles had support for the attribute added.

closes #3339

### DIFF
--- a/lib/standards/aria-roles.js
+++ b/lib/standards/aria-roles.js
@@ -18,12 +18,11 @@
 const ariaRoles = {
   alert: {
     type: 'widget',
-    allowedAttrs: ['aria-expanded'],
     superclassRole: ['section']
   },
   alertdialog: {
     type: 'widget',
-    allowedAttrs: ['aria-expanded', 'aria-modal'],
+    allowedAttrs: ['aria-modal'],
     superclassRole: ['alert', 'dialog'],
     accessibleNameRequired: true
   },
@@ -38,12 +37,11 @@ const ariaRoles = {
   },
   article: {
     type: 'structure',
-    allowedAttrs: ['aria-posinset', 'aria-setsize', 'aria-expanded'],
+    allowedAttrs: ['aria-posinset', 'aria-setsize'],
     superclassRole: ['document']
   },
   banner: {
     type: 'landmark',
-    allowedAttrs: ['aria-expanded'],
     superclassRole: ['landmark']
   },
   blockquote: {
@@ -71,8 +69,7 @@ const ariaRoles = {
       'aria-colindex',
       'aria-colspan',
       'aria-rowindex',
-      'aria-rowspan',
-      'aria-expanded'
+      'aria-rowspan'
     ],
     superclassRole: ['section'],
     nameFromContent: true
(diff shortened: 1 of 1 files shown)


WilcoFiers (commented):
I'm a little hesitant to just pull attributes. I would prefer to flag them as deprecated. I'm not entirely sure how to do this. Perhaps a new check is needed for that. I'll talk to Steve about this. It's looking like axe-core 4.5 is going to get a more "sync with latest standards" focus, so I think we'll include it in that. I'll leave this open while we finish up 4.4 (3 more weeks).

scottaohara:
just for my own clarity, you mean that axe would want to report these as deprecated, even though their removal isn't an actual deprecation by the spec, but rather a fix for erroneously marking their use on these roles as allowed?

and understood on the timetable.  I made this PR knowing you had said it wouldn't make it into 4.4.

WilcoFiers:
@scottaohara That's what I'm thinking. More importantly, I don't want to just remove them like this PR is doing. I'm concerned this is going to be perceived as a regression. I think this needs some kind message explicitly calling out the change in the standard. Perhaps we make this more of a "deprecated or removed" type message, but something so that users don't think axe-core broke on them.

I would've preferred ARIA WG deprecate these TBH, but we're probably well passed the point that that's a reasonable request to make. This change is pretty well locked into 1.2.

scottaohara:
That all makes sense. Thanks. 

I’ll leave this PR open in case y’all want to use it as a quick reference to grab all the roles where aria-expanded was removed (and the few where it was added). But otherwise feel free to close it down, and again thanks for the explanation.

querkmachine:
Hello!

In the GOV.UK Design System we use `aria-expanded` for our [conditional checkbox pattern](https://design-system.service.gov.uk/components/checkboxes/#conditionally-revealing-a-related-question). Services using the Design System frequently receive feedback from accessibility auditors using Axe that this is a violation, as Axe hasn't been updated to incorporate this part of ARIA 1.2 yet. 

Just thought I'd give a poke as the original target of version 4.5 seems to have passed without comment.

Thanks!

colinrotherham:
@querkmachine Confirmed fixed with [`@axe-core/puppeteer`](https://www.npmjs.com/package/@axe-core/puppeteer) now including [axe-core](https://www.npmjs.com/package/axe-core) ~`^4.7.2`~ `~4.8.2`

* https://github.com/alphagov/govuk-frontend/pull/4335

Thanks all

Issue #3346, opened 2022-01-12
aria-prohibited-attr should always report incomplete if there is text in the subtree


WilcoFiers:
The following should have been reported as incomplete, but is failed instead:

```html
<div role="code" aria-label="foo">Hello world</div>
```

It should only fail if the content is missing, like in the following:

```html
<div role="code" aria-label="foo"></div>
```

This already works without the role. It goes bad because `subtreeText` isn't called correctly.

padmavemulapati:
Validated with the latest dev branch code base, and the test snippet 
`<div role="code" aria-label="foo">Hello world</div>` , it is showing `incomplete` results as shown in the screenshot:
![image](https://user-images.githubusercontent.com/47409579/149789155-3d16bb5c-f6e6-438f-b69b-620e1ecbd911.png)
It was `failing` earlier with the same code snippet instead `incomplete`
![image](https://user-images.githubusercontent.com/47409579/149789344-49103ad1-a4bc-4710-b0c1-3d970fbca2cd.png)

It fails when `<div role="code" aria-label="foo"></div>`

![image](https://user-images.githubusercontent.com/47409579/149789693-d1570fff-adea-4241-9b8d-c87413c676c3.png)

PR #3347, opened 2022-01-12
fix(prohibited-attr): always report incomplete if there is text in the subtree

WilcoFiers:
Subtree text wasn't computed for elements that are not named from content. To force this, subtreeDescendant: true had to be passed.

This PR also significantly improves the messages of the prohibited-attr check, which was the cause of some confusion.

Closes issue: #3346


test/checks/aria/aria-prohibited-attr.js
```diff
@@ -9,39 +9,79 @@ describe('aria-prohibited-attr', function() {
     checkContext.reset();
```
straker, line 9: Need tests for checking when the messageKey is set to `noRoleSingular` when returned as `false` and `undefined`

straker (changes requested):

straker (approved):

straker (approved):

Issue #3157, opened 2021-09-14
False positive: "Color-contrast" rule fails when using particular text-shadow values(2)


iamrafan:
```color-contrast``` rule fails when an element has text-shadow values as mentioned in Case-1 in the example. 

Demo link: https://codepen.io/iamrafan/full/MWoEpZj

<pre><code>axe-core version: 4.3.3</code></pre>

straker:
Thanks for the issue. We'll look into it.

WilcoFiers:
I'm doing some polling to decide if this should actually pass. I'm not convinced anymore. It's definitely on the edge. The inner shape of letters don't have a 4.5:1 contrast on this, but the contrast around the word does. I think there's a reasonable case to be made either way. Generally with axe we try to stick to the more permissive side of things, but I'd like to get a few more opinions in before deciding that.

WilcoFiers:
Update; I've asked a good number of folks about this and most think this should pass. Working on a fix today.

PR #3350, opened 2022-01-13
fix(color-contrast): check bg on fg contrast for thin text-shadows

WilcoFiers:
The initial text-shadow implementation overlooked that the shadow color can have the luminance between that of the foreground and background. Now, if you have a dark text, light background and a thin shadow that sits in between those two colors, axe will grab the contrast of the foreground and background color.

Related to that, the shadowOutlineEmMax threshold is bumped to .2em. I've showed a few people the .15em example and most seemed okay with comparing background and foreground for a shadow of that width.

This change also addressed some "false positive" edge cases, making the rule in general slightly more permissive when it comes to text-shadow.

Closes issue: #3157


lib/checks/color/color-contrast-evaluate.js
```diff
@@ -71,14 +71,16 @@ export default function colorContrastEvaluate(node, options, virtualNode) {
...
-    // Thin shadows can pass either by contrasting with the text color
-    // or when contrasting with the background.
     shadowColor = [...shadowColors, bgColor].reduce(flattenShadowColors);
-    const bgContrast = getContrast(bgColor, shadowColor);
-    const fgContrast = getContrast(shadowColor, fgColor);
-    contrast = Math.max(bgContrast, fgContrast);
-    contrastContributor =
-      bgContrast > fgContrast ? 'shadowOnBgColor' : 'fgOnShadowColor';
+    // Compare shadow, bgColor, textColor. Check passes if any is sufficient
+    const fgBgContrast = getContrast(bgColor, fgColor);
```
WilcoFiers, line 76: This is the critical change here. Considering fgColor / bgColor contrast even if there's a thin shadow.

lib/checks/color/color-contrast.json
```diff
@@ -17,7 +17,7 @@
       }
     },
     "pseudoSizeThreshold": 0.25,
-    "shadowOutlineEmMax": 0.1
+    "shadowOutlineEmMax": 0.2
```
WilcoFiers, line 20: After asking a few people I discovered the current threshold was slightly too strict. Increasing this number means the slightly larger shadows will be allowed to serve as text outlines, giving more opportunities for it to pass the contrast ratio.

test/integration/rules/color-contrast/text-shadows.html
```diff
@@ -303,220 +320,433 @@
 <section class="s1" id="b1">
   <div class="r0"><i id="sample1">Sample</i></div>
   <div class="r1">
-    <i>b1-r1-i1</i> <i>b1-r1-i2</i> <i>b1-r1-i3</i> <i>b1-r1-i4</i>
+    <i id="b1-r1-i1">b1-r1-i1</i>
```
WilcoFiers, line 323: I added IDs to make these easier to work with.

WilcoFiers (commented):
To see the results locally go to http://localhost:9876/test/integration/rules/color-contrast/text-shadows.html

The script on that page adds icons to what passes and what fails.

straker (approved):

Issue #2806, opened 2021-02-16
Color contrast does not accurately assess overflow


WilcoFiers:
This code causes a color contrast issue to be reported, even though the element with low contrast is hidden because of overflow. This type of thing can happen in carousels for example, when the opacity is reduced as the end state of a fade-out effect.

```html
<style>
  .container {
    width: 200px;
    height: 200px;
    overflow: hidden;
  }
  .foo * {
    width: 200px;
    height: 200px;
  }
  #two {
    color: #eee;
  }
</style>
<div class="container">
  <div class="foo" id="foo">
    <div id="one">hello</div>
    <div id="two">goodbye</div>
  </div>
</div>
```

waabid:
Hi Wilco,

We're seeing the transition issue as well. Here's a simple codepen that can repro it: https://codepen.io/dbjorge/pen/mdRopqe

The related issue on our end pertains to a carousel specifically, here: https://github.com/microsoft/accessibility-insights-web/issues/4144

WilcoFiers:
Unfortunately this hasn't quite made the cut for 4.3. We'll push it to 4.4.

PR #3351, opened 2022-01-14
feat: deprecate and replace dom.isVisible, utils.isHidden, and dom.isHiddenWithCss

### LINKED ISSUE #3617: Refactor multiple visibility functions
Currently we have 3 visibility functions: `isHidden`, `isHiddenWithCss`, and `isVisible`. It's not particularly clear which one should be used in what scenarios. Also, the `isHidden` function is in `utils` which means we can't access virtual nodes which makes the work for [knowing when elements are hidden in a closed details](https://github.com/dequelabs/axe-core/pull/3562) difficult.

We've decided to refactor the functions to make things more clear and greatly simplify the code (as there is a lot of duplicated logic between them). First, the names will be updated to better reflect what the logic is trying to say about when a node is visible (or hidden): 

- **isHidden** - will be deprecated. A new function called `isVisibleToScreenreader` will be created in `commons` (so we now have access to virtual nodes).
- **isVisible** - will be deprecated and renamed to `isVisibleOnScreen`. It will not take a `screenreader` option. `isVisibleToScreenreader` should be used instead of passing a `screenreader` option.
- **isHiddenWithCss** - will be deprecated and renamed to `isHiddenForEveryone`.

As part of the refactor, we will split out all the individual visibility checks (#3351) into their own functions that can be used by the 3 new visibility functions.

If possible, we would like to be able to have the 3 visibility functions call each other when needed (so `isHiddenForEveryone` would call `isVisibleToScreenreader` and `isVisibleOnScreen`). To do this we'll need to pass a flag to not have the functions recurse up the parent tree in order to just get the answer for the passed in node. We'll also have to figure out what to do with the cache for this type of call.

straker:
A precursor to fixing #2806 and #3532.

Closes #3617


lib/commons/dom/is-visible.js
```diff
@@ -179,56 +89,194 @@ function isVisible(el, screenReader, recursed) {
...
+    visible = isVisible(parent, screenReader, true);
+  }
+
+  if (vNode) {
+    vNode[cacheName] = visible;
+  }
+
+  return visible;
+}
+
```
WilcoFiers, line 128:
```suggestion

  const sharedHiddenMethods = [elementHidden, displayHidden];
  const offScreenMethods = [opacityHidden, clipHidden, scrollHidden, overflowHidden];
  const accTreeHiddenMethods = [elementHidden, ariaHidden, displayHidden];
  const nonRecursedScreenMethods = [displayMethods, offscreenHidden];

```

lib/commons/dom/is-visible.js
```diff
@@ -179,56 +89,194 @@ function isVisible(el, screenReader, recursed) {
...
+  // different ways to hide an element
+  const hiddenMethods = [
+    displayHidden,
+    opacityHidden,
+    clipHidden,
+    scrollHidden,
+    overflowHidden,
+    visiblityHidden,
+    offscreenHidden
+  ];
```
WilcoFiers, line 101: I think you should preselect these methods based on the current scenario, rather than pass `recursed` and `screenreader` into these methods:

```suggestion
  const hiddenMethods = sharedHiddenMethods.concat(
    screenReader ? accTreeHiddenMethods : offScreenMethods
  )
  if (!screenReader && !recursed) {
    hiddenMethods.push(...nonRecursedScreenMethods);
  }
```

lib/commons/dom/is-visible.js
```diff
@@ -179,56 +89,194 @@ function isVisible(el, screenReader, recursed) {
...
+    ? vNode.attr('aria-hidden')
+    : el.getAttribute('aria-hidden');
+  return !!screenReader && ariaHiddenValue === 'true';
+}
+
+/**
+ * Determine if an element is natively hidden
+ */
+function elementHidden({ nodeName }) {
+  return ['style', 'script', 'noscript', 'template'].includes(nodeName);
```
WilcoFiers, line 165: I'm leaning to stick the isAreaVisible call in here, abstract that logic away from the higher isVisible method. WDYT?

WilcoFiers (dismissed):
The direction looks fine. I left some suggestions on refactoring this to make the logic of which method does what a little more explicit. Obviously I'd like to see tests added to this.

lib/commons/dom/is-visible.js
```diff
@@ -179,56 +89,194 @@ function isVisible(el, screenReader, recursed) {
...
+ * Determine if an element is hidden setting the display property
+ */
+function displayHidden({ style }) {
+  return style.getPropertyValue('display') === 'none';
+}
+
+/**
+ * Determine if an element is hidden by setting the visibility property
+ */
+function visiblityHidden({ style, recursed }) {
```
WilcoFiers, line 146:
```suggestion
function visibilityHidden({ style, recursed }) {
```

lib/commons/dom/is-visible.js
```diff
@@ -139,14 +55,8 @@ function isVisible(el, screenReader, recursed) {
...
   // hidden from screen readers
-  if (screenReader) {
-    const ariaHiddenValue = vNode
-      ? vNode.attr('aria-hidden')
-      : el.getAttribute('aria-hidden');
-
-    if (ariaHiddenValue === 'true') {
-      return false;
-    }
+  if (ariaHidden({ screenReader, el, vNode })) {
```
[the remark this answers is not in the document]
straker: This has to come before the `(!el)` check for disconnected nodes

lib/commons/dom/is-visible.js
```diff
@@ -179,56 +89,194 @@ function isVisible(el, screenReader, recursed) {
     return isAreaVisible(el, screenReader, recursed);
```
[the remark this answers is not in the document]
straker: Can't as the bug in Firefox for area elements means it has to return the result of `isAreaVisible` and not check `displayNone` (which always returns false for the area)

lib/commons/dom/is-visible.js
straker: So I tried this and I'm not sure I like the approach. Because of the divergent code paths with disconnected trees (which can check ariaHidden and elementHidden, but nothing else), the final code ended up looking something like this:

```js
let hiddenMethods = [elementHidden];
let style;
if (screenReader) {
  hiddenMethods.push(ariaHidden)
}

if (el) {
  if (nodeType === DOCUMENT_FRAGMENT_NODE) {
    el = el.host; // grab the host Node
  }

  style = window.getComputedStyle(el, null);
  if (style === null) {
    return false;
  }

  /**
   * check visibility of `AREA`
   * Note:
   * Firefox's user-agent always sets `AREA` element to `display:none`
   * hence excluding the edge case, for visibility computation
   */
  if (nodeName === 'area') {
    return isAreaVisible(el, screenReader, recursed);
  }

  hiddenMethods.push(displayHidden);

  if (!screenReader) {
    hiddenMethods.push(
      opacityHidden,
      clipHidden,
      scrollHidden,
      overflowHidden
    );
  }
  if (!recursed) {
    hiddenMethods.push(visibilityHidden);
  }
  if (!screenReader && !recursed) {
    hiddenMethods.push(offscreenHidden);
  }
}

if (
  hiddenMethods.some(method =>
    method({
      el,
      vNode,
      nodeName,
      style
    })
  )
) {
  return false;
}
```

I find it harder to see what's going on, and even harder to know which methods are allowed to look at `vNode` and which must stick to just using the `el`. Also that style is passed to all methods but isn't set unless you have `el` is kinda odd to me.

lib/commons/dom/is-visible.js
```diff
@@ -13,14 +13,187 @@ const clipRegex = /rect\s*\(([0-9]+)px,?\s*([0-9]+)px,?\s*([0-9]+)px,?\s*([0-9]+
...
+    )
+  ) {
+    return false;
+  }
+
+  const parent = el
+    ? el.assignedSlot
+      ? el.assignedSlot
+      : el.parentNode
+    : vNode.parent;
```
WilcoFiers, line 109: Why isn't this always vNode.parent? Also, shouldn't this include the host? I would suggest using `getComposedParent` if we for some reason can't use vNode.parent.

lib/commons/dom/is-visible.js
```diff
@@ -13,14 +13,187 @@ const clipRegex = /rect\s*\(([0-9]+)px,?\s*([0-9]+)px,?\s*([0-9]+)px,?\s*([0-9]+
...
+  let visible = el ? false : true;
+  if (parent) {
+    visible = isVisible(parent, screenReader, true);
+  }
```
WilcoFiers, line 113: This looks kinda odd. How about: 
```suggestion
  let visible;
  if (parent) {
    visible = isVisible(parent, screenReader, true);
  } else {
    visible = el ? false : true
  }
```

lib/commons/dom/is-visible.js
```diff
@@ -13,14 +13,187 @@ const clipRegex = /rect\s*\(([0-9]+)px,?\s*([0-9]+)px,?\s*([0-9]+)px,?\s*([0-9]+
...
+ * @param {Boolean} recursed
+ * @return {Boolean} The element's visibilty status
+ */
+export default function isVisible(el, screenReader, recursed) {
+  if (!el) {
+    throw new TypeError(
+      'Cannot determine if element is visible for non-DOM nodes'
+    );
+  }
+
```
WilcoFiers, line 31: I think this is better done with assert.
```suggestion
  assert(el, 'Cannot determine if element is visible for non-DOM nodes');
```

lib/commons/dom/is-visible.js
```diff
@@ -13,14 +13,187 @@ const clipRegex = /rect\s*\(([0-9]+)px,?\s*([0-9]+)px,?\s*([0-9]+)px,?\s*([0-9]+
...
+  const vNode = el instanceof AbstractVirtualNode ? el : getNodeFromTree(el);
+  el = vNode ? vNode.actualNode : el;
+
+  const cacheName = '_isVisible' + (screenReader ? 'ScreenReader' : '');
+  const { DOCUMENT_NODE, DOCUMENT_FRAGMENT_NODE } = window.Node ?? {};
+  const nodeType = vNode ? vNode.props.nodeType : el.nodeType;
+  const nodeName = vNode ? vNode.props.nodeName : el.nodeName.toLowerCase();
```
WilcoFiers, line 38: Why wouldn't there be a virtual node? Is this because you're recursing on document fragments? Seems like that can be avoided.

lib/commons/dom/is-visible.js
```diff
@@ -13,14 +13,187 @@ const clipRegex = /rect\s*\(([0-9]+)px,?\s*([0-9]+)px,?\s*([0-9]+)px,?\s*([0-9]+
...
+
+  const hiddenMethods = [elementHidden];
+  let style;
+  if (screenReader) {
+    hiddenMethods.push(ariaHidden);
+  }
+
+  if (el) {
+    if (nodeType === DOCUMENT_FRAGMENT_NODE) {
+      el = el.host; // grab the host Node
```
WilcoFiers, line 56: Probably best not to recurse on doc fragements and just pass the host node, no?

lib/commons/dom/is-visible.js
```diff
@@ -13,14 +13,187 @@ const clipRegex = /rect\s*\(([0-9]+)px,?\s*([0-9]+)px,?\s*([0-9]+)px,?\s*([0-9]+
...
+    }
+
+    style = window.getComputedStyle(el, null);
+    if (style === null) {
+      return false;
```
WilcoFiers, line 61: How is this possible?

WilcoFiers (dismissed):

lib/commons/dom/is-visible.js
straker: We can't. Firefox always sets the `display` style of an area element as `none`, which means it will always fail the `displayHidden` check. So we need to run the function before we call `displayHidden` and return rather than let the rest of the function continue.

lib/commons/dom/is-visible.js
straker: Added a comment. TL;DR is that we recurse up the DOM tree and can find parent elements not in the virtual node tree, but we still need to know the visible state of all elements.

lib/commons/dom/is-visible.js
straker: Seems to be an old Firefox bug that [has been in the code for a decade](https://github.com/dequelabs/axe-core/commit/b1d4b027f46a7f9d8d8cf7b600fea91f8e40830d#diff-3e7200e9a058efa36f66332b3f68b47ff659a5bf0241a1a1dd871eda19ae8ce0R39). I checked latest firefox and IE11 and they both seem to return a style object and not `null`.

lib/commons/dom/is-visible.js
```diff
@@ -13,14 +15,187 @@ const clipRegex = /rect\s*\(([0-9]+)px,?\s*([0-9]+)px,?\s*([0-9]+)px,?\s*([0-9]+
...
+        vNode,
+        nodeName,
+        style
+      })
+    )
+  ) {
+    return false;
+  }
+
+  if (nodeName === 'html') {
```
straker, line 101: `getComposedParent` doesn't return the `#document` node, so we need to stop recursing at the `html` element. However we have a test for making sure `#document` is visible, so had to still include the `nodeType` check for `#document`.

WilcoFiers (changes requested):

straker:
The tests are mostly copy paste from the 3 deprecated functions, with a few modifications to convert them to use virtual nodes and removed / added some tests (isVisible had tests removed and move to `isVisibleForScreenreader`)

lib/commons/dom/visibility-functions.js
```diff
@@ -0,0 +1,171 @@
...
+
+const clipRegex = /rect\s*\(([0-9]+)px,?\s*([0-9]+)px,?\s*([0-9]+)px,?\s*([0-9]+)px\s*\)/;
+const clipPathRegex = /(\w+)\((\d+)/;
+
+/**
+ * Determine if an element is natively hidden
+ * @param {VirtualNode} vNode
+ * @return {Boolean}
+ */
+export function nativelyHidden(vNode) {
```
WilcoFiers, line 17: I wonder if we should expand this list. There are more elements that are hidden by default. `param`, `option`, `title`, etc. We could consider putting a property on the standard for this instead. We have other rules that use this kind of list too I believe.

stephenmathieson:
Reminder: commit subjects should be 60 characters or less (not 200+). See https://github.com/dequelabs/dev-process/tree/main/git#subject

lib/commons/dom/visibility-functions.js
straker: I like the idea of adding it to the html element standard

straker: Created https://github.com/dequelabs/axe-core/issues/3641

WilcoFiers (changes requested):
Almost done, I'll let you start working on this.

lib/commons/dom/visibility-functions.js
```diff
@@ -0,0 +1,171 @@
...
+
+/**
+ * Check `AREA` element is visible
+ * - validate if it is a child of `map`
+ * - ensure `map` is referred by `img` using the `usemap` attribute
+ * @param {VirtualNode} vNode
+ * @param {Function} visibleFunction Function used to check if the image element is visible
+ * @retruns {Boolean}
+ */
+export function isAreaVisible(vNode, visibleFunction) {
```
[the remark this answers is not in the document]
straker: For the other functions the things being passed in were options that could change the output so not passing anything was still valid (they were optional), for `isAreaVisible`, the `visibleFunction` is a required parameter in order for the function to run, so I think it should stay a non-object parameter as it is not an option.

lib/commons/dom/is-offscreen.js
```diff
@@ -24,7 +24,13 @@ function noParentScrolled(element, offset) {
  * @param  {Element} element
  * @return {Boolean}
  */
-function isOffscreen(element) {
+function isOffscreen(element, recursed) {
+  if (recursed) {
+    return false;
```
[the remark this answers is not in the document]
straker: It failed a test without it due to this comment https://github.com/dequelabs/axe-core/blob/develop/lib/commons/dom/is-visible.js#L211-L216

> // visibility is only accurate on the first element and
> // position does not matter if it was already calculated

straker:
All comments about changing things outside of the refactor captured here https://github.com/dequelabs/axe-core/issues/3655

test/checks/keyboard/accesskeys.js
```diff
@@ -12,6 +12,7 @@ describe('accesskeys', function() {
 
   it('should return true and record accesskey', function() {
     fixture.innerHTML = '<div id="target" accesskey="A"></div>';
+    fixtureSetup();
     var node = fixture.querySelector('#target');
```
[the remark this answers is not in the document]
straker: acceskeys evaluate is not setup for virtual nodes and so if I did this I would still need to get the actual DOM node which kind of invalidates the whole reason to use `queryFixture`

test/checks/navigation/skip-link.js
```diff
@@ -25,6 +21,7 @@ describe('skip-link', function() {
   it('should return false if the href points to a non-existent element', function() {
     fixture.innerHTML =
       '<a href="#spacecamp">Click Here</a><h1 id="mainheader">Introduction</h1>';
+    axe._tree = axe.utils.getFlattenedTree(fixture);
```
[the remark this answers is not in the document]
straker: I was just mimicing what the rest of the tests did

lib/commons/dom/visibility-functions.js
```diff
@@ -0,0 +1,171 @@
+import {
```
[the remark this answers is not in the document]
straker: We don't make these functions public on `axe` so can't really test them. We could either make them all available like we do with the text methods we mimicked or add them to the `_thisWillBeDeletedDoNotUse` object. 

Since this is just a refactor to make the code more readable, I'm sort of OK if this file isn't tested and we have all the tests in the visibility functions.

straker:
So although I converted the functions to use options, I don't think `memoize` can memoize options. Given the code:

```js
foo = axe.utils.memoize(function(a, { b }) {
  console.log('here', {a, b})
  return true
})

foo(1, {b: 2}) // here {a: 1, b: 2}
foo(1, {b: 2}) // here {a: 1, b: 2}
```

Which shows that the memoized function is not be cached as the options object is not strictly equal for the cache to trigger. We may not be able to do options after all.

lib/commons/dom/visibility-functions.js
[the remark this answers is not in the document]
WilcoFiers: Disagree. I'm uncomfortable when we do it just for pulling a method into its own file. With whole new methods I think we need to make the effort. These are easy tests to write. Let's just get them done.

WilcoFiers (dismissed):

WilcoFiers:
Also remember to open an issue for those questionable uses of isVisible.

.eslintrc.js
```diff
@@ -110,19 +104,14 @@ module.exports = {
...
-      ],
+      files: ['test/aria-practices/**/*.js'],
       env: {
         mocha: true
       }
     },
     {
       files: ['test/**/*.js'],
       excludedFiles: 'test/aria-practices/**/*.js',
-      parserOptions: {
```
straker, line 123: We can now use es6 syntax in tests

WilcoFiers: True, but unrelated to this PR. Please pull this out.

lib/commons/dom/is-offscreen.js
```diff
@@ -35,6 +35,10 @@ function isOffscreen(element, { isAncestor } = {}) {
   element =
     element instanceof AbstractVirtualNode ? element.actualNode : element;
 
+  if (!element) {
+    return undefined;
```
WilcoFiers, line 39: Are  you sure changing the return type doesn't matter here? If so, update the docblock too.

lib/core/core.js
```diff
@@ -78,8 +89,22 @@ axe._thisWillBeDeletedDoNotUse.public = {
...
+  axe._thisWillBeDeletedDoNotUse.commons.dom || {};
+axe._thisWillBeDeletedDoNotUse.commons.dom.nativelyHidden = nativelyHidden;
+axe._thisWillBeDeletedDoNotUse.commons.dom.displayHidden = displayHidden;
+axe._thisWillBeDeletedDoNotUse.commons.dom.visibilityHidden = visibilityHidden;
+axe._thisWillBeDeletedDoNotUse.commons.dom.ariaHidden = ariaHidden;
+axe._thisWillBeDeletedDoNotUse.commons.dom.opacityHidden = opacityHidden;
+axe._thisWillBeDeletedDoNotUse.commons.dom.scrollHidden = scrollHidden;
+axe._thisWillBeDeletedDoNotUse.commons.dom.overflowHidden = overflowHidden;
+axe._thisWillBeDeletedDoNotUse.commons.dom.clipHidden = clipHidden;
+axe._thisWillBeDeletedDoNotUse.commons.dom.areaHidden = areaHidden;
```
WilcoFiers, line 107: This is getting messy. Lets pull this into a separate file.

.eslintrc.js
```diff
@@ -112,9 +112,6 @@ module.exports = {
     {
       files: ['test/**/*.js'],
       excludedFiles: ['test/act-rules/**/*.js', 'test/aria-practices/**/*.js'],
-      parserOptions: {
```
WilcoFiers, line 115: This should go into a separate PR.

WilcoFiers (changes requested):
Couple minor points

lib/core/core.js
straker: I will add this to the tech debt ticket https://github.com/dequelabs/axe-core/issues/3655

WilcoFiers (approved):

Issue #3381, opened 2022-02-04
Rule violation in valid use case (role="gridcell" aria-roledescription="button")


williamernest:
Product: axe-core@4.4

Expectation: No error

Actual:
```
    Violated rule: "aria-roledescription"
    Reasoning: "aria-roledescription must be on elements with a semantic role"
```
Motivation: Tested on various browsers/screen readers and behavior is expected (correctly reads roledescription)
VO/Chrome : Correctly reads roledescription
VO/Firefox: Correctly reads roledescription
VO/Safari: Correctly reads roledescription
NVDA/Chrome: Correctly reads roledescription
NVDA/Firefox: Correctly reads roledescription

https://codepen.io/williamernest/full/GROjddd

WilcoFiers:
Are you seeing this reported as a violation? I am seeing this come up as "needs review". If you're using the axe extension, you can tell from the flag icon.

The reason axe-core flags this is because this will not work in internet explorer with JAWS. Depending on how it's used, missing information from aria-roledescription can cause significant problems. Use of IE is a good bit higher in among screen reader users than it is in the general population, so it's something just to keep an eye out for.

WilcoFiers:
No response. We have done the testing, there are definitely still cases where aria-roledescription in combination with role=gridcell causes issues. Going to leave this as is.

PR #3382, opened 2022-02-04
fix(roledescription): add gridcell as supported role

williamernest:
Adds 'gridcell' as a supported role for aria-roledescription.

Closes issue: #3381

### DIFF
--- a/test/integration/rules/aria-roledescription/aria-roledescription.html
+++ b/test/integration/rules/aria-roledescription/aria-roledescription.html
@@ -15,6 +15,11 @@
 ></div>
 <input type="checkbox" aria-roledescription="my checkbox" id="pass8" />
 <input type="radio" aria-roledescription="my radio" id="pass9" />
+<div role="grid">
+  <div role="gridrow">
+    <div role="gridcell" aria-roledescription="my radio" id="pass10" />
+  </div>
+</div>
 
 <h1 aria-roledescription="my heading" id="incomplete1">heading</h1>
 <div role="rowgroup" aria-roledescription="my row" id="incomplete2"></div>
--- a/doc/check-options.md
+++ b/doc/check-options.md
@@ -191,7 +191,8 @@ All checks allow these global options:
   'radio',
   'combobox',
   'menuitemcheckbox',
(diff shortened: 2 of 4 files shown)


test/integration/rules/aria-roledescription/aria-roledescription.html
```diff
@@ -15,6 +15,11 @@
 ></div>
 <input type="checkbox" aria-roledescription="my checkbox" id="pass8" />
 <input type="radio" aria-roledescription="my radio" id="pass9" />
+<div role="grid">
+  <div role="gridrow">
```
straker, line 19: `gridrow` is not a valid role.

```suggestion
  <div role="row">
```

straker (commented):
Sorry for the delay. We were busying getting the 4.4.0 release of axe and our apis out. Testing with a grid using the `row` role, and support doesn't look great. 

```html
<div role="grid">
  <div role="row">
    <div role="gridcell" aria-roledescription="my radio"/>Hello World</div>
  </div>
</div>
```

Works in:

* JAWS 2020.2012.13 / IE 11 (surprisingly) - "Grid with 1 columns and 1 rows. Hello World my radio"
* JAWS / Chrome - "Grid with 1 columns and 1 rows. Hello World my radio"

Does not work in:
* NVDA / Firefox - "Table with 1 rows and 1 columns. Row 1 column 1. Hello World"

Partial support:
* VO / Safari - "Table 1 column, 1 row. Hello World; You are currently on a my radio, inside of a table." I'd say this is partial support since the "my radio" instructions are not announced unless you have hints turned on. Anyone without hints will not hear anything about the role change. Using table navigation or text navigation does not announce the role without hints.

Due to support I'm not sure this works as well as we would need. @WilcoFiers what do you think?

WilcoFiers:
Thanks for testing Steve. This isn't properly supported yet, so I don't think we can we can make this change.

WilcoFiers:
No response. We have done the testing, there are definitely still cases where aria-roledescription in combination with role=gridcell causes issues. Going to leave this as is.

Issue #3384, opened 2022-02-06
violation message string contains interpolated value, but is verbatim in JSON report


danielweck:
Product: axe-core

Expectation: violation message string interpolation should work

Actual: message appears verbatim, variable not substituted

Motivation: user text is not useful 

<pre><code>
axe-core version: 4.4.1

Any OS, any web browser
</code></pre>

Source reference:

https://github.com/dequelabs/axe-core/blob/f583c70b191513f5bd116a3ed091e937ea2cd65d/lib/checks/aria/deprecatedrole.json#L8

Verbatim message: `The role used is deprecated: ${data.values}`:

```json
    {
        "id": "aria-roles",
        "impact": "minor",
        "tags": [
            "cat.aria",
            "wcag2a",
            "wcag412"
        ],
        "description": "Ensures all elements with a role attribute use a valid value",
        "help": "ARIA roles used must conform to valid values",
        "helpUrl": "https://dequeuniversity.com/rules/axe/4.4/aria-roles?application=axeAPI",
        "nodes": [
            {
                "any": [],
                "all": [],
                "none": [
                    {
                        "id": "deprecatedrole",
                        "data": "doc-endnote",
                        "relatedNodes": [],
                        "impact": "minor",
                        "message": "The role used is deprecated: ${data.values}"
                    }
                ],
                "impact": "minor",
                "html": "<p xmlns=\"http://www.w3.org/1999/xhtml\" role=\"doc-endnote\" xmlns:epub=\"http://www.idpf.org/2007/ops\" epub:type=\"rearnote\">bla bla</p>",
                "target": [
                    "p"
                ],
                "failureSummary": "Fix all of the following:\n  The role used is deprecated: ${data.values}"
            }
        ]
    }
```

To reproduce, create `index.html` and open in web browser + inspect debugger console to see Axe's JSON output:
```html
<?xml version="1.0" encoding="utf-8"?>
<!DOCTYPE HTML>
<html xmlns="http://www.w3.org/1999/xhtml" xmlns:epub="http://www.idpf.org/2007/ops" lang="en" xml:lang="en">
<head>

<title>TITLE</title>

<script src="https://unpkg.com/axe-core@4.4.1/axe.min.js"></script>

</head>
<body>

<section role="doc-chapter">
<p role="doc-endnote" epub:type="rearnote">bla bla</p>
</section>

<script type="javascript">
    window.axe
    .run()
    .then(results => {
[...]

PR #3387, opened 2022-02-07
fix(deprecatedrole,color-contrast): fix message to properly include deprecated role, improve color-contrast pass messages

straker:
Missed during review (my fault) that `data.values` wouldn't work for non-array `data` string. Might try my hand at extending our [check message validation script](https://github.com/dequelabs/axe-core/blob/develop/test/testutils.js#L443) to include verifying that the data object matches all interpolated parts of the string (probably using [process-message](https://github.com/dequelabs/axe-core/blob/develop/lib/core/utils/process-message.js))

Closes issue: #3384


WilcoFiers (changes requested):
I was thinking about this the other day. We don't really have a way to test this, which has caused a few issues. How about if we add something to `testUtils.MockCheckContext` that lets us write tests for these messages?

straker:
Doing this revealed that color-contrast had a few pass cases where the message wasn't passed any data. Fixed those with the test.

WilcoFiers (dismissed):
LGTM, just update the title, since you're not just touching deprecatedrole.

WilcoFiers (approved):

Issue #3315, opened 2021-11-29
a[href][disabled] is still a focusable element


WilcoFiers:
Noticed this in reviewing #3165. In determining if an element is focused, axe-core looks at the disabled attribute. However it doesn't check if that attribute is allowed for the node. So the following link is incorrectly determined as not focusable:

```html
<a href="foo.html" disabled>Hello</a>
```

This bug exists in `lib/commons/dom/focus-disabled.js`. The fix is to check `vNode.props.nodeName` before looking for the `disabled` attribute.

dan-tripp:
I would like to fix this one.  But it might take me a few weeks to get around to it.

WilcoFiers:
@dan-tripp You are welcome to it. Much appreciated!

OmiCoding:
I was looking for issues to fix and I noticed this issue was already fixed in https://github.com/dequelabs/axe-core/commit/bb8b5ca760f45755ba628ac5873350b5dc4c47d1, could a maintainer mark this as closed?

padmavemulapati:
Verified with the latest dev branch code , role=text is allowing to have "tabindex=-1" as a descendants

and with the `nested-interactive/aria-text` allowing "tabindex=-1" on elements when role is not there.
and role conflict is appearing, when `tabindex=-1` and `role=presentation`



![Image](https://user-images.githubusercontent.com/47409579/196704442-a3f14df8-d34b-4cb5-b9a2-73333c90bf51.png)



![Image](https://user-images.githubusercontent.com/47409579/196703826-ac251a67-0d68-4b23-bd26-8383394fb2fd.png)

PR #3393, opened 2022-02-21
fix(checks/no-focusable-disable): don't count non-disableable elements as disabled

dan-tripp:
focusDisabled() now checks the element name to see if "disabled" attribute is valid on that element.

Closes issue #3315


lib/commons/dom/focus-disabled.js
```diff
@@ -2,6 +2,13 @@ import AbstractVirtualNode from '../../core/base/virtual-node/abstract-virtual-n
 import { getNodeFromTree } from '../../core/utils';
 import isHiddenWithCSS from './is-hidden-with-css';
 
+function isDisabledAttrAllowed(nodeName) {
+  // Source: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled 
+  const allowedNodeNames = new Set(["button", "command", "fieldset", "keygen", "optgroup", 
+      "option", "select", "textarea", "input"]);
+  return allowedNodeNames.has(nodeName);
```
straker, line 9: Thanks for listing the source, that was very helpful.

Since the `allowedNodeNames` never changes, we should move the creation of the array outside of the function so it's not recreated every time the function is called. Also `Set` is a bit much and we can just make do with an array using `.includes`.

```suggestion
// Source: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled 
const allowedDisableNodeName = [
  'button', 
  'command', 
  'fieldset', 
  'keygen', 
  'optgroup', 
  'option', 
  'select', 
  'textarea', 
  'input'
];

function isDisabledAttrAllowed(nodeName) {
  return allowedDisableNodeName.includes(nodeName);
}
```

straker (changes requested):
Thanks for taking on the pr. Just have 1 suggestion, but otherwise it looks good.

dan-tripp:
Right on.  Done.

straker (approved):

straker:
reviewed for security

dan-tripp:
Thanks!

Issue #3282, opened 2021-11-12
Minor unsupportedrole check bugs


WilcoFiers:
I don't think this impacts anyone today, since there are currently no unsupported roles listed in axe-core. But I noticed the following bugs.

- The value is not set, even though it's used in the fail message
- dpub roles can not be listed as unsupported, because they are ignored by default in getRole
- a role attribute using a fallback roles will not recognise the unsupported role, even if it's the first (`role="unsupported button"`)

We can easily solve this by replicating what we did in the deprecatedrole check.

WilcoFiers:
This looks fixed.

PR #3395, opened 2022-02-21
fix(rules): unsupportedrole check bugs

dan-tripp:
fix(rules): unsupportedrole check bugs

Closes issue #3282


straker (changes requested):
Thanks for the pr. Overall it looks good, the only request I have is that we'll want to check the `checkContext._data` to ensure that the role is set (since we weren't setting it before).

So for tests that return true, we'll check that the role is set:

```js
assert.deepEqual(checkContext._data, [role]);
```

and for tests that return false, we'll want to check that it's null

```js
assert.isNull(checkContext._data);
```

See https://github.com/dequelabs/axe-core/blob/develop/test/checks/aria/deprecatedrole.js for examples of doing this.

dan-tripp:
That makes sense.  I like axe-core's test philosophy.  Does my latest push reflect what you have in mind?

straker (approved):
Yep! Looks great.

dan-tripp:
Nice.  Thanks again for the guidance.

Issue #3367, opened 2022-01-28
Fix flakey Firefox tests


straker:
Firefox has been failing recently. Need to investigate why.

https://app.circleci.com/pipelines/github/dequelabs/axe-core/3335/workflows/8b6a5416-3696-476b-98c6-75471285f2fe/jobs/40638

straker:
https://app.circleci.com/pipelines/github/dequelabs/axe-core/3351/workflows/ac85d186-103a-458f-a6ee-decf00724ad5/jobs/40766

straker:
https://app.circleci.com/pipelines/github/dequelabs/axe-core/3360/workflows/4191fc9d-3df7-4cb5-b64e-5c61254df2b5/jobs/40838

straker:
https://app.circleci.com/pipelines/github/dequelabs/axe-core/3364/workflows/f7ad725f-ae47-445a-9eb1-acade592c9a2/jobs/40886

straker:
https://app.circleci.com/pipelines/github/dequelabs/axe-core/3437/workflows/44bf0328-aaf6-40c6-8be9-21ffadf2f7a6/jobs/41588

straker:
This seems to be resolved.

PR #3398, opened 2022-02-25
tests: resolve isolated-env race condition

straker:
Turns out there was a race condition in the isolated-env tests between when the iframe `load` event was triggered and when the script contents to fetch axe and inject it into the page via `eval` would complete. Tests would fail if the script couldn't execute faster than the `load` event.

Fixed that by properly waiting for the post message the script sends in addition to waiting for the nested load function.

Closes issue: #3367


WilcoFiers (approved):
LGTM; I did change your PR title though.

PR #3402, opened 2022-03-04
fix(html-elms): update role allowances for nav element

### LINKED ISSUE #3401: Allow role=none|presentation on nav element
the `nav` element was the only native landmark element that didn't allow the explicit use of `role=none|presentation` so developers could mitigate prior author errors.

[ARIA in HTML](https://github.com/w3c/html-aria/pull/404) is being updated to allow authors to specify these roles on the `nav` element.

scottaohara:
closes #3401

allows `none` and `presentation` role on `<nav>` element.

### DIFF
--- a/lib/standards/html-elms.js
+++ b/lib/standards/html-elms.js
@@ -633,6 +633,8 @@ const htmlElms = {
       'doc-toc',
       'menu',
       'menubar',
+			'none',
+			'presentation',
       'tablist'
     ],
     shadowRoot: true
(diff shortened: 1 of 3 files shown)


WilcoFiers (changes requested):
Spacing thing, and then would you mind adding a test case to test/integration/rules/aria-allowed-role/aria-allowed-role.html, so that we can prevent regression.

Also, as a question, how come `<main>` isn't allowed role=none, but footer, header, aside and now nav are?

WilcoFiers (approved):

WilcoFiers:
Reviewed for security. I'm going to merge this despite the failing tests. Those are unrelated to this PR. A fix for that is merging into develop today too.

PR #3403, opened 2022-03-04
ci(act): Add nightly job to run latest ACT rules

WilcoFiers:
We had the ACT job fail because of an update to one of the rules. This PR makes it so we run tests against a pinned version of ACT. We'll have a nightly job run against the latest ACT rules so we can see if something in ACT broke our implementation.

I haven't looked at what was failing specifically. I'll do that in a separate PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -119,7 +119,6 @@ jobs:
     steps:
       - checkout
       - <<: *restore_dependency_cache_unix
-      # Always run on the latest master branch, regardless of cache
       - run: npm run build
       - run: npm run test:act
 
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -120,7 +120,6 @@ jobs:
       - checkout
       - <<: *restore_dependency_cache_unix
       # Always run on the latest master branch, regardless of cache
```
straker, line 122:
```suggestion
```

straker (changes requested):

straker (approved):

Issue #3419, opened 2022-03-22
Add note to invalid IdRef value error that it may be due to a Shadow DOM boundary


joppekroon:
Product: axe-core

Expectation: Indicate that the referenced id is located beyond a Shadow DOM boundary. Even a message that states that this may be potentially the reason of the error would be very helpful.

Actual: Indicates that the value of the IdRef attribute is invalid

```
    "ARIA attributes must conform to valid values (aria-valid-attr-value)"

    Fix all of the following:
      Invalid ARIA attribute value: aria-controls="id-in-shadow"
```

Motivation: It is currently quite a confusing error message if you're not acutely aware that referencing an id that is located on the other side of a Shadow DOM boundary is problematic. 

<pre><code>
axe-core version: 4.2.0
</code></pre>

padmavemulapati:
Validated with the latest develop branch axe-core code base, 
test snippet is:
```
<div></div>
  <button id="target1" aria-describedby="test1">Button1</button>
  <div></div>
  <button id="target2" aria-labelledby="test2">Button2</button>
```
so that the failure message now is:
<img width="855" alt="image" src="https://user-images.githubusercontent.com/47409579/175059066-74f22876-5e53-4469-bf43-01cd8d2a9633.png">

PR #3421, opened 2022-03-22
fix(aria-valid-attr-value): add note about element id being in a different shadow DOM tree

straker:
Closes issue: https://github.com/dequelabs/axe-core/issues/3419

### DIFF
--- a/lib/checks/aria/aria-valid-attr-value-evaluate.js
+++ b/lib/checks/aria/aria-valid-attr-value-evaluate.js
@@ -68,7 +68,8 @@ function ariaValidAttrValueEvaluate(node, options, virtualNode) {
           'aria-describedby'
         )}"`;
         // TODO: es-modules_tree
-        messageKey = axe._tree[0]._hasShadowRoot ? 'noIdShadow' : 'noId';
+        messageKey =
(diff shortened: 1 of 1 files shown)


lib/checks/aria/aria-valid-attr-value.json
```diff
@@ -11,7 +11,7 @@
         "plural": "Invalid ARIA attribute values: ${data.values}"
       },
       "incomplete": {
-        "noId": "ARIA attribute element ID does not exist on the page: ${data.needsReview}",
+        "noId": "ARIA attribute element ID does not exist on the page or is a descendant of a different shadow DOM tree: ${data.needsReview}",
```
WilcoFiers, line 14: I'm not sure about this. Could we test that shadow DOM exists on the page before we include this? I don't think we should tell people that shadow DOM might be the problem if we they have no shadow DOM on the page.

WilcoFiers (changes requested):
See comment.

lib/core/utils/get-flattened-tree.js
```diff
@@ -20,6 +20,8 @@ import cache from '../base/cache';
  *    the spec for this)
  */
 
+let hasShadowRoot;
```
WilcoFiers, line 23: Little hackish... but alright.

WilcoFiers (approved):

Issue #3252, opened 2021-11-02
Documentation for `aria-hidden-focus` for single element is misleading and not fully correct


kospl:
Product: axe-core

Expectation: `<input aria-hidden="true">` failing test for `aria-hidden-focus` and having clear help message explaining why this element is failing.

Actual: axe core 4.3.5 reports `aria-hidden-focus` violation with description about its elements (which don't exist).

Motivation: misleading help text made some tech support to think there is false positive with axe rule, because it's mentioning `do not contain elements`, and there are just one single DOM element, with no other elements that it contains that may be failing this test.
___

I know it sounds a little bit nerdy, but such help text have to be clear, as it's getting even worse when it's got to Lighthouse: https://github.com/GoogleChrome/lighthouse/blob/8a29854072202807ba54276beb9cdd9f244007b5/lighthouse-core/audits/accessibility/aria-hidden-focus.js#L18

__

minimal repro example:

    <input aria-hidden="true">

axe help text: "ARIA hidden element must not contain focusable elements"
Lighthouse title: "`[aria-hidden="true"]` elements do not contain focusable descendents"

straker:
Thanks for the issue. We'll try to update the message to make it more clear.

TechThomas:
Hey @straker how is the process going on improving the messaging. We actually just tweeted about this issue https://twitter.com/TechThomas/status/1488413276400922624

straker:
We weren't able to get this into the 4.4.0 release which just went out, but we can try to get it added to 4.4.1.

TechThomas:
@straker when will 4.4.1 be expected?

straker:
Sorry, 4.4.1 was released not too long ago but did not include this fix (it was a quick release to fix a bug in 4.4.0). I'm not sure when we plan on releasing the next release, so I'm unable able to say when this will be addressed.

padmavemulapati:
Validated with the latest axe-core, develop branch `axe.js` source, 
It was showing in 4.3.5 , for `aria-hidden-focus` rule failing description is ""Ensures aria-hidden elements do not contain focusable elements" , below is the screenshot

<img width="725" alt="image" src="https://user-images.githubusercontent.com/47409579/169813490-40e85eaf-da38-4f3e-9808-847b61a15123.png">

and now it got fixed and the description is appearing as "Ensures aria-hidden elements are not focusable nor contain focusable elements"
[...]

PR #3422, opened 2022-03-24
fix(aria-hidden-focus): Update rule help & description

mjawili:
The current `help` value does not include the scenario in which `aria-hidden="true"` is on a focusable element. I updated it to include that scenario. Here's the PR, per @WilcoFiers suggestion. Slack thread: https://deque.slack.com/archives/C0LA5E545/p1646762357136919

### DIFF
--- a/lib/rules/aria-hidden-focus.json
+++ b/lib/rules/aria-hidden-focus.json
@@ -7,7 +7,7 @@
   "actIds": ["6cfa84"],
   "metadata": {
     "description": "Ensures aria-hidden elements do not contain focusable elements",
-    "help": "ARIA hidden element must not contain focusable elements"
+    "help": "ARIA hidden element must not be focusable or contain focusable elements"
   },
   "all": [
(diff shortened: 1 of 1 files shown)


lib/rules/aria-hidden-focus.json
```diff
@@ -7,7 +7,7 @@
   "actIds": ["6cfa84"],
   "metadata": {
     "description": "Ensures aria-hidden elements do not contain focusable elements",
```
WilcoFiers, line 9:
```suggestion
    "description": "Ensures aria-hidden elements are not focusable nor contain focusable elements",
```

straker (approved):
Thanks for the PR!

straker:
Reviewed for security

straker:
Ps. this pr closes issue https://github.com/dequelabs/axe-core/issues/3252

mjawili:
@straker that's good to know!

PR #3423, opened 2022-03-24
fix(utils): greatly improve the speed of our querySelectorAll code

straker:
Following the QSA proposal, this greatly improves the speed of QSA as it no longer has to scan the entire DOM tree for top-level queries. For large sites the speed increase is substantial.

For https://js.tensorflow.org/api/latest/, the top slowest performanceTimer results were as follow:

Timer Name | Time (ms)
-- | --
rule_heading-order#gather | 136.40
rule_empty-heading#gather | 137.20
rule_empty-heading | 137.70
rule_list | 139.30
rule_aria-command-name#gather | 140.40
rule_aria-command-name | 140.70
rule_aria-allowed-attr#gather | 148.30
runchecks_heading-order | 151.10
runchecks_page-has-heading-one | 151.30
rule_page-has-heading-one | 151.40
rule_identical-links-same-purpose#gather | 172.60
runchecks_link-name | 197.30
rule_valid-lang#gather | 197.70
rule_valid-lang | 198.10
rule_landmark-unique#gather | 208.90
rule_landmark-unique | 210.80
rule_aria-allowed-attr | 216.40
rule_identical-links-same-purpose#matches | 273.50
rule_link-name | 278.60
rule_heading-order | 287.60
runchecks_bypass | 338.80
rule_bypass | 405.80
rule_identical-links-same-purpose | 575.70
runchecks_region | 871.80
rule_color-contrast#matches | 967.10
rule_region | 997.40
runchecks_color-contrast | 11,530.40
rule_color-contrast | 12,497.60
audit_start_to_end | 21,541.50

We can see that the `#gather` step is what causes most rules to be slow (and not so much the `#runchecks` step), causing axe-core to take 20 seconds to just run through all the rules.

By caching node information ahead of time, the top slowest results are now as follows:

Timer Name | Time (ms)
-- | --
rule_page-has-heading-one | 31.90
runchecks_aria-hidden-body | 31.90
rule_aria-hidden-body | 32.00
rule_autocomplete-valid | 36.60
rule_aria-valid-attr-value | 36.70
rule_aria-valid-attr | 38.80
runchecks_list | 52.90
rule_list | 53.90
rule_scrollable-region-focusable#matches | 69.00
rule_nested-interactive#matches | 71.30
runchecks_bypass | 75.30
rule_bypass | 75.50
rule_aria-allowed-attr#gather_axe.utils.isHidden | 81.50
rule_nested-interactive | 91.80
rule_aria-allowed-attr#gather | 94.70
rule_scrollable-region-focusable | 99.20
rule_region#gather | 102.60
rule_aria-allowed-attr | 115.80
runchecks_identical-links-same-purpose | 121.90
rule_identical-links-same-purpose#matches | 181.30
runchecks_link-name | 186.10
rule_link-name | 193.50
rule_identical-links-same-purpose | 308.60
runchecks_region | 806.00
rule_region | 914.70
rule_color-contrast#matches | 1,054.00
runchecks_color-contrast | 11,780.90
rule_color-contrast | 12,835.00
audit_start_to_end | 15,093.50
axe | 16,805.50

This change shaves off about 7-10 seconds from the time needed to run all of axe rules on the site (`audit_start_to_end`).

As the page size increases, so too does the savings. Testing an extremely large site (152,000 nodes), this is the final time difference:

Timer Name | Normal Run Time (ms) | Cache Run Time (ms)
-- | -- | --
audit_start_to_end | 62,128.60 | 45,251.90
axe | 75,526.50 | 60,693.90

This is about a 20 second savings. On medium size pages, the savings is minimal (almost no difference).

To ensure the results are the same, I ran axe-core against the tensorflow site, recorded the results, then ran the caching verison of axe core and compared the two results for equality. Here's the code I used. 

```js
const { AxePuppeteer } = require('@axe-core/puppeteer');
const puppeteer = require('puppeteer');
const fs = require('fs');
const chai = require('chai');
const processArgs = require('./process-args');

const { assert } = chai;
const args = processArgs(process.argv.slice(2));

if (!args.url) {
  throw new Error('Must include a url to test');
}
if (!args.numTests) {
  args.numTests = 1;
}

function assertShallowEqual(obj1, obj2, parentKey = '') {
  let keys;
  if (Array.isArray(obj1)) {
    keys = obj1.map((value, index) => index);

  }
  else {
    keys = Object.keys(obj1);
  }

  keys.forEach(key => {
    const valueObj1 = obj1[key];
    const valueObj2 = obj2[key];

    // don't compare html output, just selector
    if (key === 'html') return;

    // don't compare url props (can contain date-time...)
    if (key === 'urlProps') return;

    // don't compare axe testEngine since they are named differently
    // to ensure correct tests
    if (key === 'testEngine') return;

    // don't look for null
    if (valueObj1) {
      assert.exists(valueObj2, `${key} does not exist on ${parentKey}`);
    }

    if (valueObj1 && (Array.isArray(valueObj1) || typeof valueObj1 === 'object')) {
      assertShallowEqual(valueObj1, valueObj2, `${parentKey}.${key}`);
    }
    else {
      assert.equal(valueObj1, valueObj2, `${parentKey}.${key} not equal`);
    }
  });
}

(async () => {
  const browser = await puppeteer.launch();
  const page = await browser.newPage();
  await page.setBypassCSP(true);

  const axe = new AxePuppeteer(page);
  const results = [];

  const axeCorePath = require.resolve('axe-core');
  const origAxeCore = fs.readFileSync(axeCorePath, 'utf8');
  const devAxeCore = fs.readFileSync('./axe-develop.js', 'utf8');
  const perfAxeCore = fs.readFileSync('./axe-perf.js', 'utf8');

  fs.rmSync('./control', { recursive: true, force: true });
  fs.rmSync('./test', { recursive: true, force: true });

  fs.mkdirSync('./control');
  fs.mkdirSync('./test');

  async function runAxe(dir) {
    for (let i = 0; i < args.numTests; i++) {
      page.goto(args.url);

      // wait for content to settle
      await new Promise((resolve) => {
        setTimeout(resolve, 5000)
      });

      const result = await axe.analyze();
      delete result.timestamp;
      fs.writeFileSync(`./${dir}/result-${i}.json`, JSON.stringify(result, null, 2), 'utf8');
      results.push(result);
    }
  }

  fs.writeFileSync(axeCorePath, devAxeCore, 'utf8');
  await runAxe('control');
  fs.writeFileSync(axeCorePath, perfAxeCore, 'utf8');
  await runAxe('test');
  fs.writeFileSync(axeCorePath, origAxeCore, 'utf8');

  try {
    for (let i = 0; i < args.numTests; i++) {
      assertShallowEqual(results[i], results[i+1]);
    }
  } finally {
    await page.close();
    await browser.close();
  }
})();
```

### DIFF
--- a/lib/core/utils/matches.js
+++ b/lib/core/utils/matches.js
@@ -224,6 +224,10 @@ export function convertSelector(selector) {
  * @returns {Boolean}
  */
 function optimizedMatchesExpression(vNode, expressions, index, matchAnyParent) {
+  if (!vNode) {
+    return false;
+  }
+
   const isArray = Array.isArray(expressions);
   const expression = isArray ? expressions[index] : expressions;
   let matches = matchExpression(vNode, expression);
--- a/lib/core/utils/matches.js
+++ b/lib/core/utils/matches.js
@@ -128,6 +128,7 @@ function convertAttributes(atts) {
     return {
       key: attributeKey,
       value: attributeValue,
+      type: typeof att.value === 'undefined' ? 'attrExist' : 'attrValue',
       test: test
     };
   });
--- a/lib/core/utils/selector-cache.js
+++ b/lib/core/utils/selector-cache.js
@@ -1,10 +1,29 @@
 import { convertSelector, matchesExpression } from './matches';
+import tokenList from './token-list';
 
+// since attribute names can't contain whitespace, this will be
+// a reserved list for ids
+const idsKey = ' [idsMap]';
 let selectorMap = {};
 
-function cacheSelector(key, vNode) {
-  selectorMap[key] = selectorMap[key] || [];
-  selectorMap[key].push(vNode);
+function isGlobalSelector(exp) {
+  return exp.tag === '*' && !exp.attributes && !exp.id && !exp.classes;
+}
+
+function cacheSelector(key, vNode, map = selectorMap) {
+  map[key] = map[key] || [];
+  map[key].push(vNode);
+}
+
+/**
+ * Inner join two arrays.
+ */
+function innerJoin(a, b) {
+  if (!b.length) {
+    return a;
+  }
+
+  return a.filter(node => b.includes(node));
 }
 
 /**
@@ -26,6 +45,14 @@ export function cacheNodeSelectors(vNode) {
   cacheSelector('*', vNode);
 
   vNode.attrNames.forEach(attrName => {
+    // element ids are the only values we'll match
+    if (attrName === 'id') {
+      selectorMap[idsKey] = selectorMap[idsKey] || {};
+      tokenList(vNode.attr(attrName)).forEach(value => {
+        cacheSelector(value, vNode, selectorMap[idsKey]);
+      });
+    }
+
     cacheSelector(`[${attrName}]`, vNode);
   });
 }
@@ -47,6 +74,19 @@ export function getNodesMatchingSelector(domTree, selector, filter) {
 
   const shadowId = domTree[0].shadowId;
   const expressions = convertSelector(selector);
+
+  // if the selector uses a global selector with a combinator
+  // (A *, A > *) it's actually faster to use our QSA code than
+  // getting all nodes and using matchesExpression
+  for (let i = 0; i < expressions.length; i++) {
+    if (
+      expressions[i].length > 1 &&
+      expressions[i].some(expression => isGlobalSelector(expression))
+    ) {
+      return;
+    }
+  }
+
   let matchedNodes = [];
 
   // find nodes that match just a part of the selector in order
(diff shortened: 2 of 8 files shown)


test/core/utils/qsa.js
```diff
@@ -57,190 +57,240 @@ describe('axe.utils.querySelectorAllFilter', function() {
...
-    assert.equal(result[1].actualNode.nodeName, 'DIV');
-    assert.equal(result[2].actualNode.nodeName, 'UL');
-  });
-  it('should filter the returned nodes when passed a filter function', function() {
-    var result = axe.utils.querySelectorAllFilter(dom, 'ul, #one', function(
-      node
-    ) {
-      return node.actualNode.nodeName !== 'UL';
+
+  var tests = ['without cache', 'with cache'];
```
straker, line 61: The first 30 lines are new code to the file, the rest is whitespace difference.

test/core/utils/qsa.js
```diff
@@ -57,190 +57,240 @@ describe('axe.utils.querySelectorAllFilter', function() {
...
+      it('should find nodes using id', function() {
+        var result = axe.utils.querySelectorAllFilter(dom, '#one');
+        assert.equal(result.length, 1);
+      });
+
+      // can only select shadow dom nodes when we're not using the
+      // top-level node. but since the top-level node is the one
+      // with the cache, this only works when we are testing the full
+      // tree (i.e. without cache)
+      if (describeName === 'without cache') {
```
straker, line 144: This wrapping `if` is also new to handle the test for shadowDOM id querying (which only works when not using the cache for this particular test).

lib/core/utils/matches.js
```diff
@@ -128,6 +128,7 @@ function convertAttributes(atts) {
     return {
       key: attributeKey,
       value: attributeValue,
+      type: typeof att.value === 'undefined' ? 'attrExist' : 'attrValue',
```
straker, line 131: This was added to tell the difference between the selectors `[href]` and `[href=""]` (both returned the same structure of `value: ''` even though one is only checking existence of the attribute and the other is checking for an empty string).

lib/core/utils/selector-cache.js
```diff
@@ -0,0 +1,196 @@
...
+}
+
+/**
+ * Get nodes from the selector cache that match the selector.
+ * @param {VirtualTree[]} domTree flattened tree collection to search
+ * @param {Object} expressions
+ * @param {Function} filter function (optional)
+ * @return {Mixed} Array of nodes that match the selector or undefined if the selector map is not setup
+ */
+export function getNodesMatchingExpression(domTree, expressions, filter) {
```
WilcoFiers, line 81: Can you put this method at the top of the file.

lib/core/utils/selector-cache.js
```diff
@@ -0,0 +1,196 @@
...
+    ) {
+      return;
+    }
+  }
+
+  const nodeSet = new Set();
+
+  // find nodes that match just a part of the selector in order
+  // to speed up traversing the entire tree
+  expressions.forEach(expression => {
```
WilcoFiers, line 107: I think the content of this callback should be put into a separate function.

lib/core/utils/selector-cache.js
```diff
@@ -0,0 +1,196 @@
...
+      nodeSet.add(node);
+    });
+  });
+
+  // ie11 does not work with Array.from without a polyfill (missing
+  // `.entries`) but does have forEach. using a set saves time when
+  // wanting to make sure the same node isn't added twice (~3 seconds
+  // total over using array.includes)
+  let matchedNodes = [];
+  nodeSet.forEach(node => matchedNodes.push(node));
```
WilcoFiers, line 189: If you're going to return an array, why bother with a set? Wouldn't it be easier to just use an array and then filter it for unique values before you return?

lib/core/utils/selector-cache.js
```diff
@@ -0,0 +1,196 @@
+import { matchesExpression } from './matches';
+import tokenList from './token-list';
+
+// since attribute names can't contain whitespace, this will be
+// a reserved list for ids so we can perform virtual id lookups
+const idsKey = ' [idsMap]';
+let selectorMap = {};
```
WilcoFiers, line 7: Do we need to put this in the axe-core cache, so that it can be cleaned up after a run?

lib/core/utils/selector-cache.js
```diff
@@ -0,0 +1,196 @@
...
+ * Get nodes from the selector cache that match the selector.
+ * @param {VirtualTree[]} domTree flattened tree collection to search
+ * @param {Object} expressions
+ * @param {Function} filter function (optional)
+ * @return {Mixed} Array of nodes that match the selector or undefined if the selector map is not setup
+ */
+export function getNodesMatchingExpression(domTree, expressions, filter) {
+  // check to see if the domTree is the root and has the selector
+  // map. if not we just return and let our QSA code do the finding
+  const selectorMap = domTree[0]._selectorMap;
```
WilcoFiers, line 84: It's a little confusing you're reusing this variable name. You now have this name both in the function scope and in the file scope.

test/core/utils/selector-cache.js
```diff
@@ -0,0 +1,242 @@
...
+
+    it('should return nodes for each expression', function() {
+      fixture.innerHTML =
+        '<div role="button" aria-label="other"></div><span id="foo"></span>';
+      var nodes = createTree();
+      var expression = convertSelector('[role], [id]');
+      assert.deepEqual(getNodesMatchingExpression(tree, expression), nodes);
+    });
+
+    it('should return nodes for child combinator selector', function() {
```
WilcoFiers, line 149: This should probably also test indirect descendants.

WilcoFiers (dismissed):

lib/core/utils/selector-cache.js
straker: I attached the selector cache to the root node so once the tree is cleaned up after a run the cache goes away as well.

lib/core/utils/selector-cache.js
```diff
@@ -0,0 +1,196 @@
...
+const idsKey = ' [idsMap]';
+let selectorMap = {};
+
+/**
+ * Non-tag selectors use `*` for the tag name so a global selector won't have any other properties of the expression.
+ * @param {Object} exp Selector Expression
+ * @returns {Boolean}
+ */
+function isGlobalSelector(exp) {
+  return exp.tag === '*' && !exp.attributes && !exp.id && !exp.classes;
```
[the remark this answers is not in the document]
straker: Since we don't cache the content of a pseudo selector, they are still considered a global selector such that `*:not([foo])` will use the global selector cache and then `matchesExpression` to match the pseudo selector.

lib/core/utils/selector-cache.js
```diff
@@ -0,0 +1,196 @@
...
+
+  const nodeSet = new Set();
+
+  // find nodes that match just a part of the selector in order
+  // to speed up traversing the entire tree
+  expressions.forEach(expression => {
+    // use the last part of the expression to find nodes as it's more
+    // specific. e.g. for `body h1` use `h1` and not `body`
+    const exp = expression[expression.length - 1];
+    let nodes = [];
```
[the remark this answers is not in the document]
straker: Good catch

lib/core/utils/selector-cache.js
straker: Trying to make it part of the cache interfered with other tests as creating any virtual node adds the node to the cache. The cache is attached to the tree so it at least is clean up that way, but the object will continue to be around after a run.

WilcoFiers (changes requested):

lib/core/utils/get-flattened-tree.js
```diff
@@ -38,6 +39,20 @@ function getSlotChildren(node) {
...
+/**
+ * Create a virtual node
+ * @param {Node} node the current node
+ * @param {VirtualNode} parent the parent VirtualNode
+ * @param {String} shadowId, optional ID of the shadow DOM that is the closest shadow ancestor of the node
+ * @return {VirtualNode}
+ */
+function createNode(node, parent, shadowId) {
+  const vNode = new VirtualNode(node, parent, shadowId);
+  cacheNodeSelectors(vNode, cache.get('selectorMap'));
```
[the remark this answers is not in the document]
straker: It messed up a few tests when having the cache of the node set and when tests would create a VirtualNode that wasn't part of the tree itself. I think it's better that this only cache the selector when in the flatten tree script since it's specifically caching the information about the tree (and not one off vNodes).

lib/core/utils/selector-cache.js
```diff
@@ -0,0 +1,210 @@
...
+    }
+  }
+
+  // it turned out to be more performant to use a Set to generate a
+  // unique list of nodes rather than an array and array.includes
+  // (~3 seconds total on a benchmark site)
+  const nodeSet = new Set();
+
+  expressions.forEach(expression =>
+    findMatchingNodes(expression, selectorMap, shadowId, nodeSet)
```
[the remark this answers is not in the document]
straker: It at least needs to be passed in since each expression needs to share the same Set in order to remove duplicates.

lib/core/utils/selector-cache.js
```diff
@@ -0,0 +1,210 @@
...
+ * @param {Object} expression Selector Expression
+ * @param {Object} selectorMap Selector map cache
+ * @param {String} shadowId ShadowID of the root node
+ * @param {Set} nodeSet
+ */
+function findMatchingNodes(expression, selectorMap, shadowId, nodeSet) {
+  // use the last part of the expression to find nodes as it's more
+  // specific. e.g. for `body h1` use `h1` and not `body`
+  const exp = expression[expression.length - 1];
+  let nodes = [];
```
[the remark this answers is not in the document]
straker: Since the function no longer returns the empty array if passed in and only returns nodes that are in both arrays, it fixed the issue at the same time.

At step 3 when nothing is found, the empty array is passed to step 4, but since there are no nodes in the array the script won't add any nodes from the `.foo` array.

```
function getSharedValues(a, b) {
  return a.filter(node => b.includes(node));
}

getSharedValues([], ['a', 'b', 'c']);  // []
getSharedValues(['a', 'b', 'c'], []);  // []
```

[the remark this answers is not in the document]
WilcoFiers: I don't think it is....? If somewhere along this algorithm you end up with the array getting emptied again, because of the ternary, you're not calling getSharedValues. You're doing `nodes = cachedNodes`, effectively.

This doesn't feel like a safe algorithm. It's hard to track. If you want this setup, have `nodes` be `null` until it is first assigned, and have your ternary be:

```js
 nodes = nodes === null ? cachedNodes : getSharedValues(nodes, cachedNodes);
```

WilcoFiers (changes requested):

lib/core/utils/selector-cache.js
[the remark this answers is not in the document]
straker: I see what you mean. Added a test to catch this problem now.

lib/core/utils/matches.js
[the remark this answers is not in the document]
straker: Currently the `convertSelector` API does not get exported to a public axe function and is only used internally. We could add it to the `_thisWillBeDeletedDoNotUse` test api, but I think adding a suite of tests for it should be a separate pr.

WilcoFiers (approved):

straker:
Tested latest changes using my compare script, tensorflow.org passed, and so did a few other sites.

Issue #3404, opened 2022-03-04
Empty table headers are flagged as a failure of WCAG 1.3.1


fstrr:
Product: axe Extension

Expectation:
I don't think an empty table header (e.g. `<th></th>`) is a failure of WCAG 1.3.1 Info And Relationships, which is what Axe flags and tags this code as. It's definitely best practice to have text in a table heading, but I don't think this is a WCAG failure. 1.3.1 says:

>"Information, structure, and relationships conveyed through presentation can be programmatically determined or are available in text."

If a header cell is marked up as a header cell, is in the same same row as all the other header cell, and styled differently from non-header cells, I'd argue that the presentational information, structure, and relationships are programmatically available. There's no text, so there's nothing to present visually or programmatically.

I'd expect to see this flagged as Best Practice rather than a WCAG conformance issue.

Actual:
This example is flagged in Axe with an issue tag of "wcag131":

```html
<table>
  <tr>
   <th>First Name</th>
   <th>Last Name</th>
   <th>Email Address</th>
   <th></th>
  </tr>
  <tr>
    <td>Sarah</td>
    <td>Smith</td>
    <td>sarah.smith@example.com</td>
    <td><button type="button">Edit</button><button type="button">Delete</button></td>
 </tr>
</table>
```
Motivation:
I don't believe that this is a WCAG failure, so would like to see this moved to a best practice recommendation.

<pre><code>
axe-core version: 4.4.1
Browser: Chrome 98
</code></pre>

straker:
Thanks for the issue. So the reason we flag empty table headers as a failure of WCAG 1.3.1 is because the ARIA spec requires `columnheader` and `rowheader` [to have an accessible name](https://www.w3.org/TR/wai-aria-1.2/#columnheader). Because it's required, it can't be a best practice.

straker:
In this particular case, to fix the issue you could replace the empty `th` with an empty `td` since there is no header for the column.

straker:
So when mapping HTML elements to ARIA roles, the [ARIA in HTML spec](https://w3c.github.io/html-aria/#el-th) is used to determine the implicit AIRA role of the element.

As for the criterion, I think @WilcoFiers will have to answer that one.

fstrr:
[...]

PR #3427, opened 2022-03-29
feat(empty-table-headers): rule is now best-practice and fails instead of incompletes

fstrr:
References #3404

Updated empty table headers rule to be best practice rather than a WCAG failure.

Closes issue:

### DIFF
--- a/lib/rules/empty-table-header.json
+++ b/lib/rules/empty-table-header.json
@@ -2,7 +2,6 @@
   "id": "empty-table-header",
   "selector": "th, [role=\"rowheader\"], [role=\"columnheader\"]",
   "tags": ["cat.name-role-value", "best-practice"],
-  "impact": "minor",
   "metadata": {
     "description": "Ensures table headers have discernible text",
     "help": "Table header text should not be empty"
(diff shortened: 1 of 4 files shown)


straker:
Thanks for the pr. The tests are failing because `reviewOnFailed` was removed, which is causing the integration tests to fail (it was looking for incomplete before). 

The way the integration tests work is they use [a json file](https://github.com/dequelabs/axe-core/blob/develop/test/integration/rules/empty-table-header/empty-table-header.json), which is paired with [an html file](https://github.com/dequelabs/axe-core/blob/develop/test/integration/rules/empty-table-header/empty-table-header.htm). The json file lists the expected output of html elements on the html file (listed by the element ids). In this case, we expect there to be 3 incomplete results.

To fix the tests, we'll just need to update the json file to use `violations` instead of `incomplete` and update the id names for each of the elements (we typically use `pass`, `fail`, and `incomplete` for the id names).  Something like the following:

```js
{
  "description": "empty-table-header tests",
  "rule": "empty-table-header",
  "violations": [["#fail1"], ["#fail2"], ["#fail3"]],
  "passes": [["#pass1"], ["#pass2"], ["#pass3"]]
}
```

WilcoFiers:
Reviewed for security

Issue #3440, opened 2022-04-22
add help text to link-in-text-block rule


shanisebarona:
Add help text with more context to link-in-text-block rule. 

Suggestion: Sufficient color contrast ratio of 3:1 is necessary to distinguish link text color from surrounding text color.

WilcoFiers:
Thank you for the suggestion. I think this rule need some work TBH. The W3C took the position a while back that a 3:1 contrast between the link and it surrounding text would be sufficient. That's not how we understood this criterion when we wrote the rule. Now that a 3:1 contrast ratio is explicitly permitted I think the rule needs some revision.

Out of curiosity, this rule is experimental and so isn't on by default. What are you running axe-core with that has this rule enabled?

shanisebarona:
The rule is currently in Accessibility for Insights for Web as a "needs review" check.

DaveTryon:
@WilcoFiers, as discussed offline, we're picking this up with the following proposed approach:
1. Replace the current link-in-text check with 4 pseudoClass checks--in each case, fall back to the non-pseudoClass colors if colors can't be found in the pseudoClass
  - `link-in-text-block-link` mirrors `link-in-text-block` but checks the `link` pseudoClass
  - `link-in-text-block-hover` mirrors `link-in-text-block` but checks the `hover` pseudoClass
  - `link-in-text-block-visited` mirrors `link-in-text-block` but checks the `visited` pseudoClass
  - `link-in-text-block-active` mirrors `link-in-text-block` but checks the `active` pseudoClass
2. Add the new checks to 'all' array of the `link-in-text-block` rule
3. Update tests appropriately

The net result should be that the `link-in-text-block` rule is able to account for each of the pseudoClass types, then provide a meaningful answer.

Two observations on this approach:
1. Looking at https://www.w3.org/WAI/WCAG21/Understanding/use-of-color, a 3:1 contrast ratio ([G183](https://www.w3.org/WAI/WCAG21/Techniques/general/G183)) is just 1 of 4 possible ways to satisfy the requirement. [G14](https://www.w3.org/WAI/WCAG21/Techniques/general/G14) would satisfy the requirement equally well. If someone were to adopt [G14](https://www.w3.org/WAI/WCAG21/Techniques/general/G14, this rule would essentially be a false positive.
[...]

PR #3441, opened 2022-04-22
add help text

shanisebarona:
This PR adds more context to the help text for the `link-in-text-block` rule. 

Closes issue: #3440


WilcoFiers:
My apologies, this one slipped through the cracks. I'm afraid this isn't going to work. The rule itself doesn't correctly account for the 3:1 contrast ratio. Or rather, the way the rule works today differs from how the W3C has recently begun to explain this. We'll need to make change to the implementation of the rule, not just its message. Since this rule is experimental, addressing this also doesn't have a particularly high priority.

Thank you for the contribution, but I don't think we can use this. I'm going to close this PR for now. If you're interested in helping us make the code changes please ping me. It's a fairly involved change though.

Issue #3469, opened 2022-05-19
Have axe.getRules() return actIds


WilcoFiers:
A while back we included an `actIds` property to rules for easier mapping between axe-core and ACT. This data is in axe-core, but there is no way to get it out.

This info is too niche I think to be useful in results, but there should be some way to get at it at least. I propose we use `axe.getRules()` for that. Seems like the easiest way to get that done.

padmavemulapati:
Validated with the latest axe-core develop branch code base source, `axe.js` , with `axe.getRules()` it is displaying the total ruleIds, 97 totally.
![image](https://user-images.githubusercontent.com/47409579/170037166-8c1ea5fc-0804-435e-9533-f429f86587d2.png)

PR #3470, opened 2022-05-19
feat(getRules): return actIds when set

WilcoFiers:
Return `actIds` from `axe.getRules()`.

Closes issue #3469


straker (approved):
LGTM, just a comment that you can decide what to do with.

straker (approved):

Issue #3466, opened 2022-05-19
frame-focusable-content should ignore negative tabindex


WilcoFiers:
The `frame-focusable-content` rule should not fail [Passed Example 2](https://www.w3.org/WAI/content-assets/wcag-act-rules/testcases/akn7bn/aa153f6799d28563054ce66bcf7dfcedf9b75288.html) of the [iframe with negative tabindex has no interactive elements]() ACT rule.

```html
<iframe tabindex="-1" srcdoc="
  <a href='/' tabindex='-1'>Home</a>
"></iframe>
```

padmavemulapati:
Not seeing any findings of `frame-focusible-content` except in `inapplicable`
<img width="1225" alt="image" src="https://user-images.githubusercontent.com/47409579/187349834-1a5f378b-79a1-41a8-95a6-6a9164e273b1.png">

straker:
@padmavemulapati what does your test HTML look like?

padmavemulapati:
Validated with the latest develop branch code base, when we run axe inside iframes and frame has its focussible content , then 
`frame-focussible-content` rule is failing instead passing

PR #3493, opened 2022-06-01
fix(frame-focusable-content): don't fail for elements with negative tabindex

straker:
Updated the `isFocusable` code since we were looking at it to 1) default export at top and 2) update the description as we spent a bit of time trying to figure out why an element with tabindex=-1 was still considered focusable.

Also, having the `frame-focusable-content` test have all the elements wrapped in `buttons` was strange so changed them to `divs`

Closes issue: #3466

### DIFF
--- a/lib/checks/keyboard/frame-focusable-content-evaluate.js
+++ b/lib/checks/keyboard/frame-focusable-content-evaluate.js
@@ -1,7 +1,26 @@
 import isFocusable from '../../commons/dom/is-focusable';
 
+export default function frameFocusableContentEvaluate(
+  node,
+  options,
+  virtualNode
+) {
+  if (!virtualNode.children) {
+    return undefined;
+  }
+
+  try {
+    return !virtualNode.children.some(child => {
+      return focusableDescendants(child);
+    });
+  } catch (e) {
+    return undefined;
+  }
+}
+
 function focusableDescendants(vNode) {
-  if (isFocusable(vNode)) {
+  const tabIndex = parseInt(vNode.attr('tabindex'), 10);
+  if (isFocusable(vNode) && (isNaN(tabIndex) || tabIndex > -1)) {
     return true;
   }
 
@@ -17,19 +36,3 @@ function focusableDescendants(vNode) {
     return focusableDescendants(child);
   });
 }
-
-function frameFocusableContentEvaluate(node, options, virtualNode) {
-  if (!virtualNode.children) {
-    return undefined;
-  }
-
-  try {
-    return !virtualNode.children.some(child => {
-      return focusableDescendants(child);
-    });
-  } catch (e) {
-    return undefined;
-  }
-}
-
-export default frameFocusableContentEvaluate;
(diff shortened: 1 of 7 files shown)


test/checks/keyboard/frame-focusable-content.js
```diff
@@ -45,13 +45,6 @@ describe('frame-focusable-content tests', function() {
     assert.isTrue(frameFocusableContent(null, null, vNode));
   });
 
-  it('should return false if element is natively focusable but has tabindex', function() {
```
straker: Turns out IE11 will take the empty `tabindex` and default it to `-3876`. But that value of tabindex also didn't prevent the browser from tabbing to the element so ¯\_(ツ)_/¯

lib/checks/keyboard/frame-focusable-content-evaluate.js
```diff
@@ -1,7 +1,26 @@
...
+    });
+  } catch (e) {
+    return undefined;
+  }
+}
+
 function focusableDescendants(vNode) {
-  if (isFocusable(vNode)) {
+  const tabIndex = parseInt(vNode.attr('tabindex'), 10);
+  if (isFocusable(vNode) && (isNaN(tabIndex) || tabIndex > -1)) {
```
WilcoFiers, line 23: Probably put these in reverse order, for performance:

```suggestion
  if ((isNaN(tabIndex) || tabIndex > -1) && isFocusable(vNode)) {
```

lib/commons/dom/is-focusable.js
```diff
@@ -4,15 +4,14 @@ import AbstractVirtualNode from '../../core/base/virtual-node/abstract-virtual-n
...
+ * Determines if an element is keyboard or programmatically focusable.
  * @method isFocusable
  * @memberof axe.commons.dom
  * @instance
  * @param {HTMLElement} el The HTMLElement
  * @return {Boolean} The element's focusability status
  */
-
-function isFocusable(el) {
+export default function isFocusable(el) {
```
WilcoFiers, line 14: I wonder if maybe we should add some option here to say `isFocusable(el, { noNegativeTabindex: true })`?

This seem like sort of a common occurance.

test/checks/keyboard/frame-focusable-content.js
```diff
@@ -10,32 +10,45 @@ describe('frame-focusable-content tests', function() {
   });
 
   it('should return true if element has no focusable content', function() {
-    var vNode = queryFixture('<button id="target"><span>Hello</span></button>');
+    var vNode = queryFixture('<div id="target"><span>Hello</span></div>');
```
WilcoFiers, line 13: Why did you have to change these?

WilcoFiers (changes requested):

lib/commons/dom/is-focusable.js
straker: I agree. Though maybe the option could be called `programmatically`?

```
const focusable = isFocusable(el, { programmatically: false })
```

straker: https://github.com/dequelabs/axe-core/issues/3500

PR #3522, opened 2022-06-27
docs(code-submission-guidelines): add coding guidelines

straker:


doc/code-submission-guidelines.md
```diff
@@ -4,11 +4,170 @@ We've enacted standards for commits and pull requests to effectively manage the
...
+Note that the id added to the `violations` array is inside an array. This is because the `violations` and `passes` arrays are axe-core selectors, which follow the format:
+
+- a single string - the string is the CSS selector
+- multiple strings
+  - The last string is the final CSS selector
+  - All other's are the nested structure of iframes inside the document
+
 ## Git Commits
 
 We follow Angular's code contribution style with precise rules for formatting git commit messages.
```
michael-siek, line 173: Should we add the link to conventional commits here?

michael-siek (changes requested):

doc/code-submission-guidelines.md
```diff
@@ -4,11 +4,170 @@ We've enacted standards for commits and pull requests to effectively manage the
...
 - [Git Commits](#git-commits)
 - [Submitting a pull request](#submitting-a-pull-request)
 - [Merging a pull request](#merging-a-pull-request)
 - [Squashing Commits](#squashing-everything-into-one-commit)
 
+## Code Guidelines
+
+### Default Export at Top
+
+When coding using [JavaScipt modules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules), the `default export` should be at the top of the file right after any import statements and module level variables. This ensures that when you open the file the main code path is the first thing you read.
```
stephenmathieson, line 17: This seems arbitrary to me, as it doesn't change how the module works/reads. It also won't work when using function expressions:

Block-scoped variable 'myFunction' used before its declaration

doc/code-submission-guidelines.md
```diff
@@ -4,11 +4,170 @@ We've enacted standards for commits and pull requests to effectively manage the
...
+### Return Early / Happy Path Coding
+
+[Return Early Coding](https://medium.com/swlh/return-early-pattern-3d18a41bba8) (also called "[Happy Path Coding](https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88)") is a coding pattern where you try to keep the main execution flow of the function aligned to a single column edge. Doing so allows someone to quickly scan down the column to see the expected flow or happy path of the function.
+
+In return early coding, the idea is to keep the main execution flow to the left column and use if statements to primarily handle errors or edge cases. This typically means that the function will include many `return` statements instead of the concept of a single return.
+
+In return early coding, we also try to declare variables at the moment they are needed rather than listing them all at the top of the function.
+
+<details>
+  <summary>Example of nested execution flow</summary>
```
stephenmathieson, line 30: Are these examples necessary? The Mat Ryer article does a **great** job at explaining this.

doc/code-submission-guidelines.md
```diff
@@ -4,11 +4,170 @@ We've enacted standards for commits and pull requests to effectively manage the
...
+
+  return validateReportStructure(results);
+}
+```
+
+</details>
+
+### Docblock Comments
+
+We use [Docblock comments](https://en.wikipedia.org/wiki/Docblock) in our code. Docblock comments are a way to describe what a function does, it's signature, and describe it's inputs.
```
stephenmathieson, line 114:
```suggestion
We use [Docblock comments](https://en.wikipedia.org/wiki/Docblock) in our code. Docblock comments are a way to describe what a function does, its signature, and describe its inputs.
```

stephenmathieson (changes requested):

doc/code-submission-guidelines.md
straker: I guess we should clarify, it should be this:

```js
default export myfunction() {
  return "Hello World"
}
```

I'm not sure how to distinguish that though (the technical term for it)

stephenmathieson: Why should we require people not use function expressions (and declarations instead)? I don't think it's worthwhile to police what JS features contributors use.

This whole block seems like a matter of personal preference, which we intentionally got away from by adding Prettier. Prior to that, there was _tons_ of back-and-forth about code style/spacing/quotes/etc, none of which helped axe-core be what it is today.

straker: The team has adopted this policy, so I'm just codifying it in our repo as per our team discussion.

stephenmathieson: Right below this, you say:

> If you encounter any code that we maintain that does not put the `default export` at the top, you should update the file to do so.

So I'm not sure this has been adopted.

Again, why should we prefer declarations over expressions? Why does this matter?

WilcoFiers: I don't think this is going to prevent anyone who want to contribute from doing so, and it does make the code base easier to work in. I think a consistent code base is more important than they very occasional "Can you put this function at the top of the file" comment.

WilcoFiers (approved):

doc/code-submission-guidelines.md
stephenmathieson: > I don't think this is going to prevent anyone who want to contribute from doing so

It's not going to prevent contributions, but it is extremely tedious and somewhat discouraging to be told _"This is great, but redo it for an arbitrary reason"_.

> it does make the code base easier to work in

Based on what? It requires people to use function declarations rather than expressions, which are **much** more common since "arrow functions" became a thing. If anything, I think this makes it _harder_.

> I think a consistent code base is more important

It's not consistent at all. There are _tons_ of files that use arrow functions and `export` things at the bottom.

WilcoFiers:
No response, merging.

Issue #3465, opened 2022-05-19
valid-lang should check that there is actual content


WilcoFiers:
For consistency with the [Element with `lang` attribute has valid language tag](https://w3.org/WAI/standards-guidelines/act/rules/de46e4) ACT rule, axe-core should make sure that the `valid-lang` rule does not fail elements that are empty, or that only contain whitespace characters.

Here are some of the ACT example we're failing, when we shouldn't be:

### [Passed example 4](https://www.w3.org/WAI/content-assets/wcag-act-rules/testcases/de46e4/8376f95166a75a8541217b77ae6a235f1aac6c3d.html)

```html
<article lang="invalid">
  <div lang="en">
    They wandered into a strange Tiki bar on the edge of the small beach town.
  </div>
</article>
```

### [Inapplicable example 3](https://www.w3.org/WAI/content-assets/wcag-act-rules/testcases/de46e4/a44f5e11d20feec4ae39e2db0336ddef0a8e04ec.html) 

```html
<p lang="hidden">
  <span style="display: none;">
    They wandered into a strange Tiki bar on the edge of the small beach town.
  </span>
</p>
```

### [Inapplicable example 4](https://www.w3.org/WAI/content-assets/wcag-act-rules/testcases/de46e4/471e3f82cdd9122e2886d2d7bcfc8cda1397a51d.html) 

```html
<div lang="invalid">
  <img src="/WAI/content-assets/wcag-act-rules/test-assets/shared/fireworks.jpg" alt="" />
</div>
```

MelSumner:
👋  we used [is-language-code](https://www.npmjs.com/package/is-language-code) when implementing the `--lang` flag in ember-cli. Sharing in case it's useful! Thank you for all you do! 👍

WilcoFiers:
Thanks Melanie! That looks a little too strict for what we want in axe-core though. Plus the size of it is a bit of an issue. Either way, appreciate the suggestion.

Glad you like our work, if you're ever interested to help, get in touch. I hear good things about your work on Ember too!

padmavemulapati:
Validate with the latest develop branch code base, 
test Script: ` <div id="target"><span lang="en"> text </span></div>`
<img width="1038" alt="image" src="https://user-images.githubusercontent.com/47409579/188797174-2b7903c9-6ae8-41d6-8964-e615c6ba244e.png">

PR #3523, opened 2022-06-28
fix(valid-lang): ignore lang on elements with no text

WilcoFiers:
Have the valid-lang rule ignore lang attributes on elements without text, of where its text is in a child node with yet another lang attribute.

Closes issue: #3465

Following this PR, i'll make not-html-matches deprecated in #3524, to get the tests on this one passing though I'm already taking out the tests for it.


straker (commented):
While we're here, do we need to handle `aria-owns` applying the `lang` attribute to the owned children?

```html
<div role="list" lang="fr" aria-owns="contents"></div>
<div id="contents" role="none">
  <div role="listitem">Ceci est un paragraphe</div>
</div>
```

In JAWS / Chrome and JAWS / Edge, JAWS applies the French accent to the owned listitems. However our code would treat the `lang` node as having no children text nodes.

For completeness:
* NVDA / Firefox - only applies French accent when the `#contents` element has the `lang`
* JAWS / IE11  - only applies French accent when the `#contents` element has the `lang`
* VO / Safari - I couldn't get this one to work as VoiceOver seems to ignore the `lang` attribute and autodetect the language from the sentence?

test/commons/dom/has-lang-text.js
```diff
@@ -0,0 +1,83 @@
...
+  'use strict';
+  var hasLangText = axe.commons.dom.hasLangText;
+  var fixture = document.getElementById('fixture');
+  var tree;
+
+  it('returns true when the element has a non-empty text node as its content', function () {
+    fixture.innerHTML = '<div id="target">  text  </div>';
+    tree = axe.utils.getFlattenedTree(fixture);
+    var target = axe.utils.querySelectorAll(tree, '#target')[0];
+    assert.isTrue(hasLangText(target, { skipNestedLang: true }));
```
straker, line 11: I don't see `skipNestedLang` as an option for `hasLangText`. Should it be here for all the function calls?

straker (dismissed):

test/commons/dom/has-lang-text.js
WilcoFiers: Good spot. This is a holdover from something I tried to do before.

lib/commons/dom/is-visual-content.js
```diff
@@ -1,8 +1,13 @@
...
+import AbstractVirtualNode from '../../core/base/virtual-node/abstract-virtual-node';
+
 const visualRoles = [
   'checkbox',
   'img',
+  'meter',
+  'progressbar',
+  'scrollbar',
   'radio',
-  'range',
```
WilcoFiers, line 5: Range is an abstract role. This should've never been here.

test/integration/full/isolated-env/isolated-env.html
```diff
@@ -79,7 +79,7 @@ <h2>Ok</h2>
           <title>I am a circle</title>
           <circle cx="50" cy="50" r="40" fill="yellow"></circle>
         </svg>
-        <div lang="en"></div>
+        <div lang="en">English</div>
```
WilcoFiers, line 82: Since this is now inapplicable, rather than a pass.

test/rule-matches/not-html-matches.js
```diff
@@ -1,20 +0,0 @@
-describe('not-html-matches', function() {
-  'use strict';
-
-  var fixture = document.getElementById('fixture');
-  var rule;
-
-  beforeEach(function() {
-    rule = axe.utils.getRule('valid-lang');
```
WilcoFiers, line 8: There's no way to run these tests, since there's no way to get at the matcher. I'm deprecating the matcher in #3524

lib/commons/dom/has-lang-text.js
```diff
@@ -0,0 +1,24 @@
...
+ * @param {VirtualNode} virtualNode 
+ * @return boolean
+ */
+export default function hasLangText(virtualNode) {
+  if (typeof virtualNode.children === 'undefined' || hasChildTextNodes(virtualNode)) {
+    return true;
+  }
+  if (virtualNode.props.nodeType === 1 && isVisualContent(virtualNode)) {
+    // See: https://github.com/dequelabs/axe-core/issues/3281
+    return !!axe.commons.text.accessibleTextVirtual(virtualNode);
```
dylanb, line 17: this call here is going to be slow

lib/rules/valid-lang.json
```diff
@@ -1,7 +1,7 @@
 {
   "id": "valid-lang",
-  "selector": "[lang], [xml\\:lang]",
-  "matches": "not-html-matches",
+  "selector": "[lang]:not(html), [xml\\:lang]:not(html)",
+  "matches": "has-lang-text-matches",
```
dylanb, line 4: Could we switch these around? I.e. only look for the text if the lang value is invalid?

lib/commons/dom/has-lang-text.js
WilcoFiers: Possibly, although it'll be rare. You generally don't see people put lang attributes on images with aria-labelledby or whatever.

lib/checks/language/valid-lang-evaluate.js
```diff
@@ -1,5 +1,6 @@
 import { isValidLang, getBaseLang } from '../../core/utils';
 import { sanitize } from '../../commons/text';
+import { hasLangText } from '../../commons/dom'
 
 function validLangEvaluate(node, options, virtualNode) {
   const invalid = [];
```
dylanb, line 6: nit: `invalid` is now a bad name for this variable

dylanb (approved):
I would prefer if we renamed the variable `invalid` to `valid` to correctly represent what is happening now

PR #3524, opened 2022-06-28
feat(matches): deprecate not-html-matches, use :not(html) instead

WilcoFiers:


WilcoFiers:
Too many conflicts. I'm just going to cherry-pick on a new branch.

Issue #3497, opened 2022-06-02
New rule: Meta-refresh-no-exceptions


WilcoFiers:
The current meta-refresh rule doesn't actually check that the refresh exceeds the 20 hour limit, but once #3496 is resolved it will. We should have a version of this rule for WCAG level AAA that does not permit any delay. This would be a rule that is consistent with the [Meta element has no refresh delay (no exception)
](https://wai-wcag-act-rules.netlify.app/standards-guidelines/act/rules/bisz58/proposed/) ACT rule.

padmavemulapati:
Validated with latest develop branch code base,
`meta-refresh-no-exceptions` rule for pass and fail cases:
```
<head>
    <meta charset="utf8" />
    <title>Meta-refresh-no-exceptions fail 1</title>
    <meta http-equiv="refresh" content="72001: https://w3.org" />
    <meta http-equiv="refresh" content="90000; https://deque.com/">
    
    <link
      rel="stylesheet"
      type="text/css"
      href="/node_modules/mocha/mocha.css"
    />
```
<img width="1581" alt="image" src="https://user-images.githubusercontent.com/47409579/187353568-8c2062f1-36e3-4539-83cc-d018d4f75c5f.png">

PR #3525, opened 2022-06-29
fix(meta-refresh): Add WCAG's 20-hour exception

WilcoFiers:
1. Ensure that redirects after 20 hours are passed
2. Improve the accuracy of the redirect number parsing
3. Add options for minDelay and maxDelay, to make the check configurable
4. Make the rule inapplicable when there's no content attribute
5. Add more integration tests

Closes issue: #3496

PR #3530, opened 2022-07-04
feat(new-rule): Add meta-refresh-no-exceptions (wcag2aaa, disabled by default)

WilcoFiers:
Adds a WCAG 2 AAA variant of the meta-refresh rule. Prior to #3525, the meta-refresh rule wouldn't handle the 20+ hour exception in WCAG. Now that it does, we should have a AAA rule that doesn't allow the exception.

Closes issue: https://github.com/dequelabs/axe-core/issues/3497


dylanb:
Seems arbitrary that this AAA rule is off by default and the other one we have is on by default, shouldn't we be more consistent?

Issue #3467, opened 2022-05-19
avoid-inline-spacing should consider the values


WilcoFiers:
Axe-core should not just look at whether or not `!important` is used, but also whether or not the value is below the threshold WCAG sets for it. For example the following should not fail the `avoid-inline-spacing` rule, because 0.15em is sufficient to satisfy WCAG:

```html
<p style="letter-spacing: 0.15em !important">
  The toy brought back fond memories of being lost in the rain forest.
</p>
```

There are quite a few ACT examples axe-core is getting wrong because it's not considering the value, or how it's set:

- [Letter spacing in `style` attributes is not `!important`](https://www.w3.org/WAI/standards-guidelines/act/rules/24afc2/)
   - [Passed example 2](https://www.w3.org/WAI/content-assets/wcag-act-rules/testcases/24afc2/9e9382901f59c7dd476717a55bf5c5a37ed76bbc.html)
   - [Passed example 3](https://www.w3.org/WAI/content-assets/wcag-act-rules/testcases/24afc2/43f8fe88b8e7365db7aa251b263b5d00c7a47ae9.html)
   - [Passed example 4](https://www.w3.org/WAI/content-assets/wcag-act-rules/testcases/24afc2/787f24a573fa422e24ab72312f7306253bb83a4f.html)
   - [Passed example 5](https://www.w3.org/WAI/content-assets/wcag-act-rules/testcases/24afc2/f000a9c495f11a4a11a4314871b91f4173e4589a.html)
   - [Passed example 7](https://www.w3.org/WAI/content-assets/wcag-act-rules/testcases/24afc2/6aa2034507dc16e6ae0d16f1b6f2a14d3dfadc18.html)
   - [Passed example 8](https://www.w3.org/WAI/content-assets/wcag-act-rules/testcases/24afc2/64b25817b3d3909ab7f4acaee061875ebac1cee3.html)
   - [Inapplicable example 8](https://www.w3.org/WAI/content-assets/wcag-act-rules/testcases/24afc2/88d6ea5706ed8ae188caa166879c381e64e5077a.html)

- [Word spacing in `style` attributes is not `!important`](https://www.w3.org/WAI/standards-guidelines/act/rules/9e45ec/)
   - [Passed example 2](https://www.w3.org/WAI/content-assets/wcag-act-rules/testcases/24afc2/45e5a588c3e8977fa0e83074d7f7c89738e8ec42.html)
[...]

PR #3533, opened 2022-07-05
fix(avoid-inline-spacing): add spacing threshold

WilcoFiers:
- Replaces the avoid-inline-styling check three new checks, one for each CSS property
- Add a important-css-property evaluate method
- Enable ACT rule tests related to the avoid-inline-spacing axe-core rule
- For the line-height property, only fail multi-line texts

Closes issue: https://github.com/dequelabs/axe-core/issues/3467


lib/checks/shared/inline-style-property-evaluate.js
```diff
@@ -0,0 +1,71 @@
+import { getLineCount } from '../../commons/text';
+
+/**
+ * Check if a CSS property
```
stephenmathieson, line 4:
```suggestion
 * Check if a CSS property [does/has something]
```

lib/commons/text/get-line-count.js
```diff
@@ -0,0 +1,29 @@
+/**
```
straker, line 1: Missing tests for `get-line-count`

lib/commons/text/get-line-count.js
```diff
@@ -0,0 +1,29 @@
...
+ * Return the number of line wraps in a DOM element
+ * 
+ * IMPORTANT: This function assumes single-column flowing text.
+ * Floating images, flex-box, and positioned content can throw off
+ * the count. 
+ * @param {Element} domNode 
+ * @param {number} margin (default: 2)
+ * @returns {number}
+ */
+export default function getLineCount(domNode, margin = 2) {
```
straker, line 11: This function could be simplified to something like `isMultilineText` and then return if `range.getClientRects() > 1`. Calculating the exact number of lines is not necessary and could be error prone. All we really need to know is if there is more than 1 line.

straker (changes requested):

lib/commons/text/get-line-count.js
WilcoFiers: Whaaaa??? I swear I wrote those :-/

test/checks/shared/inline-style-property.js
```diff
@@ -0,0 +1,343 @@
...
+        '<p style="line-height: 1.2em; max-width: 200px;" id="target">' + 
+        '	The toy brought back fond memories of being lost in the rain forest.'+
+        '</p>'
+      );
+      var result = checkEvaluate.apply(checkContext, params);
+      assert.isTrue(result);
+      assert.isNull(checkContext._data);
+    });
+
+    (isIE11 ? xit : it)('is false when below 1.5em and !important', function () {
```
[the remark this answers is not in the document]
WilcoFiers: Beat me... IE reported the value as 0.19. Guess IE just cannot count. Didn't think it worth debugging since we're pulling IE soon.

test/integration/rules/avoid-inline-spacing/avoid-inline-spacing.html
```diff
@@ -20,18 +17,31 @@
 >
   Look, a distraction!
 </p>
+<p id="pass6" style="line-height: 1.5em !important;">Banana<br>pass</p>
```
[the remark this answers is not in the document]
WilcoFiers: Seems reasonably well handled by the check test. What's the benefit?

lib/checks/shared/inline-style-property-evaluate.js
```diff
@@ -0,0 +1,71 @@
...
+    return true;
+  }
+  return false;
+}
+
+function getNumberValue(domNode, { cssProperty, absoluteValues }) {
+  const computedStyle = window.getComputedStyle(domNode);
+  const cssPropValue = computedStyle.getPropertyValue(cssProperty);
+  if (cssPropValue === 'normal') {
+    return 0;
```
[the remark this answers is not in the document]
WilcoFiers: Good catch. Yeah line-height shouldn't be 0.

test/integration/rules/avoid-inline-spacing/avoid-inline-spacing.html
```diff
@@ -20,18 +17,41 @@
...
+<!-- one-liner   -->
+<p id="pass7" style="line-height: 1.4em !important;">Banana</p>
+<p id="pass8" style="letter-spacing: 0.12em !important;">
+  We need more cheeeeeeessseeeee!!!
+</p>
+<p id="pass9" style="word-spacing: 0.16em !important;">
+  The cheese grater is in the way!
+</p>
+
+<!-- CSS keywords -->
```
WilcoFiers, line 30: I didn't add "initial", since IE chokes on that.

straker (approved):

PR #3536, opened 2022-07-11
chore: add locale/_template.json

WilcoFiers:
Adding a template for localisations, this should hopefully make reviewing a little easier.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -148,6 +148,13 @@ module.exports = function(grunt) {
       }
     },
     'add-locale': {
+      template: {
+        options: {
+          lang: 'xyz'
+        },
+        src: ['tmp/core/core.js'],
+        dest: './locales/_template.json'
+      },
       newLang: {
         options: {
           lang: grunt.option('lang')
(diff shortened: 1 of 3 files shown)


Gruntfile.js
```diff
@@ -249,7 +256,7 @@ module.exports = function(grunt) {
     }
   });
 
-  grunt.registerTask('translate', ['validate', 'esbuild', 'add-locale']);
+  grunt.registerTask('translate', ['validate', 'esbuild', 'add-locale:newLang']);
```
dylanb, line 259: Shouldn't we generate this template every time we build?

dylanb (commented):
Should we generate the template every time we build?

Gruntfile.js
WilcoFiers: That happens here:
https://github.com/dequelabs/axe-core/pull/3536/files#diff-8e7e5f941c67b0cf481a9d650295119d4922b96117e7adc069ba30a81a78c65fR270

Because I added a `template` prop to `add-locale`, I need to explicitly call `add-locale:newLang`, otherwise both commands will run in `translate`.

dylanb (approved):

PR #3539, opened 2022-07-11
chore: update dependencies

dbowling:
Updates project dependencies to current versions when possible.

The following dependencies are one or more major versions behind and have not been upgraded to latest, but should be safe for minor and patch updates.

- **chalk**: `4.x` (latest is `5.0.1`) 
Version `5` is breaking due to ESM only https://github.com/chalk/chalk/releases/tag/v5.0.0
- **execa**: `5.x` (latest is `6.1.0`) 
Version `6` is breaking due to ESM only https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c
- **globby**: `11.x` (latest is `13.1.2`)
Version `12` is breaking due to ESM only https://github.com/sindresorhus/globby/releases/tag/v12.0.0
- **html-entities**: `1.x` (latest is `2.3.3`)
Version `2` is breaking due to ESM only https://github.com/mdevils/html-entities/releases/tag/v2.0.0
- **markdown-table**: `2.x` (latest is `3.0.2`)
Version `3` is breaking due to ESM only https://github.com/wooorm/markdown-table/releases/tag/3.0.0
- **mocha**: `9.x` (latest is `10.0.0`)
Version `10` drops IE 11 and Node 12 support https://github.com/mochajs/mocha/blob/master/CHANGELOG.md#1000--2022-05-01
- **sinon**: `11.x` (latest is `14.0.0`)
Version `12` is breaking due to ESM only https://github.com/sinonjs/sinon/blob/main/CHANGES.md#1200
Version `10` is breaking due to IE 11 not supporting template literals https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#browser_compatibility

Seleniumn is pinned due to a bug, but should be easily updated in the future.
- **selenium-webdriver** is pinned to `4.3.0` because `4.3.1` introduces a breaking change from 4.3.0->4.3.1
`selenium-webdriver/chrome#setDefaultService()`  https://github.com/SeleniumHQ/selenium/pull/10796/files#diff-6c87d95a2288e92e15a6bb17710c763c01c2290e679beb26220858f3218b6a62L260

Also, it appears the selenium webdriver browser message "Chrome is being controlled by automated software" changes the window height when it appears between running the two axe runs. To fix `deepEqual`, we assign the height to be identical: `axeRunPartialResult.testEnvironment = axeRunResult.testEnvironment;` 

**ESBuild** is upgraded to `0.14.7` (`0.14.50 is latest`) due to flaky windows tests. This is the version that passed CI and was reasonably up-to-date.

replaces: PR #3535


package.json
```diff
@@ -105,79 +105,76 @@
...
+    "mocha": "9.x",
+    "node-notifier": "^10.0.1",
     "npm-run-all": "^4.1.5",
-    "prettier": "^1.17.1",
+    "prettier": "^2.7.1",
     "proxyquire": "^2.1.3",
     "revalidator": "~0.3.1",
-    "selenium-webdriver": "~3.6.0",
-    "sinon": "^7.5.0",
+    "selenium-webdriver": "3.x",
```
michael-siek, line 169: I think we should upgrade this to use `4.3.0`

michael-siek (dismissed):

package.json
dbowling: I'll take a look. I remember now that you shared an update about a breaking change with that package; maybe that was my issue with the newer 4.3.1.

straker (commented):
LGTM. @WilcoFiers should also approve since I helped a bit.

.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+npx --no grunt configure && npx --no lint-staged
```
WilcoFiers, line 4: Why put this on the same line?

Also, should we stage rule-descriptions.md? We can call `grunt configure` all we want, if we don't stage it, it won't end up in the commit.

doc/aria-supported.md
```diff
@@ -8,6 +8,5 @@ For a detailed description about how accessibility support is decided, see [How
 
 ## Attributes
 
-| aria-attribute   | axe-core support |
-| ---------------- | ---------------- |
-| aria-describedat | No               |
+| aria-attribute | axe-core support |
+| -------------- | ---------------- |
```
WilcoFiers, line 12: This seems odd. Should we turn off this part of the task? An empty table doesn't seem useful to me.

test/integration/full/run-partial/after-method.js
```diff
@@ -23,6 +23,11 @@ describe('run-partial, after-method', function() {
         var axeRunResult = results[1];
         assert.lengthOf(axeRunPartialResult.violations, 0);
 
+        // it appears the selenium webdriver browser message "Chrome is being controlled by
+        // automated software" changes the window height when it appears
+        // between running the two axe runs
+        axeRunPartialResult.testEnvironment = axeRunResult.testEnvironment;
+
```
WilcoFiers, line 30: I don't think these comments are necessary. Put 'm on the PR if you like, so we can find it. It is not relevant for understanding this code.

WilcoFiers (changes requested):
Couple minor points

.husky/pre-commit
stephenmathieson: > if we don't stage it, it won't end up in the commit

Doesn't lint-staged take care of this?

dbowling: `lint-staged` only runs on files that are staged. If you run `npx lint-staged` without staging anything, it outputs `No staged files found.`

These commands existed on the same line in the original hook. Since `husky` will fail the commit if any command returns non-zero, I don't think there's any practical difference. I agree that it would be more clear to have them on different lines because it currently implies something that isn't really there using `&&`.

doc/aria-supported.md
dbowling: I couldn't find any other docs page that had an empty table, so this does stand out. Since it is generated dynamically, would disabling the rule have unexpected consequences in the future? 

My thoughts now are to not disable toe task, but instead alter `aria-supported.js` to output an empty string if this might bite us later. If not, disabling the task or culling the code would certainly be easiest.

dbowling:
@WilcoFiers Other than fixing mocha test timeouts happening in CI, do you have anything else you want addressed in this PR? Your aria-supported doc feedback is all that I see outstanding from you before you are on holiday.

doc/aria-supported.md
WilcoFiers: Something needs to be done there. It's definitely outdated. Perhaps we should change that page to ARIA 1.3. I'll open a ticket.

package.json
```diff
@@ -105,79 +105,76 @@
...
-    "memoizee": "^0.4.14",
+    "karma-spec-reporter": "0.0.34",
+    "lint-staged": "^13.0.3",
+    "make-dir": "^3.1.0",
+    "markdown-table": "2.x",
+    "memoizee": "^0.4.15",
     "minami": "^1.2.3",
-    "mocha": "^9.2.0",
-    "node-notifier": "^8.0.1",
+    "mocha": "9.x",
```
WilcoFiers, line 163: Why not Mocha 10?

WilcoFiers (changes requested):

package.json
dbowling: Because version 10 drops IE 11 and Node 12 support.
 
https://github.com/mochajs/mocha/blob/master/CHANGELOG.md#1000--2022-05-01

doc/aria-supported.md
dbowling: Ok, I'll let you open a ticket and will mark this as resolved as-is.

.husky/pre-commit
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+npx --no grunt configure
+npx --no lint-staged
```
straker, line 5: I think it's a safe requirement to say if you are commit code you need to have dev dependencies installed. `lint-staged` also runs prettier and eslint, which are required dev dependencies 

```suggestion
npx grunt configure
npx lint-staged
```

straker (changes requested):

package.json
dbowling: I spoke with @straker yesterday. Since we need the IE 11 support, we won't upgrade to Mocha 10 and will leave this comment as unresolved in the PR.

dbowling:
This PR is ready for security review, even though it has pending changes--those will be left open, see above.

straker (approved):

doc/aria-supported.md
dbowling: New issue for this is #3590.

Issue #3132, opened 2021-08-24
put back act test aria-props-permitted


straker:
We had to remove the test so we could release 4.3.3. We need to put it back when we do a 4.4 release which will include the [appropriate fix](https://github.com/dequelabs/axe-core/pull/3102).

padmavemulapati:
Validated with the latest develop branch code base, `aria-allowed-attribute` rule is now passing for attribute `aria-busy` and `aria-owns` and failing with earlier attribute `aria-selected` and `audio`
Script:
```
<div id="target" tabindex="1" aria-selected="true" aria-checked="true"></div>
      <div id="target" tabindex="1" aria-busy="true" aria-owns="foo"></div>
      <div id="target" tabindex="1" aria-selected="true" aria-owns="foo"></div>
      <audio
  src="/test/assets/moon-speech.mp3"
  controls
  aria-orientation="horizontal"
  id="failCase"
></audio>
```
<img width="1186" alt="image" src="https://user-images.githubusercontent.com/47409579/187334277-8a6890da-897d-44ac-ae67-019a54a61bac.png">

PR #3558, opened 2022-07-21
test(aria-allowed-attr): put back act test aria-props-permitted

pschwendy:
Closes issue: #3132


test/act-mapping/aria-allowed-attr.json
```diff
@@ -0,0 +1,5 @@
+{
+  "id": "5c01ea",
+  "title": "Aria attributes are used validly on elements depending on their semantic role",
```
straker, line 3: The title should be the name of the ACT rule.

```suggestion
  "title": "ARIA state or property is permitted",
```

straker (dismissed):

straker (approved):

Issue #3532, opened 2022-07-04
color-contrast incorrectly calculating background color of an element in closed `<details>`


wojtekmaj:
Product: axe-core

Expectation: When calculating contrast ratio for an element in closed `<details>`, either:
* take background color of open `<details>` element (might not be possible without page interaction, which could be a roadblock), 
* or don't take any background color at all, effectively skipping the color-contrast test for this element.
* or if we really really would like to take background color of where the element _would have been_ if the `<details>` element was open, ignore `<details>`'s siblings when calculating background colors, because they will be pushed down when `<details>` is eventually opened.

Actual: background color is taken from the place where the element _would have been_  if the `<details>` element was open. Unfortunately, in our example, in this place we have a dark blue button which, when combined with black text, fails the test.

Motivation: Current behavior is obviously faulty.

<pre><code>
axe-core version: checked on: 4.0.0, …, 4.1.0, …, 4.2.0, …, 4.3.0, …, 4.4.0, …, 4.4.2

Browser and Assistive Technology versions
Chrome 103.0.5060.53
</code></pre>

Repro:

```html
<html lang="en-US">
  <head>
    <style>
      span {
        position: relative;
      }

      button {
        background-color: darkblue;
        color: #fff;
      }
    </style>
    <title>color-contrast bug</title>
  </head>

  <body>
    <header>
      <h1>color-contrast bug</h1>
    </header>
    <main>
      <details>
        <summary>Summary</summary>
        <span>Some text</span>
      </details>
      <button>A blue button</button>
    </main>
  </body>
</html>

```

wojtekmaj:
It appears that the problem is in `filteredRectStack` function, which in this case, returns `[span, button, main, body, html]`.

WilcoFiers:
Thanks for reporting. We'll take a closer look ASAP.

straker:
Looking at the issue, it appears we don't treat elements inside the `<details>` as hidden when the element is collapsed. As is, we think the `<span>` element is currently visible and positioned right above the button, so that's why we reporting the overlapping element.

I believe `<details>` elements have the `open` attribute when expanded, so we should be able to use that to determine visibility.

straker:
Need to discuss with Wilco the ramifications of saying elements inside closed detail elements are hidden as we may not run axe-core rules against them.

straker:
[...]

PR #3562, opened 2022-07-25
[WIP] fix: do not run axe against elements inside closed details

straker:
Closes issue: #3532


lib/commons/dom/is-visible.js
```diff
@@ -96,6 +96,20 @@ function isAreaVisible(el, screenReader, recursed) {
   );
 }
 
+/**
+ * Check `DETAILS` element is visible
+ * @param {Element} el `DETAIL` element
+ * @retruns {Boolean}
+ */
```
stephenmathieson, line 103:
```suggestion
/**
 * Check if a `<details>` element is visible.
 * @param {HTMLDetailsElement} el The element to check
 * @retruns {boolean}
 */
```

stephenmathieson (approved):
LGTM other than some minor JSDoc stuff.

straker:
This currently breaks `link-name` rule if there is a link inside of a closed `<details>` element. We may need to do a different strategy or modify `is-hidden` to account for this as well since currently the `link-name` rule does not think any anchor elements inside closed details are hidden, then tries to run accessible name on the link which uses `is-visible` and returns an empty string since the text nodes are not visible.

wojtekmaj:
Remember that `<summary>` remains visible even on closed `<details>` element.

straker:
We discussed this and the resolution is we'll update the `is-hidden` function to account for this. This means rules _will not run_ against elements inside of a closed `<details>` elements and users will need to expand the details in order to get accessibility checks.

lib/commons/dom/is-visible.js
```diff
@@ -179,6 +193,10 @@ function isVisible(el, screenReader, recursed) {
     return isAreaVisible(el, screenReader, recursed);
   }
 
+  if (nodeName === 'details') {
+    return isDetailsVisible(el);
```
WilcoFiers, line 197: I'm a little confused. This seems to suggest to me that the details element is hidden when it isn't open, rather than its content being hidden -- except for the summary element. I'd like to see some tests that show this difference.

lib/commons/dom/is-visible.js
```diff
@@ -96,6 +96,20 @@ function isAreaVisible(el, screenReader, recursed) {
...
+ * Check if a `<details>` element is visible.
+ * @param {HTMLDetailsElement} el The element to check
+ * @retruns {boolean}
+ */
+function isDetailsVisible(el) {
+  // IE11 does not support details (without a
+  // polyfill), so we need to check for the `open`
+  // property in the element to determine if
+  // the browser supports hiding content when
+  // the details element is closed
```
WilcoFiers, line 109: This seems more than enough

```suggestion
  // Handle lack of support in IE
```

WilcoFiers (changes requested):

lib/commons/dom/is-visible.js
```diff
@@ -226,6 +209,15 @@ function isVisible(el, screenReader, recursed) {
     return false;
   }
 
+  if (nodeName === 'summary') {
```
WilcoFiers, line 212: What happens if there's more than one summary element?

lib/commons/dom/is-visible.js
```diff
@@ -226,6 +209,15 @@ function isVisible(el, screenReader, recursed) {
...
+    return true;
+  }
+
+  // Handle lack of support in IE
+  if (nodeName === 'details' && 'open' in el && !el.hasAttribute('open')) {
+    return false;
```
WilcoFiers, line 218: I think maybe here we need to look at the nodeName of the parent, rather than the current node. The details element isn't hidden when it's closed. There are all sorts of ways in which that can make a difference, for example if there's an invalid ARIA attribute on a closed details element, we'll still want to report that I think.

WilcoFiers (changes requested):

lib/commons/dom/is-visible.js
```diff
@@ -208,6 +209,15 @@ function isVisible(el, screenReader, recursed) {
     return false;
   }
 
+  if (nodeName === 'summary') {
+    return true;
+  }
+
+  // Handle lack of support in IE
+  if (nodeName === 'details' && 'open' in el && !el.hasAttribute('open')) {
```
WilcoFiers, line 217: I'm not sure what to suggest here, but it jumped out at me that if `el` is somehow not an object, this is going to throw. That doesn't seem possible in this function as written today, but that's the kind of thing that could slip through in a refactor. I fee like a little defensive programming here could save us from a headache in the future.

WilcoFiers (changes requested):

lib/commons/dom/is-visible.js
wojtekmaj: It's then partially ignored. By "partially" I mean it's hidden in closed `<details>`, but has some `<summary>`'s styling applied to it. This, however, is invalid HTML - there has to be exactly one `<summary>` element according to MDN.

<details>
<summary>Summary 1</summary>
<summary>Summary 2</summary>
Content
</details>

straker:
This work will be done in #3351 as part of the refactor in #3617. Going to close and move the work into the other pr.

lib/commons/dom/is-visible.js
```diff
@@ -208,6 +209,15 @@ function isVisible(el, screenReader, recursed) {
     return false;
   }
 
+  if (nodeName === 'summary') {
+    return true;
+  }
+
+  // Handle lack of support in IE
```
wojtekmaj, line 216: Should we care about IE if Microsoft has already deprecated it…?

PR #3564, opened 2022-07-26
chore: add target assert to queryFixture

dbowling:
- Adds an assert to the `axe.testUtils.queryFixture()` to guide developers in usage of the fixture helper.
- Refactor to help VS Code pick up correct signatures
- Add JSDoc
- Whitespace fixes

### DIFF
--- a/test/checks/navigation/heading-order.js
+++ b/test/checks/navigation/heading-order.js
@@ -1,14 +1,15 @@
-describe('heading-order', function() {
+describe('heading-order', function () {
   'use strict';
 
   var checkContext = axe.testUtils.MockCheckContext();
   var queryFixture = axe.testUtils.queryFixture;
+  var fixtureSetup = axe.testUtils.fixtureSetup;
 
-  afterEach(function() {
+  afterEach(function () {
     checkContext.reset();
   });
 
-  it('should store the heading order path and level for [role=heading] elements and return true', function() {
+  it('should store the heading order path and level for [role=heading] elements and return true', function () {
     var vNode = queryFixture(
       '<div role="heading" aria-level="1" id="target">One</div><div role="heading" aria-level="3">Three</div>'
     );
@@ -31,7 +32,7 @@ describe('heading-order', function() {
     });
   });
 
-  it('should handle incorrect aria-level values', function() {
+  it('should handle incorrect aria-level values', function () {
     var vNode = queryFixture(
       '<div role="heading" aria-level="-1" id="target">One</div><div role="heading">Two</div>'
     );
@@ -54,7 +55,7 @@ describe('heading-order', function() {
     });
   });
 
-  it('should allow high aria-level values', function() {
+  it('should allow high aria-level values', function () {
     var vNode = queryFixture(
       '<div role="heading" aria-level="12" id="target">One</div>'
     );
@@ -73,7 +74,7 @@ describe('heading-order', function() {
     });
   });
 
-  it('should store the correct header level for hn tags and return true', function() {
+  it('should store the correct header level for hn tags and return true', function () {
     var vNode = queryFixture('<h1 id="target">One</h1><h3>Three</h3>');
     assert.isTrue(
       axe.testUtils
@@ -94,7 +95,7 @@ describe('heading-order', function() {
     });
   });
 
-  it('should allow aria-level to override semantic level for hn tags and return true', function() {
+  it('should allow aria-level to override semantic level for hn tags and return true', function () {
     var vNode = queryFixture(
       '<h1 aria-level="2" id="target">Two</h1><h3 aria-level="4">Four</h3>'
     );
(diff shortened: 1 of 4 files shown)


test/testutils.js
```diff
@@ -420,24 +420,43 @@ testUtils.assertStylesheet = function assertStylesheet(
  * Injecting content into a fixture and return queried element within fixture
  *
  * @param {String|Node} content to go into the fixture (html or DOM node)
+ * @param {String?} query - the CSS selector query to find target DOM node
```
straker, line 423:
```suggestion
 * @param {String} [query=#target] - the CSS selector query to find target DOM node
```

test/testutils.js
```diff
@@ -420,24 +420,43 @@ testUtils.assertStylesheet = function assertStylesheet(
...
   testOptions = testOptions || {};
 
-  return function evaluateWrapper(node, options, virtualNode, context) {
+  /**
+   * Wraps a check for evaluation using .call()
+   * @param {HTMLElement} node
+   * @param {*} options
+   * @param {VirtualNode} virtualNode
+   * @param {Context} context
+   * @returns {string}
```
straker, line 455:
```suggestion
   * @returns {Boolean|undefined}
```

test/testutils.js
```diff
@@ -420,24 +420,43 @@ testUtils.assertStylesheet = function assertStylesheet(
...
+  query = query || '#target';
   var rootNode = testUtils.fixtureSetup(html);
+  var vNode = axe.utils.querySelectorAll(rootNode, query)[0];
+  assert.exists(
+    vNode,
+    'Node does not exist in query `' +
+      query +
+      '`. This is usually fixed by adding the default target (`id="target"`) to your html parameter. If you do not intend on querying the fixture for #target, consider using `axe.testUtils.fixtureSetup()` instead.'
+  );
   return axe.utils.querySelectorAll(rootNode, query || '#target')[0];
```
straker, line 436:
```suggestion
  return vNode;
```

test/commons/text/form-control-value.js
```diff
@@ -408,30 +408,30 @@ describe('text.formControlValue', function() {
...
       var target = queryFixture(
         '<div id="target" role="combobox">' + comboboxContent + '</div>'
       );
       assert.equal(ariaComboboxValue(target), 'bar');
     });
 
-    it('returns `` for elements without role=combobox', function() {
-      queryFixture('<div role="combobox">' + comboboxContent + '</div>');
+    it('returns `` for elements without role=combobox', function () {
+      fixtureSetup('<div role="combobox">' + comboboxContent + '</div>');
```
straker, line 419: This would be better written by using `fixtureSetup` and passing a custom query. I can't suggest the next 4 lines, so this should replace the `querySelectorAll` for the `[role=listbox]` as well

```js
var target = queryFixture('<div role="combobox">' + comboboxContent + '</div>', '[role=listbox]');
```

test/checks/navigation/heading-order.js
```diff
@@ -117,8 +118,8 @@ describe('heading-order', function() {
     });
   });
 
-  it('should ignore aria-level on iframe when not used with role=heading', function() {
-    var vNode = queryFixture('<iframe aria-level="2"></iframe>');
+  it('should ignore aria-level on iframe when not used with role=heading', function () {
+    var vNode = fixtureSetup('<iframe aria-level="2"></iframe>');
```
straker, line 122: Even though `heading-order` doesn't make use of `vNode` anymore, for now we should be consistent, so adding a `#target` would be more appropriate rather than trying to save off `fixtureSetup` which returns the entire virtual node tree (which would break the test if we ever change how `heading-order` works in the future).

```suggestion
    var vNode = queryFixture('<iframe aria-level="2" id="target"></iframe>');
```

test/checks/navigation/heading-order.js
```diff
@@ -1,14 +1,15 @@
-describe('heading-order', function() {
+describe('heading-order', function () {
   'use strict';
 
   var checkContext = axe.testUtils.MockCheckContext();
   var queryFixture = axe.testUtils.queryFixture;
+  var fixtureSetup = axe.testUtils.fixtureSetup;
```
straker, line 6:
```suggestion
```

test/commons/text/native-text-alternative.js
```diff
@@ -1,46 +1,47 @@
...
-  it('returns `` when no method matches', function() {
+  it('returns `` when no method matches', function () {
     var vNode = queryFixture('<div id="target">baz</div>');
     assert.equal(nativeTextAlternative(vNode), '');
   });
 
-  it('returns `` when no accessible text method returned something', function() {
-    queryFixture('<div class="baz">baz</div>');
+  it('returns `` when no accessible text method returned something', function () {
+    fixtureSetup('<div class="baz">baz</div>');
```
straker, line 25: Should replace this line and the querySelector for the class.

```js
    var div = queryFixture('<div class="baz" id="target">baz</div>');
```

straker (changes requested):

test/testutils.js
dbowling: This has been added, though my VS Code install doesn't display #target

test/testutils.js
dbowling: Done, and updated JSDoc to `@return {VirtualNode}`

test/checks/navigation/heading-order.js
dbowling: Updated in 36bbd049b1a1c2d7ff517a44d0f74317191fabbf

test/commons/text/native-text-alternative.js
dbowling: Updated in 9e4d1fdd41f947c111b9cf85783bc23f5ae2c5a9

straker (approved):

Issue #3563, opened 2022-07-26
Axe is struggling to determinate the background color when floats are used.


Zauberbutter:
I've found the following issue with the determination of background colors by axe. The source code contains two examples where it fails. I don't know if its fixable but I want to report it nevertheless.

**How to reproduce**:
<details>
<summary>HTML</summary>

```html

<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>Test</title>
    <style>
    *, ::after, ::before {
        -webkit-box-sizing: border-box;
        -moz-box-sizing: border-box;
        box-sizing: border-box;
    }
    body {
        font-family: Arial,Helvetica,sans-serif;
        font-size: 14px;
        line-height: 1.6em;
    }
    </style>
</head>
<body>

<ul style="list-style: none">
    <li style="display: table">
        <span style="margin-bottom: -1.7em;padding-left: 264px;display:block">Foo</span>
        <h5 style="margin:0">
            <a href="#">

                <img
                    alt=""
[...]

PR #3565, opened 2022-07-26
fix(color-contrast): correctly determine color contrast for element that exactly overlaps midpoint of node

straker:
In https://github.com/dequelabs/axe-core/pull/2003/files we changed the algorithm from `<` to `<=` to handle nodes returning as incomplete in the code highlighting. The tests added for `test/integration/full/contrast/code-highlighting.html` assert that only 1 incomplete is returned, so this change does not affect the reason we moved from `<` to `<=` in the first place.

Closes issue: #3563


straker:
Looks like Firefox in linux is giving the element an extra 2 px of height, so instead of the element being close to the midpoint (<1 px difference) it's now over the midpoint by a little more than a pixel. Not sure what we want to do about that at the moment.

lib/commons/dom/get-rect-stack.js
```diff
@@ -124,13 +126,27 @@ export function getRectStack(grid, rect, recursed = false) {
...
+            el: gridCellNode.actualNode.cloneNode(false).outerHTML,
+            y: y,
+            floorY: floorY,
+            'rectY': rectY,
+            height: clientRect.height,
+            'rectY+height': rectY + clientRect.height,
+            'floor(rectY+height)': floor(rectY + clientRect.height)
+          },null,2));
+        }
+
```
WilcoFiers, line 140:
```suggestion
```

test/checks/color/color-contrast.js
```diff
@@ -395,13 +395,25 @@ describe('color-contrast', function() {
...
     });
   });
 
-  describe('with pseudo elements', function() {
-    it('should return undefined if :before pseudo element has a background color', function() {
+  it('should return true for element which exactly overlaps midpoint of another element', function () {
+    var params = checkSetup(
+      '<div>' +
+        '<div id="target">' +
+        '<h4 id="debugNode">Foo</h4>' +
```
WilcoFiers, line 407:
```suggestion
        '<h4>Foo</h4>' +
```

test/checks/color/color-contrast.js
```diff
@@ -395,13 +395,25 @@ describe('color-contrast', function() {
...
+    assert.doesNotThrow(function () {
       contrastEvaluate.apply(checkContext, params);
     });
   });
 
-  describe('with pseudo elements', function() {
-    it('should return undefined if :before pseudo element has a background color', function() {
+  it('should return true for element which exactly overlaps midpoint of another element', function () {
+    var params = checkSetup(
+      '<div>' +
```
WilcoFiers, line 405: Should this be here? Why is this tag not closed?

test/checks/color/color-contrast.js
```diff
@@ -395,13 +395,25 @@ describe('color-contrast', function() {
...
     );
-    assert.doesNotThrow(function() {
+    assert.doesNotThrow(function () {
       contrastEvaluate.apply(checkContext, params);
     });
   });
 
-  describe('with pseudo elements', function() {
-    it('should return undefined if :before pseudo element has a background color', function() {
+  it('should return true for element which exactly overlaps midpoint of another element', function () {
```
WilcoFiers, line 403: I feel this test is too indirect. I think I'd rather see us test getRectStack directly.

lib/commons/dom/get-rect-stack.js
WilcoFiers, line 140:
```suggestion
```

WilcoFiers (changes requested):

lib/commons/dom/get-rect-stack.js
straker: Ops. forgot I still had those in when trying to figure out why IE11 was broken.

WilcoFiers (approved):

Issue #3208, opened 2021-10-13
ARIA 1.2 required owned elements when aria-busy=true


Zidious:
Ref: https://github.com/dequelabs/axe-core-npm/issues/292

> There may be times that required owned elements are missing, for example, while editing or while loading a data set. When a widget is missing required owned elements due to script execution or loading, authors MUST mark a containing element with aria-busy equal to true. For example, until a page is fully initialized and complete, an author could mark the document element as busy.

source: [ARIA 1.2 - Required Owned Elements](https://www.w3.org/TR/wai-aria-1.2/#mustContain)

Example provided: 

```JS
<div id="while-loading" aria-label="while-loading" role="list">
	<div class="loader" aria-busy="true"><Loading List... /></div>
</div>
```
Output: 
```
Fix any of the following: Expecting ARIA children role to be added: group, listitem
```

This results in axe marking this as incomplete. Do we want to assume if a required owned elements is missing but its child has `aria-busy=true` we pass it? Or better, update the incomplete message to inform the user that we cannot determine if element is accessible due to `aria-busy`?

WilcoFiers:
Revisiting this issue, as it is now also relevant for axe-core's consistency with the [ARIA required owned elements]() ACT rule.

I agree with Dylan, although I'm not sure we should try to do this all in one go. I'd prefer we do the required children thing first, and in a later stage figure out what to do with other rules.

Support for aria-busy still seems fairly limited. That's something to consider here as well.

dylanb:
I think they should just be ignored rather than generating more incompletes

straker:
@dylanb the [support for `aria-busy` isn't great](https://a11ysupport.io/tech/aria/aria-busy_attribute) (only works in JAWS), so I'm not sure if we want to completely ignore the issues.

straker:
Instead of a global `aria-busy` check, we could scope this to only look at required child rules (`list`, `required-children`, etc.). So if the parent element or child has `aria-busy` we can mark the rule as pass.

padmavemulapati:
Validated with the latest develop branch code base,
reproducing with 4.3.5 version as below: `aria-required-children` was into `incomeplete` result as well the the fix details is having `Fix any of the following:\n  Expecting ARIA children role to be added: group, listitem`
Now with `4.4.3` , seeing `aria-required-children` into failure as well fix information as 
[...]

PR #3569, opened 2022-07-27
feat(aria-required-children): add aria-busy check

pschwendy:
Closes issue: #3208

### DIFF
--- a/lib/checks/aria/aria-busy.json
+++ b/lib/checks/aria/aria-busy.json
@@ -1,0 +1,11 @@
+{
+  "id": "aria-busy",
+  "evaluate": "aria-busy-evaluate",
+  "metadata": {
+    "impact": "serious",
+    "messages": {
+      "pass": "Element has aria-busy tag",
+      "fail": "Element does not have aria-busy tag"
+    }
+  }
+}
--- a/lib/checks/aria/aria-busy.json
+++ b/lib/checks/aria/aria-busy.json
@@ -4,7 +4,7 @@
   "metadata": {
     "impact": "serious",
     "messages": {
-      "pass": "Element has aria-busy attribute",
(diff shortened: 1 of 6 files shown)


test/checks/aria/aria-busy.js
```diff
@@ -0,0 +1,23 @@
+describe('aria-busy', function() {
```
straker, line 1: An additional test making sure that it returns false if `aria-busy=false`

straker (changes requested):

lib/checks/aria/aria-busy.json
```diff
@@ -0,0 +1,11 @@
+{
+  "id": "aria-busy",
+  "evaluate": "aria-busy-evaluate",
+  "metadata": {
+    "impact": "serious",
+    "messages": {
+      "pass": "Element has aria-busy tag",
+      "fail": "Element does not have aria-busy tag"
```
WilcoFiers, line 8:
```suggestion
      "pass": "Element has aria-busy attribute",
      "fail": "Element has no aria-busy=\"true\" attribute"
```

WilcoFiers (changes requested):

lib/checks/aria/aria-busy.json
```diff
@@ -0,0 +1,11 @@
+{
+  "id": "aria-busy",
+  "evaluate": "aria-busy-evaluate",
+  "metadata": {
+    "impact": "serious",
+    "messages": {
+      "pass": "Element has aria-busy attribute",
```
WilcoFiers, line 7:
```suggestion
      "pass": "Element has an aria-busy attribute",
```

WilcoFiers (approved):

Issue #3561, opened 2022-07-25
Flakey tests


straker:
Recently our tests have been super flaky and timeout a lot. Going to start keeping track of the failing tests to see if I can figure out what's going on.

https://app.circleci.com/pipelines/github/dequelabs/axe-core/3825/workflows/b29a4944-bda0-40a8-8984-adc893a08c36/jobs/44782

https://app.circleci.com/pipelines/github/dequelabs/axe-core/3825/workflows/b29a4944-bda0-40a8-8984-adc893a08c36/jobs/44780

PR #3574, opened 2022-07-28
tests: support fail case for testutils.awaitNestedLoad

straker:
The `awaitNestedLoad` function never took a fail callback, so if the queue for whatever reason failed, it would go to the default fail function which was just to log the error to console. Since we don't see console logs when we run the full tests, and since the fail case was never reported back to the promise, our tests could easily timeout with no clue as to why.

Added a fail callback and made sure each promise in the isolated-env code (which [had a timeout](https://app.circleci.com/pipelines/github/dequelabs/axe-core/3855/workflows/8c57f2be-1832-4a49-b963-dc0f15416242/jobs/45110)) had a `catch` clause that reported the error back Mocha. Hopefully if the code does fail now we'll get a helpful error rather than a test timeout.

Part of #3561

### DIFF
--- a/test/integration/full/isolated-env/isolated-env.html
+++ b/test/integration/full/isolated-env/isolated-env.html
@@ -4,7 +4,7 @@
     <title>all rules test</title>
     <meta charset="utf8" />
     <meta http-equiv="refresh" content="foo" />
-    <meta name="viewport" content="maximum-scale=2;" />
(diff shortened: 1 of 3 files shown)


straker:
Reran multiple times. The test failures for IE11 vaildLangs that https://github.com/dequelabs/axe-core/pull/3571 should fix

stephenmathieson (approved):

Issue #2817, opened 2021-02-24
Align link-in-text-block with W3C's new understanding


WilcoFiers:
W3C has updated its understanding docs about how to deal with colored links in blocks of text. You can now pass if text has a 3;1 contrast with the link, regardless of hover / focus states for those links.

After this change we should change this rule not to be experimental anymore.

WilcoFiers:
This was released in 4.5, it seems we didn't QA it :(

padmavemulapati:
Vaidated with the latest axe-core develop branch code base:

Test script:
```
<p style="color: black">
    paragraph of text (fail: text color has insufficient contrast against
    paragraph)
    <a
      style="text-decoration: none; color: blue"
      href="#"
      id="fail-insufficient-text-contrast"
    >
      Link text</a
    >
  </p>
  
  <p style="color: black">
    paragraph of text (fail: background color has insufficient contrast against
    paragraph)
    <a
      style="text-decoration: none; color: black; background-color: pink"
      href="#"
      id="fail-insufficient-background-contrast"
    >
      Link text</a
    >
  </p>

  <h1>Incomplete tests</h1>

<p style="color: black; background-image: linear-gradient(red, yellow)">
  paragraph of text (incomplete: link has insufficient text contrast and
  paragraph has a background gradient)
  <a
    style="text-decoration: none; color: black"
    href="#"
    id="incomplete-low-contrast-parent-has-gradient"
  >
    Link test</a
  >
</p>



```

Incomeplete Result
![Image](https://user-images.githubusercontent.com/47409579/200480714-413cfa36-9980-4c2c-9510-50672100507f.png)


Failure Result
![Image](https://user-images.githubusercontent.com/47409579/200480762-b6568224-826f-4dcd-ba89-b0822bc14cbd.png)

PR #3575, opened 2022-07-28
fix(link-in-text-block): Update rule to match current guidance, revise tests

DaveTryon:
This PR updates the `link-in-text-block` rule as described in #2817. The new rule converts a single check into 2 checks (`link-in-text-block-style` and `link-in-text-block`), as suggested. We use `any`, so the rule passes if either rule passes.

`link-in-text-block-style` check:
- return `false` if the link is not actually a link (unchanged from before)
- return `true` if the link and the containing text block have visually distinct styling independent of color

`link-in-text-block` check:
- return `false` if the link is not actually a link (unchanged from before)
- return `true` if _either_ the foreground contrast ratio or the background contrast ratio is at least 3.0
- return `undefined` if we couldn't get enough data to make a good decision. Possible values of `messageKey`:
  - `default` if we fail to obtain the foreground color of either the link or the containing text block
  - `bgContrast` if we fail to obtain background colors of either the link or the containing text block and no value for `bgColor` was set via `incompleteData`
  - `bgImage`, `bgGradient`, `imgNode`, or `bgOverlap` if a value for `bgColor` was set via `incompleteData` when trying to obtain a background color
- return `false` with an appropriate value for `messageKey`:
  - `bgContrast` if the foreground color is unchanged and the background color changed
  - `fgContrast` if the foreground color has changed or if the background color is unchanged

Strings in the json file have been updated to include more information about the colors that were used to compute the ratios. Unit tests and integration tests have been updated (and slightly regrouped) to pin the updated behavior.

This covers the rule changes for #2817, but it leaves the rule categorized as experimental. Changing the rule category will need a separate change.


straker:
Thanks for the pr! We'll need @WilcoFiers to look over it, which may take awhile as he is out.

lib/checks/color/link-in-text-block.json
```diff
@@ -5,10 +5,14 @@
     "impact": "serious",
     "messages": {
       "pass": "Links can be distinguished from surrounding text in some way other than by color",
-      "fail": "Links need to be distinguished from surrounding text in some way other than by color",
+      "fail": {
+        "default": "Links need to be distinguished from surrounding text in some way other than by color",
```
WilcoFiers, line 9: Doesn't look like we can ever really hit this code path still. The only return false that gets to this is here: https://github.com/dequelabs/axe-core/pull/3575/files#diff-ca029eef8fd1995a0339c5f17452abc9b469d1f874a12d01ad89a2d1f42d819bR30 and that one shouldn't be hit either, since the matches should prevent non-block nodes from getting into the check.

lib/checks/color/link-in-text-block-evaluate.js
```diff
@@ -37,57 +37,58 @@ function linkInTextBlockEvaluate(node) {
...
-      });
-      incompleteData.clear();
-      return undefined;
-    }
+  if (textContrast >= 3.0 || backgroundContrast >= 3.0) {
+    return true;
+  }
+
+  // Message reflects the highest contrast ratio since it's the smallest fix
+  if (textContrast >= backgroundContrast) {
```
WilcoFiers, line 76: I see where you're coming from, but I feel this may not be the right direction. For a link with 1. no underline, and 2. bad contrast, this is going to end up telling it to improve the contrast. I think what might be better is that we pull the contrast testing into a separate check, and make those "any" checks, so we can report it like so:

> Fix one of the following:
> - The link has has no styling (such as underline) to distinguish it from the surrounding text
> - The link text insufficient contrast of 2:1 with the surrounding text. (Minimum contrast 3:1, link text: #ABC123, surrounding text: #CDE234)

For the background, I think that can be combined with the text color check. People don't use backgrounds on inline links all too often, we don't have to explicitly call it out unless there is a background on the link IMO. 

Alternatively, we could do the background and foreground thing in the same check, and report whichever has the highest number like you're doing here. Changing background colors on inline links is rare, so it may not be necessary to point that option out unless someone is already using this. Might help on performance a bit too.

Re performance; When splitting this check it'll be good to do some performance testing, make sure we don't significantly slow down this rule. Color contrast can be expensive after all. If it does slow down we'll probably have to look at doing some more caching / memoization.

WilcoFiers (changes requested):
We'll need integration tests for this.

WilcoFiers:
@DaveTryon Is this ready again?

lib/checks/color/link-in-text-block.json
```diff
@@ -1,14 +1,20 @@
...
+  "options": {
+    "requiredContrastRatio": 3
+  },
   "metadata": {
     "impact": "serious",
     "messages": {
       "pass": "Links can be distinguished from surrounding text in some way other than by color",
-      "fail": "Links need to be distinguished from surrounding text in some way other than by color",
+      "fail": {
+        "fgContrast": "The link has insufficient color contrast of ${data.contrastRatio}:1 with the surrounding text. (Minimum contrast is ${data.requiredContrastRatio}:1, link text: ${data.nodeColor}, surrounding text: ${data.parentColor})",
```
madalynrose, line 12: A thought that just occurred to me is that these messages mention color contrast when what we're actually analyzing is relative luminance contrast. Worth capturing in the error message to avoid confusion?

lib/checks/color/link-in-text-block-evaluate.js
```diff
@@ -37,57 +37,58 @@ function linkInTextBlockEvaluate(node) {
...
-        reason = 'bgContrast';
-      }
-      incompleteData.set('fgColor', reason);
-      this.data({
-        messageKey: incompleteData.get('fgColor')
-      });
-      incompleteData.clear();
-      return undefined;
-    }
+  if (textContrast >= 3.0 || backgroundContrast >= 3.0) {
```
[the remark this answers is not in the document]
DaveTryon: Moved the ratio into a check option

lib/checks/color/link-in-text-block-evaluate.js
```diff
@@ -37,57 +37,58 @@ function linkInTextBlockEvaluate(node) {
...
-    nodeColor = getForegroundColor(node);
-    parentColor = getForegroundColor(parentBlock);
+  }
+
+  // Capture colors, exiting early in case of error
+  var nodeColor = getForegroundColor(node);
+  var parentColor = getForegroundColor(parentBlock);
+
+  if (!nodeColor || !parentColor) {
+    return undefined;
```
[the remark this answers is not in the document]
DaveTryon: We now return success if the text color has adequate contrast, even if we can't access the background colors

lib/checks/color/link-in-text-block.json
DaveTryon: Remove the default case for failure

lib/checks/color/link-in-text-block-evaluate.js
DaveTryon: Refactored to use any. I'll reach out to you offline for details on how you generally handle perf testing

test/integration/rules/link-in-text-block/link-in-text-block.html
```diff
@@ -25,19 +26,107 @@
...
+  paragraph of text (incomplete: link has insufficient text contrast and
+  paragraphs has a background gradient)
+  <a
+    style="text-decoration: none; color: black"
+    href="#"
+    id="incomplete-low-contrast-parent-has-gradient"
+  >
+    Link test</a
+  >
 </p>
```
DaveTryon, line 132: @WilcoFiers, I added a local integration test for this where the `linear-gradient` style is on the link instead of the paragraph. The browser rendered it correctly, but the rule doesn't return `undefined`. Debugging showed that the call to [`elementHasImage`](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/color/element-has-image.js) is returning false. That seems strange to me, but I didn't try to fix it in this PR. Just for reference, here's the change that I tested:

```
<p style="color: black">
  paragraph of text (incomplete: link has insufficient text contrast and
  paragraphs has a background gradient)
  <a
    style="text-decoration: none; color: black; background-image: linear-gradient(red, yellow)"
    href="#"
    id="incomplete-low-contrast-parent-has-gradient"
  >
    Link test</a
  >
</p>
```

DaveTryon:
> @DaveTryon Is this ready again?

@WilcoFiers: Yes, please take another pass. I left a comment of some behavior that I didn't expect in the integration tests

WilcoFiers (changes requested):
Looks great. Apologies for the delayed review. I've been kind of swamped by WCAG 2.2 / 3.0 work the last couple weeks. I think we're almost ready to merge this in.

WilcoFiers (approved):
Reviewed for security. Thanks @DaveTryon!

PR #3577, opened 2022-07-28
refactor(accessible-text-virtual): finalize virtualNode

straker:
Noticed that the `accessible-text-virtual` functions didn't all completely use virtual node and sometimes accessed the `actualNode` just to look at the `nodeType` and call `isVisible`. Updated to just use virtual nodes.

### DIFF
--- a/lib/commons/text/accessible-text-virtual.js
+++ b/lib/commons/text/accessible-text-virtual.js
@@ -19,7 +19,6 @@ import isIconLigature from '../text/is-icon-ligature';
  * @return {string}
  */
 function accessibleTextVirtual(virtualNode, context = {}) {
-  const { actualNode } = virtualNode;
   context = prepareContext(virtualNode, context);
 
   // Step 2A, check visibility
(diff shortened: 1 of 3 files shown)


lib/commons/text/accessible-text-virtual.js
```diff
@@ -119,12 +118,11 @@ function shouldIgnoreIconLigature(virtualNode, context) {
  * @return {Object} context object with defaults applied
  */
 function prepareContext(virtualNode, context) {
-  const { actualNode } = virtualNode;
   if (!context.startNode) {
     context = { startNode: virtualNode, ...context };
   }
 
-  if (!actualNode) {
+  if (!virtualNode) {
```
WilcoFiers, line 125: virtualNode will not be falsey. This creates a dead code path. Might as well remove this if statement.

There is a very subtle change this introduces when computing the accessible name from aria-labelledby, referencing a visible element that then contains a hidden element. This change causes that hidden element to be excluded from the name. We should probably just add a test for this. Don't know if you meant to fix a bug, but you did.

WilcoFiers (dismissed):

lib/commons/text/accessible-text-virtual.js
straker: We believe this is the code path to test. Since before we exited before ever entering `isVisible`, this change allows the `isVisible` code to run, meaning we need to test when the text of the element is hidden:

```html
<div aria-labelledby="foo"></div>
<div id="foo">
  <span aria-hidden="true">thing</span>
</div>
```

WilcoFiers (approved):

PR #3579, opened 2022-08-01
chore: add issue forms for issues, feature requests, documentation, and questions

straker:
Recently discovered Github template forms and thought it would be a great addition to the repo over our singular issue template. [eleventy uses them](https://github.com/11ty/eleventy/issues/new/choose) and where I took a lot of inspiration on how to separate out the issues. The important aspect of this is being able to auto label different categories of issues (`feat`, `docs`, `questions`) and being able to change content based on the issue type.

You can see a running example at my test repo https://github.com/straker/issue-template-tests/issues/new/choose.

After this is merged, I want to create a Github action that will automatically add the `extension` or `linting` labels based on the value of the product dropdown.


.github/ISSUE_TEMPLATE/possible-bug.yml
```diff
@@ -0,0 +1,57 @@
...
+        Describe what you expected the product to do.
+
+        #### Actual
+        Describe what the product actually does.
+
+        #### How to Reproduce
+        Provide a code sample or link to a webpage that reproduces the issue. Without this, your issue may be closed without investigation.
+
+        #### Additional context
+        Any thing else we should know about the issue?
```
stephenmathieson, line 44: Users are going to `ctrl+a` and delete this (or at least want to). Why not create a separate field for each of these?

.github/ISSUE_TEMPLATE/possible-bug.yml
```diff
@@ -0,0 +1,57 @@
...
+    id: product-versions
+    attributes:
+      label: Product Versions
+      description: If applicable, please provide any additional information about the versions of products, browsers, or assistive technologies.
+      render: bash
+      value: |
+        axe-core version: X.Y.Z
+        axe Exension version: X.Y.Z
+        axe Linter version: X.Y.Z
+        Browser and Assistive Technology versions:
```
stephenmathieson, line 57: Should this just be `Product Version`? We already know which product (as result of the dropdown), so listing them again seems redundant.

stephenmathieson, line 57: Should this be a separate (optional) field?

stephenmathieson (commented):
LGTM. Minor suggestions left inline 👍

.github/ISSUE_TEMPLATE/possible-bug.yml
straker: I had it original as a separate field, but taking a look at past issues opened and not every bug needs every one of the fields (Wilco and I certainly don't need every field). It also felt a bit overwhelming to have 3 additional required fields just to fill in the bug. 

User are able to select a different template and use the blank issue one, but it's not prominent in the UI so I didn't want to discourage anyone from filing a bug.

stephenmathieson: Fair enough 🤷

stephenmathieson (approved):

.github/ISSUE_TEMPLATE/documentation.yml
```diff
@@ -0,0 +1,25 @@
+name: Improve our documentation
+description: Report issues in our documentation or things we could document better.
+labels: ['docs']
```
WilcoFiers, line 3: I'm not sure we should add tags this way. I'd rather we add tags as part of grooming. By adding tags now it'll be more difficult to tell which issues are groomed and which are not. Thoughts?

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,4 @@
+contact_links:
+  - name: Join the axe Slack Community
+    url: https://accessibility.deque.com/axe-community
+    about: Connect with other people passionate about digital accessibility and axe tools.
```
WilcoFiers, line 4: Nitpick, should these end with a newline? We do that everywhere else in the repo.

WilcoFiers (dismissed):

.github/ISSUE_TEMPLATE/documentation.yml
straker: That's a fair point. What do you think about adding the "label" as part of the title? would that be helpful at all?

```
[Docs]: <title>
[Feat]: <title>
```

We could also create a new label `needs-triage ` or `ungroomed` or something that we remove from tickets to help ensure we've groomed them.

WilcoFiers: Hmm, I guess "ungroomed" could work.

straker: Add ungroomed label https://github.com/dequelabs/axe-core/labels?q=ungroomed

WilcoFiers (approved):

Issue #3544, opened 2022-07-13
Warn during test when Node 12+ isn't used


WilcoFiers:
PR https://github.com/dequelabs/axe-core/pull/3538, axe-core is back to using Node v4 as its minimum. That's true enough when it comes to importing and using axe, but Node 12+ is required for building / testing. We should add an explicit warning of sorts to call out when an older version of Node is used to build / test axe.

I am very open to how to do this. Easiest might be to add some little script that returns exit code 1 in the `pretest` and `prebuild` scripts in package.json.

pandu-supriyono:
Hi, I have the time and motivation to look into this.

Do I understand correctly that we want to produce a fatal error with exit code 1 if the minimum Node version isn't found instead of a warning message that exits with exit code 0?

straker:
Thanks for taking this on @pandu-supriyono. I believe having the build script or test script exit with exit code 1 if Node version is < 12 would be fine. Adding a note in the developer guide on minimum Node version would also be very helpful.

PR #3580, opened 2022-08-01
chore: warn when node 12+ isn't used

pandu-supriyono:
Added a node script in the project root directory as opposed to creating a bin/scripts directory since there aren't any scripts of similar nature around (yet).

Closes issue: #3544


check-node-version.js
```diff
@@ -0,0 +1,32 @@
+#! /usr/bin/env node
+
+const minimumVersion = '12.0.0';
+const currentVersion = process.version.replace('v', '');
+
+const minimumVersionMajor = parseInt(minimumVersion.split('.')[0]);
```
stephenmathieson, line 6: Why not just do:

```suggestion
const minimumVersionMajor = 12
```

check-node-version.js
```diff
@@ -0,0 +1,32 @@
...
+const currentVersionMajor = parseInt(currentVersion.split('.')[0]);
+
+const usesMinimumVersion = currentVersionMajor >= minimumVersionMajor;
+
+if (usesMinimumVersion) {
+  process.exit();
+}
+
+console.log(
```
stephenmathieson, line 15: IMO this belongs on stderr:

```suggestion
console.error(
```

check-node-version.js
```diff
@@ -0,0 +1,32 @@
...
+    ', but you need ' +
+    'Node.js version ' +
+    minimumVersion +
+    ' or higher to build and/or test axe-core.' +
+    '\n\n' +
+    'Install Node.js version ' +
+    minimumVersion +
+    ' or higher and try again.' +
+    '\n\n' +
+    'You can use nvm (https://github.com/creationix/nvm) to update your Node.js version.' +
```
stephenmathieson, line 28: We ought to create an [`.nvmrc` file](https://github.com/nvm-sh/nvm/blob/master/README.md#nvmrc), so users can just `cd axe-core; nvm use`.

Would you mind creating one?

stephenmathieson (changes requested):
Thanks! This is great. Minor things left inline.

check-node-version.js
stephenmathieson: Not a blocker btw, just seems simpler to declare the variable as what we want (rather than dynamically building it).

pandu-supriyono:
@stephenmathieson thanks for having a look and the feedback! I can see to them tomorrow again.

One last request as I'd appreciate thoughts on the following:

Using an .nvmrc is a good idea -- in that case we could dynamically load that file into the check-node-version script and assign it to `minimumVersion`.

In my view, this would make maintenance easier since it constrains the source of truth with regards to the minimum node version to the .nvmrc file (excluding the docs).

.nvmrc
```diff
@@ -0,0 +1 @@
+12.0.0
```
stephenmathieson, line 1: We should not specify a minor or patch version. This will force users using 12.0.1 to change versions.

```suggestion
12
```

check-node-version.js
```diff
@@ -0,0 +1,42 @@
+#! /usr/bin/env node
+
+const fs = require('fs');
+const path = require('path');
+
+fs.readFile(path.join(__dirname, '.nvmrc'), 'utf-8', function (err, data) {
```
stephenmathieson, line 6: What is the benefit of using the async version of `readFile` here? This would be easier to maintain and faster if we used `fs.readFileSync()`.

doc/developer-guide.md
```diff
@@ -31,7 +31,7 @@ Axe 3.0 supports open Shadow DOM: see our virtual DOM APIs and test utilities fo
 
 ### Environment Pre-requisites
 
-1. You must have NodeJS installed.
+1. You must have NodeJS version 12 or higher installed.
```
stephenmathieson, line 34: Based on the current `.nvmrc`, this is not true; you must have 12.0.0 installed.

stephenmathieson (changes requested):

check-node-version.js
stephenmathieson: Nope, it'd make it Just Work™

straker: But wouldn't that force them to go back to using node 12 instead of a later version like 16 if that's their default?

pandu-supriyono: Looking back, I realize the implications this could have if someone does have an automatic script.

With an .nvmrc, you could warn if someone is not using 12.0.0. But in doing so, we are implying that 12.0.0 should be used for development (instead of at least/minimum 12.0.0).

straker:
@pandu-supriyono Let's go ahead and drop the `.nvmrc` file so we don't force users to use an older version of node. If you remove the file and revert the changes to the script to use the `.nvmrc` file, I think this should be good to merge.

Thanks again for working on this

pandu-supriyono:
@straker Got it -- done in [387c3fc](https://github.com/dequelabs/axe-core/commit/387c3fcd630f50ee8273d73a140b6b8e8dc87dd5) 

Thanks again 😃

straker (dismissed):
Verified the file runs as expected, just need to update the variable name to the right one. Other than that, could you move the file to the `build` directory? That houses all our node files that get run for various things.

straker:
Apologies on the back-and-forth trying to figure out the `.nvmrc` file. Do you think you'll have some time this week to address https://github.com/dequelabs/axe-core/pull/3580#pullrequestreview-1072878912?

WilcoFiers (approved):

WilcoFiers:
Oh, right! Reviewed for security.

Issue #3473, opened 2022-05-20
Migrate more single-node rules over to virtual nodes


WilcoFiers:
Virtual nodes are faster and let axe run synchronously. We migrated a good number of rules over last year, but here are a couple more that I think could be migrated fairly quickly:

- [x] aria-roledescription (https://github.com/dequelabs/axe-core/pull/3570)
- [x] aria-text (https://github.com/dequelabs/axe-core/pull/3581)
- [x] empty-table-header (https://github.com/dequelabs/axe-core/pull/3621)
- [x] empty-heading (https://github.com/dequelabs/axe-core/pull/3582)
- [x] presentation-role-conflict (https://github.com/dequelabs/axe-core/pull/3596)
- [x] table-duplicate-name (https://github.com/dequelabs/axe-core/pull/3604)

padmavemulapati:
After migrated `aria-roledescription` general fixture to virtual node, no change in the behaviour but refactered.
Script:
```
<button aria-roledescription="Awesome Button" id="target">Click</button>
  <div role="radio" aria-roledescription="Awesome Radio" id="target">Click</div>
  <div role="main" aria-roledescription="Awesome Main" id="target">The main element</div>
  <div aria-roledescription="Awesome Main" id="target">The main element</div>
  <div role="presentation" aria-roledescription="Awesome Main" id="target">The main element</div>
  <div role="none" aria-roledescription="Awesome Main" id="target">The main element</div>
```
<img width="1488" alt="image" src="https://user-images.githubusercontent.com/47409579/187441676-8f618825-c090-4ff3-8358-038e981b21b9.png">

PR #3581, opened 2022-08-02
refactor(aria-text): use virtualNode

dbowling:
* use virtual node

references: #3473


test/integration/virtual-rules/aria-text.js
```diff
@@ -0,0 +1,198 @@
...
+      nodeName: 'div',
+      attributes: {
+        role: 'text'
+      }
+    });
+    var child = new axe.SerialVirtualNode({
+      nodeName: 'button',
+      attributes: {
+        role: 'widget'
+      }
```
WilcoFiers, line 28: Did you maybe mean something like this?

```suggestion
      nodeName: 'div',
      attributes: {
        role: 'button'
      }
```

role=widget doesn't do anything, and putting it on a button makes this basically the same as your next test.

WilcoFiers (changes requested):

WilcoFiers (approved):

Issue #3464, opened 2022-05-18
Color contrast should ignore overlapping elements that have transparent background


straker:
Take the following example:

```html
<style>
.overlap {
  position: absolute;
  top: 0;
  left: 0;
  width: 100px;
  height: 100px;
}
</style>

<span>Hello World</span>
<div class="overlap"></div>
```

This causes an incomplete result for color contrast because the `.overlap` element is on top of the `<span>` element in the DOM stack. However since the `.overlap` element has a transparent background it should just be ignored. 

This same problem happens with [Material Button components](https://material.io/components/buttons) due to the ripple effect element that overlaps the button text.

```html
<button class="mdc-button mdc-button--raised">
  <span class="mdc-button__label">Button</span>
  <div class="mdc-button__ripple"></div>
</button>
```

straker:
We should look into restricting exactly what is ignored, as images or iframes that overlap won't have a background color but will interfere with the element. So possible solution is to ignore the element if it meets all of the following conditions:

* has a fully transparent background
* has no text content
* has a presentational or no role

WilcoFiers:
It also can't have CSS filters, thick border, pseudo elements, embedded graphical content like img, canvas, object. There are a ton of ways in which something that at first glance may seem like an empty div, for it to have an impact on content that's behind it.

I think that 9 out of 10 times when we see a transparent element on top of text content, it's there to have some kind of effect on things that are behind it. Why would people put completely empty divs on top of their content?

straker:
A good example of this is the home page of https://arstechnica.com/. They use empty anchor tags with `aria-hidden=true` and `tabindex=-1` to cover the entire card for the article links which makes every single node in the card return as needs review (165 total color contrast needs review items).

We could scope the requirement to ignore the element based on:

* does not have text content
* does not have a background color (i.e. transparent)
* does not have children or pseudo elements (pseudo element will be flagged anyway)
* does not have a border, outline, filter, etc.

straker:
Another place this happens on is Material Inputs https://material.angular.io/components/input/overview. The inputs each contain a span which in turn contains the label element for the input. The span overlaps the input, but the label does not. Since the span itself has no background and no direct text children, we should ignore it as well.

CHE1RON:
I know this was closed some time ago, but is there any chance to revive #3583 in order to prevent incomplete tests? 😕

WilcoFiers:
This was closed by mistake. We're going to look at this again.

[...]

PR #3583, opened 2022-08-04
fix(color-contrast): ignore overlapping elements that do not interfere with the background color

straker:
This should drastically decrease the number of overlapping element incompletes axe-core generates, especially for anyone who uses Material Button on their site. I moved `findPseudoElement` to it's own function so it could be used in both `color-contrast-evaluate` and `calculateObscuringElement`. 

I also moved `getBackgroundStack` to the top to match our default export style. The main parts of that file change are the addition of `isEmptyElement` that goes with the `isInlineDescendant` check.

Closes issue: #3464

### DIFF
--- a/lib/commons/color/get-background-stack.js
+++ b/lib/commons/color/get-background-stack.js
@@ -3,50 +3,37 @@ import elementHasImage from './element-has-image';
 import getOwnBackgroundColor from './get-own-background-color';
 import incompleteData from './incomplete-data';
 import reduceToElementsBelowFloating from '../dom/reduce-to-elements-below-floating';
+import visibleTextNodes from '../text/visible-text-nodes';
+import findPseudoElement from '../dom/find-pseudo-element';
+import { getNodeFromTree } from '../../core/utils';
 
 /**
- * Determine if element B is an inline descendant of A
- * @private
- * @param {Element} node
- * @param {Element} descendant
- * @return {Boolean}
+ * Get all elements rendered underneath the current element,
+ * In the order they are displayed (front to back)
+ *
+ * @method getBackgroundStack
+ * @memberof axe.commons.color
+ * @param {Element} elm
+ * @param {Object} options
+ * @return {Array}
  */
-function isInlineDescendant(node, descendant) {
-  const CONTAINED_BY = Node.DOCUMENT_POSITION_CONTAINED_BY;
-  // eslint-disable-next-line no-bitwise
-  if (!(node.compareDocumentPosition(descendant) & CONTAINED_BY)) {
-    return false;
-  }
-  const style = window.getComputedStyle(descendant);
-  const display = style.getPropertyValue('display');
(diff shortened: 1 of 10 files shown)


lib/commons/dom/find-pseudo-element.js
```diff
@@ -0,0 +1,71 @@
...
+  pseudo
+) {
+  const style = window.getComputedStyle(node, pseudo);
+  const matchPseudoStyle = (prop, value) =>
+    style.getPropertyValue(prop) === value;
+  if (
+    matchPseudoStyle('content', 'none') ||
+    matchPseudoStyle('display', 'none') ||
+    matchPseudoStyle('visibility', 'hidden') ||
+    matchPseudoStyle('position', 'absolute') === false
```
WilcoFiers, line 42: This line works well enough when this was part of the color-contrast-evaluate method, but I'm not so sure about this now that it's in the commons.

lib/commons/color/get-background-stack.js
```diff
@@ -95,31 +82,89 @@ function sortPageBackground(elmStack) {
...
+ * @see https://github.com/dequelabs/axe-core/issues/3464
+ * @private
+ * @param {Element} node
+ * @param {Object} options
+ */
+function isEmptyElement(node, { pseudoSizeThreshold, ignorePseudo }) {
+  const vNode = getNodeFromTree(node);
+  const style = window.getComputedStyle(node);
+  const bgColor = getOwnBackgroundColor(style);
+  const hasPseudoElement = !!findPseudoElement(vNode, {
```
WilcoFiers, line 144: This method doesn't seem right here. For example, if the pseudo element isn't position:absolute, `hasPsuedoElement` would become false, even if there's a background color, or content, or a bunch of other things on that.

I'm not sure how to fix that, except perhaps maybe have this be "true" unless `content:none` is true for both `:before` and `:after`? I'm not sure that `findPseudoElement` should have been added to `commons`. There are a bunch of things it does that make sense for inside color-contrast-evaluate, but that create problematic gotchas if used in other places.

test/commons/dom/find-pseudo-element.js
```diff
@@ -0,0 +1,169 @@
...
+        );
+        var actual = findPseudoElement(vNode);
+        assert.equal(vNode, actual);
+      });
+
+      it('should return a parent element with a pseudo element', function () {
+        var vNode = queryFixture(
+          '<style>.parent' +
+            pseudoElm +
+            '{ content: ""; width: 100%; height: 100px; background: red; position: absolute; }</style><div class="parent"><div><div id="target" style="height: 100px;"></div></div></div>'
```
WilcoFiers, line 35: Would you mind putting in a line break into these longer lines?

WilcoFiers (changes requested):

straker:
Closing as I don't think we'll get to this any time soon, pr code will still be here if we need it.

PR #3589, opened 2022-08-09
chore: create action to auto-label extension and linting issues

straker:
With the new issue forms, we can now auto-assign opened issues with the `extension` or `linting`  labels based on the Product dropdown. It was difficult to figure out the proper matching of newline characters. 

You can see it working https://github.com/straker/issue-template-tests/issues and https://github.com/straker/issue-template-tests/actions

### DIFF
--- a/.github/workflows/label-extension-linter-issues.yml
+++ b/.github/workflows/label-extension-linter-issues.yml
@@ -3,8 +3,6 @@ name: Add extension and linting labels to associated opened issues
 on:
   issues:
     types: [opened]
-env:
-  MY_GITHUB_TOKEN: ${{ secrets.MY_GITHUB_TOKEN }}
 
 jobs:
   label_extension_linting_issues:
--- a/.github/workflows/label-extension-linter-issues.yml
+++ b/.github/workflows/label-extension-linter-issues.yml
@@ -1,0 +1,26 @@
+name: Add extension and linting labels to associated opened issues
+
+on:
+  issues:
+    types: [opened]
+env:
+  MY_GITHUB_TOKEN: ${{ secrets.MY_GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/label-extension-linter-issues.yml
```diff
@@ -0,0 +1,26 @@
+name: Add extension and linting labels to associated opened issues
+
+on:
+  issues:
+    types: [opened]
+env:
+  MY_GITHUB_TOKEN: ${{ secrets.MY_GITHUB_TOKEN }}
```
stephenmathieson, line 7: This isn't used.

```suggestion
```

.github/workflows/label-extension-linter-issues.yml
```diff
@@ -0,0 +1,26 @@
...
+
+jobs:
+  label_extension_linting_issues:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - name: Label Extension Issues
+        uses: andymckay/labeler@1.0.4
+        if: contains(toJson(github.event.issue.body), '### Product\n\naxe Extension\n\n')
```
stephenmathieson, line 17: Why convert this to JSON?

stephenmathieson (dismissed):

.github/workflows/label-extension-linter-issues.yml
straker: It was the only way to get matching of newlines to work. Just the body would not match any newline character combinations I tried (`\n`, `\r\n`, etc.). https://stackoverflow.com/questions/61238849/github-actions-if-contains-function-not-working-with-env-variable

WilcoFiers (approved):

Issue #3485, opened 2022-05-31
frame-title rule should account for presentational conflict


WilcoFiers:
The following should fail the frame-title rule:

```html
<iframe title=" " src="/test-assets/SC4-1-2-frame-doc.html" role="none"> </iframe>
```

https://wai-wcag-act-rules.netlify.app/standards-guidelines/act/rules/cae760/proposed/#failed-example-4

straker:
To resolve this, we need to update the `isNativelyFocusable` code to say that `iframe` and `frame` elements are natively focusable. That will then trigger the conflict resolution code in the `getRole` function.

padmavemulapati:
Validated with the latest axe-core develop branch code base, the test script:
` <iframe title=" " src="/test-assets/SC4-1-2-frame-doc.html" role="none"> </iframe>`
is failing for `frame-title` rule


![Image](https://user-images.githubusercontent.com/47409579/191693110-08b16da3-106d-44ee-bd5d-0acd086960f4.png)

PR #3594, opened 2022-08-10
fix(frame-title): return incomplete for presentational iframe with empty title

straker:
Closes issue: #3485


test/integration/rules/frame-title/frame-title.html
```diff
@@ -64,6 +54,16 @@
...
 ></iframe>
+<iframe
+  src="/integration/rules/frame-title/frames/level1a.html"
+  id="violation8"
+  role="presentation"
+></iframe>
+<iframe
+  src="/integration/rules/frame-title/frames/level1a.html"
+  id="violation9"
+  role="none"
```
WilcoFiers, line 65: Add in the test from the issue. I'd like to see that working.

```suggestion
  role="none"
  title="   "
```

lib/commons/dom/is-natively-focusable.js
```diff
@@ -31,6 +31,7 @@ function isNativelyFocusable(el) {
     case 'select':
     case 'summary':
     case 'button':
+    case 'iframe':
```
WilcoFiers, line 34: I would also like to enable the `cae760` ACT rule (Iframe element has non-empty accessible name). With this fix in, that rule should be consistent.

WilcoFiers (changes requested):

WilcoFiers (approved):

Issue #3468, opened 2022-05-19
aria-required-children should not permit mixing child roles


WilcoFiers:
Axe-core's aria-required-children rule shouldn't pass the following:

```html
<div role="list">
  <li>Item 1</li>
  <span role="link">Item 2</span>
</div>
```

WAI-ARIA is a little vague about whether or not you're allowed to do this. There are scenarios, such as for radiogroup where owned elements might not all need to be of the required owned elements type, but for something like lists and trees, that doesn't seem right.

Failing the above example is required for consistency with the [ARIA required owned elements](https://www.w3.org/WAI/standards-guidelines/act/rules/bc4a75/proposed/) ACT rule.

straker:
We can do this by filtering the `ownedRoles` array and seeing which ones are not required

```js
const invalidRoles = ownedRoles.filter(role => !required.includes(role));
  if (invalidRoles) {
    this.data({

    })
    return false;
  }
```

padmavemulapati:
Validated with the latest develop branch code base, the script below is failing `arai-required-children` rule
```
<div role="list">
    <li>Item 1</li>
    <span role="link">Item 2</span>
  </div>
```
<img width="1446" alt="image" src="https://user-images.githubusercontent.com/47409579/187229819-9bd2052c-17fa-4ee5-bd19-e720f1cbde38.png">

Issue #3462, opened 2022-05-18
aria-required-children: be less permissive of divs as child nodes


WilcoFiers:
ARIA required context role, failed example 3

In the [ARIA required context role](https://w3.org/WAI/standards-guidelines/act/rules/ff89c9/proposed) ACT rules, [failed example 3](https://www.w3.org/WAI/content-assets/wcag-act-rules/testcases/ff89c9/52508dc0ac389108301d7cbd7f931be45a45741f.html) currently isn't being failed by axe-core. Specifically, I think aria-required-children should have caught this:

```html
<div role="list">
  <div aria-live="polite">
    <div role="listitem">List item 1</div>
    <div role="listitem">List item 2</div>
  </div>
</div>
```

Because of its global attribute, that `div[aria-live]` should not be ignored in the accessibility tree. How browsers do this varies. Firefox considers this a `section` inside a `list`.

straker:
Should fail due to #3468, just need to add to the `getOwnedRoles` function to check for tabindex > -1 or is focusable (should we make this it's own function called `includedInAccessibilityTree`? We do this type of check often enough it might make sense). Adding `null` to the list or return roles should be sufficient in this case to trigger the error message.

padmavemulapati:
Validated with the latest develop branch code base,
`aria-required-children` failing now for the child elements which are unlisted

<img width="1328" alt="image" src="https://user-images.githubusercontent.com/47409579/187247064-c6687b4b-ca2a-4ab7-ba57-256d518e2536.png">
 
and with 4.3.5 we are reproducing the issue

PR #3597, opened 2022-08-10
feat(aria-required-children): fail for children which are not listed as required

straker:
Closes issue: #3468 and #3462

### DIFF
--- a/lib/checks/aria/aria-required-children-evaluate.js
+++ b/lib/checks/aria/aria-required-children-evaluate.js
@@ -28,7 +28,7 @@ function getOwnedRoles(virtualNode, required) {
     // this means intermediate roles between a required parent and
     // child will fail the check
     if (
-      (!role && !includedInAccessibilityTree) ||
+      (!role && ! hasGlobalAriaOrFocusable) ||
       (['group', 'rowgroup'].includes(role) &&
         required.some(requiredRole => requiredRole === role))
     ) {
--- a/lib/checks/aria/aria-required-children-evaluate.js
+++ b/lib/checks/aria/aria-required-children-evaluate.js
@@ -4,7 +4,8 @@ import {
   getExplicitRole,
   getOwnedVirtual
 } from '../../commons/aria';
-import { hasContentVirtual, idrefs } from '../../commons/dom';
+import { getGlobalAriaAttrs } from '../../commons/standards';
+import { hasContentVirtual, idrefs, isFocusable } from '../../commons/dom';
 
 /**
  * Get all owned roles of an element
@@ -16,18 +17,24 @@ function getOwnedRoles(virtualNode, required) {
     const ownedElement = ownedElements[i];
     const role = getRole(ownedElement, { noPresentational: true });
 
+    const hasGlobalAria = getGlobalAriaAttrs().some(attr =>
+      ownedElement.hasAttr(attr)
+    );
+    const includedInAccessibilityTree =
+      hasGlobalAria || isFocusable(ownedElement);
+
     // if owned node has no role or is presentational, or if role
     // allows group or rowgroup, we keep parsing the descendant tree.
     // this means intermediate roles between a required parent and
     // child will fail the check
     if (
-      !role ||
+      (!role && !includedInAccessibilityTree) ||
       (['group', 'rowgroup'].includes(role) &&
         required.some(requiredRole => requiredRole === role))
     ) {
       ownedElements.push(...ownedElement.children);
-    } else if (role) {
-      ownedRoles.push(role);
+    } else if (role || includedInAccessibilityTree) {
+      ownedRoles.push({ role, ownedElement });
     }
   }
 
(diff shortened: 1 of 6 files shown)


test/integration/rules/aria-required-children/aria-required-children.html
```diff
@@ -76,8 +76,18 @@
...
+</div>
+<div role="list" id="fail12">
+  <div aria-live="polite">
+    <div role="listitem">List item 1</div>
+    <div role="listitem">List item 2</div>
+  </div>
+</div>
 
-<div role="list" id="incomplete10">
+<div role="list" id="fail13">
```
WilcoFiers, line 90: I don't think this should fail. Could keep incomplete, or pass. I'd guess @dylanb would rather see us just passing here.

lib/checks/aria/aria-required-children.json
```diff
@@ -21,7 +21,8 @@
       "pass": "Required ARIA children are present",
       "fail": {
         "singular": "Required ARIA child role not present: ${data.values}",
-        "plural": "Required ARIA children role not present: ${data.values}"
+        "plural": "Required ARIA children role not present: ${data.values}",
+        "unallowed": "Element has children which are not allowed"
```
WilcoFiers, line 25: I wonder if we should maybe do something like this:
```suggestion
        "unallowed": "Element has children which are not allowed (see related nodes)"
```

test/integration/rules/aria-required-children/aria-required-children.html
```diff
@@ -76,8 +76,18 @@
...
+<div role="list" id="fail11">
+  <li>Item 1</li>
+  <span role="link">Item 2</span>
+</div>
+<div role="list" id="fail12">
+  <div aria-live="polite">
+    <div role="listitem">List item 1</div>
+    <div role="listitem">List item 2</div>
+  </div>
+</div>
```
WilcoFiers, line 88: I would like to see some more tests thrown at this. Looking at the code, i don't think these work yet.

```suggestion
</div>

<div role="list" id="pass14">
  <script></script>
  <style></style>
  <canvas hidden></canvas>
  <div><div role="listitem">Item 1</div></div>
  <div role="generic"><div role="listitem">item 2</div></div>
  <div role="presentation"><div role="listitem">item 2</div></div>
  <div role="none"><div role="listitem">item 2</div></div>
</div>

<div id="radiogroup" id="pass15">
  <div role="heading" aria-level="2">Heading</div>
  <ul>
    <li><div role="radio"></div></li>
    <li><div role="radio"></div></li>
  </ul>
</div>
```

lib/checks/aria/aria-required-children-evaluate.js
```diff
@@ -16,18 +17,24 @@ function getOwnedRoles(virtualNode, required) {
     const ownedElement = ownedElements[i];
     const role = getRole(ownedElement, { noPresentational: true });
 
+    const hasGlobalAria = getGlobalAriaAttrs().some(attr =>
+      ownedElement.hasAttr(attr)
+    );
+    const includedInAccessibilityTree =
+      hasGlobalAria || isFocusable(ownedElement);
```
WilcoFiers, line 24: Wrong name. I think what you're testing for is presentational role conflict, not if it's in the tree.

WilcoFiers (changes requested):

test/integration/rules/aria-required-children/aria-required-children.html
straker: We want to handle this by removing the `requiredOwned` from the `radiogroup` role and allowing `menu` as a required child of `menubar`.

lib/checks/aria/aria-required-children-evaluate.js
```diff
@@ -16,18 +17,24 @@ function getOwnedRoles(virtualNode, required) {
     const ownedElement = ownedElements[i];
     const role = getRole(ownedElement, { noPresentational: true });
 
+    const hasGlobalAria = getGlobalAriaAttrs().some(attr =>
+      ownedElement.hasAttr(attr)
+    );
+    const includedInAccessibilityTree =
```
straker, line 23:
```suggestion
    const hasGlobalAriaOrFocusable =
```

lib/checks/aria/aria-required-children-evaluate.js
```diff
@@ -16,18 +17,24 @@ function getOwnedRoles(virtualNode, required) {
...
+      hasGlobalAria || isFocusable(ownedElement);
+
     // if owned node has no role or is presentational, or if role
     // allows group or rowgroup, we keep parsing the descendant tree.
     // this means intermediate roles between a required parent and
     // child will fail the check
     if (
-      !role ||
+      (!role && !includedInAccessibilityTree) ||
```
straker, line 31:
```suggestion
      (!role && ! hasGlobalAriaOrFocusable) ||
```

lib/checks/aria/aria-required-children-evaluate.js
```diff
@@ -16,18 +17,24 @@ function getOwnedRoles(virtualNode, required) {
...
       (['group', 'rowgroup'].includes(role) &&
         required.some(requiredRole => requiredRole === role))
     ) {
       ownedElements.push(...ownedElement.children);
-    } else if (role) {
-      ownedRoles.push(role);
+    } else if (role || includedInAccessibilityTree) {
```
straker, line 36:
```suggestion
    } else if (role || hasGlobalAriaOrFocusable) {
```

lib/checks/aria/aria-required-children.json
straker: This would be a first I believe to point to the related nodes. Is that a pattern we want to adopt?

test/integration/rules/aria-required-children/aria-required-children.html
straker: There's not a current way to make an exception to this since it's using a global aria attribute

WilcoFiers (approved):

WilcoFiers (approved):

PR #3598, opened 2022-08-11
ci: use correct environment variable

michael-siek:
Ref: https://github.com/srggrs/assign-one-project-github-action/issues/65

### DIFF
--- a/.github/workflows/auto-assign-issues.yml
+++ b/.github/workflows/auto-assign-issues.yml
@@ -14,5 +14,5 @@ jobs:
       - name: Auto assign axe-core integration issues to Axe API team project board
         uses: srggrs/assign-one-project-github-action@1.2.1
         with:
-          project: https://github.com/orgs/dequelabs/projects/66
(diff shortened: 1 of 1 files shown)


stephenmathieson (changes requested):
Why?

stephenmathieson:
See https://github.com/dequelabs/axe-core/runs/7757770044?check_suite_focus=true. It appears the issue is not the lack of quotes, but something with `jq` instead:

image

straker:
Cause the action is currently failing and we don't know why other than looking at the docs they use single quotes for the url so we're trying that to see if it works https://github.com/dequelabs/axe-core/runs/7757770044?check_suite_focus=true

stephenmathieson:
Wrapping a string in quotes is redundant in YAML. Look at the logs from that action:

image


Note that the URL _is_ wrapped in quotes.

straker:
We found the issue https://github.com/srggrs/assign-one-project-github-action/issues/82. Looks like we need to switch to a different action

PR #3599, opened 2022-08-11
ci: remove auto assign issues

michael-siek:
Have to remove this due to the project board being used is beta not classic
https://github.com/srggrs/assign-one-project-github-action/issues/82


.github/workflows/auto-assign-issues.yml
```diff
@@ -1,18 +0,0 @@
-name: Auto assign axe-core integration issues to Axe API team project board
-
-on:
-  issues:
-    types: [opened]
-env:
-  MY_GITHUB_TOKEN: ${{ secrets.MY_GITHUB_TOKEN }}
```
stephenmathieson, line 7: My guess is the reason this doesn't work is the token doesn't have sufficient permissions/scopes.

Also, I recommend using https://github.com/actions/add-to-project instead. It was authored by GitHub, so it probably works.

WilcoFiers (approved):

PR #3600, opened 2022-08-11
feat(doc-has-single-title): add doc-has-single-title

cmswalker:
This adds a check that the html document only has 1 title. It's impossible to assign multiple titles to the document api, however it is still possible to have multiple `<title>` elements within the HTML itself, which is a bad practice

### DIFF
--- a/lib/checks/shared/doc-has-single-title-evaluate.js
+++ b/lib/checks/shared/doc-has-single-title-evaluate.js
@@ -1,0 +1,6 @@
+function docHasSingleTitleEvaluate() {
+  var titles = document.getElementsByTagName('title');
+  return titles.length === 1;
+}
+
+export default docHasSingleTitleEvaluate;
--- a/lib/checks/shared/doc-has-single-title.json
+++ b/lib/checks/shared/doc-has-single-title.json
@@ -1,0 +1,11 @@
+{
+  "id": "doc-has-single-title",
+  "evaluate": "doc-has-single-title-evaluate",
+  "metadata": {
+    "impact": "serious",
+    "messages": {
+      "pass": "Document has only one <title> element",
+      "fail": "Document does not have only one <title> element"
+    }
+  }
+}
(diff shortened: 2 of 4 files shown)


lib/checks/shared/doc-has-single-title-evaluate.js
```diff
@@ -0,0 +1,6 @@
+function docHasSingleTitleEvaluate() {
+  var titles = document.getElementsByTagName('title');
+  return titles.length === 1;
```
WilcoFiers, line 3: There's already a rule that checks there is a single title. We don't need to fail missing titles twice.

```suggestion
  return titles.length >= 1;
```

WilcoFiers (changes requested):
Howdy! I really appreciate the pull request. I think this is a good idea for a best practice rule, but you've so far only added the check. Each rule runs some number of checks, but a check in itself won't do anything. You can look at lib/rules/document-title.json for an example.

lib/checks/shared/doc-has-single-title-evaluate.js
cmswalker: Which rule is that? `doc-has-title`? It checks the document API, not the html/DOM. So if I added `>= 1` here we'd be allowing more than 1 title to exist in the html, which is bad practice

cmswalker: I could edit/update `doc-has-title` to assert both the document api and the markup, but this might cause some breaking changes as it edits a rule

cmswalker: I think adding multiple checks to `doc-has-title` would be a bit easier vs separate rule

WilcoFiers: @cmswalker Wouldn't work I'm afraid. The `document-title` rule is required for WCAG, whereas this would be a best practice. We have to report them differently, we can only do that by putting them into separate rules.

WilcoFiers:
@cmswalker I just wanted to check in with you. Have you had a chance to look at this PR again?

WilcoFiers:
It's been over a month. I'm going to close this PR due to inactivity. If anyone wants to pick this back up they are welcome to open a new PR.

Issue #3276, opened 2021-11-12
Add generator method to cache.get()


WilcoFiers:
There are a good bunch of places in axe where we have a pattern like this:

```js
let x;
if (cache.get('x') === undefined) {
  x = 12345
  cache.set('x', x);
} else {
  x = cache.get('x')
}
```

I think a cleaner way that express this would be with a method that either returns the value from cache, or generates it if the value isn't cached. I would like to see us replace those patterns with a slightly higher-level solution like this:

```js
const x = cache.pull('x', () => 12345);
```

In other words. If `x` is defined, get it from the cache. If it's not, run the callback, store that as `x`, and then return it. This has a few advantages. Firstly, it's more expressive. It's also shorter, and I think it discourages using `set` and `get` in different places, using the cache as some kind of global scope, which is not a good practice.

padmavemulapati:
tech debt.. No QA required

padmavemulapati:
Moved it to "In Progress" as nothing merged against this ticket yet.

straker:
Completed in https://github.com/dequelabs/axe-core/pull/3608

PR #3608, opened 2022-08-15
feat(cache): add default param to cache.get()

dbowling:
- `cache.get()` gains second parameter, enabling a default value to be defined on first call for a given key

Closes issue: #3276


dbowling:
@straker I had to spread the params of `get()` to account for the possibility of an `undefined` default value. This does change the way the cache behaves a bit though--any params passed in excess of the first were ignored, and passing no params gave you some unexpected behavior.

It could be smart to add a guard to `get()` and throw if there are the wrong number of args, considering how cache methods can be a real pain if they aren't predictable. Adding a throw, though, feels like it is at a greater risk of introducing a breaking change than adding this feature without.

If we do want to add it, here's the patch:

```
diff --git a/lib/core/base/cache.js b/lib/core/base/cache.js
index ec414ec6..d0773055 100644
--- a/lib/core/base/cache.js
+++ b/lib/core/base/cache.js
@@ -21,6 +21,10 @@ const cache = {
     const args = [...arguments];
     const hasDefaultValue = args.length === 2;
 
+    if (args.length > 2 || args.length === 0) {
+      throw new Error('incorrect number of arguments');
+    }
+
     if (
       hasDefaultValue &&
       !(key in _cache) &&
diff --git a/test/core/base/cache.js b/test/core/base/cache.js
index cb0befe9..b9f51018 100644
--- a/test/core/base/cache.js
+++ b/test/core/base/cache.js
@@ -96,4 +96,18 @@ describe('axe._cache', function () {
   it('should return undefined for key lookup miss when no default value is provided', function () {
     assert.strictEqual(typeof axe._cache.get('foo'), 'undefined');
   });
+
+  it('should throw if there are too many arguments', function () {
+    function fn() {
+      axe._cache.get('foo', 'bar', 'baz');
+    }
+    assert.throws(fn);
+  });
+
+  it('should throw if there are no arguments', function () {
+    function fn() {
+      axe._cache.get();
+    }
+    assert.throws(fn);
+  });
 });
```

WilcoFiers (changes requested):
I'd like to see this get used as well. No point adding this if we're not going to use it.

lib/core/base/cache.js
```diff
@@ -13,9 +13,24 @@ const cache = {
...
    * Retrieve an item from the cache.
    * @param {String} key - Name of the key the value was stored as.
+   * @param {*} defaultValue - Default value to set if there is a cache miss. Functions are evaluated before caching. To override a value already saved, use `set()`.
    * @returns {*} The item stored
    */
-  get(key) {
+  get(key, defaultValue) {
+    // undefined is a valid cache value, so figure this out from args instead of a simple truthy check
+    const args = [...arguments];
+    const hasDefaultValue = args.length === 2;
```
WilcoFiers, line 22: This seems simpler:

```suggestion
  get(key, ...args) {
    const defaultValue = args[0];
    const hasDefaultValue = args.length === 1;
```

straker:
I would argue that it is not valid to set the cache with a value of `undefined`. `undefined` is the only way we know if the cache was not set before, so it should not be a valid default argument value either. That is, passing `undefined` would do nothing.

dbowling:
Although passing `undefined` and passing nothing are equivalent when reading the value directly, they are not equivalent in this PR because of the arg spreading, which can determine the difference between the caller's usage. In one case it writes the key with a value of `undefined`, and in the other (no second arg passed) it would not set the key at all. 

The `set()` function does not distinguish though. It writes to the cache with no checks.

I would propose that if we consider `undefined` to not be a valid cache value, we codify that decision in both the `get()` and `set()` logic and tests as the current behavior would then be a bug. Probably via a separate issue/PR for the fix, and putting this on-hold until it is merged into develop.

straker:
I rescind my comment on `undefined being a valid value.

lib/core/base/cache.js
```diff
@@ -13,9 +13,24 @@ const cache = {
...
+
+    if (
+      hasDefaultValue &&
+      !(key in _cache) &&
+      typeof defaultValue === 'function'
+    ) {
+      this.set(key, defaultValue.call());
```
straker, line 29:
```suggestion
      this.set(key, defaultValue());
```

lib/core/base/cache.js
```diff
@@ -13,9 +13,24 @@ const cache = {
...
+    } else if (hasDefaultValue && !(key in _cache)) {
+      this.set(key, defaultValue);
+    }
```
straker, line 32:
```suggestion
    if (hasDefaultValue && !(key in _cache)) {
      this.set(key, typeof defaultValue === 'function' ? defaultValue() : defaultValue);
    }
```

WilcoFiers (changes requested):
This: https://github.com/dequelabs/axe-core/pull/3608#pullrequestreview-1074064357

dbowling:
The following files in `lib` use `cache.set()`, but I elected not to include them in the refactor

- `lib/core/utils/get-flattened-tree.js` - only sets new values and does not use return
- `lib/core/public/run/globals-setup.js `- set needs to overwrite
- `lib/checks/generic/page-no-duplicate-evaluate.js` - refactor of this would likely be substantially harder to read

lib/core/base/cache.js
```diff
@@ -13,9 +13,24 @@ const cache = {
   /**
    * Retrieve an item from the cache.
    * @param {String} key - Name of the key the value was stored as.
+   * @param {*} defaultValue - Default value to set if there is a cache miss. Functions are evaluated before caching. To override a value already saved, use `set()`.
    * @returns {*} The item stored
    */
-  get(key) {
+  get(key, ...args) {
+    const defaultValue = args[0];
+    const hasDefaultValue = args.length === 1;
```
stephenmathieson, line 21: Why do we need to rely on arity here? This smells like magic.

stephenmathieson: Also, what if someone accidentally passes 2 parameters to this function?

dbowling: Since `undefined` is treated as a value that could be cached, we can't just check for `undefined` of the second parameter. 

The best way I can think of doing it is to look at the length of arguments passed by the caller by spread or `arguments.length`.

The params have had a couple different ways of getting defined suggested. Personally, I'm not a huge fan of using the spread of args as the second parameter since only the first value would be used. I'll toss another one and see if that floats higher in the water for us:

```suggestion
    get(key, defaultValue) {
        const hasDefaultValue = arguments.length === 2;
```
IMO, a named parameter instead of `...args` addresses most of the magic smell. The other thing that seems more obvious to me is to make the second parameter an options object. `options = { default: null }`

If someone sends a second argument in accidentally, they would be setting the cache for that key if the key was not previously defined. It does not override an existing cache value though.

We could also leave `cache.get()` untouched and add Wilco's requested logic to a new method, something like `cache.getWithDefaultValue()`.

```suggestion
    getWithDefaultValue(key, defaultValue) {
        if( !get(key) ){
            set(key, defaultValue); 
            //plus function detection with .call()
        };
        return get(key);
```

stephenmathieson: Why would we need to cache `undefined`?

dbowling: My understanding is that Axe throws around `undefined` as a meaningful value by convention, and we may want to be caching that return. `cache.set()` will let you do so currently.

dbowling:
I converted this PR back to draft while I work through the expanded scope of updating `lib` to use the default value poperty.

lib/core/base/cache.js
stephenmathieson: axe-core uses `undefined` to denote an incomplete check. I can't think of any reason we'd cache that though.

Unless we have an actual use case for doing `cache.set('foo', undefined)`, I do not think we should support it.

stephenmathieson:
What is the benefit of passing a default value? Why can't callers do:

```
thing = cache.get('thing') || defaultValue
```

lib/core/base/cache.js
dbowling: Tagging @WilcoFiers and @straker

dbowling:
Stephen, that gets back to what Wilco put in https://github.com/dequelabs/axe-core/issues/3276.

Your example `thing = cache.get('thing') || defaultValue` is close, but would need to also set the default value to the cache. In practice, that creates a pattern that us trying to be generalized.

Would this be more clear if we left `get()` as is, and added the new function call? This is a bit wordy, but gets the discussion started: `cache.getAndSetDefaultIfNotSet(key, default)`?

stephenmathieson:
Looks like the issue is asking for:

```js
var thing = cache.pull('thing', () => {
  // This is only run if `thing` is not in the cache.
  return 7
})
```

This PR is not doing that. The issue also never makes any mention of `undefined`.

lib/core/base/cache.js
straker: @stephenmathieson There are a few functions in axe-core where `undefined` is a valid cache value, such as [`is-modal-open](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/dom/is-modal-open.js#L100-L101). The point of the cache is to denote an operation (potentially expensive) that we do not want to do again. What the operation returns is not important.

dbowling:
Yes, that and a CI/local test discrepancy is why I took this back to draft status. If we want to change the intended behavior, thats okay too. Cache is a potential nightmare for bugs to hide, so I am all about being intentional. 

The undefined discussion started out of a review of finding appropriate test cases for the change. The code isn't currently explicitly defining how undefined should be treated and it is up to the caller. It appears there are mixed opinions on that, so it's good we are discussing as we evaluate how we want to use the cache.

lib/core/base/cache.js
stephenmathieson: What is the difference between not setting anything in the cache and setting `undefined`? I don't understand why we'd ever need this.

straker: Setting anything in the cache means we've performed an operation and do not wish to do it again. If it helps, think of it as memoizing a function, but instead of caching the value for a set of parameters, we can cache the value for the entire run. It doesn't matter if the memoized function returned `undefined` (or anything really), the function will not be run again for the same set of parameters.

lib/checks/navigation/heading-order-evaluate.js
```diff
@@ -3,6 +3,39 @@ import { querySelectorAllFilter, getAncestry } from '../../core/utils';
...
+  // find all headings, even ones that are outside the current
+  // context. also need to know where iframes are so we can insert
+  // the results of any in-context iframes into their proper order
+  // @see https://github.com/dequelabs/axe-core/issues/728
+
+  const headingOrder = cache.get('headingOrder', getHeadingOrder());
+
+  this.data({ headingOrder });
+
+  return true;
```
straker, line 18: I would say this function doesn't benefit from the cache get default value pattern. The main reason is that the setting of `data` should happen inside `getHeadingOrder` and not in the evaluate function (we only want the data on the first node, not every node), which means you need to pass `this` to `getHeadingOrder` or use an arrow function instead of a named function. That complicates things and only adds further nesting, so I think we should revert and leave this case as is.

lib/checks/navigation/region-evaluate.js
```diff
@@ -81,38 +83,31 @@ function findRegionlessElms(virtualNode, options) {
...
-      return vNode;
-    })
-    // Remove duplicate containers
-    .filter((vNode, index, array) => {
-      return array.indexOf(vNode) === index;
-    });
-  cache.set('regionlessNodes', regionlessNodes);
+  const regionlessNodes = cache.get(
+    'regionlessNodes',
+    findRegionlessElms(axe._tree[0], options)
```
straker, line 92: I think I would prefer this as it's own function for how complex it is.

```js
const regionlessNodes = cache.get(
  'regionlessNodes',
  () => getRegionlessNodes(options)
);

function getRegionlessNodes(options) {
  regionlessNodes = findRegionlessElms(axe._tree[0], options)
    // Find first parent marked as having region descendant (or body) and
    // return the node right before it as the "outer" element
    .map(vNode => {
      while (
        vNode.parent &&
        !vNode.parent._hasRegionDescendant &&
        vNode.parent.actualNode !== document.body
      ) {
        vNode = vNode.parent;
      }

      return vNode;
    })
    // Remove duplicate containers
    .filter((vNode, index, array) => {
      return array.indexOf(vNode) === index;
    });
}
```

lib/commons/dom/is-modal-open.js
```diff
@@ -34,26 +37,30 @@ function isModalOpen(options) {
...
+function definiteModalsExists(roles) {
+  return Boolean(
+    querySelectorAllFilter(
+      // TODO: es-module-_tree
+      axe._tree[0],
+      roles,
+      vNode => isVisible(vNode.actualNode)
+    ).length
+  );
+}
```
straker, line 56: This is an antipattern to use the `Boolean` constructor. Instead return a comparison.

```suggestion
  return querySelectorAllFilter(
      // TODO: es-module-_tree
      axe._tree[0],
      roles,
      vNode => isVisible(vNode.actualNode)
    ).length > 0;
}
```

lib/commons/dom/is-modal-open.js
```diff
@@ -34,26 +37,30 @@ function isModalOpen(options) {
...
-
-  const definiteModals = querySelectorAllFilter(
-    // TODO: es-module-_tree
-    axe._tree[0],
-    'dialog, [role=dialog], [aria-modal=true]',
-    vNode => isVisible(vNode.actualNode)
+  return cache.get(
+    'isModalOpen',
+    definiteModalsExists(definiteModalRoles) ||
+      probableModalExists(modalPercent)
```
straker, line 43: I feel that this change could lead to causing a subtle bug. If the order of the OR conditions were changed, the code would not work as expected since `probableModalExists` would return `undefined` and then still execute `definiteModalsExists`

lib/commons/dom/is-skip-link.js
```diff
@@ -54,3 +35,22 @@ export default function isSkipLink(element) {
...
+  } else {
+    firstPageLink = querySelectorAll(
+      axe._tree,
+      'a[href]:not([href^="javascript:"])'
+    ).find(link => !isCurrentPageLink(link.actualNode));
+  }
+
+  // null will signify no first page link
+  cache.set('firstPageLink', firstPageLink || null);
+  return firstPageLink;
```
straker, line 55:
```suggestion
  return firstPageLink || null;
```

lib/core/base/cache.js
```diff
@@ -13,10 +13,21 @@ const cache = {
   /**
    * Retrieve an item from the cache.
    * @param {String} key - Name of the key the value was stored as.
+   * @param {*} defaultValue - Default value to set if there is a cache miss. Functions are evaluated before caching. To override a value already saved, use `set()`.
```
straker, line 16:
```suggestion
   * @param {*} [defaultValue] - Default value to set if there is a cache miss. Functions are evaluated before caching. To override a value already saved, use `set()`.
```

lib/checks/navigation/heading-order-evaluate.js
```diff
@@ -3,6 +3,39 @@ import { querySelectorAllFilter, getAncestry } from '../../core/utils';
 import { isVisible } from '../../commons/dom';
 import { getRole } from '../../commons/aria';
 
+export default headingOrderEvaluate;
+
+function headingOrderEvaluate() {
```
straker, line 8:
```suggestion
export default function headingOrderEvaluate() {
```

lib/core/utils/filter-html-attrs.js
```diff
@@ -73,3 +62,18 @@ function filterHtmlAttrs(element, filterAttrs) {
 }
 
 export default filterHtmlAttrs;
```
straker, line 64: The default export function itself should be moved to the top with an inlined `default export`.

test/core/base/cache.js
```diff
@@ -1,14 +1,18 @@
-describe('axe._cache', function() {
+describe('axe._cache', function () {
   'use strict';
 
-  it('should set items from the cache without error', function() {
+  beforeEach(function () {
+    axe._cache.clear();
+  });
+
```
straker, line 7: This is handled by a [global `afterEach` in testutils](https://github.com/dequelabs/axe-core/blob/develop/test/testutils.js#L565)

```suggestion
```

straker (changes requested):

lib/core/base/cache.js
stephenmathieson: Perhaps I don't understand the code, but it looks like the function will run again if `undefined` is cached based on these lines: https://github.com/dequelabs/axe-core/blob/develop/lib/commons/dom/is-modal-open.js#L37-L39

straker: Thanks for pointing that out, looks like a bug in the `is-modal-open` function.

stephenmathieson: I still don't believe there's a reason to cache `undefined`.

lib/commons/dom/is-skip-link.js
```diff
@@ -54,3 +35,21 @@ export default function isSkipLink(element) {
     element.DOCUMENT_POSITION_FOLLOWING
   );
 }
+
+function generateFirstPageLink(firstPageLink) {
```
WilcoFiers, line 39:
```suggestion
function generateFirstPageLink() {
    let firstPageLink;
```

lib/core/base/cache.js
```diff
@@ -13,10 +13,21 @@ const cache = {
...
+  get(key, defaultValue) {
+    if (key in _cache) {
+      return _cache[key];
+    }
+
+    if (arguments.length === 2) {
+      this.set(
+        key,
+        typeof defaultValue === 'function' ? defaultValue() : defaultValue
+      );
```
WilcoFiers, line 28: I think we may need to throw if `creator` returns undefined. Because if it does we're just going to end up rerunning the creator method until it doesn't. That kind of defeats the purpose of the cache. Seeing how easy it is to get this wrong, I think it'd be safer not to allow undefined.

I spotted we had a bug because we're allowing this in isModalOpen (see my comments there).

lib/commons/dom/is-modal-open.js
```diff
@@ -34,26 +40,32 @@ function isModalOpen(options) {
...
   // just our best guess of being one based on convention.
 
-  if (cache.get('isModalOpen')) {
-    return cache.get('isModalOpen');
+  if (definitelyModalExists()) {
+    return true;
+  } else if (couldBeModalExists(modalPercent)) {
+    return true;
   }
+  return undefined;
```
WilcoFiers, line 48: If we pass undefined into the cache, it's not going to know this has been set. `detectModal` is going to run over and over again.

```suggestion
  return false;
```

WilcoFiers (changes requested):

lib/core/base/cache.js
```diff
@@ -13,10 +13,21 @@ const cache = {
   /**
    * Retrieve an item from the cache.
    * @param {String} key - Name of the key the value was stored as.
+   * @param {*} [defaultValue] - Default value to set if there is a cache miss. Functions are evaluated before caching. To override a value already saved, use `set()`.
    * @returns {*} The item stored
    */
-  get(key) {
-    return _cache[key];
+  get(key, defaultValue) {
```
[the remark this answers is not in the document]
dbowling: > I'm coming back to the idea of allowing a default that isn't a function. Just in this PR, there are a dozen places where it wasn't used correctly. I can't think of a single use case where this shouldn't be a function. I think we're better off enforcing a function. If we have any case where this shouldn't be a function, we can always use `.set()`.

I'm concerned that allowing `set()` to bypass the validation rules might led to unintentional bugs. I'm going to take a little inspiration from the React team and try to deliver your goal, but with protections by default and including an escape hatch with something like `dangerouslySet()`.

test/core/base/cache.js
```diff
@@ -1,32 +1,160 @@
...
+          axe._cache.get('boolean', function () {
+            return true;
+          })
+        );
+        assert.isNull(
+          axe._cache.get('null', function () {
+            return null;
+          })
+        );
+        assert.deepEqual(
```
WilcoFiers, line 85: We should check equality, not just deepEqual. We want it to be the same object that's returned.

lib/core/base/cache.js
```diff
@@ -7,16 +7,27 @@ const cache = {
...
    * @returns {*} The item stored
    */
-  get(key) {
-    return _cache[key];
+  get(key, creator) {
+    if (key in _cache) {
+      return _cache[key];
+    }
+
+    if (creator) {
```
WilcoFiers, line 26:
```suggestion
    if (typeof creator !== 'undefined') {
```

test/core/base/cache.js
```diff
@@ -1,32 +1,160 @@
...
+        assert.isBoolean(
+          axe._cache.get('Foo', function () {
+            return 'should not be set to a string because Foo is in the cache';
+          })
+        );
+      });
+
+      it('should not override a value from `set()`', function () {
+        axe._cache.set('foo', 'bar');
+        axe._cache.get('foo', 'baz');
```
WilcoFiers, line 138: It's weird for this not to throw. I think we should always check the creator type, regardless of if there's a value already or not. It's odd for this function to accept an invalid argument if a value was previously set.

lib/commons/dom/is-modal-open.js
```diff
@@ -14,6 +16,14 @@ function isModalOpen(options) {
   options = options || {};
   const modalPercent = options.modalPercent || 0.75;
 
+  try {
```
WilcoFiers, line 19: What's this for?

lib/commons/dom/is-skip-link.js
```diff
@@ -54,3 +35,22 @@ export default function isSkipLink(element) {
     element.DOCUMENT_POSITION_FOLLOWING
   );
 }
+
+function generateFirstPageLink() {
+    let firstPageLink;
```
WilcoFiers, line 40:
```suggestion
  let firstPageLink;
```

WilcoFiers (dismissed):

lib/commons/dom/is-modal-open.js
dbowling: The creator `detectModal` returns undefined on line `52`, which causes `cache.get` to throw. That looks to be entirely normal behavior with this, so the error is being caught.

I can toss in a comment, check for that specific error type, or both. It doesn't look like we would want to change the return behavior.

lib/commons/dom/is-skip-link.js
dbowling: Strange, I guess my prettier didn't work on this commit.

test/core/base/cache.js
dbowling: I'll update to do that check, though we won't be able to do 100% of the validation every time because that would require executing the function to determine the return type (since `undefined` is not allowed.)

dbowling: I was mistaken about your intent on undefined so ignore that comment. This is now resolved.

lib/commons/dom/is-modal-open.js
dbowling: As above, I misread the date that your comment on undefined was. This try/catch comes out when considering that so I am resolving.

test/core/base/cache.js
```diff
@@ -1,32 +1,173 @@
...
+      }
+      assert.doesNotThrow(fn);
+    });
+
+    // undefined is a valid result type of a potentially expensive operation in axe-core
+    it('should set `undefined` without error', function () {
+      axe._cache.set('foo', undefined);
+
+      function fn() {
+        assert.isUndefined(axe._cache.get('foo'));
```
straker, line 17: You're testing the `get` function in the `set` describe block. We should make sure that setting `undefined` does not throw, not that `get` does not throw

```suggestion
      function fn() {
        axe._cache.set('foo', undefined);
```

test/core/base/cache.js
```diff
@@ -1,32 +1,173 @@
...
+            return obj;
+          }),
+          obj
+        );
+
+        assert.sameOrderedMembers(
+          axe._cache.get('array', function () {
+            return [1, 2, 3];
+          }),
+          [1, 2, 3]
```
straker, line 100: You'll want to check that it returns the same array like you did for the obj

```suggestion
        var array = [1, 2, 3];
        assert.equal(
          axe._cache.get('array', function () {
            return array;
          }),
          array
```

straker (dismissed):

test/core/base/cache.js
```diff
@@ -1,32 +1,180 @@
...
-  });
+  describe('set', function () {
+    it('should set items without error', function () {
+      function fn() {
+        axe._cache.set('foo', 'bar');
+      }
+      assert.doesNotThrow(fn);
+    });
+
+    // undefined is a valid result type of a potentially expensive operation in axe-core
```
stephenmathieson, line 12: You all have convinced me that you believe this is true. You have yet to provide a valid use case for caching nothing though. Why is this necessary?

stephenmathieson (changes requested):

test/core/base/cache.js
straker: Both @WilcoFiers and I have agreed that we want this functionality

stephenmathieson: Why? I don't understand why an explanation is too much to ask.

lib/core/base/cache.js
```diff
@@ -7,16 +7,29 @@ const cache = {
...
-    return _cache[key];
+  get(key, creator) {
+    validateCreator(creator);
+
+    if (key in _cache) {
+      return _cache[key];
+    }
+
+    if (typeof creator === 'function') {
+      const value = creator();
```
WilcoFiers, line 29: Did we not say we were going to do this?

```suggestion
      const value = creator();
      assert(value !== 'undefined', 'Cache creator function should not return undefined');
```

lib/core/base/cache.js
```diff
@@ -27,4 +40,35 @@ const cache = {
...
+ * @returns {Void} Void if the key is valid.
+ * @throws {TypeError} If the key is not a string.
+ */
+function validateKey(key) {
+  if (typeof key !== 'string') {
+    throw new TypeError('key must be a string, ' + typeof key + ' given');
+  }
+  if (key === '') {
+    throw new TypeError('key must not be empty');
+  }
```
WilcoFiers, line 55: Did you know axe-core has an `assert`? Might be good to use:

```suggestion
  assert(typeof key !== 'string', 'key must be a string, ' + typeof key + ' given');
  assert(key === '', 'key must not be empty');
```

WilcoFiers (dismissed):

lib/core/base/cache.js
stephenmathieson: Without the extra quotes:

```suggestion
      const value = creator();
      assert(value !== undefined, 'Cache creator function should not return undefined')
```

test/core/base/cache.js
```diff
@@ -9,35 +9,34 @@ describe('axe._cache', function () {
...
-    // undefined is a valid result type of a potentially expensive operation in axe-core
     it('should set `undefined` without error', function () {
       function fn() {
         axe._cache.set('foo', undefined);
       }
       assert.doesNotThrow(fn);
     });
 
-    it('should throw TypeError for non-string keys', function () {
+    it('should throw for non-string keys', function () {
```
stephenmathieson, line 19: Probably not a big deal, but IMO this should still be a `TypeError` (the error is due to the wrong type being used).

dbowling: Agreed, I would rather throw a more specific error. I changed this because I was asked to use the internal assert, which only throws errors.

@straker and @WilcoFiers  How would you feel about me adding an error type param to the axe assert helper so we can continue to assert the more specific type?

dbowling:
@straker or @WilcoFiers  - how would you like me to handle the current failed tests for places where the [creator is returning undefined and throwing an error](https://circleci.com/gh/dequelabs/axe-core/48476?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)?

If we want to continue to throw, I'll need to touch these functions and check before sending, put in try/catch, or add an options object that permits cache to quietly ignore (e.g. not setting a value and not throwing) when it encounters this case.

Happy to do whatever you want, I'm mindful of scope bloat on this PR though. If that scope increase concerns you, we can still remove the throw (or make it a non-default setting) as that is new behavior for this PR.

lib/core/base/cache.js
stephenmathieson: Asked on Slack (but didn't get a response), so I'll do it again here...

If we are not allowing:

```js
cache.get('foo', function(){ return undefined })
```

Then why do we allow the below?

```js
cache.set('foo', undefined)
```

test/core/base/cache.js
dbowling: I spoke with Wilco and Steve earlier today and discussed adding an error type to the internal assert helper to improve test quality. The rest of axe-core doesn't differentiate between error types, so they would rather not change the pattern for just the cache. I prefer the more specific error types, but it is probably a solution in search of a problem here.

dbowling:
The [reply to my last comment can be found here](https://github.com/dequelabs/axe-core/pull/3608#discussion_r966448810).

Since the `assert` is throwing for `is-modal-open`, we also decided that a better solution than treating that as a bug is to memoize.

lib/core/base/cache.js
dbowling: The assert helper has been integrated. Stephen, I think Steven got to chat with you on Slack last week. Thanks for the feedback, but it looks like we're going anther direction on caching undefined so I'm going to resolve this conversation to get the PR ready for re-review with current scope.

stephenmathieson: The question above was never answered (neither on Slack nor here).

test/core/base/cache.js
dbowling: My understanding is that the two of you spoke offline on this question. I'm resolving this thread so we can finalize review.

stephenmathieson: This has still not been resolved. We did speak on Slack, but the question posted at https://github.com/dequelabs/axe-core/pull/3608#discussion_r965041066 has never been answered.

lib/commons/dom/is-modal-open.js
```diff
@@ -10,7 +10,7 @@ import { querySelectorAllFilter } from '../../core/utils';
  * @instance
  * @return {Boolean|undefined} True if we know (or our best guess) that a modal is open, undefined if we can't tell (doesn't mean there isn't one open)
  */
-function isModalOpen(options) {
+export default memoize(function isModalOpen(options) {
```
WilcoFiers, line 13: The options object is going to prevent memoize from working. This PR isn't about memoization anyway, so I don't think this belongs here. If we can't use the creator method for isModalOpen, best to just leave it as is.

lib/core/utils/filter-html-attrs.js
```diff
@@ -2,25 +2,6 @@ import cache from '../base/cache';
 import getNodeAttributes from './get-node-attributes';
 import matchesSelector from './element-matches';
 
-/**
```
WilcoFiers, line 5: Why was this file changed?

WilcoFiers (changes requested):

lib/core/utils/filter-html-attrs.js
dbowling: I was confused at first too, but there's nothing special about the changes to this file. It refactors the same thing the rest of them do, removing this pattern:

```JavaScript
if (cache.get(outerHTML)) {
    node = cache.get(outerHTML);
  }
```

It looks strange because the first function wasn't the default export, so I had switched their order to comply with the new code standards.

test/core/base/cache.js
```diff
@@ -1,32 +1,169 @@
...
+          axe._cache.get('foo', function () {
+            return undefined;
+          });
+        }
+        assert.throws(fn);
+      });
+
+      it('should not evaluate creator if key already exists', function () {
+        var spy = sinon.spy();
+        axe._cache.set('foo', spy);
```
WilcoFiers, line 120: Passing the spy in here is confusing.

```suggestion
        axe._cache.set('foo', 'bar');
```

test/core/base/cache.js
```diff
@@ -1,32 +1,169 @@
...
+        axe._cache.get('foo', spy);
+        assert.isFalse(spy.calledOnce);
```
WilcoFiers, line 122: This makes it look as though this could be called more than once. It shouldn't have been called at all.

WilcoFiers (changes requested):
Couple minor points

test/core/base/cache.js
dbowling: Makes sense, thanks for the feedback.

test/core/base/cache.js
dbowling: Good point, this has been clarified.

WilcoFiers (approved):

Issue #3500, opened 2022-06-03
add option to isFocusable to take into account negative tabindex


straker:
The [isFocusable function](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/dom/is-focusable.js) currently does not take into account negative tabindex (and for the default use case it shouldn't take into account negative tabindex). This causes us to pair the function call with a check on tabindex to ensure that the element is currently keyboard focusable and not programmatically focusable.

```js
const tabindex = vNode.attr('tabindex');
if ((isNaN(tabindex) || tabindex >  -1) && isFocusable(element)) {
  // ...
}
```

We should add an option to the function that enables looking for negative tabindex and changes the return to `false` for the element. Something like `programmatically: false` or `noNegativeTabindex: false`.

```js
if (isFocusable(element, { programmatically: false }) {
  // ...
}
```

padmavemulapati:
Validated using latest develop branch code base, 
If negative tab index is there within the frame it is failing but if it is in `focussible` area, then it is ignoring
Script:
```<iframe title="frame"  srcdoc="<a href='/' tabindex='-1'>Home</a>"></iframe>
<script src="/axe.js"></script>
```
![image](https://user-images.githubusercontent.com/47409579/188117741-e8ed0c0b-7814-4d34-a85f-f7d0973f2cb6.png)

PR #3609, opened 2022-08-16
refactor(is-focusable): account for negative tabindex

dbowling:
- add option arg that enables looking for negative tabindex (defaults to false)

references: #3500

### DIFF
--- a/.github/ISSUE_TEMPLATE/possible-bug.yml
+++ b/.github/ISSUE_TEMPLATE/possible-bug.yml
@@ -27,7 +27,7 @@ body:
   - type: checkboxes
     id: latest-axe
     attributes:
-      label: Lastest Version
+      label: Latest Version
       options:
         - label: I have tested the issue with the latest version of the product
           required: true


lib/commons/dom/is-focusable.js
```diff
@@ -9,22 +9,33 @@ import { getNodeFromTree } from '../../core/utils';
  * @memberof axe.commons.dom
  * @instance
  * @param {HTMLElement} el The HTMLElement
+ * @param {object} options Optional. If set to {programatically: false}, the function will return true
  * @return {Boolean} The element's focusability status
  */
-export default function isFocusable(el) {
+export default function isFocusable(el, options = { programmatically: false }) {
```
straker, line 15: No need to set the default

```suggestion
export default function isFocusable(el, options = {}) {
```

lib/commons/dom/is-focusable.js
```diff
@@ -9,22 +9,33 @@ import { getNodeFromTree } from '../../core/utils';
...
   if (vNode.props.nodeType !== 1) {
     return false;
   }
 
+  var tabindex = vNode.attr('tabindex');
+
+  if (
+    tabindex &&
+    options.programmatically === true &&
+    parseInt(tabindex, 10) === -1
```
straker, line 27: `-1` is not the only valid negative tabindex value

```suggestion
    parseInt(tabindex, 10) < 0
```

straker (dismissed):

lib/commons/dom/is-focusable.js
straker: You'll also want to modify one of the tests for another negative value other than `-1`

lib/commons/dom/is-focusable.js
```diff
@@ -9,22 +9,33 @@ import { getNodeFromTree } from '../../core/utils';
...
   }
 
+  var tabindex = vNode.attr('tabindex');
+
+  if (
+    tabindex &&
+    options.programmatically === true &&
+    parseInt(tabindex, 10) < 0
+  ) {
+    return false;
```
WilcoFiers, line 29: Elements with a negative tabindex are programmatically focucsable.

```suggestionI
    return true;
```

WilcoFiers (changes requested):
1. This is not a feature, it's a refactor
2. Please use the option, otherwise you've just added dead code.

test/commons/dom/is-focusable.js
```diff
@@ -184,89 +184,221 @@ describe('is-focusable', function() {
...
+    it('should return false for a div with no tabindex', function () {
       fixture.innerHTML = '<div id="target"></div>';
       var el = document.getElementById('target');
       flatTreeSetup(fixture);
 
       assert.isFalse(axe.commons.dom.isFocusable(el));
     });
+    describe('options', function () {
+      describe('programatically', function () {
+        describe('true (default)', function () {
```
WilcoFiers, line 196: This is kind of a personal preference, but when it comes to boolean options I prefer defaults to be falsey. It's not intuitive (to me) for undefined and true to behave the same, but for false to be different. Much of axe-core works that way, so it's more consistent doing this too.

I propose using `ifProgrammatic` with a falsey default.

WilcoFiers (dismissed):

dbowling:
Just documenting a discussion for this PR:

We're going to create a new function and leave isFocusable unchanged. This should drastically simplify future implementations. Also, implementations will go into a new issue as needed to avoid scope creep.

I'll force push the new approach, so I'm resolving all requests for changes now as outdated.

straker:
We seem to have somehow confused the GitHub UI. This pr was not merged into develop, does [not appear in develops commit history](https://github.com/dequelabs/axe-core/commits/develop), and GitHub even says 0 commits were merged. The commit that says was merged (49ccb00) is [an entirely different pr](https://github.com/dequelabs/axe-core/commit/49ccb007c63cc3f1dce7b82cb7630856fb17605c). Not sure what happened.

Screenshot from GitHub showing: 
dbowling merged 0 commits into develop from 3500-isFocusable-negative-tabindex

Issue #3487, opened 2022-06-01
Make frame-title-unique a reviewOnly WCAG412 rule


WilcoFiers:
Aze-core's `frame-title-unique` rule is currently inconsistent with ACT Rules because it is far too strict. A better approach to this rule would be to change the tags, give it the `wcag412` and `wcag2a` tags, and remove `best-practice`, then add `"reviewOnFail": true` to the rule so it never flags things as violations. We simply cannot tell these are WCAG violations. This makes the rule partially consistent with the [Iframe elements with identical accessible names have equivalent purpose](https://wai-wcag-act-rules.netlify.app/standards-guidelines/act/rules/4b1c6c/proposed/) ACT rule.

padmavemulapati:
Validated in the latest develop branch code base, 
`frame-title-unique` is showing in the `incomplete` results now and also having no `best-practice` tag value.
We are seeing only tags, 
```
0: "cat.text-alternatives"
1: "wcag412"
2: "wcag2a"
```
![image](https://user-images.githubusercontent.com/47409579/187217878-a94ef4ae-7260-4079-9f17-b663437e2bdc.png)

Earlier this rule is in list of `violations`, and also having only `best-practice` tag 

![image](https://user-images.githubusercontent.com/47409579/187218080-c573365d-9bb4-40a0-a1f9-703aa618b4c0.png)

PR #3610, opened 2022-08-16
fix(frame-title-unique): Make frame-title-unique reviewOnly and WCAG412

dbowling:
Aze-core's `frame-title-unique` rule is currently inconsistent with ACT Rules because it is far too strict. 

This makes the rule partially consistent with the [Iframe elements with identical accessible names have equivalent purpose](https://wai-wcag-act-rules.netlify.app/standards-guidelines/act/rules/4b1c6c/proposed/) ACT rule.

- add tags: `wcag412` and `wcag2a` 
- remove  tag: `best-practice`
- add `"reviewOnFail": true`

references: #3487


lib/rules/frame-title-unique.json
```diff
@@ -2,13 +2,14 @@
...
   "metadata": {
     "description": "Ensures <iframe> and <frame> elements contain a unique title attribute",
-    "help": "Frames should have a unique title attribute"
+    "help": "Frames must have a unique title attribute"
   },
   "all": [],
   "any": [],
-  "none": ["unique-frame-title"]
+  "none": ["unique-frame-title"],
+  "reviewOnFail": true
```
WilcoFiers, line 14: Don't know why, but this isn't working, otherwise the integration tests would've failed. This is not something that can be changed without updating integration tests.

WilcoFiers (dismissed):

test/core/base/rule.js
```diff
@@ -1662,144 +1662,209 @@ describe('Rule', function() {
         assert.lengthOf(result.nodes, 1);
       });
     });
+
+    it('should mark checks as incomplete if reviewOnFail is set to true', function () {
```
WilcoFiers, line 1666: I think you should test each type separately. This test can pass if two of the three types are broken.

WilcoFiers (changes requested):
Good catch on the Rule class change. This turned out more complicated than I expected it would be.

padmavemulapati:
Validated in the latest develop branch code base, 
`frame-title-unique` is showing in the `incomplete` results now and also having no `best-practice` tag value.
We are seeing only tags, 
```
0: "cat.text-alternatives"
1: "wcag412"
2: "wcag2a"
```
image

Earlier this rule is in list of `violations`, and also having only `best-practice` tag 

image

Issue #2652, opened 2020-11-19
Add target-spacing rule


WilcoFiers:
I wrote a load of test cases on how to do this:

https://codepen.io/wilcofiers/pen/abZxPow

I'd like to implement this rule while WCAG 2.2 is still in draft, so that it can be used for testing. We can take it out of "experimental" as soon as WCAG 2.2 becomes a recommendation.

padmavemulapati:
@straker , can you please give QA notes to validate this ticket

padmavemulapati:
Validated with the latest axe-core develop branch code base, 
test script for the failure in `target-size` rule:
```
<!-- Failed examples -->
  <p>
    <button id="failCase1">x</button
    ><span role="button" tabindex="0" id="failCase2">x</span>
  </p>
  
  <ul>
    <li><a href="#" id="failCase3">Item1</a></li>
    <li><a href="#" id="failCase4">Item2</a></li>
  </ul>
  
  <p>
    <input type="checkbox" id="failCase5" /><input type="checkbox" id="failCase6" /><input
      type="text"
      id="pass-adjacent"
    />
  </p>
  
```


![Image](https://user-images.githubusercontent.com/47409579/192687246-11e9feae-e53a-4625-aee1-1718b00fe070.png)

Issue #3281, opened 2021-11-12
Odd errors when importing aria/get-role


WilcoFiers:
While working on #3275 I found that when I imported `aria/get-role` into `dom/is-skip-link`, axe-core for some reason completely broke. Just importing would do it. Did not require invoking getRole. I couldn't figure out the problem, and so moved passed it. I don't see how something like that could cause problems. There's clearly something going on in our build system that we don't understand. I would like to know what that is, so we don't run into this at a time where we don't have a simple work-around.

straker:
I've determined the problem occurs with this import chain:

get-role -> implicit-role -> implicit-html-roles -> is-column-header / is-row-header -> get-scope -> to-grid

Memoize by that point should be defined up higher in the code since it's in utils, but for some reason is being put below the code that now needs it

PR #3616, opened 2022-08-19
feat(new-rule): Add WCAG 2.2 target-size rule

WilcoFiers:
- [x] Better handle rounding errors, so that the minimum is not 23.95, but actually 24px
- [x] Tests for target-offset
- [x] Report the offset to the closest neighbour, and indicate the expected offset
- [x] Update the selector to ensure all focusable widgets are included -- may need a matcher too
- [x] Have a matcher filter out inline controls
- [x] Add more integration tests that deal with the selector / matcher
- [x] Document the new check options

Closes issue: #2652


test/commons/dom/visually-sort.js
```diff
@@ -0,0 +1,28 @@
+// This method is mostly tested through color-contrast integrations
```
WilcoFiers, line 1: @straker I could use some help figuring out what needs to be tested here.

lib/rules/widget-not-inline-matches.js
```diff
@@ -0,0 +1,30 @@
...
+  (node, vNode) => !svgNamespaceMatches(node, vNode),
+  (node, vNode) => isFocusable(vNode),
+  node => !isInTextBlock(node, { noLengthCompare: true })
+];
+
+function isWidgetType(vNode) {
+  const role = getRole(vNode);
+  if ('option' === role) {
+    return false;
+  } else if (['combobox', 'listbox'].includes(role)) {
```
WilcoFiers, line 21: For consideration; I could potentially change the role type of these three so that everything we want is a widget, and everything we don't isn't. I didn't explore what the wider implications of that could be.

lib/commons/dom/is-in-text-block.js
```diff
@@ -84,24 +92,27 @@ function isInTextBlock(node) {
...
+  if (options?.noLengthCompare) {
+    return parentText.length !== 0;
+  }
+
+  widgetText = sanitize(widgetText);
+  return parentText.length > widgetText.length;
+}
 
-  return parentText.length > linkText.length;
+function isWidget(domNode) {
```
straker, line 112: This is the second instance of this function (the other being in `widget-not-inline-matches.js`, but they are slightly different. Thwidget-not-inline-matches.js` only accounts for `listbox` and `combobox` composite roles, but this one includes all composite roles. Should this function be pulled out and made a shared function that both use?

test/commons/dom/find-nearby-elms.js
```diff
@@ -0,0 +1,41 @@
...
+      if (vNode.props.id && vNode.props.id !== 'fixture') {
+        ids.push(vNode.props.id);
+      }
+    });
+    return ids;
+  }
+
+  beforeEach(function () {
+    fixture = fixtureSetup(
+      '<div id="n0" style="height:30px; margin-bottom:30px;">0</div>' +
```
straker, line 19: This test fails when debugging in a browser due to a Mocha element. 

Test result for 'returns node from the same grid cell' showing failure due to 'mocha-stats' node being included in the results

Should probably set the width of these nodes to something small (like 30px) so it doesn't try to include extraneous elements outside of fixture.

straker (changes requested):
I'm not done reviewing this, but figured I get this out there now so you could work on the things I did find.

straker (changes requested):
More things (still working my way through)

lib/checks/mobile/target-size.json
```diff
@@ -0,0 +1,17 @@
+{
+  "id": "target-size",
+  "evaluate": "target-size-evaluate",
+  "options": {
+    "minSpacing": 24
```
straker, line 5: `minSpacing` doesn't make much sense for target size. This should be `minSize` or `targetSize`

test/checks/mobile/target-size.js
```diff
@@ -0,0 +1,106 @@
...
+    );
+    assert.isTrue(check.evaluate.apply(checkContext, checkArgs));
+    assert.deepEqual(checkContext._data, {
+      minSpacing: 24,
+      width: 40,
+      height: 30
+    });
+  });
+
+  it('returns true for obscured targets with insufficient space', function () {
```
straker, line 47: Should this be `sufficient space`? The test below this one has the same title but a different return.

straker (dismissed):
Finished

lib/commons/math/has-visual-overlap.js
```diff
@@ -0,0 +1,44 @@
+import { visuallySort } from '../dom';
+
+/**
+ * Check if node A overlaps B
+ * @param {VirtualNode} vNodeA
+ * @param {VirtualNode} vNodeB
+ * @returns 0 | 1 | -1
+ */
+export default function hasVisualOverlap(vNodeA, vNodeB) {
+  const overlapRect = getRectOverlap(
```
straker, line 10: This could be a much simpler check (and easier to understand) if you used an [axis-aligned bounding box collision check](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/dom/get-rect-stack.js#L129-L137). Then you just check for collision and call `visualSort` if they collide.

```js
return vNodeA.left < vNodeB.left + vNodeB.width &&
    vNodeA.left + vNodeA.width > vNodeB.left &&
    vNodeA.top < vNodeB.top + vNodeB.height &&
    vNodeA.top + vNodeA.height > vNodeB.top &&
    visuallySort(vNodeA, vNodeB) > 0;
```

lib/checks/mobile/target-offset-evaluate.js
```diff
@@ -0,0 +1,34 @@
...
+
+const roundingMargin = 0.05;
+
+export default function targetOffsetEvaluate(node, options, vNode) {
+  const minOffset = options?.minOffset || 24;
+  const closeNeighbors = [];
+  let closestOffset = minOffset;
+  for (const vNeighbor of findNearbyElms(vNode, minOffset)) {
+    const role = getRole(vNeighbor);
+    if (getRoleType(role) !== 'widget') {
```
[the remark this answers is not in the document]
WilcoFiers: Good point, although not exactly. Disabled buttons should probably be excluded... that's a fair enough point, but I don't think we should exclude inline links for example, or area / SVG widgets. If you put a button too close to your SVG, the button's still the problem, even if axe passes on the SVG.

lib/checks/mobile/target-offset.json
```diff
@@ -0,0 +1,14 @@
+{
+  "id": "target-offset",
+  "evaluate": "target-offset-evaluate",
+  "options": {
+    "minOffset": 24
+  },
+  "metadata": {
+    "impact": "serious",
+    "messages": {
+      "pass": "Target is sufficiently offset from its closest neighbor (${data.closestOffset}px should be at least ${data.minOffset}px)",
```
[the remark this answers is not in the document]
WilcoFiers: I put it there because I figured if someone is actually reading these, they're probably trying to figure out why axe-core is passing something they thought should fail. We can do that without putting it in the pass message though. Let me know if you want this removed. It'll still be in `.data`, even if it's not in the message. I can go either way.

lib/checks/mobile/target-size.json
WilcoFiers: Agreed, should be `minSize`.

lib/commons/dom/is-in-text-block.js
WilcoFiers: Noticed that did you?... You're right. Need to fix that. I'll go find a better solution. I don't quite like how wonky this is with the listbox / combobox exception. I might just change what type we give to them and avoid needing a function at all.

lib/rules/widget-not-inline-matches.js
```diff
@@ -0,0 +1,30 @@
...
+
+export default function widgetNotInline(node, vNode) {
+  return matchesFns.every(fn => fn(node, vNode));
+}
+
+const matchesFns = [
+  (node, vNode) => isWidgetType(vNode),
+  (node, vNode) => isNotAreaElement(vNode),
+  (node, vNode) => !svgNamespaceMatches(node, vNode),
+  (node, vNode) => isFocusable(vNode),
```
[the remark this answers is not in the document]
WilcoFiers: Tricky. An element with opacity: 0 is still clickable. All those other on screen visibility methods do prevent focus, but this one doesn't... Also, maybe I need to think about  CSS `pointer-events`, which makes a target non-clickable....

lib/commons/dom/create-grid.js
```diff
@@ -324,11 +324,13 @@ function findScrollRegionParent(vNode, parentVNode) {
...
-  // can find it even if it's in a subgrid
-  vNode._grid = grid;
-
   vNode.clientRects.forEach(rect => {
+    if (rect.right <= 0 || rect.bottom <= 0) {
+      return;
+    }
+    // save a reference to where this element is in the grid so we
+    // can find it even if it's in a subgrid
+    vNode._grid ??= grid;
```
WilcoFiers, line 333: IMPORTANT: These lines have been modified to fix a bug where elements outside the viewport could still be added to the grid.

lib/commons/math/get-offset.js
```diff
@@ -0,0 +1,147 @@
...
+    }
+    // Work out which edge of A is farthest away from the center of B
+    const centerB = rectB[start] + rectB[diameter] / 2;
+    const startDistance = Math.abs(centerB - rectA[start]);
+    const endDistance = Math.abs(centerB - rectA[end]);
+    if (startDistance >= endDistance) {
+      farthestPoint[axis] = rectA[start]; // left | top
+    } else if (startDistance < endDistance) {
+      farthestPoint[axis] = rectA[end]; // right | bottom
+    }
```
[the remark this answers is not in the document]
WilcoFiers: Why would that be an improvement? Xor'ing a bit on the int is going to be faster than squaring it, no? My version of this (minus the elseif, I agree) better expresses what's happening. Even if yours is faster, which I doubt, it's confusing.

lib/commons/math/get-offset.js
```diff
@@ -0,0 +1,147 @@
...
+    adjacentRect = ownRect;
+  }
+
+  const { top, right, bottom, left } = adjacentRect;
+  // Is the adjacent rect horizontally or vertically aligned
+  const xAligned = x >= left && x <= right;
+  const yAligned = y >= top && y <= bottom;
+  // Find the closest edge of the adjacent rect
+  const closestX = Math.abs(left - x) < Math.abs(right - x) ? left : right;
+  const closestY = Math.abs(top - y) < Math.abs(bottom - y) ? top : bottom;
```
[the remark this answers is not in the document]
WilcoFiers: As above. I doubt this is faster, but it's definitely not expressing what's going on as well.

lib/commons/math/get-offset.js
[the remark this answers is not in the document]
WilcoFiers: Going to leave these. Pow is slightly faster, but not enough that it's worth making the code less readable.

lib/checks/keyboard/no-focusable-content-evaluate.js
```diff
@@ -41,8 +41,7 @@ function getFocusableDescendants(vNode) {
 
   const retVal = [];
   vNode.children.forEach(child => {
-    const role = getRole(child);
-    if (getRoleType(role) === 'widget' && isFocusable(child)) {
+    if (getRoleType(child) === 'widget' && isFocusable(child)) {
```
WilcoFiers, line 44: If you want me to pull these into their own PR let me know. It's not unrelated to the PR... but it's not very related either.

lib/commons/math/has-visual-overlap.js
```diff
@@ -0,0 +1,44 @@
+import { visuallySort } from '../dom';
```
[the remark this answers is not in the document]
WilcoFiers: Nope... but this is a good bit confusing. `dom.visuallyOverlaps` looks that a rect intersects with a particular element. It doesn't check whether A is on top of B, or B is on top of A. Probably `dom.visuallyOverlaps` should have been called `rectIntersectsElement` or something like that? Don't think it impacts this PR.

lib/commons/math/get-offset.js
```diff
@@ -0,0 +1,147 @@
...
+  dimensionProps.forEach(([axis, start, end, diameter]) => {
+    if (rectB[start] < rectA[start] && rectB[end] > rectA[end]) {
+      farthestPoint[axis] = rectA[start] + rectA[diameter] / 2; // center | middle
+      return;
+    }
+    // Work out which edge of A is farthest away from the center of B
+    const centerB = rectB[start] + rectB[diameter] / 2;
+    const startDistance = Math.abs(centerB - rectA[start]);
+    const endDistance = Math.abs(centerB - rectA[end]);
+    if (startDistance >= endDistance) {
```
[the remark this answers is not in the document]
WilcoFiers: If the distance is the same it doesn't matter.

test/commons/dom/find-nearby-elms.js
WilcoFiers: How do I get to this? And should we care?

lib/rules/widget-not-inline-matches.js
```diff
@@ -15,14 +15,7 @@ const matchesFns = [
...
 function isWidgetType(vNode) {
-  const role = getRole(vNode);
-  if ('option' === role) {
-    return false;
-  } else if (['combobox', 'listbox'].includes(role)) {
-    return true;
-  }
-  const roleType = getRoleType(role);
-  return roleType === 'widget';
+  return getRoleType(vNode) === 'widget';
```
straker, line 18: So this will return `true` for the `option` role, is that OK?

lib/checks/mobile/target-size-evaluate.js
```diff
@@ -10,32 +10,44 @@ const roundingMargin = 0.05;
...
+  for (const vNeighbor of findNearbyElms(vNode)) {
+    if (
+      !hasVisualOverlap(vNode, vNeighbor) ||
+      getCssPointerEvents(vNeighbor) === 'none'
+    ) {
+      continue;
+    }
+    if (isEnclosedRect(vNode, vNeighbor)) {
+      this.data({ messageKey: 'obscured' });
+      return true;
```
straker, line 30: Should add the node as a related node

test/checks/mobile/target-size.js
```diff
@@ -39,45 +38,139 @@ describe('target-size tests', function () {
...
+          'display: inline-block; width:40px; height:30px; margin-left:30px;' +
+          '">x</button>' +
+          '<button style="' +
+          'display: inline-block; width:40px; height:30px; margin-left: -10px;' +
+          '">x</button>' +
+          '<span tabindex="0" style="' +
+          'display: inline-block; width:40px; height:30px; margin-left: -100px;' +
+          '">x</span>'
+      );
+      assert.isTrue(check.evaluate.apply(checkContext, checkArgs));
```
straker, line 99: I feel like this should be `false`. the `span` with `tabindex=0` captures the click on the left side of the target, making the clickable width effectively 20px.

image

test/checks/mobile/target-size.js
```diff
@@ -39,45 +38,139 @@ describe('target-size tests', function () {
...
+          'display: inline-block; width:40px; height:30px; margin-left:30px;' +
+          '">x</button>' +
+          '<button style="' +
+          'display: inline-block; width:40px; height:30px; margin-left: -10px;' +
+          '">x</button>' +
+          '<button disabled style="' +
+          'display: inline-block; width:40px; height:30px; margin-left: -100px;' +
+          '">x</button>'
+      );
+      assert.isTrue(check.evaluate.apply(checkContext, checkArgs));
```
straker, line 119: I feel this should be `false` as well since the disabled button captures the click on the left side of the target, making the clickable width effectively 20px. 

This can be debatable though as the disabled button doesn't cause a misclick to happen.

image

straker (changes requested):

lib/commons/aria/get-role-type.js
```diff
@@ -1,21 +1,23 @@
...
 function getRoleType(role) {
-  const roleDef = standards.ariaRoles[role];
-
-  if (!roleDef) {
-    return null;
+  if (
+    role instanceof AbstractVirtualNode ||
+    (window?.Node && role instanceof window.Node)
+  ) {
+    role = axe.commons.aria.getRole(role);
```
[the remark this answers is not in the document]
WilcoFiers: I didn't noticed that. Have I really? I used instanceof here because that's what we usually do when we want to check nodes. There are lots of `instanceof AbstractVirtualNode` calls.. and a couple `instanceof window.Node` calls too.

Looking at where/how this function has been used, it's going to get `null` passed into it a good bit too, since getRole() can return null. I'll update it to reflect that.

lib/commons/dom/is-in-text-block.js
```diff
@@ -33,10 +34,16 @@ function getBlockParent(node) {
...
+ * widgets. When false, compares the length of non-widget text to widget text
+ *
  * @param  {Element} node [description]
+ * @param  {Object} options Optional
+ * @property {Bool} noLengthCompare
  * @return {Boolean}      [description]
  */
-function isInTextBlock(node) {
+function isInTextBlock(node, options) {
+  const { getRoleType } = axe.commons.aria;
```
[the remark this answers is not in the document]
WilcoFiers: Ran into this error again: https://github.com/dequelabs/axe-core/issues/3281

Looks like I can work around it by importing directly from get-role-type, rather than from aria/index. Doing that.

lib/rules/widget-not-inline-matches.js
WilcoFiers: They do, but they're pulled out for not being focusable.

test/checks/mobile/target-size.js
WilcoFiers: This is no different than if that was just a narrower button. If someone accidentally hits the span instead of the button, nothing happens. Or at least, we can't assume something is going to happen. If that span is a widget that is missing a role, a different accessibility violation exists in that page, and after they've fixed that (by adding a role to the span) this rule will work correctly.

We could maybe report this as incomplete? I'd prefer we do this later though. This PR is much larger than I want it to be already. Please open an issue for this.

test/checks/mobile/target-size.js
WilcoFiers: Same answer. I don't think we can fail this, maybe incomplete, but I'd prefer we leave that for another time.

straker (approved):

straker (approved):

PR #3619, opened 2022-08-24
feat(is-in-tab-order): add isInTabOrder to commons

dbowling:
- add `axe.commons.isInTabOrder()`
- replace uses of `isFocusable` with `isInTabOrder` when `isFocusable` also checked for negative tabindex

references: https://github.com/dequelabs/axe-core/issues/3500

### DIFF
--- a/lib/checks/keyboard/no-focusable-content-evaluate.js
+++ b/lib/checks/keyboard/no-focusable-content-evaluate.js
@@ -1,5 +1,6 @@
 import isFocusable from '../../commons/dom/is-focusable';
 import { getRole, getRoleType } from '../../commons/aria';
+import { isInTabOrder } from '../../commons/dom';
 
 export default function noFocusableContentEvaluate(node, options, virtualNode) {
   if (!virtualNode.children) {
@@ -14,7 +15,7 @@ export default function noFocusableContentEvaluate(node, options, virtualNode) {
     }
 
     const notHiddenElements = focusableDescendants.filter(
-      usesUnreliableHidingStrategy
+      vNode => !isInTabOrder(vNode)
     );
 
     if (notHiddenElements.length > 0) {
@@ -49,9 +50,4 @@ function getFocusableDescendants(vNode) {
     }
   });
   return retVal;
-}
-
-function usesUnreliableHidingStrategy(vNode) {
-  const tabIndex = parseInt(vNode.attr('tabindex'), 10);
-  return !isNaN(tabIndex) && tabIndex < 0;
 }
--- a/lib/checks/keyboard/no-focusable-content-evaluate.js
+++ b/lib/checks/keyboard/no-focusable-content-evaluate.js
@@ -1,6 +1,5 @@
 import isFocusable from '../../commons/dom/is-focusable';
 import { getRole, getRoleType } from '../../commons/aria';
-import { isInTabOrder } from '../../commons/dom';
 
 export default function noFocusableContentEvaluate(node, options, virtualNode) {
   if (!virtualNode.children) {
@@ -15,7 +14,7 @@ export default function noFocusableContentEvaluate(node, options, virtualNode) {
     }
 
     const notHiddenElements = focusableDescendants.filter(
-      vNode => !isInTabOrder(vNode)
+      usesUnreliableHidingStrategy
     );
 
     if (notHiddenElements.length > 0) {
@@ -50,4 +49,9 @@ function getFocusableDescendants(vNode) {
     }
   });
   return retVal;
+}
+
+function usesUnreliableHidingStrategy(vNode) {
+  const tabIndex = parseInt(vNode.attr('tabindex'), 10);
+  return !isNaN(tabIndex) && tabIndex < 0;
 }
--- a/lib/commons/dom/get-tabbable-elements.js
+++ b/lib/commons/dom/get-tabbable-elements.js
@@ -1,4 +1,3 @@
-import { isInTabOrder } from '.';
 import { querySelectorAll } from '../../core/utils';
 
 /**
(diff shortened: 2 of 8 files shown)


lib/commons/dom/is-in-tab-order.js
```diff
@@ -0,0 +1,23 @@
...
+ * @return {Boolean} The element's tabindex status
+ */
+export default function isInTabOrder(el) {
+  const vNode = el instanceof AbstractVirtualNode ? el : getNodeFromTree(el);
+
+  if (vNode.props.nodeType !== 1) {
+    return false;
+  }
+
+  const tabindex = parseInt(vNode.attr('tabindex', 10));
```
WilcoFiers, line 20: Your code will return `true` for hidden elements with a non-negative tabindex. 

```suggestion
  const tabindex = parseInt(vNode.attr('tabindex', 10));
  if (tabindex <= -1) {
    return false; // Elements with negative tabindex are never in the tab order
  }
```

lib/commons/dom/is-in-tab-order.js
```diff
@@ -0,0 +1,23 @@
...
+
+  return isFocusable(vNode) && (isNaN(tabindex) || tabindex > -1);
```
WilcoFiers, line 22:
```suggestion
  return isFocusable(vNode);
```

test/commons/dom/is-in-tab-order.js
```diff
@@ -0,0 +1,90 @@
...
+  it('should return false for hidden inputs', function () {
+    var target = queryFixture('<input type="hidden" id="target"></input>');
+    assert.isFalse(isInTabOrder(target));
+  });
+
+  it('should return false for non-element nodes', function () {
+    var target = queryFixture('<span id="target">Hello World</span>');
+    assert.isFalse(isInTabOrder(target));
+  });
+});
```
WilcoFiers, line 90: I think one of these tests fail with the function written as is.

```suggestion

  it('should return false natively focusable hidden element', function () {
    var target = queryFixture('<button id="target" hidden></button>');
    assert.isTrue(isInTabOrder(target));
  });

  it('should return false hidden element with tabindex 1', function () {
    var target = queryFixture('<div id="target" tabindex="1" hidden></div>');
    assert.isTrue(isInTabOrder(target));
  });
});
```

WilcoFiers (changes requested):
Where appropriate, can you replace uses of `isFocusable` with `isInTabOrder`?

test/commons/dom/is-in-tab-order.js
```diff
@@ -0,0 +1,100 @@
...
+    );
+    assert.isFalse(isInTabOrder(target));
+  });
+
+  it('should return false for hidden inputs', function () {
+    var target = queryFixture('<input type="hidden" id="target"></input>');
+    assert.isFalse(isInTabOrder(target));
+  });
+
+  it('should return false for non-element nodes', function () {
```
straker, line 86: Did you mean to test the `#text` node rather than the `span`? If so you'll need to pass `target.children[0]`.

test/commons/dom/is-in-tab-order.js
```diff
@@ -0,0 +1,100 @@
...
+    var target = queryFixture('<span id="target">Hello World</span>');
+    assert.isFalse(isInTabOrder(target));
+  });
+
+  it('should return false natively focusable hidden element', function () {
```
straker, line 91:
```suggestion
  it('should return false for natively focusable hidden element', function () {
```

test/commons/dom/is-in-tab-order.js
```diff
@@ -0,0 +1,100 @@
...
+    var target = queryFixture('<button id="target" hidden></button>');
+    assert.isFalse(isInTabOrder(target));
+  });
+
+  it('should return false hidden element with tabindex 1', function () {
```
straker, line 96:
```suggestion
  it('should return for false hidden element with tabindex 1', function () {
```

straker (changes requested):

test/commons/dom/is-in-tab-order.js
dbowling: Yes, thanks for the catch! I will get that changed.

lib/checks/keyboard/no-focusable-content-evaluate.js
```diff
@@ -14,7 +15,7 @@ export default function noFocusableContentEvaluate(node, options, virtualNode) {
     }
 
     const notHiddenElements = focusableDescendants.filter(
-      usesUnreliableHidingStrategy
+      vNode => !isInTabOrder(vNode)
```
WilcoFiers, line 18: By calling into isFocusable again, this is going to be slower than it needs to be. Best to revert.

lib/commons/dom/get-tabbable-elements.js
```diff
@@ -12,12 +13,7 @@ function getTabbableElements(virtualNode) {
   const nodeAndDescendents = querySelectorAll(virtualNode, '*');
 
   const tabbableElements = nodeAndDescendents.filter(vNode => {
-    const isFocusable = vNode.isFocusable;
```
WilcoFiers, line 15: We should continue to use the cached version on vNode. Best to revert this one too.

lib/commons/dom/inserted-into-focus-order.js
```diff
@@ -12,13 +12,7 @@ import isNativelyFocusable from './is-natively-focusable';
...
  */
 function insertedIntoFocusOrder(el) {
-  const tabIndex = parseInt(el.getAttribute('tabindex'), 10);
-
-  // an element that has an invalid tabindex will return 0 or -1 based on
-  // if it is natively focusable or not, which will always be false for this
-  // check as NaN is not > 1
-  // @see https://www.w3.org/TR/html51/editing.html#the-tabindex-attribute
-  return tabIndex > -1 && isFocusable(el) && !isNativelyFocusable(el);
+  return isInTabOrder(el) && !isNativelyFocusable(el);
```
WilcoFiers, line 15: While this is functionally the same, by doing it this way you're running a bunch of unnecessary code in isFocusable. Best to revert here too.

WilcoFiers (changes requested):

WilcoFiers (approved):

PR #3620, opened 2022-08-24
chore: add action to create pr for generated files

straker:
We have two files that are automatically generated on build: `doc/rule-descriptions.md` and `locales/_template.json`. These files can easily become out-of-sync when a pull request makes changes to a rule or check metadata file. If a metadata file gets suggestions and the suggestions are accepted, the generated files would need to be rebuilt and redployed to the pr. Most of the time this doesn't happen and the changes aren't applied to the generated files. This causes any new builds on local branches to have new changes from develop.

This action will run the build on pushes to develop and make a pr with the built files if there are changes. You can see it working here: https://github.com/straker/issue-template-tests/pull/22. If there are no changes, the action will fail and not create a pr (as seen here https://github.com/straker/issue-template-tests/actions/runs/2922890704).

### DIFF
--- a/.github/workflows/update-generated-files.yaml
+++ b/.github/workflows/update-generated-files.yaml
@@ -18,7 +18,9 @@ jobs:
           fetch-depth: 0
 
       - name: Build
-        run: npm run build
+        run: |
+          npm install
+          npm run build
 
       - name: Check for changes
         id: changes
--- a/.github/workflows/update-generated-files.yaml
+++ b/.github/workflows/update-generated-files.yaml
@@ -3,7 +3,7 @@ name: Update generated files
 on:
   push:
     branches:
-      - main
+      - develop
 
 jobs:
   update_genereated_files:
@@ -27,6 +27,6 @@ jobs:
           git checkout -b $branch_name
           git commit -m "chore: sync generated files"
           git push origin $branch_name
-          gh pr create --base main --head $branch_name --title "chore: sync generated files" --body ""
+          gh pr create --base develop --head $branch_name --title "chore: sync generated files" --body ""
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/update-generated-files.yaml
```diff
@@ -0,0 +1,32 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Build
+        run: npm run build
+
+      - name: Create pull pull-requests
```
straker, line 20: I wanted to use a condition to run this step, but [there's currently a bug in action runners](https://github.com/actions/runner/issues/2033) that prevents 3rd party actions from running.

WilcoFiers (changes requested):
I suggest we do something to avoid opening multiple PRs. Perhaps use a constant branch name. If the branch exists in remote we push the update, if it doesn't then we create the PR. The repo is set up to delete branches after merge, so if there's a remote branch, we can be confident there's a PR.

.github/workflows/update-generated-files.yaml
```diff
@@ -0,0 +1,66 @@
...
+      - name: Check branch exists
+        id: branchExists
+        if: steps.changes.outputs.changes
+        run: |
+          exists=$(git ls-remote --heads origin $BRANCH_NAME)
+          echo $exists | grep . && echo "Branch '$BRANCH_NAME' already exists on remote" || echo "Branch does not exists in remote"
+          echo ::set-output name=exists::"$exists"
+
+      - name: Create pull request
+        if: ${{ steps.changes.outputs.changes && !steps.branchExists.outputs.exists }}
```
straker, line 40: You can see a working example of using the the step outputs:

- [no changes](https://github.com/straker/issue-template-tests/runs/8076907049?check_suite_focus=true)
- [changes and branch does not exist](https://github.com/straker/issue-template-tests/runs/8077228637?check_suite_focus=true)
- [changes and branch already exists](https://github.com/straker/issue-template-tests/runs/8077261494?check_suite_focus=true)

WilcoFiers (approved):

PR #3621, opened 2022-08-26
refactor(empty-table-header): use virtual node

dbowling:
use virtual node

references: #3473


test/integration/virtual-rules/empty-table-header.js
```diff
@@ -0,0 +1,142 @@
+describe('empty-table-header virtual-rule', function () {
```
straker, line 1: We'll want a test that should return incomplete if the `th` element does not have the `children` set to an empty array. Similar to [button-name](https://github.com/dequelabs/axe-core/blob/develop/test/integration/virtual-rules/button-name.js#L116-L125)

straker (changes requested):
We'll also want to add a test in `test/commons/table/get-scope.js` that verifies `auto` is returned if passed an element without an `actualNode` or in the tree.

WilcoFiers (approved):

PR #3631, opened 2022-09-02
chore: run prettier on generated files

straker:
The `locales/_template.json` file was always dirty when running `npm run build`. This caused there to always be changes in local development. The reason for this was that the file is built without a newline character at the end of the file. However, when the file was run through prettier, prettier would add the newline character and thus git would detect no changes. This is why the new action to [sync generated files](https://github.com/dequelabs/axe-core/runs/8157911949?check_suite_focus=true) is failing with detected changes but then when it tries to commit those changes there are no longer any changes.

To fix this, I created a new grunt task that just runs prettier on our generated files so that they won't be built anymore with changes.


WilcoFiers (approved):

stephenmathieson:
Any reason not to have Prettier run after each commit, [like we do elsewhere](https://github.com/dequelabs/sax-team/blob/master/.github/workflows/format.yml)?

straker:
We already run it before each commit so files should be pretty before a pull request is made

stephenmathieson:
You're right, it _should_ be. I've seen countless comments similar to "_Oh weird, Prettier didn't run for me_". Letting a robot take care of it would prevent this entirely.

Gruntfile.js
```diff
@@ -253,10 +253,26 @@ module.exports = function (grunt) {
           return ['./axe' + lang + '.js', './axe' + lang + '.min.js'];
         })
       }
+    },
+    prettier: {
+      genearted: {
```
stephenmathieson, line 258:
```suggestion
      generated: {
```

stephenmathieson (changes requested):
Typo.

Also, why do we need Grunt for this?

straker:
> You're right, it should be. I've seen countless comments similar to "Oh weird, Prettier didn't run for me". Letting a robot take care of it would prevent this entirely.

Sure, that is true. However this would not prevent the problem that this pr is trying to solve.

> Also, we've been talking about getting rid of Gruntfile.js for years now. I don't think we should be adding things to it. Why do we need grunt for this?

Until we get rid of it it's the current way to build our files and code. It's the best place to put this as the code that generates the files is in the build pipeline

build/tasks/prettier.js
```diff
@@ -0,0 +1,19 @@
...
+    'Task to run the prettier on generated files',
+    function () {
+      const files = grunt.task.current.data.files;
+
+      files.forEach(file => {
+        const src = Array.isArray(file.src) ? file.src : [file.src];
+
+        src.forEach(entry => {
+          execSync(`npx prettier --write ${entry}`, { stdio: 'inherit' });
+        });
```
stephenmathieson, line 15: Why start a new process for each file?

```suggestion
        execSync(`npx prettier --write ${src.join(' ')`, { stdio: 'inherit' })
```

Or better yet, just use `require('prettier').format()` and keep everything in the same process.

straker: I tried, but It wouldn't run the correct parser on the files, it always tried to use `babel` and throw errors. I couldn't figure out how to load the proper config even with `.resolveConfig`

stephenmathieson:
> Until we get rid of it it's the current way to build our files and code. It's the best place to put this as the code that generates the files is in the build pipeline

Why not add `"postbuild": "prettier --write ."` to `package.json`? That way, everything is "prettier" and we don't have to refactor anything when we remove Grunt.

WilcoFiers:
I like Stephen's suggestion. Lets stay not do things in Gunt if we can avoid it.

package.json
```diff
@@ -108,7 +108,8 @@
     "fmt": "prettier --write .*.json *.{json,md,js} **/*.ts './{.circleci,.github,build,doc,lib,locales,test,typings}/**/*.{json,md,js,ts,html}'",
     "prepare": "husky install",
     "prebuild": "node ./build/check-node-version.js",
-    "pretest": "node ./build/check-node-version.js"
+    "pretest": "node ./build/check-node-version.js",
+    "postbuild": "prettier --write ./locales/_template.json ./doc/rule-descriptions.md"
```
stephenmathieson, line 112: I see no reason why we shouldn't format everything here.

```suggestion
    "postbuild": "prettier --write ."
```

stephenmathieson (dismissed):

package.json
straker: Doing this will add so much noise to prs. We only run prettier on files changed so we only include files that were changed.

stephenmathieson: Doing this will cause 1 PR to generate noise; after that, everything will be formatted (therefore no more noise).

straker: This also removes the need to run prettier on precommit. I'd rather we get this small change in to fix a problem and have a larger discussion about how we want to do prettier in a separate ticket.

stephenmathieson: This does not remove the need for the precommit hook. It's very possible to change a file in this repository without running `npm run build`.

What is your hesitation here? I don't understand why this change isn't being accepted.

straker: A few reasons:

1. running prettier on every file as part of this pr is going to cause merge conflicts to every open pr, especially the pr which refactors the entire codebase away from our visibility functions. I've been waiting for prs to be settled before running prettier on the entire code base. 
2. this pr, which was a 5 minute change, has now taken multiple days in order to fix something that should have been trivial. I'd like to fix the problem before scope creeping to a larger discussion about how prettier is run on the repo

package.json
```diff
@@ -108,7 +108,8 @@
     "fmt": "prettier --write .*.json *.{json,md,js} **/*.ts './{.circleci,.github,build,doc,lib,locales,test,typings}/**/*.{json,md,js,ts,html}'",
```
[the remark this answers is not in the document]
straker: This is not a valid call to `prettier` and results in the usage output:

```
> prettier --write

Usage: prettier [options] [file/dir/glob ...]

By default, output is written to stdout.
Stdin is read if it is piped to Prettier and no files are given.
```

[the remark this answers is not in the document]
straker: This will lint things we haven't been linting before like CSS, YAML, and XHTML files. I see no reason not to be linting those though, but just wanted to document that.

stephenmathieson (approved):

Issue #3633, opened 2022-09-05
valid-lang rule should test aria-hidden text


WilcoFiers:
### Product

axe-core

### Product Version

4.4.3

### Latest Version

- [X] I have tested the issue with the latest version of the product

### Issue Description

For the valid-lang rule, axe-core reports the following as passed:

```html
<article lang="english">
  <p aria-hidden="true">
    They wandered into a strange Tiki bar on the edge of the small beach town.
  </p>
</article>
```

This is a failed example of ACT, and so axe-core should fail it too. Text in aria-hidden sections aren't entirely hidden to assistive technologies. Plus there are non-AT benefits to ensuring correct language is used.

padmavemulapati:
Validated with the latest axe-core develop branch code base,
test script:
```
<article lang="english">
  <p aria-hidden="true">
    They wandered into a strange Tiki bar on the edge of the small beach town.
  </p>
</article>
```
and `valid-lang` rule for the specified script is not passing now, it is listing under failure.


![Image](https://user-images.githubusercontent.com/47409579/190316704-db75ace8-c910-4642-9234-6666a613ca76.png)

PR #3634, opened 2022-09-05
fix(valid-lang): run on aria-hidden text

WilcoFiers:
Closes issue: #3633

### DIFF
--- a/lib/commons/aria/get-element-unallowed-roles.js
+++ b/lib/commons/aria/get-element-unallowed-roles.js
@@ -3,7 +3,7 @@ import getImplicitRole from './implicit-role';
 import getRoleType from './get-role-type';
 import isAriaRoleAllowedOnElement from './is-aria-role-allowed-on-element';
 import { tokenList, isHtmlElement, getNodeFromTree } from '../../core/utils';
-import AbstractVirtuaNode from '../../core/base/virtual-node/abstract-virtual-node';
+import AbstractVirtualNode from '../../core/base/virtual-node/abstract-virtual-node';
 
 // dpub roles which are subclassing roles that are implicit on some native
 // HTML elements (img, link, etc.)
@@ -55,7 +55,7 @@ function getRoleSegments(vNode) {
  */
 function getElementUnallowedRoles(node, allowImplicit = true) {
   const vNode =
-    node instanceof AbstractVirtuaNode ? node : getNodeFromTree(node);
+    node instanceof AbstractVirtualNode ? node : getNodeFromTree(node);
   // by pass custom elements
   if (!isHtmlElement(vNode)) {
     return [];
--- a/lib/commons/dom/is-hidden-with-css.js
+++ b/lib/commons/dom/is-hidden-with-css.js
@@ -1,5 +1,6 @@
 import getComposedParent from './get-composed-parent';
 import { getNodeFromTree } from '../../core/utils';
+import AbstractVirtualNode from '../../core/base/virtual-node/abstract-virtual-node';
 
 /**
  * Determine whether an element is hidden based on css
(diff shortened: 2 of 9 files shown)


test/commons/dom/is-hidden-with-css.js
```diff
@@ -1,10 +1,11 @@
-describe('dom.isHiddenWithCSS', function() {
+describe('dom.isHiddenWithCSS', function () {
   'use strict';
 
   var fixture = document.getElementById('fixture');
   var shadowSupported = axe.testUtils.shadowSupport.v1;
-  var isHiddenWithCSSFn = axe.commons.dom.isHiddenWithCSS;
+  var isHiddenWithCSS = axe.commons.dom.isHiddenWithCSS;
```
WilcoFiers, line 6: I have no idea why we did this, but I thought it should be changed.

lib/commons/aria/get-element-unallowed-roles.js
```diff
@@ -3,7 +3,7 @@ import getImplicitRole from './implicit-role';
 import getRoleType from './get-role-type';
 import isAriaRoleAllowedOnElement from './is-aria-role-allowed-on-element';
 import { tokenList, isHtmlElement, getNodeFromTree } from '../../core/utils';
-import AbstractVirtuaNode from '../../core/base/virtual-node/abstract-virtual-node';
+import AbstractVirtualNode from '../../core/base/virtual-node/abstract-virtual-node';
```
WilcoFiers, line 6: Unrelated, but I spotted this typo. Figured I might as well fix it.

test/integration/rules/valid-lang/valid-lang.html
```diff
@@ -10,11 +10,17 @@
 <p xml:lang="en-US" lang="gibberish-US" id="fail3">Mix</p>
 <p lang=" " id="fail4">English</p>
 
-<p lang="invalid" id="pass7">  <!-- ignore -->  </p>
+<p lang="invalid" id="pass7"><!-- ignore --></p>
```
WilcoFiers, line 13: Hmm, wonder why prettier did that.... Not sure it matters?

locales/_template.json
```diff
@@ -965,7 +965,8 @@
         "default": "Element's default semantics were not overridden with role=\"none\" or role=\"presentation\"",
         "globalAria": "Element's role is not presentational because it has a global ARIA attribute",
         "focusable": "Element's role is not presentational because it is focusable",
-        "both": "Element's role is not presentational because it has a global ARIA attribute and is focusable"
+        "both": "Element's role is not presentational because it has a global ARIA attribute and is focusable",
+        "iframe": "Using the \"title\" attribute on an ${data.nodeName} element with a presentational role behaves inconsistently between screen readers"
```
straker, line 969: This has been fixed in #3632 and doesn't need to be in the pr.

lib/commons/dom/is-hidden-with-css.js
```diff
@@ -10,8 +11,10 @@ import { getNodeFromTree } from '../../core/utils';
  * @param {Boolean} descendentVisibilityValue (Optional) immediate descendant visibility value used for recursive computation
  * @return {Boolean} the element's hidden status
  */
-function isHiddenWithCSS(el, descendentVisibilityValue) {
-  const vNode = getNodeFromTree(el);
+function isHiddenWithCSS(node, descendentVisibilityValue) {
```
straker, line 14: This change isn't necessary once the isVisible refactor is in (the function name changed and it already handles virtual nodes). Do we still want to make this change if `isHiddenWithCSS` is being deprecated and not used anymore in our code?

straker (changes requested):

lib/commons/dom/is-hidden-with-css.js
WilcoFiers: I didn't want to create a dependency on your refactor. I don't think doing both is a problem.

locales/_template.json
WilcoFiers: Doesn't matter -- but I updated to develop.

straker (approved):

Issue #3489, opened 2022-06-01
aria-valid-attr-value should permit null values for attributes


WilcoFiers:
Code like the following should not fail the `aria-valid-attr-value` rule:

```html
<div role="checkbox" aria-checked>Accept terms and conditions</div>
```

We could consider reporting this as incomplete, possibly. This is sort of an odd thing to happen, but it isn't a guaranteed failure. `aria-checked` is not a required attribute for `checkbox`, so this is sort of like not having the attribute there at all.

I'm not sure this holds for all ARIA attributes, we'll need to check that. And I think we'll need to confirm that aria-required-attr isn't passed with null attributes. It seems to me that if we're going to fail null attributes its only for attributes that are required.

WilcoFiers:
WAI-ARIA 1.2 changes this a bit. checkbox no longer has a default aria-checked value.

I did some testing
|           | NVDA+FF     | JAWS+GC     | VO+Safari   |
|-----------|-------------|-------------|-------------|
| no attr   | not checked | not checked | Not checked |
| null attr | not checked | not checked | Not checked |
| ""        | not checked | not checked | Not checked |
| "false"   | not checked | not checked | Not checked |
| "FALSE"   | checked     | not checked | Not checked |
| " false " | checked     | checked     | checked     |
| "invalid" | checked     | checked     | checked     |

On Google Chrome, NVDA and Narrator behave the same as JAWS.


I think how we address this issue depends on how we deal with #3461. If we change this rule to be a best practice I think it's okay to be stricter. If not, we could potentially just flag everything as "needs review". The way browsers have implemented this attribute is pretty smart.

Something like the following is will work consistently, even though both states are trigged through invalid values:

```jsx
const MyCheckbox = (props) => (<div role="checkbox" aria-checked={props.checked ? 'checked' : ''} />
```

WilcoFiers:
We're going with the less strict option.

padmavemulapati:
Validated with the latest axe-core develop branch code base,
Test script:
```
<div role="checkbox" aria-checked>Accept terms and conditions</div>
<div aria-invalid></div>
<div aria-hidden></div>
<div aria-checked=""></div>
```
And `aria-valid-attr-value` rule is part of incomplete on empty and null attributes.


![Image](https://user-images.githubusercontent.com/47409579/190192653-203cfc39-ff73-42f4-9ca5-8380b8f8f279.png)

PR #3635, opened 2022-09-05
fix(aria-valid-attr-value): report empty values as incomplete

WilcoFiers:
There is no real difference between an empty ARIA attribute, a null attribute, and not having the attribute at all. Failing empty ARIA attributes therefor seems a little strict. They shouldn't be passed either though, since it is pretty easy to mistakenly assume `<input aria-invalid>` means the input is invalid. Reporting `incomplete` seems more appropriate. This PR makes that change, along with:

- String type attributes, like aria-valuenow are still not allowed empty
- `aria-invalid` is now reported as incomplete when empty, rather than passed
- Extra tests were added to aria-required-attr to ensure empty attributes when required are failed there

Closes issue: #3489

### DIFF
--- a/lib/checks/aria/aria-valid-attr-value-evaluate.js
+++ b/lib/checks/aria/aria-valid-attr-value-evaluate.js
@@ -1,4 +1,5 @@
 import { validateAttrValue } from '../../commons/aria';
+import standards from '../../standards';
 
 /**
  * Check that each ARIA attribute on an element has a valid value.
@@ -24,7 +25,7 @@ import { validateAttrValue } from '../../commons/aria';
  * @memberof checks
  * @return {Mixed} True if all ARIA attributes have a valid value. Undefined for invalid `aria-current` or `aria-describedby` values. False otherwise.
  */
-function ariaValidAttrValueEvaluate(node, options, virtualNode) {
+export default function ariaValidAttrValueEvaluate(node, options, virtualNode) {
   options = Array.isArray(options.value) ? options.value : [];
 
   let needsReview = '';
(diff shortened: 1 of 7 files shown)


lib/checks/aria/aria-valid-attr-value-evaluate.js
```diff
@@ -106,30 +107,35 @@ function ariaValidAttrValueEvaluate(node, options, virtualNode) {
     } catch (e) {
       needsReview = `${attrName}="${attrValue}"`;
       messageKey = 'idrefs';
+      return;
```
WilcoFiers, line 110: Avoid getting into that if statement below, as we don't want both `needsReview` and `invalid` set.

lib/checks/aria/aria-valid-attr-value-evaluate.js
```diff
@@ -106,30 +107,35 @@ function ariaValidAttrValueEvaluate(node, options, virtualNode) {
...
+      if (attrValue === '' && !isStringType(attrName)) {
+        needsReview = attrName;
+        messageKey = 'empty';
+      } else {
+        invalid.push(`${attrName}="${attrValue}"`);
+      }
     }
   });
 
-  if (needsReview) {
```
WilcoFiers, line 119: Report invalid before needsReview.

test/checks/aria/aria-required-attr.js
```diff
@@ -1,48 +1,72 @@
-describe('aria-required-attr', function() {
+describe('aria-required-attr', function () {
```
WilcoFiers, line 1: Check didn't change, just want to better prove empty values are not allowed under this check.

straker (approved):

Issue #3499, opened 2022-06-03
Update axe-core ACT tests


WilcoFiers:
For axe-core 4.5 we're doing a lot of work to improve consistency with ACT rules. This is going to let us run more rules in the act-test job. We'll need to make sure to enable those.

straker:
Use https://wai-wcag-act-rules.netlify.app/standards-guidelines/act/implementations/axe-core/ to determine which rules to bring over (based on which ones we are conformant with).

Issue #3636, opened 2022-09-05
ACT Tests should run inside iframes


WilcoFiers:
Currently the ACT tests don't inject axe-core into any iframes. Because of that some of the rule mappings don't quite work. Notably, the "Iframe with negative tabindex has no interactive elements" ACT rule cannot be tested because its test cases require axe-core to run inside iframes.

We should update the test runner to inject axe-core into each frame, and only run the tests until the page and all its frames are fully loaded.

PR #3637, opened 2022-09-05
chore: Update ACT test cases to run from W3C site

WilcoFiers:
Closes issue: #3499

Enable tests for all rules axe-core has currently implemented, except for 1 (see https://github.com/dequelabs/axe-core/issues/3636)

### DIFF
--- a/test/act-mapping/aria-props-allowed.json
+++ b/test/act-mapping/aria-props-allowed.json
@@ -1,0 +1,5 @@
+{
+  "id": "5c01ea",
+  "title": "ARIA state or property is permitted",
+  "axeRules": ["aria-allowed-attr"]
+}
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -197,7 +197,7 @@ jobs:
       # install ACT rules
       # install first as for some reason installing a single package
       # also re-installs all repo dependencies as well
-      - run: npm install act-rules/act-rules.github.io#master
+      - run: npm install w3c/wcag-act-rules#main
       - run: npx browser-driver-manager install chromedriver --verbose
       - run: npm run build
       - run: npm run test:act
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,6 @@
         "@babel/preset-env": "^7.18.6",
         "@babel/runtime-corejs3": "^7.18.6",
         "@deque/dot": "^1.1.5",
-        "act-rules.github.io": "github:act-rules/act-rules.github.io#a759e99f",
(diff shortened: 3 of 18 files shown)


test/act-mapping/aria-props-allowed.json
```diff
@@ -0,0 +1,5 @@
+{
+  "id": "5c01ea",
```
straker, line 2: Question: how do we want to name these files? We don't appear to have a naming convention for axe-core, and I'm not sure where the `aria-props-allowed` name comes from. Is that what it's called in ACT (but not listed in the rule itself)?

straker (changes requested):
See question above

test/act-mapping/aria-props-allowed.json
WilcoFiers: I was sort of just winging it. ACT rules do have a file name. I could copy those I suppose.

test/act-rules/README.md
```diff
@@ -0,0 +1,5 @@
+# ACT Rules
+
+Each ACT rule that axe-core is consistent with has a test file in this directory. These tests use the `act-runner.js` script, which loads the test cases from `node_modules/wcag-act-rules`. The ACT runner accepts a `id` and `title` from the ACT rule, and an array of `axeRules` that map to this ACT rule.
```
straker, line 3:
```suggestion
Each ACT rule that axe-core is consistent with has a test file in this directory named [after the ACT rule](https://github.com/act-rules/act-rules.github.io/tree/develop/_rules). These tests use the `act-runner.js` script, which loads the test cases from `node_modules/wcag-act-rules`. The ACT runner accepts a `id` and `title` from the ACT rule, and an array of `axeRules` that map to this ACT rule.
```

test/act-rules/act-runner.js
```diff
@@ -0,0 +1,61 @@
...
+    after(async () => {
+      await driver.close();
+      chromedriver.stop();
+    });
+
+    testcases.forEach(testcase => {
+      const shouldRun = testcase.relativePath.match(/\.(xhtml|html?)$/);
+      (shouldRun ? it : xit)(testcase.testcaseTitle, async () => {
+        await driver.get(`${addr}/${testcase.relativePath}`);
+        const builder = new AxeBuilder(driver, axeSource);
```
straker, line 47: Nice, I like this way of doing the tests

straker (changes requested):

straker (approved):

Issue #3534, opened 2022-07-06
Drop support for using axe-core in Internet Explorer


WilcoFiers:
IE11 is no longer supported by Microsoft, many dependencies we use no longer support it. It's time to let go. In the next version of axe-core, IE will no longer be supported. The main thing this means is:

1. Documentation needs to be updated
2. IE tests will no longer run in CI
3. Most, if not all polyfills can be pulled.
4. There's a color parser somewhere that was only created for IE which I think we can pull
5. We can remove all the axe.testUtils.isIE11 things in our tests

There may be a bunch of IE specific code paths. Pulling those out is going to take a while, we'll probably do that gradually, rather than as one big chunk. Those are going to be difficult to find.

**important**: This is different from us dropping IE11 as an accessibility supported browser (see #3474). This ticket is specifically about dropping support for running axe-core in IE11, it is unrelated to testing for accessibility issues that only occur in JAWS+IE.

WilcoFiers:
Dropping support for unmaintained projects is not a breaking change IMO. Our [backward compabitiiliby](https://github.com/dequelabs/axe-core/blob/develop/doc/backwards-compatibility-doc.md) doc doesn't list this as a breaking change.

WilcoFiers:
So conclusion, we're deprecating support for IE11 as of axe-core 4.5. It will be fully removed as of 5.0. With that deprecation though, we'll also stop running automated tests against IE11. Anyone experiencing issues in using axe-core in IE11 can still report them, so we can investigate.

PR #3644, opened 2022-09-07
feat: Deprecated Internet Explorer support

WilcoFiers:
As of axe-core 4.5, Internet Explorer 11 is deprecated. Support will be fully dropped with the release of axe-core 5.0, which will come at least 6 months after the 5.0 release to give sufficient time to upgrade.

As of this pull request, we will no longer run automated tests on Internet Explorer. Any bugs reported against IE11 will be investigated and addressed, and the axe-core team will continue to keep limitations imposed by Internet Explorer in mind.

Closes issue: #3534


.circleci/config.yml
```diff
@@ -52,18 +52,6 @@ jobs:
           key: v9-cache-unix-{{ checksum "package-lock.json" }}
```
straker, line 52: There's also the `orbs: win`, `restore_dependency_cache_win`, and `win_box` steps that can be removed

test/commons/color/get-background-color.js
```diff
@@ -807,28 +802,31 @@ describe('color.getBackgroundColor', function() {
...
-  (shadowSupported ? it : xit)('finds colors in shadow boundaries', function() {
-    fixture.innerHTML = '<div id="container"></div>';
-    var container = fixture.querySelector('#container');
-    var shadow = container.attachShadow({ mode: 'open' });
-    shadow.innerHTML =
-      '<div style="background-color: black;">' +
-      '<span id="shadowTarget" style="color: #ccc;">Text</span>' +
-      '</div>';
-    axe.testUtils.flatTreeSetup(fixture);
+  (shadowSupported ? it : xit)(
```
straker, line 805: With IE11 gone all our tests run in browsers that support shadow dom (Chrome and Firefox). We should get rid of `shadowSupported` as well

test/integration/full/preload-cssom/preload-cssom.js
```diff
@@ -429,23 +426,17 @@ describe('preload cssom integration test', function() {
     commonTestsForRootNodeAndNestedFrame();
   });
 
-  describe('tests for nested document', function() {
-    before(function() {
-      if (!nestedFrame || isIE11) {
```
straker, line 434: Is there a chance that `nestedFrame` will ever be `false` that we need the if statement still?

straker (changes requested):

test/integration/full/preload-cssom/preload-cssom.js
WilcoFiers: I have no idea why this was ever there. It wasn't necessary. (otherwise the tests would have failed)

straker (approved):

PR #3657, opened 2022-09-09
feat: add axe.utils.memoizeObjectArgs

straker:
This adds a function which allows us to memoize functions that take objects as parameters which are created at call time.

For example, the following does not memoize correctly and would call the function two times

```js
const foo = memoize(function({ optionA, optionB }) {})

foo({ optionA: 1, optionB: 2 })
foo({ optionA: 1, optionB: 2 })
```

The fix was to pass the `normalizer` option to memoize which lets us specify how to resolve the arguments. The normalizer uses `deepEqual` to compare objects rather than strict equal comparison, letting us memoize objects correctly.

This PR also removes the ES5 requirement from our tests now that we no longer run in IE11

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -71,10 +71,7 @@ module.exports = {
   overrides: [
     {
       files: ['lib/**/*.js'],
-      excludedFiles: [
-        'lib/core/reporters/**/*.js',
-        'lib/**/*-after.js'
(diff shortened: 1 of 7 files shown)


straker:
Going to write a proposal instead as this looks like it needs a larger discussion

PR #3663, opened 2022-09-20
fix(html-lang-valid): only run rule when attribute has value

### LINKED ISSUE #3624: Empty <html> lang attributes report as fail of html-has-lang, pass for html-lang-valid
### Product

axe-core

### Product Version

4.4.2

### Latest Version

### Issue Description

For `<html lang="">`, `html-has-lang` fails, and `html-lang-valid` passes. This is a similar issue to #3623, where I would expect the rules to be consistent with one-another.

#### Expectation

Both rules should be consistent. If `lang=""` is indeed a valid lang attribute value, then `html-has-lang` should pass (both rules should pass). If it’s not a valid value, then either both rules should fail, or only `html-lang-valid` should fail.

#### Actual

Currently, `<html lang="">` passes `html-lang-valid`.

#### How to Reproduce

Here are two sites where this happens, and Axe CLI commands:

```bash
axe -r html-has-lang,html-lang-valid https://www.koldijkbv.nl/ --save empty-lang.json
axe -r html-has-lang,html-lang-valid https://loyalty.posnania.eu/ --save empty-lang.json
```

In each case, see the `html-has-lang` violation while `html-lang-valid` is marked as a "pass".

#### Additional context

Similarly to #3623, I was also tempted to report this to Lighthouse instead / in addition to Axe, as its overly simplified audit messages make this even more confusing. It would report:

- An error "`<html> element does not have a [lang] attribute`"
- And a pass "`<html> element has a valid value for its [lang] attribute`"

thibaudcolas:
**Depends on #3669**. Closes issue #3624. This updates the rule’s selector as discussed so the rule is reported as inapplicable when encountering empty `lang` or `xml:lang` attributes.

It would be nice if this detail of the implementation was documented somewhere but I’m not sure where would be a good place.

For tests, I added another "applicability" test case similar to `html-lang-valid virtual-rule – is inapplicable without lang or xml:lang`. I also considered adding a test case to `test/integration/full/html-lang-valid/html-lang-valid.js`, however we’re not checking applicability in there currently, and it seems the only thing we could do is add another frame and check it _isn’t_ in the result – so I decided to leave this out. Happy to revisit if needed.

---

This is failing currently in tests, with existing passes and violations not getting picked up by the new selector. I’m not entirely sure why – could be because of how Axe converts selectors to its own expression format but that’s just a guess.


thibaudcolas:
Thank you for looking into this @straker. I’m feeling out of my depth here so I’ll leave this as-is until someone else can devise  a fix.

straker:
Have a pr that should fix the issue.

thibaudcolas:
Rebased onto the latest `develop`, with the fix so this now seems to be working as expected. I believe this is now ready for review, though note I’ve only tested the change with the added unit test so far.

straker (changes requested):
Looking good! We should also update the [html-valid-lang integration tests](https://github.com/dequelabs/axe-core/tree/develop/test/integration/full/html-lang-valid) to check for this. If you create two more iframes inside the `frames` dir, one using `lang` and the other for `xml:lang` and make them both empty, and then add the frames to the `html-valid-lang.html` file, that should be enough (since it shouldn't modify the passes/violation count of the test)

straker:
@thibaudcolas anything I can do to help with the pr?

straker:
@thibaudcolas I'll go ahead and add the tests so this can be reviewed and merged.

WilcoFiers:
[x] Reviewed for security

thibaudcolas:
Thank you for picking this up! I had a busy couple months and wouldn’t have been able to get to this until now, it’s great to see it having gone through.

PR #3668, opened 2022-09-22
chore: run prettier after develop build

straker:
In #3631, moving the prettier step out of the Gruntfile and into a `postbuild` script only solved half the problem. Running `npm run develop` would not cause the `build` npm script to run so the `locales/_template` file would still be out-of-sync. 

This change is minimal in order to use what we have already created and still get prettier to run after running `develop`.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,3 +1,5 @@
+var execSync = require('child_process').execSync;
+
 /*eslint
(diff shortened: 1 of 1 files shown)


WilcoFiers (approved):
Yuck... But yeah

stephenmathieson:
😅

PR #3669, opened 2022-09-22
fix(utils.matches): fix attribute exists selector to match empty attributes

straker:
Fixes the issue discovered in #3663 where we couldn't properly match `[attr=""]`

### DIFF
--- a/test/core/utils/matches.js
+++ b/test/core/utils/matches.js
@@ -121,7 +121,7 @@ describe('utils.matches', function () {
       assert.isTrue(matches(virtualNode, '[foo~="baz"]'));
     });
 
-    it('returns true if attribute matches not having value', function () {
+    it('returns true if attribute matches having an empty value', function () {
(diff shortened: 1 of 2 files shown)


test/core/utils/matches.js
```diff
@@ -1,224 +1,245 @@
...
 
-    it('returns true if attribute has value', function() {
+    it('returns true if attribute has value', function () {
       var virtualNode = queryFixture(
         '<span id="target" foo="bar baz" bar="foo" baz="bar"></span>'
       );
       assert.isTrue(matches(virtualNode, '[foo~="baz"]'));
     });
+
+    it('returns true if attribute matches not having value', function () {
```
WilcoFiers, line 124:
```suggestion
    it('returns true if attribute matches having an empty value', function () {
```

WilcoFiers (dismissed):

WilcoFiers (approved):

PR #3671, opened 2022-09-23
docs: add code-of-conduct.md

WilcoFiers:


straker (changes requested):
I believe we're following the old *nix pattern to uppercase important files, so this should be `CODE_OF_CONDUCT.md`

straker (approved):

Issue #3247, opened 2021-11-02
element with 'overflow: hidden' still has children spans reported in color-contrast violations


karanbirsingh:
Product: axe-core

Expectation: If a span is visually hidden because its parent specifies "overflow: hidden", the span should be excluded from color contrast results.

Actual: axe-core reports color-contrast violations on hidden spans in the following codepen: https://codepen.io/karansin/pen/QWMOKbj 

Motivation: We (Accessibility Insights) received a report from a user whose website looks something like the codepen. The color-contrast result is correct in one sense (the foreground & background are not accessible), but the elements are hidden. The reporting user expects the hidden spans to be excluded from color-contrast results. Does this seem like a false positive? Thanks!

<pre><code>
axe-core version: 4.3.2, I reproduced the error using a locally built axe.min.js from develop as well, commit 112b960ee95b6a6abfb38a15b7092d9847512f0f

Browser and Assistive Technology versions
Chrome 95.0.4638.54
Windows

straker:
I believe this is similar to #2806

PR #3676, opened 2022-09-26
fix(is-visible-on-screen): ignore elements hidden by overflow:hidden

straker:
This also fixes a bug in `svg-non-empty-title` that was discovered by this change in that it used `visibleVirtual` to get the title text. `title` elements aren't visible so it should instead use `subtreeText` and allow hidden elements.

For the previous `overflowHidden` function, I found that it was added to handle [a screen reader only styling](https://github.com/dequelabs/axe-core/pull/2985). I made sure the code still works with the new functionality.

Closes issue: #3247 and #2806

### DIFF
--- a/lib/commons/dom/visibility-methods.js
+++ b/lib/commons/dom/visibility-methods.js
@@ -3,7 +3,8 @@ import {
   closest,
   getRootNode,
   querySelectorAll,
-  escapeSelector
+  escapeSelector,
+  memoize
 } from '../../core/utils';
 
 const clipRegex =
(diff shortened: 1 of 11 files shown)


lib/commons/dom/visibility-methods.js
```diff
@@ -88,21 +90,58 @@ export function scrollHidden(vNode) {
...
+    return false;
+  }
 
-  return (
-    vNode.getComputedStylePropertyValue('position') === 'absolute' &&
-    (elHeight < 2 || elWidth < 2) &&
-    vNode.getComputedStylePropertyValue('overflow') === 'hidden'
-  );
+  const rect = vNode.boundingClientRect;
+  const nodes = getOverflowHiddenNodes(vNode);
```
WilcoFiers, line 107:
```suggestion
  const nodes = getOverflowHiddenAncestors(vNode);
```

lib/commons/dom/visibility-methods.js
```diff
@@ -88,21 +90,58 @@ export function scrollHidden(vNode) {
...
+    return ancestors;
+  }
+
+  const overflow = vNode.getComputedStylePropertyValue('overflow');
+
+  if (overflow === 'hidden') {
+    ancestors.push(vNode);
+  }
+
+  return ancestors.concat(getOverflowHiddenNodes(vNode.parent));
```
WilcoFiers, line 142: Double points for the clean recursion 😄.

WilcoFiers (dismissed):
Mostly done, wrapping up for the day though.

test/commons/math/rects-touch.js
```diff
@@ -0,0 +1,50 @@
...
+    rectA.right = 40;
+
+    assert.isFalse(rectsTouch(rectA, rectB));
+  });
+
+  it('returns false when rects are next to one another', () => {
+    rectA.left = 20;
+    rectA.right = 25;
+
+    assert.isFalse(rectsTouch(rectA, rectB));
```
WilcoFiers, line 39: Going off the name, this seems like it should be true. I think if you want this to be false when the edges touch, this should probably be called `rectsOverlap`. WDYT?

test/commons/math/rects-touch.js
```diff
@@ -0,0 +1,50 @@
...
+
+  it('returns true when rects touch', () => {
+    assert.isTrue(rectsTouch(rectA, rectB));
+  });
+
+  it('should not matter on order', () => {
+    assert.isTrue(rectsTouch(rectB, rectA));
+  });
+
+  it('returns false when rects do not touch', () => {
```
WilcoFiers, line 28: You should probably test vertical misalignment too, for the sake of completeness.

test/commons/math/rects-touch.js
```diff
@@ -0,0 +1,50 @@
+describe('math.rects-touch', function () {
+  const rectsTouch = axe.commons.math.rectsTouch;
+
+  let rectA, rectB;
+  beforeEach(() => {
+    rectA = {
```
WilcoFiers, line 6: Not strictly necessary, but you could consider using DOMRect in these tests. (I only learned of these a week ago, otherwise I might've used them in target-size stuff.)

WilcoFiers (changes requested):
Couple minor points

test/commons/math/rects-touch.js
straker: Not a bad idea. If the tests were more complicated I might consider that, but since it's a simple test I'll just leave it as is.

WilcoFiers (approved):

Issue #3655, opened 2022-09-08
Code quality improvements from isVisible refactor


straker:
From https://github.com/dequelabs/axe-core/pull/3351, listing changes we can make to improve the code that we didn't want to make in the large pr

- [x] `lib/checks/keyboard/accesskeys-evaluate.js` - `isVisibleOnScreen` does not need to be used to determine if label is visible (https://github.com/dequelabs/axe-core/pull/3351#discussion_r964707715)
- [x] `lib/checks/label/multiple-label-evaluate.js` - based on the comment, `!isHiddenForEveryone` may be more appropriate than `isVisibleOnScreen` (https://github.com/dequelabs/axe-core/pull/3351#discussion_r964709669)
- [x] `lib/commons/dom/get-rect-stack.js` - may not need need width/height check if `isVisibleOnScreen` already does that (https://github.com/dequelabs/axe-core/pull/3351#discussion_r964715832)
- [x] `lib/checks/shared/is-on-screen-evaluate.js` - may not need `isOffScreen` as `isVisibleOnScreen` already checks for that (https://github.com/dequelabs/axe-core/pull/3351#discussion_r964713787)
- [x] `lib/rules/is-visible-matches.js` - file name is different than matches function name (https://github.com/dequelabs/axe-core/pull/3351#discussion_r964720738)
- [x] `lib/rules/is-visible-matches.js` - deprecate and replace with `is-visible-on-screen-matches` (https://github.com/dequelabs/axe-core/pull/3351#discussion_r964722624)
- [x] `lib/rules/landmark-unique-matches.js` - remove `isVisibleForScreenreader` call as it may not be useful (https://github.com/dequelabs/axe-core/pull/3351#discussion_r964724044)
- [x] `lib/commons/dom/visiblity-methods.js` - add tests for the functions when we can build the tests to allow ES6 imports (https://github.com/dequelabs/axe-core/pull/3351#discussion_r964810439)
- [x] `lib/core/core.js` - move all uses of `axe.__thisWillBeDeletedDoNotUse` to its own file (https://github.com/dequelabs/axe-core/pull/3351#discussion_r975172022)

straker:
> lib/rules/landmark-unique-matches.js - remove isVisibleForScreenreader call as it may not be useful (https://github.com/dequelabs/axe-core/pull/3351#discussion_r964724044)

[...]

PR #3678, opened 2022-09-26
refactor: change or remove visibility functions as needed

straker:
Also updated `landmark-unique-matches.js` to use `closest` instead of the deprecated `findUp` while I was in the code looking at the use of `isVisibleForScreenreader`

Part of: https://github.com/dequelabs/axe-core/issues/3655


WilcoFiers (changes requested):
Kinda thinking each of these should have a test case to go with it.

WilcoFiers (approved):

PR #3679, opened 2022-09-26
feat(matches): deprecate is-visible-matches, use is-visible-on-screen-matches

straker:
Just copied the tests from `is-visible-matches` to `is-visible-on-screen-matches`.

Part of https://github.com/dequelabs/axe-core/issues/3655

### DIFF
--- a/lib/rules/is-visible-on-screen-matches.js
+++ b/lib/rules/is-visible-on-screen-matches.js
@@ -1,5 +1,5 @@
 import { isVisibleOnScreen } from '../commons/dom';
 
-export default function isVisibleOnScreenMatches(node) {
-  return isVisibleOnScreen(node);
+export default function isVisibleOnScreenMatches(node, virtualNode) {
+  return isVisibleOnScreen(virtualNode);
 }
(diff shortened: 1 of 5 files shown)


WilcoFiers (approved):

test/rule-matches/is-visible-on-screen-matches.js
```diff
@@ -0,0 +1,32 @@
+describe('is-visible-on-screen-matches', function () {
+  'use strict';
+
+  var rule;
+  var fixture = document.getElementById('fixture');
+  var fixtureSetup = axe.testUtils.fixtureSetup;
+
+  beforeEach(function () {
+    fixture.innerHTML = '';
```
WilcoFiers, line 9: What's this here for? We usually put it in after, although that's not needed anymore either since we have an after in testutils.js that does it for us.

lib/rules/is-visible-on-screen-matches.js
```diff
@@ -0,0 +1,5 @@
+import { isVisibleOnScreen } from '../commons/dom';
+
+export default function isVisibleOnScreenMatches(node) {
+  return isVisibleOnScreen(node);
```
WilcoFiers, line 4: Since we're at it, should we just swap this over to a virtual node?

straker: Good call, I'll do that

WilcoFiers (approved):

Issue #3656, opened 2022-09-09
Object-alt rule should ignore objects that don't load things


WilcoFiers:
The following is from an [ACT inapplicable example](https://wai-wcag-act-rules.netlify.app/standards-guidelines/act/rules/8fc3b6/proposed/#inapplicable-example-8), axe shouldn't fail this:

```html
<object data="/invalid/url/index.html">
  <img src="/WAI/content-assets/wcag-act-rules/test-assets/shared/w3c-logo.png" alt="W3C logo" />
</object>
```

I'm not sure how to directly detect whether an object loaded something, but detecting that its pass-through content is displayed can easily be done by just checking that it's children have a position, for example by calling `getClientRects()` on them. There are probably other ways to do it too.

To do this we'll probably need a new matches method for objects that can filter out objects which show their content.

padmavemulapati:
Validated with the latest axe-core develop branch code base, 
for the test script:
```
<object data="/invalid/url/index.html">
  <img src="/WAI/content-assets/wcag-act-rules/test-assets/shared/w3c-logo.png" alt="W3C logo" />
</object>
```
`object-alt` rule is appearing in `inapplicable` , not failing


![Image](https://user-images.githubusercontent.com/47409579/195615460-ed8c8601-8b2f-4e3b-b64c-4951b3136fc3.png)

PR #3680, opened 2022-09-27
fix(object-alt): ignore unloaded objects

WilcoFiers:
Skip objects on the object-alt rule if they either have no data attribute, or if any of their fallback content is visible.

Closes issue: #3656

I was having difficulty getting the ACT tests to pass because the server doesn't correctly map the URLs. I opened a separate PR for that: #3681

### DIFF
--- a/test/integration/rules/object-alt/object-alt.html
+++ b/test/integration/rules/object-alt/object-alt.html
@@ -1,18 +1,75 @@
-<object id="pass1" title="This object has text"></object>
-<object id="pass2" aria-label="this object has text"></object>
-<span id="label1">this object has text</span>
-<object id="pass3" aria-labelledby="label1"></object>
-<object id="pass4" role="presentation"></object>
-<object id="pass5" role="none"></object>
+<object
+  id="pass1"
+  title="This object has text"
+  data="data:text/html,Object%20content"
+></object>
+<object
+  id="pass2"
+  aria-label="this object has text"
+  data="data:text/html,Object%20content"
+></object>
+<span id="label1" data="data:text/html,Object%20content"
+  >this object has text</span
+>
+<object
+  id="pass3"
+  aria-labelledby="label1"
+  data="data:text/html,Object%20content"
+></object>
+<object
+  id="pass4"
+  role="presentation"
+  data="data:text/html,Object%20content"
+></object>
+<object id="pass5" role="none" data="data:text/html,Object%20content"></object>
 
-<object id="violation1"></object>
-<object id="violation2"><div></div></object>
-<object id="violation3">This object has text.</object>
-<object id="violation4" role="none" tabindex="0"></object>
-<object id="violation5" role="presentation" tabindex="0"></object>
-<object id="violation6" role="none" aria-live="assertive"></object>
-<object id="violation7" role="presentation" aria-live="assertive"></object>
+<object id="violation1" data="data:text/html,Object%20content"></object>
+<object id="violation2" data="data:text/html,Object%20content">
+  <div></div>
+</object>
+<object id="violation3" data="data:text/html,Object%20content">
+  This object has text.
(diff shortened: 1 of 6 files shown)


test/integration/rules/object-alt/object-alt.html
```diff
@@ -1,18 +1,75 @@
...
+<object id="violation1" data="data:text/html,Object%20content"></object>
+<object id="violation2" data="data:text/html,Object%20content">
+  <div></div>
+</object>
+<object id="violation3" data="data:text/html,Object%20content">
+  This object has text.
+</object>
+<object
+  id="violation4"
+  role="none"
```
straker, line 35: This fails [inapplicable example 5](https://wai-wcag-act-rules.netlify.app/standards-guidelines/act/rules/8fc3b6/proposed/#inapplicable-example-5)

test/integration/rules/object-alt/object-alt.html
```diff
@@ -1,18 +1,75 @@
...
+  tabindex="0"
+  data="data:text/html,Object%20content"
+></object>
+<object
+  id="violation5"
+  role="presentation"
```
straker, line 41: This fails [inapplicable example 5](https://wai-wcag-act-rules.netlify.app/standards-guidelines/act/rules/8fc3b6/proposed/#inapplicable-example-5)

test/integration/rules/object-alt/object-alt.html
```diff
@@ -1,18 +1,75 @@
...
+></object>
+<object
+  id="violation7"
+  role="presentation"
+  aria-live="assertive"
+  data="data:text/html,Object%20content"
+></object>
+<object
+  id="violation8"
+  role="img"
```
straker, line 59: This fails [inapplicable example 1](https://wai-wcag-act-rules.netlify.app/standards-guidelines/act/rules/8fc3b6/proposed/#inapplicable-example-1)

test/integration/rules/object-alt/object-alt.html
```diff
@@ -1,18 +1,75 @@
...
-<object id="violation3">This object has text.</object>
-<object id="violation4" role="none" tabindex="0"></object>
-<object id="violation5" role="presentation" tabindex="0"></object>
-<object id="violation6" role="none" aria-live="assertive"></object>
-<object id="violation7" role="presentation" aria-live="assertive"></object>
+<object id="violation1" data="data:text/html,Object%20content"></object>
+<object id="violation2" data="data:text/html,Object%20content">
+  <div></div>
+</object>
+<object id="violation3" data="data:text/html,Object%20content">
```
straker, line 30: This fails [inapplicable example 6](https://wai-wcag-act-rules.netlify.app/standards-guidelines/act/rules/8fc3b6/proposed/#inapplicable-example-6)

straker (dismissed):
I don't thing this is the right approach. Before we only failed 1 of the inapplicable examples (the one this pr fixed), but with the change we now fail 3 others. From reading the ACT rule I don't think it requires looking at if a resource loaded or not. The applicability states that it should only be run on object elements that match the following criteria:

> * The object element is included in the accessibility tree; and
> * The object element has no explicit role; and
> * The object element embeds a resource whose MIME type is either image, or audio or video.

If we want to be more conformant, should we look at the 3rd point rather than if content is loaded or not?

test/integration/rules/object-alt/object-alt.html
WilcoFiers: ACT Covers this in a different rule. We don't.

test/integration/rules/object-alt/object-alt.html
WilcoFiers: ACT doesn't cover this, but this is still an issue.

straker (approved):

Issue #3674, opened 2022-09-24
target-size rule should ignore nested interactive controls.


WilcoFiers:
Spotted this beauty on a video streaming site:

```html
  <a role="link" aria-label="play" tabindex="0">
    <button tabindex="-1" style="margin:1px;">Play</button>
  </a>
```

This is failing the rule because the link is considered to be largely obscured by the button inside of it. I think the solution here is that when we check for obscuring elements that we exclude descendants of the widget.

padmavemulapati:
`target-size` rule off for now, as per the [ticket](https://github.com/dequelabs/axe-core/issues/3705).
As @straker suggessted, tried to fetch the results for `target-size` using `axe.run({ runOnly: 'target-size' });` but getting,
strange results: 
![Image](https://user-images.githubusercontent.com/47409579/195761876-7da62b48-8904-489e-abfb-fd0e919bd61c.png)

will wait for the dev confirmation.

padmavemulapati:
@straker with the code snippet:

```
<a role="link" aria-label="play" tabindex="0">
    <button tabindex="-1" style="margin:1px;">Play</button>
  </a>
```
I am seeing `inapplicable` results , but it should pass `target-size`
![Image](https://user-images.githubusercontent.com/47409579/196844171-574ffbf6-af0e-426e-b74a-088b3d318bce.png)

padmavemulapati:
Tried with the test snippet:
```
<main>
    <a role=“link” aria-label=“play” tabindex=“0” style=“padding: 5px;“>
      <button tabindex=“-1”>Play</button>
    </a>
  </main>
```
and now `target-size` is passing for the nested interactive controls



![Image](https://user-images.githubusercontent.com/47409579/197193254-86962eba-ac15-4259-8816-d5a1f81591f7.png)

PR #3682, opened 2022-09-27
chore(target-size): Avoid testing nested widgets

WilcoFiers:
Closes issue: #3674

### DIFF
--- a/test/checks/mobile/target-size.js
+++ b/test/checks/mobile/target-size.js
@@ -177,6 +177,17 @@ describe('target-size tests', function () {
           '#obscurer2'
         ]);
       });
+
+      it('returns true if the obscuring widget is a descendant', function () {
+        var checkArgs = checkSetup(
+          `<a role="link" aria-label="play" tabindex="0" style="display:inline-block" id="target">
+            <button tabindex="-1" style="margin:1px; line-height:20px">Play</button>
+          </a>`
+        );
+        const out = check.evaluate.apply(checkContext, checkArgs);
+        console.log(checkContext._data);
+        assert.isTrue(out);
+      });
     });
   });
 
--- a/test/rule-matches/widget-not-inline-matches.js
+++ b/test/rule-matches/widget-not-inline-matches.js
@@ -40,6 +40,14 @@ describe('widget-not-inline-matches', function () {
       assert.isTrue(rule.matches(node, vNode));
     });
 
+    it('returns false for button with tabindex="-1"', function () {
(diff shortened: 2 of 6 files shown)


test/checks/mobile/target-size.js
```diff
@@ -177,6 +177,16 @@ describe('target-size tests', function () {
           '#obscurer2'
         ]);
       });
+
+      it('returns true if the obscuring widget is a descendant', function () {
+        var checkArgs = checkSetup(
+          `<a role="link" aria-label="play" tabindex="0" style="display:inline-block" id="target">
+            <button tabindex="-1" style="margin:1px; line-height:20px">Play</button>
```
straker, line 184: If the button were to not have `tabindex=-1` (so failing the `nested-interactive-role` check), would we still want to fail this rule? Should this rule instead ignore descendant widgets of the widget?

WilcoFiers: That's where I started too, but you can correctly nest controls. Links in particular can contain other controls. I don't think we can ignore those. It could be used in cards for example, where the card is a link, and then inside it there are more controls.

I don't think we can outright ignore nested controls. That'll miss things it shouldn't. There is a potential for false positives on this, I don't know if those are going to happen in practice though. I've only ever seen this nested control pattern with the inner one with tabindex="-1". We could report it as incomplete, but I suspect it's not going to be necessary.

straker: I feel that this is too broad of a solution. This only checks if the element is not tabbable, so non-descendant elements would be ignored as well. I think if we should also check that the element is a descendant of the element it's obscuring before we ignore it.

```html
<div style="position: relative">
  <button>hello</button>
  <button tabindex="-1" style="position: absolute; left: 0; top: 0>world</button>
</div>
```

test/rule-matches/widget-not-inline-matches.js
```diff
@@ -1,124 +1,197 @@
...
+        const node = vNode.actualNode;
+        assert.isTrue(rule.matches(node, vNode));
+      });
+
+      it('is true when the target is not in the tab order', () => {
+        const vNode = queryFixture(`
+          <span role="link" tabindex="-1">
+            Link text
+            <a href="#" id="target" tabindex="-1">Nested link</a>
+          </span>`);
```
straker, line 191: What does this as an integration test produce? Would the child link report a target-size violation but the parent would pass?

straker (approved):

Issue #3606, opened 2022-08-15
aria-roles rule does not support space separated tokens


besenwagen:
### Product

axe-core

### Product Version

4.4.3

### Lastest Version

- [X] I have tested the issue with the latest version of the product

### Issue Description

In the a11y slack somebody recently asked wether Axe failing space separated token values for the `role` attribute is a WCAG or a best practices failure. I assume they used a GUI that doesn't disclose details, axe-core fails it under WCAG SC 4.1.2.

The axe-core **aria-roles** rule
https://github.com/dequelabs/axe-core/blob/develop/doc/rule-descriptions.md#wcag-20-level-a--aa-rules

should implement

ACT rule **Role attribute has valid value**
https://act-rules.github.io/rules/674b10

which clearly describes space separated tokens and has a passing test example for it. 

#### Expectation

The ACT rule's passing test example should not cause a WCAG error in axe-core.

#### Actual

The ACT rule's passing test example causes a WCAG error in axe-core.

#### How to Reproduce

Create a test page that includes the ACT example for a passing test case with space separated tokens in the `role` attribute value and use it as input for axe-core.

besenwagen:
> Auto assign axe-core integration issues to Axe API team project board: All jobs have failed

WilcoFiers:
Axe-core fails this because of issues with Internet Explorer. We're planning to change this fairly soon, since IE is now no longer officially supported.

besenwagen:
Thanks. I don't really understand that, though. It’s categorized as a WCAG SC failure, shouldn’t issues related to browser support be put under best practices? Maybe I have some basic misunderstanding there.

aardrian:
I think this change should also address the _aria-allowed-attr_ rule and maybe the _region_ rule, Right now, the space-separated roles seem to confuse the check for what attributes are allowed.

For example, this passes an axe test:
```
<div aria-label="test" role="dialog">
  Test
</div>
```

As does this:
```
<div aria-label="test" role="region">
  Test
</div>
```

But when I add the space-separated roles as such:
```
<div aria-label="test" role="dialog region">
  Test
</div>
```

It fails:
- [aria-roles](https://dequeuniversity.com/rules/axe/4.4/aria-roles)
- [aria-allowed-attr](https://dequeuniversity.com/rules/axe/4.4/aria-allowed-attr)
- [region](https://dequeuniversity.com/rules/axe/4.4/region) (which I suspect is related)

[...]

PR #3683, opened 2022-09-27
feat(aria-roles): permit fallback roles

WilcoFiers:
Remove the `fallbackrole` from the aria-roles rule. Since IE11 has been deprecated in axe-core 4.5, we're no longer going to fail fallback roles. Because people may still want to test for this we're not going to deprecate the fallbackroles check.

Closes issue: #3606

There are lots of additional changes needed for axe-core to properly deal with fallback roles, such as changing the default of aria.getRole(). Those are out of scope for this PR, and will not be part of axe-core 4.5.


straker (changes requested):
We should also deprecated the `fallbackroles` check while we're doing this

straker (changes requested):
We should add a comment to the fallbackrole check (evaluate?) that we're keeping this around for custom rules and it should not be removed. Otherwise we'll forget why we have a check that is not used by any of our rules.

straker (approved):

Issue #3654, opened 2022-09-08
Unexpected error raised for table marked as presentation


scottaohara:
### Product

axe-core

### Product Version

4.4.2

### Latest Version

- [X] I have tested the issue with the latest version of the product

### Issue Description

The following code snippet was flagged as a serious error with axe for use of the headers attribute.  

```
<table role=presentation>
 <tr><th hidden>...</th><th>...</th></tr>
  <tr>
     <td headers=foo hidden>...</td>
     <td>...</td>
  </tr>
</table>
```

However, this doesn't make any sense to flag as an error because:
1. the table has been marked as role=presentation so table semantics are kicked to the curb anyway.
2. the cell with the erroneous `headers` attribute on it also has the `hidden` attribute, so even if the table was still exposed as a table, the cell is hidden so still can't be an accessibility issue.

padmavemulapati:
Valdiated with the latest axe-core develop branch code base, 
in the test script table has been marked as role=presentation  and td atrribute as hidden so test snippet should not fail for `td-headers-attr` rule
```
<table role=presentation>
 <tr><th hidden>...</th><th>...</th></tr>
  <tr>
     <td headers=foo hidden>...</td>
     <td>...</td>
  </tr>
</table>

```



![Image](https://user-images.githubusercontent.com/47409579/192967378-5e882652-a341-4108-8afa-f8425ee7b3e3.png)

PR #3684, opened 2022-09-27
fix(td-headers-attr): ignore table elements with their role changed

WilcoFiers:
- Only test tables with an implicit or explicit role of table, grid, or treegrid
- Ignore hidden cells with a headers attribute
- Add integration tests

Closes issue: #3654

### DIFF
--- a/lib/checks/tables/td-headers-attr-evaluate.js
+++ b/lib/checks/tables/td-headers-attr-evaluate.js
@@ -1,4 +1,5 @@
 import { tokenList } from '../../core/utils';
+import { isVisibleForScreenreader } from '../../commons/dom';
 
 function tdHeadersAttrEvaluate(node) {
   const cells = [];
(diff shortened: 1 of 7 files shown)


lib/checks/tables/td-headers-attr-evaluate.js
```diff
@@ -24,7 +25,7 @@ function tdHeadersAttrEvaluate(node) {
     let isSelf = false;
     let notOfTable = false;
 
-    if (!cell.hasAttribute('headers')) {
+    if (!cell.hasAttribute('headers') || !isVisibleForScreenreader(cell)) {
```
straker, line 28: This should be it's own pr since it's going to start passing nodes it previously failed but not be in the changelog. That our the pr title needs to be updated to reflect both changes

straker (changes requested):

straker (approved):

Issue #3692, opened 2022-10-02
scrollable-region-focusable: false positive for disabled multiline select / listbox


MichaelKetting:
### Product

axe-core

### Product Version

3.3.4 ... 4.4.2

### Latest Version

- [X] I have tested the issue with the latest version of the product

#### Expectation
scrollable-region-focusable should not report for multiline select elements / listbox (when they are diabled)

#### Actual
I have a disabled select element showing a scrollbar because it contains more items than than it has rows configured and I get scrollable-region-focusable in my aXe result.

#### How to Reproduce
```
<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml" lang="en">
<head>
<title>Example</title>
<meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
</head>
<body>
  <select size="2" name="select" disabled="disabled" aria-label="Test">
    <option value="First">First</option>
    <option value="Second">Second</option>
    <option value="Third">Third</option>
  </select>
</body>
</html>

```

#### Additional context
Since the browser does not allow focusing the element, it is not within the web-developer's scope to solve this issue for disabled elements.

see also #3691

padmavemulapati:
Validated with the latest axe-core develop branch code base,
The `Scrollable-region-focussible` rule is not failing when `select` `size` makes scrollable for the list of items.
Test snippet:
```
<select size="2" name="select" disabled="disabled" aria-label="Test">
    <option value="First">First</option>
    <option value="Second">Second</option>
    <option value="Third">Third</option>
  </select>
```



![Image](https://user-images.githubusercontent.com/47409579/214528224-5b412be0-6393-4c43-a8f6-222918fde818.png)

Issue #3691, opened 2022-10-02
scrollable-region-focusable: false positive for disabled textarea


MichaelKetting:
### Product

axe-core

### Product Version

3.3.4 ... 4.4.2

### Latest Version

- [X] I have tested the issue with the latest version of the product

#### Expectation
scrollable-region-focusable should not report for textareas (when they are diabled)

#### Actual
I have a disabled textarea element showing a scrollbar because it contains more text than it has vertical space and I get scrollable-region-focusable in my aXe result

#### How to Reproduce
```
<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml" lang="en">
<head>
<title>Example</title>
<meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
</head>
<body>
  <textarea name="textarea" rows="2" cols="20"  disabled="disabled" aria-label="Test">
      test
      test
      test
  </textarea>
</body>
</html>
```

#### Additional context
Since the browser does not allow focusing the element, it is not within the web-developer's scope to solve this issue for disabled elements.

see also #3692

straker:
Thanks for the issues. I believe you are correct and we should ignore disabled controls for this rule.

PR #3701, opened 2022-10-06
fix(scrollable-region-focusable): report inapplicable if scrollable regions are disabled

Zidious:
This PR checks if the scrollable content has the `disabled` attribute, if so we can ignore it as it is not focusable by the keyboard

Closes issue: https://github.com/dequelabs/axe-core/issues/3692, https://github.com/dequelabs/axe-core/issues/3691


lib/rules/scrollable-region-focusable-matches.js
```diff
@@ -54,6 +55,14 @@ function scrollableRegionFocusableMatches(node, virtualNode) {
     }
   }
 
+  /**
+   * if scrollable content is disabled (not focusable)
+   * we can return false
+   */
+  if (attrNames.includes('disabled')) {
```
straker, line 62: We should use `commons.forms.isDisabled`. It'll cover things like `disabled` being inherited from a parent (such as `fieldset`), and using `aria-disabled`

test/rule-matches/scrollable-region-focusable-matches.js
```diff
@@ -1,24 +1,60 @@
...
+        '<option value="First">First</option>' +
+        '<option value="Second">Second</option>' +
+        '<option value="Third">Third</option>' +
+        '</select>'
+    );
+    var actual = rule.matches(target.actualNode, target);
+    assert.isFalse(actual);
+  });
+
+  it('returns false when the scrollable content is disabled="true" and not focusable', function () {
```
straker, line 33: Should add a test for parent fieldset having `disabled` as well as `aria-disabled`

straker (changes requested):

test/rule-matches/scrollable-region-focusable-matches.js
Zidious: If the parent element is disabled does that not mean all children are disabled by default?

Zidious: edit:  added test for this scenario: 

```html
  <fieldset disabled>
    <legend>Do you like receiving wrenches?</legend>
    <input
      type="checkbox"
      id="chbx"
      name="agree"
      value="Yes!"
      aria-disabled="false"
    />
    <label for="chbx">Well, yes of course!</label>
  </fieldset>
  ```

**Comment by Zidious** [comment]
After looking into this for [firefox CI failure](https://app.circleci.com/pipelines/github/dequelabs/axe-core/4339/workflows/632d1756-f16a-4373-b68b-052bc844b17d/jobs/49833?invite=true#step-109-9244). 

With the following snippet: 

```html
<textarea id="pass8" name="textarea" aria-label="test">
  hazaar
  hazaar
  hazaar
</textarea>
```

Within `scrollable-region-focusable` we call `getScroll()` which calls this function [isScrollable()](https://github.com/dequelabs/axe-core/blob/develop/lib/core/utils/get-scroll.js#L37-L40). 

Here we check the `scrollableX` and `scrollableY` values have the `scroll` or `overflow` prop;

With Chrome we get: 

```
{overflowProp: 'auto'}
```

Interestingly, with Firefox we get: 

```
{ overflowProp: "visible" }
```

I was not sure if adding `visible` to the `overflowProp` had some possible negative side effects - possibly requires more discussion 👀

Issue #3626, opened 2022-08-31
axe is struggling to determinate the background color when using nested lists


Zauberbutter:
### Product

axe-core

### Product Version

4.4.3

### Latest Version

- [X] I have tested the issue with the latest version of the product

### Issue Description

I've found the following issue with the determination of background colors by axe. The source code contains an example where it fails and one where it succeeds. 

**How to reproduce**:
<details>
<summary>HTML</summary>

```html

<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>Test</title>
    <style>
    *, ::after, ::before {
        -webkit-box-sizing: border-box;
        -moz-box-sizing: border-box;
        box-sizing: border-box;
    }
    body {
	    font-family: Arial,Helvetica,sans-serif;
        font-size: 14px;
        line-height: 1.6em;
    }
    </style>
</head>
<body>

<ul>
    <li>
        Test
        <ul>
            <li></li>
        </ul>
    </li>

    <!-- when you remove the dot at the end it fails too -->
    <li>Lorem <strong>ipsum</strong>. 
        <ul>
    	    <li></li>
        </ul>
    </li>
</ul>

</body>
</html>

```

</details>

**Product**: axe-core
**Expectation**: It should be able to determine the background color.
**Actual**: It can't.

<pre><code>axe-core version: 4.4.3
DevTools: 4.36.2</code></pre>

straker:
Thanks for the  issue. It looks like we'll need to figure out how to handle this type of code as we currently think the nested `li` is overlapping the parent `li`.  Even `document.elementsFromPoint` reports it the same way.

padmavemulapati:
Validated with the latest code base of axe-core develop branch, 
when using nested list elements with background color, it is taking care of `color-contrast` things not `list` which is expected.
When Nested list elements are presented, issue saying that, error is getting for "Text elements must have sufficient color contrast."
Not seeing that error now, but seeing incomplete results for color-contrast saying
"Element's background color could not be determined because it is overlapped by another element"
<img width="1298" alt="image" src="https://user-images.githubusercontent.com/47409579/206959946-aa4aba8d-5868-4f32-8b60-ca08d2b2734a.png">

Issue #3442, opened 2022-04-26
Colour contrast checking ignores content being clipped by overflow:hidden


Shepard:
Product: axe-core and axe Extension

Fiddle for reproducing this: https://jsfiddle.net/Loe7pacx/

Expectation: Limit contrast checking of content that is cut off by overflow:hidden only to the visible part. No issue is reported by axe as this is black text on a white background. The black background of the outer element (body) should not matter.

Actual: Axe reports an issue on the span element. Contrast is checked as if the content visibly extended into an area with a different background colour.

Motivation: This was not reported in previous versions and visually there is no contrast issue for the user.

<pre><code>
axe-core version: 4.4.1
extension version: 4.25.3

Browser and Assistive Technology versions
Chrome 100

For Tooling issues:
- Platform:  Windows
</code></pre>

straker:
Thanks for the issue. Confirmed that this is an issue with our stacking algorithm as `.wrapper` and  `body` are missing from the final stack calculation.

```js
axe.setup()
axe.commons.dom.getElementStack(document.querySelector('span'))
// [span, html]
```

Shepard:
The fiddle is a very simplified version of the code where I was experiencing this issue. In my code, the darker background colour is set on the html element and there's some divs inbetween them.
I wonder if this would make a difference when fixing this bug?

straker:
Some more info on this ticket. This isn't limited to `overflow`, this same issue can be caused by the following DOM:

```html
<div style="background-color: #222; height: 40px; width: 100px"></div>
<div style="margin-top: -40px;">Hello world</div>
```

Because our code only really looks at the bounding box size of the element, it doesn't pick up the smaller element behind the text contributing to the background color. In this case no issue is reported even though there is one.

padmavemulapati:
Validated with the latest axe-core develop branch code base,
Not seeing any failure results for `color-contrast` either in `violations` or in `incomplete` when `overflow:hidden`



![Image](https://user-images.githubusercontent.com/47409579/206957885-424d22d5-6fca-4989-ac67-4baeb6edbd12.png)

Issue #3400, opened 2022-02-28
Contrast-error on list with more than one level


kjetilhansen0:
List with more than one level generates error: "Text elements must have sufficient color contrast."

**How to reproduce:**
List with two levels or more. No CSS needed.

```
<ul>
  <li>Lorem
    <ul>
      <li>Ipsum</li>
    </ul>
  </li>
</ul>
```
Plugin says: "Element's background color could not be determined because it is overlapped by another element."
Tested with Axe-core and browser plugin. Chrome on MacOS.

straker:
Thanks for the issue. We'll see if we can get this fixed for the next release.

padmavemulapati:
Validated with the latest code base of axe-core develop branch,

with the code snippet:
```
<ul>
  <li>Lorem
    <ul>
      <li>Ipsum</li>
    </ul>
  </li>
</ul>
```
where nested `list` elements are presented, issue saying that, error is getting for "Text elements must have sufficient color contrast." 
Not seeing that error now, but seeing incomplete results for `color-contrast` saying 
"Element's background color could not be determined because it is overlapped by another element"
![image](https://user-images.githubusercontent.com/47409579/206958978-533afee0-3e78-42ba-89d4-164bef8e6caf.png)

Issue #3257, opened 2021-11-05
Color contrast incomplete for label and input


straker:
This returns incomplete for the `label` element because we try to get the midpoint of the label, which is way outside the text content.

```html
<label>Do you like peas?
  <input name="peas" style="width: 100%">
</label>
```

padmavemulapati:
Validated with the latest code base of axe-core develop branch,
with the code snippet:
```
<label>Do you like peas?
  <input name="peas" style="width: 100%">
</label>
```
`label` element is getting `incomplete` results 
<img width="1306" alt="image" src="https://user-images.githubusercontent.com/47409579/206959286-87efd879-ccf4-49b6-8b37-702da0faabeb.png">

PR #3703, opened 2022-10-07
fix(color-contrast): correctly calculate background color of text nodes with different size than their container

straker:
Had an idea on how to improve the accuracy of our color contrast algorithm by looking at just the text rects (instead of trying to look at the bounding box of the element). This not only simplifies the algorithm a bit by removing a few in-between functions, but also fixes incorrect calculations of background color, such as in the following example which would report no contrast issues even though the text is fully on top of a dark background:

```html
<div id="background-1" style="background-color: #222; height: 40px; width: 100px"></div>
<div id="foreground-div" target style="margin-top: -40px;">Hello world<br/>Goodbye Fate</div>
```

Also fixes a lot of different cases reporting partial overlap when they didn't need to.

This also deprecates `visually-contains`, `filteredRectStack`, and `getRectStack`, but I'll deprecate them in a separate pr.

Closes issue: https://github.com/dequelabs/axe-core/issues/3626
Closes issue: https://github.com/dequelabs/axe-core/issues/3442
Closes issue: https://github.com/dequelabs/axe-core/issues/3400
Closes issue: https://github.com/dequelabs/axe-core/issues/3257


WilcoFiers (changes requested):

lib/commons/dom/get-visible-text-rects.js
```diff
@@ -0,0 +1,84 @@
...
+      const verticalMidpoint = rect.top + rect.height / 2;
+
+      return (
+        horizontalMidpoint < nodeRect.left ||
+        horizontalMidpoint > nodeRect.right ||
+        verticalMidpoint < nodeRect.top ||
+        verticalMidpoint > nodeRect.bottom
+      );
+    });
+    if (outsideRectBounds) {
```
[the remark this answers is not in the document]
stephenmathieson: Just curious here...

The logic is in a separate function. Do you mean hoisting the `rects.some` callback out? It doesn't look reusable, so I'm wondering what the benefit is. Perhaps just decreasing the indent level?

lib/commons/color/get-background-color.js
```diff
@@ -53,40 +53,46 @@ function _getBackgroundColor(elm, bgElms, shadowOutlineEmMax) {
...
+    }
+
     // Get the background color
     const bgColor = getOwnBackgroundColor(bgElmStyle);
+    if (bgColor.alpha === 0) {
+      return;
+    }
 
     if (
       // abort if a node is partially obscured and obscuring element has a background
```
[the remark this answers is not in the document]
stephenmathieson: Agreed. The comment is describing the `if`, so it should be the line immediately above it.

lib/commons/dom/get-visible-text-rects.js
```diff
@@ -0,0 +1,84 @@
...
+import getOverflowHiddenAncestors from './get-overflow-hidden-ancestors';
+
+/**
+ * Get the visible text client rects of a node.
+ * @method getVisibleTextRects
+ * @memberof axe.commons.dom
+ * @instance
+ * @param {Element} node
+ */
+export default function getVisibleTextRects(node) {
```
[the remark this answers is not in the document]
straker: The problem is I need to use this function in `getBackgroundColor` as well to determine when an overlapping node visually contains all the visible text of the target node. And it has to account for `overflow` of those text nodes. So it's written to pass a single node and return a list of rects for each of the text children.

lib/commons/color/get-background-stack.js
```diff
@@ -95,31 +84,24 @@ function sortPageBackground(elmStack) {
...
- *
- * @method getBackgroundStack
- * @memberof axe.commons.color
- * @param {Element} elm
- * @return {Array}
+ * Check to see if two arrays are equal
+ * @see https://stackoverflow.com/a/16436975/2124254
  */
-function getBackgroundStack(elm) {
-  let elmStack = filteredRectStack(elm);
```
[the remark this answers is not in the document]
straker: Interesting. That doesn't seem to have any tests associated with it. I'll add some in.

lib/commons/color/get-background-color.js
```diff
@@ -124,20 +130,41 @@ function _getBackgroundColor(elm, bgElms, shadowOutlineEmMax) {
...
-    incompleteData.set('bgColor', 'elmPartiallyObscured');
+function fullyEncompasses(node, rects) {
+  rects = Array.isArray(rects) ? rects : [rects];
+
+  const nodeRect = node.getBoundingClientRect();
+  const style = window.getComputedStyle(node);
+  const overflow = style.getPropertyValue('overflow');
+
+  if (style.getPropertyValue('display') === 'inline') {
+    return true;
```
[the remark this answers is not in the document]
straker: This function was a modified copy of [visuallyContains](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/dom/visually-contains.js#L67), not sure why it's there either. Where would it be better placed?

WilcoFiers (dismissed):

lib/commons/color/get-background-color.js
```diff
@@ -53,40 +53,47 @@ function _getBackgroundColor(elm, bgElms, shadowOutlineEmMax) {
   }
 
   const elmStack = getBackgroundStack(elm);
+  const textRects = getVisibleChildTextRects(elm);
```
WilcoFiers, line 56: I'm not sure about this. We're now calling getVisibleChildTextRects twice. Once here, and once inside `getBackgroundStack > getTextElementStack`. What if we passed these rects into getBackgroundStack instead so we don't need to make the call twice?

```suggestion
  const textRects = getVisibleChildTextRects(elm);
  const elmStack = getBackgroundStack(elm, textRects);
```

lib/commons/color/get-background-color.js
```diff
@@ -124,20 +131,37 @@ function _getBackgroundColor(elm, bgElms, shadowOutlineEmMax) {
...
-  var obscured =
-    elm !== bgElm && !visuallyContains(elm, bgElm) && bgColor.alpha !== 0;
-  if (obscured) {
-    incompleteData.set('bgColor', 'elmPartiallyObscured');
+function fullyEncompasses(node, rects) {
+  rects = Array.isArray(rects) ? rects : [rects];
+
+  const nodeRect = node.getBoundingClientRect();
+  const style = window.getComputedStyle(node);
+  const overflow = style.getPropertyValue('overflow');
```
WilcoFiers, line 145: Should this handle overflow-x and overflow-y separately?

lib/commons/color/get-background-color.js
WilcoFiers: Alternatively, memoize for getVisibleChildTextRects could work too.

lib/commons/color/get-background-stack.js
```diff
@@ -95,31 +84,24 @@ function sortPageBackground(elmStack) {
...
-
-  if (elmStack === null) {
-    return null;
+function arraysEqual(a, b) {
```
WilcoFiers, line 90: How about: 
```suggestion
function shallowArraysEqual(a, b) {
```

lib/commons/dom/get-text-element-stack.js
```diff
@@ -22,51 +21,7 @@ function getTextElementStack(node) {
...
-        if (rect.width >= 1 && rect.height >= 1) {
-          clientRects.push(rect);
-        }
-      }
-    }
-  });
-
+  const clientRects = getVisibleChildTextRects(node);
   if (!clientRects.length) {
     return [getElementStack(node)];
```
WilcoFiers, line 26: I got a little confused at this point. If there's no text visible text, why doesn't this go into incomplete straight away, or return an empty array?

WilcoFiers (changes requested):

lib/commons/dom/get-text-element-stack.js
straker: The main reason is that it will handle cases where the text node is larger than the bounds of the element. That way we will use the node bounds for our calculations rather than the next node bounds. (All [the issues listed here](https://github.com/dequelabs/axe-core/pull/3703/files#diff-a443be069bc48f7824cb773157079a125dd30cd512172620ad750df4817199dbR60-R71))

```js
<div width="10px">Hello world</div>
```

lib/commons/color/get-background-color.js
straker: For what reason?

WilcoFiers: Sorry, never mind. I didn't know this, but it seems you did. So if I set overflow-x to `auto`, then overflow-y is also `auto`, even if it was set to something else. Weird.

lib/commons/dom/get-visible-child-text-rects.js
```diff
@@ -0,0 +1,80 @@
...
+ * @param {Element} node
+ */
+const getVisibleChildTextRects = memoize(
+  function getVisibleChildTextRectsMemoized(node) {
+    const vNode = getNodeFromTree(node);
+    const nodeRect = vNode.boundingClientRect;
+    const clientRects = [];
+    const overflowHiddenNodes = getOverflowHiddenAncestors(vNode);
+
+    Array.from(node.childNodes).forEach(elm => {
```
WilcoFiers, line 20: Nitpick; this doesn't need Array.from. NodeList has a forEach method. Or you could use `for ... of`.

lib/commons/dom/get-visible-child-text-rects.js
```diff
@@ -0,0 +1,80 @@
...
+      if (elm.nodeType !== 3 || sanitize(elm.textContent) === '') {
+        return;
+      }
+
+      const contentRects = getContentRects(elm);
+      if (isOutsideNodeBounds(contentRects, nodeRect)) {
+        return;
+      }
+
+      contentRects.forEach(contentRect => {
```
WilcoFiers, line 30: I keep tripping up in this part of the function trying to work out what this does. I think moving this forEach into a new function would help a lot with that.

```suggestion
      const visibleRects = removeHiddenRects(contentRects, overflowHiddenNodes)
      clientRects.push(...visibleRects);
```

lib/commons/dom/get-text-element-stack.js
WilcoFiers: I see. It's not intuitive to me that clientRects would be empty if the child texts are larger than their parent. It seems to me like maybe we should then deal with that. That's still text that can be on an element with bad contrast. For the sake of not introducing further scope creep I'd be okay with punting that to an issue.


I think if we don't want to handle child texts that are larger than their parent in this PR, then maybe `getVisibleChildTextRects` should return the bounding box of the parent, so that it can also be used here:
https://github.com/dequelabs/axe-core/pull/3703/files#diff-eaabb887841937caf303235ad7ee968907755aa5ff002929ab03096cf848b2e5R78

If I read this right, if the text is larger than the element, it will never get reported as `elmPartiallyObscured`. Since if you pass an empty `clientRect` into `fullyEncompasses` it always returns `true`. If `clientRect` had the bounding box of the element that wouldn't be a problem.

lib/commons/dom/get-visible-child-text-rects.js
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * Get the visible text client rects of a node.
+ * @method getVisibleChildTextRects
+ * @memberof axe.commons.dom
+ * @instance
+ * @param {Element} node
+ */
+const getVisibleChildTextRects = memoize(
+  function getVisibleChildTextRectsMemoized(node) {
```
WilcoFiers, line 14: Should this method handle CSS clip, in addition to overflow:hidden? If so I think a separate issue might be in order.

WilcoFiers (dismissed):

lib/commons/dom/get-visible-child-text-rects.js
straker: Possibly. We don't have any tests around this so a separate ticket would be good.

lib/commons/dom/get-text-element-stack.js
straker: Great suggestion

lib/commons/dom/get-overflow-hidden-ancestors.js
```diff
@@ -0,0 +1,28 @@
...
+  function getOverflowHiddenAncestorsMemoized(vNode) {
+    const ancestors = [];
+
+    if (!vNode) {
+      return ancestors;
+    }
+
+    const overflow = vNode.getComputedStylePropertyValue('overflow');
+
+    if (overflow === 'hidden') {
```
WilcoFiers, line 20: Should have a ticket for overflow-x: hidden overflow-y: visible.

straker: https://github.com/dequelabs/axe-core/issues/3818

lib/commons/dom/get-visible-child-text-rects.js
WilcoFiers: Clip is deprecated, skipping.

WilcoFiers (approved):

WilcoFiers (approved):

Issue #3559, opened 2022-07-22
Typo in axe extension for Rule ID: list


KB55:
Product: axe Extension
Expectation: Correctly describe how to fix the issue
Actual: Typo in the text 
Motivation: The typo has caused confusion and caused developers to look in the wrong place for the error

Extension version: v4.32.0
axe-core version: v4.4.2
Chrome browser version 103.0.5060.134 (Official Build) (64-bit)

In the axe extension an error has been identified for:

> Rule ID: list 
> Ruleset: axe-core 4.4
> `<ul>` and `<ol>` must only directly contain `<li>`, `<script>` or `<template>` elements

The issue was triggered because a `<span>` was nested within a `<ul>`. The extension says: 

> Fix the following: List element has direct children that are not allowed inside `<li>` elements

I believe this is a typo and that it should say:
> List element has direct children that are not allowed inside `<ul>` elements

The rule is concerned with errors due to incorrect direct children of `<ul>` and `<ol>` elements, not `<li>` elements and therefore this fix shown in the extension is misleading. 

Screenshots attached:

- Axe issue
- Axe issue description
- Code for issue

![AxeExtensionScreenshot](https://user-images.githubusercontent.com/26979111/180433775-90b6c897-b9ed-49c2-bbc8-11a4e372c013.png)
![AxeIssueDescScreenshot](https://user-images.githubusercontent.com/26979111/180433803-9de06360-e0e9-415e-9efb-a2d455c2423b.png)
![AxeCodeExample](https://user-images.githubusercontent.com/26979111/180433814-b5eb829b-1fe6-483a-a608-f6244c3303a6.png)

straker:
Thanks for the issue. I can see where the confusion is. I'll talk to @WilcoFiers about it and we'll see what we can do.

WilcoFiers:
Agreed, this is wrong. `only-listitems` and `only-dlitems` need to be updated. Thank you for raising this.

padmavemulapati:
@straker , I am still seeing the fix description as 
" List element has direct children that are not allowed inside <li> elements" 
for the script:
```
<ul id="scriptproperul">
    <script></script>
    <template></template>
    <li>One.</li>
    <li>Two.</li>
  </ul>
  <ul class="sample"></ul>
  <ol id="emptyol"></ol>
  <ol id="scriptemptyol">
```
[...]

PR #3707, opened 2022-10-08
feat(list/definition-list): only allow required owned roles

WilcoFiers:
Improve on messages reported by the `list` and `definition-list` checks. Since they had so much in common I unified them into a new `invalid-children-evaluate` method. That got me 95% of the way to running their rules on virtual nodes, so I then also updated the matches method to run on virtual. 

This rule gets rid of the (minor) exception on `list`, where if you have `li` elements, where some have a role and others do not, the rule will pass. I don't recall why we put that in, but it doesn't make too much sense to me.

The `only-dlitems-evaluate` and `only-listitems-evaluate` methods need to be deprecated, but this belongs in a separate PR.

Closes issue: #3559

### DIFF
--- a/lib/checks/lists/invalid-children-evaluate.js
+++ b/lib/checks/lists/invalid-children-evaluate.js
@@ -1,0 +1,75 @@
+import { isVisibleToScreenReaders } from '../../commons/dom';
+import { getExplicitRole } from '../../commons/aria';
+
+export default function invalidChildrenEvaluate(
+  node,
+  options = {},
+  virtualNode
+) {
+  const relatedNodes = [];
+  const issues = [];
+  if (!virtualNode.children) {
+    return undefined;
+  }
+
+  const vChildren = mapWithNested(virtualNode.children);
+  while (vChildren.length) {
+    const { vChild, nested } = vChildren.shift();
+    if (options.divGroups && !nested && isDivGroup(vChild)) {
+      if (!vChild.children) {
+        return undefined;
+      }
+      const vGrandChildren = mapWithNested(vChild.children, true);
+      vChildren.push(...vGrandChildren);
+      continue;
+    }
+
+    const issue = isInvalidChild(vChild, nested, options);
+    if (!issue) {
+      continue;
+    }
+    if (!issues.includes(issue)) {
+      issues.push(issue);
+    }
+    if (vChild?.actualNode?.nodeType === 1) {
+      relatedNodes.push(vChild.actualNode);
+    }
+  }
+  if (issues.length === 0) {
+    return false;
+  }
+
+  this.data({ values: issues.join(', ') });
+  this.relatedNodes(relatedNodes);
+  return true;
+}
+
+function isInvalidChild(
+  vChild,
+  nested,
+  { validRoles = [], validNodeNames = [] }
+) {
+  const { nodeName, nodeType, nodeValue } = vChild.props;
+  const selector = nested ? 'div > ' : '';
+  if (nodeType === 3 && nodeValue.trim() !== '') {
+    return selector + `#text`;
+  }
+  if (nodeType !== 1 || !isVisibleToScreenReaders(vChild)) {
+    return false;
+  }
+
+  const role = getExplicitRole(vChild);
+  if (role) {
+    return validRoles.includes(role) ? false : selector + `[role=${role}]`;
+  } else {
+    return validNodeNames.includes(nodeName) ? false : selector + nodeName;
+  }
+}
+
+function isDivGroup(vNode) {
+  return vNode.props.nodeName === 'div' && getExplicitRole(vNode) === null;
+}
+
+function mapWithNested(vNodes, nested = false) {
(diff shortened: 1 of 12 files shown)


lib/checks/lists/invalid-children-evaluate.js
```diff
@@ -0,0 +1,75 @@
...
+  if (issues.length === 0) {
+    return false;
+  }
+
+  this.data({ values: issues.join(', ') });
+  this.relatedNodes(relatedNodes);
+  return true;
+}
+
+function isInvalidChild(
```
straker, line 47: I was pretty confused on what this function was doing because the name of the function indicated it would return a boolean, but then it returned mixed boolean and strings. I think the function should just return strings (empty strings for false) and the name should be changed to reflect it's not a boolean return function.

lib/checks/lists/invalid-children-evaluate.js
```diff
@@ -0,0 +1,75 @@
...
+  const { nodeName, nodeType, nodeValue } = vChild.props;
+  const selector = nested ? 'div > ' : '';
+  if (nodeType === 3 && nodeValue.trim() !== '') {
+    return selector + `#text`;
+  }
+  if (nodeType !== 1 || !isVisibleToScreenReaders(vChild)) {
+    return false;
+  }
+
+  const role = getExplicitRole(vChild);
```
straker, line 61: We should we use `getRole` instead to account for role conflict. Something like this would fail even though each is treated as a listitem (this currently does not fail today)

```html
<ul>
  <li role="none" aria-label="hello"></li>
  <li role="none" tabindex="0">Hello</li>
</ul>
```

test/checks/lists/only-listitems.js
```diff
@@ -1,314 +1,248 @@
...
       '<ol id="target"><link rel="stylesheet" href="theme.css"><li role="menuitem">Not a list item</li></ol>'
     );
 
-    assert.isTrue(
-      axe.testUtils
-        .getCheckEvaluate('only-listitems')
-        .apply(checkContext, checkArgs)
-    );
+    assert.isTrue(checkEvaluate.apply(checkContext, checkArgs));
+    console.log(checkContext._data);
```
straker, line 125:
```suggestion
```

test/integration/virtual-rules/definition-list.js
```diff
@@ -0,0 +1,53 @@
...
+  it('fails when there no dd / dt pair', () => {
+    const dl = new axe.SerialVirtualNode({ nodeName: 'dl' });
+    const dd1 = new axe.SerialVirtualNode({ nodeName: 'dd' });
+    const dd2 = new axe.SerialVirtualNode({ nodeName: 'dd' });
+    dl.children = [dd1, dd2];
+    dd1.parent = dl;
+    dd2.parent = dl;
+
+    const results = axe.runVirtualRule('definition-list', dl);
+    console.log(JSON.stringify(results.violations, null, 2));
```
straker, line 25:
```suggestion
```

test/integration/virtual-rules/definition-list.js
```diff
@@ -0,0 +1,53 @@
+describe('definition-list virtual-rule', () => {
+  it('passes when there is an invalid child nodes', () => {
```
straker, line 2:
```suggestion
  it('passes when there are no invalid child nodes', () => {
```

test/integration/virtual-rules/list.js
```diff
@@ -0,0 +1,36 @@
+describe('list virtual-rule', () => {
+  it('passes when there is an invalid child nodes', () => {
```
straker, line 2:
```suggestion
  it('passes when there are no invalid child nodes', () => {
```

test/integration/virtual-rules/list.js
```diff
@@ -0,0 +1,36 @@
...
+    ul.children = [li];
+    li.parent = ul;
+
+    const results = axe.runVirtualRule('list', ul);
+    assert.lengthOf(results.passes, 1);
+    assert.lengthOf(results.violations, 0);
+    assert.lengthOf(results.incomplete, 0);
+  });
+
+  it('fails when there is an invalid child nodes', () => {
```
straker, line 14:
```suggestion
  it('fails when there is an invalid child node', () => {
```

straker (dismissed):

lib/checks/lists/invalid-children-evaluate.js
WilcoFiers: I don't think people should rely on fallback roles to pass. We're failing this after all. The one thing I do think is worth considering is to allow role=none on empty elements. We don't allow that currently, so I think that's a separate PR.

One thing I'd like your input on is in how this rule is becoming stricter. Something like the following was passing before. I don't think it should have been, but I'm not sure if we need to more explicitly call it out in the PR title, or even create a separate PR so its separated in the changelog. WDYT?

```html
<!-- This passes today, because it's an li and we then don't look at the role -->
<ul><li role="button">Button!</li></ul>
```

straker (approved):

PR #3709, opened 2022-10-09
feat: Deprecate only-dlitems-evaluate & only-listitems-evaluate methods

WilcoFiers:
As per #3707, these evaluate methods are no longer in use.


straker (approved):

WilcoFiers:
Close because of conflicts, creating a new PR.

Issue #3672, opened 2022-09-24
target-size rule should handle fixed position elements


WilcoFiers:
An issue I'm spotting with the new target-size rule is that it doesn't properly account for fixed position elements. An element with sufficient size could be failed if it is partially obscured by another focusable element with a fixed position. 

I spotted an example of this on the axe-con website:
https://www.deque.com/axe-con/sessions/changing-the-way-you-think-and-work-agile-accessibility-audits-in-the-product-life-cycle/

![Axe-con's wide screen button partially obscured by the profile link in the fixed position top bar](https://user-images.githubusercontent.com/530687/192096480-e3cc8928-ba35-4f39-99ef-6e9638177986.png)

I think the fix for this will be to create a separate grid for all fixed position elements, so that fixed elements aren't considered to be adjacent to scrollable elements. We'd also need to consider what sticky elements do in this. I'm less sure of what to do with those, since they sort of live in two worlds. They are scrollable, but become fixed to their parent container once the scrollTop is larger then their boundingClientRect.top,

padmavemulapati:
Validated with the latest develop branch code base, 
the test script:
```
<p style="position: fixed">
      <a href="#" id="pass-fixed">Wide link</a>
      <a href="#" id="fail-fixed">x</a>
    </p>
```
`target-size` rule failing for the fixed element when `<a href="#" id="fail-fixed">x</a>` 
and 
passing `<a href="#" id="pass-fixed">Wide link</a>`



![Image](https://user-images.githubusercontent.com/47409579/196845323-6ed7a7f3-a300-4b2b-bd7e-8366b104689c.png)

PR #3711, opened 2022-10-09
chore(target-size): consider fixed and non-fixed position nodes separately

WilcoFiers:
Have the target-size rule test fixed position elements separate from those not that are not fixed.

Closes issue: #3672


lib/commons/dom/find-nearby-elms.js
```diff
@@ -37,3 +45,13 @@ function loopGridCells(gridCells, boundaries, cb) {
     }
   }
 }
+
+const hasFixedPosition = memoize(vNode => {
+  if (!vNode) {
+    return false;
+  }
+  if (vNode.getComputedStylePropertyValue('position') === 'fixed') {
```
straker, line 53: We should also consider `sticky` position 

```suggestion
  if (['fixed', 'sticky'].includes(vNode.getComputedStylePropertyValue('position')) {
```

straker (changes requested):

lib/commons/dom/find-nearby-elms.js
WilcoFiers: I don't think this works for sticky elements. This is already a fairly naive approach to fixed elements... It's sort of just hoping that fixed elements are never in a subgrid, and that within a fixed position element there isn't a scrollable region.

But for sticky this is not the way to go. Sticky element's aren't like fixed position elements. They are sticky to their container, not fixed to the viewport. Sticky elements are much more like position:relative then they are like position:fixed. While not optimal, I think treating them like relative elements is better than treating them like sticky elements.

Ideally the solve for sticky would be for the grid to treat them as though they are on their original position, rather than take their current bounding box as the correct placement for them. That's out of scope for this PR, and for the 4.5 release I think too. I think we need to revisit the grid for 4.6 to address these problems, but we're too late at this stage.

straker: > Ideally the solve for sticky would be for the grid to treat them as though they are on their original position, rather than take their current bounding box as the correct placement for them

I don't think that's possible. The grid positions the elements where they are at the time it is created. It has no way of telling if the elements current position is not its original position before scrolling occurred. Once the sticky elements scroll boundary is passed, a sticky element is no different than a fixed element in terms of the bounding box and position. The problem is there's no way to tell where the elements scroll boundary is unless the grid were to take scrolling into account, which it doesn't as it treats the entire page as one long document.

I don't mind saying we don't want to add code to deal with sticky elements into this fix, so long as we understand that the fix for sticky elements is probably not going to be in the grid itself.

Issue #3612, opened 2022-08-17
Avoid duplicates between color-contrast and color-contrast-enhanced


WilcoFiers:
### Product

axe-core

### Feature Description

Currently the color-contrast-enhanced rule fails nodes that also fail the color-contrast rule. This creates duplicates when both rules are on. This isn't ideal. A better solution would be to pass nodes that are below the contrast threshold of color-contrast.

There are already options available in color-contrast-evaluate to do this. We should add caching to this evaluate method to avoid calculating the contrast of a node more than once though, since this is a fairly slow operation.

WilcoFiers:
Can probably memoize `getBackgroundColor`, that'll be the quickest way to cache this.

padmavemulapati:
Validated with the latest develop branch code base, 
when run both `color-contrast-enhanced` and `color-contrast` , now it is not resulting both incomplete results instead it is giving only `color-contrast` incomplete result and avoiding duplicate

Test snippet:
```
<p id="target" style="color: #666; background: linear-gradient(to right, #FFF, #0FF); width: 300px">
    Some text in English
  </p>
  <p id="target" style="color: #666; background: linear-gradient(to right, #FFF, #0FF); width: 300px; font-size: 18pt;">
    Some text in English
  </p>
  
```




![Image](https://user-images.githubusercontent.com/47409579/196725090-de60f60f-f99f-450e-bc38-ee70c2875471.png)

PR #3714, opened 2022-10-11
fix(color-contrast-enhanced): avoid duplicates between color-contrast and color-contrast-enhanced

straker:
Closes issue: https://github.com/dequelabs/axe-core/issues/3612


lib/commons/color/get-background-color.js
```diff
@@ -25,6 +26,34 @@ export default function getBackgroundColor(
...
+
+  const bgColor = _getBackgroundColor(elm, bgElms, shadowOutlineEmMax);
+
+  // TODO: this is a temporary solution to the problem
+  // of needing to cache the getBackgroundColor call
+  // but not being able to use `memoize` since this
+  // is not a pure function due to the `bgElms` array
+  // and `incompleteData`.
+  // remove after 4.5 release when we have time to
+  // properly update the function to be more cacheable
```
WilcoFiers, line 46: Please put this in an issue instead.

test/checks/color/color-contrast.js
```diff
@@ -723,6 +723,25 @@ describe('color-contrast', function () {
       assert.deepEqual(checkContext._relatedNodes, []);
     });
 
+    it('should report incomplete for options.contrastRatio.normal.minThreshold', function () {
```
WilcoFiers, line 726: I'm confused about what these test titles mean / what this test is supposed to prove. Can you clarify?

WilcoFiers (changes requested):

test/checks/color/color-contrast.js
straker: I was just copying the tests we already had (the prior test to each of these) and modifying it to return incomplete to prove that it reports incomplete since the threshold would have prevent that from happening in the code.

lib/commons/color/get-background-color.js
WilcoFiers: ... No, I mean take this out of the code and put it into an issue.

test/checks/color/color-contrast.js
WilcoFiers: ... sorry, but I don't understand what this test proves. The way this description reads suggests that any time `normal.minThreshold` is used, the result must be `undefined`. That's clearly not what's going on. But then what is?

WilcoFiers (changes requested):
I think I figured out the ACT issue. Probably the same as this one: https://github.com/dequelabs/axe-core/pull/3719 Since we're avoiding overlap with the AAA rule in axe-core, to get the ACT rule tested, we need to map both axe-core rules to the ACT rule for AAA. You'll want to add `09o5cg` to the `actIds` prop on `rules/color-contrast.json`, and add `color-contrast` to `axeRules` in `text-contrast-enhanced-09o5cg.spec.js`.

lib/checks/color/color-contrast-evaluate.js
```diff
@@ -102,8 +102,8 @@ export default function colorContrastEvaluate(node, options, virtualNode) {
 
   // ratio is outside range
   if (
-    (typeof minThreshold === 'number' && contrast < minThreshold) ||
-    (typeof maxThreshold === 'number' && contrast > maxThreshold)
+    (contrast && typeof minThreshold === 'number' && contrast < minThreshold) ||
+    (contrast && typeof maxThreshold === 'number' && contrast > maxThreshold)
```
WilcoFiers, line 106:
```suggestion
    (typeof minThreshold === 'number' && (typeof contrast !== 'number' || contrast < minThreshold)) ||
    (typeof maxThreshold === 'number' && (typeof contrast !== 'number' || contrast > maxThreshold))
```

lib/commons/color/get-background-color.js
straker: https://github.com/dequelabs/axe-core/issues/3721

WilcoFiers (approved):

PR #3715, opened 2022-10-12
chore(target-size): consider fixed and non-fixed position nodes separately

WilcoFiers:
Have the target-size rule test fixed position elements separate from those not that are not fixed.

Closes issue: https://github.com/dequelabs/axe-core/issues/3672


straker (approved):

straker:
Just added a note since this is a different pr. We decided to not account for `sticky` elements. See https://github.com/dequelabs/axe-core/pull/3711#discussion_r991524983 for more details.

Issue #3708, opened 2022-10-08
list > group > listitem should not be allowed


WilcoFiers:
As of WAI-ARIA 1.2, groups are no longer allowed in lists. This is soon changing in ACT as well. We should make sure the following fails axe-core:

```html
<div role="list">
  <div role="group">
    <div role="listitem">list item</div>
  </div>
</div>
```

To fix this, remove `group` from `list` and `listitem` in standards/aria-roles.js, and fix any tests that fail.

WilcoFiers:
This was done a while back. Needs QA though.

padmavemulapati:
Validated with the latest axe-core develop branch code base,

`role=group` are no longer allowed inside lists and listitems.  
Test snippet:
```
 <div role="list" id="target"><ul role="group"><li role="listitem">Item</li></ul></div>
  ```
`aria-required-children` should fail here as this is holding role=group as child for role=list.

Also:
Require aria-controls on combobox. 
Require focusable separators to have aria-value now.
Require combobox, radio, seperator now require aria-checked. The default value of false is no longer assumed.
Test snippet:
```
<!--aria-req-parent should pass here-->
   <div role="tree"><div role="none"><div role="group"><div role="none"><div role="group"><div role="treeitem" id="target">Nothing here.</div></div></div></div></div></div>
   <!--aria-req-attr should fail when role=seperator--> 
   <span role="radio" id="aratrr_fail">I am BLUE!</span>
<div role="separator" id="violation" tabindex="0"></div>
<div id="target-fail" role="combobox" aria-expanded="invalid-value"></div>
```


![Image](https://user-images.githubusercontent.com/47409579/214545027-80c98bd9-97ad-45a3-aeaf-e102d6e813f9.png)



![Image](https://user-images.githubusercontent.com/47409579/214545197-5a73d6a4-7af1-4f0b-920c-a830d5bb80b9.png)




![Image](https://user-images.githubusercontent.com/47409579/214545308-a6dc4a9d-0339-4acd-87b6-f1481223ee83.png)

PR #3716, opened 2022-10-12
feat: prohibit ARIA 1.1 features not permitted in ARIA 1.2

WilcoFiers:
- Require aria-controls on combobox. aria-owns is no longer permitted as it is not sufficiently supported.
- Groups are no longer allowed inside lists. This was never properly supported by and is no longer valid in ARIA 1.2
- Require focusable separators to have aria-valuenow
- Require checkbox, radio, menuitemcheckbox, and menuitemradio now require aria-checked. The default value of false is no longer assumed

Closes issue: https://github.com/dequelabs/axe-core/issues/3708


straker (approved):

WilcoFiers:
This one's a bit late for axe-core 4.5. I'm going to defer this to 4.6. We'd be breaking policy to rush it into 4.5, I don't really see a reason for doing that.

WilcoFiers:
This PR was split.

PR #3723, opened 2022-10-17
chore: update playground

WilcoFiers:
I find myself repeatedly making changes to the playground to:

1. Add runOnly to run a single rule
2. Add a landmark and heading to ensure no violations from best practices

I also included testUtils, so I can use awaitNestedLoad, which is helpful from time to time.

### DIFF
--- a/test/playground.html
+++ b/test/playground.html
@@ -12,7 +12,7 @@
   <script>
     axe.testUtils.awaitNestedLoad(
       window,
-      async () => {
+      () => {
         axe.run(
           {
             runOnly: 'color-contrast',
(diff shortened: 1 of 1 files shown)


test/playground.html
```diff
@@ -1,18 +1,29 @@
...
-      axe.run(document, function(err, res) {
-        console.log(res.violations);
-        res.incomplete.forEach(function(issue) {
-          console.log(issue);
-        });
-      });
-    });
+    axe.testUtils.awaitNestedLoad(
+      window,
+      async () => {
```
stephenmathieson, line 15: Why is this `async` when we use a callback below?

WilcoFiers: Oh, right! Yeah because I initially used `await` on axe and then realised I needed a try/catch block, which was more code than just `err || results`

stephenmathieson (approved):

PR #3737, opened 2022-10-21
Revert "chore: run prettier"

bitcras:
Reverts dequelabs/axe-core#3725


WilcoFiers:
@bitcras why would you want this reverted?

PR #3744, opened 2022-10-24
chore: release axe-core 4.5.0 (sri-history corrected)

WilcoFiers:
Had an issue with sri-history.json because we did the release a bit out of order.


stephenmathieson (approved):
LGTM but won't this break things if the SRI is bad in 4.5.0?

straker (approved):

WilcoFiers:
@stephenmathieson No, the SRI was incorrect because we changed source after we created the sri-history file. We forgot to update it. This is the update.

Issue #3772, opened 2022-11-08
Unable to test contrast of nodes scrolled out of view


WilcoFiers:
This problem was introduced in axe-core 4.5. When an element is scrolled out of view, it is not included in the grid, which results (among other things) that it is reported as incomplete once its scrolled out of view. 

Here's how to reproduce the problem. These logs should report the same, but the second one reports incomplete instead of a violations.

```html
<h1 style="color: #ABC">Hello World</h1>
<main style="background: #CDF; height: 800px;"></main>
<script src="/axe.js"></script>
<script>
  window.onload = async () => {
    window.scrollTo(0, 0);
    const res0 =  await axe.run({ runOnly: 'color-contrast' })
    console.log({ violations: res0.violations, incomplete: res0.incomplete })

    window.scrollTo(0, 1000);
    const res1 =  await axe.run({ runOnly: 'color-contrast' })
    console.log({ violations: res1.violations, incomplete: res1.incomplete })
  }
</script>
```

dbjorge:
In case it helps with prioritization/sense of scale of the issue: We were just doing some validation of axe-core 4.5.0 in preparation for upgrading Accessibility Insights for Web and found that 4.5.0 detected about 90% fewer color-contrast issues in our big corpus of test sites compared to 4.4.1 (78 vs 836 issues detected). Based on spot checking the cases with different results I suspect this is probably the root cause, but I didn't verify every case.

straker:
@dbjorge axe-core 4.5.2 was released and should fix this issue. Would you be able to pull it down and test it to see if it returns results as expected?

dbjorge:
Yes, I'll rerun with 4.5.2 against our big batch of test sites overnight tonight!

dbjorge:
Yes, that resolved most of the differences.

I did find one class of failure that was reported in 4.4.1 but still not in 4.5.2. It involves a specific type of carousel control currently used in a few Skype marketing pages. An example is the bottom of the main content in https://www.skype.com/en/content-creators/ (I tried to make a more minimal repro but gave up after 5 minutes). This specific control has a few reported `color-contrast` issues where it renders some carousel items at `opacity: .2` as they transition onto the page. `axe-core` 4.4.1 and 4.5.2 both report those cases for which the content is visible on screen in this state as `color-contrast` failures. However, 4.4.1 *additionally* reports the items which are completely off-screen in this state as having color-contrast failures, where 4.5.2 does not report color-contrast failures for the off-screen items.

[...]

PR #3773, opened 2022-11-09
fix(create-grid): include elements scrolled out of view in the grid

WilcoFiers:
Closes issue: #3772

### DIFF
--- a/lib/commons/dom/create-grid.js
+++ b/lib/commons/dom/create-grid.js
@@ -320,7 +320,7 @@ function addNodeToGrid(grid, vNode) {
     // save a reference to where this element is in the grid so we
     // can find it even if it's in a subgrid
     vNode._grid ??= grid;
-    const gridRect = grid.getPositionOfRect(rect);
+    const gridRect = grid.getGridPositionOfRect(rect);
     grid.loopGridPosition(gridRect, gridCell => {
       if (!gridCell.includes(vNode)) {
         gridCell.push(vNode);
@@ -335,10 +335,20 @@ class Grid {
     this.cells = [];
   }
 
+  /**
+   * Convert x or y coordinate from rect, to a position in the grid
+   * @param {number}
+   * @returns {number}
+   */
   toGridIndex(num) {
     return Math.floor(num / constants.gridSize);
   }
 
+  /**
+   * Return an an array of nodes available at a particular grid coordinate
+   * @param {DOMPoint} gridPosition
+   * @returns {Array<AbstractVirtualNode>|null}
+   */
   getCellFromPoint({ x, y }) {
     const rowIndex = this.toGridIndex(y);
     const colIndex = this.toGridIndex(x);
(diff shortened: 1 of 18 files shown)


lib/commons/dom/create-grid.js
```diff
@@ -323,36 +323,66 @@ function findScrollRegionParent(vNode, parentVNode) {
...
+    startCol: Math.floor((left - margin) / gridSize),
+    endRow: Math.floor((bottom + margin - 1) / gridSize),
+    endCol: Math.floor((right + margin - 1) / gridSize)
+  };
+}
+
+export function getGridCellFromPoint({ cells, numCols }, { x, y }) {
+  const rowIndex = Math.floor(y / constants.gridSize);
+  const colIndex = Math.floor(x / constants.gridSize);
+  if (rowIndex >= cells.length || colIndex >= numCols) {
```
straker, line 383: This should take into account `_negativeIndex`. If the grid has 10 rows, but 9 of them are negative, a `rowIndex` of 2 should throw.

```suggestion
  if (rowIndex >= cells.length + cells._negativeIndex || colIndex >= numCols) {
```

Alternatively, we could set `numRows` like we do `numCols` as the largest `endRow` value so we don't have to use `length` of the  array.

straker (changes requested):

lib/commons/dom/create-grid.js
```diff
@@ -323,36 +324,75 @@ function findScrollRegionParent(vNode, parentVNode) {
...
+export function getGridCellFromPoint({ cells, boundaries }, { x, y }, debug) {
+  const rowIndex = toGridIndex(y);
+  const colIndex = toGridIndex(x);
+  const { top, right, bottom, left } = boundaries;
 
-      for (let col = startCol; col <= endCol; col++) {
-        grid.cells[row][col] = grid.cells[row][col] || [];
+  if (debug) {
+    console.log({ x, y }, { rowIndex, colIndex }, { top, right, bottom, left });
+  }
```
WilcoFiers, line 388: Oops, will fix

lib/commons/dom/create-grid.js
```diff
@@ -323,36 +316,75 @@ function findScrollRegionParent(vNode, parentVNode) {
...
+  toGridIndex(num) {
+    return Math.floor(num / constants.gridSize);
+  }
 
-    for (let row = startRow; row <= endRow; row++) {
-      grid.cells[row] = grid.cells[row] || [];
+  getCellFromPoint({ x, y }) {
+    const rowIndex = this.toGridIndex(y);
+    const colIndex = this.toGridIndex(x);
+    if (!isPointInRect({ y: rowIndex, x: colIndex }, this.boundaries)) {
```
straker, line 345: I think either the `pointInRect` function or this one should error or something if `this.boundries` is not set. This can happen when you can `getCellFromPoint` before calling `loopGridPosition` (which setups up the grid and boundries). Should inform when you call things out of order so we don't run into the problem in the future.

Or better yet, return early and return `null` if `this.boundaries` does not exist.

test/commons/dom/create-grid.js
```diff
@@ -4,13 +4,18 @@ describe('create-grid', function () {
   var fixture;
   var createGrid = axe.commons.dom.createGrid;
   var fixtureSetup = axe.testUtils.fixtureSetup;
+  var gridSize = axe.constants.gridSize;
 
   function findPositions(grid, vNode) {
     var positions = [];
+
     grid.cells.forEach(function (rowCells, rowIndex) {
```
straker, line 12: This can now be `grid.loopGridPosition(grid.boundries, ())`. However, we'll need the row/col position of the item and not just the `index`, so the `loopGridPosition` function should be updated to pass that back.

lib/commons/dom/create-grid.js
```diff
@@ -323,36 +316,75 @@ function findScrollRegionParent(vNode, parentVNode) {
...
+      throw new Error('Element midpoint exceeds the grid bounds');
+    }
+    const row = this.cells[rowIndex - this.cells._negativeIndex] ?? [];
+    return row[colIndex - row._negativeIndex] ?? null;
+  }
 
-      for (let col = startCol; col <= endCol; col++) {
-        grid.cells[row][col] = grid.cells[row][col] || [];
+  loopGridPosition(gridRect, callback) {
```
straker, line 352: Minor, we could name this `loopPosition` since the grid part is implied with the class.

lib/commons/math/get-bounding-rect.js
```diff
@@ -0,0 +1,7 @@
+export function getBoundingRect(rectA, rectB) {
```
straker, line 1: JSDoc comments. Also need the `@memberof` comment to link it to the Math functions.

test/commons/math/is-point-in-rect.js
```diff
@@ -0,0 +1,27 @@
...
+    assert.isTrue(isPointInRect(new DOMPoint(20, 40), rect));
+  });
+
+  it('returns false when the point is vertically outside the rect', () => {
+    const rect = new DOMRect(10, 20, 10, 20);
+    const point = new DOMPoint(15, 50);
+    assert.isFalse(isPointInRect(point, rect));
+  });
+
+  it('returns false when the point is vertically outside the rect', () => {
```
straker, line 22:
```suggestion
  it('returns false when the point is horizontally outside the rect', () => {
```

straker (dismissed):

straker (commented):
Direction LGTM

lib/commons/dom/create-grid.js
```diff
@@ -323,36 +317,100 @@ function findScrollRegionParent(vNode, parentVNode) {
...
+  getCellFromPoint({ x, y }) {
+    assert(this.boundaries, 'Grid does not have cells added');
+    const rowIndex = this.toGridIndex(y);
+    const colIndex = this.toGridIndex(x);
+    assert(
+      isPointInRect({ y: rowIndex, x: colIndex }, this.boundaries),
+      'Element midpoint exceeds the grid bounds'
+    );
+    const row = this.cells[rowIndex - this.cells._negativeIndex] ?? [];
+    return row[colIndex - row._negativeIndex] ?? null;
```
straker, line 362: Small thing. Maybe this should return an empty array instead of null? That way when you use it in get-rect-stack we wouldn't need to do default it to an empty array using `|| []` after the call?

straker (approved):

straker (approved):

Issue #3771, opened 2022-11-06
Webpack Error: axe-core cannot be extracted


TheJaredWilcurt:
### Product

axe-core

### Product Version

4.5.0 and 4.5.1

### Latest Version

- [X] I have tested the issue with the latest version of the product

### Issue Description

**Steps to Reproduce:**

I'm using `vue-axe` in a Vue 3, Vue-CLI project.

1. Install the Vue-CLI to generate an app from a boilerplate
   * `npm install @vue/cli`
   * `vue create bug-repro` > "Default Vue 3"
1. Follow the instructions in the `vue-axe` README:
   * https://github.com/vue-a11y/vue-axe-next#readme
1. Start webpack
   * `npm run serve`

**Expected:**

* Webpack starts up with no warnings/errors.

**Actual:**

```
WARNING  Compiled with 1 warning

warning  in ./node_modules/axe-core/axe.js

Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
```

**Additional information:**

Issue was not present in `v4.4.3`. It is present in `v4.5.0` and `v4.5.1`.

seanBlommaert:
The error seems to occur on 2 places. Hope this helps.

```
./node_modules/axe-core/axe.js 379:46-53
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
./node_modules/axe-core/axe.js 386:15-22
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
```

padmavemulapati:
dev task

PR #3777, opened 2022-11-10
fix: do not warn when using webpack

straker:
Following the steps in the ticket, this fixes the webpack output to not warn of the critical dependency. Seems something about esbuild v0.11+ causes the problem. For now this is a fix to the problem, but I'll investigate further at some later date to see if I can figure out what's going on. 

image

Closes issue: #3771

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,14 +20,14 @@
         "browser-driver-manager": "1.0.4",
         "chai": "~4.3.6",
         "chalk": "^4.x",
-        "chromedriver": "^105.0.1",
+        "chromedriver": "latest",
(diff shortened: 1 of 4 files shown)


stephenmathieson (commented):
I'm hesitant to downgrade this. IMO we ought to consider using a different build tool. Is 0.10 maintained?

TheJaredWilcurt:
Perhaps an issue should be created in the `esbuild` repo, so that this can be resolved there. Then once resolved, `axe-core` could update to the latest esbuild version with the fix. For now though, downgrading is the quickest/simplest solution.

WilcoFiers (approved):

Issue #3733, opened 2022-10-19
Runtime issue: Cannot read properties of undefined (reading 'source')


AlyssaLemon:
<< Please describe how you encountered this issue >>

axe-core version: `4.4.2`
axe-extension version: `4.37.1`

```
TypeError: Cannot read properties of undefined (reading 'source')
    at chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/axe-versions/latest/axe.js:12:167249
    at Array.map (<anonymous>)
    at chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/axe-versions/latest/axe.js:12:167219
    at Array.forEach (<anonymous>)
    at chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/axe-versions/latest/axe.js:12:167169
    at Array.forEach (<anonymous>)
    at chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/axe-versions/latest/axe.js:12:167065
    at Array.map (<anonymous>)
    at chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/axe-versions/latest/axe.js:12:166733
    at Array.map (<anonymous>)
```

straker:
Thanks for the issue. Is the site where you encountered this issue public that we could try to replicate the issue? If not, could your try running axe-core directly on the page and/or using one of our [axe-core integrations](https://github.com/dequelabs/axe-core-npm/tree/develop/packages) to see if it replicates the error? That would help us narrow down where the issue is occurring.

Issue #3627, opened 2022-08-31
Runtime issue: Cannot read properties of undefined (reading 'source')


JackieHayduk:
<< Please describe how you encountered this issue >>

axe-core version: `4.4.2`
axe-extension version: `4.36.2`

```
TypeError: Cannot read properties of undefined (reading 'source')
    at chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/axe-versions/latest/axe.js:12:167249
    at Array.map (<anonymous>)
    at chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/axe-versions/latest/axe.js:12:167219
    at Array.forEach (<anonymous>)
    at chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/axe-versions/latest/axe.js:12:167169
    at Array.forEach (<anonymous>)
    at chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/axe-versions/latest/axe.js:12:167065
    at Array.map (<anonymous>)
    at chrome-extension://lhdoppojpmngadmnindnejefpokejbdd/axe-versions/latest/axe.js:12:166733
    at Array.map (<anonymous>)
```

WilcoFiers:
Can you provide some information on how you got this error?

JackieHayduk:
Hello,

I was attempting to "scan part of the page" on a local hosted storybook
instance.

Thanks,
Jackie

Please Note: My working hours may be different than yours. Please do not
feel obligated to respond outside your normal working hours. (Philadelphia
Based | EST)

JACKIE HAYDUK

______________________________________________

Senior Front End Engineer • Certified Web Accessibility Specialist
Philadelphia,
PA


[image: Hero Digital Logo] <https://herodigital.com/>      [image: IAAP Web
Accessibility Certification]
<https://www.accessibilityassociation.org/s/wascertification>

WilcoFiers:
That'll be difficult to debug. I can't think of anything either related to running on localhost, or running storybook that could cause this issue. @scurker / @schne324 any thoughts?

padmavemulapati:
Cannt reproduce with axe-core also it is tracking from the extension, https://github.com/dequelabs/axe-extension/issues/3185 . So closing here

PR #3778, opened 2022-11-14
fix: prevent undefined relatedNodes from halting axe

WilcoFiers:
This PR fixes the reporter throwing an error when the relatedNode is undefined.

Instead the source is now reported as "Undefined"

Closes issues #3733 and #3627

### DIFF
--- a/lib/core/reporters/helpers/process-aggregate.js
+++ b/lib/core/reporters/helpers/process-aggregate.js
@@ -10,19 +10,19 @@ function normalizeRelatedNodes(node, options) {
       .forEach(checkRes => {
         checkRes.relatedNodes = checkRes.relatedNodes.map(relatedNode => {
           var res = {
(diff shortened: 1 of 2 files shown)


straker (approved):

lib/core/reporters/helpers/process-aggregate.js
```diff
@@ -10,19 +10,19 @@ function normalizeRelatedNodes(node, options) {
       .forEach(checkRes => {
         checkRes.relatedNodes = checkRes.relatedNodes.map(relatedNode => {
           var res = {
-            html: relatedNode.source
+            html: relatedNode?.source ?? 'Undefined'
```
dylanb, line 13: I would have made this all lowercase so it looks like A Javascript stringification of a variable with a value of `undefined`, but nbd

PR #3782, opened 2022-11-15
feat(checks/label-content-name-mismatch): deprecate occuranceThreshold option in favor of occurrenceThreshold to fix typo

nolandanley:
Fixes typos of the word "occurrence" throughout axe-core.

### DIFF
--- a/lib/checks/label/label-content-name-mismatch-evaluate.js
+++ b/lib/checks/label/label-content-name-mismatch-evaluate.js
@@ -38,7 +38,7 @@ function curateString(str) {
 }
 
 function labelContentNameMismatchEvaluate(node, options, virtualNode) {
-  const { pixelThreshold, occuranceThreshold } = options || {};
+  const { pixelThreshold, occurrenceThreshold } = options || {};
 
   const accText = accessibleText(node).toLowerCase();
   if (isHumanInterpretable(accText) < 1) {
@@ -50,7 +50,7 @@ function labelContentNameMismatchEvaluate(node, options, virtualNode) {
       subtreeDescendant: true,
       ignoreIconLigature: true,
       pixelThreshold,
-      occuranceThreshold
+      occurrenceThreshold
     })
   ).toLowerCase();
   if (!visibleText) {
--- a/lib/checks/label/label-content-name-mismatch-evaluate.js
+++ b/lib/checks/label/label-content-name-mismatch-evaluate.js
@@ -38,7 +38,8 @@ function curateString(str) {
 }
 
 function labelContentNameMismatchEvaluate(node, options, virtualNode) {
-  const { pixelThreshold, occurrenceThreshold } = options || {};
+  const pixelThreshold = options?.pixelThreshold;
+  const occurrenceThreshold = options?.occurrenceThreshold ?? options.occuranceThreshold;
 
   const accText = accessibleText(node).toLowerCase();
   if (isHumanInterpretable(accText) < 1) {
--- a/lib/checks/label/label-content-name-mismatch-evaluate.js
+++ b/lib/checks/label/label-content-name-mismatch-evaluate.js
@@ -39,7 +39,7 @@ function curateString(str) {
 
 function labelContentNameMismatchEvaluate(node, options, virtualNode) {
   const pixelThreshold = options?.pixelThreshold;
-  const occurrenceThreshold = options?.occurrenceThreshold ?? options.occuranceThreshold;
+  const occurrenceThreshold = options?.occurrenceThreshold ?? options?.occuranceThreshold;
 
   const accText = accessibleText(node).toLowerCase();
   if (isHumanInterpretable(accText) < 1) {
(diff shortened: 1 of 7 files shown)


lib/checks/label/label-content-name-mismatch.json
```diff
@@ -3,7 +3,7 @@
   "evaluate": "label-content-name-mismatch-evaluate",
   "options": {
     "pixelThreshold": 0.1,
-    "occuranceThreshold": 3
+    "occurrenceThreshold": 3
```
stephenmathieson, line 6: Unfortunately this is a breaking change. I'm not sure we can do this until we're ready to ship v5.

WilcoFiers: Agreed. Although what we could do is allow either variation, and deprecate the option with the typo.

lib/commons/text/accessible-text-virtual.js
```diff
@@ -104,11 +104,11 @@ function shouldIgnoreHidden(virtualNode, context) {
  * @return {Boolean}
  */
 function shouldIgnoreIconLigature(virtualNode, context) {
-  const { ignoreIconLigature, pixelThreshold, occuranceThreshold } = context;
+  const { ignoreIconLigature, pixelThreshold, occurrenceThreshold } = context;
```
WilcoFiers, line 107: This should avoid a break in the API.

```suggestion
  const { ignoreIconLigature, pixelThreshold } = context;
  const occurrenceThreshold = context.occurrenceThreshold ?? context.occurenceThreshold
```

WilcoFiers (changes requested):

lib/checks/label/label-content-name-mismatch.json
stephenmathieson: I like that idea!

lib/commons/text/accessible-text-virtual.js
nolandanley: Thanks for this! I think you mean "occuranceThreshold" since that was the only version of the typo used in code. "Occurence" was only used in a comment. Added this modified fix in [22d9b95a1d851fae3c850f0a5f125a4ba8bf0438](https://github.com/dequelabs/axe-core/pull/3782/commits/22d9b95a1d851fae3c850f0a5f125a4ba8bf0438).

lib/checks/label/label-content-name-mismatch-evaluate.js
```diff
@@ -38,7 +38,7 @@ function curateString(str) {
 }
 
 function labelContentNameMismatchEvaluate(node, options, virtualNode) {
-  const { pixelThreshold, occuranceThreshold } = options || {};
+  const { pixelThreshold, occurrenceThreshold } = options || {};
```
straker, line 41: This will also need to account for the old spelling as it passes the option into `subtreeText`

straker (changes requested):
Thanks, looking good so far, just need to update the name in one more place. Also, since we're going to deprecate the old name we'll need to update the pr title to match [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/). We deprecate things as a `feat` and want to make sure the title says which thing is deprecated, and either the pr title or body should say what to use instead.

lib/checks/label/label-content-name-mismatch-evaluate.js
nolandanley: I believe I've fixed that. Since pixelThreshold was being unpacked alone, I opted for just directly assigning it rather than by unpacking. Hope this is ok!

lib/checks/label/label-content-name-mismatch-evaluate.js
```diff
@@ -38,7 +38,8 @@ function curateString(str) {
 }
 
 function labelContentNameMismatchEvaluate(node, options, virtualNode) {
-  const { pixelThreshold, occuranceThreshold } = options || {};
+  const pixelThreshold = options?.pixelThreshold;
+  const occurrenceThreshold = options?.occurrenceThreshold ?? options.occuranceThreshold;
```
straker, line 42:
```suggestion
  const occurrenceThreshold = options?.occurrenceThreshold ?? options?.occuranceThreshold;
```

straker (changes requested):
Everything looks great. Looks like the tests are failing due to missing optional chaining. Tests should pass once it's added.

lib/checks/label/label-content-name-mismatch-evaluate.js
nolandanley: Committed! Sorry about all the simple issues. This is my first GitHub contribution so still have a lot to pick up. Thanks for helping me with it.

straker:
No problem! Thanks for contributing to axe-core.

lib/checks/label/label-content-name-mismatch-evaluate.js
straker: No problem! Thanks for contributing to axe-core.

straker (approved):

straker:
Reviewed for security

PR #3783, opened 2022-11-15
ci: create semantic pr title GHA

michael-siek:
<< Describe the changes >>

Closes issue:

### DIFF
--- a/.github/workflows/semantic-pr-title.yml
+++ b/.github/workflows/semantic-pr-title.yml
@@ -1,0 +1,45 @@
+name: Semantic PR Title
+
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - edited
+
+jobs:
+  validate:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check PR title
+        uses: actions/github-script@v6
+        env:
+          TITLE: ${{ github.event.pull_request.title }}
+        with:
+          script: |
+            const { TITLE } = process.env
+            const types = [
+              'feat',
+              'fix',
+              'perf',
+              'style',
+              'docs',
+              'refactor',
+              'test',
+              'build',
+              'ci',
+              'chore',
+              'Merge',
+              'Revert',
+              'Release'
+            ]
+            console.log('Validating PR title: "%s"', TITLE)
+            for (const type of types) {
+              if (TITLE.startsWith(type)) {
+                console.log('Title matches conventional commits (type: "%s").', type)
+                return
+              }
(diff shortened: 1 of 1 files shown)


.github/workflows/semantic-pr-title.yml
```diff
@@ -0,0 +1,45 @@
...
+              'build',
+              'ci',
+              'chore',
+              'Merge',
+              'Revert',
+              'Release'
+            ]
+            console.log('Validating PR title: "%s"', TITLE)
+            for (const type of types) {
+              if (TITLE.startsWith(type)) {
```
straker, line 38: I'd like to see this improved so it doesn't generate false positives. Each of the following titles would pass this check even though they don't follow [conventional commit specification](https://www.conventionalcommits.org/en/v1.0.0/#specification): 

- `fixture tests`
- `fix the tests`
- `circle ci failures`
- `fix(build) update build failures`
- `fix(build):update build failures`

This also does not allow for case insensitivity, so `FIX` should be allowed as a type

straker (changes requested):

.github/workflows/semantic-pr-title.yml
stephenmathieson: @WilcoFiers a 👎 without explanation is meaningless.

WilcoFiers: I'm with Steve on this. I'd rather have no action than one that gives me undeserved confidence. We shouldn't be copy/pasting code across repos like this either. We easily can, and should do better.

Is there any reason for us not to use this?
https://github.com/marketplace/actions/semantic-pull-request

stephenmathieson: > I'd rather have no action than one that gives me undeserved confidence.

This would flag PRs that are clearly wrong. It'd also give contributors the heads up that something needs to change. The rest should be left up to the human brain IMO.

> We shouldn't be copy/pasting code across repos like this either.

[A little copying is better than a little dependency](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=9m28s). Also there's nothing stopping us from making this a public action.

> Is there any reason for us not to use this?

Nope, I'd be fine with that. This was setup before that existed and has worked well for us (and other teams) for the past several months.

----

You and Steve seem to do **a lot** of gatekeeping in this repository. Michael is simply trying to help out by taking care of some low-hanging fruit. This seems like a pointless way of using your time.

WilcoFiers (approved):

PR #3787, opened 2022-11-17
fix(aria-required-children): Add mapping to ACT rule ff89c9

WilcoFiers:
Closes issue:


WilcoFiers:
This PR was split up.

PR #3791, opened 2022-11-21
fix(color-contrast): consider -webkit-text-stroke & -webkit-text-fill-color

### LINKED ISSUE #3356: -webkit-text-stroke- vs "insufficient contrast"
Product: <<axe-core | axe Extension | axe Linter>>

Expectation: << Describe what you expect the product to do >>

I made "meme font" for h1 on my site, using -webkit-text-stroke-width: 2px and -webkit-text-stroke-color: black. Font has just slightly different color than the background, but it's still easy to read, cause there's a black border around each letter

Actual: << Describe what the product actually does >>

This is accessibility report from lighthouse, and they said to write to you.  the -webkit-text-stroke-__ triggers "insufficient contrast ratio" for h1 tags, because there's very little difference in text and background color.  h1 text looks like "meme font", so the text is visible, due to each letter having distinctive "border" color of which has sufficient contrast. basically, background and font are white, but text-stroke is 2px thick and black. 

Now I don't know if the "meme" font is considered hard to read, seeing how it has "double lines", but the "insufficient contrast" warning, doesn't seem applicable, cause black text-stroke makes all the contrast it can make 

Motivation: << Describe why you want the behavior to be changed >>

the -webkit-text-stroke- css features are new, so that explains why they are not considered, but it should be, cause these features make the "meme font" and that has to be reasonably popular feature. I mean, this font is made in such a way, cause it's easy to read, no matter the background, so lighthouse probably shouldn't recognize this as "low visibility" issue 

<pre><code>
axe-core version: X.Y.Z
axe-webdriver, extension or other integration version: X.Y.Z

For Tooling issues:
- Node version: XX  
- Platform:  
</code></pre>

WilcoFiers:
Consider `-webkit-text-stroke` and its cousin `-webkit-text-fill-color` when calculating contrast

Closes #3356


test/commons/color/get-foreground-color.js
```diff
@@ -1,190 +1,211 @@
...
-    document.getElementById('fixture').innerHTML = '';
+describe('color.getForegroundColor', () => {
+  const { getForegroundColor, Color } = axe.commons.color;
+  const { queryFixture, queryShadowFixture } = axe.testUtils;
+
+  function assertSameColor(actual, expected, margin = 0) {
+    assert.closeTo(actual.red, expected.red, margin);
+    assert.closeTo(actual.green, expected.green, margin);
+    assert.closeTo(actual.blue, expected.blue, margin);
+    // Reduce the margin for alpha, since it is 100-based, rather than 255-based
```
WilcoFiers, line 9: On second thought, this should be done differently. the colors are 0 - 255, alpha is 0 - 1. For a fair comparison need to divide this by 255.

straker (changes requested):
Just a typo

straker (approved):

PR #3794, opened 2022-11-22
refactor: break up context.js and clean it up a bit

### LINKED ISSUE #3793: ContextObject type definition does not match actual context object
Our ContextObject is defined as follows:

```ts
type BaseSelector = string;
type ContextObject = {
  include?: Node | BaseSelector | Array<Node | BaseSelector | BaseSelector[]>;
  exclude?: Node | BaseSelector | Array<Node | BaseSelector | BaseSelector[]>;
};
```

At least two of those are impossible today:

```js
// This does not work at all. This tests the entire page:
axe.run({ include: 'main' })
// Oddly enough, this only tests `header`.
// Everything after the first string is ignored:
axe.run({ include: ['header', 'main'] })
```

Neither of these options is described in our docs. That makes me thing these were never supposed to work, and we created an error in the type definition. That said, we have tons of tests that do something like `{include: ['#fixture']}`, presumably others building on top of axe have done similar things. I don't really know why these do work. Going from the code it looks intentional. And to allow multiple strings in that array, the only change would be to change a `break` in a for-loop to a `continue`. That seems like a bug.

The way I see it, we can either change our types to what we actually support, or update what we support to match the types. I'm leaning towards the latter. It adds a little more complexity to Context, but makes them more intuitive in my opinion. If we didn't want that complexity we probably stop supporting `{include: ['#fixture']}`, which I'm concerned may break other implementations.

WilcoFiers:
In refactoring the context file I found this bug:

Related to #3793

### DIFF
--- a/lib/core/base/context.js
+++ b/lib/core/base/context.js
@@ -106,19 +106,10 @@ function isPageContext({ include }) {
  * @return {Error}
  */
 function validateContext(context) {
-  if (context.include.length !== 0) {
-    return;
-  }
-  if (context.frames.length === 0) {
+  if (context.include.length === 0 && context.frames.length === 0) {
     const env = respondable.isInFrame() ? 'frame' : 'page';
     throw new Error('No elements found for include in ' + env + ' Context');
   }
-
-  context.frames.forEach((frame, i) => {
-    if (frame.include.length === 0) {
-      throw new Error('No elements found for include in Context of frame ' + i);
-    }
-  });
 }
 
 /**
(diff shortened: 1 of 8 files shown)


lib/core/base/context.js
```diff
@@ -11,208 +12,6 @@ import {
...
-      var env = respondable.isInFrame() ? 'frame' : 'page';
-      return new Error('No elements found for include in ' + env + ' Context');
-    }
-    context.frames.forEach((frame, i) => {
-      if (frame.include.length === 0) {
-        return new Error(
-          'No elements found for include in Context of frame ' + i
-        );
-      }
-    });
```
WilcoFiers, line 191: I left this out because it never worked.We're returning in a forEach. That does nothing. This is completely pointless too. If include is empty, axe tests the entire frame.

straker (approved):

Issue #3765, opened 2022-11-01
Add include shadow element


ccheung0926:
### Product

axe-core

### Question

Hi,

I am currently working on a project for my company and would like to know how I can add my shadow element under the include selector list?

straker:
Currently this isn't possible. We [had a ticket for it](https://github.com/dequelabs/axe-core/issues/3029), but closed it due to lack of use cases. Do you have aa use case for doing this?

ccheung0926:
hi @straker 
Yes, I saw that ticket. And I would like to say we have a use case for it. I am not in the position to expose too much information here. But my company's accessibility team has a roadmap to integrate with Axe-core. The product team uses [lit library](https://lit.dev/) which is very heavy on shadow DOM.
It would be great if I can choose shadow DOM as part of the inclusive/exclusive shadow DOM selector prior to the Axe-core scan.

straker:
I think the best you could do at the moment is select the top-most custom element that has the shadow DOM. We can leave this ticket open for now now that we have a use case for it and start exploring it again.

ccheung0926:
@straker Sure, i can use the top-most custom element for now. But just let you know that that top-most element is a parent that nested with a lot of shadow DOMs developed by different UI teams 😅

WilcoFiers:
I think this would be a good thing for us to resolve. For now there is a workaround you could use. You can include DOM nodes, including shadow nodes. If you wanted you could create a little shadowRoot query function. Something like this:

```js
function getShadowNode(...queries) {
  let node, query;
  while (query = queries.shift()) {
    node = (node?.shadowRoot ?? node ?? document)
      .querySelector(query);
  }
  return node;
}
const shadowNode = getShadowNode('#shadow-root', '#nested-elm', '#my-team')
axe.run(shadowNode);
```

WilcoFiers:
In hindsight... axe-core has a utility function that lets you select individual nodes from shadow DOM trees that will work today:

```js
const shadowNode = axe.utils.shadowSelect('#shadow-root', '#nested-elm', '#my-team')
axe.run(shadowNode);
```

PR #3796, opened 2022-11-23
feat(utils): new shadowSelectAll utility

WilcoFiers:
This new `axe.utils.shadowSelectAll()` method should make it a little easier to explicitly include or exclude nodes in shadow DOM trees. This is a first step in providing a context that can include/exclude shadow DOM nodes without additional code.

Related to #3765

### DIFF
--- a/test/core/utils/shadow-select-all.js
+++ b/test/core/utils/shadow-select-all.js
@@ -1,0 +1,93 @@
+describe('utils.shadowSelectAll', () => {
+  const shadowSelectAll = axe.utils.shadowSelectAll;
+  const fixture = document.querySelector('#fixture');
+  const mapNodeName = elms => elms.map(elm => elm.nodeName.toLowerCase());
+
+  it('throws when not passed a string or array', () => {
+    assert.throws(() => {
+      shadowSelectAll(123);
+    });
+  });
+
+  it('throws when passed an array with non-string values', () => {
+    assert.throws(() => {
+      shadowSelectAll([123]);
+    });
+  });
+
+  describe('given a string', () => {
+    it('returns [] if no node is found', () => {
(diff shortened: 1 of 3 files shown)


test/core/utils/shadow-select-all.js
```diff
@@ -0,0 +1,93 @@
...
+
+      addShadowTree(root1.children[0], '<a></a>');
+      addShadowTree(root1.children[1], '<b></b>');
+      addShadowTree(root2.children[0], '<i></i>');
+      addShadowTree(root2.children[1], '<s></s>');
+
+      const nodeNames = mapNodeName(
+        shadowSelectAll(['#fixture > div', 'span', '*'])
+      );
+      assert.deepEqual(nodeNames, ['a', 'b', 'i', 's']);
```
straker, line 90: It might be good to have one of these deep shadow trees not match so the test can guarantee it's not just entering the shadow tree and getting all children when it shouldn't.

```suggestion
    it('returns nodes from multiple trees deep', () => {
      fixture.innerHTML = '<div></div><div></div>';
      const root1 = addShadowTree(
        fixture.children[0],
        '<span></span><span></span>'
      );
      const root2 = addShadowTree(
        fixture.children[1],
        '<div></div><span></span>'
      );

      addShadowTree(root1.children[0], '<a></a>');
      addShadowTree(root1.children[1], '<b></b>');
      addShadowTree(root2.children[0], '<i></i>');
      addShadowTree(root2.children[1], '<s></s>');

      const nodeNames = mapNodeName(
        shadowSelectAll(['#fixture > div', 'span', '*'])
      );
      assert.deepEqual(nodeNames, ['a', 'b', 's']);
```

straker (approved):
LGTM. Just a suggestion to improve the test, but not necessary.

PR #3798, opened 2022-11-24
feat(context): allow selecting shadow DOM nodes

### LINKED ISSUE #3765: Add include shadow element

### LINKED ISSUE #3793: ContextObject type definition does not match actual context object

### LINKED ISSUE #3799: Add syntax sugar for selecting nodes in frames and shadow DOM
Related to #3793

Currently, to include the main element inside an iframe you need to use a nested array syntax like so:

```js
axe.run([['iframe', 'main']])
```

That the outer array is to allow multiple elements, and the inner one is to target inside frame is difficult to read. This gets worse when we add shadow DOM selectors, which will introduce a third level array:

```js
axe.run([[['#shadow-host', 'main']]])
```

To make understanding these arrays easier, axe-core should allow labelling what node you're actually trying to grab. Doing this will also make wrapping optional, which further clears up the API. The above examples could then be rewritten like so:

```js
axe.run({ fromFrames: ['iframe', 'main'] })
axe.run({ fromShadowDom: ['#shadow-host', 'main'] })
```

WilcoFiers:
- [x] Add support for shadow DOM selection
- [x] Look at all uses of Context to ensure it works with shadow DOM selectors
- [x] Normalise fromFrames option when constructing Context
- [x] Normalise fromShadowDom option when constructing Context
- [x] Update type definition
- [x] create doc/context.md file
- [x] Update API.md file

Closes #3793, closes #3765, and closes #3799


lib/core/utils/contains.js
```diff
@@ -7,33 +7,25 @@
...
-        return true;
-      }
-      otherVNode = otherVNode.parent;
-    } while (otherVNode);
-  }
+  // Alternative method for shadow DOM / virtual tree tests
+  do {
+    if (vNode === otherVNode) {
+      return true;
+    } else if (otherVNode.nodeIndex < vNode.nodeIndex) {
```
WilcoFiers, line 24: Maybe add a test that this works when these are undefined.

lib/core/utils/contains.js
```diff
@@ -7,33 +7,25 @@
  * @return {Boolean}           Whether `vNode` contains `otherVNode`
  */
 export default function contains(vNode, otherVNode) {
-  /*eslint no-bitwise: 0*/
-  if (vNode.shadowId || otherVNode.shadowId) {
-    do {
-      if (vNode.shadowId === otherVNode.shadowId) {
-        return true;
```
WilcoFiers, line 14: This was wrong. We never caught it because `contains` is never called with vNode as a node in the shadow DOM tree.

lib/core/public/run-virtual-rule.js
```diff
@@ -37,10 +37,15 @@ function runVirtualRule(ruleId, vNode, options = {}) {
...
   const context = {
     initiator: true,
-    include: [vNode]
+    include: [vNode],
+    exclude: [],
+    frames: [],
+    page: false,
+    focusable: true,
+    size: {},
+    flatTree: []
```
WilcoFiers, line 48: Spotted some missing props. Don't know if it can cause a problem, but I figured I'd fix it and add a test.

lib/core/utils/is-node-in-context.js
```diff
@@ -1,46 +1,38 @@
 import contains from './contains';
 
-/**
- * Get the deepest node in a given collection
- * @private
- * @param  {Array} collection Array of nodes to test
- * @return {Node}             The deepest node
- */
-function getDeepest(collection) {
-  return collection.sort((a, b) => {
```
WilcoFiers, line 10: This shouldn't have used sort. We only need the last item in the array.

lib/core/base/context/parse-selector-array.js
```diff
@@ -12,29 +12,20 @@ export function parseSelectorArray(context, type) {
   const result = [];
   for (let i = 0, l = context[type].length; i < l; i++) {
     const item = context[type][i];
-    // selector
-    if (typeof item === 'string') {
```
WilcoFiers, line 16: This cannot happen. After normalising, this is always either a node, or an array..

lib/core/base/context/parse-selector-array.js
```diff
@@ -12,29 +12,20 @@ export function parseSelectorArray(context, type) {
...
       } else {
         result.push(getNodeFromTree(item));
       }
-      continue;
-    }
 
-    // Handle Iframe selection
-    if (item && item.length) {
+      // Handle Iframe selection
+    } else if (item && item.length) {
```
WilcoFiers, line 24: This might just as well be "else". If this isn't an array, normalization has a bug. Open to suggestions.

test/core/utils/contains.js
```diff
@@ -40,95 +71,112 @@ describe('axe.utils.contains', function () {
...
+      assert.isFalse(axe.utils.contains(node2, node1));
+    });
+
+    it('is false when the nodes are in adjacent trees', () => {
+      const section = createNestedShadowDom(
+        fixture,
+        `<section id="target"></section>`
+      );
+      createNestedShadowDom(section, `<img>`);
+      createNestedShadowDom(section, `<input>`);
```
straker, line 149: Did you mean for this to not have any shadow trees? Passing a single html code to `createNestedShadowDom` just appends the code to `innerHTML` and never creates a shadow tree, so the end result is just flat nodes:

```html
<div id="fixture">
  <section id="target">
    <img>
    <input>
  </section>
</div>
```

straker (changes requested):
This should include tests for the types in `typings/axe-core-tests.ts`. 

Also, when a shadow DOM selector targets an element that is not a shadow host, axe-core should call axe.log() to log the issue. Right now it's throwing with "No elements found for include in page Context"

lib/core/utils/contains.js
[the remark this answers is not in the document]
WilcoFiers: jup

doc/context.md
```diff
@@ -0,0 +1,290 @@
...
+  },
+  exclude: {
+    fromShadowDom: ['#root', 'app-footer', '.ad-banner']
+  }
+});
+```
+
+**Note**: The `fromShadowDom` property cannot be used on the same object as `include` and `exclude`.
+
+### Slotted elements
```
WilcoFiers, line 228: I need to add tests for this. I think this will work, but I didn't try it.

lib/core/base/context/normalize-context.js
```diff
@@ -1,52 +1,181 @@
...
+    `Invalid context; ${str}\nSee: https://github.com/dequelabs/axe-core/blob/master/doc/context.md`
+  );
+}
+
+// Wrapper to prevent throwing for non-objects & null
+function objectHasOwn(obj, prop) {
+  if (!obj || typeof obj !== 'object') {
+    return false;
+  }
+  return Object.prototype.hasOwnProperty.call(obj, prop);
```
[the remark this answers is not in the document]
WilcoFiers: I don't recall where, but doing that broke something... This seemed safer. I was using Object.hasOwn first, but that apparently doesn't work in some of our examples. Old version of Node?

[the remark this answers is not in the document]
WilcoFiers: Ahh yes, this: https://github.com/dequelabs/axe-core/blob/develop/test/core/base/context.js#L331

test/core/utils/contains.js
```diff
@@ -1,37 +1,68 @@
...
-      node1 = {
+  it('is false when the nodes are a cousins', () => {
+    const tree = fixtureSetup(`<section> <input> </section> <img>`);
+    const node1 = axe.utils.querySelectorAll(tree, 'input')[0];
+    const node2 = axe.utils.querySelectorAll(tree, 'img')[0];
+    assert.isFalse(axe.utils.contains(node1, node2));
+    assert.isFalse(axe.utils.contains(node2, node1));
+  });
+
+  describe.skip('using fallbacks', () => {
```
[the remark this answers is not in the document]
WilcoFiers: Ahh yes. I got rid of .compareDocumentPosition. Don't care for maintaining that code path, but I can keep the other two tests.

doc/context.md
```diff
@@ -0,0 +1,290 @@
...
+```
+
+To select elements in nested frames, axe will need a selector for each level of nesting. Because axe tests frames recursively, this can be any number of levels deep. The following shows how to test the `form`, inside an `#inner`, inside an `#outer` iframe:
+
+```js
+// Test <form> inside, #inner, inside #outer:
+axe.run({ fromFrames: ['iframe#outer', 'iframe#inner', 'form'] });
+```
+
+The `fromFrames` object can be used as part of an `exclude` or `include` property. It can be by itself, or as part of an array along with other selectors. The following example shows how to exclude all `.ad-banner` elements on the top window, as well as any that are part of the first level of iframes:
```
enlarsen, line 154:
```suggestion
The `fromFrames` object can be used as part of an `exclude` or `include` property. It can be by itself or as part of an array along with other selectors. The following example shows how to exclude all `.ad-banner` elements on the top window, as well as any that are part of the first level of iframes:
```

doc/context.md
```diff
@@ -0,0 +1,290 @@
...
+    fromFrames: ['iframe#payment', 'form > .ad-banner']
+  }
+});
+```
+
+**Note**: The `fromFrames` property cannot be used on the same object as `include` and `exclude`.
+
+## Limit Shadow DOM Testing
+
+To include or exclude specific sections of a [shadow DOM](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM) tree can be done with a `fromShadowDom` selector object. This works similar to the [`fromFrames` selector](#limit-frame-testing). The `fromShadowDom` property takes an array of strings, the first to select the shadow DOM host element(s) and the last to select the element(s) to include or exclude. The following example shows how to test the `#search` form inside the shadow DOM tree attached to the `.app-header` element:
```
enlarsen, line 185:
```suggestion
Including or excluding specific sections of a [shadow DOM](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM) tree can be done with a `fromShadowDom` selector object. This works similar to the [`fromFrames` selector](#limit-frame-testing). The `fromShadowDom` property takes an array of strings: the first to select the shadow DOM host element(s) and the last to select the element(s) to include or exclude. The following example shows how to test the `#search` form inside the shadow DOM tree attached to the `.app-header` element:
```

doc/context.md
enlarsen, line 228:
```suggestion
### Slotted Elements
```

doc/context.md
```diff
@@ -0,0 +1,290 @@
...
+  },
+  exclude: {
+    fromShadowDom: ['#shadowForm', 'slot']
+  }
+});
+```
+
+## Combine Shadow DOM and Frame Context
+
+To select frames inside shadow DOM trees, or shadow DOM trees inside frames it is possible to use the a [`fromShadowDom`](#limit-shadow-dom-testing) as a selector in the [`fromFrames`](#limit-frame-testing) selector object. The following example shows how to test the `main` element, inside each `iframe`, that is part of the shadow DOM tree of `#appRoot`:
```
enlarsen, line 246:
```suggestion
To select frames inside shadow DOM trees or shadow DOM trees inside frames, it is possible to use [`fromShadowDom`](#limit-shadow-dom-testing) as a selector in the [`fromFrames`](#limit-frame-testing) selector object. The following example shows how to test the `main` element, inside each `iframe` that is part of the shadow DOM tree of `#appRoot`:
```

doc/context.md
```diff
@@ -0,0 +1,290 @@
...
+        }
+      ]
+    }
+  ]
+});
+```
+
+**Note**: Even though the iframe is inside the shadow DOM tree, the `fromShadowDom` selector object must be part of the `fromFrames` selector. Doing the reverse does not work and will cause an error.
+
+## Implicit Frame And Shadow DOM Selection
```
enlarsen, line 279:
```suggestion
## Implicit Frame and Shadow DOM Selection
```

doc/context.md
```diff
@@ -0,0 +1,290 @@
...
+
+In earlier versions of axe, using nested arrays was the only way to include or exclude elements inside iframes. These are still supported in axe, and is how axe works internally. This nested array syntax is used in the `target` property. For example, the following is a possible selector for a `.commentBody` element, in the shadow DOM tree of `#userComments`, inside the `#blog-comments` iframe:
```
enlarsen, line 281:
```suggestion
In earlier versions of axe, using nested arrays was the only way to include or exclude elements inside iframes. These are still supported in axe and is how axe works internally. This nested array syntax is used in the `target` property. For example, the following is a possible selector for a `.commentBody` element, in the shadow DOM tree of `#userComments`, inside the `#blog-comments` iframe:
```

doc/context.md
```diff
@@ -0,0 +1,290 @@
...
+
+```js
+result = await axe.run();
+result.violations[0].nodes[0].target; // ['#blog-comments', ['#userComments', '.commentBody']]
+```
+
+To pass a `target` property into axe it must be wrapped in another array. This creates three nested arrays. The outer array to allow multiple selectors, the middle array to select into frames, and the inner array, which is optional, to select elements inside shadow DOM trees.
```
enlarsen, line 288:
```suggestion
To pass a `target` property into axe, it must be wrapped in another array. This creates three nested arrays: the outer array to allow multiple selectors, the middle array to select into frames, and the inner array, which is optional, to select elements inside shadow DOM trees.
```

doc/context.md
```diff
@@ -0,0 +1,290 @@
...
+
+While this syntax continues to be supported, it is recommended to avoid it, and to use the [fromFrames](#limit-frame-testing) and [fromShadowDom](#limit-shadow-dom-testing) object selectors instead, since these are more clearer and don't need to be wrapped in otherwise empty arrays.
```
enlarsen, line 290:
```suggestion
While this syntax continues to be supported, it is recommended to avoid it and to use the [fromFrames](#limit-frame-testing) and [fromShadowDom](#limit-shadow-dom-testing) object selectors instead since these are clearer and don't need to be wrapped in otherwise empty arrays.
```

doc/API.md
```diff
@@ -321,106 +321,64 @@ By default, `axe.run` will test the entire document. The context object is an op
...
-    exclude: $fixture[0].firstChild
-  },
-  (err, results) => {
-    // ...
-  }
-);
-```
-
-2. Include the element with the ID of `fix` but exclude any `div`s within it
+1. Test the `#navBar` and all other `nav` elements and its content
```
enlarsen, line 332:
```suggestion
1. Test the `#navBar` and all other `nav` elements and its content.
```

doc/API.md
```diff
@@ -321,106 +321,64 @@ By default, `axe.run` will test the entire document. The context object is an op
...
-    // ...
-  }
-);
+axe.run([`#navBar`, `nav`], (err, results) => {
+  // ...
+});
 ```
 
-3. Include the whole document except any structures whose parent contains the class `exclude1` or `exclude2`
+2. Test everything except `.ad-banner` elements
```
enlarsen, line 340:
```suggestion
2. Test everything except `.ad-banner` elements.
```

doc/API.md
```diff
@@ -321,106 +321,64 @@ By default, `axe.run` will test the entire document. The context object is an op
...
+    exclude: '.ad-banner'
   },
   (err, results) => {
     // ...
   }
 );
 ```
 
-4. Include the element with the ID of `fix`, within the iframe with id `frame`
+3. Test the `form` element inside the `#payment` iframe
```
enlarsen, line 353:
```suggestion
3. Test the `form` element inside the `#payment` iframe.
```

doc/API.md
```diff
@@ -321,106 +321,64 @@ By default, `axe.run` will test the entire document. The context object is an op
...
+    fromFrames: ['iframe#payment', 'form']
   },
   (err, results) => {
     // ...
   }
 );
 ```
 
-5. Include the element with the ID of `fix`, within the iframe with id `frame2`, within the iframe with id `frame1`
+4. Exclude all `.commentBody` elements in each `.commentsShadowHost` shadow DOM tree
```
enlarsen, line 366:
```suggestion
4. Exclude all `.commentBody` elements in each `.commentsShadowHost` shadow DOM tree.
```

doc/context.md
```diff
@@ -0,0 +1,290 @@
+# Axe Testing Context
+
+Axe-core's `context` argument is a powerful tool for controlling precisely which elements are tested and which are ignored. The context lets you do many things, including:
+
+1. [Test specific elements](#test-specific-elements)
```
enlarsen, line 5:
```suggestion
1. [Test specific elements](#test-specific-elements).
```

doc/context.md
```diff
@@ -0,0 +1,290 @@
...
+1. [Test DOM nodes](#test-dom-nodes)
```
enlarsen, line 6:
```suggestion
1. [Test DOM nodes](#test-dom-nodes).
```

doc/context.md
```diff
@@ -0,0 +1,290 @@
...
+1. [Excludes elements from test](#exclude-elements-from-test)
+1. [Select from prior tests](#select-from-prior-tests)
```
enlarsen, line 8:
```suggestion
1. [Select from prior tests](#select-from-prior-tests).
```

doc/context.md
```diff
@@ -0,0 +1,290 @@
...
+1. [Limit frame testing](#limit-frame-testing)
```
enlarsen, line 9:
```suggestion
1. [Limit frame testing](#limit-frame-testing).
```

doc/context.md
```diff
@@ -0,0 +1,290 @@
...
+1. [Limit shadow DOM testing](#limit-shadow-dom-testing)
```
enlarsen, line 10:
```suggestion
1. [Limit shadow DOM testing](#limit-shadow-dom-testing).
```

doc/context.md
```diff
@@ -0,0 +1,290 @@
...
+1. [Combine shadow DOM and frame context](#combine-shadow-dom-and-frame-context)
```
enlarsen, line 11:
```suggestion
1. [Combine shadow DOM and frame context](#combine-shadow-dom-and-frame-context).
```

test/core/utils/contains.js
WilcoFiers: Ooff! Good catch. This test is completely broken. Will fix.

doc/context.md
```diff
@@ -0,0 +1,290 @@
...
+const nodes = document.querySelector('main, header');
+await axe.run(nodes);
+
+// Test an array of nodes:
+const navbar = document.getElementById('navbar');
+const cookiePopup = document.getElementById('cookie-popup');
+await axe.run([navbar, cookiePopup]);
+```
+
+### Component Frameworks
```
straker, line 56: If we're listing component frameworks, could we also list a small section for testing frameworks like Jest? Then we could add a quick example showing that you need to mount instead of shallowMount a component for testing

doc/context.md
```diff
@@ -0,0 +1,290 @@
...
+```js
+// Test each search form inside each <app-header> shadow DOM tree.
+axe.run({ fromShadowDom: ['.app-header', 'form#search'] });
+```
+
+To select elements in nested shadow DOM trees, axe will need a selector for each level of nesting. The following shows how to test the `nav` element inside the `.header` element's Shadow DOM tree, inside the `app-root` custom element's shadow DOM tree:
+
+```js
+// Test #search, inside each .header, inside each <app-root>
+axe.run({ fromShadowDom: ['app-root', '.header', '#search'] });
```
straker, line 196:
```suggestion
axe.run({ fromShadowDom: ['app-root', '.header', 'nav'] });
```

doc/context.md
```diff
@@ -0,0 +1,290 @@
...
+## Limit Shadow DOM Testing
+
+To include or exclude specific sections of a [shadow DOM](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM) tree can be done with a `fromShadowDom` selector object. This works similar to the [`fromFrames` selector](#limit-frame-testing). The `fromShadowDom` property takes an array of strings, the first to select the shadow DOM host element(s) and the last to select the element(s) to include or exclude. The following example shows how to test the `#search` form inside the shadow DOM tree attached to the `.app-header` element:
+
+```js
+// Test each search form inside each <app-header> shadow DOM tree.
+axe.run({ fromShadowDom: ['.app-header', 'form#search'] });
+```
+
+To select elements in nested shadow DOM trees, axe will need a selector for each level of nesting. The following shows how to test the `nav` element inside the `.header` element's Shadow DOM tree, inside the `app-root` custom element's shadow DOM tree:
```
straker, line 192: The example shows `#search` so just making the description match

```suggestion
To select elements in nested shadow DOM trees, axe will need a selector for each level of nesting. The following shows how to test the `#search` element inside the `.header` element's Shadow DOM tree, inside the `app-root` custom element's shadow DOM tree:
```

doc/context.md
WilcoFiers: These don't seem like they should have a period af the end.

doc/context.md
```diff
@@ -0,0 +1,290 @@
...
+Axe-core's `context` argument is a powerful tool for controlling precisely which elements are tested and which are ignored. The context lets you do many things, including:
+
+1. [Test specific elements](#test-specific-elements)
+1. [Test DOM nodes](#test-dom-nodes)
+1. [Excludes elements from test](#exclude-elements-from-test)
+1. [Select from prior tests](#select-from-prior-tests)
+1. [Limit frame testing](#limit-frame-testing)
+1. [Limit shadow DOM testing](#limit-shadow-dom-testing)
+1. [Combine shadow DOM and frame context](#combine-shadow-dom-and-frame-context)
+1. [Implicit frame and shadow DOM selection](#implicit-frame-and-shadow-dom-selection)
```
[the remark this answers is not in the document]
WilcoFiers: I see where you're coming from, but this isn't a call to action thing. I don't want people to do this. It's more about explaining some background.

doc/context.md
WilcoFiers: Good point, but I'm going for a note instead.

doc/context.md
enlarsen: I was iffy on these because I was reading them as sentences, but I can easily see *not* reading them as sentences. Might be clearer then to rewrite them to use gerunds.

doc/context.md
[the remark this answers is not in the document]
enlarsen: Makes sense.

typings/axe-core/axe-core-tests.ts
```diff
@@ -29,6 +29,38 @@ axe.run(
...
 );
+export async function runAsync() {
+  await axe.run('main'); // Single selector
+  await axe.run(['main']); // Array of one selector
+  await axe.run({ include: ['main', document.head] });
+  await axe.run(document.querySelector('main'));
+  await axe.run(document.querySelectorAll('main'));
+  // axe.run with frameContext context
+  await axe.run({ fromShadowDom: ['#app', '#main', '#inner'] });
+  // @ts-expect-error // Must be two long:
```
straker, line 40: I did not know about `@ts-expect-error`. That's really cool.

typings/axe-core/axe-core-tests.ts
```diff
@@ -29,6 +29,38 @@ axe.run(
     console.log(error || results);
   }
 );
+export async function runAsync() {
+  await axe.run('main'); // Single selector
+  await axe.run(['main']); // Array of one selector
+  await axe.run({ include: ['main', document.head] });
+  await axe.run(document.querySelector('main'));
+  await axe.run(document.querySelectorAll('main'));
+  // axe.run with frameContext context
```
straker, line 38:
```suggestion
  // axe.run with shadow DOM context
```

typings/axe-core/axe-core-tests.ts
```diff
@@ -29,6 +29,38 @@ axe.run(
     console.log(error || results);
   }
 );
+export async function runAsync() {
+  await axe.run('main'); // Single selector
```
straker, line 33: We'll also want to add tests for unlabeled frame and shadow DOM selectors.

straker (changes requested):

straker (approved):

Issue #3788, opened 2022-11-17
empty-table-header virtualised rule should run without requiring role


WilcoFiers:
This test needs to be added, and needs to pass.

```js
  it('should fail if table header has no child nodes', function () {
    var node = new axe.SerialVirtualNode({
      nodeName: 'th',
      attributes: {}
    });
    node.children = [];

    var results = axe.runVirtualRule('empty-table-header', node);

    assert.lengthOf(results.passes, 0);
    assert.lengthOf(results.violations, 1);
    assert.lengthOf(results.incomplete, 0);
  });
```

dbowling:
@WilcoFiers - as I was looking at this tonight, I see we actually [already have this test as part of my original PR](https://github.com/dequelabs/axe-core/pull/3621/files?diff=split&w=1#diff-8b7797425f99ada52d4cfcff689f3e8b56e74b5c84e90eba0c440774b33a4d3eR2-R13), but our expectation was that it incompletes:

```JavaScript
it('should incomplete when children are missing', function () {
    var thNode = new axe.SerialVirtualNode({
      nodeName: 'th'
    });
    thNode.children = [];

    var results = axe.runVirtualRule('empty-table-header', thNode);

    assert.lengthOf(results.passes, 0);
    assert.lengthOf(results.violations, 0);
    assert.lengthOf(results.incomplete, 1);
  });
```

**Note**: existing test is has `attributes: {}` vs `attributes: undefined`, and has different asserts for `results.violations` and `results.incomplete`.

The test was based on [a suggestion](https://github.com/dequelabs/axe-core/pull/3621/files?diff=split&w=1#r956142997) from Steve:

> We'll want a test that should return incomplete if the `th` element does not have the children set to an empty array. Similar to [button-name](https://github.com/dequelabs/axe-core/blob/develop/test/integration/virtual-rules/button-name.js#L116-L125) 

TTBOMK, the empty object vs `undefined` for `attributes` isn't something that this code should differentiate on, so I'm not seeing how things need to change. Let me know if that's not the case and I will look with fresh eyes, and maybe we can chat about how to make a test that is more expressive about intent.

straker:
[...]

PR #3800, opened 2022-11-28
fix(get-role): handle presentation role inheritance for vnodes with no parent

dbowling:
Closes issue: #3788


dbowling:
Wrong branch

PR #3801, opened 2022-11-28
fix(get-role): handle presentation role inheritance for vnodes with no parent

dbowling:
Closes issue: #3788


test/integration/virtual-rules/empty-table-header.js
```diff
@@ -1,17 +1,61 @@
 describe('empty-table-header virtual-rule', function () {
-  it('should incomplete when children are missing', function () {
+  it('should fail when children contain no visible text', function () {
```
WilcoFiers, line 2: Can you add a couple more tests:

With empty content:
1. it passes when the table has role=none
2. it passes when the th has role=cell

WilcoFiers (changes requested):
LGTM, although it'd be good to add a couple more tests to prove this out.

test/integration/virtual-rules/empty-table-header.js
dbowling: >  it passes when the table has role=none

Steve said that the following test needs to be a new ticket as it is not a related issue. He was going to chat with you today about it I think. 

This test currently fails:

```
it('should pass with a table of role none', function () {
	var table = new axe.SerialVirtualNode({
		nodeName: 'table',
		attributes: {
			role: 'none'
		}
	});

	var tr = new axe.SerialVirtualNode({
		nodeName: 'tr',
	});

	var th = new axe.SerialVirtualNode({
		nodeName: 'th',
	});

	tr.children = [th];
	tr.parent = table;
	th.parent = tr;
	th.children = [];
	table.children = [tr];

	var results = axe.runVirtualRule('empty-table-header', table);

	assert.lengthOf(results.passes, 1);
	assert.lengthOf(results.violations, 0);
	assert.lengthOf(results.incomplete, 0);
});
```


> it passes when the th has role=cell

Test has been added and passes.

dbowling:
Issue added, #3817

PR #3804, opened 2022-11-29
ci: Add a workflow for creating releases

stephenmathieson:
Example release PR: https://github.com/dequelabs/axe-core/pull/3814

Ref https://github.com/dequelabs/axe-api-team/issues/304

<< Describe the changes >>

Closes issue:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -50,3 +50,5 @@ jobs:
           echo "" >> /tmp/pr.txt
           echo "This PR was opened by a robot :robot: :tada:" >> /tmp/pr.txt
           hub pull-request --file /tmp/pr.txt --base main
+        env:
+          GITHUB_TOKEN: ${{ secrets.PAT }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,6 +10,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
       - uses: actions/cache@v3
         with:
           path: ./node_modules
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,7 +26,6 @@ jobs:
           echo "Branch: $Branch"
           git checkout -b "$Branch"
 
-          npm ci
           npm run release
 
           git push origin "$Branch" --force
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -31,6 +31,8 @@ jobs:
           npm ci
           npm run release
 
+          git push origin "$Branch" --force
+
           Version=$(jq -r .version ./package.json)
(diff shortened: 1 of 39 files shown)


WilcoFiers (dismissed):
LGTM, I would like @straker's review of this too before we merge.

.github/workflows/release.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Run release script and open PR
+        run: |
+          git config user.name "API Team CI User"
+          git config user.email "aciattestteamci@deque.com"
+
+          Branch="release-$(date +%Y-%m-%d)"
+          echo "Branch: $Branch"
+          git checkout -b "$Branch"
+
+          npm ci
```
straker, line 29: The release [already runs `npm ci`](https://github.com/dequelabs/axe-core/blob/develop/package.json#L60-L62) so no need to run it before

```suggestion
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,52 @@
...
+          npm run release
+
+          git push origin "$Branch" --force
```
straker, line 32: Why the force push?

.github/workflows/release.yml
```diff
@@ -0,0 +1,52 @@
...
+
+          Version=$(jq -r .version ./package.json)
+          echo "Version: $Version"
+
+          # Stolen from https://github.com/dequelabs/axe-extension/blob/0f22963c4d0a73bf853aa982ade2393e422f002c/scripts/release.sh#L51-L54
```
straker, line 37: This is a private url, shouldn't list it in public repo.

```suggestion
          # Get the additions to the changelog as the commit body and generate the PR url
```

straker (changes requested):

.github/workflows/release.yml
stephenmathieson: If the branch already has changes in it, we'll want to overwrite them.

.github/workflows/release.yml
stephenmathieson: Fair enough 🤷

.github/workflows/release.yml
straker: Wouldn't the checkout action fail then and kill the build before it got to this point?

stephenmathieson: No, the checkout action is taking whatever branch we're releasing from (usually `develop`). We then create this branch, add the release commit, then push. The `--force` is to ensure any prior pending release (maybe someone made a mistake) is overwritten.

This is, however, fixing a problem we haven't had (yet?). I'm happy to pull it out if you aren't comfortable force pushing.

straker: Wilco didn't seem to think it was a problem when he reviewed it, so I'm fine with it.

straker (approved):

WilcoFiers (approved):

PR #3808, opened 2022-12-02
chore: improve type definitions

WilcoFiers:
Address some issues introduced in #3798

1. Separate SerialContextObject from FrameContextObject
2. Allow changing the reporter type
3. Fix issue with the frameSelector type -- This should have been a CrossTreeSelector

### DIFF
--- a/axe.d.ts
+++ b/axe.d.ts
@@ -81,12 +81,28 @@ declare namespace axe {
       };
   type ElementContext = Selector | SelectorList | ContextObject;
 
-  interface SerialContextObject {
+  type SerialSelector =
+    | BaseSelector
+    | LabelledShadowDomSelector
+    | LabelledFramesSelector;
+  type SerialSelectorList = Array<SerialSelector | FramesSelector>;
+
+  type SerialContextObject =
+    | {
+        include: SerialSelector | SerialSelectorList;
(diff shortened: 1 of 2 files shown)


axe.d.ts
```diff
@@ -81,12 +81,28 @@ declare namespace axe {
       };
   type ElementContext = Selector | SelectorList | ContextObject;
 
-  interface SerialContextObject {
+  type SerialSelector =
+    | BaseSelector
+    | LabelledShadowDomSelector
+    | LabelledFramesSelector;
+  type SerialSelectorList = Array<SerialSelector | FramesSelector>;
```
straker, line 88: While we're here, it'd be helpful to have the internal type pulled out so it can be used directly. That way it's controlled by axe-core rather than duplicated across our apis, so if we update what `SerialSelectorList` is we don't have to update the type that `include` and `exclude` takes throughout all our apis.

Something like this (not sure on the name though)

```suggestion
  type SerialFrameSelector = SerialSelector | FramesSelector;
  type SerialSelectorList = Array<SerialFrameSelector>;
```

straker (dismissed):
Tested this on one of our apis and it seems to work. Would like @michael-siek to approve this as well before we merge.

michael-siek (approved):

PR #3815, opened 2022-12-05
ci: run on node 16.18

piwysocki:
run on latest node 16.x


WilcoFiers (approved):
Not sure why you created this PR, but it's appreciated!

PR #3826, opened 2022-12-12
chore: update pr template to use closes syntax

straker:


stephenmathieson (approved):
<< LGTM >>

<< Why do we do this? >>

PR #3846, opened 2023-01-04
ci: streamline ci

straker:
While looking through the circle ci script I saw a few things we could improve to speed up the process or just clean up the file a bit. In total, this shaved off about 3.5 minutes from the build.

- current config: avg. 12m (based on the last few runs, 10m min, 13m max).
- streamlined: 8m 42s

Here's the changes:

- check if the npm cache was restored before running `npm ci` and saving the cache. `npm ci` [removes `node_modules` before installing](https://docs.npmjs.com/cli/v9/commands/npm-ci), so restoring the cache before installing isn't helpful. [You can see this working here](https://app.circleci.com/pipelines/github/dequelabs/axe-core/4771/workflows/35937673-9a59-4eac-8ed0-a588b86ba21a/jobs/53463) and [when the cache doesn't exist here](https://app.circleci.com/pipelines/github/dequelabs/axe-core/4772/workflows/ec1854c0-0a79-4e2a-be85-38ebb1ca44f6/jobs/53465)
- cache the axe.js file as we build it in every step. The cache key is based on the [current git sha](https://circleci.com/docs/pipeline-variables/)
- cleaned up the dependency chain for `release` as it depended on `hold` which already depended on everything else
- made the `next_release` step depend on all the jobs instead of just some

I'm open to suggestions and/or changes if we don't think we should cache the axe.js build.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,7 +22,11 @@ restore_dependency_cache_unix: &restore_dependency_cache_unix
   restore_cache:
     keys:
       - v9-cache-unix-{{ checksum "package-lock.json" }}
-      - v9-cache-unix-
+
+restore_build: &restore_build
+  restore_cache:
+    keys:
+      - v9-cache-build-{{ pipeline.git.revision }}
 
 jobs:
   # Fetch and cache dependencies.
@@ -33,13 +37,25 @@ jobs:
       - browser-tools/install-browser-tools
       - checkout
       - <<: *set_npm_auth
-      - <<: *restore_dependency_cache_unix
       - run: npm ci
       - run: npx browser-driver-manager install chromedriver --verbose
       - save_cache:
           key: v9-cache-unix-{{ checksum "package-lock.json" }}
           paths:
             - node_modules
+
+  # Build and cache axe.js
+  build_unix:
+    <<: *defaults
+    <<: *unix_box
+    steps:
+      - checkout
+      - <<: *restore_dependency_cache_unix
+      - run: npm run build
+      - save_cache:
+          key: v9-cache-build-{{ pipeline.git.revision }}
+          paths:
+            - axe.js
 
   # Run ESLINT
   lint:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -20,27 +20,55 @@ set_npm_auth: &set_npm_auth
...
-      - browser-tools/install-browser-tools
       - checkout
-      - <<: *set_npm_auth
       - <<: *restore_dependency_cache_unix
+      - run:
+          name: Skip Install If Cache Exists
+          command: |
+            if [ -d "node_modules" ]; then
+                echo "node_modules exist"
+                circleci step halt
```
stephenmathieson, line 46: Fancy! I didn't know this was a thing.

stephenmathieson (approved):

.circleci/config.yml
```diff
@@ -20,27 +20,55 @@ set_npm_auth: &set_npm_auth
...
+    <<: *defaults
+    <<: *unix_box
+    steps:
+      - checkout
+      - <<: *restore_dependency_cache_unix
+      - run: npm run build
+      - save_cache:
+          key: v9-cache-build-<< pipeline.git.revision >>
+          paths:
+            - axe.js
```
WilcoFiers, line 70: You need to cache `axe.min.js` too, otherwise it won't be published. And then what about `metadata-function-map.js`? Might we need that cached?

.circleci/config.yml
```diff
@@ -335,28 +368,21 @@ workflows:
...
-            - dependencies_unix
-            - test_chrome
-            - test_firefox
-            - test_examples
-            - test_locales
-            - test_virtual_rules
-            - build_api_docs
-            - test_rule_help_version
-            - test_node
             - hold
```
WilcoFiers, line 386: Should we rename this to maybe `hold_release`?

WilcoFiers (changes requested):

.circleci/config.yml
straker: Good point. We don't need to cache the function map as it get's rolled into the built files.

WilcoFiers (approved):

Issue #3329, opened 2021-12-16
Color-contrast does not take into account <slot> elements and the light DOM


tmcconechy:
Product: axe-core && axe Extension
Expectation: Correctly Report the Contrast on a Manual Scan
Actual: False Reports of an Issue
Motivation: This was Previously working

<pre><code>
axe-core version: 4.3.5
extension: 4.19.0
</code></pre>

To reproduce this issue you can use our test site https://main.wc.design.infor.com/ids-radio/
With Axe Latest version use the extension and Scan the Page.

I can see the page is reporting errors on the label custom element that is almost black on a white background. This previously did work ok (noticed in last week or so). I also wonder how the rules are deployed as we have seen other wierd things like some people seeing it and some not. We see the same thing in our tests which run the puppetteer axe.

![Screen Shot 2021-12-16 at 3 04 37 PM](https://user-images.githubusercontent.com/814283/146441670-9a263de8-dd1f-4ff8-83c7-dfb59a099dd9.png)

I wonder if its because the element is a customElement and the text is linked via a slot.

![Screen Shot 2021-12-16 at 3 05 42 PM](https://user-images.githubusercontent.com/814283/146441649-70035d98-31a7-4452-bed0-c3f45558d2dd.png)

nloding:
Hey @tmcconechy - thanks for the report, and thanks for including the URL! I can confirm that I'm seeing the same results you are, and that it does appear to be a bug. I will pass thing along to the correct team members to review.

dylanb:
The issue is being raised as a "needs review" item a.k.a. incomplete, this is expected behavior

straker:
Digging into this issue it appears that our color contrast code doesn't take into account custom elements with `<slot>` correctly. Take the following example which produces a needs review due to overlapping elements:

```
<my-group>
    <my-paragraph>
        Let's have some different text!
    </my-paragraph>
  </my-group>

  <script src="/axe.js"></script>
  <script>
    customElements.define('my-group',
      class extends HTMLElement {
        constructor() {
          super();

          const shadowRoot = this.attachShadow({mode: 'open'});
          shadowRoot.innerHTML = '<div style="position: absolute;"><slot></slot></div>'
        }
      }
    );

    customElements.define('my-paragraph',
      class extends HTMLElement {
        constructor() {
          super();

          const shadowRoot = this.attachShadow({mode: 'open'});
          shadowRoot.innerHTML = '<p><slot>My default text</slot></p>'
        }
      }
[...]

PR #3847, opened 2023-01-05
fix(color-contrast): correctly compute color contrast of <slot> elements

straker:
Closes: #3329

### DIFF
--- a/test/checks/color/color-contrast.js
+++ b/test/checks/color/color-contrast.js
@@ -906,6 +906,29 @@ describe('color-contrast', function () {
     }
   );
 
+  (shadowSupported ? it : xit)('handles <slot> elements', () => {
+    fixture.innerHTML =
+      '<div id="container" style="height: 100px;"><p id="target" style="color: #333;">Slotted text</p></div>';
+    const container = fixture.querySelector('#container');
+    const shadow = container.attachShadow({ mode: 'open' });
+
+    shadow.innerHTML =
+      '<div id="shadowContainer" style="position: absolute; background: black;"><slot></slot></div>';
+    const shadowContainer = shadow.querySelector('#shadowContainer');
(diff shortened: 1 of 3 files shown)


test/checks/color/color-contrast.js
```diff
@@ -906,6 +906,29 @@ describe('color-contrast', function () {
     }
   );
 
+  (shadowSupported ? it : xit)('handles <slot> elements', () => {
+    fixture.innerHTML =
```
straker, line 910: I couldn't use `shadowCheckSetup` because I needed the final node to be the paragraph, but trying to have the paragraph in the first parameter had the code add the shadow dom to it rather than the container element.

test/checks/color/color-contrast.js
```diff
@@ -906,6 +906,29 @@ describe('color-contrast', function () {
...
+      '<div id="container" style="height: 100px;"><p id="target" style="color: #333;">Slotted text</p></div>';
+    const container = fixture.querySelector('#container');
+    const shadow = container.attachShadow({ mode: 'open' });
+
+    shadow.innerHTML =
+      '<div id="shadowContainer" style="position: absolute; background: black;"><slot></slot></div>';
+    const shadowContainer = shadow.querySelector('#shadowContainer');
+    axe.testUtils.flatTreeSetup(fixture);
+
+    const target = fixture.querySelector('#target');
```
straker, line 920: Couldn't use `axe.utils.querySelector` since the element is in the shadow dom in the virtual tree, so would need to select the shadow element and query selector from there.

WilcoFiers (changes requested):
I'm not sure I fully understand the problem. Would like to get on a call to discuss.

WilcoFiers (approved):

PR #3851, opened 2023-01-09
chore: update dependencies

straker:
Updated all dependencies, including examples. I updated the enzyme/jest example to not need the helper utilizies and instead just use `mount` and the jsdom helper script to render it to the dom.


doc/examples/qunit/Gruntfile.js
```diff
@@ -5,7 +5,18 @@ module.exports = function (grunt) {
 
   grunt.initConfig({
     qunit: {
-      all: ['test/**/*.html']
+      all: ['test/**/*.html'],
+      options: {
```
straker, line 9: Ran into this issue https://github.com/gruntjs/grunt-contrib-qunit/issues/158

WilcoFiers (approved):

PR #3855, opened 2023-01-11
tests: refactor is-focusable tests

straker:
Was starting to work on allowing the `inert` attribute to determine when things were focusable and noticed that there were no test files for `is-natively-focusable`, `inserted-into-focus-order` and `focus-disabled`. Turns out most of the tests were all part of the `is-focusable` test code so split them out into their own files. `focus-disabled` didn't have any dedicated tests even inside `is-focusable`, so had to add a bunch. This also meant that I needed to add `focus-disabled` to the `commons` object so I could access it like everything else.

Changes made:
- copied and pasted the `dom.isNativelyFocusable` and `dom.insertedIntoFocusOrder` describes into their own files (with updated es6 syntax for arrow functions and `const` instead of `var)
- created `focus-disabled` tests (all new). Will need verify they cover all cases
- added `focus-disabled` to the `commons` functions

### DIFF
--- a/test/commons/dom/inserted-into-focus-order.js
+++ b/test/commons/dom/inserted-into-focus-order.js
@@ -1,0 +1,211 @@
+describe('dom.insertedIntoFocusOrder', () => {
+  const fixture = document.getElementById('fixture');
+  const { fixtureSetup } = axe.testUtils;
+  const insertedIntoFocusOrder = axe.commons.dom.insertedIntoFocusOrder;
+
+  function hideByClipping(el) {
+    el.style.cssText =
+      'position: absolute !important;' +
+      ' clip: rect(0px 0px 0px 0px); /* IE6, IE7 */' +
+      ' clip: rect(0px, 0px, 0px, 0px);';
+  }
+
+  function hideByMovingOffScreen(el) {
+    el.style.cssText =
+      'position:absolute;' +
+      ' left:-10000px;' +
+      ' top:auto;' +
+      ' width:1px;' +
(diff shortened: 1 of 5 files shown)


test/commons/dom/inserted-into-focus-order.js
```diff
@@ -0,0 +1,211 @@
+describe('dom.insertedIntoFocusOrder', () => {
```
WilcoFiers, line 1: Feel you could probably have combined some of these tests. It works though. Ok if you want to leave it like this.

WilcoFiers (changes requested):
Couple minor points

test/commons/dom/is-natively-focusable.js
```diff
@@ -0,0 +1,285 @@
...
+  it('should return false for an anchor with no href', () => {
+    fixture.innerHTML = '<a name="anchor" id="target"></a>';
+    const el = document.getElementById('target');
+    flatTreeSetup(fixture);
+
+    assert.isFalse(isNativelyFocusable(el));
+  });
+
+  it('should return false for a div with a tabindex with spaces', () => {
+    fixture.innerHTML = '<div id="target" tabindex="0"></div>';
```
WilcoFiers, line 229:
```suggestion
    fixture.innerHTML = '<div id="target" tabindex="    0     "></div>';
```

WilcoFiers (changes requested):
Did you forget to push or something?

WilcoFiers (approved):

PR #3857, opened 2023-01-16
feat(dom.focusDisabled,dom.isVisibleForScreenreader): support the inert attribute

### LINKED ISSUE #3448: Support the HTML inert attribute
In Chrome 102 we're expecting the inert attribute to be supported. This has a number of implications for axe-core. I think we'll need to update a number of internal methods to understand this new attribute. I think since it's not going to be available to a number of other browsers yet a new rule that fails the use of inert unless it's polyfilled is probably also useful here.

straker:
Also adds `dom.isInert`.

Closes #3448


WilcoFiers (approved):

WilcoFiers (changes requested):
Actually, you know what, we should have some integration tests for this. 

I'm also wondering if we should do anything about firefox's lack of support. Thoughts?

WilcoFiers (approved):

Issue #3659, opened 2022-09-12
`color-contrast` fails on `aria-hidden="true"` elements


NickAb:
### Product

axe-core

### Product Version

4.4.3

### Latest Version

- [X] I have tested the issue with the latest version of the product

#### Expectation

Elements marked `aria-hidden="true"` should be skipped by `color-contrast` rule as purely decorative (https://www.w3.org/TR/WCAG22/#contrast-minimum).

#### Actual
`color-contrast` rule fails on elements with low contrast intended to be decorative and marked `aria-hidden="true"`.

#### How to Reproduce

html snippet

```html
<i class="material-icon-i material-icons-extended _ngcontent-agc-196" role="img" aria-hidden="true">brightness_1</i>
```

sample unit test

```js
// test/rule-matches/color-contrast-matches.js

it('should not match aria-hidden=true', function () {
    fixture.innerHTML = '<i aria-hidden="true">hi</i>';
    var target = fixture.querySelector('i');
    axe.testUtils.flatTreeSetup(fixture);
    assert.isFalse(rule.matches(target, axe.utils.getNodeFromTree(target)));
});
```

#### Additional context

https://www.w3.org/TR/WCAG22/#contrast-minimum

> The visual presentation of text and images of text has a contrast ratio of at least 4.5:1, except for the following:
> ...
>
>- Incidental: Text or images of text that are part of an inactive user interface component, that are **pure decoration**, that are not visible to anyone, or that are part of a picture that contains significant other visual content, have no contrast requirement.
> ...

NickAb:
Please let me know if `aria-hidden="true"` is not the correct way to mark purely decorative elements and they should be marked differently according tostandards.

WilcoFiers:
@NickAb That looks like a font icon, not just a text correct? Can you clarify if this is reported as a violation or if it is reported as "needs review"? If you're using the browser extension, it is indicated with a little yellow flag icon.

Something like a font icons do have a minimum contrast. It'd 3:1 though. Axe-core can't definitively determine what's going on in such cases so it lets you know they still need to be reviewed.

straker:
The problem is that there is no standard way to mark a visual element as purely decorative (that I know of). `aria-hidden` doesn't make it decorative, it just removes it from the accessibility tree. Visual users may still need the information. Without something to key off of, axe-core can't make any determinations about what color contrast should or should not run against.

WilcoFiers:
[...]

PR #3867, opened 2023-01-20
fix(color-contrast): skip ligature icons

WilcoFiers:
Closes: #3659

### DIFF
--- a/test/rule-matches/color-contrast-matches.js
+++ b/test/rule-matches/color-contrast-matches.js
@@ -537,4 +537,32 @@ describe('color-contrast-matches', function () {
       assert.isTrue(rule.matches(input, axe.utils.getNodeFromTree(input)));
     });
   }
+
+  describe('with font icons', () => {
+    before(async () => {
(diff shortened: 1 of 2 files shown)


test/rule-matches/color-contrast-matches.js
```diff
@@ -537,4 +537,32 @@ describe('color-contrast-matches', function () {
       assert.isTrue(rule.matches(input, axe.utils.getNodeFromTree(input)));
     });
   }
+
+  describe('with font icons', () => {
```
straker, line 541: Needs a true case

straker (dismissed):

test/rule-matches/color-contrast-matches.js
WilcoFiers: The way I figured it, pretty much every other test in this suite shows that text that isn't an icon font can pass or fail the check. Whatever I add here would basically just be a copy of the default "true" case.

straker (approved):

Issue #3860, opened 2023-01-18
Listbox should not require a name when its used in a combobox pattern


WilcoFiers:
This example should not fail. It comes straight out of WAI-ARIA 1.2:

https://www.w3.org/TR/wai-aria-1.2/#combobox

```html
<label for="tag_combo">Tag</label>
<input type="text" id="tag_combo"
      role="combobox" aria-autocomplete="list"
      aria-haspopup="listbox" aria-expanded="true"
      aria-controls="popup_listbox" aria-activedescendant="selected_option">
<ul role="listbox" id="popup_listbox">
   <li role="option">Zebra</li>
   <li role="option" id="selected_option">Zoom</li>
</ul>
```

I think the fix would be to skip listbox if its part of a combobox pattern in the `no-naming-method-matches` method.

There's an issue open on WAI-ARIA for it: https://github.com/w3c/aria/issues/1769#issuecomment-1386857482

padmavemulapati:
Validated with the latest axe-core develop branch code base,  with the "role=listbox" , it is not needed any name when it is used in a combobox. No `aria` rules failing
Test snippet:
```
<label for="tag_combo">Tag</label>
<input type="text" id="tag_combo"
      role="combobox" aria-autocomplete="list"
      aria-haspopup="listbox" aria-expanded="true"
      aria-controls="popup_listbox" aria-activedescendant="selected_option">
<ul role="listbox" id="popup_listbox">
   <li role="option">Zebra</li>
   <li role="option" id="selected_option">Zoom</li>
</ul>
```


![Image](https://user-images.githubusercontent.com/47409579/217759746-499717a2-cff8-4481-b2b2-23aafaa75398.png)

PR #3886, opened 2023-01-24
fix(aria-input-field-name): skip combobox popups

WilcoFiers:
Closes: #3860

### DIFF
--- a/lib/commons/aria/is-combobox-popup.js
+++ b/lib/commons/aria/is-combobox-popup.js
@@ -1,0 +1,40 @@
+import getExplicitRole from './get-explicit-role';
+import ariaAttrs from '../standards/aria-attrs';
+import closest from '../core/utils/closest';
+import getRootNode from '../core/utils/get-root-node';
+import tokenList from '../core/utils/token-list';
+
+export default function isComboboxPopup(virtualNode) {
+  const popupRoles = ariaAttrs['aria-haspopup'].values;
+  const node = virtualNode.actualNode;
+
+  const role = getExplicitRole(virtualNode);
+  if (!popupRoles.includes(role)) {
+    return false;
+  }
+
+  // in ARIA 1.1 the container has role=combobox
+  if (closest(virtualNode, '[role~="combobox"]')) {
+    return true;
+  }
+
+  // in ARIA 1.0 and 1.2 the combobox owns (1.0) or controls (1.2) the listbox
+  const id = virtualNode.attr('id');
+  if (!id) {
+    return false;
+  }
+
+  const doc = getRootNode(node);
+  const owned = Array.from(
+    doc.querySelectorAll(`[aria-owns~="${id}"], [aria-controls~="${id}"]`)
+  );
+  const comboboxOwned = owned.some(el => {
+    const roles = tokenList(el.getAttribute('role'));
+    return roles.includes('combobox');
+  });
+  if (!comboboxOwned) {
+    return false;
+  }
+
+  return true;
+}
--- a/lib/rules/scrollable-region-focusable-matches.js
+++ b/lib/rules/scrollable-region-focusable-matches.js
@@ -1,75 +1,20 @@
-import { hasContentVirtual } from '../commons/dom';
-import { getExplicitRole } from '../commons/aria';
-import {
-  querySelectorAll,
-  getScroll,
-  closest,
-  getRootNode,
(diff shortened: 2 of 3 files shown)


lib/commons/aria/get-role.js
```diff
@@ -3,7 +3,7 @@ import getImplicitRole from './implicit-role';
 import getGlobalAriaAttrs from '../standards/get-global-aria-attrs';
 import isFocusable from '../dom/is-focusable';
 import { getNodeFromTree } from '../../core/utils';
-import AbstractVirtuaNode from '../../core/base/virtual-node/abstract-virtual-node';
+import AbstractVirtualNode from '../../core/base/virtual-node/abstract-virtual-node';
```
WilcoFiers, line 6: Found a typo

lib/commons/aria/is-combobox-popup.js
```diff
@@ -0,0 +1,55 @@
...
+    `[aria-owns~="${id}"][role~="combobox"]:not(select),
+     [aria-controls~="${id}"][role~="combobox"]:not(select)`
+  );
+
+  return Array.from(ownedCombobox).some(isCombobox);
+}
+
+const isCombobox = node => node && getRole(node) === 'combobox';
+
+function nearestParentWithRole(vNode) {
```
WilcoFiers, line 48: This is not a perfect solution. We should really have a method that can tell you the owner. I spent more time on this PR than I expected to already, and didn't want to bloat this further. I opened a tech debt ticket: #3888

lib/commons/aria/is-combobox-popup.js
```diff
@@ -0,0 +1,55 @@
...
+    return true;
+  }
+
+  const { id } = virtualNode.props;
+  if (!id) {
+    return false;
+  }
+
+  if (!virtualNode.actualNode) {
+    throw new Error('Unable to determine combobox popup without an actualNode');
```
WilcoFiers, line 34: I didn't think we needed additional virtualNode tests, but let me know if I missed anything.

test/rule-matches/no-naming-method-matches.js
```diff
@@ -69,6 +69,24 @@ describe('no-naming-method-matches', function () {
     assert.isFalse(actual);
   });
 
+  it('returns false for the listbox popup of a role=`combobox`', function () {
```
straker, line 72: If the suggestion above about using `popupRoles` is accepted, we should add a test that shows only `listbox` returns false and at least one other popup role returns true.

straker (changes requested):

lib/rules/no-naming-method-matches.js
```diff
@@ -10,14 +11,18 @@ function noNamingMethodMatches(node, virtualNode) {
...
   // Additionally, ignore combobox that get their name from a descendant input:
   if (
     getExplicitRole(virtualNode) === 'combobox' &&
     querySelectorAll(virtualNode, 'input:not([type="hidden"])').length
   ) {
     return false;
   }
+  // Ignore listboxes that are referenced by a combobox
+  // Other roles don't require a name at all, or require one anyway
+  if (isComboboxPopup(virtualNode, ['listbox'])) {
```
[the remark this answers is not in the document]
WilcoFiers: Yikes! Yeah I did.

/me hides in shame.

lib/commons/aria/is-combobox-popup.js
```diff
@@ -0,0 +1,55 @@
...
+     [aria-controls~="${id}"][role~="combobox"]:not(select)`
+  );
+
+  return Array.from(ownedCombobox).some(isCombobox);
+}
+
+const isCombobox = node => node && getRole(node) === 'combobox';
+
+function nearestParentWithRole(vNode) {
+  while ((vNode = vNode.parent)) {
```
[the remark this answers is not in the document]
WilcoFiers: I don't plan on getting #3888 into 4.7, no. Hopefully 4.8. Still, I don't agree this needs to be cached. This isn't going to run much. Pages don't have 100 visible listboxes on them.

test/rule-matches/no-naming-method-matches.js
WilcoFiers: That was supposed to be the next test, but I mixed up `modal` and `dialog`. Modal isn't an ARIA role.

Issue #3728, opened 2022-10-17
Deprecate Bower support


straker:
Bower has been deprecated for awhile now and is no longer being maintained. We should drop our bower support for 5.0.

padmavemulapati:
dev task

PR #3889, opened 2023-01-25
feat: deprecate bower

WilcoFiers:
Since bower has been deprecated for a while, it makes no sense for axe-core to continue support for it. When the next major version of axe-core is released the bowers.json file will be removed.

Closes: #3728

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,7 @@
 {
   "name": "axe-core",
   "version": "4.6.3",
+  "deprecated": true,
   "contributors": [
     {
       "name": "David Sturley",


bower.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "axe-core",
   "version": "4.6.3",
+  "deprecated": true,
```
WilcoFiers, line 4: This is not actually a supported prop. I put something in so that when we get to 5.0 and search for deprecated things to remove, we'll find this.

stephenmathieson (approved):

PR #3894, opened 2023-01-25
fix(color-contrast): do not check contrast for inert elements

straker:
Finishes the work on supporting inert. One question I have is if we want to not check the color contrast of labels (implicit, explicit, etc.) for inputs that are inerted. So this example:

```html
<div id="t1"><span>Test</span></div>
<div role="textbox" aria-labelledby="bob t1 fred" inert></div>
```

We prevent checking this type of setup if the input is disabled (or `aria-disabled`). Do we need to do the same if the input has `inert`?

Closes: #3448


WilcoFiers (approved):

straker:
We determined not to worry about inert inputs and their labels for now. If someone brings up an issue about it we'll reconsider.

Issue #3307, opened 2021-11-21
Update allowance of aria-checked on HTML checkbox and radios


scottaohara:
ARIA in HTML (w3c/html-aria#372) is tightening the allowance of `aria-checked` on native checkboxes and radios, particularly due to w3c/aria#1622 and the fact that `aria-checked=mixed` should never be able to be exposed by browsers that correctly follow the ARIA spec.

we would request that this at the very least this be called out as an author error if using `aria-checked=mixed` on a checkbox, and either an author error or a warning to indicate that authors should not use this attribute on native checkboxes and radios.

For instance, there is a presently bug in chromium where `aria-checked` incorrectly takes precedent over the native checked / dirty checked state of an radio or checkbox element, which can cause inaccurate information to be exposed if the actual checked state and the aria-checked attribute are out of alignment: https://bugs.chromium.org/p/chromium/issues/detail?id=1260371

WilcoFiers:
Interesting, yeah that makes sense. We'll see what we can do with that.

dylanb:
Can we expand the rule to check that the aria semantics match the checkbox semantics and only complain when they don't match until the Chromium bug is fixed and then we switch it to always complain?

WilcoFiers:
So, what I think I'm going to do here is to prohibit when aria-checked contradicts the true state, which because mixed isn't possible on checkboxes prohibits mixed on checkboxes.

scottaohara:
Agree with not failing when the aria-checked attribute and whatever checked state the input is actually set to don't contradict.  

Not sure I understand why you'd just prohibit the true state though?  Or that mixed isn't possible on checkboxes?  No attribute for the latter, but it's definitely possible per the `indeterminate` idl attribute and check for the indeterminate state/pseudo selector. 

One could very well have aria-checked=false on an actually checked checkbox.  or aria-checked=mixed on an unchecked checkbox.  maybe i misunderstood what you were saying though?

WilcoFiers:
Well, looks like I have a gap in my HTML knowledge! I had never heard of indeterminate checkboxes. Alright, I can handle that. But then, you say

> One could very well have aria-checked=false on an actually checked checkbox.

It would seem to me that because of that Chrome issue, having `aria-checked=false` on a `checked` checkbox creates an inconsistency between browsers. Seems to me that could cause very significant accessibility issues. That seems like something to should definitely fail.

WilcoFiers:
[...]

PR #3895, opened 2023-01-26
feat(aria-allowed-attr): no inconsistent aria-checked on HTML checkboxes

WilcoFiers:
Because I want to add explicit messages without confusing the missing messages of the aria-allowed-attr check I decided to create a new `aria-conditional-attr` check. I moved existing code that checks that row attributes that are allowed on treegrid but not on table or grid into that new check too.

I put the checkbox code and row code into separate evaluate methods, that then get invoked by ariaConditionalAttrEvaluate. Alternatively they could have been different checks entirely. I'm not keen on adding a new check every time we have a role for which we want some extra logic, that felt excessive. Then again there's only two of them so maybe they could have just been two checks? I'm open to changing this.

Closes: #3307


lib/checks/aria/aria-allowed-attr-evaluate.js
```diff
@@ -1,7 +1,6 @@
-import { uniqueArray, closest, isHtmlElement } from '../../core/utils';
+import { uniqueArray, isHtmlElement } from '../../core/utils';
 import { getRole, allowedAttr, validateAttr } from '../../commons/aria';
 import { isFocusable } from '../../commons/dom';
-import cache from '../../core/base/cache';
```
WilcoFiers, line 4: I didn't think this cache was useful. If we ever want to cache getRole, it should be memoized. And unlike this current implementation I'm only looking up the table role once per row, not for every attribute.

lib/checks/aria/aria-allowed-attr-evaluate.js
```diff
@@ -30,62 +29,36 @@ import cache from '../../core/base/cache';
...
-  ariaAttr.forEach(attr => {
-    preChecks[attr] = validateRowAttrs;
-  });
-  if (allowed) {
-    for (let i = 0; i < attrs.length; i++) {
-      const attrName = attrs[i];
-      if (validateAttr(attrName) && preChecks[attrName]?.()) {
-        invalid.push(attrName + '="' + virtualNode.attr(attrName) + '"');
-      } else if (validateAttr(attrName) && !allowed.includes(attrName)) {
-        invalid.push(attrName + '="' + virtualNode.attr(attrName) + '"');
```
WilcoFiers, line 75: This was really weird and messy. Took some time refactoring and cleaning this up. Much easier to grok now. The gist of what this does is that if you're using one of these `validTreeRowAttrs` on a table or grid, it should fail because those attrs are only allowed on treegrid.

lib/checks/aria/aria-conditional-attr-evaluate.js
```diff
@@ -0,0 +1,20 @@
...
+export default function ariaConditionalAttrEvaluate(
+  node,
+  options,
+  virtualNode
+) {
+  const role = getRole(virtualNode);
+  if (!conditionalRoleMap[role]) {
+    return true;
+  }
+  return conditionalRoleMap[role].call(this, node, options, virtualNode);
```
WilcoFiers, line 19: I'm not sure if this was over-engineering or not. Having two new checks would be fine, but then if we wanted to add more that becomes kind of bloated. These things are mutually exclusive because they run for different roles.

lib/rules/aria-allowed-attr.json
```diff
@@ -7,7 +7,7 @@
     "description": "Ensures ARIA attributes are allowed for an element's role",
     "help": "Elements must only use allowed ARIA attributes"
   },
-  "all": [],
-  "any": ["aria-allowed-attr"],
+  "all": ["aria-allowed-attr", "aria-conditional-attr"],
```
WilcoFiers, line 10: I really don't like "none" checks.  Always messes with my head that `false` means `pass`. Can't undo that on the other checks but I don't want to make it worse.

lib/checks/aria/aria-conditional-checkbox-attr-evaluate.js
```diff
@@ -0,0 +1,30 @@
...
+    return false;
+  }
+
+  return true;
+}
+
+const xor = (a, b) => (a || b) && !(a && b);
+
+const isChecked = vNode =>
+  vNode.actualNode ? !!vNode.actualNode.checked : vNode.hasAttr('checked');
```
WilcoFiers, line 30: Using `checked` because if you toggle a checkbox it doesn't update the `checked` attribute.

WilcoFiers:
Back to draft, need to handle indeterminate checkboxes.

straker (changes requested):

lib/checks/aria/aria-conditional-row-attr-evaluate.js
```diff
@@ -0,0 +1,36 @@
+import getRole from '../../commons/aria/get-role';
+import { closest } from '../../core/utils';
+
+export default function ariaConditionalRowAttr(
+  node,
+  { invalidTableRowAttrs } = {},
+  virtualNode
+) {
+  const invalidAttrs =
+    invalidTableRowAttrs?.filter?.(invalidAttr => {
```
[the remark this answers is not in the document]
WilcoFiers: Well it SHOULD be an array, but since its configurable we can't really be certain. This felt slightly safer.

lib/checks/aria/aria-conditional-row-attr-evaluate.js
```diff
@@ -0,0 +1,36 @@
...
+
+  const messageKey = `row${invalidAttrs.length > 1 ? 'Plural' : 'Singular'}`;
+  this.data({ messageKey, invalidAttrs, ownerRole });
+  return false;
+}
+
+function getRowOwner(virtualNode) {
+  // check if the parent exists otherwise a TypeError will occur (virtual-nodes specifically)
+  if (!virtualNode.parent) {
+    return true;
```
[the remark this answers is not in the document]
WilcoFiers: Fair! This was a boolean function that I ended up rewriting to find the parent. Should be `null`. We always use `null` for missing nodes.

test/checks/aria/aria-conditional-attr.js
```diff
@@ -0,0 +1,265 @@
...
+            );
+            assert.isFalse(
+              ariaConditionalCheck.call(checkContext, null, null, vNode)
+            );
+            assert.deepEqual(checkContext._data, {
+              messageKey: 'checkbox',
+              checkState: 'mixed'
+            });
+          }
+        });
```
WilcoFiers, line 261:
```suggestion
        for (const prop of ['true', 'TRUE', 'false', 'invalid']) {
          const vNode = asIndeterminateVirtualNode(
            `<input type="checkbox" aria-checked="${prop}" id="target">`
          );
          assert.isFalse(
            ariaConditionalCheck.call(checkContext, null, null, vNode)
          );
          assert.deepEqual(checkContext._data, {
            messageKey: 'checkbox',
            checkState: 'mixed'
          });
        }
```

straker (approved):

PR #3912, opened 2023-02-13
chore(origin): fix CDP Page.setDocumentContent origin null crashing a…

j-mendez:
* fix issue with  test hanging on allowedOrigins unhandled exception when using CDP `Page.setDocumentContent`

--- Notes

The following issue occurs when using axe-core to test HTML markup created over  CDP `Page.setDocumentContent` or `puppeteer.setContent`. 

```
[0213/121447.712165:INFO:CONSOLE(15)] "Uncaught (in promise) Error: allowedOrigins value "null" is not a valid origin", source: pptr://__puppeteer_evaluation_script__ (15)
```

Example test showing the crash repro https://github.com/a11ywatch/a11y/blob/main/tests/basic-axe.ts.
You also need to comment out the following block https://github.com/a11ywatch/a11y/blob/main/lib/runners/axe.ts#L16 that allowing test to run after setting the `origin` manually.

### DIFF
--- a/lib/core/base/audit.js
+++ b/lib/core/base/audit.js
@@ -18,11 +18,12 @@ const dotRegex = /\{\{.+?\}\}/g;
 function getDefaultOrigin() {
   // @see https://html.spec.whatwg.org/multipage/webappapis.html#dom-origin-dev
   // window.origin does not exist in ie11
-  if (window.origin) {
+  // prevent origin default "null" string on CDP `Page.setDocumentContent` https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-setDocumentContent
+  if (window.origin && window.origin !== "null") {
     return window.origin;
   }
   // window.location does not exist in node when we run the build
-  if (window.location && window.location.origin) {
+  if (window.location && window.location.origin && window.location.origin !== "null") {
     return window.location.origin;
   }
 }


straker:
@j-mendez did you mean to close this? I was looking into this pr and was able to validate the bug and the fix. Was just about to ask for a test if that was possible (was seeing if it was possible to change `window.origin` and `window.locaiton.origin` to `'null'` in a test env without using CDP)

j-mendez:
@straker Hi, I actually did not mean to close down the PR. Changing the 'null' for a test seems adequate as well. Is there a certain location or file for the test additions?

straker:
Alight, talked with @WilcoFiers on how best to test this. What we think we'd like is a new example in the `docs/examples/puppeteer` directory that uses puppeteers `page. setContent` function to create the page with an iframe. Then trying to run `axe` on that page should work (meaning you get results back without an error). That should allow the code to be tested and prevent regressions.

Something like this (you'll need to update the example `pacakge.json` `test` script to also run the new file).

```js
// doc/examples/puppeteer/set-content.js
const assert = require('assert');

const browser = await puppeteer.launch();
const page = await browser.newPage();
const axeSource = await fs.readFile('./axe.js', 'utf8');

await page.setContent(`
  <html lang="en">
  <head>
    <title>Test Page</title>
  </head>
  <body>
    <main>
      <h1>Hello World</h1>
      <button id="empty-button"></button>
      <iframe title="iframe" srcdoc="<button id='iframe-empty-button'></button>"></iframe>
    </main>
  </body>
  </html>
`);

await page.evaluate(axeSource);
const frames = page.frames();
for (let i = 0; i < frames.length; i++) {
  await frames[i].evaluate(axeSource);
}

const results = await page.evaluate(`window.axe.run()`);
assert(results.violations.length)
```

j-mendez:
@straker feel free to drive this one home, currently do not have the bandwidth.

j-mendez:
@straker you can use the GitHub cli to checkout the pr or git.

j-mendez:
> Wo, didn't know the `gh` command could do that. Thanks for the reference.

no problem, it should be `gh pr checkout 3912` for this PR.

straker:
So now that I've made the changes, how do I push back to this pr? I don't see a `push` command for prs in `gh`, and there's not a branch for this pr techincally.

j-mendez:
> So now that I've made the changes, how do I push back to this pr? I don't see a `push` command for prs in `gh`, and there's not a branch for this pr techincally.

I think ‘gh pr push’, not sure only since I have only recently used it to checkout a PR for testing.

straker:
Hmm, didn't work. This is an interesting problem.

j-mendez:
> Hmm, didn't work. This is an interesting problem.

Ahh, I’m not at the computer right now. Feel free to merge the changes locally to another branch and re push the PR or etc and thank you for taking it over!

straker:
Moved pr to https://github.com/dequelabs/axe-core/pull/3921.

PR #3921, opened 2023-02-22
fix(respondable): work with CDP `Page.setDocumentContent`

straker:
Using Google Devtools Protocol to set the page content using `Page.setDocumentContent` would set the `origin` and `location.origin` to the string `"null"`. This allows respondable to skip posting to origins of `"null"`.

Original pr: https://github.com/dequelabs/axe-core/pull/3912

### DIFF
--- a/doc/examples/puppeteer/package.json
+++ b/doc/examples/puppeteer/package.json
@@ -4,10 +4,10 @@
   "private": true,
   "main": "axe-puppeteer.js",
   "scripts": {
-    "test": "node axe-puppeteer.js https://deque.com"
+    "test": "node axe-puppeteer.js https://deque.com && node set-content.js"
   },
   "devDependencies": {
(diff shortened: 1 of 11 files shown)


doc/examples/puppeteer/package.json
```diff
@@ -4,10 +4,10 @@
   "private": true,
   "main": "axe-puppeteer.js",
   "scripts": {
-    "test": "node axe-puppeteer.js https://deque.com"
+    "test": "node axe-puppeteer.js https://deque.com && node set-content.js"
```
WilcoFiers, line 7: Nitpick; Maybe put this into two tests jobs? `test:url` and `test:set-content`?

WilcoFiers (dismissed):

WilcoFiers (approved):

Issue #3929, opened 2023-03-02
Color-contrast issue with element stack


straker:
Our grid code seems to miscalculate the element stack levels. Take the following HTML:

```html
<div id="opacity-parent" style="opacity: 0.9;">
  <div id="z-index" style="position: relative; z-index: 2">
   <h1 id="target">Hello World</h1>
  </div>
</div>
<div id="opacity-oos" style="opacity: 0.8;">
  <div id="absolute" style="position: absolute; top: 20px; z-index: -1;">
    <div id="red" style="height: 40px; width: 100vw; background: red"></div>
  </div>
</div>
```

Which results in a stack of:

```
[
  h1#target
  div#z-index
  div#opacity-parent
  div#red
  div#absolute
  main
  body
  html
]
```

Which is incorrect as the red background should be on top of the target h1. The stacking order our code produces is:

```
[
  [0,2,0],
  [0,2,0],
  [0,0],
  [0,-1,0],
  [0,-1,0],
  [0],
  [0],
  [0]
]
```

From the [stacking order code](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/dom/create-grid.js#L251), it looks like we need to create a stacking layer first before applying z-index. That would allow the opacity to create it's own stacking context before the positioned elements change it based on z-index.

PR #3930, opened 2023-03-03
fix(create-grid): correctly compute stack order for non-positioned stacking contexts

straker:
While working on the code, I realized that the spec said that if _any_ element that creates a stacking context it would remove the fake stacks created from float or positioned elements that use auto or 0 z-index. So I was able to merge those functions into a single if statement. I also simplified the loop to remove fake stacks so it only needs to look through the array and `splice` once (since any new stack removes all previous fake stacks).

Closes: #3929

### DIFF
--- a/lib/commons/dom/create-grid.js
+++ b/lib/commons/dom/create-grid.js
@@ -302,7 +302,7 @@ function createStackingOrder(vNode, parentVNode, nodeIndex) {
   while (float.length < 10) {
     float = '0' + float;
   }
-  stackingOrder.push(DEFAULT_ORDER + parseFloat('0.0' + float));
+  stackingOrder.push(parseFloat(`${DEFAULT_ORDER}${float}`));
 
   return stackingOrder;
 }
--- a/lib/commons/dom/create-grid.js
+++ b/lib/commons/dom/create-grid.js
@@ -49,6 +49,7 @@ export default function createGrid(
 
   // IE11 requires the first 3 parameters
   // @see https://developer.mozilla.org/en-US/docs/Web/API/Document/createTreeWalker
+  let nodeIndex = 0;
   const treeWalker = document.createTreeWalker(
     root,
     window.NodeFilter.SHOW_ELEMENT,
@@ -79,7 +80,7 @@ export default function createGrid(
       vNode = new axe.VirtualNode(node, parentVNode);
     }
 
-    vNode._stackingOrder = createStackingOrder(vNode, parentVNode);
+    vNode._stackingOrder = createStackingOrder(vNode, parentVNode, nodeIndex++);
 
     const scrollRegionParent = findScrollRegionParent(vNode, parentVNode);
     const grid = scrollRegionParent ? scrollRegionParent._subGrid : rootGrid;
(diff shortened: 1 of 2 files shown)


lib/commons/dom/create-grid.js
```diff
@@ -255,26 +255,41 @@ function getStackingOrder(vNode, parentVNode) {
...
+    // if an element creates a stacking context, find the first
     // true stack (not a "fake" stack created from positioned or
     // floated elements without a z-index) and create a new stack at
     // that point (step #5 and step #8)
     // @see https://www.w3.org/Style/css2-updates/css2/zindex.html
-    while (stackingOrder.find(value => value % 1 !== 0)) {
-      const index = stackingOrder.findIndex(value => value % 1 !== 0);
-      stackingOrder.splice(index, 1);
+    const index = stackingOrder.findIndex(
+      value => (value < 1 && value > 0.2) || value === 0
```
WilcoFiers, line 265: These magic numbers are difficult to understand without studying the code. I switched them over to constants in my refactor PR. Please have a look

WilcoFiers (changes requested):
1. I tried to refactor this function a bit, in order to help myself understand it better, please use that: https://github.com/dequelabs/axe-core/pull/3932/files

2. I think we should move getStackingOrder into its own file. It's only tangentially related to creating the grid. I think separating these two could make things a little clearer.

Overall, I can't seem to keep this all straight in my head. I hope I didn't miss anything, but I'm not sure.

test/commons/dom/get-element-stack.js
```diff
@@ -399,6 +399,25 @@ describe('dom.getElementStack', () => {
...
+        <div id="3" style="opacity: 0.8;">
+          <div id="4" style="position: absolute; top: 20px; z-index: -1;">
+            <div id="5" style="height: 40px; width: 100vw; background: red"></div>
+          </div>
+        </div>
+      `;
+      axe.testUtils.flatTreeSetup(fixture);
+      const target = fixture.querySelector('#target');
+      const stack = mapToIDs(getElementStack(target));
+      assert.deepEqual(stack, ['5', '4', 'target', '2', '1', 'fixture']);
```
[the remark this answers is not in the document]
straker: Because it has an absolutely positioned child so has 0 width/height.

lib/commons/dom/create-grid.js
```diff
@@ -255,26 +255,41 @@ function getStackingOrder(vNode, parentVNode) {
...
 
-  // flex and grid items can use z-index even if position: static
-  // @see https://www.w3.org/TR/css-flexbox-1/#painting
-  // @see https://www.w3.org/TR/css-grid-1/#z-order
-  if (
-    !['auto', '0'].includes(zIndex) &&
-    (positioned || isFlexOrGridContainer(parentVNode))
-  ) {
-    // if a positioned element has a z-index > 0, find the first
+  if (isStackingContext(vNode, parentVNode)) {
```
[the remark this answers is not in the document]
straker: It's because the grid creates nodes and parents for elements that aren't in the virtual tree, but never assigns parents unless it creates a new node. So if `axe.run(#fixture)`, `#fixture` will still have a null parent but for the grid the parent element is created and passed around.

lib/commons/dom/create-grid.js
```diff
@@ -248,50 +253,68 @@ function isFlexOrGridContainer(vNode) {
...
+    stackingOrder.push(parseInt(zIndex));
+    return stackingOrder;
+  }
+  // since many things can create a new stacking context without position or
+  // z-index, we need to know the order in the dom to sort them by. Use the
+  // nodeIndex property to create a number less than the "fake" stacks from
+  // positioned or floated elements but still larger than 0
+  // 10 pad gives us the ability to sort up to 1B nodes (padStart does not
+  // exist in ie11)
+  let float = vNode.nodeIndex.toString();
```
WilcoFiers, line 298: Can we rely on this? I was under the impression that we were always building the entire tree, but since you say that isn't the case, wouldn't we run into problems here if we were to compare a node that's in the tree with one that isn't?

straker: Changed it to use a nodeIndex from looping over the treewalker, which is guaranteed to get everything in DOM order.

WilcoFiers (approved):
Would have preferred to see createStackingOrder move into its own file, but

Issue #3900, opened 2023-01-27
Font icon detection results in Canvas2D warning


WilcoFiers:
I've been getting this warning when running axe-core. I think this happened when we added checking for font icons to `color-contrast-matches`:

```
Canvas2D: Multiple readback operations using getImageData are faster with the willReadFrequently attribute set to true
```

dbjorge:
@WilcoFiers Yes, confirming that this is coming from font-icon checking - I noticed this in `label-content-name-mismatch` recently as well

PR #3931, opened 2023-03-04
fix(is-icon-ligature): prevent canvas2d warning willReadFrequently

Rrroman:
Removes the console warning related to font-icon checking and improves the speed of multiple readback operations, as mentioned in the warning.

> Canvas2D: Multiple readback operations using getImageData are faster with the willReadFrequently attribute set to true. See: https://html.spec.whatwg.org/multipage/canvas.html#concept-canvas-will-read-frequently

Closes: [3900](https://github.com/dequelabs/axe-core/issues/3900)


straker (approved):

straker:
Thanks for the pr! Approved for security

Issue #3924, opened 2023-02-24
Color contrast miscalculates foreground color with opacity


straker:
Given the following HTML:

```html
<style>
button {
  color: #fff;
  background: #00633D;
  border-color: #00633D;
  opacity: .65;
}
</style>
<button><span>Hello</span></button>
```

The foreground color is calculated as `#c5dcd3`, which is not quite correct. If we put the calculated color next to the rendered HTML, the calculated foreground color is not the same.

<img alt="" src="https://user-images.githubusercontent.com/2433219/221213323-48d7a1cd-dcc4-41f7-bc90-362554763931.png"/>

I'll need to dig into the color spec again to see how opacity works on text nodes as it appears it should use the background behind the element with opacity instead of blending into all backgrounds.

padmavemulapati:
Validated with the latest develp brranch code base,
for the code snippet:
```
<style>
    button {
      color: #fff;
      background: #00633D;
      border-color: #00633D;
      opacity: .65;
    }
    </style>
     <button><span>Hello</span></button>
     <br>
     <br>
     <div id="target" style="height:20px; width:15px; color:rgb(0 0 128); background-color:green;">Hello World</div>
   ```
1. foregrond color (text color)is calculated as- `#FFFFFF` and the background color as `#6A9A81` then it is passing at only for `wcagaa` - `Large Text,UI Components, & Graphical Objects (3:1)` and remaining all are failing for SC - "Regular Text(1.4.3a)"

<img width="692" alt="image" src="https://user-images.githubusercontent.com/47409579/230814445-8058b9ac-e8b6-4964-9f0a-14162660531b.png">

2. When this border-color color is - `#6A9A81` and background color also same then "Regular Text(1.4.3a) fialing in all the onditions.
![image](https://user-images.githubusercontent.com/47409579/230815022-ab512b89-1003-4ff6-a248-1b0c8376dc31.png)

3. for the color:rgb(0 0 128); background-color:green; also 1.4.3a failing on all the conditions.
![image (2)](https://user-images.githubusercontent.com/47409579/230815141-b3575e4f-887c-4a48-a8b1-1a64afe472ac.png)

PR #3944, opened 2023-03-14
fix(color-contrast): correctly compute background color for elements with opacity

straker:
This is the 2nd part of fixing #3924. With this I have the ability to grab the stacking context for the foreground element and then compute the separate background colors where the opacity is applied so I can compute the text color to the background behind the text's element stack.


lib/commons/color/stacking-context.js
```diff
@@ -0,0 +1,112 @@
...
+  const stackingContext = [];
+  const contextMap = new Map();
+  elmStack = elmStack ?? getBackgroundStack(elm);
+
+  elmStack.forEach(bgElm => {
+    const bgVNode = getNodeFromTree(bgElm);
+    const bgColor = getOwnBackgroundColor(bgVNode);
+    const stackingOrder = bgVNode._stackingOrder.filter(({ vNode }) => !!vNode);
+
+    stackingOrder.forEach(({ vNode }, index) => {
```
straker, line 21: The final structure of this code looks like this.

```html
<div id="parent" style="background-color: red; opacity: 0.8">
  <div id="target" style="background-color: rgba(0,255,0,0.5)">Text</div>
<div>
```

```js
[
  { 
    vNode: #parent,
    opacity: 0.8,
    blendMode: 'normal',
    bgColor: Color(255,0,0,1),
   descendants: [
     { 
        vNode: #target,
        opacity: 1,
        blendMode: 'normal',
        bgColor: Color(0,255,0,0.5),
       descendants: []
      }
   ]
  }
]
```

lib/commons/color/get-background-color.js
```diff
@@ -47,59 +48,57 @@ export default function getBackgroundColor(
...
   if (bgColors.length) {
     bgColors = [{ color: bgColors.reduce(flattenShadowColors) }];
   }
 
-  const elmStack = getBackgroundStack(elm);
-  const textRects = getVisibleChildTextRects(elm);
-
   // Search the stack until we have an alpha === 1 background
-  (elmStack || []).some(bgElm => {
+  for (let i = 0; i < elmStack.length; i++) {
```
straker, line 63: By turning this into a loop instead of a `some` we can return early inside the loop rather than trying to return after the `some` has already gone through all elements of the stack.

lib/commons/dom/create-grid.js
```diff
@@ -267,11 +267,11 @@ function createStackingOrder(vNode, parentVNode, nodeIndex) {
   if (!isStackingContext(vNode, parentVNode)) {
     if (vNode.getComputedStylePropertyValue('position') !== 'static') {
       // Put positioned elements above floated elements
-      stackingOrder.push(POSITION_STATIC_ORDER);
+      stackingOrder.push(createContext(POSITION_STATIC_ORDER, vNode));
```
straker, line 270: Adding virtual node info to the stacking order allows me to use that virtual node to create the stacks for the node.

lib/commons/color/color.js
```diff
@@ -1,68 +1,7 @@
...
-    colors = [high, 0, low];
-  }
-
-  return colors
-    .map(color => {
-      return Math.round((color + base) * 255);
-    })
-    .concat(alpha);
-}
+const hexRegex = /^#[0-9a-f]{3,8}$/i;
```
straker, line 3: I took this opportunity to do something I've been wanting to do for awhile now which is convert the `Color` constructor function into a true class. The advantage to this was that I was able to do `deepEquals` on two Color objects whereas before that was not possible (thus greatly simplifying the tests). Another advantage is that all Color objects now share memory for all their functions thanks to the prototype whereas before each Color object had their own memory instance for each function.

WilcoFiers (dismissed):

lib/commons/color/stacking-context.js
```diff
@@ -0,0 +1,205 @@
...
+    const color = context.bgColor;
+    color.alpha *= context.opacity;
+
+    return {
+      color,
+      blendMode: context.blendMode
+    };
+  }
+
+  const fgColor = context.descendants.reduce(
```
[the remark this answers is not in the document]
straker: The spec calls these ["source color" and "backdrop color"](https://www.w3.org/TR/compositing-1/#simplealphacompositing) so we should update the terms in our code to match (same for the parameter names for `flattenColors`)

lib/commons/color/stacking-context.js
```diff
@@ -0,0 +1,205 @@
...
+ */
+export function getStackingContext(elm, elmStack) {
+  const vNode = getNodeFromTree(elm);
+  if (vNode._stackingContext) {
+    return vNode._stackingContext;
+  }
+
+  const stackingContext = [];
+  const contextMap = new Map();
+  elmStack = elmStack ?? getBackgroundStack(elm);
```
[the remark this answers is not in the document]
straker: We'll want to do some performance testing of `getBackgroundStack` to see if it's an expensive call that warrants caching the result for when foreground color needs it.

WilcoFiers (approved):

Issue #3649, opened 2022-09-07
False positives when using aria-roledescription


tesk9:
### Product

axe-core

### Product Version

4.4.3

### Latest Version

- [X] I have tested the issue with the latest version of the product

#### Expectation
`aria-roledescription` should be allowed on any element that has "a valid WAI-ARIA role or has an implicit WAI-ARIA role," according to [ARIA 1.1's roledescription section](https://www.w3.org/TR/wai-aria-1.1/#aria-roledescription)

#### Actual
I believe that the [checks for `aria-roledescription`](https://github.com/dequelabs/axe-core/blob/develop/lib/checks/aria/aria-roledescription.json#L5) limit the supported WAI-ARIA roles to:
```
"supportedRoles": [
      "button",
      "img",
      "checkbox",
      "radio",
      "combobox",
      "menuitemcheckbox",
      "menuitemradio"
    ]
```
which leaves out a fair number of roles, including the "region" role that's used in the [ARIA 1.1. example of a "slide"](https://www.w3.org/TR/wai-aria-1.1/#aria-roledescription).

#### How to Reproduce
Provide a code sample or link to a webpage that reproduces the issue. Without this, your issue may be closed without investigation.

Add the following HTML node to a page:
`<mark aria-roledescription="Claim highlight">Bats are vital pollinators</mark>`

Run axe devtools or any of the other tools that use axe-core, and you'll get an error:
<img width="514" alt="Screen Shot of axe devtools output with Issue 'Ensure aria-roledescription is only used on elements with an implicit or explicit role' highlighting `mark[aria-roledescription='Claim highlight']`" src="https://user-images.githubusercontent.com/8811312/188992164-f0aaa14e-9605-4fe4-a241-123fc54be4a1.png">

straker:
Thanks for the issue. When we created the check two years ago, [assistive technologies didn't support it very well](https://github.com/dequelabs/axe-core/issues/1216#issuecomment-454104190) for non-interactive elements. That may have changed in the past years, so we can recheck support.

straker:
After rechecking, the attribute is fully supported so we'll disable the rule and deprecate it..

tesk9:
Thanks!

padmavemulapati:
Validated with the latest axe-core develop branch code base, `aria-roledescription` is deprecated , so that with the latest `axe.run` is not showing `aria-roledescription` with the below code snippet:
```
[...]

PR #3948, opened 2023-03-20
feat(aria-roledescription): deprecate and disable rule

straker:
After testing, the `aria-roledescription` attribute seems to be fully supported* in the various browsers / screen readers. We can remove this rule now, so will do so in the next major release.

Closes: #3649

* There are a few inconsistencies between browsers, but not enough to keep the rule.


WilcoFiers (approved):

straker:
@kensgists We currently do not have another rule that tests if `aria-roledescription` is used correctly (used on a valid role). #3955 is set up to track validating usage of ARIA roledescription and braille attrs.

kensgists:
@straker Appreciate the information. Will track the progress of that.

Issue #3850, opened 2023-01-07
False positive in aria-required-children "unknown" path for child not in accessibility tree with aria-hidden="true"


dbjorge:
### Product

axe-core

### Product Version

4.6.2

### Latest Version

- [X] I have tested the issue with the latest version of the product

#### Expectation

The following pattern should not result in an `aria-required-children` failure:

```
<div role="menu">
  <div role="menuitem">menu item</div>
  <div aria-hidden="true">shouldn't be flagged but is</div>
</div>
```

#### Actual

This results in a failure with the following `failureSummary`:

```
Fix any of the following:
  Element has children which are not allowed (see related nodes)
  Element has no aria-busy="true" attribute
```

#### How to Reproduce
[Minimal repro](https://codepen.io/dbjorge/pen/NWBbmKL) (+ an example where the same case without `aria-hidden="true"` does not trigger the failure)

#### Additional context

Per #3462, the "unknown" failure path introduced in `aria-required-children` in #3597 attempts to detect cases similar to [this ACT rule](https://www.w3.org/WAI/standards-guidelines/act/rules/ff89c9/proposed/)'s "failed example 3", where a child **in the accessibility tree** (emphasis mine) does not have a role that matches one of the requiredOwned roles for the parent.

However, we think the logic introduced in #3597 to determine whether a child is "in the accessibility tree" for the purposes of this rule is a bit off - it does so by asking "does the child have any global ARIA attribute OR is it focusable". However, `aria-hidden` is a global ARIA attribute which currently triggers the first clause of this check, which is inconsistent with the rule. [Per the glossary of the ACT rule in question](https://www.w3.org/WAI/standards-guidelines/act/rules/ff89c9/proposed/#included-in-the-accessibility-tree), an element which is programmatically hidden should be treated as "not included in the accessibility tree" for the purposes of this check:

> ### Included in the accessibility tree
>
> [Programmatically hidden](https://www.w3.org/WAI/standards-guidelines/act/rules/ff89c9/proposed/#programmatically-hidden) elements are removed from the accessibility tree. However, some browsers will leave [focusable](https://www.w3.org/WAI/standards-guidelines/act/rules/ff89c9/proposed/#focusable) elements with an aria-hidden attribute set to true in the accessibility tree. Because they are hidden, these elements are considered not included in the accessibility tree. This may cause confusion for users of assistive technologies because they may still be able to interact with these focusable elements using sequential keyboard navigation, even though the element should not be included in the accessibility tree.

[...]

PR #3949, opened 2023-03-20
fix(aria-required-children): do not fail for children with aria-hidden

straker:
Closes: #3850


lib/commons/dom/index.js
```diff
@@ -40,7 +40,7 @@ export { default as isNode } from './is-node';
 export { default as isOffscreen } from './is-offscreen';
 export { default as isOpaque } from './is-opaque';
 export { default as isSkipLink } from './is-skip-link';
-export { default as isVisibleToScreenReaders } from './is-visible-for-screenreader';
+export { default as isVisibleToScreenReaders } from './is-visible-to-screenreader';
```
straker, line 43: I noticed that the file names were still using `for` instead of `to` so changed those to match the name of the function while I was there.

lib/checks/aria/aria-required-children-evaluate.js
```diff
@@ -15,6 +20,10 @@ function getOwnedRoles(virtualNode, required) {
   const ownedElements = getOwnedVirtual(virtualNode);
   for (let i = 0; i < ownedElements.length; i++) {
     const ownedElement = ownedElements[i];
+    if (ownedElement.props.nodeType !== 1) {
```
straker, line 23: Needed to ignore non-element nodes as `isVisibletoScreenreaders` calls into `isHiddenForEveryone` which tries to get the computed style for `display`, which fails for non-element nodes

straker:
```html
<ul>
  <div aria-hidden="true">foo</div>
  <li>bar</li>
</ul>
```

passes the `list` rule, and

```html
<dl>
  <span aria-hidden="true">foo</span>
  <dt>bar</dt>
  <dd>baz</dd>
</dl>
```

passes the `definition-list` rule.

ilantom:
Has this fix been released? I upgraded axe-core to 4.7.2 and I still the issue. Please let me know.

straker:
@ilantom This should have been fixed in 4.7.1 I believe. Could you provide a code sample of it failing? It was validated for the original issue https://github.com/dequelabs/axe-core/issues/3850#issuecomment-1486468420.

ilantom:
I was using code in https://developer.microsoft.com/en-us/fluentui#/controls/web/detailslist/customitemcolumns

straker:
@ilantom I exported the code to codepen and don't see any `aria-required-children` violations

screenshot of violations for the page showing empty-table-header, image-alt, landmark-one-main, page-has-heading-one, and region

ilantom:
You are right - I'm unable to repro the issue after exporting the code to codepen but I see the issue here:

image

same for my site

straker:
Is your site public that I could take a look at it? Running axe-core on the microsoft page itself still does not show a violation for `aria-required-children`. However the microsoft page has a bunch of CORS and CSP issues when trying to run axe-core so I'm not exactly confident in the final results.

PR #3954, opened 2023-03-23
chore: Updating aria-busy error message.

### LINKED ISSUE #3934: Clarify aria-busy remediation message
We had some confusion about this message:

```
Element has no aria-busy="true" attribute
```

This doesn't explain when aria-busy would be appropriate. I think the following would be better:

```
Element uses aria-busy="true" while showing a loader
```

This required a small change to `lib/checks/aria/aria-busy.json`.

1Copenut:
Changed the `aria-busy` fail message per guidance in issue #3934.

Closes:
This PR closes #3934.


WilcoFiers (approved):
All good. Thanks for the PR @1Copenut!

WilcoFiers:
Reviewed for security.

Issue #3892, opened 2023-01-25
Update duplicate-id rules to align with WCAG 2.2


WilcoFiers:
In WCAG 2.2, success criterion 4.1.1 has been removed. This raises lots of questions on how to handle that in axe-core. Here's my proposal on how this should effect axe-core. I'm looking for feedback on this approach:

### `duplicate-id-aria`

This rule fails when a duplicate ID is used in ARIA, or in form labelling. The result of it can be that controls have incorrect names, or that other relationships are incorrect. Per the standard, only the first of an ID in a page is valid. So if the first ID happens to be the correct one then there's no accessibility problem. If not, then there is almost certainly some issue, but without SC 4.1.1 to report to it is difficult to say what it actually is. This rule reports `serious` issues.

- Replace the `wcag2a` tag with a `wcag2a-obsolete` tag.
- Add the `best-practice` tag, so that this rule will only run when best practices are turned on

**Future**: We've been talking about possibly replacing this rule with a more focused `idref-labels-and-aria` rule that doesn't just look at duplicates, but considers whether based on DOM order that's likely to cause an accessibility problem. As part of resolving this issue we should have a proposal of that rule written.

### `duplicate-id`

It is a technical failure of WCAG SC 4.1.1 Parsing, but in practice it has no impact on accessibility. Axe-core reports these as `minor` issues.

- Deprecate and disable the `duplicate-id` rule by default. This rule will not run unless explicitly configured, and will be removed in axe-core 5.0.
- Replace the `wcag2a` tag with a `wcag2a-obsolete` tag.

### `duplicate-id-active`

This rule was created to catch a problem with Dragon Naturally Speaking, where if active controls like buttons shared an ID, Dragon could activate the first element with that ID, even if you asked it to activate the second. Axe reports this as a `serious` issue. After some testing we've discovered **this problem no longer exists**. So for this rule we'll deprecate and tag it the same way we're doing for `duplicate-id`.

dbjorge:
I agree 100% with your proposals for `duplicate-id` and `duplicate-id-active`. We already disable `duplicate-id` in Accessibility Insights because it generates a lot of noise with no practical impact on accessibility. I think your assessment of `duplicate-id-active` makes sense.

[...]

PR #3957, opened 2023-03-23
feat(rules): deprecate duplicate-id, duplicate-id-active, and duplicate-id-aria rules

straker:
Closes: #3892


WilcoFiers (changes requested):
This PR title is a bit misleading in suggesting duplidate-id-aria gets deprecated. How about we put that one in its own PR?

WilcoFiers (approved):

WilcoFiers (approved):

Issue #3920, opened 2023-02-22
css-orientation-lock rule should support CSS rotate


WilcoFiers:
Our ACT nightly tests started failing after the CSS orientation rule added examples with the CSS `rotate` prop:

https://app.circleci.com/pipelines/github/dequelabs/axe-core/4921/workflows/53584a64-11a1-4bb5-bd24-872f02c1cece/jobs/54868

The following example should fail axe-core:

```html
<html lang="en">
<head>
  <style>
    @media (orientation: portrait) {
      html { rotate: 90deg; }
    }
  </style>
</head>
<body>
  Page Content
</body>
</html>
```

WilcoFiers:
Here's the PR that changes this in ACT: https://github.com/act-rules/act-rules.github.io/pull/1742/files

padmavemulapati:
Valdiated with the latest axe-core develop branch code base, with the below code snippet:
```

  <p id="fail1" class="fail1">This is my first paragraph with this ID.</p>
<div style="display: none">
  <p id="fail1">This is my second paragraph with this ID.</p>
</div>
<head>
  <style>
    @media (orientation: portrait) {
      html { rotate: 90deg; }
    }
  </style>
</head>
<body>
  Page Content
</body>
```
`css-orientation-lock` will fail 


![Image](https://user-images.githubusercontent.com/47409579/230264544-a4367d95-e54a-469c-8583-c36cc44c998e.png)

PR #3958, opened 2023-03-23
fix(css-orientation-lock): support the css rotate property

straker:
Closes: #3920

### DIFF
--- a/lib/checks/mobile/css-orientation-lock-evaluate.js
+++ b/lib/checks/mobile/css-orientation-lock-evaluate.js
@@ -103,21 +103,31 @@ function cssOrientationLockEvaluate(node, options, virtualNode, context) {
 
     const transformStyle =
       style.transform || style.webkitTransform || style.msTransform || false;
-    if (!transformStyle) {
+    if (!transformStyle && !style.rotate) {
       return false;
     }
+
+    let transformFn;
+    let transformFnValue;
 
     /**
      * get last match/occurrence of a transformation function that can affect rotation along Z axis
      */
-    const matches = transformStyle.match(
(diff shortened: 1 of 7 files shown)


lib/checks/mobile/css-orientation-lock-evaluate.js
```diff
@@ -103,21 +103,31 @@ function cssOrientationLockEvaluate(node, options, virtualNode, context) {
 
     const transformStyle =
       style.transform || style.webkitTransform || style.msTransform || false;
-    if (!transformStyle) {
+    if (!transformStyle && !style.rotate) {
       return false;
     }
 
+    let transformFn;
+    let transformFnValue;
```
WilcoFiers, line 111: Nitpick; I suggest pulling this code out into its own function.

```suggestion
    const { transformType, transformValue } = getCssTransform(style);
```

lib/checks/mobile/css-orientation-lock-evaluate.js
```diff
@@ -103,21 +103,31 @@ function cssOrientationLockEvaluate(node, options, virtualNode, context) {
...
+
     /**
      * get last match/occurrence of a transformation function that can affect rotation along Z axis
      */
-    const matches = transformStyle.match(
-      /(rotate|rotateZ|rotate3d|matrix|matrix3d)\(([^)]+)\)(?!.*(rotate|rotateZ|rotate3d|matrix|matrix3d))/
-    );
-    if (!matches) {
-      return false;
+    if (transformStyle) {
```
WilcoFiers, line 116: We should probably account for someone using both `transform` and `rotate`. I can well imagine someone using them together, for example using transform to adjust the position of an element and rotate to turn it.

WilcoFiers (changes requested):
We should also update wcag-act-rules to the latest to prove this passes the latest ACT tests.

WilcoFiers (approved):

PR #3973, opened 2023-04-03
fix(color-contrast): correcly apply opacity to foreground color

straker:
Closes: #3924

### DIFF
--- a/lib/commons/color/get-foreground-color.js
+++ b/lib/commons/color/get-foreground-color.js
@@ -3,7 +3,6 @@ import getBackgroundColor from './get-background-color';
 import incompleteData from './incomplete-data';
 import flattenColors from './flatten-colors';
 import getTextShadowColors from './get-text-shadow-colors';
-import { getNodeFromTree } from '../../core/utils';
 import { getStackingContext, stackingContextToColor } from './stacking-context';
 
 /**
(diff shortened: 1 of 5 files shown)


lib/commons/color/get-foreground-color.js
```diff
@@ -21,42 +21,49 @@ import { getNodeFromTree } from '../../core/utils';
...
   if (bgColor === null) {
     const reason = incompleteData.get('bgColor');
     incompleteData.set('fgColor', reason);
     return null;
   }
-  fgColor.alpha = fgColor.alpha * opacity;
-  return flattenColors(fgColor, bgColor);
+
+  const stackingContexts = getStackingContext(node);
+  const context = findContext(stackingContexts, node);
```
WilcoFiers, line 61: How about this?

```suggestion
  const context = findNodeInContexts(stackingContexts, node);
```

lib/commons/color/get-foreground-color.js
```diff
@@ -83,26 +90,75 @@ function getStrokeColor(nodeStyle, { textStrokeEmMin = 0 }) {
...
+
+/**
+ * Find the stacking context that belongs to the passed in node
+ * @param {Object} contexts - Array of stacking contexts
+ * @param {Element} node
+ * @returns {Object}
+ */
+function findContext(contexts, node) {
+  for (let i = 0; i < contexts.length; i++) {
+    const context = contexts[i];
```
WilcoFiers, line 154: Is there a reason not to do this?

```suggestion
  for (const context of contexts) {
```

WilcoFiers (dismissed):
LGTM. Two minor suggestions for you though.

WilcoFiers (approved):

PR #3976, opened 2023-04-05
ci: fix nightly apg tests

straker:
The apg tests weren't running due to the github repo completely changing structure in https://github.com/w3c/aria-practices/pull/2417. I updated the paths to the new structure and was able to remove a lot of the skipped rules for specific pages as well. I also set the tags to only run wcag21aa and below so we don't get best practice rules or aaa rules running. 

Lastly, I wanted to pin the repo to their last release, but they changed the structure after that release so instead just pinned to the latest commit (as it would be a lot of spaghetti code to try to maintain two directory structures for the test).

### DIFF
--- a/test/aria-practices/apg.spec.js
+++ b/test/aria-practices/apg.spec.js
@@ -6,12 +6,15 @@ const { assert } = require('chai');
 const globby = require('globby');
 
 describe('aria-practices', function () {
-  // Use path.resolve rather than require.resolve because APG has no package.json
+  // Use path.resolve rather than require.resolve because APG package.json main file does not exist
   const apgPath = path.resolve(__dirname, '../../node_modules/aria-practices/');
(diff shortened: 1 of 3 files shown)


test/aria-practices/apg.spec.js
```diff
@@ -32,56 +35,40 @@ describe('aria-practices', function () {
...
+        await driver.get(`${addr}content/patterns/${filePath}`);
 
-        const builder = new AxeBuilder(driver, axeSource);
-        builder.disableRules([
-          ...disabledRules['*'],
-          ...(disabledRules[filePath] || [])
-        ]);
+        const builder = new AxeBuilder(driver, axeSource)
+          // Support table has no title and has duplicate ids
+          .exclude('#at-support')
```
WilcoFiers, line 59: Yikes! Fair enough though.

WilcoFiers (approved):

Issue #3897, opened 2023-01-26
Update axe-core release policy


WilcoFiers:
We're going to release more frequently, with less noticed (not 6 weeks). The I'll need to update the documentation to reflect that.

padmavemulapati:
Will take QA accordingly and follw the releases

PR #3983, opened 2023-04-12
docs(release-and-support): Adjust  release cadence

WilcoFiers:
- Remove wording suggesting a major release every 12 to 18 months. Major releases are done as needed. We only want to do them when they are necessary
- Minor releases have a 3 week lead-time instead of 6
- Added a section on the use of milestones and code freeze
- Fix a couple typos

Closes: #3897

### DIFF
--- a/doc/release-and-support.md
+++ b/doc/release-and-support.md
@@ -2,7 +2,7 @@
 
 ## Release Cadence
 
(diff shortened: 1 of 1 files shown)


straker (approved):
Minor nit, but nothing that would stop a merge. Take as you will.

straker (approved):

PR #3994, opened 2023-04-19
fix(aria-allowed-attrs): add aria-expanded to allowed attrs for menuitemcheckbox and menuitemradio

### LINKED ISSUE #3992: False positive aria-allowed-attr for aria-expanded on menuitemcheckbox and menuitemradio
### Product

axe-core

### Product Version

_No response_

### Latest Version

#### Expectation
`aria-expanded` is an allowed attribute on both menuitemcheckbox and menuitemradio (ref: https://w3c.github.io/aria/#menuitemcheckbox).

#### Actual
axe-core currently returns an aria-allowed-attr failure when it is added to either role.

#### How to Reproduce
Simplest repro: https://jsfiddle.net/fu9ve7gp/show

#### Additional context
Looks like the issue is here: https://github.com/dequelabs/axe-core/blob/develop/lib/standards/aria-roles.js#L377-L396

#3343 would fix it along with some other `aria-expanded` support mismatches, but since it's over a year old I wasn't sure if it was still active.

smhigley:
Fixes #3992

Adds `aria-expanded` to allowed attributes for `menuitemcheckbox` and `menuitemradio`, and updates the `passes.html` aria-allowed-attr tests for that attribute on those two roles.

Past context: #3343


straker:
Awesome, thanks for this. I noticed #3343 added `aria-expanded` to more roles, such as checkbox and switch. Do you think those should be part of this pr or a separate pr?

straker (approved):

straker:
Reviewed for security

Issue #3896, opened 2023-01-26
Deprecate all "virtual" commons


WilcoFiers:
When we were working on axe-core 3.0 we imagined we'd have different methods for DOM nodes and virtual DOM nodes. The later would be called with methods ending with `Virtual`. Because of that we ended up with methods like `accessibleText` and `accessibleTextVirtual`. We've moved away from this approach more recently. More and more of axe runs exclusively on VirtualNodes. It's faster, safer, lets us normalise more consistently, etc. It's far better. Where DOM nodes are needed we can easily get at them from `vNode.actualNode`.

So, I would like us to find all `xxxVirtual` methods, make sure they're non-virtual counterpart can accept virtual nodes, then deprecate all those `xxVirtual` methods, and remove their use from the code base.

straker:
I looked into what it would take to do this and there's a slight snag in that the virtual function is where most of the code for the function is and the non-virtual function is just a pass through to the virtual one. For example, all `accessibleText` does is get the virtualNode from the tree and then calls `accessibleTextVirtual`. This means we can't just deprecate `accessibleTextVirtual` since that's where all the code is.

Instead what I propose we do is do a bit of file renaming and transfer the virtual function into the non-virtual name. So rename the `accessibleTextVirtual` to `accessibleText` and do the same for the current `accessibleText` (renaming it to `accessibleTextVirtual`), then we can deprecate the "new" virtual function. Do this with the git rename command `git mv`. The reason I propose the work this way is in order to keep the commit and version history of the file that contains all the code. That way we can still do blame and see the history of what's changed. If we instead just moved the code from the virtual function into the non-virtual function we would lose all that once the old virtual function file is deleted.

PR #3998, opened 2023-04-27
refactor: add util function to normalize node argument

straker:
In preparation for #3896 I decided that instead of needing to import two different things (`AbstractVirtualNode` and `getNodeFromTree`) and copying and pasting the same if statement, we should instead have a helper function that does the things for us. So that's what I did.


lib/core/utils/normalize-node.js
```diff
@@ -0,0 +1,22 @@
+import AbstractVirtualNode from '../base/virtual-node/abstract-virtual-node';
+import getNodeFromTree from './get-node-from-tree';
+
+/**
+ * Get the virtual node and actual node from a parameter that could be either.
+ *
+ * @param {VirtualNode|Element} element
+ * @return {Object} The Virtual Node and actual node
+ */
+export default function normalizeNode(element) {
```
WilcoFiers, line 10: I'm not sure normalize is the right term for this. How about `nodeLookup()`?

lib/core/utils/normalize-node.js
```diff
@@ -0,0 +1,22 @@
...
+  if (element instanceof AbstractVirtualNode) {
+    return {
+      vNode: element,
+      node: element.actualNode
```
WilcoFiers, line 14: I would like us to get a little more consistent about what we're calling different nodes. I very much regret "actualNode" as a term. `domNode` would have been much better. How about this:

1. We barely use virtualNode anymore, I think that's okay so lets stick with vNode
2. When we talk about a DOM node, lets call it that; domNode, domParent, whatever.
3. When something can be either a domNode or a virtualNode we just use `node`. That's only for passing this in

```suggestion
      domNode: element.actualNode
```

lib/commons/aria/arialabel-text.js
```diff
@@ -8,14 +7,11 @@ import { getNodeFromTree } from '../../core/utils';
...
  * @return {string} ARIA label
  */
-function arialabelText(vNode) {
-  if (!(vNode instanceof AbstractVirtualNode)) {
-    if (vNode.nodeType !== 1) {
-      return '';
-    }
-    vNode = getNodeFromTree(vNode);
+export default function arialabelText(element) {
+  const { vNode } = normalizeNode(element);
```
WilcoFiers, line 11: Bummer decorators aren't available yet. I'd love to have this work:

```suggestion
@nodeLookup
export default function arialabelText({ vNode }) {
```

WilcoFiers (dismissed):
Had some ideas, but we should probably discuss before you make changes.

WilcoFiers (approved):

PR #4009, opened 2023-05-05
perf(aggregate): add fast oneshot aggregate

j-mendez:
# Whats Changed

**Major** performance improvement for `aggregate` handling.

The aggregate util simply needs the map value at the last index. Iterating to get that value and breaking the loop as quick as possible to return brings huge performance boosts since prior the solution was cloning and iterating over the array multiple times. Expect to see performance increase around 1/3 - 1/4 baseline for a medium real world test case.

The logic for initial does not handle mapping to the array `map` param for the custom test cases.

TODO:

For the initial custom cases we need to convert the array to an object to map accordingly, we can do this when we notice the initial index of the array is not any of the predefined `criteriaMap` values. We could also look into removing the custom mapping since it does not seem to be used in the actual core beyond making the `aggregate` method a bit more dynamic.

-- Edit

~~Pr was done on phone, can look into the formatter later and the failing test. May have mapped one of the values incorrectly.~~

### DIFF
--- a/lib/core/utils/aggregate.js
+++ b/lib/core/utils/aggregate.js
@@ -20,25 +20,30 @@ const criteriaMap = {
  */
 function aggregate(map, values, initial) {
   const valSize = values.length;
+
+  if (valSize === 1) {
+    return map[criteriaMap[values[0]]];
+  }
+
   const maxPoint = valSize / 2;
-  const tailEnd = valSize ? valSize - 1 : 0;
+  const tailEnd = valSize - 1;
   let mapIndex = initial ? criteriaMap[initial] : 0;
 
   if (mapIndex !== 3) {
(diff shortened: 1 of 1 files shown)


j-mendez:
I did the PR on my phone from the github app. I see that theres formatting issues, I can take a look later on how to run the formatter. The testing was done on another project for the changes. Thanks for the library!

j-mendez:
Hi, @WilcoFiers what are your thoughts on the 4 test case failures here https://app.circleci.com/pipelines/github/dequelabs/axe-core/5125/workflows/63f9ff11-221e-48f5-91fd-b2739b8c57bd/jobs/56923. The test uses the method in a way that the current core is not to my knowledge. We can add this fix if needed or remove the custom test case?

PR #4011, opened 2023-05-08
chore(import): fix memoize import color-contrast-evalulate

j-mendez:
The memoize is exported as a default module. This PR fixes the import on `color-contrast-evaluate`. Theres some other errors with the import in other files.


j-mendez:
This memoize is not being imported from the default root that handles the re-export. Using typescript helps with finding errors like this.

PR #4013, opened 2023-05-09
perf(lang): add built in language validation

j-mendez:
# What Changed

1. add native language validation handling - better on performance.
1. drastic reduce on bundle size.
1. remove unused validLangs util.
1. remove grunt language file building.

Displaying the changes since the dif is large.

```js
function isValidLang(lang) {
  let valid = true;
 
  try {
    valid = !!Intl.Collator.supportedLocalesOf(lang, {
      localeMatcher: 'lookup'
    }).length;
  } catch (err) {
    valid = false;
  }
 
  return valid;
}
```

Before:

```sh
./axe.min.js
Size:           459.51 Kb
Gzip Size:      117.55 Kb
```

After:
```sh
./axe.min.js
Size:           436.26 Kb
Gzip Size:      114.19 Kb
```
-- Notes

Some of my PR's may close upon deleting the forks on my github. If a pr is closed, double check for items added.
A lot of the logic in the js could be done natively by using browser API methods. If older IE support is needed a bundler should be used to bundle that certain script instead of keeping the core bloated.

— Edit

In order to support all locales we can add the list of matches that are not in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale. A script can pull in the locales once to add in the fallback runtime adding collations. Languages are set in for speaking, not sure if we need a script to manage it beyond a one time thing since languages are being removed instead of added over the years.


lib/core/utils/valid-langs.js
```diff
@@ -61,45 +6,12 @@ const langs = [,[,[1,1,1,1,1,1,1,1,1,1,,1,1,1,1,1,1,1,,1,1,1,1,1,1,,1],[1,1,1,1,
...
-    lang += '`';
-  }
-  for (let i = 0; i <= lang.length - 1; i++) {
-    const index = lang.charCodeAt(i) - 96;
-    array = array[index];
-    if (!array) {
-      return false;
-    }
+  try {
+    // @ts-ignore getCanonicalLocales not in draft but, available on all main browsers
```
j-mendez, line 10: This sets up the project for future typescript changes. A lot of unused modules in the library adding some weight that ts could eliminate during development.

PR #4026, opened 2023-05-19
perf(get-element-stack): improve getElementStack performance on large pages

qazwsxedcrfvtgb1111:
This pr improves the performance of getElementStack on large pages by caching the results of getScroll.
Tested with axe-extension keyboard igt on this page: https://bop.nv.gov/Board/BoardMtgs/
Saves ~200-400 ms on every getElementStack call


straker (dismissed):
Thanks for the pr. I can see how caching `getScroll` would greatly improve performance when the grid is created as we call it for every ancestor in `findScrollRegionParent`. A better way to do the caching would be to wrap the function using `memoize` instead of the using the internal cache. You can see an example of that in [get-cell-position](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/table/get-cell-position.js).

Amandeque:
Hi @straker Do we have a target Axe-core release and timeline, this PR will be the part of ? A customer is waiting to get this related [issue](https://github.com/dequelabs/axe-extension/issues/3518) fixed.

Issue #4027, opened 2023-05-22
Why is aria-required on input with type="range" a critical issue?


cstrobbe:
### Product

axe-core

### Question

When an `input` element has `type="range"` (i.e. a slider input), the presence of the attribute `aria-required="true"` triggers the Axe rule [Elements must only use allowed ARIA attributes](https://dequeuniversity.com/rules/axe/4.7/aria-allowed-attr). This rule has the [impact level "Critical"](https://github.com/dequelabs/axe-core/blob/develop/doc/issue_impact.md) (the highest level), i.e. "Results in blocked content for people with disabilities, and will definitely prevent them from accessing fundamental features or content."

I understand that incorrect roles can be blocking in certain cases, but why would  `aria-required="true"` be blocking on a slider / range input? I also understand that marking a slider input as required makes little sense, since the input always has a default value, even without user interaction. In fact, the HTML 5 specification even [prohibits the native attribute `required` on a slider](https://html.spec.whatwg.org/#range-state-(type=range)). But that still does not explain why  `aria-required="true"` would be a blocking issue on a range input.

Is this really a blocking issue or does the rule "Elements must only use allowed ARIA attributes" catch too many things?

straker:
Thanks for the issue. It looks like the [slider role](https://www.w3.org/TR/wai-aria-1.2/#slider) does not allow `aria-required` (slider being the [implicit role of a range input](https://www.w3.org/TR/html-aria/#el-input-range)), but the [switch role](https://www.w3.org/TR/wai-aria-1.2/#switch) does. So axe-core is correct according to spec for the OP, but incorrect for the followup comment. 

We'd love a pr for the `role=switch` issue. You'd just have to update the [aria-roles file for the switch role](https://github.com/dequelabs/axe-core/blob/develop/lib/standards/aria-roles.js#L686) to add `aria-required` to the list of `allowedRoles`, then update the [aria-allowed-attrs integration file](https://github.com/dequelabs/axe-core/blob/develop/test/integration/rules/aria-allowed-attr/passes.html#L1900-L1925) to add `aria-required` to the `role=switch` element.

straker:
Talked to @WilcoFiers about `slider` not allowing `aria-required`. We decided that even if the spec doesn't allow it axe-core should since it won't cause any problems for screen readers or users. We'll also make sure other input type roles do the same since having code that generates the `type` dynamically can easily generate this problem.

PR #4028, opened 2023-05-22
Add 'aria-required' to switch role

mikeybinns:
Add 'aria-required' to switch role

Related: #4027


mikeybinns:
Something odds going on with my commit author, I'm gonna close and resubmit

PR #4029, opened 2023-05-22
fix(aria-allowed-attr): Add 'aria-required' to switch role

mikeybinns:
Related: #4027


straker:
Approved for security

PR #4034, opened 2023-05-23
fix(aria-allowed-attr): allow aria-required on role=slider

straker:
Tested in VO/Safari, JAWS/Chrome, and NVDA/Firefox.

Closes: #4027


straker:
Going to open this again with the correct email for license.

PR #4035, opened 2023-05-23
fix(aria-allowed-attr): allow aria-required on role=slider

straker:
Tested in VO/Safari, JAWS/Chrome, and NVDA/Firefox.

Closes: https://github.com/dequelabs/axe-core/issues/4027


WilcoFiers (approved):
Assuming you looked at other input-like roles that ARIA may not have omitted this on.

PR #4050, opened 2023-06-12
test: work around firefox 115 text-stroke-width regression

dbjorge:
This PR fixes a test failure against FirefoxNightly in our nightly CI build. Specifically, it works around a regression introduced in Firefox 115 tracked by https://bugzilla.mozilla.org/show_bug.cgi?id=1837692 where Firefox started clamping `-webkit-text-stroke-width` values to integer numbers of pixels. This broke test case `uses stroke color when thickness is equal to the minimum` in `get-foreground-color` because it involved verifying a test fixture that used a stroke width of `.2em` against a font-size of `16px` (ie, `3.2px`).

To address this, this PR pins that test case to use `16px` font-size and `.25em` stroke width, which resolves to an integer number of pixels.

To avoid the possibility of a similar issue being introduced in the future, I also updated all test cases in this file to pin font-size explicitly (to the currently-used value of `16px`) everywhere that stroke width is customized. @straker is separately investigating a more robust version of that fix to pin `font-size` (and probably some other "base" CSS) in our tests to a fixed value at a higher level than per-test, but that'll be addressed in a separate PR - this is just a minimal workaround to unblock our nightly test runs.

### DIFF
--- a/test/commons/color/get-foreground-color.js
+++ b/test/commons/color/get-foreground-color.js
@@ -56,7 +56,7 @@ describe('color.getForegroundColor', () => {
   describe('text-stroke', () => {
     it('ignores stroke when equal to 0', () => {
       const target = queryFixture(
-        '<div style="color: rgb(0 0 128); -webkit-text-stroke: 0 #CCC" id="target">Hello World</div>'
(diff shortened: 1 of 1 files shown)


dbjorge:
Closing as unnecessary - firefox already pushed a fix to nightly, which unblocks our CI build

PR #4053, opened 2023-06-13
test: fix implicit test dependencies on local dev environment

dbjorge:
This PR fixes a few test failures I noticed when running `npm run test:unit` and `npm run test:integration` locally on my dev machine, which is running MacOS 13.4 with the system appearance preference set to dark mode. Before this PR, the integration tests wouldn't run and there were 11 bogus unit test failures on my local dev machine; after this PR, they are consistent with CI and pass.

The main issues this PR fixes are:
* `npm run start` used `http-server` without explicitly setting its `-a` argument, which means that it listens only on IPv4 localhost interfaces, not IPv6. This caused `npm run test:integration` to get stuck in `start-server-and-test`'s health check of the localhost server, which tried to connect on `::1` by default. Adding `-a ""` works around this. See https://github.com/http-party/http-server/issues/832 and https://github.com/http-party/http-server/pull/833/files for context.
* The CSS that we inherit from Mocha's `mocha.css` in our unit and integration tests uses a `prefers-color-scheme` media query to change the test page's background color according to system color scheme preference. However, many test cases in the `get-foreground-color` and `get-background-color` unit tests make assumptions about color blending that rely on the default light-scheme background color. This PR adds `before`/`after` hooks to those test files to normalize to the `#fff` background the tests assume. These hooks intentionally set the normalized color in `beforeEach` (to ensure each test case has an consistent starting state) and reset to the original color in `after` rather than `afterEach` (to limit the amount of unnecessary color flashing).
* The `dialog` integration test attempted to induce a color-contrast failure by overriding only `background-color`, but in dark mode, the new background color didn't actually cause a failure on its own. Updated that test to also override `color` to ensure the intended failure appears regardless of dark mode settings.

This PR also included a few refactorings that didn't turn out to be necessary to the fixes, but seemed nice enough and separate enough from the other changes to include. They are isolated to their own commits; feel free to review commit-by-commit to see them in isolation from the functional changes.

@straker and I originally explored an alternative for this that involved adding our own `/test/global-styles.css` that would override `mocha.css` and force a consistent background/text color across all tests all the time, ignoring prefers-color-scheme. I didn't end up going with this strategy because it conflicted too much with the test status messages that mocha includes on the page during `npm run test:debug` execution; in dark mode, it resulted in a lot of unreadable white-on-white text.

Closes: n/a

### DIFF
--- a/test/commons/color/get-background-color.js
+++ b/test/commons/color/get-background-color.js
@@ -7,6 +7,37 @@ describe('color.getBackgroundColor', function () {
   var origBodyBg;
   var origHtmlBg;
 
+  /**
+   * Assert that two Colors are equal.
+   * @param {axe.commons.color.Color[]} actual
+   * @param {axe.commons.color.Color[]} expected
+   */
+  function assertColorsEqual(actual, expected) {
+    assert.equal(actual.red, expected.red, 'red');
+    assert.equal(actual.green, expected.green, 'green');
+    assert.equal(actual.blue, expected.blue, 'blue');
+    assert.equal(actual.alpha, expected.alpha, 'alpha');
+  }
+
+  /**
+   * Assert that two Colors are close-to-equal.
+   * @param {axe.commons.color.Color[]} actual
+   * @param {axe.commons.color.Color[]} expected
+   * @param {number} threshold How much each RGB value may differ by
+   * @param {number} alphaThreshold How much the alpha channel may differ by
+   */
+  function assertColorsClose(
+    actual,
+    expected,
+    threshold = 0.5,
+    alphaThreshold = 0.1
+  ) {
+    assert.closeTo(actual.red, expected.red, threshold, 'red');
+    assert.closeTo(actual.green, expected.green, threshold, 'green');
+    assert.closeTo(actual.blue, expected.blue, threshold, 'blue');
+    assert.closeTo(actual.alpha, expected.alpha, alphaThreshold, 'alpha');
+  }
+
   before(function () {
     origBodyBg = document.body.style.background;
     origHtmlBg = document.documentElement.style.background;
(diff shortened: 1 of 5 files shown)


test/commons/color/get-background-color.js
```diff
@@ -7,19 +7,59 @@ describe('color.getBackgroundColor', function () {
   var origBodyBg;
   var origHtmlBg;
 
+  /**
+   * Assert that two Colors are equal.
+   * @param {axe.commons.color.Color[]} actual
+   * @param {axe.commons.color.Color[]} expected
+   */
+  function assertColorsEqual(actual, expected) {
```
straker, line 15: We actually don't need this anymore. In [a recent pr](https://github.com/dequelabs/axe-core/pull/3944#discussion_r1137758374) I updated `axe.commons.color.Color` to be a true class so we could do straight deep equal comparisons between two colors

```js
assert.deepEqual(actual, expected)
```

test/commons/color/get-background-color.js
```diff
@@ -7,19 +7,59 @@ describe('color.getBackgroundColor', function () {
...
+    document.documentElement.style.background = 'unset';
+  });
 
+  afterEach(function () {
     axe.commons.color.incompleteData.clear();
     axe._tree = undefined;
   });
 
+  after(function () {
+    document.body.style.background = origBodyBg;
```
straker, line 59: This shouldn't be necessary anymore either as the [global `afterEach` function](https://github.com/dequelabs/axe-core/blob/develop/test/testutils.js#L601) from testutils removes the `style` attribute from body

straker (changes requested):
Just a few small things

test/commons/color/get-background-color.js
dbjorge: Per offline discussion, added `window.chai.config.truncateThreshold = 0;` to `testutils.js` in concert with this update to ensure assertion failure output actually contains all the properties of the color if an assertion failure occurs

test/commons/color/get-background-color.js
dbjorge: Updated the global afterEach to also clear `<html>` styles similarly to `<body>` ones

straker (approved):

PR #4055, opened 2023-06-15
fix(aria-allowed-role): add meter to allowed roles for named img

### LINKED ISSUE #4054: Update allowed roles (meter) for img element
Presently the img element allows the role of `progressbar` to be set to it.  Similarly, the role `meter` should also be allowed.

ARIA in HTML is being updated to indicate this is an allowed role.

scottaohara:
closes #4054

as stated in the PR title, this adds the meter element to be an allowed role for the img element when it is provided an accessible name.

### DIFF
--- a/test/integration/rules/aria-allowed-role/aria-allowed-role.html
+++ b/test/integration/rules/aria-allowed-role/aria-allowed-role.html
@@ -246,7 +246,7 @@
 <div id="fail-dpub-4" role="doc-noteref">ok</div>
 <!-- images -->
 <div id="fail-dpub-5" role="doc-cover">ok</div>
-<img src="#" role="meter" aria-valuenow="0" aria-valuemin="0" aria-valuemax="100" alt="test">
+<img src="#" role="meter" aria-valuenow="0" aria-valuemin="0" aria-valuemax="100" alt="test" id="pass-img-valid-role-meter">
 <img role="doc-cover" aria-label="foo" id="pass-img-valid-role-aria-label" />
 <img role="menuitem" title="bar" id="pass-img-valid-role-title" />
 <div id="image-baz">hazaar</div>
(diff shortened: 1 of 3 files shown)


straker (changes requested):
Thanks for the pr. Would you be able to add a test for this is [test/integration/aria-allowed-role.html](https://github.com/dequelabs/axe-core/blob/develop/test/integration/rules/aria-allowed-role/aria-allowed-role.html#L247-L265). Just adding an image tag with the new role and and id that will need to be added to the companion json file pass section.

scottaohara:
added

straker:
Thanks. The element will need an `id` attribute and then the `id` will need to be added to the companion [aria-allowed-role.json](https://github.com/dequelabs/axe-core/blob/develop/test/integration/rules/aria-allowed-role/aria-allowed-role.json#L4) file to the `passes` section (that should resolve the tests).

test/integration/rules/aria-allowed-role/aria-allowed-role.html
```diff
@@ -246,6 +246,7 @@ <h1 id="pass-h1-valid-role" role="none"></h1>
 <div id="fail-dpub-4" role="doc-noteref">ok</div>
 <!-- images -->
 <div id="fail-dpub-5" role="doc-cover">ok</div>
+<img src="#" role="meter" aria-valuenow="0" aria-valuemin="0" aria-valuemax="100" alt="test">
```
scottaohara, line 249:
```suggestion
<img src="#" role="meter" aria-valuenow="0" aria-valuemin="0" aria-valuemax="100" alt="test" id="pass-img-valid-role-meter">
```

straker (approved):
Awesome! Thanks for doing that. 

Reviewed for security.

Issue #4051, opened 2023-06-13
Add EN301.549 tags


WilcoFiers:
As this is an important international standard, it is useful to add tags for  EN 301 549. I think we can do that very similar to how we do WCAG tags. Each rule that maps to EN 301 549 gets an `EN.301549` tag, and a second tag referencing the section. These are all part of section 9, so for `wcag111` that would be `EN.9.1.1.1`

One open question here is what to do with WCAG 2.2. That will likely be part of of EN before the end of the year. But since there's a transition period before it becomes required, I'm not sure whether we should tag that rule as soon as it's in EN, or if we should wait with that.

PR #4063, opened 2023-06-23
feat: add EN.301.549 tags to rules

WilcoFiers:
- [x] Update tags
- [x] Update docs
- [x] Add tag validation

Closes: #4051

- Make new tags available for DQU documentation (after merging)


build/tasks/validate.js
```diff
@@ -307,5 +301,170 @@ function validateRule({ tags, metadata }) {
...
+  {
+    name: 'WCAG',
+    standardRegex: /^wcag2(1|2)?a{1,3}$/,
+    criterionRegex: /^wcag\d{3,4}$/
+  },
+  {
+    name: 'Section 508',
+    standardRegex: /^section508$/,
+    criterionRegex: /^section508\.\d{1,2}\.[a-z]$/,
+    wcagLevelRegex: /wcag2aa?/
```
dbjorge, line 337: Should the `wcagLevelRegex`es be terminated by `^` and `$` to avoid, say, matching `wcag2aaa` by accident?

build/tasks/validate.js
```diff
@@ -307,5 +301,170 @@ function validateRule({ tags, metadata }) {
...
+  },
+  {
+    name: 'Trusted Tester',
+    standardRegex: /^TTv5$/,
+    criterionRegex: /^TT\d{1,3}\.[a-z]$/,
+    wcagLevelRegex: /wcag2aa?/
+  },
+  {
+    name: 'EN 301 549',
+    standardRegex: /^EN.301.549$/,
```
dbjorge, line 347:
```suggestion
    standardRegex: /^EN\.301\.549$/,
```

build/tasks/validate.js
```diff
@@ -307,5 +301,170 @@ function validateRule({ tags, metadata }) {
...
+  'semantics',
+  'sensory-and-visual-cues',
+  'structure',
+  'tables',
+  'text-alternatives',
+  'time-and-media'
+];
+
+const standardsTags = [
+  {
```
dbjorge, line 328: It's non-obvious that `findTagIssues` relies on the `WCAG` entry being first in the list, maybe leave a comment noting that that's important?

build/tasks/validate.js
```diff
@@ -307,5 +301,170 @@ function validateRule({ tags, metadata }) {
...
+    name: 'WCAG',
```
straker, line 329: Seems odd to leave out validating wcag tags since a rule can only belong to one wcag level. E.g. you can't have `wcag2aa` and `wcag2a` on the same rule. Also I think wcag tags should come before all other standard tags now that we're already enforcing an order.

build/tasks/validate.js
```diff
@@ -307,5 +301,170 @@ function validateRule({ tags, metadata }) {
...
+      ) {
+        issues.push(
+          `Expect WCAG and EN criteria numbers to match: ${wcagCriteria.join(
+            ', '
+          )} vs ${enCriteria.join(', ')}}`
+        );
+      }
+    }
+
+    if (!startsWith(tags, bestPracticeTags)) {
```
straker, line 442: You've already validated against best practice tags up above (line 401) so we shouldn't need to do so again here.

straker (dismissed):
I think the tag validation should be its own PR and that we had some tests around the validation (might be hard since it's in a build step, but it's getting complex now that it's not strictly using the schema validator).

doc/API.md
```diff
@@ -76,23 +76,25 @@ Each rule in axe-core has a number of tags. These provide metadata about the rul
...
+| `best-practice`  | Common accessibility best practices                                                                                           |
+| `wcag***`        | WCAG success criterion e.g. wcag111 maps to SC 1.1.1                                                                          |
+| `ACT`            | W3C approved Accessibility Conformance Testing rules                                                                          |
+| `section508`     | Old Section 508 rules                                                                                                         |
+| `section508.*.*` | Requirement in old Section 508                                                                                                |
+| `TTv5`           | Trusted Tester v5 rules                                                                                                       |
+| `TT*.*`          | Test ID in Trusted Tester                                                                                                     |
+| `EN.301.549`     | Rule required under [EN 301 549](https://www.etsi.org/deliver/etsi_en/301500_301599/301549/03.02.01_60/en_301549v030201p.pdf) |
```
dbjorge, line 94: The `.` separators here feel a bit inconsistent with other tags as a separator for what are spaces in the official standard being referenced. Other existing cases either omit a separator (`Section 508` -> `section508`) or use a hyphen (`best-bractice`). I think `EN-301-549` is the nicest-looking consistent option, but I think any of `EN301549`, `en301549`, `en-301-549` would be fine.

This feedback is minor; I don't think creates big backcompat issues, so if anyone else felt strongly about the `.`s, I can live with the inconsistency.

doc/API.md
```diff
@@ -76,23 +76,25 @@ Each rule in axe-core has a number of tags. These provide metadata about the rul
...
+| `EN-9.*`         | Section in EN 301 549 listing the requirement                                                                                 |
```
dbjorge, line 95: Using just `EN` without any other prefix makes me nervous about creating future backcompat issues; I think future revisions of EN 301 549 are likely to maintain consistent section numbering, but if there's ever a future different EN standard to, say, cover WCAG 3, it's reasonably likely it'd use conflicting numbering. That leads me to prefer using a fuller prefix here, like `EN-301-549-9.1.2.3` or `EN-301-549.9.1.2.3`.

dbjorge (changes requested):
Thanks Wilco! I love the new validation, gave me much more confidence that my eyes didn't glaze over something during review. A few suggestions inline.

build/tasks/validate.js
```diff
@@ -307,5 +301,170 @@ function validateRule({ tags, metadata }) {
...
+  'parsing',
+  'semantics',
+  'sensory-and-visual-cues',
+  'structure',
+  'tables',
+  'text-alternatives',
+  'time-and-media'
+];
+
+const standardsTags = [
```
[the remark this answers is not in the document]
WilcoFiers: It does a little, yes. The order in which we list tags in the rule is the order in which they show up in other places. Having a consistent order will make it a little easier for people to find the tag they are looking for. It's minor, but it helps I think.

build/tasks/validate.js
WilcoFiers: Yes it should :) Fixed

build/tasks/validate.js
```diff
@@ -307,5 +301,170 @@ function validateRule({ tags, metadata }) {
...
+    };
+    if (bestPracticeTags.length) {
+      issues.push(`${name} tags cannot be used along side best-practice tag`);
+    }
+    if (standardTags === 0) {
+      issues.push(`Expected one ${name} tag, got 0`);
+    } else if (standardTags.length > 1) {
+      issues.push(`Expected one ${name} tag, got: ${standardTags.join(', ')}`);
+    }
+    if (!criterionTags.length) {
```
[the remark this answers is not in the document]
WilcoFiers: Fair! Yeah I think I prefer === 0. I'll go with that.

build/tasks/validate.js
```diff
@@ -307,5 +301,170 @@ function validateRule({ tags, metadata }) {
...
+    if (name === 'EN 301 549') {
+      const wcagCriteria = standards.WCAG.criterionTags.map(tag =>
+        tag.slice(4)
+      );
+      const enCriteria = criterionTags.map(tag =>
+        tag.slice(5).replaceAll('.', '')
+      );
+      if (
+        wcagCriteria.length !== enCriteria.length ||
+        !startsWith(wcagCriteria, enCriteria)
```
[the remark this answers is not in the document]
WilcoFiers: ?? This checks that array 1 starts with the items in array 2. It's nothing like a deepEqual. Open to better names, but deepEqual isn't right.

build/tasks/validate.js
```diff
@@ -307,5 +301,170 @@ function validateRule({ tags, metadata }) {
...
+
+    if (!startsWith(tags, bestPracticeTags)) {
+      issues.push(`Tag ${bestPracticeTags[0]} must be before ${tags[0]}`);
+    }
+    tags = removeTags(tags, [...standardTags, ...criterionTags]);
+  }
+
+  // Other tags
+  const usedMiscTags = miscTags.filter(tag => tags.includes(tag));
+  if (!startsWith(tags, usedMiscTags)) {
```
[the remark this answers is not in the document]
WilcoFiers: This checks misc tags are in the correct order. The issue message can be better here though. I'll change that.

doc/API.md
WilcoFiers: We have that problem with other tags too. `wcag111` won't be relevant when WCAG 3 comes out. `TT1.a` will probably be obsolete when Trusted Tester v6 comes out. Don't quite know what we're going to do at that point but it seems likely we need to come up with something different by that time.

I can well imagine we deprecate the criteria/section specific tags all together and put that information somewhere else. That's what we did with the ACT Rules, we added an `actId` prop to the rule.

doc/API.md
WilcoFiers: Dashes make sense. @straker what do you think of `EN-301-549`?

build/tasks/validate.js
```diff
@@ -307,5 +301,170 @@ function validateRule({ tags, metadata }) {
...
+      issues.push(`Expected one ${name} tag, got: ${standardTags.join(', ')}`);
+    }
+    if (!criterionTags.length) {
+      issues.push(
+        `Expected at least one ${name} criterion tag, got ${criterionTags.length}`
+      );
+    }
+
+    if (wcagLevelRegex) {
+      const wcagLevel = standards.WCAG.standardTag;
```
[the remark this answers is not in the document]
WilcoFiers: Comment added

build/tasks/validate.js
straker: Oh you're right. I was basing it solely on the `if` statements looking at section508 and en only, but there's the line that checks that the standard tag doesn't match more than 1 which would cover this case.

doc/API.md
straker: I'm good with dashes

doc/API.md
WilcoFiers: As discussed, the plan is to eventually not include SC numbers / section numbers into tags but make them available in some other format in the output. For now we're going with what we've been doing up until this point.

dbjorge (approved):
LGTM!

straker (approved):

PR #4071, opened 2023-06-28
feat: deprecate & disable duplicate-id / duplicate-id-active

WilcoFiers:
Disable, remove, and update tags for `duplicate-id` and `duplicate-id-active`.

Closes: #3892


dylanb:
Don't we also need to disable these by default?

dylanb (approved):

dylanb:
> Don't we also need to disable these by default?

They are disabled by default

adamraine:
@WilcoFiers I'm testing this out on the latest canary release with `duplicated-id` explicitly enabled and it doesn't appear to be catching any violations. Is this intended behavior?

PR #4075, opened 2023-06-30
feat(duplicate-id-aria): set to review on fail and tag as wcag412

WilcoFiers:
Closes: #3892


test/integration/rules/runner.js
```diff
@@ -95,7 +95,11 @@
               // check that all nodes are removed
               assert.equal(JSON.stringify(targets), '[]');
             } else {
-              assert(false, 'there are no ' + collection);
+              assert.lengthOf(
+                test[collection],
+                0,
```
WilcoFiers, line 100: This lets us say not to expect any violations. Previously that was just erroring.

dbjorge:
This is marked as closing 3892, but that issue includes this note that I don't think has been resolved yet:

> Future: We've been talking about possibly replacing [duplicate-id-aria] with a more focused idref-labels-and-aria rule that doesn't just look at duplicates, but considers whether based on DOM order that's likely to cause an accessibility problem. As part of resolving this issue we should have a proposal of that rule written.

Are we still planning that? Do we at least want to extract that bit to a separate rule suggestion issue before allowing this PR to close it?

doc/rule-descriptions.md
```diff
@@ -37,7 +37,7 @@
...
+| [duplicate-id-aria](https://dequeuniversity.com/rules/axe/4.7/duplicate-id-aria?application=RuleDescription)                     | Ensures every id attribute value used in ARIA and in labels is unique                                                                                 | Critical                 | cat.parsing, wcag2a, wcag412, EN-301-549, EN-9.4.1.2                                                                               | needs&nbsp;review          | [3ea0c8](https://act-rules.github.io/rules/3ea0c8)                                                                                                                                                                                                                 |
```
dbjorge, line 40: Are we still happy with `Critical` for this rule?

WilcoFiers: Yeah, I think that's right. Getting the ID wrong on something like a form label is potentially a vey big problem.

dbjorge (approved):
Code changes LGTM!

WilcoFiers:
Here's the issue you asked for: https://github.com/dequelabs/axe-core/issues/4076

Issue #4064, opened 2023-06-26
Text stroke (shadow) edge case not getting picked up


WilcoFiers:
This should pass, however when I run it through axe the text-stroke is seen as too thin to be relevant. Not sure why that is, but this end up with a 1px stroke around text that is really fairly readable. My guess is this comes because it has move overlapping shadows than I initially accounted for:

```html
<style>
  h3 {
    text-shadow:
      -1px -1px 0 #000,
      0 -1px 0 #000,
      1px -1px 0 #000,
      1px 0 0 #000,
      1px 1px 0 #000,
      0 1px 0 #000,
      -1px 1px 0 #000,
      -1px 0 0 #000;
    color: #ffcc00;
    font-size: 24px;
    letter-spacing: normal;
    text-transform: uppercase;
  }
</style>
<h3>Hello World</h3>
```

padmavemulapati:
Validated using axe-core develop branch code base,

the code with text-shadow stroke with 1px around it is passing as it is readable



https://github.com/dequelabs/axe-core/assets/47409579/e1ceb56b-946e-4684-b9af-6adc2e714f8c

PR #4079, opened 2023-07-03
fix(color-contrast): get text stoke from offset shadows

WilcoFiers:
If several text shadows with some offset and with minimal blur are used together, they can combine into a text stroke effect. This PR attempts to identify such cases and combine these shadows into a single color.

Closes: #4064


test/commons/color/get-text-shadow-colors.js
```diff
@@ -192,4 +192,154 @@ describe('axe.commons.color.getTextShadowColors', function () {
...
+  describe('with shadows that combine to a stroke', () => {
+    const opt = { minRatio: 0.01 };
+    it('combines multiple shadows without blur into a single stroke', () => {
+      fixture.innerHTML = `
+        <span style="text-shadow:
+          0 -2px #F00,
+          2px 0 #F00,
+          0 2px #F00,
+          -2px 0 #F00;
+        ">Hello wold</span>
```
dbjorge, line 205: nit for all of these tests:

```suggestion
        ">Hello world</span>
```

test/commons/color/get-text-shadow-colors.js
```diff
@@ -192,4 +192,154 @@ describe('axe.commons.color.getTextShadowColors', function () {
...
+      `;
+      const shadowColors = getTextShadowColors(fixture.firstElementChild, opt);
+      assert.lengthOf(shadowColors, 0);
+    });
+
+    it('skips shadows offset with 0.5px or less', () => {
+      fixture.innerHTML = `
+        <span style="text-shadow:
+          0.5px 0.5px #000,
```
dbjorge, line 250: This test case assumes that 2 shadows on opposite corners would be recognized correctly as a normally-combined case; it'd be good to have a test case verifying that explicitly ahead of this one

lib/commons/color/get-text-shadow-colors.js
```diff
@@ -22,29 +31,96 @@ function getTextShadowColors(node, { minRatio, maxRatio } = {}) {
...
+  if (thinShadows.length > 0) {
+    // Append any remaining stroke colors
+    const strokeColors = getStrokeColor(thinShadows);
+    shadowColors.push(...strokeColors);
+  }
+
   return shadowColors;
 }
 
+function getStrokeColor(thinShadows) {
```
dbjorge, line 73:
```suggestion
function getStrokeColors(thinShadows) {
```

lib/commons/color/get-text-shadow-colors.js
```diff
@@ -1,18 +1,27 @@
...
+/** Magic numbers **/
+// Alpha value to use when text shadows are offset between .5px and 1.5px
+const SHADOW_STROKE_ALPHA = 0.54;
+// Shadows offset by less than this are not visible enough no matter how much you stack them
+const VISIBLE_SHADOW_MIN_PX = 0.5;
+// Shadows offset by more than this have full opacity
+const OPAQUE_STROKE_OFFSET_MIN_PX = 1.5;
+
 /**
  * Get text-shadow colors that can impact the color contrast of the text
```
dbjorge, line 13: It would be nice for the doc comment to summarize the total behavior

```suggestion
 * Get text-shadow colors that can impact the color contrast of the text, including from:
 * - Shadows which are individually thick enough to distinguish as characters
 * - Groups of thin shadows that collectively act as a pseudo-text-stroke (see #4064)
```

lib/commons/color/get-text-shadow-colors.js
```diff
@@ -1,18 +1,27 @@
...
  * @param {Element} node  DOM Element
  * @param {Object} options (optional)
  * @property {Bool} minRatio Ignore shadows smaller than this, ratio shadow size divided by font size
```
dbjorge, line 16:
```suggestion
 * @property {Bool} minRatio Treat shadows smaller than this as "thin", ratio shadow size divided by font size
```

test/commons/color/get-text-shadow-colors.js
```diff
@@ -192,4 +192,154 @@ describe('axe.commons.color.getTextShadowColors', function () {
...
+          -2px 0 ${minRatio}em #000;
+        ">Hello wold</span>
+      `;
+      const shadowColors = getTextShadowColors(fixture.firstElementChild, {
+        minRatio
+      });
+      assert.lengthOf(shadowColors, 4);
+    });
+
+    it('skips raised-letter effects (shadows on 1 or 2 sides)', () => {
```
dbjorge, line 234: Would be good to have a test that verifies that exactly 3 sides is sufficient to be not-skipped

lib/commons/color/get-text-shadow-colors.js
```diff
@@ -22,29 +31,96 @@ function getTextShadowColors(node, { minRatio, maxRatio } = {}) {
...
+    // Average the number of shadows around the text
+    let density = 0;
+    let isSolid = true;
+    edges.forEach(edge => {
+      density += sides[edge].length / 4;
+      isSolid &&= sides[edge].every(
+        ([x, y]) =>
+          Math.abs(x) > OPAQUE_STROKE_OFFSET_MIN_PX ||
+          Math.abs(y) > OPAQUE_STROKE_OFFSET_MIN_PX
+      );
```
dbjorge, line 111: This is assuming that a 1.4px shadow and a 0.6px shadow should contribute equally to density, but I don't think that's accurate in practice based on playing with the `--thickness` CSS var in [this test playground](https://developer.mozilla.org/en-US/play?id=b4w6cdYxibzUleKHyvYH%2BRXJbVLocqnjdC82xafXh21t4sfT4M%2FM%2F24pjp5%2BvuaSJm7TToYJTgsBuSrq). Perhaps better to weight a side's contribution to density according to its relevant x/y value, as opposed to each side contributing a flat 1? I think the below covers that such that the behavior is the same as the original magic algorithm for 1px thick shadows

```suggestion
      for (const side of sides[edge]) {
        const relevantSideProperty = { left: 'x', right: 'x', top: 'y', bottom: 'y' }[edge];
        const sideThickness = Math.abs(side[relevantSideProperty])
        isSolid &&= sideThickness > OPAQUE_STROKE_OFFSET_MIN_PX
        density += Math.min(sideThickness, OPAQUE_STROKE_OFFSET_MIN_PX)
      }
```

dbjorge (changes requested):
Thanks Wilco! Some suggestions inline

lib/commons/color/get-text-shadow-colors.js
WilcoFiers: The reason I did that is because I noticed when you want half-pixels shadows aren't distributed evenly. When I did a .6px I got 70% alpha on one side of the text and 16% on the opposite side. On the other hand we do sort of still pull an average, so this makes sense.

dbjorge: That's a convincing enough reason to give these cases uniform density. Maybe leave a comment explaining that reasoning?

I do think the functional change I've made here to `isSolid` would be good to pull in regardless, though; I think the original logic would treat a `.5px 2px #000, -2px -.5px #000` shadow as solid when it doesn't intend to. Maybe a cleaner fix for that would be to update further up when you're building up `colorMap` (the 4 lines like `borders.N.push(pixels)`) to push only the relevant `offsetN` into the side's array instead of `pixels`?

lib/commons/color/get-stroke-colors-from-shadows.js
```diff
@@ -0,0 +1,90 @@
...
+ * Using colorStr and thickness of sides, create a color object
+ */
+function shadowGroupToColor({ colorStr, sides, edgeCount }) {
+  if (edgeCount !== 4) {
+    return null; // ignore thin shadows and shadows on one side of the text
+  }
+  const strokeColor = new Color();
+  strokeColor.parseString(colorStr);
+
+  // Average the number of shadows around the text
```
dbjorge, line 69:
```suggestion
  // Detect whether any sides' shadows are thin enough to be considered
  // translucent, and if so, calculate an alpha value to apply on top of
  // the parsed color.
```

lib/commons/color/get-stroke-colors-from-shadows.js
```diff
@@ -0,0 +1,90 @@
...
+    }
+  }
+  return colorMap;
+}
+
+/**
+ * Using colorStr and thickness of sides, create a color object
+ */
+function shadowGroupToColor({ colorStr, sides, edgeCount }) {
+  if (edgeCount !== 4) {
```
dbjorge, line 63: It feels off that this check is still here when the caller is now *also* doing an overlapping edgeCount check. Maybe more clear to move this check out of `shadowGroupToColor` and update the `.filter` on L32 to filter out based on edgeCount before mapping this function, instead of mapping this function and then filtering based on its return value.

test/commons/color/get-stroke-colors-from-shadows.js
```diff
@@ -0,0 +1,120 @@
...
+    assert.lengthOf(shadowColors, 1);
+    assert.equal(shadowColors[0].red, 0);
+    assert.equal(shadowColors[0].green, 0);
+    assert.equal(shadowColors[0].blue, 0);
+    assert.closeTo(shadowColors[0].alpha, 0.7, 0.01);
+  });
+
+  it('double-counts shadows on corners', () => {
+    // Corner-shadows overlap on the sides of letters, increasing alpha
+    const shadows = parseTextShadows(`
```
dbjorge, line 107: This is the only test case that's covering shadow-on-corners behavior; it'd be good to have a case that verifies that "2 opposing corner shadows is enough to not return empty/null"

lib/commons/color/get-background-color.js
```diff
@@ -54,7 +54,8 @@ function _getBackgroundColor(elm, bgElms, shadowOutlineEmMax) {
   }
 
   const textRects = getVisibleChildTextRects(elm);
-  let bgColors = getTextShadowColors(elm, { minRatio: shadowOutlineEmMax });
+  let bgColors =
+    getTextShadowColors(elm, { minRatio: shadowOutlineEmMax }) ?? [];
```
dbjorge, line 58: For the `?? []` to come up in `color-contrast` and not have already resulted in a `complexTextShadows` exit in `color-contrast-evaluate`, an element would need to have a "complex" (2-3 edge compound) shadow of blur radius between .001em and 0.2em. I think if that were to actually happen in practice *and* there was additionally a higher-blur-radius shadow(s), it probably doesn't make sense to silently throw away those higher-blur-radius shadow(s). I can be persuaded that either of "bail as incomplete" or "ignore those cases but still consider the thick shadow behind as a possible background color" would be reasonable.

`link-in-text-block` also uses `getBackgroundColor` and I think the `?? []` is even sketchier there, since it doesn't early-exit with incomplete for blur radius < .001 complex compound shadows like `color-contrast` does. Generally, it seems a bit strange to me that `link-in-text-block-evaluate` isn't reusing the same support for shadows providing contrast that color contrast is.

dbjorge (dismissed):

lib/commons/color/get-background-color.js
WilcoFiers: I can't sneak anything past you can I? I realised that too when I wrote this. Wasn't sure whether I should care, but fair enough.

dbjorge (approved):
:shipit:

Issue #4082, opened 2023-07-07
Update to allow role doc-glossary on aside element


scottaohara:
ARIA in HTML is updating to allow the `doc-glossary` dpub aria role to the aside element.

https://github.com/w3c/html-aria/pull/455

PR #4083, opened 2023-07-07
Add doc-glossary to allowed roles for aside element

scottaohara:
updates html-elms.js to add doc-glossary to the allowed roles list.  

Updates for aria-allowed-role.html/.json files are incoming.

Closes: #4082


straker (approved):
Awesome. Thanks for always keeping us up-to-date on these things.

Issue #4041, opened 2023-05-25
Incorrect element stack for flex and position absolute


straker:
For the page https://www.sony.com/en/SonyInfo/csr/, `axe.commons.dom.getElementStack` gives an incorrect stack for the links in the navbar (e.g. "Contact Us" `a[href$="SupportAll/"]`) and reports them as needs review. I'll have to debug further as to the reason, but at a quick glance it looks like theres another header on the page that is claiming to be on top of the stack. The headers are using flex and some position: absolute.

straker:
@padmavemulapati the fix did not resolve the needs review issue, so not much will have changed while running the page. However axe-core is reporting it correctly as there is indeed another element on top of the links that if it had a background would hide everything underneath if it had a background color. Going to say this is a dev task.

PR #4086, opened 2023-07-11
fix(color-contrast): correctly handle flex and position

straker:
Mostly it's rearranging the order of the `createStackingContext` function to handle elements that created stacking context and needed to be positioned or floated. The code didn't handle elements that were both positioned with `z-index: auto` and another property that made it a stacking context (like `display: flex`). 

Also changed how to handle adding the node index to every context by making it it's own property instead of combining it with the stacking context value. Made things simpler when creating / comparing the numbers. This was needed to allow sorting elements from different (but equal) stacking contexts correctly. Before they would be sorted as equal (stacking contexts were the same value) and then relied on the document order and positional sorting (based on float or inline), even when they belonged to different stacking context and shouldn't be sorted by position. Now the elements will only be positionally sorted when they belong to the same stacking context.

It should be noted that this fix doesn't actually result in resolving the incomplete in #4041. There is another element (`.tmpl-header_searchBtnContainer`) that is on top of the others (with a transparent background) and so the incomplete is correct that there is another element overlapping the links. I didn't realize this until after I made the stack correct.

Closes: #4041

### DIFF
--- a/lib/commons/dom/create-grid.js
+++ b/lib/commons/dom/create-grid.js
@@ -10,7 +10,7 @@ import assert from '../../core/utils/assert';
 const ROOT_ORDER = 0;
 const DEFAULT_ORDER = 0.1;
 const FLOAT_ORDER = 0.2;
-const POSITION_STATIC_ORDER = 0.3;
+export const POSITION_ORDER = 0.3;
 let nodeIndex = 0;
 
 /**
@@ -39,7 +39,7 @@ export default function createGrid(
     }
 
     nodeIndex = 0;
-    vNode._stackingOrder = [createContext(ROOT_ORDER, null)];
+    vNode._stackingOrder = [createContext(ROOT_ORDER, null, nodeIndex++)];
     rootGrid ??= new Grid();
     addNodeToGrid(rootGrid, vNode);
 
@@ -240,7 +240,7 @@ function isStackingContext(vNode, parentVNode) {
  * @param {VirtualNode} vNode
  * @return {Boolean}
  */
-function isFlexOrGridContainer(vNode) {
+export function isFlexOrGridContainer(vNode) {
   if (!vNode) {
     return false;
   }
@@ -254,65 +254,61 @@ function isFlexOrGridContainer(vNode) {
  * zIndex values for each stacking context parent.
  * @param {VirtualNode} vNode
  * @param {VirtualNode} parentVNode
- * @param {Number} nodeIndex
+ * @param {Number} index
  * @return {Number[]}
  */
-function createStackingOrder(vNode, parentVNode, nodeIndex) {
+function createStackingOrder(vNode, parentVNode, index) {
   const stackingOrder = parentVNode._stackingOrder.slice();
-
-  // if a positioned element has z-index: auto or 0 (step #8), or if
-  // a non-positioned floating element (step #5), treat it as its
-  // own stacking context
-  // @see https://www.w3.org/Style/css2-updates/css2/zindex.html
-  if (!isStackingContext(vNode, parentVNode)) {
-    if (vNode.getComputedStylePropertyValue('position') !== 'static') {
-      // Put positioned elements above floated elements
-      stackingOrder.push(createContext(POSITION_STATIC_ORDER, vNode));
-    } else if (vNode.getComputedStylePropertyValue('float') !== 'none') {
-      // Put floated elements above z-index: 0
-      // (step #5 floating get sorted below step #8 positioned)
-      stackingOrder.push(createContext(FLOAT_ORDER, vNode));
-    }
-    return stackingOrder;
-  }
 
   // if an element creates a stacking context, find the first
   // true stack (not a "fake" stack created from positioned or
   // floated elements without a z-index) and create a new stack at
   // that point (step #5 and step #8)
   // @see https://www.w3.org/Style/css2-updates/css2/zindex.html
-  const index = stackingOrder.findIndex(({ value }) =>
-    [ROOT_ORDER, FLOAT_ORDER, POSITION_STATIC_ORDER].includes(value)
-  );
-  if (index !== -1) {
-    stackingOrder.splice(index, stackingOrder.length - index);
+  if (isStackingContext(vNode, parentVNode)) {
+    const index = stackingOrder.findIndex(({ value }) =>
+      [ROOT_ORDER, FLOAT_ORDER, POSITION_ORDER].includes(value)
+    );
+    if (index !== -1) {
+      stackingOrder.splice(index, stackingOrder.length - index);
+    }
   }
 
   const zIndex = getRealZIndex(vNode, parentVNode);
   if (!['auto', '0'].includes(zIndex)) {
-    stackingOrder.push(createContext(parseInt(zIndex), vNode));
+    stackingOrder.push(createContext(parseInt(zIndex), vNode, index));
     return stackingOrder;
   }
-  // since many things can create a new stacking context without position or
-  // z-index, we need to know the order in the dom to sort them by. Use the
-  // nodeIndex property to create a number less than the "fake" stacks from
-  // positioned or floated elements but still larger than 0
-  // 10 pad gives us the ability to sort up to 1B nodes (padStart does not
-  // exist in ie11)
-  let float = nodeIndex.toString();
-  while (float.length < 10) {
-    float = '0' + float;
-  }
-  stackingOrder.push(
-    createContext(parseFloat(`${DEFAULT_ORDER}${float}`), vNode)
-  );
 
+  // if a positioned element has z-index: auto or 0 (step #8), or if
+  // a non-positioned floating element (step #5), treat it as its
+  // own stacking context
+  // @see https://www.w3.org/Style/css2-updates/css2/zindex.html
+  if (vNode.getComputedStylePropertyValue('position') !== 'static') {
+    // Put positioned elements above floated elements
+    stackingOrder.push(createContext(POSITION_ORDER, vNode, index));
+    return stackingOrder;
+  }
+
+  if (vNode.getComputedStylePropertyValue('float') !== 'none') {
+    // Put floated elements above z-index: 0
+    // (step #5 floating get sorted below step #8 positioned)
+    stackingOrder.push(createContext(FLOAT_ORDER, vNode, index));
+    return stackingOrder;
+  }
+
+  if (!isStackingContext(vNode, parentVNode)) {
+    return stackingOrder;
+  }
+
+  stackingOrder.push(createContext(DEFAULT_ORDER, vNode, index));
   return stackingOrder;
 }
 
-function createContext(value, vNode) {
+function createContext(value, vNode, index) {
   return {
     value,
+    nodeIndex: index,
     vNode
   };
 }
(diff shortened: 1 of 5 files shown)


test/commons/dom/get-element-stack.js
```diff
@@ -465,6 +465,27 @@ describe('dom.getElementStack', () => {
       assert.deepEqual(stack, []);
     });
 
+    it('should correctly position children of different stacking contexts', () => {
+      fixture.innerHTML =
+        '<header id="1" style="position: absolute; z-index: 999; height: 50px; width: 100%; top: 0">' +
```
dbjorge, line 470: It's a bit hard to grok the test without indentation on the `innerHTML`; maybe use the style from `it('should correctly order elements outside of the axe tree', ...)` instead?

lib/commons/dom/create-grid.js
```diff
@@ -254,65 +254,61 @@ function isFlexOrGridContainer(vNode) {
...
+    // Put floated elements above z-index: 0
+    // (step #5 floating get sorted below step #8 positioned)
+    stackingOrder.push(createContext(FLOAT_ORDER, vNode, index));
+    return stackingOrder;
+  }
+
+  if (!isStackingContext(vNode, parentVNode)) {
     return stackingOrder;
   }
-  // since many things can create a new stacking context without position or
```
dbjorge, line 296: This is a nice change, I think including the `nodeIndex` in the comparator is much cleaner than the old way.

test/commons/dom/get-element-stack.js
straker: Good point. I sometimes forget that we can use newer es6 syntax in tests now.

lib/commons/dom/create-grid.js
```diff
@@ -254,65 +254,61 @@ function isFlexOrGridContainer(vNode) {
  * zIndex values for each stacking context parent.
  * @param {VirtualNode} vNode
  * @param {VirtualNode} parentVNode
- * @param {Number} nodeIndex
+ * @param {Number} index
```
WilcoFiers, line 257: Can you give this a better name? If this is no longer an index of the node in the flat tree, what is it? I don't think it's an index at all anymore. You can't use this number to look something up in a table / tree.

lib/commons/dom/create-grid.js
```diff
@@ -254,65 +254,61 @@ function isFlexOrGridContainer(vNode) {
...
   // floated elements without a z-index) and create a new stack at
   // that point (step #5 and step #8)
   // @see https://www.w3.org/Style/css2-updates/css2/zindex.html
-  const index = stackingOrder.findIndex(({ value }) =>
-    [ROOT_ORDER, FLOAT_ORDER, POSITION_STATIC_ORDER].includes(value)
-  );
-  if (index !== -1) {
-    stackingOrder.splice(index, stackingOrder.length - index);
+  if (isStackingContext(vNode, parentVNode)) {
+    const index = stackingOrder.findIndex(({ value }) =>
```
WilcoFiers, line 269: Please don't have two variables with the same name in one function. This code is complicated enough 😁

lib/commons/dom/create-grid.js
```diff
@@ -254,65 +254,61 @@ function isFlexOrGridContainer(vNode) {
...
-  );
 
+  stackingOrder.push(createContext(DEFAULT_ORDER, vNode, index));
   return stackingOrder;
 }
 
-function createContext(value, vNode) {
+function createContext(value, vNode, index) {
   return {
     value,
```
WilcoFiers, line 310: Could we rename this, maybe to `stackOrderValue` or something as well, for clarity?

lib/commons/dom/create-grid.js
```diff
@@ -254,65 +254,61 @@ function isFlexOrGridContainer(vNode) {
  * zIndex values for each stacking context parent.
  * @param {VirtualNode} vNode
  * @param {VirtualNode} parentVNode
- * @param {Number} nodeIndex
+ * @param {Number} index
  * @return {Number[]}
  */
-function createStackingOrder(vNode, parentVNode, nodeIndex) {
+function createStackingOrder(vNode, parentVNode, index) {
```
WilcoFiers, line 260: I took a stab at refactoring this. Since you have a couple of repeated `stackingOrder.push(createContext())` calls I pulled finding that order value into its own function. I think this helps explain what's going on.

```js
function createStackingOrder(vNode, parentVNode, gridIndex) {
  const stackingOrder = parentVNode._stackingOrder.slice();

  // if an element creates a stacking context, find the first
  // true stack (not a "fake" stack created from positioned or
  // floated elements without a z-index) and create a new stack at
  // that point (step #5 and step #8)
  // @see https://www.w3.org/Style/css2-updates/css2/zindex.html
  if (isStackingContext(vNode, parentVNode)) {
    const index = stackingOrder.findIndex(({ value }) =>
      [ROOT_ORDER, FLOAT_ORDER, POSITION_ORDER].includes(value)
    );
    if (index !== -1) {
      stackingOrder.splice(index, stackingOrder.length - index);
    }
  }

  const stackOrderValue = getStackOrderValue(vNode, parentVNode);
  if (stackOrderValue !== null) {
    stackingOrder.push(createContext(stackOrderValue, vNode, gridIndex));
  }
  return stackingOrder;
}

function getStackOrderValue(vNode, parentVNode) {
  const zIndex = getRealZIndex(vNode, parentVNode);
  if (!['auto', '0'].includes(zIndex)) {
    return parseInt(zIndex);
  }
  // if a positioned element has z-index: auto or 0 (step #8), or if
  // a non-positioned floating element (step #5), treat it as its
  // own stacking context @see https://www.w3.org/Style/css2-updates/css2/zindex.html
  if (vNode.getComputedStylePropertyValue('position') !== 'static') {
    return POSITION_ORDER;
  }
  // Put floated elements above z-index: 0
  // (step #5 floating get sorted below step #8 positioned)
  if (vNode.getComputedStylePropertyValue('float') !== 'none') {
    return FLOAT_ORDER;
  }
  if (isStackingContext(vNode, parentVNode)) {
    return DEFAULT_ORDER;
  }
  return null;
}
```

lib/commons/dom/create-grid.js
WilcoFiers, line 1: Still feel this needs to be split up. We've had that discussion before though, no need to get into it if you still disagree.

lib/commons/dom/create-grid.js
```diff
@@ -254,65 +254,61 @@ function isFlexOrGridContainer(vNode) {
...
 
+  stackingOrder.push(createContext(DEFAULT_ORDER, vNode, index));
   return stackingOrder;
 }
 
-function createContext(value, vNode) {
+function createContext(value, vNode, index) {
   return {
     value,
+    nodeIndex: index,
```
WilcoFiers, line 311: This tripped me up. Since VirtualNode also has a nodeIndex prop, I think we should call this something different. Are you okay with using `gridIndex` instead?

```suggestion
    gridIndex,
```

lib/commons/dom/create-grid.js
```diff
@@ -254,65 +254,61 @@ function isFlexOrGridContainer(vNode) {
...
   const stackingOrder = parentVNode._stackingOrder.slice();
 
-  // if a positioned element has z-index: auto or 0 (step #8), or if
-  // a non-positioned floating element (step #5), treat it as its
-  // own stacking context
-  // @see https://www.w3.org/Style/css2-updates/css2/zindex.html
-  if (!isStackingContext(vNode, parentVNode)) {
-    if (vNode.getComputedStylePropertyValue('position') !== 'static') {
-      // Put positioned elements above floated elements
-      stackingOrder.push(createContext(POSITION_STATIC_ORDER, vNode));
```
WilcoFiers, line 270: Wow, did we call the position for when something's NOT static the `POSITION_STATIC_ORDER`. Whoops! 🙄 That was my idea too wasn't it?

lib/commons/dom/create-grid.js
WilcoFiers, line 311: Also, I think you need tests in `test/commons/dom/create-grid.js` to check this value is set correctly.

lib/commons/dom/visually-sort.js
```diff
@@ -27,6 +27,11 @@ export default function visuallySort(a, b) {
     if (b._stackingOrder[i].value < a._stackingOrder[i].value) {
       return -1;
     }
+
+    // stacks share the same stacking context value so compare document order
+    if (b._stackingOrder[i].nodeIndex !== a._stackingOrder[i].nodeIndex) {
+      return b._stackingOrder[i].nodeIndex - a._stackingOrder[i].nodeIndex;
```
WilcoFiers, line 33: I think this should be tested directly in `test/commons/dom/visually-sort.js`

WilcoFiers (changes requested):

lib/commons/dom/create-grid.js
dbjorge: Thoughts on a separate PR enabling the `no-shadow` eslint rule to catch this sort of issue proactively? (111 pre-existing issues, not trivial but manageable)

lib/commons/dom/create-grid.js
dbjorge: I agree with avoiding the name conflict with virtualNodes, but I'm not a fan of `gridIndex`; that to me implies that it's an index into the grid, which isn't accurate.

Does the nodeIndex on VirtualNode already have the same semantics? If so, can this just be dropped and have `visuallySort` use that, since it already has references to virtual nodes? If not, maybe `domPosition`, `domOrder`, `domOrderComparator`, or similar?

straker: Since the grid can [create elements that are missing from the virtual tree](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/dom/create-grid.js#L80-L82) this `nodeIndex` is not guaranteed to be the same as the one on the virtual tree so I can't use that for sorting. I can call it `domPosition` as it makes it relate to the `compareDocumentPosition` function in my mind.

lib/commons/dom/create-grid.js
straker: I'd rather do that in a separate pr

lib/commons/dom/create-grid.js
straker: I think I'll call this `treeOrder` as it matches the [spec definition](https://www.w3.org/Style/css2-updates/css2/zindex.html)

> Tree Order
> Preorder depth-first traversal of the rendering tree, in logical (not visual) order for bidirectional content, after taking into account properties that move boxes around.

lib/commons/dom/create-grid.js
straker: Renamed it to `stackLevel` to [match spec](https://www.w3.org/Style/css2-updates/css2/visuren.html#layers).

lib/commons/dom/create-grid.js
```diff
@@ -251,72 +253,94 @@ function isFlexOrGridContainer(vNode) {
...
 
-function createContext(value, vNode) {
+/**
+ * Create a stacking context, keeping track of the stack level, tree order, and virtual
+ * node container.
+ * @see https://www.w3.org/Style/css2-updates/css2/zindex.html
+ * @see https://www.w3.org/Style/css2-updates/css2/visuren.html#layers
+ * @param {Number} stackLevel - The stack level of the stacking context
+ * @param {Number} treeOrder - The elements depth-first traversal order
+ * @param {VirtualNode} vNode - The virtual node that is the container for the stacking context
```
WilcoFiers, line 293: Thank you!

WilcoFiers (approved):

lib/commons/dom/create-grid.js
WilcoFiers: Please

Issue #4067, opened 2023-06-28
Put aria-prohibited-attr (from aria-allowed-attr) into its own rule


WilcoFiers:
Part of https://github.com/dequelabs/axe-core/issues/4066

The `aria-prohibited-attr` check should run in its own rule, instead of as part of `aria-allowed-attr`.

padmavemulapati:
Validated using latest axe-core develop branch code base,
seeing `aria-prohibited-attr` as a seperate rule instead `aria-allowed-attr` for the prohibited attributes like caption, code, deletion, emphsis, insertion ... etc.,



https://github.com/dequelabs/axe-core/assets/47409579/24659caf-0399-4008-9a6f-5a1e0fbd343b

PR #4088, opened 2023-07-12
feat(aria-prohibited-attr): add rule aria-prohibited-attr

straker:
Closes: #4067


test/integration/rules/aria-allowed-attr/failures.html
WilcoFiers, line 1: Maybe add a few tests showing aria-allowed-attr no longer fails prohibited attributes. You've done the reverse, but this way around would be good to have IMO, just like we did on aria-deprecated-attr.

WilcoFiers (changes requested):
Minor

test/integration/rules/aria-allowed-attr/passes.html
```diff
@@ -2153,3 +2153,8 @@
 
 <input type="checkbox" aria-checked="false" id="pass92" />
 <input type="checkbox" aria-checked="true" checked id="pass93" />
+
+<div role="caption" aria-label="value" id="pass94"></div>
```
WilcoFiers, line 2157:
```suggestion
<!-- Prohibited attributes fail in aria-prohibited-attr -->
<div role="caption" aria-label="value" id="pass94"></div>
```

test/integration/rules/aria-prohibited-attr/aria-prohibited-attr.html
```diff
@@ -0,0 +1,40 @@
...
+<div role="paragraph" aria-labelledby="value" id="fail12"></div>
+<div role="strong" aria-label="value" id="fail13"></div>
+<div role="strong" aria-labelledby="value" id="fail14"></div>
+<div role="subscript" aria-label="value" id="fail15"></div>
+<div role="subscript" aria-labelledby="value" id="fail16"></div>
+<div role="superscript" aria-label="value" id="fail17"></div>
+<div role="superscript" aria-labelledby="value" id="fail18"></div>
+<div aria-label="value" id="fail19"></div>
+<div aria-labelledby="value" id="fail20"></div>
+<!-- technically presentation and none roles do not allow aria-label and aria-labelledby, but since those are global attributes the presentation role conflict will not resolve the roles to none or presentation -->
```
WilcoFiers, line 27: Because ARIA loves throwing curve balls, this is no longer true. `aria-label(ledby)` is global "except where prohibited". Which Chromium took to mean that if you use aria-label on role="none", role conflict doesn't trigger. It still does in Firefox. Axe-core would fail it regardless, we're just maybe failing it in the wrong rule?

Maybe we should change this to:

```suggestion
<!- aria-label(ledby) is prohibited on none / presentation. Axe-core considers this to trigger presentation role 
conflict, which was true in ARIA 1.1. This changed in ARIA 1.2 but so far has only been implemented in Chomium. -->
```

WilcoFiers (approved):
LGTM. 2 suggestions for you, if you want them.

WilcoFiers (approved):

Issue #3752, opened 2022-10-28
header in section with role should not have role banner


lexanth:
### Product

axe-core

### Product Version

4.4.1

### Latest Version

- [X] I have tested the issue with the latest version of the product

#### Expectation
Given a dialog which is labelled by a header:
```html
<section role="dialog" aria-labelledby="dialog-header">
  <header id="dialog-header">My header</header>
  <div>Some dialog content</div>
</section>
```
I would expect the accessible name of the dialog to be identified as "My header"

#### Actual
I get a violation on "ARIA dialog and alertdialog nodes should have an accessible name (aria-dialog-name)"

On further investigation:
* It is correctly identifying the `aria-labelledby` link
* In determining the text of the `header`, it first considers its role
* The role is identified as `banner`, because https://github.com/dequelabs/axe-core/blob/522bdfb57e/lib/commons/standards/implicit-html-roles.js#L17 only links the header to the section if the section does not have a role attribute.
* Banners are not named from contents (https://github.com/dequelabs/axe-core/blob/522bdfb57e/lib/commons/aria/named-from-contents.js#L36) so the contents are ignored, so the element linked by `aria-labelledby` is identified as empty, hence the violation

#### How to Reproduce
See HTML in Expectation

#### Additional context
I haven't verified if a screen reader identifies the header as the dialog header.

I haven't seen anywhere that says that sectioning HTML elements are only sectioning when they don't have a role. https://www.w3.org/TR/2010/WD-html5-20101019/content-models.html#sectioning-content-0

All of the valid roles for a section in https://www.w3.org/TR/2010/WD-html5-20101019/content-models.html#annotations-for-assistive-technology-products-aria seem like they should still act as sectioning elements. I think invalid roles would be caught by the corresponding rule?

I can work around this by explicitly setting `role="none"` on my `header`.

straker:
Interesting. So testing this out:

```html
<section tabindex="0" role="dialog" aria-labelledby="dialog-header">
    <div>Some dialog content</div>
  </section>

<header id="dialog-header" hidden>My header</header>
```

Each Browser / Screen Reader reports the `[role=dialog]` as "My header, dialog".

So even though [the `header` should be a banner role](https://www.w3.org/TR/html-aria/#el-header) since it's not a descendant of another sectioning element, seems screen readers are allowing named from contents on it even though it's suppose to be named from author.

WilcoFiers:
[...]

PR #4097, opened 2023-07-20
fix(access-name): get name from header elements

WilcoFiers:
The content of landmarks and other elements that have "name from author" should not be ignored when referenced by aria-labelledby or another label. 

This is a bug, where axe was inconsistent with the accessible name spec. Specifically, [step F](https://www.w3.org/TR/accname-1.1/#step2F):
> if the current node's role allows name from content, or if the current node is referenced by aria-labelledby, aria-describedby, or is a native host language text alternative element (e.g. label in HTML), or is a descendant of a native host language text alternative element

Previously axe-core was only considering the first part of this sentence, and skipping name from author elements while following aria-labelledby and `<label>` IDREFs. This was a problem on dialogs (as the issue reported), but had broader implications too. Notably, column and rowheaders in a table could not be referenced with aria-labelledby.

That created a second problem; Elements with a value (like select and progressbar) should never use name from content. I had to add a condition in for that. I ended up testing the use of element values  in names again. This improved a good bit, and so now only the value of progressbar still can't be relied on to give an accessible name.

Closes: #3752

### DIFF
--- a/lib/commons/text/unsupported.js
+++ b/lib/commons/text/unsupported.js
@@ -1,5 +1,7 @@
-const unsupported = {
-  accessibleNameFromFieldValue: ['combobox', 'listbox', 'progressbar']
+export default {
+  // Element's who's value is not consistently picked up in the accessible name
+  // Supported in Chrome 114, Firefox 115, but not Safari 16.5:
+  // <input aria-labelledby="lbl">
+  // <div id="lbl" role="progressbar" aria-valuenow="23"></div>
+  accessibleNameFromFieldValue: ['progressbar']
 };
-
-export default unsupported;
--- a/test/integration/rules/label/label.html
+++ b/test/integration/rules/label/label.html
@@ -5,67 +5,72 @@
   <input type="submit" id="na4" />
   <input type="reset" id="na5" />
   <input type="text" id="fail1" />
-  <textarea id="fail3"></textarea>
+  <textarea id="fail2"></textarea>
   <input type="text" aria-label="label" id="pass1" />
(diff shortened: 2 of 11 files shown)


test/integration/rules/label/label.html
```diff
@@ -5,67 +5,85 @@
...
+  <input id="pass9" title="Label" />
+  <textarea id="pass10" title="Label"></textarea>
 
   <label
-    ><label><input type="text" id="fail22" /></label
+    ><label><input type="text" id="fail8" /></label
   ></label>
 
   <div>
-    <label>
```
WilcoFiers, line 35: This label does nothing, and fail24 is not tested by this rule; this is not the select-name rule.

lib/commons/text/unsupported.js
```diff
@@ -1,5 +1,7 @@
-const unsupported = {
-  accessibleNameFromFieldValue: ['combobox', 'listbox', 'progressbar']
+export default {
+  // Element's who's value is not consistently picked up in the accessible name
+  // Supported in Chrome 114, Firefox 115, but not Safari 16.5:
+  // <input aria-labelledby="lbl">
+  // <div id="lbl" role="progressbar" aria-valuenow="23"></div>
+  accessibleNameFromFieldValue: ['progressbar']
```
straker, line 6: Should we have a test for `combobox` and `listbox` now being supported? I see you changed a test from `combobox` to `listbox` but not any added tests to show they are supported.

straker (changes requested):

straker (approved):

PR #4098, opened 2023-07-21
fix: Use correct copyright year

stephenmathieson:
This patch corrects the copyright year included in the built `axe.js` and `axe.min.js` files. Deque Systems, Inc has held the copyright for axe-core since 2015.

### DIFF
--- a/lib/intro.stub
+++ b/lib/intro.stub
@@ -1,5 +1,5 @@
 /*! axe v<%= pkg.version %>
- * Copyright (c) <%= grunt.template.today("yyyy") %> Deque Systems, Inc.
+ * Copyright (c) 2015 Deque Systems, Inc.
  *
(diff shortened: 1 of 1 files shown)


lib/intro.stub
```diff
@@ -1,5 +1,5 @@
 /*! axe v<%= pkg.version %>
- * Copyright (c) <%= grunt.template.today("yyyy") %> Deque Systems, Inc.
+ * Copyright (c) 2015 Deque Systems, Inc.
```
WilcoFiers, line 2: Lets do this, so we have the date of the last update in here. Makes the files not seem old.

```suggestion
 * Copyright (c) 2015 - <%= grunt.template.today("yyyy") %> Deque Systems, Inc.
```

WilcoFiers (approved):

Issue #3923, opened 2023-02-23
createGrid consumes excessive memory and crashes on pages with extremely large elements


dbjorge:
### Product

axe-core

### Product Version

4.6.3

### Latest Version

- [X] I have tested the issue with the latest version of the product

#### Overview

Today, `createGrid` understands to avoid expanding grids for elements which are not visible:

https://github.com/dequelabs/axe-core/blob/d4522cdd7a90018336098f9b2119e353bd5a71ee/lib/commons/dom/create-grid.js#L87-L93

However, for elements which are partially visible and extremely large, this check isn't quite enough to avoid an undesirable explosion of grid size.

The motivating example is [the Monaco editor](https://microsoft.github.io/monaco-editor/playground.html?source=v0.35.0#example-creating-the-editor-hello-world). To render its main content area's background, it uses an element which is styled with `width: 1e6; height: 1e6` (the linked playground page contains 3 editor areas, so it has 3 such elements).

The presence of these elements causes `axe-core` scans to eat up an enormous amount of memory (~2.5GB) and time (~4s wall time on my dev box), and often causes the whole target page to crash. Almost all of the perf hit comes from call stack `createGrid > addNodeToGrid > loopGridPosition > loopNegativeIndexMatrix > loopNegativeIndexMatrix`, where `axe-core` is allocating a separate JS object for each cell of a roughly 5000x5000 grid, almost every cell of which is not visible on screen.

I would imagine that making `createGrid` cooperate with elements like this would probably require coming up with a means to calculate the specific clipping of an element's clientRect(s) which are visible on screen and only add that to the grid, rather than adding the entire clientRect(s) whenever any part of the element is visible. I bet that's trickier than it sounds, though...

#### Expectation
`axe-core` should be able to scan [the repro page in question](https://microsoft.github.io/monaco-editor/playground.html?source=v0.35.0#example-creating-the-editor-hello-world) with performance comparable to `axe-core@4.4.3`

#### Actual
Scanning the repro page takes ~2.5GB JS heap, ~4s wall time on my dev box, and often crashes the page entirely

#### How to Reproduce

* In Chromium, open [the Monaco editor](https://microsoft.github.io/monaco-editor/playground.html?source=v0.35.0#example-creating-the-editor-hello-world)
* Open the F12 dev tools' "Performance" tab and start a trace
[...]

PR #4101, opened 2023-07-24
fix(commons/dom/createGrid): only add the visible, non-overflow areas of an element to the grid

straker:
Implementing https://github.com/dequelabs/axe-core/issues/3923#issuecomment-1624013523

Closes: #3923

### DIFF
--- a/test/integration/full/contrast/memory.html
+++ b/test/integration/full/contrast/memory.html
@@ -1,0 +1,45 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <title>Test Page</title>
+    <link
+      rel="stylesheet"
+      type="text/css"
+      href="/node_modules/mocha/mocha.css"
+    />
+    <script src="/node_modules/mocha/mocha.js"></script>
+    <script src="/node_modules/chai/chai.js"></script>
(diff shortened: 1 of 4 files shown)


WilcoFiers (dismissed):
I think it would be good to have an integration test that demonstrates the memory problem, and shows it's solved. Have you tried that?

test/integration/full/contrast/memory.html
```diff
@@ -0,0 +1,45 @@
...
+      rel="stylesheet"
+      type="text/css"
+      href="/node_modules/mocha/mocha.css"
+    />
+    <script src="/node_modules/mocha/mocha.js"></script>
+    <script src="/node_modules/chai/chai.js"></script>
+    <script src="/axe.js"></script>
+    <script>
+      mocha.setup({
+        timeout: 10000,
```
WilcoFiers, line 15: This seems a little high to me. I think I'd prefer to start this off lower, if it turns out flakey we can bump it up. WDYT?

```suggestion
        timeout: 2000,
```

WilcoFiers (approved):
Left a suggestion though.

WilcoFiers (approved):

Issue #3913, opened 2023-02-13
isIconLigature fails on leading invisible character


SirUppyPancakes:
### Product

axe-core

### Product Version

4.6.3

### Latest Version

- [X] I have tested the issue with the latest version of the product

#### Expectation
Expected `axe-core` to be able to handle any valid HTML (the HTML in this case is 100% valid according to the W3C validator)

#### Actual
The `isIconLigature` function fails to consider invisible characters at the start of the input, resulting in a `SupportError` being thrown.

#### How to Reproduce
Project which reproduces the issue: 
[axe-core-invisible-issue.zip](https://github.com/dequelabs/axe-core/files/10725677/axe-core-invisible-issue.zip)

Steps:
1. Run `npm install`
2. Start a server to serve the html file (I used the `html-server` npm package, for example) or simply open the html file in a browser
3. Optionally put some breakpoints (in `node_modules/axe-core/axe.js`) to observe the exact source of the issue, on the following lines:
    - 11378 `var compareData = new Uint32Array(canvasContext.getImageData(0, 0, width, fontSize).data.buffer);`
    - 11379 `if (!compareData.some(function(pixel) {`
    - 22005 `reject(new SupportError({`

After hitting the break point right before `isIconLigature` calls `canvasContext.getImageData`, you'll be able to see that `width` and `fontSize` have been determined to be `NaN` and `Infinity` which are invalid parameters for this function, causing it to throw an exception and result in the `SupportError` being generated. This happens because the first character in the link content of my HTML in this case is [U+2069] which is an invisible character relating to the left-to-right or right-to-left nature of the text (we ran into this issue when checking one of our Arabic-translated files).

Would it be possible to change `isIconLigature` to look for the first visible character? I suppose there would also need to be a case to handle if there isn't any visible characters in the span as well...

straker:
Thanks for the issue. For security reasons I can't open a zip file. However, thanks to the description I can reproduce this on my end using the HTML string `&#2069Hello World`.

SirUppyPancakes:
Thanks! For future reference for myself, what would have been a preferable way to share a test project? A gist maybe?

straker:
A gist or HTML snippet work best.

padmavemulapati:
Validated with the latest axe-core develop branch code base, not seeing any failure with the ligature icon on invisible charecters, 
 for ex: ` <main> <a href="https://www.google.com">⁩البطاطس</a > </main>`



[...]

PR #4102, opened 2023-07-24
fix(color-contrast): ignore format unicode characters

straker:
Format characters such as zero-width spaces were not checked in the `nonBMP` section of the unicode functions. I generated the regex by using [regenerate](https://github.com/mathiasbynens/regenerate) and adding all [format unicode to it](https://www.compart.com/en/unicode/category/Cf).

In making this change I also discovered a bug in `hasUnicode` in that it only checked the first option that matched and not all options due to the early return style. So trying to check if a string contained punctuation while also checking if it had emoji would only ever check if it had emoji. Fixed that while I was here.

Closes: #3913

### DIFF
--- a/test/commons/text/unicode.js
+++ b/test/commons/text/unicode.js
@@ -1,86 +1,93 @@
-describe('text.hasUnicode', function () {
-  describe('text.hasUnicode, characters of type Non Bi Multilingual Plane', function () {
-    it('returns false when given string is alphanumeric', function () {
-      var actual = axe.commons.text.hasUnicode('1 apple', {
+describe('text.hasUnicode', () => {
+  describe('text.hasUnicode, characters of type Non Bi Multilingual Plane', () => {
+    it('returns false when given string is alphanumeric', () => {
+      const actual = axe.commons.text.hasUnicode('1 apple', {
         nonBmp: true
       });
       assert.isFalse(actual);
     });
 
-    it('returns false when given string is number', function () {
-      var actual = axe.commons.text.hasUnicode('100', {
+    it('returns false when given string is number', () => {
+      const actual = axe.commons.text.hasUnicode('100', {
         nonBmp: true
       });
       assert.isFalse(actual);
     });
 
-    it('returns false when given string is a sentence', function () {
-      var actual = axe.commons.text.hasUnicode('Earth is round', {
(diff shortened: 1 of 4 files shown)


test/commons/text/unicode.js
```diff
@@ -1,86 +1,94 @@
...
+    it('returns true for a string with characters in supplementary private use area A', () => {
+      const actual = axe.commons.text.hasUnicode('\uDB80\uDFFE', {
+        nonBmp: true
+      });
+      assert.isTrue(actual);
+    });
+
+    it('returns true when given string has format unicode', () => {
+      // the front of this string contains a zero-width character
+      const actual = axe.commons.text.hasUnicode('⁩Hello World', {
```
WilcoFiers, line 82: Instead of putting a comment and hoping this work, how about we use `\u` here?

```suggestion
      // zero-width spacer character U+200B
      const actual = axe.commons.text.hasUnicode('\u200BHello World', {
```

test/commons/text/unicode.js
```diff
@@ -235,4 +262,12 @@ describe('text.removeUnicode', function () {
     );
     assert.equal(actual, 'The  is orange the  is white');
   });
+
+  it('returns string removing format unicode', () => {
+    // the front of this string contains a zero-width character
+    const actual = axe.commons.text.removeUnicode('⁩Hello World', {
```
WilcoFiers, line 268:
```suggestion
    // zero-width spacer character U+200B
    const actual = axe.commons.text.removeUnicode('\u200BHello World', {
```

test/commons/text/unicode.js
```diff
@@ -171,61 +179,80 @@ describe('text.hasUnicode', function () {
       );
       assert.isTrue(actual);
     });
+
+    it('returns true when given format unicode characters', () => {
+      // the front of this string contains a zero-width character
+      const actual = axe.commons.text.hasUnicode('⁩Hello World', {
```
WilcoFiers, line 185:
```suggestion
      // zero-width spacer character U+200B
      const actual = axe.commons.text.hasUnicode('\u200BHello World', {
```

WilcoFiers (approved):
I'm okay with this approach, but had an idea you might appreciate.

WilcoFiers (approved):

Issue #3918, opened 2023-02-21
color-contrast check: getImageData fails with value is not of type long


mattzeunert:
### Product

axe-core

### Product Version

4.6.3

### Latest Version

- [X] I have tested the issue with the latest version of the product

#### Expectation

Run an accessibility audit without errors.

#### Actual

Getting this error: 

> "`TypeError: Failed to execute 'getImageData' on 'CanvasRenderingContext2D': Value is not of type 'long'.` - feature unsupported in your environment. Skipping color-contrast rule."

#### How to Reproduce

1. Go to https://simpletire.com/
2. Paste axe-min code
3. Run this code:

```
axe
  .run()
  .then(results => {
    if (results.violations.length) {
        console.log(results.incomplete.find(item => item.id === "color-contrast"))
    }
  })
  ```

![](https://user-images.githubusercontent.com/1303660/220357133-80ab1ff4-5e80-46b5-bdea-8999e948e8aa.png)

#### Additional context

Width is zero / font size is infinity. Maybe because the custom font can't be found?

![](https://user-images.githubusercontent.com/1303660/220357220-b5090b12-7a69-42fc-ad4e-c0fe29add5b2.png)

**Comment by straker** [comment]
Thanks for the detailed issue. Looks like the affirm font is a font icon and doesn't have all the characters, which makes this similar to https://github.com/dequelabs/axe-core/issues/3913. Since that one deals with leading invisible unicode characters, I'll leave this one open since it deals with a font icon with missing characters.

**Comment by adamraine** [comment]
Hello we are seeing this error on https://cloud.google.com/gcp as well. I haven't tested on all axe versions but this error is not present in v4.4.1.

**Comment by Shagon1k** [comment]
Hello,
Any updates regarding this one? Seems like still not working with the "low-contrast" issue returning NaN.

**Comment by WilcoFiers** [comment]
@Shagon1k A fix for this is planned to be done in the 4.8 release scheduled for August 7th.

**Comment by straker** [comment]
@mattzeunert Would you happen to know the copyright of the Affirm logo font?

**Comment by mattzeunert** [comment]
No idea sorry.

**Comment by padmavemulapati** [comment]
Validated with axe-core develop branch code base, seeing `color-contrast` issues on the page, `https://simpletire.com/` when running with latest `axe.min.js`
and executed the axe.run as follows:
```
axe
  .run()
  .then(results => {
    if (results.violations.length) {
[...]

PR #4103, opened 2023-07-24
fix(color-contrast): ignore zero width characters

straker:
I grabbed the Affirm font from the provided website. I couldn't figure out how to generate a font that had 0 width characters so I hope it's OK for us to grab it for test purposes. 

Closes: #3918


lib/commons/text/is-icon-ligature.js
```diff
@@ -143,6 +139,14 @@ export default function isIconLigature(
   const firstChar = nodeValue.charAt(0);
   let width = canvasContext.measureText(firstChar).width;
 
+  // if the font uses zero-width characters that are not unicode characters
+  // it has to be an icon ligature (e.g. Affirm)
+  // @see https://github.com/dequelabs/axe-core/issues/3918
+  if (width === 0) {
+    font.numLigatures++;
+    return true;
```
WilcoFiers, line 147: Why do you think a zero-width char is a ligature font? Unicode has at least one character that is supposed to have zero width (the zero-width space). There might be others, I'm not particularly well versed on unicode. This seems like a false assumption. Would it be better to loop over characters until you found one that has space?

test/assets/Affirm.woff
WilcoFiers, line 1: Are we allowed to copy this into our repo?

WilcoFiers (dismissed):

lib/commons/text/is-icon-ligature.js
straker: https://github.com/dequelabs/axe-core/pull/4102 will solve the zero-width unicode problem

test/assets/Affirm.woff
dbjorge: I went ahead and replaced it with a purpose-built font that just has a zero-width `0` character and nothing else

lib/commons/text/is-icon-ligature.js
```diff
@@ -143,6 +139,14 @@ export default function isIconLigature(
   const firstChar = nodeValue.charAt(0);
   let width = canvasContext.measureText(firstChar).width;
 
+  // if the font uses zero-width characters that are not unicode characters
+  // it has to be an icon ligature (e.g. Affirm)
```
dbjorge, line 143:
```suggestion
  // we already checked for typical zero-width unicode formatting characters further up,
  // so we assume that any remaining zero-width characters are part of an icon ligature
```

dbjorge (approved):
Updated inline with a non-copyright-encumbered custom font per our discussion offline. Minor suggestion on one of the new comments, otherwise LGTM!

dbjorge (approved):

Issue #3943, opened 2023-03-13
"Document should not have more than one banner landmark" when page has no banners


luis-pato:
### Product

axe-core

### Product Version

Chrome extension + axe-core 4.6.3 on Cypress

### Latest Version

- [X] I have tested the issue with the latest version of the product

#### Expectation
I am getting the "_Document should not have more than one banner landmark_" error on `header` elements placed inside webcomponents with the `role="article"`

When i test the elements on a storybook page, both with the Chrome extension and with Cypress (using axe-core), i get this error:

![image](https://user-images.githubusercontent.com/42058106/224625753-647e5924-9794-463a-8779-d032f7cc4446.png)

I understand that the `<header>` element has a role similar to the banner, unless it is a descendent of an `<aside>`, `<article>`, `<main>`, ... (like mentioned [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/banner_role))

These `<header>` elements are not direct descendents of any of these tags, but they are a direct descendent of a webcomponent with a `role="article"`, so i expected them to have the same behavior as if they were inside an `<article>` tag

![image](https://user-images.githubusercontent.com/42058106/224626470-e7763290-b045-44a0-a7e6-46c93573102d.png)

#### How to Reproduce
Unless i am understanding this issue incorrectly, a `<header>` element inside an `<article>` should not be a problem. I also think if it is inside a webcomponent with the `role="article"` it shouldn't throw the error so:

1. This should throw the error
```
<my-component>
    <header>....</header>
</my-component>

<my-component>
    <header>....</header>
</my-component>
```

2. This should not throw the error
```
<article>
    <header>....</header>
</article>

<article>
    <header>....</header>
</article>
```

3. This also should not throw the error
```
<my-component role="article">
    <header>....</header>
</my-component>

<my-component role="article">
    <header>....</header>
</my-component>
```

straker:
Thanks for the issue. Looks like this is a problem with the [page-no-duplicate-banner check](https://github.com/dequelabs/axe-core/blob/develop/lib/checks/keyboard/page-no-duplicate-banner.json#L7). We don't look at any aria roles in the `nativeScopeFilter` option (which is used to exclude elements which are descendants of any of those selectors).

padmavemulapati:
Validated with the latest axe-core develop branch code base,

code: is failing for  `landmark-no-duplicate-banner` rule
```
<my-component>
    <header>....</header>
</my-component>

<my-component>
[...]

PR #4105, opened 2023-07-25
feat(page-no-duplicate-banner/contentinfo): deprecate options.nativeScopeFilter, take into ancestors with sectioning roles

straker:
Our `nativeScopeFilter` selector only accounted for native elements and not elements with sectioning roles. Instead of adding sectioning roles to the selector, I  decided to add the new `role` option to let the check look at the role of the element, which computes the implicit html role based on [sectioning ancestors](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/standards/implicit-html-roles.js#L16-L25). That way our selectors can't become out-of-sync and cause issues in the future.

I also deprecated the `nativeScopeFilter` option since we shouldn't be using it over the `role` option. 

Closes: #3943


WilcoFiers (dismissed):
Can you add integration tests plz.

WilcoFiers (approved):

Issue #3955, opened 2023-03-23
New rule: restrict use of aria-description and aria-braille*


straker:
Axe-core 4.7 will add the `aria-description` and all `aria-brille*` attributes as valid ARIA attributes. Once that is done, we'll need to create a new rule (or two) that limits their use to elements that have an ARIA role and an accessible name.

WilcoFiers:
I looked at this more closely. I think the right way to do this would be for ARIA to prohibit these attributes for `none` / `presentation` / `generic`. That covers the "role" scenario. I've opened an issue for it here: https://github.com/w3c/aria/issues/1988

What I think is left is a rule that checks that you can only use aria-braillelabel on elements with a non-empty accessible name, and that you can only use aria-brailleroledescription on elements with a non-empty aria-roledescription.

padmavemulapati:
Verified on axe-core develop branch code base, 
for the test code snippet

```
<button id="fail1" aria-braillelabel="hello"></button>
<aside
  id="fail2"
  aria-roledescription=""
  aria-brailleroledescription="table of contents"></aside>

```

Seeing a new rule `aria-braille-equivalent`

```
description: "Ensure aria-braillelabel and aria-brailleroledescription have a non-braille equivalent"
help: "aria-braille attributes must have a non-braille equivalent"
helpUrl: "https://dequeuniversity.com/rules/axe/4.7/aria-braille-equivalent?application=axeAPI"
id: "aria-braille-equivalent"
impact: "serious"
nodes: Array(2)0: {any: Array(0), all: Array(1), none: Array(0), impact: 'serious', html: '<button id="fail1" aria-braillelabel="hello"></button>', …}
1: {any: Array(0), all: Array(1), none: Array(0), impact: 'serious', html: '<aside id="fail2" aria-roledescription="" aria-brailleroledescription="table of contents"></aside>', …}
tags:  ['cat.aria', 'wcag2a', 'wcag412', 'EN-301-549', 'EN-9.4.1.2']

```



![Image](https://github.com/dequelabs/axe-core/assets/47409579/c2a04783-5e8f-43f6-a6b6-a25e7d22a69e)

PR #4107, opened 2023-07-28
feat(new-rule): aria-braille-equivalent finds incorrect uses of aria-braille attributes

WilcoFiers:
This rule effectively checks two things:

1. aria-braillelabel is only ever used on elements with an accessible name
2. aria-brailleroledescription is only ever used on elements with aria-roledescription

Closes: #3955


lib/checks/aria/braille-label-equivalent-evaluate.js
```diff
@@ -0,0 +1,18 @@
...
+ * @memberof checks
+ * @return {Boolean}
+ */
+export default function brailleLabelEquivalentEvaluate(
+  node,
+  options,
+  virtualNode
+) {
+  const brailleLabel = virtualNode.attr('aria-braillelabel') ?? '';
+  if (!brailleLabel.trim()) {
```
straker, line 14: Should we use `sanitize` from our text functions here instead (similar to aria-label and aria-labelledby checks)?

lib/checks/aria/braille-label-equivalent.json
```diff
@@ -0,0 +1,11 @@
+{
+  "id": "braille-label-equivalent",
+  "evaluate": "braille-label-equivalent-evaluate",
+  "metadata": {
+    "impact": "serious",
+    "messages": {
+      "pass": "aria-braillelabel is not used on an element with no accessible text",
```
straker, line 7: Just simplifying the double negative.

```suggestion
      "pass": "aria-braillelabel is used on an element with accessible text",
```

lib/checks/aria/braille-roledescription-equivalent-evaluate.js
```diff
@@ -0,0 +1,27 @@
...
+ * @memberof checks
+ * @return {Boolean}
+ */
+export default function brailleRoleDescriptionEquivalentEvaluate(
+  node,
+  options,
+  virtualNode
+) {
+  const brailleRoleDesc = virtualNode.attr('aria-brailleroledescription') ?? '';
+  if (!brailleRoleDesc.trim()) {
```
straker, line 13: Ditto for `sanitize`

lib/checks/aria/braille-roledescription-equivalent-evaluate.js
```diff
@@ -0,0 +1,27 @@
...
+    return true;
+  }
+  const roleDesc = virtualNode.attr('aria-roledescription');
+  if (typeof roleDesc !== 'string') {
+    this.data({ messageKey: 'noRoleDescription' });
+    return false;
+  }
+
+  if (roleDesc.trim() === '') {
```
straker, line 22: Should keep consistent use of if you're using `!brailleRoleDesc.trim()` or `brailleRoleDesc.trim() === '')`

lib/checks/aria/braille-roledescription-equivalent.json
```diff
@@ -0,0 +1,14 @@
+{
+  "id": "braille-roledescription-equivalent",
+  "evaluate": "braille-roledescription-equivalent-evaluate",
+  "metadata": {
+    "impact": "serious",
+    "messages": {
+      "pass": "aria-brailleroledescription is not used on an element with no accessible text",
```
straker, line 7:
```suggestion
      "pass": "aria-brailleroledescription is used on an element with aria-roledescription",
```

test/integration/virtual-rules/aria-braille-equivalent.js
```diff
@@ -0,0 +1,61 @@
+describe('aria-braille-equivalent virtual-rule', () => {
+  afterEach(() => {
+    axe.reset();
+  });
```
straker, line 4: I believe this isn't necessary 

```suggestion
```

test/integration/virtual-rules/aria-braille-equivalent.js
```diff
@@ -0,0 +1,61 @@
...
+        'aria-brailleroledescription': 'Hello world'
+      }
+    });
+
+    assert.lengthOf(results.passes, 1);
+    assert.lengthOf(results.violations, 0);
+    assert.lengthOf(results.incomplete, 0);
+  });
+
+  it('fails when roledescription is empty but brailleroledescription is not', () => {
```
straker, line 48: I believe `accessibleTextVirtual` can throw in subtree text if the element doesn't have a child (See https://github.com/dequelabs/axe-core/blob/develop/test/integration/virtual-rules/button-name.js#L116-L125). If that's the case, we should add a test for incomplete.

straker (dismissed):

lib/checks/aria/braille-label-equivalent-evaluate.js
WilcoFiers: We probably shouldn't use it for aria-labelledby either. What sanitize does that trim doesn't do is replace duplicate whitespace characters in between words with single space characters. We don't need that here.

test/checks/aria/braille-label-equivalent.js
```diff
@@ -0,0 +1,51 @@
...
+    const params = checkSetup(
+      '<img id="target" alt="" aria-braillelabel=" \r\t\n " />'
+    );
+    assert.isTrue(checkEvaluate.apply(checkContext, params));
+  });
+
+  describe('when aria-braillelabel has text', () => {
+    it('returns false when the accessible name is empty', () => {
+      const params = checkSetup(`
+        <img id="target" alt="" aria-braillelabel="foo" />
```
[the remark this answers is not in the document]
WilcoFiers: I think this should fail. There's no reason someone would put `aria-braillelabel` on an element like this. We might change how that works if this attribute ever goes on the prohibited attrs list, which is what I asked the ARIA WG about. I think that would be more appropriate. Until then I think this is reasonable.

test/integration/virtual-rules/aria-braille-equivalent.js
WilcoFiers: Configuration isn't reverted until axe.reset() is called. It's not in testutils. Where would you think this might happen?

straker (dismissed):

straker (approved):

PR #4110, opened 2023-07-31
fix(aria-allowed-attr): pass aria-expanded on checkbox & switch

### LINKED ISSUE #3339: roles that allow aria-expanded and aria-roles.js question

WilcoFiers:
WAI-ARIA 1.3 has proposed to dramatically change which roles allow aria-expanded. Per this PR, `checkbox` and `switch` now allow this attribute. Quite a few more roles no longer allow `aria-expanded`.

I've opened up an issue requesting that ARIA 1.3 deprecate aria-expanded on those roles, instead of simply removing them. I think simply removing them would create a lot of unnecessary friction. If the W3C decides not to stay backward compatible we may have to figure out a different way to present this, but by far the best solution here would be to use deprecation, which is what they've also done in other places. See https://github.com/w3c/aria/issues/1990

Closes #3339, closes #3343

### DIFF
--- a/lib/standards/aria-roles.js
+++ b/lib/standards/aria-roles.js
@@ -75,9 +75,9 @@ const ariaRoles = {
     allowedAttrs: [
       'aria-colindex',
       'aria-colspan',
-      'aria-expanded',
       'aria-rowindex',
-      'aria-rowspan'
+      'aria-rowspan',
+      'aria-expanded'
     ],
     superclassRole: ['section'],
     nameFromContent: true
@@ -197,9 +197,9 @@ const ariaRoles = {
   },
   feed: {
     type: 'structure',
+    requiredOwned: ['article'],
     // Aria-expanded removed as of 1.3
     allowedAttrs: ['aria-expanded'],
-    requiredOwned: ['article'],
     superclassRole: ['list']
   },
   figure: {
--- a/lib/standards/aria-roles.js
+++ b/lib/standards/aria-roles.js
@@ -812,8 +812,6 @@ const ariaRoles = {
   },
   time: {
     type: 'structure',
-    // Aria-expanded removed as of 1.3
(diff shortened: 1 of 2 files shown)


lib/standards/aria-roles.js
```diff
@@ -775,15 +812,19 @@ const ariaRoles = {
   },
   time: {
     type: 'structure',
+    // Aria-expanded removed as of 1.3
+    allowedAttrs: ['aria-expanded'],
```
straker, line 816: Did you mean to add this?

```suggestion
```

straker (changes requested):

straker (approved):

dbjorge (approved):

Issue #4066, opened 2023-06-28
Give rules a consistent impact


WilcoFiers:
The impact of a violation is currently determined by which checks fail. The idea of it was to give some flexibility in setting the impact within the same rule. This never worked well though. Since impact is bound up with the checks of a rule, and  which checks run is driven by the remediation message we want to give, axe-core's architecture gives very little control over the impact.

As variable impact is also a common source of confusion, we're going to change this so that a single rule only ever outputs one level of impact. It is already possible to set the impact on a rule, although we aren't using it much.

Since impact is used in many UIs, and we don't want the numbers to drastically change by this release we're going to have to put a few checks into separate rules

## Prerequisites

- [x] #1325
- [x] #4067
- [x] #4068
- [x] #4069
- [x] #4070

## Make impact a required prop on rules/*.json

Once all the above tickets are done, the work for this ticket will be to give each rule an `impact` property, and make it required in the build tool. We can then do some cleanup in how rule-descriptions.md is generated, as that won't need to look at checks anymore.

PR #4114, opened 2023-08-01
feat: Deprecate impact on checks, use rules instead

WilcoFiers:
Impact is now required on rules, rather than on checks. This makes impact on checks obsolete. As of axe-core 4.8, impact on checks is **deprecated**.

Closes: #4066


straker (approved):

straker (changes requested):
Missing the following rules, which means the rule validator is not working / running?

- `aria-roles`
- `aria-valid-attr-value`
- `scope-attr-valid`

straker (approved):
Turns out they were already updated (but my local branch was not)

PR #4116, opened 2023-08-01
fix(jsdom): allow axe.setup() without a global window

### LINKED ISSUE #3962: axe.setup should set globals
Axe-core's setup and teardown methods should be able to work with JSDOM:

```js
import axe from 'axe-core'
import { JSDOM } from 'jsdom'
const { window } = new JSDOM(`<!DOCTYPE html>\n<html></html>`);
axe.setup(window)
```

Currently the only way to use `axe.setup()` with JSDOM is if `globalThis.window` and `globalThis.document` are defined before axe-core is imported. This is annoying and shouldn't be necessary. We handle this in `axe.run()` by calling `setupGlobals`. Probably this call should just be moved into `axe.setup()`, and the `resetGlobals()` call should go into `axe.teardown()`.

Another problem I'm spotting is that `axe.runPartial()` doesn't set up globals either. Moving these calls to setup/teardown will solve that too. A test for this would be useful.

WilcoFiers:
`axe.setup()` should set up temporary globals when `globalThis.window` and/or `globalThis.document` aren't set up. This PR fixes that.

Closes #3962

### DIFF
--- a/lib/core/public/run.js
+++ b/lib/core/public/run.js
@@ -1,6 +1,6 @@
 import { getReporter } from './reporter';
 import normalizeRunParams from './run/normalize-run-params';
-import { setupGlobals, resetGlobals } from './run/globals-setup';
+import { setupGlobals } from './run/globals-setup';
 import { assert } from '../utils';
 
 const noop = () => {};
@@ -36,10 +36,10 @@ export default function run(...args) {
     axe.utils.performanceTimer.start();
   }
 
-  function handleRunRules(rawResults, cleanup) {
+  function handleRunRules(rawResults, teardown) {
     const respond = results => {
       axe._running = false;
-      cleanup();
+      teardown();
       try {
         callback(null, results);
       } catch (e) {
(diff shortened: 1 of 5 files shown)


lib/core/public/run.js
```diff
@@ -36,10 +36,10 @@ export default function run(...args) {
     axe.utils.performanceTimer.start();
   }
 
-  function handleRunRules(rawResults, cleanup) {
+  function handleRunRules(rawResults, teardown) {
```
WilcoFiers, line 39: Since this is actually just the teardown method, it's confusing to have it named something else here.

lib/core/public/run.js
```diff
@@ -66,7 +66,6 @@ export default function run(...args) {
       axe.utils.performanceTimer.end();
     }
     axe._running = false;
-    resetGlobals();
```
WilcoFiers, line 69: This is already called in `teardown`, which is invoked earlier. Calling `resetGlobals` here does nothing, and wouldn't actually have worked because axe's cache has been cleaned up by this point. Same is true for the other `resetGlobals()` call in this file.

straker (approved):

Issue #3891, opened 2023-01-25
Update target-size to the latest WCAG spec


WilcoFiers:
Since we created `target-size` the WCAG spec has been updated, specifically some improvements were made around not measuring diagonally when elements are horizontally or vertically adjacent. That was one of the difficulties we had.

We'll need to update axe-core to align.

PR #4117, opened 2023-08-02
fix(target-size): update to match new spacing requirements

straker:
This mostly coverts the target-offset check to use the updated spacing exception from the spec. However, going over this we did notice that we still might be off in determining the center of the bounding rect is. The code currently only looks at the largest target rect but we are thinking we need to take into account the bounding box for all target rects and center on that. Will leave that as a tech debt ticket for after 4.8 so we can get this merged in on time. 

Another tech debt ticket that needs to be created is to update the target-size check to use the new `getTargetSize` function.

Closes: #3891


lib/commons/math/get-offset.js
```diff
@@ -95,55 +69,5 @@ function getClosestPoint({ x, y }, ownRect, adjacentRect) {
...
-  } else if (!closestX) {
-    return { x, y: closestY };
-  }
-  if (Math.abs(x - closestX) < Math.abs(y - closestY)) {
-    return { x: closestX, y };
-  } else {
-    return { x, y: closestY };
-  }
-}
+  return Math.hypot(pointA.x - pointB.x, pointA.y - pointB.y);
```
straker, line 68: This function exists in IE11 (I've verified)

lib/commons/math/split-rects.js
```diff
@@ -37,19 +37,33 @@ function splitRect(inputRect, clipRect) {
...
 
 function computeRect(baseRect) {
-  return {
-    ...baseRect,
-    x: baseRect.left,
-    y: baseRect.top,
-    height: baseRect.bottom - baseRect.top,
-    width: baseRect.right - baseRect.left
-  };
+  return new window.DOMRect(
```
straker, line 54: I though it was weird that split rects works on DOMRects but then doesn't return one, which meant that doing `splitRect(rect1, rect2).left` would return `undefined` instead of the `x` value.

test/commons/math/get-offset.js
```diff
@@ -1,104 +1,66 @@
-describe('getOffset', function () {
+describe('getOffset', () => {
   'use strict';
```
WilcoFiers, line 2:
```suggestion
```

test/commons/math/get-offset.js
```diff
@@ -1,104 +1,66 @@
...
-    assert.closeTo(getOffset(nodeA, nodeB), getDiagonal(40), round);
-    assert.closeTo(getOffset(nodeB, nodeA), getDiagonal(30), round);
+  it('returns center to corner or square when at a diagonal', () => {
+    const fixture = fixtureSetup(`
+      <button style="width: 10px; height: 10px; margin: 0; position: absolute; top: 0; left: 0">&nbsp;</button>
+      <button style="width: 30px; height: 30px; margin: 0; position: absolute; top: 50px; left: 50px">&nbsp;</button>
+    `);
+    const nodeA = fixture.children[1];
+    const nodeB = fixture.children[3];
+    assert.closeTo(getOffset(nodeA, nodeB), 61.5, round);
```
WilcoFiers, line 34: Is that right? You're going from x:5,y:5 to x:50,y:50. `Math.hypot(45, 45)` is 63.63.... Why is this 61.5?

lib/commons/math/get-offset.js
```diff
@@ -1,91 +1,61 @@
...
-    const endDistance = Math.abs(centerB - rectA[end]);
-    if (startDistance >= endDistance) {
-      farthestPoint[axis] = rectA[start]; // left | top
-    } else {
-      farthestPoint[axis] = rectA[end]; // right | bottom
-    }
-  });
-  return farthestPoint;
-}
+export default function getOffset(vNodeA, vNodeB, radius = 12) {
```
WilcoFiers, line 12: Radius could do with a more explicit name.

```suggestion
export default function getOffset(vNodeA, vNodeB, minRadiusNeighbour = 12) {
```

test/commons/math/get-offset.js
```diff
@@ -1,104 +1,66 @@
...
-    var fixture = fixtureSetup(
-      '<a style="padding:10px; display:inline-block">' +
-        '<b style="width:20px; height: 30px; display:inline-block">&nbsp;</b>' +
-        '</a>'
-    );
-    var nodeA = fixture.children[0];
-    var nodeB = nodeA.children[0];
-    assert.closeTo(getOffset(nodeA, nodeB), getDiagonal(10), round);
-    assert.closeTo(getOffset(nodeB, nodeA), 10, round);
+  it('allows passing radius', () => {
```
WilcoFiers, line 57: This test name isn't very clear. Maybe:

```suggestion
  it('subtracts minNeighbourRadius from center-to-center calculations', () => {
```

test/commons/math/split-rects.js
```diff
@@ -1,92 +1,98 @@
-describe('splitRects', function () {
-  var splitRects = axe.commons.math.splitRects;
+describe('splitRects', () => {
+  const splitRects = axe.commons.math.splitRects;
   function createRect(x, y, width, height) {
```
WilcoFiers, line 3: I don't think we need this if all we're doing is wrap a constructor. I'm not sure but I assume this was before I knew about DOMRect?

WilcoFiers (changes requested):

WilcoFiers (changes requested):
Needs updated integration tests

test/commons/math/get-offset.js
```diff
@@ -1,104 +1,66 @@
...
-    assert.closeTo(getOffset(nodeA, nodeB), getDiagonal(40, 5), round);
-    assert.closeTo(getOffset(nodeB, nodeA), 30, round);
+  it('returns center to edge of circle when both are undersized', () => {
+    const fixture = fixtureSetup(`
+      <button style="width: 10px; height: 10px; margin: 0; position: absolute; top: 0; left: 0">&nbsp;</button>
+      <button style="width: 10px; height: 10px; margin: 0; position: absolute; top: 50px; left: 0">&nbsp;</button>
+    `);
+    const nodeA = fixture.children[1];
+    const nodeB = fixture.children[3];
+    assert.closeTo(getOffset(nodeA, nodeB), 38, round);
```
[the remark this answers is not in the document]
straker: It's because the distance between the the centers of both buttons gets [subtracted by the circle radius](https://github.com/dequelabs/axe-core/pull/4117/files#diff-7056fdbaecbbd8da82e47e708803e085002edc843821a06fa5930c68514f201fR33) of 12 (50 - 12).

lib/commons/dom/get-target-size.js
```diff
@@ -0,0 +1,70 @@
...
+    return (
+      vNeighbor.getComputedStylePropertyValue('pointer-events') !== 'none' &&
+      hasVisualOverlap(vNode, vNeighbor)
+    );
+  });
+
+  if (!overlappingVNodes.length) {
+    return {
+      rect: nodeRect
+    };
```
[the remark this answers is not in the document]
straker: I left this as an object because we'll probably need to add more information to this return later, such as the bounding rect of the target and the largests target size rect.

test/commons/math/get-offset.js
```diff
@@ -1,104 +1,66 @@
...
-  it('returns corner to corner distance for diagonal elms', function () {
-    var fixture = fixtureSetup(
-      '<a style="width: 30px; height:30px; margin:10px 30px; display:block">&nbsp;</a>' +
-        '<b style="width: 20px; height:20px; display:block">&nbsp;</b>'
-    );
-    var nodeA = fixture.children[0];
-    var nodeB = fixture.children[1];
-    assert.closeTo(getOffset(nodeA, nodeB), getDiagonal(40), round);
-    assert.closeTo(getOffset(nodeB, nodeA), getDiagonal(30), round);
+  it('returns center to corner or square when at a diagonal', () => {
```
straker, line 27:
```suggestion
  it('returns center to corner of square when at a diagonal', () => {
```

test/commons/math/get-offset.js
straker: Padding apparently. Good catch.

straker:
Create https://github.com/dequelabs/axe-core/issues/4119 and https://github.com/dequelabs/axe-core/issues/4120

WilcoFiers (approved):

Issue #4119, opened 2023-08-07
Update `math.getOffset` algorithm to better align with spec


straker:
As part of #4117 we determined that the `math.getOffset` calculation is not correct and needs to be updated. It didn't produce any false-positives so we decided to update the code in another pr in order to get 4.8 out the door.

The algorithm should be as follows:

For targets with target size >= 24x24px, we should make sure we skip the (potentially expensive) offset calculation since the spacing exception isn't relevant. Target size === largest unobscured rect from `math.splitRects`. E.g. the result of calling `dom.getTargetSize`.

Otherwise, for undersized `target`s:

1. determine bounding rect of `target` and get the center (we'll call `centerTarget`). [Bounding rect](https://www.w3.org/TR/WCAG22/#dfn-bounding-boxes) === the result of calling `math.getBoundingRect` for each rect in `splitRects`. (should be a new function)
1. determine the target's "neighbors" using grid (any other target that might be within 24px of any part of `target`'s bounding rect is a neighbor)
1. for each `neighbor`:
    1. determine `neighborSplitRects` using `math.splitRects(neighbor)`
    1. for each `neighborSplitRect`:
        1. calculate distance from `centerTarget` to the nearest point in `neighborSplitRect` (may be 0 if `centerTarget` lies within `neighborSplitRect`)
    1. let `neighborSpacing` be the minimum `neighborSplitRect` distance
    1. if the `neighbor` is *also* undersized (target size < 24x24px):
        1. calculate `centerNeighbor` similarly to `centerTarget` using the neighbor's bounding box
        1. if the distance from `centerTarget` to `centerNeighbor` is < `neighborSpacing`, use that as the new `neighborSpacing`
1. return the minimum `neighborSpacing` from among all neighbors

padmavemulapati:
Validated with the latest axe-core develop branch code base,
on running ` await axe.run({runOnly:'target-size'});` at the console for latest `axe.js` script

seeing the target-size failure when the element - 

```
<button style=\"width: 10px; height: 10px; margin: 0; padding: 0; position: absolute; top: 0; left: 0\">&nbsp;</button>"

```
and also failure summary is more precised now:

```
failureSummary:
"Fix any of the following:\n  Target has insufficient size (10px by 10px, should be at least 24px by 24px)\n  Target has insufficient space to its closest neighbors. Safe clickable space has a diameter of 0px instead of at least 24px."

```


[...]

Issue #4120, opened 2023-08-07
Use getTargetSize in target-size check


straker:
We didn't do this as part of #4117 but we should update the target-size check to use the new `dom.getTargetSize` function and replace all uses of `nodeRect` with the target size.

straker:
Part of this ticket will need to be commonizing the handling of descendants of targets that can change the true clickable size of the target (such as an image inside a link) and also shouldn't be counted as an obscuring element to the target.

Issue #3673, opened 2022-09-24
target-size should account for the size of its descendants


WilcoFiers:
The new target-size rule looks at the area of the focusable element to determine its size and position relative to other elements. This is making the incorrect assumption that the clickable area of an element cannot exceed its bounding box. This is correct. Any descendant of a link is also clickable, and will activate that parent link. There are numerous ways in which descendants can be positioned outside the boundary of its parent, and so result in a much larger target area than the original link has. One such example is the axe-con logo in the side navigation:

https://www.deque.com/axe-con/sessions/changing-the-way-you-think-and-work-agile-accessibility-audits-in-the-product-life-cycle/
 
![Screenshot axe-con logo who's parent link is much smaller than the actual image it wraps](https://user-images.githubusercontent.com/530687/192097753-84bd6426-3c13-4217-880c-ee13ae4ebf12.png)

Solving this could be pretty complicated. I can think of a few scenarios here:

- If there is no overlap, we could test the widget and the descendants separately, and make sure at least one of them passes
- If the descendant completely encloses the widget, we could test the descendant instead of the widget
- I don't know what to do in case of partial overlap like in axe-con's logo, where the link is wide but small, and the image is less wide, but tall.

dylanb:
How about if they overlap, test them each independently first, if they do NOT pass (i.e. none is, on its own, the required size), then take the [largest common rectangle](https://www.sciencedirect.com/science/article/pii/S0022000017300648) and test that. I assume that this satisfies the intent of the S.C.

padmavemulapati:
Validated with the latest code base,
for the test script, seeing incomplete results when targets have overflowing content.
```
<a href="#" id="target"><img width="24" height="24"></a>
  <a href="#" id="target" style="font-size:24px;"><img width="24" height="24"></a>

  <a href="#" id="target" style="font-size:24px;">
     <img width="24" height="36" style="vertical-align: bottom;">
    </a><br>
    <a href="" style="margin-top:-10px; position:absolute; width:24px;">&nbsp;</a>

    <a href="#" id="target" style="font-size:24px;">
     <img width="24" height="36" style="vertical-align: bottom;">
      </a><br>
      <a href="" style="margin-top:-2px; position:absolute; width:24px;">&nbsp;</a>

      <a href="#" id="target" style="font-size:24px;">
        <img width="24" height="36" style="vertical-align: bottom;">
        </a><br>
[...]

PR #4125, opened 2023-08-09
fix(target-size): correctly calculate bounding box

straker:
Closes: #4119

### DIFF
--- a/lib/commons/dom/get-target-rects.js
+++ b/lib/commons/dom/get-target-rects.js
@@ -1,0 +1,28 @@
+import findNearbyElms from './find-nearby-elms';
+import { splitRects, hasVisualOverlap } from '../math';
+import memoize from '../../core/utils/memoize';
+
+export default memoize(getTargetRects);
+
+/**
+ * Compute the unobscured
+ * @see https://www.w3.org/TR/WCAG22/#dfn-bounding-boxes
+ */
+function getTargetRects(vNode) {
+  const nodeRect = vNode.boundingClientRect;
+  const overlappingVNodes = findNearbyElms(vNode).filter(vNeighbor => {
+    return (
+      vNeighbor.getComputedStylePropertyValue('pointer-events') !== 'none' &&
+      hasVisualOverlap(vNode, vNeighbor)
+    );
+  });
+
+  if (!overlappingVNodes.length) {
+    return [nodeRect];
+  }
+
+  const obscuringRects = overlappingVNodes.map(
+    ({ boundingClientRect: rect }) => rect
+  );
+  return splitRects(nodeRect, obscuringRects);
+}
--- a/lib/commons/math/get-offset.js
+++ b/lib/commons/math/get-offset.js
@@ -1,38 +1,53 @@
-import { getTargetSize } from '../dom';
+import { getTargetRects } from '../dom';
+import { getBoundingRect } from './get-bounding-rect';
+import { isPointInRect } from './is-point-in-rect';
+import { getRectCenter } from './get-rect-center';
+import rectHasMinimumSize from './rect-has-minimum-size';
 
 /**
- * Get the offset between node A and node B
+ * Get the offset between target A and neighbor B. This assumes that the target is undersized and needs to check the spacing exception.
  * @method getOffset
  * @memberof axe.commons.math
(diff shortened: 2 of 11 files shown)


lib/commons/dom/get-target-rects.js
```diff
@@ -0,0 +1,30 @@
...
+export default memoize(getTargetRects);
+
+/**
+ * Return all unobscured rects of a target.
+ * @see https://www.w3.org/TR/WCAG22/#dfn-bounding-boxes
+ * @param {VitualNode} vNode
+ * @return {DOMRect[]}
+ */
+function getTargetRects(vNode) {
+  const nodeRect = vNode.boundingClientRect;
```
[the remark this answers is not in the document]
dbjorge: #3673 is closely related, but the fix that closed it only covered the `target-size` check, I think the `target-offset` check is still vulnerable to the same issue

lib/commons/math/get-offset.js
```diff
@@ -1,38 +1,52 @@
...
+    if (isPointInRect(targetCenter, rect)) {
+      return 0;
+    }
+
+    const closestPoint = getClosestPoint(targetCenter, rect);
+    const distance = pointDistance(targetCenter, closestPoint);
+    minDistance = Math.min(minDistance, distance);
+  }
+
+  if (rectHasMinimumSize(minRadiusNeighbour * 2, targetBoundingBox)) {
```
dbjorge, line 39: This should be checking against the target-size of the neighbor, not the bounding box size of the target

lib/commons/dom/get-target-rects.js
```diff
@@ -0,0 +1,30 @@
...
+  const overlappingVNodes = findNearbyElms(vNode).filter(vNeighbor => {
```
dbjorge, line 15: I know this is just moved and not written in this PR, but I'm noticing as I'm looking at it that it's probably incorrectly considering non-separate-target-descendants as obscuring the target rather than being part of the target. I think this needs to be more complex, along the same lines as `filterByElmsOverlap` in `target-size-evaluate`

dbjorge (changes requested):
Thanks @straker ! Some notes inline

lib/commons/dom/get-target-rects.js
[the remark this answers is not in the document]
straker: Added this as part of #4120

lib/commons/dom/get-target-rects.js
straker: Added this as part of #4120

lib/commons/math/get-offset.js
straker: Good catch

dbjorge: Also maybe good to leave a comment explaining the motivation for the early exit path

```suggestion
  // This early exit accounts for the final "undersized" qualifier in SC 2.5.8's spacing exception,
  // "...the circles do not intersect another target or the circle for another *undersized* target"
  if (rectHasMinimumSize(minRadiusNeighbour * 2, neighborTargetSize)) {
```

dbjorge (dismissed):
LGTM modulo the work deferred for #4120, but would be good to have @WilcoFiers 's eyes on it too

lib/commons/dom/get-target-rects.js
```diff
@@ -0,0 +1,30 @@
...
+    return (
```
WilcoFiers, line 16: You're going to need to filter the non-focusable descendants from this, just like we do in target-size-evaluate. Otherwise something like an icon link ends up with no size. Descendants are part of the target, they don't obscure it.

WilcoFiers (changes requested):

lib/commons/dom/get-target-rects.js
straker: Yep. We're proposing doing that work as part of https://github.com/dequelabs/axe-core/issues/4120 since target size evaluate has the code already and I doing the refactoring.

straker: We agreed that this should be brought into this pr by updating the `getTargetRects` function to filter out descendants that are not in tab order before calling `splitRects`.

WilcoFiers (approved):

PR #4131, opened 2023-08-22
chore(rules/aria-text): fix typo in the description

ma10:
Change \"role=text\" to role=\"text\"


WilcoFiers (approved):
Ahh that's where it came from. Thank you @ma10!

Issue #4126, opened 2023-08-11
link-in-text-block should consider parent nodes & descending images


WilcoFiers:
The following should pass the `link-in-text-block` rule:

```html
<p>
  Lorem ipsum 
  <u><a href="#">dolor</a></u> 
  sit amet consectetur adipisicing elit. Ea officiis
  <a href="#"><img width="16" height="16"> exercitationem</a> 
  ullam inventore facere quaerat, itaque eaque. Eaque 
  <a href="#"><svg width="16" height="16"></svg>consectetur</a>,
  doloremque nostrum, vel animi quis pariatur, qui omnis inventore cum harum!
</p>
```

PR #4135, opened 2023-08-24
fix(link-in-text-block): take into account ancestor inline element styles

straker:
First of 2 for #4126 allowing `link-in-text-block` to use an ancestor inline element's styles when considering the style of the link. While doing this I noticed that the unit tests didn't test any of the styles we consider to be distinct, so went ahead and added them.

### DIFF
--- a/lib/checks/color/link-in-text-block-style-evaluate.js
+++ b/lib/checks/color/link-in-text-block-style-evaluate.js
@@ -15,8 +15,10 @@ export default function linkInTextBlockStyleEvaluate(node) {
     return false;
   }
 
-  var parentBlock = getComposedParent(node);
+  let parentBlock = getComposedParent(node);
+  const inlineNodes = [node];
   while (parentBlock && parentBlock.nodeType === 1 && !isBlock(parentBlock)) {
+    inlineNodes.push(parentBlock);
     parentBlock = getComposedParent(parentBlock);
   }
 
@@ -26,18 +28,22 @@ export default function linkInTextBlockStyleEvaluate(node) {
 
   this.relatedNodes([parentBlock]);
 
-  if (elementIsDistinct(node, parentBlock)) {
-    return true;
+  for (const inlineNode of inlineNodes) {
+    if (elementIsDistinct(inlineNode, parentBlock)) {
+      return true;
+    }
+
+    if (hasPseudoContent(inlineNode)) {
+      this.data({ messageKey: 'pseudoContent' });
+      return undefined;
+    }
   }
-  if (hasPseudoContent(node)) {
-    this.data({ messageKey: 'pseudoContent' });
-    return undefined;
-  }
+
   return false;
 }
 
 function isBlock(elm) {
-  var display = window.getComputedStyle(elm).getPropertyValue('display');
+  const display = window.getComputedStyle(elm).getPropertyValue('display');
   return blockLike.indexOf(display) !== -1 || display.substr(0, 6) === 'table-';
 }
 
--- a/lib/checks/color/link-in-text-block-style-evaluate.js
+++ b/lib/checks/color/link-in-text-block-style-evaluate.js
@@ -28,11 +28,11 @@ export default function linkInTextBlockStyleEvaluate(node) {
 
   this.relatedNodes([parentBlock]);
 
-  for (const inlineNode of inlineNodes) {
-    if (elementIsDistinct(inlineNode, parentBlock)) {
-      return true;
-    }
+  if (elementIsDistinct(node, parentBlock)) {
+    return true;
+  }
 
+  for (const inlineNode of inlineNodes) {
     if (hasPseudoContent(inlineNode)) {
       this.data({ messageKey: 'pseudoContent' });
       return undefined;
--- a/lib/commons/color/element-is-distinct.js
+++ b/lib/commons/color/element-is-distinct.js
@@ -1,4 +1,5 @@
 import Color from './color';
+import { getComposedParent } from '../dom';
 
 /**
  * Creates a string array of fonts for given CSSStyleDeclaration object
@@ -25,7 +26,7 @@ function _getFonts(style) {
  * @return {Boolean}
  */
 function elementIsDistinct(node, ancestorNode) {
-  var nodeStyle = window.getComputedStyle(node);
+  const nodeStyle = window.getComputedStyle(node);
 
   // Check if the link has a background
   if (nodeStyle.getPropertyValue('background-image') !== 'none') {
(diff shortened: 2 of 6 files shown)


test/integration/rules/link-in-text-block/link-in-text-block.html
```diff
@@ -55,7 +55,7 @@ <h1>Non-color change tests</h1>
 <p style="color: black">
   paragraph of text (pass: link is bolder than paragraph)
   <a
-    style="text-decoration: none; font-weight: bolder; color: black"
+    style="text-decoration: none; font-weight: bolder; color: #222"
```
straker, line 58: This would pass even without the change in style due to using the same color as the paragraph.

lib/checks/color/link-in-text-block-style-evaluate.js
```diff
@@ -26,18 +28,22 @@ export default function linkInTextBlockStyleEvaluate(node) {
...
-  }
-  if (hasPseudoContent(node)) {
-    this.data({ messageKey: 'pseudoContent' });
-    return undefined;
+  for (const inlineNode of inlineNodes) {
+    if (elementIsDistinct(inlineNode, parentBlock)) {
+      return true;
+    }
+
+    if (hasPseudoContent(inlineNode)) {
```
dbjorge, line 36: If one `inlineNode` passes `elementIsDistinct` and a different one doesn't but `hasPseudoContent`, should this check return `true` or `undefined`? In this implementation, the answer depends on which order we analyze the `inlineNodes` in, which doesn't seem right to me; I think it would probably make more sense to do one loop for the `elementIsDistinct` checks, pass if any `inlineNode` is distinct, and only then do a second loop for the `hasPseudoContent` checks.

lib/checks/color/link-in-text-block-style-evaluate.js
```diff
@@ -26,18 +28,22 @@ export default function linkInTextBlockStyleEvaluate(node) {
...
   this.relatedNodes([parentBlock]);
 
-  if (elementIsDistinct(node, parentBlock)) {
-    return true;
-  }
-  if (hasPseudoContent(node)) {
-    this.data({ messageKey: 'pseudoContent' });
-    return undefined;
+  for (const inlineNode of inlineNodes) {
+    if (elementIsDistinct(inlineNode, parentBlock)) {
```
dbjorge, line 32: I think checking each inline node individually is a little too simplified; it gives a false negative if an inline node specifies a sufficiently distinct style which is then overridden by an insufficiently distinct style in a descendant.

Here's an example of an integration test which ought to fail, but passes because of this issue:

```html
<p style="color: black">
  paragraph of text (fail: ancestor span is bolder than paragraph, but link reverts ancestor style)
  <span style="font-weight: bolder">
    <a
      style="text-decoration: none; color: #222; font-weight: normal"
      href="#"
      id="fail-ancestor-irrelevant-different-weight"
    >
      Link text</a
    >
  </span>
</p>
```

Probably this needs to do something different for cases where a child doesn't inherit/can't override the distinguishing property (like borders and text decoration) vs cases where a child can override the property (like font weight)

dbjorge (changes requested):

test/commons/color/element-is-distinct.js
```diff
@@ -1,271 +1,355 @@
...
+  const fixture = document.querySelector('#fixture');
+
+  function convertStylePropsToString(props = {}) {
+    return Object.entries(props)
+      .map(([name, value]) => `${name}: ${value}`)
+      .join(';');
+  }
 
-  const fixture = document.getElementById('fixture');
+  function createTestFixture({ target, root, inline } = {}) {
```
straker, line 12: Did some cleanup of this file. It didn't need to create actual style sheets for the elements so just made a simpler function to create inline styles.

test/commons/color/element-is-distinct.js
```diff
@@ -1,271 +1,438 @@
...
```
WilcoFiers, line 12: I think I would prefer for `inline` to be called `targetWrap`. `inline` feels a little vague, especially since I don't think all inline ancestors should be considered (see comments above). A sentence like `returns true if inline adds outline` also messes with my brain a bit, that might just be me though.

WilcoFiers (dismissed):
It seems I forgot to press "submit review" last time. Some of these are from a week ago. Sorry about that.

lib/checks/color/link-in-text-block-style-evaluate.js
```diff
@@ -15,8 +15,10 @@ export default function linkInTextBlockStyleEvaluate(node) {
     return false;
   }
 
-  var parentBlock = getComposedParent(node);
+  let parentBlock = getComposedParent(node);
+  const inlineNodes = [node];
   while (parentBlock && parentBlock.nodeType === 1 && !isBlock(parentBlock)) {
+    inlineNodes.push(parentBlock);
```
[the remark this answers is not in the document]
straker: Not dealing with this in this pr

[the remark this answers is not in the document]
straker: https://github.com/dequelabs/axe-core/issues/4167

lib/commons/color/element-is-distinct.js
```diff
@@ -24,64 +11,105 @@ function _getFonts(style) {
...
   }
 
-  var hasStyle = [
-    'text-decoration-line',
-    'text-decoration-style',
-    'font-weight',
-    'font-style',
-    'font-size'
-  ].reduce((result, cssProp) => {
+  let hasStyle = ['font-weight', 'font-style', 'font-size'].some(cssProp => {
```
WilcoFiers, line 44: Maybe this stuff that inherits should go above the while loop. Also, a comment here would be nice to indicate this isn't in the while loop because it inherits. After not having thought about this for three weeks, it took me a bit to realise why you did it this way.

lib/commons/color/element-is-distinct.js
```diff
@@ -24,64 +11,105 @@ function _getFonts(style) {
...
-
-  if (hasBorder) {
-    return true;
+    // Check if the link has a background
+    if (curNodeStyle.getPropertyValue('background-image') !== 'none') {
+      return true;
+    }
+
+    if (
+      hasSameText(node, curNode) &&
```
WilcoFiers, line 28: I'm a little confused, why is hasSameText() not in the while condition? I.e. 
```js
while (curNode !== ancestorNode && hasSameText(node, curNode)) {
```

My intuition says that that's what should be happening. Remind me why an ancestor with a background-image that has different text would count as making the link distinct?

WilcoFiers (changes requested):

straker:
Closing as I don't think we'll get to this any time soon. pr code will still be here if we need it.

Issue #4137, opened 2023-08-25
Shadow Root with inline style fails to fetch cssom


EduSantosBrito:
### Product

axe-core

### Product Version

4.7.2

### Latest Version

- [X] I have tested the issue with the latest version of the product

#### Expectation
No errors when testing Lit components

#### Actual
When I have inline style on my shadowroot, it triggers an error
![CleanShot 2023-08-25 at 13 42 44](https://github.com/dequelabs/axe-core/assets/19692650/8a578ed9-1ebb-492e-92cb-db5bc95faa7a)
Removing the inline style, it will not trigger an error.
![CleanShot 2023-08-25 at 13 43 37](https://github.com/dequelabs/axe-core/assets/19692650/65df9ae1-eebd-40a5-b93f-82630b13ef74)

#### How to Reproduce
Create a simple Lit component
```ts
import { LitElement, css, html } from "lit";
import { customElement, property } from "lit/decorators.js";

@customElement("primitive-accordion")
export class PrimitiveAccordion extends LitElement {
  static styles = css`
    :host {
      color: blue;
    }
  `;

  @property({ reflect: true })
  name: string;

  constructor() {
    super();
    this.name = "World";
  }

  // Render the UI as a function of component state
  render() {
    return html`<div>
      <h1>Hello, ${this.name}!</h1>
    </div>`;
  }
}

declare global {
  interface HTMLElementTagNameMap {
    "primitive-accordion": PrimitiveAccordion;
  }
}

```

And run this simple test

```ts
import { afterEach, beforeEach, describe, it } from "vitest";
import "../lib/accordion";
import { expect, fixture, html } from "@open-wc/testing";

describe("given a single accordion", async () => {
  let root: HTMLDivElement;

  beforeEach(() => {
    root = document.createElement("div");
    document.body.appendChild(root);
  });

  afterEach(() => {
    root?.remove();
  });

  it("should have no accessibility violations in default state", async () => {
    const el = await fixture(
      html` <primitive-accordion name="Testing"></primitive-accordion> `
    );
    console.log(el.shadowRoot?.innerHTML);
    await expect(el).to.be.accessible();
  });
});
```

#### Additional context
The error is originated on the function `getStylesheetsFromDocumentFragment` inside the file `preload-cssom.js`. The function `convertDataToStylesheet` will return null on `stylesheet.sheet`. Checking the value of `stylesheet.sheet` should fix the issue. I'm creating a PR for that fix right now :)

PR #4138, opened 2023-08-25
fix: ignore stylesheets fetching when is a inline style

EduSantosBrito:
Closes: #4137

This fixes the issue when there's a inline style tag inside the shadowRoot. Since the inline style doesn't have a href, there's no need to fetch it. This is usual on Lit components.


WilcoFiers:
@EduSantosBrito Quick ping, have you had a chance to look at Steve's comment?

EduSantosBrito:
Hey! I have time this week to look at it. I was on holiday last week hahaha

EduSantosBrito:
Ok, after debugging for some hours I found that document.implementation.createHTMLDocument breaks HTMLStyleElement.sheet. There's some issues opened related to that https://github.com/jsdom/jsdom/issues/3179

This makes sense why the browser using `selenium-webdriver` was working and the test using `jsdom` wasn't.

Quick ping @WilcoFiers @straker

straker:
@EduSantosBrito Thanks for that investigation. That helps understand why `style` tags don't have a `.sheet` property, but I'd still like to know how the `style` tag gets through the filter method. `filerStyleAndLinkAttributesInDocumentFragment` [looks for any node with `nodeName` of `style`](https://github.com/dequelabs/axe-core/blob/develop/lib/core/utils/preload-cssom.js#L215-L216) and should remove it from the list before passing to the function that checks for `.sheet`. Somehow that's not happening in your case when it should be.

straker:
That makes sense. Thanks for the looking into that for us.

straker (changes requested):
Just updating the comment to reflect the investigations on when and why this is needed.

straker (approved):

straker:
Approved for security.

PR #4143, opened 2023-08-31
ci: pin chrome version

michael-siek:
This should temporarily fix the circleci issue with chromedriver version not existing since it is an upcoming stable and not yet stable but is stable for chrome

Also I made this into a command to easily revert or fix in the future 

Ref: https://github.com/CircleCI-Public/browser-tools-orb/issues/90

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,7 +33,7 @@ restore_build: &restore_build
 commands:
   browser-tools-job:
     steps:
-      -browser-tools/install-browser-tools:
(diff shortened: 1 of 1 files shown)


straker (approved):

wojsmol:
IMHO we can use [[@puppeteer/browsers](https://pptr.dev/browsers-api)](https://pptr.dev/browsers-api/) described at the end of [this](https://developer.chrome.com/blog/chrome-for-testing/) article

PR #4144, opened 2023-09-01
docs: make quotes usage consistent in messages of rules and checks

ma10:
* Replace name='value' with name=\"value\"
* don't enclose name=value in single quotes


WilcoFiers (approved):
Great stuff! Thank you for the contribution @ma10

Issue #4148, opened 2023-09-06
`target-size` rule contains typos


adamraine:
### Product

axe-core

### Product Version

4.8.0

### Latest Version

- [X] I have tested the issue with the latest version of the product

### Expectation

Failure message ends with:
"Safe clickable space has a diameter of 10px instead of at least 24px"

10px is just an example value.

### Actual

Failure message ends with:
"Safe clickable space has a diameter of {$data.closestOffset}px instead of at least 24px)"

It appears that the actual `data.closestOffset` value is not being filled in. There is also an unpaired `)` at the end of the string that I think should be removed.

### How to Reproduce

Any code that fails the `target-size` rule.

#### Additional context

https://github.com/dequelabs/axe-core/pull/4117

straker:
Ops, you're right. Thanks for spotting that. We'll get it fixed for the next release.

PR #4151, opened 2023-09-07
fix(target-size): show closest offset in message

WilcoFiers:
Fixes a typo in the remediation message. I checked the localisations, none of them have the `{$` instead of `${` problem, although we do need the translators to fix the bracket issue I think. Hoping @ma10 can pick that up for Japanese. 

Closes: #4148, closes #4149

### DIFF
--- a/lib/checks/mobile/target-offset.json
+++ b/lib/checks/mobile/target-offset.json
@@ -7,11 +7,11 @@
   "metadata": {
     "impact": "serious",
     "messages": {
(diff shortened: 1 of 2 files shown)


straker (approved):

ma10:
> Hoping @ma10 can pick that up for Japanese.

Yes, I will prepare a pull request and send it once this one gets finalized and merged.

BTW, while it's quite easy to notice newly added messages, catching updates to existing messages is not so.  I hope we can come up with some mechanism to make it easy to track the translation status, or maybe it's just that I'm not aware of some common techniques for this kind of tasks?

PR #4060, opened 2023-06-20
fix: avoid problems from element IDs that exist on object prototype

WilcoFiers:
Closes: #4000

Issue #4127, opened 2023-08-11
v4.7.1 broke node <= 6


ljharb:
### Product

axe-core

### Product Version

v4.7.1

### Latest Version

- [X] I have tested the issue with the latest version of the product

#### Expectation
Be requireable in node < 6.

#### Actual
Crashes, because `Object.values` was added in node 7.

#### How to Reproduce
Run `require('axe-core')` in node 6.

#### Additional context
I suggest using https://npmjs.com/object.values until you're ready to release a major version that drops support for node 4-6 (please fix v4 for node <= 6 before doing that). I'd be happy to submit a PR.

straker:
Thanks for the issue. Looks like the problem is in [colorjs.io](https://www.npmjs.com/package/colorjs.io) which we brought in to fix https://github.com/dequelabs/axe-core/issues/4007.

ljharb:
I’m confused why it would have ended up bundled into the required entrypoint node if it’s a separate package.

ljharb:
Looks like v4.8.0 also broke node < 16 by using `Object.hasOwn` somewhere.

straker:
@ljharb Thanks for pointing that out. Looks like that was our fault and added in https://github.com/dequelabs/axe-core/pull/4060

ljharb:
@straker thanks for the pointer; filed #4152 to fix that.

straker:
Closed via v4.8.2

ljharb:
@straker i don't think this was fully fixed, since colorjs.io hasn't been updated to have a polyfill yet. https://github.com/dequelabs/axe-core/issues/4127#issuecomment-1710748305 was a separate but related issue.

straker:
Yep sorry, I closed because we fixed the `Object.hasOwn` problem and forgot this was talking about `Object.values`. We should probably bring in an `Object.values` polyfill to fix the color.js issue.

ljharb:
Unfortunately if it's bundled inside color.js, i think it might need to be fixed there, but I haven't had a chance to dig into it yet.

PR #4152, opened 2023-09-07
fix: polyfill Object.hasOwn for node 14

ljharb:
From #4060; fixes https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/953.

### DIFF
--- a/lib/core/imports/index.js
+++ b/lib/core/imports/index.js
@@ -7,6 +7,11 @@ import Color from 'colorjs.io';
 import es6promise from 'es6-promise';
 import { Uint32Array } from 'typedarray';
 import 'weakmap-polyfill';
+import hasOwn from 'core-js-pure/actual/object/has-own';
+
+if (!('hasOwn' in Object)) {
+  Object.hasOwn = hasOwn;
+}
 
 // prevent striping newline characters from strings (e.g. failure
 // summaries). value must be synced with build/configure.js


straker (dismissed):
Thanks for taking on fixing < node 14.

It looks like [core-js provides an implementation for Object.hasOwn](https://github.com/zloirock/core-js/blob/master/packages/core-js/actual/object/has-own.js), which we already depend on. We should use that one rather than bring in another dependency.

Also, the way we handle doing these types of polyfills in axe-core is to add them to the [imports file](https://github.com/dequelabs/axe-core/blob/develop/lib/core/imports/index.js) and only apply the polyfill if the current implementation does not exist. That way we don't have to rewrite every use of the function and can easily remove it when we don't need the polyfill anymore.

```js
import hasOwn from 'core-js-pure/actual/object/has-own';
 
if (!('hasOwn' in Object) {
  Object.hasOwn = hasOwn;
}
```

ljharb:
Sounds good, done. It seems like tests aren't running in every supported version of node, which would have caught this as well as #4127. I'm not familiar enough with circle to fix that myself, but it'd be great if someone was able to do that separately :-)

straker:
Yep, was actually looking into that myself to see what it would entail.

lib/core/imports/index.js
```diff
@@ -7,6 +7,11 @@ import Color from 'colorjs.io';
 import es6promise from 'es6-promise';
 import { Uint32Array } from 'typedarray';
 import 'weakmap-polyfill';
+import hasOwn from 'core-js-pure/actual/object/has-own';
+
+if (!('hasOwn' in Object) {
```
straker, line 12: My bad, seems I forgot the end bracket :D

```suggestion
if (!('hasOwn' in Object)) {
```

ljharb: serves me right for naively copypasting

dbjorge (approved):

straker:
Approved for security

ljharb:
Thanks! I ran `npm run fmt` locally and no diff was generated, so i'm not sure why the prettier task was failing.

ljharb:
Hopefully this will make it into v4.8.2?

straker:
Yep, I forgot I wanted this into the 4.8.1 release before we made it.

ljharb:
Thank you!

PR #4179, opened 2023-10-11
chore: add `dependabot.yml` configuration

straker:
This PR adds dependabot configuration for automatically updating dependencies for NPM and GitHub Actions. While I was updating the actions for sha pinning, I also updated to not use `actions/cache` for npm cache and instead use `actions/setup-node` as is recommended, and updated the actions to run on node 20.


.github/dependabot.yml
```diff
@@ -0,0 +1,42 @@
...
+    schedule:
+      interval: 'monthly'
+    open-pull-requests-limit: 10
+    commit-message:
+      prefix: 'chore'
+    ignore:
+      # Prevent updates to ESM-only versions
+      - dependency-name: 'chalk'
+        versions: ['>=5.0.0']
+      - dependency-name: 'esbuild'
```
dbjorge, line 29: I don't think `esbuild` should be in the same block here - we pin it for a different reason than "being ESM-only". (I think we should investigate whether we still need to do that, or at least have a tech debt ticket for it)

.github/dependabot.yml
```diff
@@ -0,0 +1,42 @@
...
+        versions: ['>=0.11.0']
+      - dependency-name: 'execa'
+        versions: ['>=6.0.0']
+      - dependency-name: 'globby'
```
dbjorge, line 33: Can we replace this instead of needing to pin it? It isn't used in many spots and there are several alternatives that I expect would be viable

.github/dependabot.yml
```diff
@@ -0,0 +1,42 @@
...
+        versions: ['>=12.0.0']
+      - dependency-name: 'html-entities'
```
dbjorge, line 35: It doesn't look like v2 of this is ESM-only

.github/dependabot.yml
```diff
@@ -0,0 +1,42 @@
...
+        versions: ['>=2.0.0']
+      - dependency-name: 'markdown-table'
```
dbjorge, line 37: Can we just rip this out instead of pinning it? the one table it's being used to generate is trivial (all values are known statically to not contain special characters), and any changes it generates will be part of a PR and reviewed anyway. The whole dependency could be replaced by something like:

```js
      # Before:
      const attributesTableMarkdown = mdTable([
        headings.attributesMdTableHeader,
        ...attributesTable
      ]);

      # After:
      const formatMarkdownTableRow = (columnValues) => `| ${columnValues.join(' | ')} |`;
      const attributesTableWithHeader = [
        headings.attributesMdTableHeader,
        ['---', '---'],
        ...attributesTable
      ];
      const attributesTableMarkdown = attributesTableWithHeader.map(formatMarkdownTableRow).join('\n');
```

.github/workflows/format.yml
```diff
@@ -6,26 +6,21 @@ jobs:
...
-        uses: actions/cache@v3
+      - uses: actions/setup-node@v3
         with:
-          path: |
-            ./node_modules
-          key: npm-workspace-cache-${{ runner.os }}-${{ hashFiles('./package-lock.json') }}
-      - uses: actions/setup-node@v1
-        with:
-          node-version: 16
+          node-version: 20
```
dbjorge, line 16: I'd rather update node versions in its own PR. Ideally I'd rather see us test against a matrix of supported versions, but if we're only doing one, I think it's probably less risky to keep it at "the earliest version we support".

dbjorge (changes requested):

straker:
Created https://github.com/dequelabs/axe-core/pull/4180 to update or replace all the pins

.github/workflows/format.yml
```diff
@@ -6,26 +6,21 @@ jobs:
...
           node-version: 16
+          cache: 'npm'
       # Workflows are not allowed to edit workflows. As result, we need to prevent Prettier from formatting them.
       - name: Prevent workflows from being formatted
         run: echo ".github" >> .prettierignore
       - run: npm run fmt
       # Prevent the prettierignore change from being committed.
       - run: git checkout .prettierignore
-      - uses: stefanzweifel/git-auto-commit-action@v4
+      - uses: stefanzweifel/git-auto-commit-action@8756aa072ef5b4a080af5dc8fef36c5d586e521d # tag=v5
```
WilcoFiers, line 24: I assume you did a security review on these. Is it worth sticking a comment in here to document that?

WilcoFiers (approved):
LGTM. Left one suggestions

.github/workflows/format.yml
dbjorge: Checking in a tag + pinned SHA like this is exactly comparable to checking in a version range + lockfile for an npm dependency - why require comments here but not there?

dbjorge (approved):
LGTM, thanks for leaving the comment breadcrumb for #3771

PR #4193, opened 2023-10-12
fix(axe.d.ts): update typings

straker:
Adds more (and more correct) types to axe-core. Also refactored the typescript tests directory so it's not just one large file and moved it the the `tests` dir.


straker:
@dbjorge pointed out an issue with the `utils.publishMetadata` function  that I need to address, but this is mostly in a reviewable state

Issue #3942, opened 2023-03-12
Make it more clear what different impact levels mean


j9t:
### Product

axe-core

### Description

The [rule descriptions](https://github.com/dequelabs/axe-core/blob/develop/doc/rule-descriptions.md) feature rules that have several impact levels (and issue types).

For example, “label” is “Minor, Critical”.

This is probably context-dependent and a rule violation may be either “minor” or “critical,” but stated as is, it’s confusing (because contradictory).

Something like a brief qualifier, like “Minor (in case of…), Critical (in case of…)” may help.

PR #4207, opened 2023-10-20
docs: add links to issue impact

ortonpaul:
Add links to the [Issues Impact](https://github.com/dequelabs/axe-core/blob/develop/doc/issue_impact.md) page at a few spots in the documentation in response to #3942. Let me know if there are other places you can think of that make sense 👍🏼


straker:
Thanks for the pr. The `docs/rule-descriptions.md` file is a generated file and so your changes will not keep. Instead you'll need to update the file that generates it which is https://github.com/dequelabs/axe-core/blob/develop/build/configure.js. Specifically you'll need to [modify the `impact` variable](https://github.com/dequelabs/axe-core/blob/develop/build/configure.js#L369) to add the links to the specific impact levels.

However, that'll add a lot of duplicate links to the page, so we should discuss with @WilcoFiers how we want to handle that.

ortonpaul:
Ahhh I see! Thanks for the heads up. Maybe a link in the [table header](https://github.com/dequelabs/axe-core/blob/0080a7255eb7f246bb7b6f53974a95b65983b83a/build/configure.js#L19C30-L19C36) then or a footnote somewhere?

WilcoFiers (changes requested):
@ortonpaul Thank you for the contribution, but I think this is a little much. One link in the API.md file, and one link in the rule description file is probably enough. As Steve pointed out the rule description file is generated. To update it you'll have to change some code in build/configure.js. 

You may also want to consider putting a link into the readme somewhere. I'm open to suggestions on where.

straker:
@ortonpaul are you still interested in doing this work or could I close the pr?

straker:
Closing due to inactivity

Issue #4217, opened 2023-10-25
Typo in d.ts on rawEnv


WilcoFiers:
Instead of `rawEnv` the allowed reporter type is `raw-env`. This is incorrect. More generally, it is a little bit of a problem that the types only allow pre-existing axe-core reporters, since someone could add their own. Perhaps this just needs to allow all strings? That's quite permissive though. I wouldn't mind some intermediate where you can use a reporter name if you give some indication its available, but if you haven't, only the defaults are available?

PR #4218, opened 2023-10-25
fix(d.ts): RunOptions.reporter can be any string

WilcoFiers:
`raw-env` is not a valid reporter.

Closes: #4217


axe.d.ts
```diff
@@ -132,7 +132,7 @@ declare namespace axe {
   interface RunOptions {
     runOnly?: RunOnly | TagValue[] | string[] | string;
     rules?: RuleObject;
-    reporter?: ReporterVersion;
+    reporter?: ReporterVersion | string;
```
straker, line 135: I don't think this adds any value in typescript. The resolved type of this is just `string` and it looks like it overrides any value of the `ReporterVersion`. Might just make this `string` and not have the set values.

screenshot showing that the resolved type of a set list of strings OR string is just string

screenshot of the resolved type when using a variable of type that is a set list of strings OR string is just string

WilcoFiers: Sure, but can we do better? I looked at using a generic for this, something like:

```ts
interface RunOptions <T extends string | void = void>{
    reporter?: ReporterVersion | T;
}
```

That's fine, but then you can't pass this into axe.run because it expects `RunOptions<void>`. To fix that you'd have to make the reporter something that is passed into axe.run, and to AxeResults (for toolOptions), and AxeReporter, and getFrameContexts... It's a lot. I don't think that's worth the complexity. Open to suggestions though.

dbjorge: Looking at a GitHub-wide code search for `"axe.addReporter"`, there are exactly zero folks who are actually using `axe.addReporter` in practice in an open source project (all hits are copies of axe-core or its tests).

Based on that, I definitely don't think making this more accurate is worth making the entire `RunOptions` type much more complex, and I don't think it's worth making the intellisense worse (like steve's screenshots point out) either; my inclination is to optimize for folks that aren't using `addReporter` and actually just do this:

```ts
  /** If you've added a custom ReporterVersion with `axe.addReporter`, that is also valid here */
  reporter?: ReporterVersion
```

In practice, I think if we're improving this type, the more important improvement would be to add the existing `AxeReporter` function as an allowable value.

dbjorge: (I can also live with `ReporterVersion | string` if Wilco feels strongly about it - I only feel strongly that we shouldn't pollute the whole `RunOptions` type with a generic over this)

WilcoFiers: I do prefer `ReporterVersion | string` here.

straker (approved):

PR #4221, opened 2023-10-25
ci: update to node v18

straker:
In preparation for #4219


michael-siek:
Should this instead be using node LTS (20.9.0)?

WilcoFiers (approved):

PR #4229, opened 2023-11-01
fix(d.ts): RawNodesResult issues

WilcoFiers:
Ensure that RawResult type is correct.

### DIFF
--- a/axe.d.ts
+++ b/axe.d.ts
@@ -1,6 +1,5 @@
 // Type definitions for axe-core
 // Project: https://github.com/dequelabs/axe-core
-// Definitions by: Marcy Sutton <https://github.com/marcysutton>
 
 declare namespace axe {
   type ImpactValue = 'minor' | 'moderate' | 'serious' | 'critical' | null;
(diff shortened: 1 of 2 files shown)


axe.d.ts
```diff
@@ -333,6 +332,14 @@ declare namespace axe {
     xpath: string[];
     ancestry: UnlabelledFrameSelector;
   }
+  interface DqElement extends axe.SerialDqElement {
+    element: Element;
+    toJSON(): axe.SerialDqElement;
```
straker, line 337: Do these need to reference the namespace, or should this just be `SerialDqElement`?

straker (approved):

axe.d.ts
```diff
@@ -1,6 +1,5 @@
 // Type definitions for axe-core
 // Project: https://github.com/dequelabs/axe-core
-// Definitions by: Marcy Sutton <https://github.com/marcysutton>
```
WilcoFiers, line 3: For the record.  Removing this because Marcy should not be the person someone reaches out to for issues with this file as this comment might suggest. No other code file in axe-core includes info on who did the initial commit. It has been updated and expanded on since then many times. I'm removing this line because its out of place.

stephenmathieson: The other past employees should also be removed from `package.json`.

PR #4233, opened 2023-11-02
chore: update ACT tests

WilcoFiers:
Updated to the latest ACT tests (with some notes on things that are no longer consistent). This should prevent nightly from failing.

### DIFF
--- a/test/act-rules/word-spacing-not-important-9e45ec.spec.js
+++ b/test/act-rules/word-spacing-not-important-9e45ec.spec.js
@@ -1,5 +1,11 @@
 require('./act-runner.js')({
   id: '9e45ec',
   title: 'Word spacing in style attributes is not !important',
-  axeRules: ['avoid-inline-spacing']
+  axeRules: ['avoid-inline-spacing'],
+  // See: https://github.com/dequelabs/axe-core/issues/4232
+  skipTests: [
+    '15905a239d6755102be6a60aa152ad963d5b1dbb',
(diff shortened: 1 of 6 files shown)


test/act-rules/word-spacing-not-important-9e45ec.spec.js
```diff
@@ -1,5 +1,11 @@
 require('./act-runner.js')({
   id: '9e45ec',
   title: 'Word spacing in style attributes is not !important',
-  axeRules: ['avoid-inline-spacing']
+  axeRules: ['avoid-inline-spacing'],
+  // See: https://github.com/dequelabs/axe-core/issues/4232
+  skipTests: [
```
WilcoFiers, line 6: Heads up, there's a problem in the testcases.json file. Due to a mistake all but these three test cases are missing for this rule. I'm working to have that fixed. I don't know if the fix of that is going to reveal new issues here. We'll see.

straker: Is there an issue open for the problem? Would hate for it to slip through the cracks.

straker (approved):

test/act-rules/word-spacing-not-important-9e45ec.spec.js
WilcoFiers: There is: https://github.com/w3c/wcag-act-rules/issues/251

Issue #4248, opened 2023-11-17
Locales incomplete


EvilKarter:
### Product

axe-core

### Latest Version

- [X] I have tested the issue with the latest version of the product

#### Expectation
I have set the locale to German as described in the documentation. However, not all rules and checks are translated. 

#### Actual
I expected that all rules and checks are translated

#### How to Reproduce
Set axeSource to german localized build.

#### Additional context
Any thing else we should know about the issue?

straker:
Thanks for the issue. Axe-core's translations are provided by the community as a curtesy and so are only as up-to-date as the last provided translation. I see you opened a pr to update the translations. Thanks for your contribution. I'll go ahead and close this issue and try to review the pr next week.

PR #4249, opened 2023-11-17
chore: Update local de.json

EvilKarter:
I have written a small script, that synchronizes the offered _template.json in locales with all languages. 
If the _template.json file is incomplete, then this PR is also incomplete

I have added to the translation:
- new rules 
- new checks 
- added and removed child entries of rules and checks

I have noticed that all the locales provided are missing entries. 
However, I cannot translate them (with the exception of German)

Closes: #4248

### DIFF
--- a/locales/de.json
+++ b/locales/de.json
@@ -170,7 +170,7 @@
       "help": "Überschriften dürfen nichtleer sein."
     },
     "empty-table-header": {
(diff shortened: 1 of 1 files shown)


straker (changes requested):

EvilKarter:
@straker Many thanks for the feedback. I have adopted your suggested changes. I have also searched for all translations of "have discernible text" and harmonised them

straker (approved):

straker:
Approved for security

PR #4275, opened 2023-12-15
ci: add matrix node test

straker:
Adds a matrix test to test running axe-core in all supported node versions. Currently does not test node 4 support as it turns out axe-core does not work in node 4 as we use `array.prototype.includes` without a polyfill which doesn't exist in node 4. Will make a new pr after this is merged to add that as a `fix`.


test/node/node.js
```diff
@@ -0,0 +1,97 @@
+// this file is purposefully written without mocha and in es5 syntax in order
+// to be compatible with node 4+
+
+var axe = require('../../');
+var assert = require('assert');
+var spawn = require('child_process').spawn;
+
+var domStr =
```
WilcoFiers, line 8: Is it worth important that all-rules.html file and running against that? It won't run the iframe stuff, but that's fine.

.circleci/config.yml
```diff
@@ -360,9 +376,15 @@ workflows:
...
             - build_unix
+      - test_jsdom:
+          requires:
+            - build_unix
       - test_node:
           requires:
             - build_unix
+          matrix:
+            parameters:
+              node-version: ['6', '8', '10', '12', '14', '16', '18', '20']
```
WilcoFiers, line 387: Do we really need all of these?

WilcoFiers (dismissed):
Couple small questions.

.circleci/config.yml
straker: Yes. This is every version of node we claim to support so we should have a test for it. They only need 30s of time each and run in parallel so don't make the build any longer.

test/node/node.js
straker: That might be a good idea, though that file requires that `mocha` exist on the page. If we wanted to use it for this case we'd need to conditionally load most of the JS based on `mocha` existence.

straker: Forgot we're not loading it in a running env so it won't load the js stuff.

.github/workflows/format.yml
```diff
@@ -17,7 +17,7 @@ jobs:
           cache: 'npm'
       # Workflows are not allowed to edit workflows. As result, we need to prevent Prettier from formatting them.
       - name: Prevent workflows from being formatted
-        run: echo ".github" >> .prettierignore
+        run: echo ".github" >> .prettierignore && cat .prettierignore
```
straker, line 20: This line wasn't working correctly and was failing the build as the `.prettierignore` file didn't have a newline character at the end of the file. So instead of appending a new line item this was appending to the end of the last item

```
node_modules/
doc/api.github
```

dbjorge (approved):
LGTM! I verified that locally reverting the fix from #4274 results in the new test correctly failing.

PR #4296, opened 2024-01-09
chore(dependabot): ignore esm chai >= 5.0.0

straker:
[Chai 5.0.0 moved to ESM](https://github.com/chaijs/chai/releases/tag/v5.0.0).

Closes: https://github.com/dequelabs/axe-core/pull/4293

### DIFF
--- a/doc/examples/mocha/package.json
+++ b/doc/examples/mocha/package.json
@@ -13,6 +13,7 @@
   },
   "devDependencies": {
     "axe-core": "^4.6.2",
+    "chai": "^4.4.0",
     "karma": "^6.4.1",
     "karma-chai": "^0.1.0",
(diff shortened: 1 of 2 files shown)


doc/examples/mocha/package.json
```diff
@@ -13,6 +13,7 @@
   },
   "devDependencies": {
     "axe-core": "^4.6.2",
+    "chai": "^4.4.0",
```
straker, line 16: Without the pin `karma-chai` installs the latest version of `chai` which would install v5.

dbjorge: Pinning it to solve that problem is reasonable, but can we keep the version used here and the version used by the root package/lockfile consistent?

dbjorge (changes requested):

PR #4299, opened 2024-01-11
feat(i18n): Add Taiwanese Mandarin translation

PMinn:
* Create zh_TW.json


straker (changes requested):
Thanks for the translation! Below are two typos in in an aria property name. Once these are fixed we can merge this in.

straker (approved):

straker:
Reviewed for security

Issue #4311, opened 2024-01-26
label-content-name-mismatch should not treat hyphenated labels as matching non-hyphenated content


dbjorge:
A [recent ACT update](https://github.com/act-rules/act-rules.github.io/pull/2101) added test case `e9bbdbec137223e2973c6d2896050770c84c26e5` (Failure Example 4) to [proposed ACT rule `Visible label is part of accessible name (2ee8b8)`](https://www.w3.org/WAI/standards-guidelines/act/rules/2ee8b8/proposed/). axe-core currently passes this example where the updated ACT rule expects it to fail:

```html
<a href="#" aria-label="non-standard">nonstandard</a>
```

This happens because axe-core is much more forgiving about punctuation than the ACT rule. The ACT rule's expectation wording is:

> For each target element, all [text nodes](https://dom.spec.whatwg.org/#text) in the [visible text content](https://www.w3.org/WAI/standards-guidelines/act/rules/2ee8b8/proposed/#visible-text-content) [match characters](https://www.w3.org/WAI/standards-guidelines/act/rules/2ee8b8/proposed/#matching-characters) and are contained within the [accessible name](https://www.w3.org/WAI/standards-guidelines/act/rules/2ee8b8/proposed/#accessible-name) of this target element, except for characters in the [text nodes](https://dom.spec.whatwg.org/#text) used to express [non-text content](https://www.w3.org/TR/WCAG21/#dfn-non-text-content). Leading and trailing [whitespace](https://www.w3.org/WAI/standards-guidelines/act/rules/2ee8b8/proposed/#whitespace) and difference in case sensitivity should be ignored.

The critical difference is in the interpretation of the "except for characters in the [text nodes](https://dom.spec.whatwg.org/#text) used to express [non-text content](https://www.w3.org/TR/WCAG21/#dfn-non-text-content)" clause. `label-content-name-mismatch` current removes all punctuation from both label and content before performing its comparison, presuming that punctuation always qualifies as "non-text content", but this isn't accurate in the case where punctuation creates a meaningful difference in language (eg, "government monitoring program" vs "government-monitoring program").

[...]

PR #4312, opened 2024-01-26
test: suppress "visible label in accessible name" ACT test issue

dbjorge:
Temporarily suppresses the ACT mismatch described by #4311 to avoid nightly ACT test noise until we can fix the inconsistency

~Reviewer note: this suppression only has any effect if you're using ACT content that we haven't yet picked up; you'll need to `npm install w3c/wcag-act-rules#main` (like the nightly build does) to meaningfully validate the change locally.~ As of 9a370b6, this can be tested without updating to main

No QA required

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -74,7 +74,7 @@
         "typedarray": "^0.0.7",
         "typescript": "^5.2.2",
         "uglify-js": "^3.17.4",
-        "wcag-act-rules": "github:w3c/wcag-act-rules#485104c",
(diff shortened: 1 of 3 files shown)


WilcoFiers (approved):

WilcoFiers (changes requested):
On second thought, this should probably pull in the latest wcag-act-rules commit. Looks like that's [`dc90495`](https://github.com/w3c/wcag-act-rules/commit/dc90495a5533d326b300ee5a9487afdfc6d493c0)

Issue #4340, opened 2024-02-23
aria-required-children: failures for disallowed child roles shouldn't include an aria-busy message


dbjorge:
### Product

axe-core

### Product Version

4.8.4

### Latest Version

- [X] I have tested the issue with the latest version of the product

### Issue Description

When `aria-required-children` fails due to an element containing a disallowed child (as opposed to *not* containing a required child), the `aria-busy` check still runs separately.

This is problematic for two reasons:
- (common, we've seen confusion about this) It makes the error message confusing. Adding `aria-busy` is almost never the right solution to solving a practical instance of an `aria-required-children` failure, but because it's a separate check, it contributes to the failure summary as a viable top-level option.
- (uncommon) It would create a false negative if an element contained disallowed children but also listed `aria-busy="true"`. [The relevant bit of aria spec](https://www.w3.org/TR/wai-aria-1.2/#mustContain) that talks about `aria-busy` creating a Required Owned Element exemption only allows required elements to be *missing*.

@straker noted this in https://github.com/dequelabs/axe-core/issues/3934#issuecomment-1481609191, but it looks like the comment got lost without being addressed when a PR resolved.

I think the right way to resolve both issues is to just fold the aria-busy logic into the main `aria-required-children-evaluate` function - the logic is pretty minimal and only used in this one rule. That way we can defer the aria-busy check only to the path where required children are missing. I think even in that case, it'd probably be preferable to omit mention of it from failure messages - failure messages don't necessarily need to note every edge case we handle, and I think it causes more confusion than it helps for the message to mention aria-busy even in cases where it *would* create an exemption.

#### Expectation

1. (definite) cases where `aria-required-children-evaluate` would fail with messageKey `'unallowed'` should not consider `aria-busy`. It shouldn't be an option for passing the rule, and it shouldn't appear in failure messages
2. (definite) cases where current `aria-required-children-evaluate` would fail with its default messageKey still consider `aria-busy="true"` as a means of passing the rule
3. (more open for debate) cases where `aria-required-children-evaluate` would fail with its default messageKey should not mention `aria-busy="true"` in the failure message, despite 2

#### Actual

1. Not met
2. Already met (we should make sure it remains met while addressing this issue)
3. Not met

#### How to Reproduce
existing aria-required-children test cases

#### Additional context
n/a

PR #4347, opened 2024-02-28
fix(aria-required-children): avoid confusing aria-busy message in failures

dbjorge:
Removes the confusing aria-busy message from aria-required-children failures (which we've seen several reports of).

Also fixes a false negative (the new #fail13 integration test), which was related but hasn't been reported before.

~~Marks the separate `aria-busy` check as deprecated since no rule uses it after this change.~~ Moving this to separate PR per wilco's feedback

@straker , note that this will have a logical merge conflict with the PRs adding new translations you're working on

Closes: #4340


locales/da.json
```diff
@@ -385,7 +385,9 @@
       }
     },
     "aria-required-children": {
-      "pass": "Krævet ARIA-under-elementer er til stede",
+      "pass": {
```
dbjorge, line 388: I know we normally don't update translations with this sort of change, but I thought it made sense to fix the structure here - with these updates, `npx grunt translate` will preserve the old translations, and without these updates, it'll eat the existing translations and replace them with english.

lib/checks/aria/aria-busy.json
```diff
@@ -1,6 +1,7 @@
 {
   "id": "aria-busy",
   "evaluate": "aria-busy-evaluate",
+  "deprecated": true,
```
WilcoFiers, line 4: Deprecation should probably go into its own PR so it's a separate line in the changelog.

WilcoFiers (changes requested):

WilcoFiers (approved):

PR #4351, opened 2024-03-01
fix(target-size): support translucent elements within pseudo-stacking-context ancestors with implicit z-order

### LINKED ISSUE #4350: target-size: false positive when translucent target is inside a positioned "fake" stacking context
### Product

axe-core

### Product Version

4.8.4

### Latest Version

### Issue Description

We got a customer report today of a target-size false positive (see minimal repro below). This turned out to be because the stacking context tracking used by the target-offset check has an issue where it only has a partial understanding of the rules for elements with `opacity` < 1. Per https://www.w3.org/TR/css-color-3/#transparency:

>Since an element with opacity less than 1 is composited from a single offscreen image, content outside of it cannot be layered in z-order between pieces of content inside of it. For the same reason, implementations must create a new stacking context for any element with opacity less than 1. If an element with opacity less than 1 is not positioned, then it is painted on the same layer, within its parent stacking context, as positioned elements with stack level 0. If an element with opacity less than 1 is positioned, the ‘z-index’ property applies as described in [[CSS21]](https://www.w3.org/TR/css-color-3/#ref-CSS21), except that if the used value is ‘auto’ then the element behaves exactly as if it were ‘0’. See [section 9.9](https://www.w3.org/TR/CSS21/visuren.html#layers) and [Appendix E](https://www.w3.org/TR/CSS21/zindex.html) of [[CSS21]](https://www.w3.org/TR/css-color-3/#ref-CSS21) for more information on stacking contexts. The rules in this paragraph do not apply to SVG elements, since SVG has its own [rendering model](https://www.w3.org/TR/SVG11/render.html) ([[SVG11]](https://www.w3.org/TR/css-color-3/#ref-SVG11), Chapter 3).

Currently, `create-grid.js`'s `isStackingContext` does understand that opacity < 1 should results in a stacking context being created, but `getStackLevel` doesn't understand the rule buried in the middle about non-positioned translucent elements being painted alongside *positioned* elements. The code in `createStackingOrder` that distinguished fake and true stacks also doesn't understand to treat opacity-induced stacking contexts as "true" contexts, which the spec suggests we're supposed to do even when we draw them in POSITION_LEVEL order.

In the repro case below (and the internally-reported customer site), this results in `hasVisualOverlap` (via `visuallySort`) incorrectly determining that the positioned parent completely overlaps and obscures the translucent child, which isn't what happens in browsers in practice. This in turn results in `getTargetRects` returning an empty list such that #4194 gets triggered (which in turn results in the `target-size` false positive).

I titled this bug according to the user-facing impact, but let's have this one just track the bad stack context behavior; #4194 already tracks the bad `getTargetRects`/`targetOffsetEvaluate` behavior.

#### Expectation
Repro case below shouldn't violate the `target-size` rule.

#### Actual
It does.

#### How to Reproduce
```html
<div style="position: absolute; transform: scale(1)">
  <button id="t1">first tab</button>
  <button id="t2" style="opacity: .6">second tab</button>
</div>
```

#### Additional context
- https://www.w3.org/TR/css-color-3/#transparency
- #4194
- #4295

dbjorge:
Draft until researching/testing more on:

- do elements with transform properties need similar treatment (and are there other cases where we should be treating "fake" stacking contexts as true?)
- how to write more complete tests for this (particularly, distinguishing the cases that would be solved by the POSITION_LEVEL part of the change and the treat-as-true-stacking-context part)
- double check browser support (validated current cases on stable chrome, firefox, safari)

Closes: #4350

### DIFF
--- a/lib/commons/dom/create-grid.js
+++ b/lib/commons/dom/create-grid.js
@@ -139,6 +139,7 @@ function isStackingContext(vNode, parentVNode) {
   }
 
   // elements with an opacity value less than 1.
+  // See https://www.w3.org/TR/css-color-3/#transparency
   if (vNode.getComputedStylePropertyValue('opacity') !== '1') {
     return true;
   }
@@ -270,9 +271,7 @@ function createStackingOrder(vNode, parentVNode, treeOrder) {
   // that point (step #5 and step #8)
   // @see https://www.w3.org/Style/css2-updates/css2/zindex.html
   if (isStackingContext(vNode, parentVNode)) {
-    const index = stackingOrder.findIndex(({ stackLevel }) =>
-      [ROOT_LEVEL, FLOAT_LEVEL, POSITION_LEVEL].includes(stackLevel)
-    );
+    const index = stackingOrder.findIndex(isFakeStackingContext);
     if (index !== -1) {
       stackingOrder.splice(index, stackingOrder.length - index);
     }
@@ -302,6 +301,23 @@ function createStackingContext(stackLevel, treeOrder, vNode) {
   };
 }
 
+function isFakeStackingContext(stackingContext) {
+  const { stackLevel, vNode } = stackingContext;
+
+  // elements with opacity < 1 must be treated as their own stacking context,
+  // even if drawn POSITION_LEVEL layer order.
+  // See https://www.w3.org/TR/css-color-3/#transparency
+  if (vNode && vNode.getComputedStylePropertyValue('opacity') !== '1') {
+    return false;
+  }
+
+  if ([ROOT_LEVEL, FLOAT_LEVEL, POSITION_LEVEL].includes(stackLevel)) {
+    return true;
+  }
+
+  return false;
+}
+
 /**
  * Calculate the level of the stacking context.
  * @param {VirtualNode} vNode - The virtual node container of the stacking context
@@ -321,6 +337,15 @@ function getStackLevel(vNode, parentVNode) {
 
   // Put positioned elements above floated elements
   if (vNode.getComputedStylePropertyValue('position') !== 'static') {
+    return POSITION_LEVEL;
+  }
+
+  // From https://www.w3.org/TR/css-color-3/#transparency:
+  //
+  // > If an element with opacity less than 1 is not positioned, then it is
+  // > painted on the same layer, within its parent stacking context, as positioned
+  // > elements with stack level 0.
+  if (vNode.getComputedStylePropertyValue('opacity') !== '1') {
     return POSITION_LEVEL;
   }
 
--- a/test/integration/rules/target-size/target-size.html
+++ b/test/integration/rules/target-size/target-size.html
@@ -28,7 +28,7 @@ this should not effect their outcome -->
   </textarea>
 </p>
 
-<!-- Nested controls-->
+<!-- Nested controls -->
 <p>
   <a
     role="link"
@@ -52,6 +52,14 @@ this should not effect their outcome -->
     Wide link <button id="fail7">y</button>
   </a>
 </p>
+
+<!-- Regression test for stacking context behavior from #4350 -->
+<div style="margin: 30px; height: 50px">
+  <div style="position: absolute; transform: scale(1)">
+    <a href="#" id="pass9">wide enough</a>
+    <a href="#" id="pass10" style="opacity: 0.6">also wide enough</a>
+  </div>
+</div>
 
 <!-- Failed examples -->
 <p>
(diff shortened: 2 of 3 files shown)


straker (commented):
I don't think this is the correct solution. The basic setup to reproduce the bug is that the parent needs to both be positioned without using z-index and use another CSS technique that creates a stacking context (`transform` works as well). Then a child with opacity will cause the stack to think the parent is on top (without this change).

Tested with other styles besides using opacity and things like `transform` or `contain` on the child also show the same problem. So it's not just opacity this should account for but any CSS style that creates a stacking context. 

```html
<div id="one" style="width: 50xp; height: 50px; position: relative; contain: layout;">
  <div id="two" style="width: 50xp; height: 50px; transform: scale(1)"></div>
</div>
<script>
axe.setup();
axe.commons.dom.getElementStack( document.getElementById('two') ) // [#one, #two] 
</script>
```

Further testing of different setups shows that the issue seems to be how the stacking order is determined for positioned auto elements. Take the following example:

```html
<style>
  div {
    width: 100px;
    height: 100px;
  }
</style>
<div id="1" style="background: red;">
  <div id="2" style="contain: paint; background: rgba(0,0,255,0.5);"></div>
</div>
<div id="3" style="margin-top: -100px; background: green;">
  <div id="4" style="background: yellow;"></div>
</div>
```

In the browser this creates a yellow and 50% blue background. Axe's element stack and stacking order is correct. The `#2` element creates a stacking context so replaces the parent stacking context (which is what everything else uses), making it on top element.

```
// axe.commons.dom.getElementStack('#4')
[
    "2",
    "4",
    "3",
    "1",
    "fixture"
]

// stacking order of every node in the element stack
[
    [
        {
            "stackLevel": 0.1,
            "treeOrder": 5,
            "vNode": "#2"
        }
    ],
    [
        {
            "stackLevel": 0,
            "treeOrder": 0
        }
    ],
   // ...
]
```

However if we change it to use positioned parents then everything is incorrect. 

```html
<style>
  div {
    width: 100px;
    height: 100px;
  }
</style>
<div id="1" style="position: absolute; top: 0; background: red;">
  <div id="2" style="contain: paint; background: rgba(0,0,255,0.5);"></div>
</div>
<div id="3" style="position: absolute; top: 0; background: green;">
  <div id="4" style="background: yellow;"></div>
</div>
```

In the browser this shows a solid yellow background. The parent elements both use fake stacking contexts so their stack orders are now 2 deep and positioned using `POSITION_LEVEL`, but the `#2` element is a stacking context so overrides the parent stacking context, thus creating the bug.

```
// axe.commons.dom.getElementStack('#4')
[
    "2",
    "4",
    "3",
    "1",
    "fixture"
]

// stacking order
[
    [
        {
            "stackLevel": 0.1,
            "treeOrder": 5,
            "vNode": "#2"
        }
    ],
    [
        {
            "stackLevel": 0,
            "treeOrder": 0
        },
        {
            "stackLevel": 0.3,
            "treeOrder": 6,
            "vNode": "#3"
        }
    ],
    [
        {
            "stackLevel": 0,
            "treeOrder": 0
        },
        {
            "stackLevel": 0.3,
            "treeOrder": 6,
            "vNode": "#3"
        }
    ],
    [
        {
            "stackLevel": 0,
            "treeOrder": 0
        },
        {
            "stackLevel": 0.3,
            "treeOrder": 4,
            "vNode": "#1"
        }
    ],
    [
        {
            "stackLevel": 0,
            "treeOrder": 0
        }
    ],
    [
        {
            "stackLevel": 0,
            "treeOrder": 0
        }
    ],
    [
        {
            "stackLevel": 0,
            "treeOrder": 0
        }
    ]
]
```

This makes it seem that the `#2` element should be draw in the same stack level as the other nodes so it's drawn in DOM order (making it the bottom of the stack)

dbjorge:
Note for when we next pick this up: Add test cases from https://github.com/dequelabs/axe-core/issues/4629 and make sure the fix also covers those

pwsmaster (approved):

straker:
As the linked ticket notes this was fixed in a recent pr. Tested the scenarios we have listed and they all work as expected now:

```html
<div id="one" style="width: 50xp; height: 50px; position: relative; contain: layout;">
  <div id="two" style="width: 50xp; height: 50px; transform: scale(1)"></div>
</div>
<script>
axe.setup();
axe.commons.dom.getElementStack( document.getElementById('two') ) // [#one, #two] 
</script>
```

Returns `[#two, #one]` for the stack

```html
<style>
  div {
    width: 100px;
    height: 100px;
  }
</style>
<div id="one" style="position: absolute; top: 0; background: red;">
  <div id="two" style="contain: paint; background: rgba(0,0,255,0.5);"></div>
</div>
<div id="thre" style="position: absolute; top: 0; background: green;">
  <div id="four" style="background: yellow;"></div>
</div>
```

Returns `[#four, #three, #two, #one]` for the stack

Issue #2695, opened 2020-12-16
Report case sensitivity issues


WilcoFiers:
Saw a question come up today about:

```html
<ul role="list">
  <li role="listItem">...</li>
</ul>
```

The problem is that capital letter "I" in "listItem". There are two possible issues there:

1. I'm not so sure this should be case sensitive, we should test our ARIA rules to figure out where case sensitivity matters and where it doesn't. If it followed the specs, it should be case insensitive in HTML and case sensitive in XHTML.

2. In such cases, it would be good for axe-core to explicitly call out that the item was failed because of case sensitivity.

3. It would be nice to report this particular case as needs review instead of a fail, because the implicit role here is the correct role. This is a false positive.

WilcoFiers:
May be enough to typecast the result of getRole(). We'll need to test if this works reliably.

jnurthen:
> I'm not so sure this should be case sensitive, we should test our ARIA rules to figure out where case sensitivity matters and where it doesn't. If it followed the specs, it should be case insensitive in HTML and case sensitive in XHTML.

This is defined in https://w3c.github.io/html-aria/#case-sensitivity - it MUST be lower case only

dylanb:
@WilcoFiers close?

WilcoFiers:
This is an ongoing discussion, see https://github.com/w3c/html-aria/issues/280, https://github.com/w3c/html-aria/pull/287 and https://github.com/w3c/aria/issues/1264.

I ran the following through a few AT:

```html
<button role="heading">lower-case</button>
<button role="HEADING">upper-case</button>
```

The latest versions of all the AT/browser combinations axe-core supports treats the role as case insensitive. That includes Safari+VO, JAWS+IE, NVDA+Firefox, NVDA+Chrome. I found it wasn't working in my older version of Firefox, but after upgrading to the latest one, there was no issue.

In conclusion, axe-core should normalise to lower case when testing anything to do with the role attribute. When this issue was opened, I still think this was an issue, but now that Firefox has changed I think we can reasonably say that using upper case in roles doesn't matter.

jnurthen:
@dylanb looking at the extension code it should be ok 
I see the line 		`role = role ? role.toLowerCase() : "";`

dylanb:
@wilco, if Dragon does in fact support all cases, then I think we should put this into the category of `warning`

WilcoFiers:
[...]

PR #4358, opened 2024-03-05
fix(invalidrole): allow upper and mixed case role names

lsprr:
Closes: #2695


lsprr:
> Awesome! Thanks for the pr. We'll want to add a test to the [invalid role test file](https://github.com/dequelabs/axe-core/blob/develop/test/checks/shared/invalidrole.js) that verifies passing an uppercase role does not return `undefined`

Hi! Thank you for the feedback. I've added two tests to the [invalid role test file](https://github.com/dequelabs/axe-core/blob/develop/test/checks/shared/invalidrole.js) to ensure that passing an uppercase role to invalidroleEvaluate does not return undefined.

test/checks/shared/invalidrole.js
```diff
@@ -111,4 +111,19 @@ describe('invalidrole', function () {
     );
     assert.deepEqual(checkContext._data, ['foo', 'bar']);
   });
+
+  it('should return true if applied to an uppercase nonsensical role', function () {
+    var virtualNode = queryFixture(
+      '<div id="target" role="FOO">Contents</div>'
```
straker, line 117: Thanks for the test. Sorry for the delay (been super busy with the 4.9 release happening soon). So an invalid role that is capitalized would still return true if the role was lowercased. We'll want a test that shows that a valid role that is uppercase will now return false. Something like:

```html
<div id="target" role="BUTTON">Contents</div>
```

lsprr: It's alright! Everyone is busy with their work/life, so no rush, :) I did add a test that shows that a valid role that is uppercase will now return false.

test/checks/shared/invalidrole.js
```diff
@@ -111,4 +111,34 @@ describe('invalidrole', function () {
...
+    );
+    assert.isFalse(
+      checks.invalidrole.evaluate.call(
+        checkContext,
+        virtualNode.actualNode,
+        null,
+        virtualNode
+      )
+    );
+    assert.deepEqual(checkContext._data, ['BUTTON']);
```
straker, line 142: Thanks for the additional test. When the result is `false` we don't add a data property so this line needs to be removed (which is why the tests are failing).

```suggestion
```

straker (dismissed):

Issue #4170, opened 2023-09-22
color-contrast does not support `mix-blend-mode: color | luminosity | saturation`


straker:
The following HTML produces an error of `"blendFunctions[blendMode] is not a function"`. Looks like we need to add the [color blend mode](https://www.w3.org/TR/compositing-1/#propdef-mix-blend-mode) to our [list of blend functions](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/color/flatten-colors.js#L11).  We also need to add `luminosity` and `saturation`.

```html
<div style="mix-blend-mode: color;">Errors</div>
```

kundan8239:
Is this fix available on npm ?

straker:
@kundan8239 This hasn't been fixed yet. We hope to get it into the 4.9 release, but we don't have it groomed yet nor a schedule for when it will be released.

Alpdog14:
This is still causing issues on my corporate website, when will this be resolved on Lighthouse tool? or when will it be updated? Like a date?

straker:
@Alpdog14 We are planning the 4.9 release so can make sure to include this.

Alpdog14:
@straker , do u know when Google Chrome will be releasing that 4.9 version of Axe-Core in its brower? because it is still showing a ! for accessibility when I run it

straker:
I do not. Once we release 4.9 it'll still take awhile for Google to update Lighthouse and their products to ingest the changes.

PR #4365, opened 2024-03-07
fix(color-contrast): support color blend modes hue, saturation, color, luminosity

straker:
Decided to add functions to the `Color` class itself to handle adding, dividing, and multiplying a color by a value. Made it cleaner than trying to do that in functions of the flatten color code.

Also refactored the code a bit to put the default export at the top.

Closes: https://github.com/dequelabs/axe-core/issues/4170


lib/commons/color/color.js
```diff
@@ -150,6 +150,54 @@ export default class Color {
...
+  }
+
+  /**
+   * Divide a color by the value and return a new instance of the Color
+   * @method divide
+   * @memberof axe.commons.color.Color
+   * @instance
+   * @return {Color}
+   */
+  divide(value) {
```
WilcoFiers, line 177: I don't think it's a good idea for us to start having colors that are sometimes 8 bit and other times fractions. That's bound to lead to bugs. See below for more thoughts on that.

WilcoFiers (changes requested):
Code works, but I think there are some possible improvements on how to implement this.

PR #4366, opened 2024-03-08
feat(color): add color channel values and luminosity, saturation, clip functions

straker:
Decided to move the Color improvements from https://github.com/dequelabs/axe-core/pull/4365/files into their own PR. This also fixes the `clip` bug mentioned in https://github.com/dequelabs/axe-core/pull/4365/files#r1517706612

No QA needed

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,7 +2,7 @@ module.exports = {
   root: true,
   extends: ['prettier'],
   parserOptions: {
-    ecmaVersion: 2021
+    ecmaVersion: 2023
   },
   env: {
     node: true,
--- a/lib/commons/color/color.js
+++ b/lib/commons/color/color.js
@@ -1,9 +1,9 @@
 import { Colorjs } from '../../core/imports';
 
 const hexRegex = /^#[0-9a-f]{3,8}$/i;
-const hslRegex = /hsl\(\s*([\d.]+)(rad|turn)/;
 
 /**
+ * An sRGB color space Color object
  * @class Color
  * @memberof axe.commons.color
  * @param {number} red
@@ -12,9 +12,24 @@ const hslRegex = /hsl\(\s*([\d.]+)(rad|turn)/;
  * @param {number} alpha
  */
 export default class Color {
+  // color channel values typically in the range of 0-1 (can go below or above)
+  #r;
+  #g;
+  #b;
+  // color component values resolved to the sRGB color space (0-255)
+  #red;
+  #green;
+  #blue;
+
   constructor(red, green, blue, alpha = 1) {
     if (red instanceof Color) {
-      ({ red, green, blue, alpha } = red);
+      // preserve out of gamut values
+      const { r, g, b } = red;
+      this.r = r;
+      this.g = g;
+      this.b = b;
+      this.alpha = red.alpha;
+      return;
     }
 
     /** @type {number} */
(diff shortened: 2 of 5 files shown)


lib/commons/color/color.js
```diff
@@ -13,6 +13,10 @@ const hslRegex = /hsl\(\s*([\d.]+)(rad|turn)/;
  */
 export default class Color {
   constructor(red, green, blue, alpha = 1) {
+    if (red instanceof Color) {
```
straker, line 16: Now we can create a new color from another one:

```js
const color = new Color(0,0,0);
const color2 = new Color(color);
```

WilcoFiers (changes requested):

lib/commons/color/color.js
```diff
@@ -137,19 +141,174 @@ export default class Color {
...
+   * @instance
+   * @return {Color} A new color instance with RGB values between 0-1
+   */
+  normalize() {
+    return new Color(
+      this.red / 255,
+      this.green / 255,
+      this.blue / 255,
+      this.alpha
+    );
```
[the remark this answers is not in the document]
straker: Maybe a new class? An object wouldn't have all the functions attached to handling the fraction components. After thinking about the problem more I don't like the idea of using settings/getters to tie the two together. 

The problem is that values outside the [range of 0-1 are allowed](https://drafts.csswg.org/css-color/#color-function) and while outside the color gamut are not invalid channel values. This means we can't easily translate a value of 1.1 back to RGB as 255 is the maximum. 

What it really seems is that we need two different similar yet different color objects. We have the RGB Color (what it is now) and then need a separate Color to represent 0-1 and all it's functions (functions that also don't really work when applied to an RGB color either).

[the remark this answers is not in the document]
straker: Decided to mimic how Colorjs.io does this by using color coordinates (mapped to `r`, `g`, `b`), also called color channels.

.eslintrc.js
```diff
@@ -2,7 +2,7 @@ module.exports = {
   root: true,
   extends: ['prettier'],
   parserOptions: {
-    ecmaVersion: 2021
+    ecmaVersion: 2023
```
straker, line 5: Allows us to use private properties and functions in classes (supported by babel).

lib/commons/color/color.js
```diff
@@ -57,28 +129,12 @@ export default class Color {
    * @instance
    */
   parseString(colorString) {
-    // Colorjs currently does not support rad or turn angle values
```
straker, line 60: [Colorjs 0.5.0](https://github.com/color-js/color.js/releases/tag/v0.5.0) now fully supports grad, rad, and turn so we don't need to do this anymore.

lib/commons/color/color.js
```diff
@@ -137,32 +193,138 @@ export default class Color {
...
     return 0.2126 * r + 0.7152 * g + 0.0722 * b;
   }
+
+  /**
+   * Add a value to the color channels
+   * @private
+   * @param  {number}  value  The value to add
+   * @return {Color} A new color instance
+   */
+  #add(value) {
```
straker, line 214: Nothing outside of the color class should add a value to the color components.

WilcoFiers (approved):

evelynmcr:
Hi @WilcoFiers and @straker - I've raised a bug [here ](https://github.com/dequelabs/axe-core-nuget/issues/169) which I think could potentially relate to this change? Would you mind taking a look? Thanks!

Issue #4388, opened 2024-03-27
locale specific to Simplified Chinese and Taiwanese Mandarin not translating as expected


padmavemulapati:
### Product

axe-core

### Product Version

4.9.0

### Latest Version

- [X] I have tested the issue with the latest version of the product

#### Expectation
With the generated `axe.zh_CN.min.js` and `axe.zh_TW.min.js` (generated using `npm run build -- --lang=zh_CN` and `npm run build -- --lang=zh_TW`) source snippet , unable to see the descriptions in respective translations

#### Actual
In Simplified Chinese and Taiwanese Mandarin should be traslated

#### How to Reproduce
1. Created a test file for the 4.9.0 Change - `aria-hiddden-body` tag added `wcag131`

```
<!DOCTYPE html PUBLIC "-//W3C//DTD HTML+ARIA 1.0//EN" "http://www.w3.org/WAI/ARIA/schemata/html4-aria-1.dtd">
<html lang="en">
  <head>
    <title>aria-hidden on body Test</title>
 <body aria-hidden="true">
    <h2>Some title.</h2>
    <a href="http://www.deque.com">Deque</a>
    <div id="mocha"></div>
    
  </body>
 </head>
 </html> 
```
2. Run the axe-core with `npm run build` 
3. With the axe-core_4.9.0.min.js source snippet run axe using `await axe.run()`
4. Seeing the expected violations
5. Created the traslation files for `Greek` , `Italian`, `Simplified Chinese`, `Taiwanese Mandarin` using 
```
npm run build -- --lang=el
npm run build -- --lang=it
npm run build -- --lang=zh_CN
npm run build -- --lang=zh_TW
```
respectively.
6. I am able to the proper translations for `Greek` and `Italian`
7. Unable to see the trnasaltions for `Chinese and Taiwanese`

### Recording:

https://github.com/dequelabs/axe-core/assets/47409579/147b23d3-359b-40dc-bca6-518f00dc0df0

straker:
Found the problem. In our rule configure build script we generate the lang/locale from looking at the file name [using a regex](https://github.com/dequelabs/axe-core/blob/develop/build/tasks/configure.js#L21-L24). The regex only looks at alpha characters of length 2 or 3, so a lang like `pt` works but a lang of `pt_BR` is not matched and so defaults to english as the locale.

padmavemulapati:
[...]

PR #4396, opened 2024-04-03
fix: fix building axe-core translation files with region locales

straker:
Figured a simple `split` would do the job without needing to write a complicated regex to do it. Otherwise we'd need something like `/\.([a-zA-Z_]+)\.js$/`

Closes: #4388


dbjorge (approved):
LGTM. Verified that `npm run build -- --lang=zh_CN` produces the expected language build after this change.

Issue #4397, opened 2024-04-04
Should slider role be allowed for button?


msereniti:
### Product

axe-core

### Latest Version

- [X] I have tested the issue with the latest version of the product

#### Expectation
As I see from [html aria specification update](https://github.com/w3c/html-aria/issues/444), the `slider` role should be allowed for the button elements.

#### Actual
As I see in [the axe core standards configuration file](https://github.com/dequelabs/axe-core/blob/105016cfe9d82876cfed2ff5c656a7842c5b3761/lib/standards/html-elms.js#L140), the `slider` role is not in the list of roles, allowed for the button elements.

Is intended or the axe core standards file just needs an update?

msereniti:
If needs an update I can open a pull request but just want to ensure that it's not intended.

straker:
Thanks for the issue. It is not intentional that `slider` is not list, thanks for asking. A pr would be fantastic!

PR #4398, opened 2024-04-04
fix(aria-allowed-role): add gridcell, separator, slider and treeitem to allowed roles of button element

msereniti:
Closes: https://github.com/dequelabs/axe-core/issues/4397

### DIFF
--- a/lib/standards/html-elms.js
+++ b/lib/standards/html-elms.js
@@ -140,14 +140,18 @@ const htmlElms = {
     allowedRoles: [
       'checkbox',
       'combobox',
+      'gridcell',
       'link',
       'menuitem',
       'menuitemcheckbox',
       'menuitemradio',
       'option',
       'radio',
+      'separator',
+      'slider',
       'switch',
-      'tab'
+      'tab',
+      'treeitem'
     ],
     // 5.4 button Element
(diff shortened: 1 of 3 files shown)


straker (dismissed):

straker:
Thanks for the pr. I noticed we're also missing roles of `gridcell`, `separator`, and `treeitem`. It makes sense to update them in one go, so if you're willing we could add them to this pr (with an update to the pr title to reflect that).

If not though I've gone ahead and approved the pr and we can merge as is as well

straker:
Awesome, thanks for updating those. There is a single failing test in https://github.com/dequelabs/axe-core/blob/develop/test/integration/rules/aria-allowed-role/aria-allowed-role.html#L222-L227 where we were failing a button with `role=gridcell` but that's allowed now. The fix would be to update the id to `pass-button-role-gridcell` and then update the companion [json file](https://github.com/dequelabs/axe-core/blob/develop/test/integration/rules/aria-allowed-role/aria-allowed-role.json#L105) to remove `[#fail-button-role-gridcell]` from the violations list and add the new `[#pass-button-role-gridcell]`

straker (approved):
Awesome! 

Reviewed for security

Issue #4386, opened 2024-03-27
label-content-name-mismatch incorrectly fails close button with 


WilcoFiers:
This should not fail the label-content-name-mismatch rule:

```html
<button aria-label="Close">×</button>
```

I think sort of generally this rule should not fail on single character content elements. For this case I think the "×" needs to be added to the list of chars never to fail. I'm not sure why it isn't already. It would be good to go over the list of unicode characters we're always allowing. There are clearly some gaps there.

More generally, even if we encounter single-character texts that are not on the list it would be better to incomplete them instead of failing. We do a similar thing in the color-contrast rule just to be on the safe side.

PR #4402, opened 2024-04-09
fix(label-content-name-mismatch): better dismiss and wysiwyg symbolic text characters

gaiety-deque:
adds exceptions for (dismiss) `×`, wysiwyg characters `b`, `aA`, `abc`

fix: #4386

---

This does _not_ handle all potential use cases. Potential shortcomings worth discussing and opening further issues for are as follows:

- Considering other languages, such as [this example in German including "UT", "AD" or "DGS"](https://github.com/w3c/wcag/issues/3304)
- The original issue #4386 mentioned this being handled as a character limit, but it's unclear if that's a valid option. Consider [the WCAG `ABC` example](https://www.w3.org/WAI/WCAG21/Understanding/images-of-text-no-exception#examples)
- **EDIT**: Does not specifically have test cases for [math expressions and formulae](https://www.w3.org/WAI/WCAG21/Understanding/label-in-name.html#mathematical-expressions-and-formulae)

### DIFF
--- a/lib/checks/label/label-content-name-mismatch-evaluate.js
+++ b/lib/checks/label/label-content-name-mismatch-evaluate.js
@@ -41,12 +41,7 @@ function labelContentNameMismatchEvaluate(node, options, virtualNode) {
   const pixelThreshold = options?.pixelThreshold;
   const occurrenceThreshold =
     options?.occurrenceThreshold ?? options?.occuranceThreshold;
-
   const accText = accessibleText(node).toLowerCase();
-  if (isHumanInterpretable(accText) < 1) {
-    return undefined;
-  }
-
   const visibleText = sanitize(
     subtreeText(virtualNode, {
       subtreeDescendant: true,
@@ -55,13 +50,15 @@ function labelContentNameMismatchEvaluate(node, options, virtualNode) {
       occurrenceThreshold
     })
   ).toLowerCase();
+
   if (!visibleText) {
     return true;
   }
-  if (isHumanInterpretable(visibleText) < 1) {
-    if (isStringContained(visibleText, accText)) {
-      return true;
-    }
+
+  if (
+    isHumanInterpretable(accText) < 1 ||
+    isHumanInterpretable(visibleText) < 1
+  ) {
     return undefined;
   }
 
--- a/lib/commons/text/is-human-interpretable.js
+++ b/lib/commons/text/is-human-interpretable.js
@@ -13,27 +13,42 @@ import sanitize from './sanitize';
 function isHumanInterpretable(str) {
   /**
    * Steps:
-   * 1) Check for single alpha character edge cases
+   * 1) Early escape if string is empty
    *
-   * 2) Check for symbolic text character edge cases
+   * 2) Check for single alpha character edge cases
+   *
+   * 3) Check for symbolic text character edge cases
    * 		a) handle if character is alphanumeric & within the given icon mapping
    * 					eg: 'aA' for toggling capitalization
    *
-   * 3) handle unicode from astral (non bilingual multi plane) unicode, emoji & punctuations
+   * 4) handle unicode from astral (non bilingual multi plane) unicode, emoji & punctuations
    * 					eg: Windings font
    * 					eg: '💪'
    * 					eg: I saw a shooting 💫
    * 					eg: ? (help), > (next arrow), < (back arrow), need help ?
    */
 
-  if (!str.length) {
+  if (
+    isEmpty(str) ||
+    isNonDigitCharacter(str) ||
+    isSymbolicText(str) ||
+    isUnicodeOrPunctuation(str)
+  ) {
     return 0;
   }
 
-  if (str.length === 1 && !str.match(/\d/i)) {
-    return 0;
-  }
+  return 1;
+}
 
+function isEmpty(str) {
+  return sanitize(str).length === 0;
+}
+
+function isNonDigitCharacter(str) {
+  return str.length === 1 && !str.match(/\d/i);
+}
+
+function isSymbolicText(str) {
   const symbolicTextCharactersSet = new Set([
     // toggle capitalization
     'aA',
@@ -42,20 +57,17 @@ function isHumanInterpretable(str) {
     'ABC'
   ]);
 
-  if (symbolicTextCharactersSet.has(str)) {
-    return 0;
-  }
+  return symbolicTextCharactersSet.has(str);
+}
 
+function isUnicodeOrPunctuation(str) {
   const noUnicodeStr = removeUnicode(str, {
     emoji: true,
     nonBmp: true,
     punctuations: true
   });
-  if (!sanitize(noUnicodeStr)) {
-    return 0;
-  }
 
-  return 1;
+  return !sanitize(noUnicodeStr);
 }
 
 export default isHumanInterpretable;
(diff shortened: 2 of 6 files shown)


lib/commons/text/is-human-interpretable.js
```diff
@@ -13,28 +13,37 @@ import sanitize from './sanitize';
...
-    'i' // info
-  ];
+  const symbolicTextCharactersSet = new Set([
+    // dismiss
+    'x',
+    '×',
+    // wysiwyg
+    'i', // or, info
+    'b',
+    'aa', // toggle capitalization (aA)
```
dbjorge, line 42: My recollection from last Thursday is that we decided that we wanted to make *all* single-character strings *except* digits 0-9 be treated as non-interpretable, rather than just extending the list of disallowed single-character cases.

I like your addition of aa/abc though, we should keep that in addition to the single-character thing.

test/integration/rules/label-content-name-mismatch/label-content-name-mismatch.html
```diff
@@ -35,6 +35,11 @@
 <button id="incomplete9" aria-label="help">?</button>
 <button id="incomplete10" aria-label="&#x1F354;">&#x1F354;</button>
 <button id="incomplete11" aria-label="close">&#10060;</button>
+<button id="incomplete12" aria-label="close">×</button>
```
dbjorge, line 38: These integration test updates look good, but could you please also update the unit tests at `/test/commons/text/is-human-interpretable.js`? (the unit tests would be a good place to add a test that motivates  the new `str.trim()` line you added)

dbjorge (changes requested):

lib/commons/text/is-human-interpretable.js
```diff
@@ -13,42 +13,61 @@ import sanitize from './sanitize';
...
+  return sanitize(str).length === 0;
+}
 
-  // Step 2
+function isNonDigitCharacter(str) {
+  return str.length === 1 && !str.match(/\d/i);
+}
+
+function isSymbolicText(str) {
+  const symbolicTextCharactersSet = new Set([
```
straker, line 52: Any particular reason you used Set instead of an array?

test/commons/text/is-human-interpretable.js
```diff
@@ -1,63 +1,71 @@
...
-      var actual = axe.commons.text.isHumanInterpretable(iconText);
+  it('returns 0 when given string is a single alpha character', function () {
+    const singleCharacterExamples = ['i', '×'];
+    singleCharacterExamples.forEach(function (characterExample) {
+      const actual = axe.commons.text.isHumanInterpretable(characterExample);
+      assert.equal(actual, 0);
+    });
+  });
+
+  it('returns 0 when given string is in the symbolic text characters set (blocklist)', function () {
```
straker, line 15: You'll want a companion test to this that verifies that capitalization  matters, so `AA` or `AbC` won't match the blocklist.

test/commons/text/is-human-interpretable.js
```diff
@@ -1,63 +1,71 @@
...
+    const actual = axe.commons.text.isHumanInterpretable('');
     assert.equal(actual, 0);
   });
 
-  it('returns 0 when given string is a single character that is blacklisted as icon', function () {
-    var blacklistedIcons = ['x', 'i'];
-    blacklistedIcons.forEach(function (iconText) {
-      var actual = axe.commons.text.isHumanInterpretable(iconText);
+  it('returns 0 when given string is a single alpha character', function () {
+    const singleCharacterExamples = ['i', '×'];
```
straker, line 8: Should add back the `x` character instead of replacing it with the multiplication symbol.

```suggestion
    const singleCharacterExamples = ['i', 'x', '×'];
```

lib/commons/text/is-human-interpretable.js
```diff
@@ -13,42 +13,61 @@ import sanitize from './sanitize';
...
+  return 1;
+}
+
+function isEmpty(str) {
+  return sanitize(str).length === 0;
+}
 
-  // Step 2
+function isNonDigitCharacter(str) {
+  return str.length === 1 && !str.match(/\d/i);
```
straker, line 48: the `i` isn't needed since digits can't be capitalized. Could also use `str.match(/\D/)` instead of negating it, but that's up to you (both are fine).

```suggestion
  return str.length === 1 && !str.match(/\d/);
```

straker (changes requested):

gaiety-deque:
Should switch to case insensitive for symbolic text

straker:
As discussed, we'll want to add a test that uses a single CJK (any of the 3) character for the label and a different one for the visible text and show that it returns incomplete, that way we document the behavior (previous to this change CJK would pass/fail for single characters if different).

lib/commons/text/is-human-interpretable.js
gaiety-deque: Addressed in c4aa0620

test/commons/text/is-human-interpretable.js
gaiety-deque: We decided it does not matter after all, still enhanced my tests in [30d8388](https://github.com/dequelabs/axe-core/pull/4402/commits/30d8388976118a0cdc85de2bc939a3c6f9c601ef) regardless :)

test/integration/rules/label-content-name-mismatch/label-content-name-mismatch.html
gaiety-deque: Addressed in several recent commits, several new unit tests and some improved integration tests have been added thanks!

gaiety-deque:
> As discussed, we'll want to add a test that uses a single CJK character (any of the 3) for the label and a different one for the visible text and show that it returns incomplete, that way we document the behavior (previous to this change CJK would pass/fail for single characters if different).

@straker addressed in [30d8388](https://github.com/dequelabs/axe-core/pull/4402/commits/30d8388976118a0cdc85de2bc939a3c6f9c601ef)

lib/checks/label/label-content-name-mismatch-evaluate.js
```diff
@@ -55,13 +50,15 @@ function labelContentNameMismatchEvaluate(node, options, virtualNode) {
...
     return true;
   }
-  if (isHumanInterpretable(visibleText) < 1) {
-    if (isStringContained(visibleText, accText)) {
-      return true;
-    }
+
+  if (
+    isHumanInterpretable(accText) < 1 ||
+    isHumanInterpretable(visibleText) < 1
```
dbjorge, line 60: I had to think about this for a while, but I think I agree with your changes here. I think this creates 2 (but really 1) behavior changes:

- If `accText` is not interpretable and `visibleText` is empty
    - Previously `incomplete`, now `pass`
    - This is irrelevant in practice because `label-content-name-mismatch-matches` filters out cases without visible text
- If `accText` is interpretable, `visibleText` is *not* intepretable, *and* `visibleText` is contained in `accText`
    - Previously `pass`, now `incomplete`
    - Example: `<button aria-label="bold">B</button>`

Previously I think that case would pass (return `true`) - with this change, I think it incompletes (returns `undefined`). I think it's more consistent with the change (I don't think it makes sense for a visible "x" button to give different results for labels "close" vs "exit").

dbjorge (approved):
Thanks for the updates, LGTM!

PR #4408, opened 2024-04-12
chore: Update local de.json

EvilKarter:
Update local de.json


straker (approved):
Thanks for the updates.

Approved for security.

Issue #4407, opened 2024-04-12
build on windows fails in 4.9.0


EvilKarter:
### Product

axe-core

### Product Version

4.9.0

### Latest Version

- [X] I have tested the issue with the latest version of the product

#### Expectation
A valid build/tasks/metadata-function-map.js like in version 4.8.4
```javascript
// This file is automatically generated using build/tasks/metadata-function-map.js
import abstractroleEvaluate from '../../checks/aria/abstractrole-evaluate.js';
import ariaAllowedAttrEvaluate from '../../checks/aria/aria-allowed-attr-evaluate.js';
....

const metadataFunctionMap = {
  'abstractrole-evaluate': abstractroleEvaluate,
  'accesskeys-after': accesskeysAfter,
....
``` 

#### Actual
build fails with 
Running "esbuild:core" (esbuild) task > lib/core/base/metadata-function-map.js:2:11: error: Syntax error "c"

```javascript
// This file is automatically generated using build/tasks/metadata-function-map.js
import lib\checks\visibility\hiddenContentEvaluate from '../../../lib\checks\visibility\hidden-content-evaluate.js';
import lib\checks\tables\thHasDataCellsEvaluate from '../../../lib\checks\tables\th-has-data-cells-evaluate.js';
....

const metadataFunctionMap = {
  'lib\checks\aria\abstractrole-evaluate': lib\checks\aria\abstractroleEvaluate,
  'lib\checks\aria\aria-allowed-attr-evaluate': lib\checks\aria\ariaAllowedAttrEvaluate,
....
``` 

#### How to Reproduce
Build on windows with 
```shell
npm run build -- --lang=de
``` 

#### Additional context
i think it is something with an upgraded dependency, because all buits in older versions are still working and I see nothing in the changelog that seems like an cause.

straker:
Thanks for the issue. While I work on debugging this could I ask what version of node you're running?

EvilKarter:
node -v
v20.8.1

PR #4409, opened 2024-04-15
build: pass `posix` option to build in windows

michael-siek:
Closes: https://github.com/dequelabs/axe-core/issues/4407


straker (approved):
In 4.9.0 we switched from `globby` to `glob`. Turns out glob on windows returns the path with double slashes which when passed to `path.basename` directly wasn't working. 

We tested on windows and it was able to successfully build and run tests.

Issue #4194, opened 2023-10-12
target-offset check: targets containing shadow children cause false positives


dbjorge:
### Product

axe-core

### Product Version

4.8.2

### Latest Version

- [X] I have tested the issue with the latest version of the product

#### Expectation
The `target-offset` check uses a [`get-target-rects`](https://github.com/dequelabs/axe-core/blob/449feed607a89ac0da2b06e000a7183016b4aa3f/lib/commons/dom/get-target-rects.js#L34) helper to break down a "target" into a collection of rects. Part of that work involves checking for any other elements that might be obscuring that target. That check is meant to treat children of the target as part of the target, but the way it does so isn't compatible with a target that contains shadow dom children - it assumes that `parentElement.contains(childElementWithinShadow)` will return true, but it doesn't. We already have a shadow-aware version of `contains` (at `/lib/core/utils/contains.js`) which this should use instead. We should audit for other uses of the wrong `contains` as part of fixing this issue. Bonus points for a new eslint `no-restricted-syntax` config that warns us if we accidentally try to use the native one directly again in the future.

This causes `getTargetRects` to treat the child of the target as obscuring the target. In the motivating case, this ends up treating the target as *completely* obscured, which causes `getTargetRects` to return an empty array.

This in turn triggers a separate but related issue in `target-offset` - when a target's neighbor has an empty target rect array, it treats that neighbor as being 0 distance away from the target. It should treat fully-obscured neighbors as omitted from consideration for the check, not as being 0 distance away.

#### Actual
In the repro snippet below, `axe.run({runOnly: 'target-size'})` emits a false positive violation for a case that ought to pass the SC's spacing exception.

#### How to Reproduce

```html
<script>
  const shadowTemplate = document.createElement('template')
  shadowTemplate.innerHTML = '<div id="shadow-container"><slot></slot></div>';
  class ShadowOpenWebComponent extends HTMLElement {
    connectedCallback() {
      const shadowRoot = this.attachShadow({ mode: 'open' });
      shadowRoot.appendChild(shadowTemplate.content.cloneNode(true));
    }
  }

  customElements.define('shadow-open', ShadowOpenWebComponent);
</script>
<style>
  #container {
    font-size: 16px;
  }
  #title {
    background-color: #cff;
  }
  #content {
    background-color: #cfc;
    margin-top: 10px;
  }
</style>
<div id="container">
  <div id="title">
    <!-- getTargetRects will incorrectly return [] for this element... -->
[...]

PR #4410, opened 2024-04-15
fix(target-size): ignore descendant elements in shadow dom

straker:
This also adds to the `.eslintrc` to error if `node.contains()` or `vNode.actualNode.contains()` or used. (also upgraded the `node.attributes` error to account for `vNode.actualNode.attributes` as I noticed it was missing).

Closes: https://github.com/dequelabs/axe-core/issues/4194

### DIFF
--- a/lib/checks/mobile/target-size-evaluate.js
+++ b/lib/checks/mobile/target-size-evaluate.js
@@ -5,6 +5,7 @@ import {
   rectHasMinimumSize,
   hasVisualOverlap
 } from '../../commons/math';
+import { contains } from '../../core/utils';
 
 /**
  * Determine if an element has a minimum size, taking into account
@@ -185,9 +186,7 @@ function toDecimalSize(rect) {
 }
 
 function isDescendantNotInTabOrder(vAncestor, vNode) {
-  return (
-    vAncestor.actualNode.contains(vNode.actualNode) && !isInTabOrder(vNode)
-  );
+  return contains(vAncestor, vNode) && !isInTabOrder(vNode);
 }
 
 function mapActualNodes(vNodes) {
--- a/lib/commons/dom/get-target-rects.js
+++ b/lib/commons/dom/get-target-rects.js
@@ -2,6 +2,7 @@ import findNearbyElms from './find-nearby-elms';
 import isInTabOrder from './is-in-tab-order';
 import { splitRects, hasVisualOverlap } from '../math';
 import memoize from '../../core/utils/memoize';
+import { contains } from '../../core/utils';
 
 export default memoize(getTargetRects);
 
@@ -32,7 +33,5 @@ function getTargetRects(vNode) {
 }
 
 function isDescendantNotInTabOrder(vAncestor, vNode) {
-  return (
-    vAncestor.actualNode.contains(vNode.actualNode) && !isInTabOrder(vNode)
-  );
+  return contains(vAncestor, vNode) && !isInTabOrder(vNode);
 }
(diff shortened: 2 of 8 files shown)


lib/commons/dom/get-target-rects.js
```diff
@@ -2,6 +2,7 @@ import findNearbyElms from './find-nearby-elms';
 import isInTabOrder from './is-in-tab-order';
 import { splitRects, hasVisualOverlap } from '../math';
 import memoize from '../../core/utils/memoize';
```
straker, line 4: This is a separate import on purpose as `memoize` when destructred from the utils import path causes problems for esbuild and it doesn't hoist the memoize function high enough in the build for functions to use it. We've gotten around that by importing it directly.

test/commons/dom/get-target-rects.js
```diff
@@ -74,10 +76,28 @@ describe('get-target-rects', () => {
...
     `);
     const rects = getTargetRects(vNode);
-    console.log(JSON.stringify(rects));
     assert.deepEqual(rects, [
       new DOMRect(10, 5, 30, 7),
       new DOMRect(10, 5, 7, 40)
     ]);
   });
+
+  (shadowSupport ? it : xit)(
```
WilcoFiers, line 85: We stopped doing this now that we're not testing against IE.

```suggestion
  it(
```

test/integration/full/target-size/shadow-dom.html
```diff
@@ -0,0 +1,66 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <title>Target-size too many rects Test</title>
```
WilcoFiers, line 4:
```suggestion
    <title>Target-size shadow DOM test</title>
```

test/integration/full/target-size/shadow-dom.js
```diff
@@ -0,0 +1,45 @@
...
+    axe.testUtils.awaitNestedLoad(() => {
+      const options = {
+        runOnly: ['target-size'],
+        elementRef: true
+      };
+      const context = {
+        // ignore the mocha links
+        exclude: '#mocha'
+      };
+      axe.run(context, options, (err, r) => {
```
WilcoFiers, line 15: Any reason you're not using await here?

lib/checks/mobile/target-size-evaluate.js
```diff
@@ -185,9 +186,7 @@ function toDecimalSize(rect) {
 }
 
 function isDescendantNotInTabOrder(vAncestor, vNode) {
-  return (
-    vAncestor.actualNode.contains(vNode.actualNode) && !isInTabOrder(vNode)
-  );
+  return contains(vAncestor, vNode) && !isInTabOrder(vNode);
```
WilcoFiers, line 189: Should probably have a test for this too.

WilcoFiers (changes requested):

test/integration/full/target-size/shadow-dom.js
straker: just copy pasta from all the other test files

test/checks/mobile/target-size.js
```diff
@@ -411,4 +410,19 @@ describe('target-size tests', function () {
       '#obscurer2'
     ]);
   });
+
+  it('ignores descendants of the target that are in shadow dom', () => {
+    fixture.innerHTML =
```
straker, line 415: `shadowCheckSetup` assumes that the target is either the shadow root node or inside the shadow content. This isn't true for the DOM I was testing and I didn't want to try to update the function in this PR to account for content which uses both.

WilcoFiers (approved):

Issue #4145, opened 2023-09-01
Decorative images should be ignored by the region rule


professorplumb:
### Product

axe-core

### Question

Like many website operators, we use "marketing pixels" to track page hits and conversions. These are usually `img` tags with a height and width of 1 and a `src` URL which points to a marketing partner's site and through which affiliate metrics are tracked.

These might look something like
```
<img src="https://marketing.partner.com/path/to/tracking/script/123456789/?identifier=gobbledygook&amp;random=1594174352357287200&amp;url=https://my-page.url/&amp;other-identifier=asdfjkl" height="1" width="1" alt="" />
```

Since they're often created by a third-party script, the script usually auto-appends them to the `body` tag. Because they're not inside a landmark region, they're called out by axe-core's `region` rule.

My question is: **is there any way to get axe-core to ignore these pixels being outside a landmark region**? The purpose of the rule - to have all consumable content inside landmark regions for screen reader navigation - doesn't apply, since these pixels aren't intended to be viewed by any users at all. (Most have `alt=""` to cause screen readers to ignore them.)

I could see a few ways to update axe-core to accommodate this:
1. Exclude any images with `height="1" width="1"` (or with `alt=""`, or both) from the `region` rule.
2. Allow users to pass "ignorable" rule/element pairings to axe-core: the above could be ignored by something like `ignoreIfMatches: [{ rule: 'region', elementRegex: '/img src.*height="1" width="1".*\/>/' }]`.

I'm aware that we can also fix these violations by appending the pixel to e.g. the end of the `main` element instead of the `body`. We don't always control the script which adds them, though, and we'd love to be able to ignore these in our accessibility audits.

sorin-davidoi:
> Allow users to pass "ignorable" rule/element pairings to axe-core:

Doesn't the `exclude` context property work for your use-case? https://github.com/dequelabs/axe-core/blob/develop/doc/context.md#exclude-elements-from-test

straker:
@professorplumb Unfortunately there is no way to tell axe-core to ignore a specific violation, nor is there a good way for axe-core to detect the purpose of these tracking pixels from the markup alone to know to exclude them from the rule itself. 

But as @sorin-davidoi has suggested, one way you can have axe-core not report the issue is to use the `exclude` property of the context object and that will ignore the element completely from the run.

WilcoFiers:
[...]

PR #4412, opened 2024-04-16
fix(region): Decorative images ignored by region rule

gaiety-deque:
Issue was around [decorative images](https://www.w3.org/WAI/tutorials/images/decorative/), specifically 1 pixel wide/tall marketing tracking images, that get added outside of regions failing the region rule. An easy and fairly robust solution [the issue opener agreed with](https://github.com/dequelabs/axe-core/issues/4145#issuecomment-1710302690) was ignoring images with `alt=''` and so that's what this PR implements.

Dev notes:

- Added `isPresentationGraphic/1` check which currently only handles `alt=''` and ignores them for the region rule
- role='presentation' test added as well, but this already worked previously prior to this code change
- svg I believe is handled differently already, there's a test called `treats svg elements as regions` so the new function doesn't check for svg's

fix: #4145


lib/checks/navigation/region-evaluate.js
```diff
@@ -45,14 +45,19 @@ function getRegionlessNodes(options) {
...
+  // - hidden content
+  // - a presentation graphic
   if (
     getRole(virtualNode) === 'button' ||
     isRegion(virtualNode, options) ||
     ['iframe', 'frame'].includes(virtualNode.props.nodeName) ||
     (dom.isSkipLink(virtualNode.actualNode) &&
       dom.getElementByReference(virtualNode.actualNode, 'href')) ||
-    !dom.isVisibleToScreenReaders(node)
+    !dom.isVisibleToScreenReaders(node) ||
```
straker, line 59: This brings up an interesting question. Should a presentational img using `alt=""` be considered not visible to screen readers? @WilcoFiers is the better change to update `isVisibleToScreenReaders` to return false for an img with `alt=""`? Right now `isVisibleToScreenReaders` returns true for an img with empty alt.

gaiety-deque: You know I was wondering that, it was a path I started diving down and then uncertainty got the best of me. It's a function leveraged in a lot of places.

If we think that's the right path I'm happy to try it out and see if any other test failures come up along the way that make us second guess it.

Hellothuy1 (changes requested):
T

lib/checks/navigation/region-evaluate.js
WilcoFiers: Fair question, but no I don't think so. Conceptually, there's a difference between a hidden element and an element that's ignored. It matters for example on ``. That should fail for having an inappropriate role, whereas `` should be incomplete.

lib/checks/navigation/region-evaluate.js
```diff
@@ -86,6 +91,10 @@ function findRegionlessElms(virtualNode, options) {
   }
 }
 
+function isPresentationGraphic(virtualNode) {
+  return virtualNode.props.nodeName === 'img' && virtualNode.attr('alt') === '';
```
WilcoFiers, line 95: This feels a little incomplete. There are other graphics, and there are more ways for an image to be decorative. We have methods we can use to figure this stuff out:

```suggestion
  return aria.isVisualContent(virtualNode.actualNode) && text.accessibleTextVirtual(virtualNode) === '';
```

WilcoFiers (changes requested):

Hellothuy1 (approved):

Hellothuy1:
H

lib/checks/navigation/region-evaluate.js
gaiety-deque: Hmm I agree, alas I'm a bit unsure which existing methods are the most appropriate. Your suggestion is close, however `accessibleTextVirtual` never actually checks for an alt attribute in its computational steps.

The closest I've found is `hasAltEvaluate` in `shared` but that isn't actually used anywhere and only returns a boolean.

However, I think `findTextMethods` should work as it uses `nativeTextMethods` which checks for alt text - but that gives me an empty string even without an alt attribute on the image.

Think I'll post asking for pairing assistance tomorrow to learn how to fish for the right answer on this one :)

gaiety-deque: @WilcoFiers I paired with Steve and we generally agree that it'd be nice to make this check more broad, however `isVisualContent` is perhaps too broad for what it checks for. We currently seem to lack a shared utility for checking if something is a graphic (as far as we could tell) so one path forward may be writing one.

We would both love to circle back with you on Monday @WilcoFiers about what we'd like to do

gaiety-deque: In a call we discussed we wanted to check for the tags `img`, `svg`, `object` and `canvas` and to then use `accessibleTextVirtual` but that function does _not_ work for our needs, as it seems to return an empty string regardless of if an img tag has no alt attribute (like the original issue) or an empty alt attribute.

My findings copied from Slack below:

---

It (`accessibleTextVirtual`) seems to be written _deeply_ to rely on empty strings all over the place as far as I can tell.

* `accessibleTextVirtual`'s main `reduce` initializes the value as `''`
* ^ it also tries to sanitize the value, which if the value is falsey at all returns an empty string
* ^ the reduce returns immediately if the value is _anything_ that isn't an empty string
* `nativeTextAlternative`'s main `reduce` also initializes the value as `''`
* ^'s uses several functions in its reduce including `findTextMethods` finds and uses `native-text-methods.js`'s `attrText` function which returns the attribute **or an empty string** meaning it'll **never** return `null`

I tried adjusting all of these places but it got pretty hairy and makes me think this was all set up intentionally the way it was, and it may be the wrong thing to use.

---

So this PR is stuck assuming on this comment thread currently without a clear path forward. I believe Steve and Wilco are busy so, hi @dbjorge any thoughts?

gaiety-deque: I think the real question is the following:

> What we need to determine is if the region rule should ignore both an image with an empty alt='' attribute and if an image is entirely lacking an alt attribute

dbjorge: ##### My thoughts on the problem

> What we need to determine is if the region rule should ignore both an image with an empty alt='' attribute and if an image is entirely lacking an alt attribute

I think this ought to come down to "is the image in question going to be exposed to ATs or not"; I think the result of that is that `region` should ignore most cases with empty alt and should not ignore cases with no alt, but note that there are some edge cases where the `alt` attribute isn't the whole story.

I agree with the other comment chain that it's confusing that the use of `isVisibleToScreenReaders` doesn't already cover "is the image going to be exposed to ATs or not". Even if we do decide to keep `isVisibleToScreenReaders` as-is, I think it'd be a good idea for us to at least update the comment on `isVisibleToScreenReaders` to explain this gotcha.

The difficulty with updating `isVisibleToScreenReaders` for this case is that many places that use it (including-but-not-limited-to its own recursive implementation) do so with the understanding that if it returns false for a given element, it implies that the element **and all its descendants** are going to be invisible to screen readers.

*However, that's not the case here!*

The mechanism by which `img`s with `alt=""` get hidden from ATs is that [the implicit aria role for such an element is `role="none"`/`role="presentation"` rather than `role="img"`](https://www.w3.org/TR/html-aria/#el-img-no-name)<sup>1,2</sup>.

`role="none"`/`role="presentation"` semantics are different from `isVisibleToScreenReaders`' semantics in that they *don't* hide the element's descendants from the accessibility tree; they *only* hide the semantics of the element itself. A basic `` tag with a resolved role of `none` will be omitted from the accessibility tree because its semantics as an image covers the only information about it that would appear in the tree. But in something like `<h1 role="none">content</h1>`, for example, the `content` text still appears in the accessibility tree (just as a plain text node, not treated as a heading).

##### What I suggest to resolve it

I lean towards updating `findRegionlessElms` to calculate an `isShallowlyHidden` condition for candidate elements, something like this:

```js
  const role = getRole(virtualNode, { noPresentational: true });
  const isPresentational = role === null
  // The element itself is not visible to screen readers, but its descendants might be
  const isShallowlyHidden = isPresentational && !hasChildTextNodes(elm)
```

...and then update the conditional that checks for `dom.hasContent(node, true)` (L76 of the original) to treat `!isShallowlyHidden` as an additional requirement to `return [virtualNode];`.

You could make a reasonable argument that the better fix would be to add `presentational-role` as an additional `any` check on the `region` rule. I think that (+ a change to that check's messages to special case the messaging for the `alt=""` case) would likely be a better result message in this specific motivating repro case, but I recommend against it because of the way the `region` rule currently tries to only emit failures for the topmost ancestor in a tree of regionless elements; it'll give a misleading "how to fix" recommendation for `region` violations on elements with text descendants, where a presentational role wouldn't be enough to hide that element's descendant tree as a whole.

---

##### Footnotes

1: This has a few caveats:

- If the author assigns a `role` attribute explicitly, it will take precedence over implicit role selection
- If the element is focusable or has a global aria attribute, its implicit role will be treated as `role="img"` even with `alt=""`
    - this also overrules an *explicit* `role="none"`/`role="presentation"` attribute

2: axe-core has a few places where this logic comes up, including the caveats from note 1; see [`presentational-role-evalutate.js`](https://github.com/dequelabs/axe-core/blob/develop/lib/checks/shared/presentational-role-evaluate.js) and [`implicit-html-roles.js`'s `img` entry](https://github.com/dequelabs/axe-core/blob/eac822328b21325f0c16438b69f4a39a3102d5b9/lib/commons/standards/implicit-html-roles.js#L92-L106)

gaiety-deque: Possibly addressed in 524f0be7

I played with adding some manual checks for the caveats of global aria attributes and focusable, but it didn't seem to need any additional code. Writing the tests leads me to believe the caveats are handled as expected.

romankulkovsf (approved):

lib/checks/navigation/region-evaluate.js
```diff
@@ -86,6 +91,14 @@ function findRegionlessElms(virtualNode, options) {
   }
 }
 
+function isShallowlyHidden(virtualNode) {
+  // The element itself is not visible to screen readers, but its descendants might be
+  return (
+    getRole(virtualNode, { noPresentational: true }) === null &&
+    !hasChildTextNodes(virtualNode)
```
WilcoFiers, line 98: I don't think we can just ignore anything presentational. Most elements don't have this dynamic thing going on where they are only presentational if there is nothing interesting about them (alt, tabindex, global aria). We'd have to test for all of them if you want this to work for anything except the `img` element.

I think we're better off just skipping decorative images explicitly here, and leaving the rest for another day:

```suggestion
function isDecorativeImage(virtualNode) {
  return (
    virtualNode.props.nodeName === 'img' &&
    getRole(virtualNode, { noPresentational: true }) === null
```

WilcoFiers (changes requested):
Can you add an integration test as well?

lib/checks/navigation/region-evaluate.js
```diff
@@ -86,6 +91,14 @@ function findRegionlessElms(virtualNode, options) {
   }
 }
 
+function isShallowlyHidden(virtualNode) {
+  // The element itself is not visible to screen readers, but its descendants might be
+  return (
+    getRole(virtualNode, { noPresentational: true }) === null &&
```
WilcoFiers, line 97: Talked to @dbjorge about this. We think this is the way to do it:

```suggestion
    ['none', 'presentation'].includes(getRole(virtualNode)) &&
```

This avoids us skipping elements like `<canvas aria-label="foo">`. That's not something I think we should skip. Same with focusable stuff. This change limits it to elements that explicitly have role="none", or have role=none implicitly (which only happens for img).

lib/checks/navigation/region-evaluate.js
WilcoFiers, line 1: Can you add an integration test for this.

test/checks/navigation/region.js
WilcoFiers, line 1: Can you add a few more tests here about other elements:

- `<canvas role="none">` passes
- `<object aria-label="foo">` fails
- `<svg role="none" aria-label="foo">` fails
- `<div role="none">foo</div>` fails
- - `<div role="none"></div>` passes

gaiety-deque: We intentionally allow svg's outside of regions https://github.com/dequelabs/axe-core/commit/fab58d4bf#diff-b11ab1ec32c4f5c0e17de0c9a1f3c5a9835bcc133f673b5da3ecaaf66a932077

lib/checks/navigation/region-evaluate.js
gaiety-deque: Added :)

test/checks/navigation/region.js
gaiety-deque: Added the other tests as both unit and integration tests

test/integration/full/region/region-pass.html
```diff
@@ -20,6 +20,16 @@
   </head>
   <body>
     <a href="#mainheader" id="skiplink">This is a skip link.</a>
+    <!-- Decorative image -->
+    <img id="target" src="#" alt="" />
+    <!-- Decorative image -->
+    <img id="target" src="#" role="presentation" />
```
WilcoFiers, line 26:
```suggestion
    <img src="#" alt="" />
    <!-- Decorative image -->
    <img src="#" role="presentation" />
```

WilcoFiers (changes requested):
Almost

WilcoFiers (approved):

Issue #4247, opened 2023-11-17
False positive with list containing forbidden tags


claudiu-cristea:
### Product

axe-core

### Product Version

4.8.2

### Latest Version

- [X] I have tested the issue with the latest version of the product

#### Expectation

With this HTML chunk:

```html
<ul class="pagination pager__items js-pager__items  justify-content-center">
  <li class="page-item pager__item is-active active">
    <a href="?page=0" title="Current page" aria-current="page"
       class="page-link">
            <span class="visually-hidden">
              Current page
            </span>1</a>
  </li>
  <li class="page-item pager__item">
    <a href="?page=1" title="Go to page 2" class="page-link">
            <span class="visually-hidden">
              Page
            </span>2</a>
  </li>
  <li class="page-item pager__item">
    <a href="?page=2" title="Go to page 3" class="page-link">
            <span class="visually-hidden">
              Page
            </span>3</a>
  </li>
  <li class="page-item pager__item">
    <a href="?page=3" title="Go to page 4" class="page-link">
            <span class="visually-hidden">
              Page
            </span>4</a>
  </li>
  <li class="page-item pager__item">
    <a href="?page=4" title="Go to page 5" class="page-link">
            <span class="visually-hidden">
              Page
            </span>5</a>
  </li>
  <li class="page-item pager__item">
    <a href="?page=5" title="Go to page 6" class="page-link">
            <span class="visually-hidden">
              Page
            </span>6</a>
  </li>
  <li class="page-item pager__item">
    <a href="?page=6" title="Go to page 7" class="page-link">
            <span class="visually-hidden">
              Page
            </span>7</a>
  </li>
  <li class="page-item pager__item">
    <a href="?page=7" title="Go to page 8" class="page-link">
            <span class="visually-hidden">
              Page
            </span>8</a>
  </li>
  <li class="page-item pager__item">
    <a href="?page=8" title="Go to page 9" class="page-link">
            <span class="visually-hidden">
              Page
            </span>9</a>
  </li>


  <li class="page-item pager__item pager__item--ellipsis disabled"
      role="presentation"><span class="page-link">…</span></li>
  <li class="page-item pager__item pager__item--next">
    <a href="?page=1" title="Go to next page" rel="next" class="page-link">
      <span class="visually-hidden">Next page</span>
      <span aria-hidden="true">
              <svg class="bi icon--s">
                <use
[...]

Issue #3935, opened 2023-03-07
Allowance of presentational li elements inside a list


Al-Dani:
### Product

axe-core

### Feature Description

Is it possible to include `role=presentation` to the list of allowed roles for the `<li>` inside `<ul>`?
Having a presentational element won't interrupt the user while navigating through a list, as it will be removed from the a11y tree.

A use-case for this is to have a navigation list with separators between items. Separators are decorative.

Al-Dani:
That's what I want to achieve:
```
<ul> 
 <li> <a/> </li>
 <li> <a/> </li>
 <li role='presentation'> <hr/> </li>
 <li> <a/> </li>
 <li> <a/> </li>
</ul>
```

WilcoFiers:
Putting role=presentation on an `li` means the li is hidden, but its content is not. You should use aria-hidden instead if you want it ignored.

PR #4414, opened 2024-04-18
Allow presentation elements in list

larowlan:
Drupal is using markup like this for pagination/pager - the ... element is presentational only.

```html
<ul>
  <li><a href="page-1" class="is-active">Go to page 1</a></li>
  <li><a href="page-2">Go to page 2</a></li>
  <li><a href="page-3">Go to page 3</a></li>
  <li><a href="page-4">Go to page 4</a></li>
  <li role="presentation">...</li>
  <li><a href="page-2">Go to next page</a></li>
</ul>

### DIFF
--- a/lib/checks/lists/only-listitems.json
+++ b/lib/checks/lists/only-listitems.json
@@ -2,7 +2,7 @@
   "id": "only-listitems",
   "evaluate": "invalid-children-evaluate",
   "options": {
-    "validRoles": ["listitem"],
+    "validRoles": ["listitem", "presentation"],
     "validNodeNames": ["li"]
   },
   "metadata": {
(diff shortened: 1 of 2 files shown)


straker:
Thanks for the PR. Unfortunately a list containing a `role=presentation` child causes problems for screen readers. The content of the `role=presentation` item is [still visible and can be navigated to](https://github.com/dequelabs/axe-core/issues/4247#issuecomment-1816737181). If the content of the list item should be hidden [you should use `aria-hidden`](https://github.com/dequelabs/axe-core/issues/3935#issuecomment-1468522260).

Because of that we cannot accept the PR, so going to close.

larowlan:
Thanks :+1:

Issue #4371, opened 2024-03-13
Aria-roles standards object contains many roles that are classified as widgets but are not


straker:
Role such as `tabpanel`, `timer`, `alert`, etc. are marked as widgets but they shouldn't be. This problem is contributing to https://github.com/dequelabs/axe-core/issues/4360 as each example uses a tabpanel with many overlapping child nodes.

We'll need to investigate the ramifications of changing these to their correct types before we make the change.

WilcoFiers:
An example where this is creating a false positive is on https://www.hep.ca/horse-insurance/. The close button the cookie banner is 15x15. Axe is failing this for being too close to the to the cookie banner, which has tabindex=0 and role=alert.

gaiety-deque:
Just as a note, the description I believe is _slightly_ inaccurate as `tabpanel` **does belong** to the widget abstract role https://www.w3.org/TR/wai-aria/#widget_roles

I'm playing with getting everything synced to the spec now to see where that gets us

PR #4421, opened 2024-04-22
fix(aria-roles): correct abstract roles (types) for aria-roles

gaiety-deque:
Too many aria roles were set to widget or otherwise were incorrect, added a comment source for where I got my information for which abstract roles were the type of other aria roles

Fix: #4371


gaiety-deque:
@WilcoFiers unsure how to test this PR's changes against the comment you left in the referenced issue, I don't see a cookie notice on the horse insurance website. Any guidance is appreciated.

@straker I feel reasonably good about this change but I'm a bit unsure if it fully addresses the original issue as it seemed fairly open ended

lib/standards/aria-roles.js
```diff
@@ -17,21 +18,21 @@
...
   alert: {
-    type: 'widget',
+    type: 'structure',
     // Spec difference: Aria-expanded removed in 1.2
     allowedAttrs: ['aria-expanded'],
     superclassRole: ['section']
   },
   alertdialog: {
-    type: 'widget',
+    type: 'window',
```
straker, line 27: `window` isn't a type we support. [Valid types](https://github.com/dequelabs/axe-core/blob/develop/doc/standards-object.md#structure-1) are:

* abstract
* widget
* structure
* landmark

So probably this should be `structure` since it's not a landmark. Also according to https://www.w3.org/WAI/ARIA/1.2/class-diagram/rdf_model.svg `alert` is a section, which is under structure. 

```suggestion
    type: 'structure',
```

lib/standards/aria-roles.js
```diff
@@ -119,7 +120,7 @@ const ariaRoles = {
     nameFromContent: true
   },
   combobox: {
-    type: 'widget',
+    type: 'composite',
```
straker, line 123: I think this should remain a widget. According to https://www.w3.org/WAI/ARIA/1.2/class-diagram/rdf_model.svg `combobox` is a composite, but so is a `spinbutton` and `spinbutton` is definitely a widget. So I think we're safe to say any interactable composite should be a widget.

```suggestion
    type: 'widget',
```

lib/standards/aria-roles.js
```diff
@@ -17,21 +18,21 @@
...
+    type: 'window',
     // Spec difference: Aria-expanded removed in 1.2
     allowedAttrs: ['aria-expanded', 'aria-modal'],
     superclassRole: ['alert', 'dialog'],
     accessibleNameRequired: true
   },
   application: {
     // Note: spec difference
-    type: 'landmark',
+    type: 'structure',
```
straker, line 35: This was an intentional difference

lib/standards/aria-roles.js
```diff
@@ -169,7 +170,7 @@ const ariaRoles = {
     prohibitedAttrs: ['aria-label', 'aria-labelledby']
   },
   dialog: {
-    type: 'widget',
+    type: 'window',
```
straker, line 173: Probably `structure` again as not a landmark

```suggestion
    type: 'structure',
```

lib/standards/aria-roles.js
```diff
@@ -301,7 +302,7 @@ const ariaRoles = {
     superclassRole: ['section']
   },
   listbox: {
-    type: 'widget',
+    type: 'composite',
```
straker, line 305:
```suggestion
    type: 'widget',
```

lib/standards/aria-roles.js
```diff
@@ -1,4 +1,5 @@
 // Source: https://www.w3.org/TR/wai-aria-1.1/#roles
+// Source for abstract roles: https://www.w3.org/TR/wai-aria/#abstract_roles and https://www.w3.org/WAI/ARIA/1.2/class-diagram/rdf_model.svg
```
straker, line 2: According to https://www.w3.org/WAI/ARIA/1.2/class-diagram/rdf_model.svg `tabpanel` is a section, which is under structure.

straker (changes requested):
Thanks for taking this on. 

https://www.w3.org/WAI/ARIA/1.2/class-diagram/rdf_model.svg is an interesting resource. I've not seen it before. It's also hard to follow sometimes what line belongs to what, so I hope I parsed it correctly.

lib/standards/aria-roles.js
gaiety-deque: Perhaps a good use of our 1:1 today is for us to talk about how you arrived at that conclusion, when there's two arrows coming off an arrow in the rdf model is makes it quite unclear.

lib/standards/aria-roles.js
gaiety-deque: Despite it being an abstract role under https://www.w3.org/TR/wai-aria/#abstract_roles referencing `alertdialog` and `dialog` under https://www.w3.org/TR/wai-aria/#window_roles ?

lib/rules/widget-not-inline-matches.js
```diff
@@ -18,7 +18,8 @@ const matchesFns = [
 ];
 
 function isWidgetType(vNode) {
-  return getRoleType(vNode) === 'widget';
+  const roleType = getRoleType(vNode);
+  return roleType === 'widget' || roleType === 'composite';
```
straker, line 22: Undo this change since we'll leave `listbox` and `combobox` as widgets

lib/standards/aria-roles.js
straker: Don't reference the svg instead we make spec difference for tabpanel to not be `widget`

lib/standards/aria-roles.js
straker: window is fine

lib/standards/aria-roles.js
straker: Add comment above `type` with `// Note: spec difference` as we want to keep `widget` role

lib/standards/aria-roles.js
straker: Add comment above `type` with `// Note: spec difference` as we want to keep `widget` role

straker:
Will also need to add exception to `focus-order-semantics` to allow `alert` and `alertdialog` roles (add unit & integration tests to match)

lib/standards/aria-roles.js
gaiety-deque: Since we said `window` is fine for `alertdialog` I assume it's fine here as well

lib/rules/widget-not-inline-matches.js
gaiety-deque: Undid this change 2d185f65

lib/standards/aria-roles.js
gaiety-deque: Removed svg as comment reference 2d185f65

lib/standards/aria-roles.js
gaiety-deque: Reverted this change 2d185f65

gaiety-deque:
> Will also need to add exception to `focus-order-semantics` to allow `alert` and `alertdialog` roles (add unit & integration tests to match)

@straker added an exception to `focus-order-semantics`, renamed check and added a `window` test describe for both allowed role types. Refactored the test check as well, I'm all about verbose test files with repetition but 500+ lines had me scrolling all over the place losing track so I simplified down the checks to a static easier to comprehend list facd65f6 (and [bfa95ae](https://github.com/dequelabs/axe-core/pull/4421/commits/bfa95ae2ff233de32390777f735cd35ef1e50fa4))

lib/standards/aria-roles.js
```diff
@@ -17,13 +18,13 @@
 */
 const ariaRoles = {
   alert: {
-    type: 'widget',
+    type: 'structure',
```
WilcoFiers, line 21: This creates a problem for `focus-order-semantics`. With this change, the following is going to start failing. I don't think we should want that.

```html
<div role="alert" tabindex="0"> Hello world </div>
```

This is what we talked about on Thursday. We'll need some way to exempt these roles that are no longer considered as widgets in the `focus-order-semantics` rule. I think the best way to do that is just to add them to the `VALID_ROLES_FOR_SCROLLABLE_REGIONS` object in `valid-scrollable-semantics-evaluate`.

lib/standards/aria-roles.js
```diff
@@ -816,7 +819,7 @@ const ariaRoles = {
     superclassRole: ['section']
   },
   timer: {
-    type: 'widget',
+    type: 'structure',
```
WilcoFiers, line 822: When you're adding roles to `valid-scrollable-semantics`, I think you can leave this one out. `role=timer` doesn't seem like it should be focusable.

lib/standards/aria-roles.js
```diff
@@ -340,7 +343,7 @@ const ariaRoles = {
     superclassRole: ['landmark']
   },
   marquee: {
-    type: 'widget',
+    type: 'structure',
```
WilcoFiers, line 346: When you're adding roles to `valid-scrollable-semantics`, I think you can leave this one out. `role=marquee` doesn't seem like it should be focusable.

test/checks/aria/has-widget-or-window-role.js
```diff
@@ -0,0 +1,118 @@
...
+  'use strict';
+
+  let node;
+  const fixture = document.getElementById('fixture');
+  const checkContext = axe.testUtils.MockCheckContext();
+  const evaluate = currentNode =>
+    axe.testUtils
+      .getCheckEvaluate('has-widget-or-window-role')
+      .call(checkContext, currentNode);
+  const roles = {
```
WilcoFiers, line 11: This feels a little brittle. Generally what we do is we have the unit tests just check the code works. One or two examples of each role type is enough. And then we have fairly extensive integration tests. Those are a little more telling.

WilcoFiers (changes requested):
Can you add integration tests for:
1. Show that target-size no longer applies to some of the things that are no longer widgets. Most importantly alertdialog
2. Show that target-size no longer treats elements with those roles as neighbor targets
3. Show which roles are still allowed to be focusable under valid-scrollable-semantics, and which are not

test/checks/aria/has-widget-or-window-role.js
gaiety-deque: I'll be reverting this change since the desire is to not change this check.

Though I believe this change was not brittle. This was a refactor from ~500 lines of tests to ~100 that checked the same things, but it was difficult to parse by a person before. I found myself getting quite lost in the code previously. The underlying test code was the same with this change.

lib/checks/aria/has-widget-or-window-role-evaluate.js
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Check if an elements `role` attribute uses any widget, composite, window abstract role values.
+ *
+ * Widget roles are taken from the `ariaRoles` standards object from the roles `type` property.
+ *
+ * @memberof checks
+ * @return {Boolean} True if the element uses a `widget`, `composite`, or `window` abstract role (type). False otherwise.
+ */
+// # TODO: change to abstract role for widget and window
```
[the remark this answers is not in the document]
gaiety-deque: Not intentional, reverted a07c488

WilcoFiers (approved):

Issue #4387, opened 2024-03-27
target-size must check whether nearby elements are obscured


WilcoFiers:
The "Accept all cookies" link should not fail. The nearby "go to top" link is obscured:

```html
<div style="position: fixed; bottom: 0">
  <a href="#">Go to top</a>
</div>
<div style="position: fixed; bottom: 0; left: 0; right: 0; background: #eee">
  Cookies: <a href="#">Accept all cookies</a>
</div>
```

**Note** There are open issues related to this. Review them and keep them in mind when working on this.

padmavemulapati:
Verified with the latest axe-core-develop branch code base,
`target-size` rule ignores and obscured for the elements nearby , findings can be in passes


![image](https://github.com/dequelabs/axe-core/assets/47409579/7bda1a25-80ba-4486-a74f-ea20ba4d1f5c)

PR #4422, opened 2024-04-22
fix(target-size): pass for element that is nearby elements that are obscured

straker:
Had to update how we handled the too many rects break early since it would return an empty array, which when looking at the lengths of the arrays in `getOffset` made it difficult to know which case needed to be handled (returned empty due to too many rects or returned empty because there wasn't any visible rect).

Closes: #4387

### DIFF
--- a/lib/checks/mobile/target-offset-evaluate.js
+++ b/lib/checks/mobile/target-offset-evaluate.js
@@ -20,13 +20,29 @@ export default function targetOffsetEvaluate(node, options, vNode) {
       continue;
     }
     // the offset code works off radius but we want our messaging to reflect diameter
-    const offset =
-      roundToSingleDecimal(getOffset(vNode, vNeighbor, minOffset / 2)) * 2;
-    if (offset + roundingMargin >= minOffset) {
-      continue;
+    try {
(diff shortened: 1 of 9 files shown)


test/integration/full/target-size/too-many-rects.js
```diff
@@ -9,6 +9,7 @@ describe('target-size too many rects test', () => {
         elementRef: true
       };
       const context = {
+        include: '#incomplete',
```
straker, line 12: On chrome for some reason it was failing with 2 incomplete instead of 1 where the other was the target `h1 a` (which is a mocha heading).

lib/checks/mobile/target-offset-evaluate.js
```diff
@@ -20,8 +20,27 @@ export default function targetOffsetEvaluate(node, options, vNode) {
...
+    let offset = null;
+    try {
+      offset = getOffset(vNode, vNeighbor, minOffset / 2);
+    } catch (err) {
+      if (err.message.startsWith('splitRects')) {
+        this.data({
+          messageKey: 'tooManyRects',
+          closestOffset: 0,
+          minOffset
+        });
```
straker, line 32: Decided that instead of failing this should just incomplete as well

Issue #4428, opened 2024-04-25
prototype.js breaks axe-core v4.9.0


straker:
Websites that use `prototype.js` cause axe-core v4.9.0 to not be able to load. When trying to load axe-core on the page we get an error `Uncaught TypeError: No color space found with id = "oklch"`.

In order to get v4.9.0 working we are going to revert our update to colorjs from v0.5.0. This will allow axe-core to load on the page but will cause color-contrast to error during it's run. To fix that we're going to load an `Array.from` polyfill right before we call colorjs code to fix the problem, then restore the `prototype.js` `Array.from` implementation right afterwards.

romankulkovsf:
>

padmavemulapati:
Validated with the latest axe-core dev branch code base, now on websites that uses prototype.js  (http://script.aculo.us/), we are able to load and getting the a11y issues if any. But not getting color space with id="oklch"


https://github.com/dequelabs/axe-core/assets/47409579/3154792d-757b-4246-b6d4-f4076ba02a78


## Environment:
| **Label**                        | **Value** |
| -------------------------- | ------ |
| **Product**               | _axe-core_ |
|**Version** | _4.9.0(Dev Branch Code base_ |
| **OS-Details**     | _MAC- Intel Core i7 - 11.6.8 _ |
|**BrowserDetails**|_Chrome Version 124.0.6367.119 (Official Build) (64-bit)_|

PR #4429, opened 2024-04-25
revert(color-contrast): revert upgrade to colorjs.io v0.5.0 to work with prototype.js

straker:
This [puts back the v0.4.3 code](https://github.com/dequelabs/axe-core/pull/4366) in `color.js` to handle colorjs.io not handling rad and turn values in hsl.

I also decided to use core-js `Array.from` polyfill rather than our own internal one since I needed to bring it in outside of the `polyfill.js` file and could then add it as an import and reuse it in both places.

Closes: https://github.com/dequelabs/axe-core/issues/4428

### DIFF
--- a/lib/commons/color/color.js
+++ b/lib/commons/color/color.js
@@ -1,6 +1,7 @@
-import { Colorjs } from '../../core/imports';
+import { Colorjs, ArrayFrom } from '../../core/imports';
 
 const hexRegex = /^#[0-9a-f]{3,8}$/i;
+const hslRegex = /hsl\(\s*([\d.]+)(rad|turn)/;
 
 /**
  * @class Color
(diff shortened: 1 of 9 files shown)


dbjorge (changes requested):
Mostly LGTM, just notice one issue with the old code we're reverting back to.

Verified that this doesn't negatively impact bundle size and doesn't require any LICENSE-3RD-PARTY update.

lib/commons/color/color.js
```diff
@@ -1,6 +1,7 @@
-import { Colorjs } from '../../core/imports';
+import { Colorjs, ArrayFrom } from '../../core/imports';
 
 const hexRegex = /^#[0-9a-f]{3,8}$/i;
+const hslRegex = /hsl\(\s*([\d.]+)(rad|turn)/;
```
dbjorge, line 4: This fails the following missing unit tests:

js
```
      it('supports negative rad on hue', () => {
        const c = new Color();
        c.parseColorFnString('hsl(-3.49rad, 40%, 50%)');
        assert.equal(c.red, 77);
        assert.equal(c.green, 179);
        assert.equal(c.blue, 144);
        assert.equal(c.alpha, 1);
      });

      it('supports negative turn on hue', () => {
        const c = new Color();
        c.parseColorFnString('hsl(-0.556turn, 40%, 50%)');
        assert.equal(c.red, 77);
        assert.equal(c.green, 179);
        assert.equal(c.blue, 144);
        assert.equal(c.alpha, 1);
      });
```

straker: Good catch!

dbjorge (approved):

Montiwa11:
5c5ede0c02ab6e0a62a0e13a4ff5d1e27c3db353

PR #4430, opened 2024-04-25
chore: set up vscode config to attach to test:debug

dbjorge:
This updates the existing `test:debug` run script to pass `--remote-debugging-port` to chrome, and then sets up a vscode launch profile to connect to that port.

No QA required.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,18 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "type": "chrome",
+      "request": "attach",
+      "name": "Attach to Karma test:debug",
+      "address": "localhost",
+      "port": 9765, // keep in sync with debugPort in karma.conf.js
+      "webRoot": "${workspaceFolder}",
+      "sourceMaps": true,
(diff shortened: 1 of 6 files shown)


.vscode/launch.json
```diff
@@ -0,0 +1,18 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "type": "chrome",
+      "request": "attach",
+      "name": "Attach to Karma test:debug",
+      "address": "localhost",
+      "port": 9765, // keep in sync with debugPort in karma.conf.js
```
straker, line 9: Does .vscode not mind the comment in a json file?

straker (approved):
Tested without vscode and it doesn't break running `test:debug`.

.vscode/launch.json
romankulkovsf: > Does .vscode not mind the comment in a json file?

dbjorge: Yes, it supports this. https://code.visualstudio.com/docs/languages/json#_json-with-comments

PR #4435, opened 2024-04-30
fix(aria-allowed-attr): allow meter role allowed aria-* attributes on meter element

straker:
Turns out we didn't list the `meter` element's implicit role.

Ref: https://github.com/w3c/html-aria/issues/517

QA notes: verify a `meter` element allows non-global aria attributes from the [`meter` role](https://www.w3.org/TR/wai-aria-1.2/#meter).


WilcoFiers (approved):

romankulkovsf:
Here

PR #4444, opened 2024-05-03
style: `var` -> `let` and `const` codemod

gaiety-deque:
adds `no-var` rule to eslint
(this also pairs well with the already enabled `prefer-const` and `no-redeclare`)

Who else does this?
  - Google (JS style guide) https://google.github.io/styleguide/jsguide.html#features-local-variable-declarations
  - Airbnb (popular JS style guide) https://github.com/airbnb/javascript

Why do this?

Gaiety's explanation: "`let` works the way most devs think `var` works. scope leaking and variable hoisting is a messy confusing idea, JS got updated for a reason" also it's 2024, this isn't too controversial anymore

- https://www.freecodecamp.org/news/var-let-and-const-whats-the-difference/
- https://blog.javascripttoday.com/blog/var-vs-let-vs-const/ "The var keyword was originally used for declaring variables in JavaScript. If you’ve browsed an old codebase or an older JavaScript tutorial, you have probably seen it."
- https://medium.com/@rithinmenezes/understanding-let-var-and-const-in-javascript-a-guide-to-variable-declarations-f271c5b8dc79
- Airbnb's why https://airbnb.io/javascript/#variables

<< Describe the changes >>

Closes:


.eslintignore
```diff
@@ -7,5 +7,6 @@ doc/api/*
 doc/examples/jest_react/*.js
 
 lib/core/imports/*.js
+lib/core/utils/uuid.js
```
gaiety-deque, line 10: Not _really_ our code, maybe belongs in a vendor file? Felt weird touching it.

gaiety-deque:
Closing this in favor of doing smaller PR's piece by piece over time, but I'll reference the changes here as the ultimate goal.

PR #4452, opened 2024-05-07
perf: memoize DqElement

straker:
Noticed this when trying to debug perf issues in `duplicate-id-aria`. We've run into problems on sites that have a module repeat 1000s of times on the page and the module has an aria id that is also then repeated. Axe-core would take a really long time to run the rule. Looking into it what I discovered is that a majority of the time was spent on resolving the `relatedNodes` for each check. Since each each duplicate id node was also in the `relatedNodes` for every other node, this caused the single node to [be converted to a DqElement](https://github.com/dequelabs/axe-core/blob/develop/lib/core/utils/check-helper.js#L46) _n_ times. This lead to many performance problems, but specifically calling the `getSelector` of a DqElement would call `outerHTML` for the node _n*2_ times which would be very slow. 

To solve this I decided to memoize the DqElement creation. That way a single node will only ever output a single DqElement, thus saving significant time in creation. Not only that but every time a node appears in a result (either as the check node or a related node) the memory is now shared so this change also reduces the in-memory size of the results object.

Testing a simple page with 5000 nodes of duplicate id, here are the results for running the `duplicate-id-aria` check.

| Before change (in ms)  | After change (in ms) |
| ------------- | ------------- |
| 21,280.1  | 11,841.1  |

_Flamechart before the change. The majority of the time is spent in getSelector_

_Chrome performance timer of getSelector showing it spent 12000ms total in the function_

_Flamechart after the change. Time is now spent mostly resolving the cache which results in no time spent in getSelector_

### DIFF
--- a/lib/core/utils/dq-element.js
+++ b/lib/core/utils/dq-element.js
@@ -4,6 +4,7 @@ import getXpath from './get-xpath';
 import getNodeFromTree from './get-node-from-tree';
 import AbstractVirtualNode from '../base/virtual-node/abstract-virtual-node';
 import cache from '../base/cache';
+import memoize from './memoize';
 
 const CACHE_KEY = 'DqElm.RunOptions';
 
@@ -36,7 +37,10 @@ function getSource(element) {
  * @param {Object} options Propagated from axe.run/etc
  * @param {Object} spec Properties to use in place of the element when instantiated on Elements from other frames
  */
-function DqElement(elm, options = null, spec = {}) {
+const DqElement = memoize(function DqElement(elm, options, spec) {
+  options ??= null;
+  spec ??= {};
+
   if (!options) {
     options = cache.get(CACHE_KEY) ?? {};
   }
@@ -82,7 +86,9 @@ function DqElement(elm, options = null, spec = {}) {
   if (!axe._audit.noHtml) {
     this.source = this.spec.source ?? getSource(this._element);
   }
-}
+
+  return this;
+});
 
 DqElement.prototype = {
   /**
--- a/test/core/utils/node-serializer.js
+++ b/test/core/utils/node-serializer.js
@@ -91,14 +91,19 @@ describe('nodeSerializer', () => {
     });
 
     it('skips computing props turned off with runOptions', () => {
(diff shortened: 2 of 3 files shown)


lib/core/utils/dq-element.js
```diff
@@ -36,7 +37,10 @@ function getSource(element) {
  * @param {Object} options Propagated from axe.run/etc
  * @param {Object} spec Properties to use in place of the element when instantiated on Elements from other frames
  */
-function DqElement(elm, options = null, spec = {}) {
+const DqElement = memoize(function DqElement(elm, options, spec) {
```
straker, line 40: Default parameters get transformed out of the function after babel, which meant that only `elm` was left in the argument list for memoize, so `options` and `spec` were ignored.

lib/core/utils/dq-element.js
```diff
@@ -82,7 +86,9 @@ function DqElement(elm, options = null, spec = {}) {
   if (!axe._audit.noHtml) {
     this.source = this.spec.source ?? getSource(this._element);
   }
-}
+
+  return this;
```
straker, line 90: Memoize requires something to be returned otherwise it caches the function as `undefined`

test/core/utils/node-serializer.js
```diff
@@ -91,14 +91,19 @@ describe('nodeSerializer', () => {
...
       };
-      Object.defineProperty(dqElm, 'selector', { get: throws });
-      Object.defineProperty(dqElm, 'ancestry', { get: throws });
-      Object.defineProperty(dqElm, 'xpath', { get: throws });
+
+      const dqElm = new DqElement(
+        fixture,
+        {},
+        {
+          selector: { get: throws },
```
straker, line 102: With memoize this style of define would propagate to all tests and cause issues. Since `DqElement` takes the spec as truth, just passed the functions in as the spec.

WilcoFiers (dismissed):
Can you add a test showing that when you instantiate a DqElm for the second time you get the same object back?

WilcoFiers (approved):

Issue #4400, opened 2024-04-04
Don't break on window.CSS === null


WilcoFiers:
Angular + Jest documentation recommends a solution which ends up setting `window.CSS` to null. This creates an issue when axe-core imports colorjs, which has a falsey check for `window.CSS`, but when `null` errors out.

Since ColorJS didn't accept [our patch](https://github.com/color-js/color.js/pull/472/files), we'll try to use [patch-package](https://www.npmjs.com/package/patch-package) to fix the issue ourselves. We'll need some tests to go along with that to confirm our solution continues to work after updating ColorJS.

patch-package should probably run in the build step, not as a post-install.

gaiety-deque:
Just for full context, `window.CSS` is no longer recommended in the Jest docs (thanks Michael) https://github.com/thymikee/jest-preset-angular/commit/ac30648347ab41e0cbce741f66ae2a06b766fe13 which is live on their docs site https://thymikee.github.io/jest-preset-angular/docs/getting-started/installation/#global-mocks

PR #4456, opened 2024-05-10
fix: colorio.js patch mocking CSS

gaiety-deque:
Adds `patch-package` as suggested in the issue this PR fixes. The goal is to allow `window.CSS` to be mocked `null` in situations such as using JSdom.

Tests cover a version that is patched and one that is unpatched to demonstrate the patch truly fixes the issue.

fixes: #4400

---

**Developer Notes/Questions** for review:

- Someone mentioned the patch step should _not_ be `postinstall`, what should it be instead?
- For test coverage purposes it's useful to keep an unpatched version, currently this is done manually in `patches/color.unpatched.js` but perhaps there's a better way
- It's a bit gross modifying so many pre-compiled `dist` files
  - Instead of a patch, we could pull in colorjs.io as a submodule and build it ourselves, unsure how to propagate that to consumers of axe-core however
  - Perhaps this is fine and I could document how to update the patch in the future


package.json
```diff
@@ -113,7 +113,8 @@
     "prepare": "husky",
     "prebuild": "node ./build/check-node-version.js",
     "pretest": "node ./build/check-node-version.js",
-    "postbuild": "prettier --write ./locales/_template.json ./doc/rule-descriptions.md"
+    "postbuild": "prettier --write ./locales/_template.json ./doc/rule-descriptions.md",
+    "postinstall": "patch-package"
```
dbjorge, line 117: We really want to avoid adding a postinstall script to axe-core - this would add it as a step that axe-core consumers run when doing `npm install axe-core`, which we don't want. Instead, we'll want to run the command line as a build step. This will likely mean adding a grunt task wrapping it to `/build/tasks/` that you reference from `Gruntfile.js`.

patches/colorjs.io+0.5.0.patch
dbjorge, line 1: This patch is way bigger than it needs to be; we don't need to patch the sourcemap files and we don't need to patch all of their dist variants, just the one we actually use in our build process (I'm not sure offhand which one that is, but the test should tell you when you've found it)

patches/color.unpatched.js
dbjorge, line 1: I don't think we should check this into the repo; if we need to test against this, we should emit it by copying from node_modules as part of the build step that patches the original

dbjorge (dismissed):

patches/colorjs.io+0.4.5.patch
```diff
@@ -0,0 +1,60 @@
...
+-if (typeof CSS !== "undefined" && CSS.supports) {
++if (typeof CSS !== "undefined" && CSS?.supports) {
+ 	// Find widest supported color space for CSS
+ 	for (let space of [lab, REC2020, P3]) {
+ 		let coords = space.getMinCoords();
+ 		let color = {space, coords, alpha: 1};
+ 		let str = serialize(color);
+ 
+-		if (CSS.supports("color", str)) {
++		if (CSS?.supports("color", str)) {
```
straker, line 18: This one shouldn't need to be patched since by here `CSS.supports` exists and is not `null` due to the if statement above.

patches/colorjs.io+0.4.5.patch
```diff
@@ -0,0 +1,60 @@
...
+@@ -2039,7 +2039,7 @@ if (typeof CSS !== "undefined" && CSS.supports) {
+ function display (color, {space = defaults.display_space, ...options} = {}) {
+ 	let ret = serialize(color, options);
+ 
+-	if (typeof CSS === "undefined" || CSS.supports("color", ret) || !defaults.display_space) {
++	if (typeof CSS === "undefined" || CSS?.supports("color", ret) || !defaults.display_space) {
+ 		ret = new String(ret);
+ 		ret.color = color;
+ 	}
+diff --git a/node_modules/colorjs.io/dist/color.js b/node_modules/colorjs.io/dist/color.js
```
straker, line 31: We should only need to patch one of these files (either `color.js` or `color.cjs`)

gaiety-deque: `colorio.js`'s `package.json` has the following `exports` defined:

```js
	"exports": {
		".": {
			"types": "./types/index.d.ts",
			"import": "./dist/color.js",
			"require": "./dist/color.cjs"
```

and also in the package json

```js
	"main": "./dist/color.cjs",
	"module": "./dist/color.js",
```

so it seemed safer to do both, but we can pick one if we're certain only that one will be used

straker: Our build process will only use one of them to build, so in favor of making the patch as small possible I  think we should patch the one that our build grabs and leave the other alone.

gaiety-deque: Sounds good, switched to only `color.js` then

I'll need to repatch it when https://github.com/dequelabs/axe-core/pull/4449 is merged

gaiety-deque: Nevermind, I rebased my branch so now it'll flow through naturally in theory. Switched patch to be for colorjs.io `0.4.3` instead.

Gruntfile.js
```diff
@@ -6,12 +6,14 @@ camelcase: ["error", {"properties": "never"}]
 module.exports = function (grunt) {
   'use strict';
 
+  grunt.loadNpmTasks('grunt-exec');
```
stephenmathieson, line 9: At some point, we're going to have to stop adding tasks to Grunt if we're ever going to get rid of it.

Should now be that time?

straker: Since we want this to be part of the build it needs to go in the grunt `build` step so `npm develop` builds it as well. We won't be moving off grunt any time soon and not in this PR so a grunt step is fine for now.

romankulkovsf (approved):
Run

romankulkovsf:
Gd

romankulkovsf (approved):

Gruntfile.js
WilcoFiers: I don't think I said it needed to be in grunt :P. We did a good while back agree not to add more grunt tasks if we could at all avoid it. It seems viable to put this work in prebuild / predevelop scripts instead. Did you try that @gaiety-deque? If that was more difficult I can live with a Grunt solution, but if we can avoid these extra dependencies + more grunt tasks I think that's the way to go on this.

WilcoFiers (dismissed):

Gruntfile.js
gaiety-deque: Apologies @WilcoFiers I didn't mean to imply that you explicitly said it had to be in the _grunt_ build step, just that the issue requested it be a part of the build step. Since our build step just calls "grunt" currently and the pre/post build steps were more test-like in nature it felt like we were intentionally doing these things in grunt.

Made a change [1785f2c](https://github.com/dequelabs/axe-core/pull/4456/commits/1785f2c7b5845cf1611051586960a4efb271937c) to move it to be commands run in the package json scripts.

.eslintignore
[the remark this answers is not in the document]
gaiety-deque: Weird thought I fixed this but must have snuck back in through all the rebasing, fixed [87c7845](https://github.com/dequelabs/axe-core/pull/4456/commits/87c784598918e385a61a34eb87e3b1685d154589)

package.json
```diff
@@ -111,9 +113,11 @@
     "sri-validate": "node build/sri-update --validate",
     "fmt": "prettier --write .",
     "prepare": "husky",
-    "prebuild": "node ./build/check-node-version.js",
+    "prebuild": "node ./build/check-node-version.js && npm run unpatch",
+    "predevelop": "npm run unpatch",
     "pretest": "node ./build/check-node-version.js",
-    "postbuild": "prettier --write ./locales/_template.json ./doc/rule-descriptions.md"
+    "postbuild": "prettier --write ./locales/_template.json ./doc/rule-descriptions.md && npm run patch",
```
WilcoFiers, line 119: I suspect you mean to unpatch here. I'm not sure about this unpatching at the end of the flow. It feels like there are potential gaps there, especially with develop. Might be better if we had one script that unpatches, copies, then patches, which runs before develop and build. That feels like a slightly safer way to ensure the files in `patch/unpatched` are always up to date.

package.json
```diff
@@ -111,9 +113,11 @@
     "sri-validate": "node build/sri-update --validate",
     "fmt": "prettier --write .",
     "prepare": "husky",
-    "prebuild": "node ./build/check-node-version.js",
+    "prebuild": "node ./build/check-node-version.js && npm run unpatch",
```
WilcoFiers, line 116: One more thing to consider is that we could use the `prepare` script instead of a prebuild script. The reason we can't use postinstall is because that would trigger when someone uses `npm i axe-core`, but `prepare` only run when you call npm install in the axe-core repo. It would prevent having to run patch / unpatch over and over. @straker @dbjorge, do you see any reason not to do that?

WilcoFiers (changes requested):

package.json
dbjorge: The main gotchas I'm aware of with `prepare` are that:

- It runs within a package consumer if the consumer uses a `"git:github.com/dequelabs/axe-core.git"` dependency on axe-core rather than a more typical npm dependency
- It's not treated consistently by all packaging tools; yarn v2+ ignores it, for example.

I don't think either of those matter enough to make me want to avoid it here; the former case is already broken if it isn't running our build process, and I don't think we need optimize for the latter (so long as there's a test that would catch the patch not applying as expected).

test/core/patch.js
```diff
@@ -0,0 +1,63 @@
...
+
+function resetWindowCSSMock() {
+  Object.defineProperty(window, 'CSS', { value: originalWindowCSS });
+}
+
+function mockWindowCSS() {
+  Object.defineProperty(window, 'CSS', { value: null });
+}
+
+describe('patch test', function () {
```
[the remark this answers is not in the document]
gaiety-deque: Good catch, tests added in 8b781657. I manually ensured these tests failed if I didn't run the patch steps prior to a build. Feeling confident the test coverage is there now that we are both testing the built files and also the colorjs dependency patched and unpatched. We _could_ also keep around a copy of build axe unpatched to test against but that sounds excessive in my opinion, so I think this is good.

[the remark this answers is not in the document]
gaiety-deque: Hmm darn... the very act of dynamically importing `axe.js` or `axe.min.js` is causing many other failures. If we really need this test then we may have to find another way to do it, and I don't believe a full integration test will handle it either based on my prior attempts while working on this.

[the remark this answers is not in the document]
gaiety-deque: Got it working in the latest commit :) moved to a full integration test

package.json
```diff
@@ -110,7 +112,7 @@
     "sri-update": "grunt build && node build/sri-update && git add sri-history.json",
     "sri-validate": "node build/sri-update --validate",
     "fmt": "prettier --write .",
-    "prepare": "husky",
+    "prepare": "husky && npm run unpatch && npm run patch",
```
WilcoFiers, line 115: Do we actually need to unpatch first? I thought you had done that so we could get an unpatched version to test against. You patch-package just skips if you run it on an already patched package.

WilcoFiers (commented):
Got a question

package.json
gaiety-deque: You're correct 09a36d54

WilcoFiers (approved):

PR #4451, opened 2024-05-07
style(lib/checks): var -> let and const codemod

gaiety-deque:
This is a piece of a larger creative time PR so it's easier to review, the full scope can be seen here and includes eslint adjustments not seen here https://github.com/dequelabs/axe-core/pull/4444

---

Who else does this?
  - Google (JS style guide) https://google.github.io/styleguide/jsguide.html#features-local-variable-declarations
  - Airbnb (popular JS style guide) https://github.com/airbnb/javascript

Why do this?

Gaiety's explanation: "`let` works the way most devs think `var` works. scope leaking and variable hoisting is a messy confusing idea, JS got updated for a reason" also it's 2024, this isn't too controversial anymore

- https://www.freecodecamp.org/news/var-let-and-const-whats-the-difference/
- https://blog.javascripttoday.com/blog/var-vs-let-vs-const/ "The var keyword was originally used for declaring variables in JavaScript. If you’ve browsed an old codebase or an older JavaScript tutorial, you have probably seen it."
- https://medium.com/@rithinmenezes/understanding-let-var-and-const-in-javascript-a-guide-to-variable-declarations-f271c5b8dc79
- Airbnb's why https://airbnb.io/javascript/#variables

PR #4457, opened 2024-05-10
style(lib/core): var -> let and const codemod

gaiety-deque:
Just like https://github.com/dequelabs/axe-core/pull/4451 but for `lib/core`

This does add an `.eslintignore` line, perhaps we'd rather skip a file a different way.

Smaller part of the full vision: https://github.com/dequelabs/axe-core/pull/4444


romankulkovsf (approved):

.eslintignore
```diff
@@ -7,5 +7,6 @@ doc/api/*
 doc/examples/jest_react/*.js
 
 lib/core/imports/*.js
+lib/core/utils/uuid.js
```
WilcoFiers, line 10: Did you mean to include this?

WilcoFiers (changes requested):
the eslintignore thing. Rest looks good

.eslintignore
gaiety-deque: I did @WilcoFiers. It felt weird to update a file that I don't think we wrote (judging by the comment at the top of `uuid.js`. Eventually when all of these codemods are merged the last PR will turn on the eslint rule `no-var` and unless we either update this file or ignore it, this file will throw an eslint warning.

WilcoFiers: @gaiety-deque Merge conflict, can you update?

WilcoFiers (approved):

PR #4461, opened 2024-05-14
chore: Update locale de.json

EvilKarter:
Update locale de.json


straker (approved):
Thanks for the update. Reviewed for security.

PR #4462, opened 2024-05-14
chore: upgrade to eslint9

michael-siek:
Upgrade to eslint v9 flat config. Custom eslint rules have been tested and work in the new config file. Removed unused vars in try catch blocks due to[ breaking change in eslint v9](https://github.com/eslint/eslint/pull/17932). Use [globals](https://www.npmjs.com/package/globals) to include [globals env](https://eslint.org/docs/latest/use/configure/language-options#predefined-global-variables)


package.json
```diff
@@ -137,11 +137,11 @@
     "emoji-regex": "^10.2.1",
     "es6-promise": "^4.2.8",
     "esbuild": "^0.10.x",
-    "eslint": "^8.31.0",
-    "eslint-config-prettier": "^9.0.0",
-    "eslint-plugin-mocha-no-only": "^1.1.1",
```
michael-siek, line 142: unfortunately I was unable to reuse this in eslint v9 (this package has not been updated in 4 yrs and does not support the new eslint syntax)

WilcoFiers: We could add a `no-restricted-syntax` rule that prohibits `it.only`, `test.only` and `describe.only`.

WilcoFiers (commented):
LGTM. Have a thought about the .only plugin to consider.

package.json
michael-siek: I just wrote the fix to the plugin https://github.com/lewazo/eslint-mocha-no-only/pull/6

eslint.config.js
```diff
@@ -0,0 +1,193 @@
...
+      globals: {
+        axe: true,
+        Promise: true,
+        ...globals.node,
+        ...globals.es2015
+      }
+    }
+  },
+  {
+    files: ['lib/**/*.js'],
```
straker, line 106: This needs [to ignore the reporters and check after methods](https://github.com/dequelabs/axe-core/blob/develop/.eslintrc.js#L99C23-L99C72) as we don't want them to have access to `window` or `document`.

```suggestion
    files: ['lib/**/*.js'],
    // reporters and check after methods should not have access to window or document
    ignores: ['lib/core/reporters/**/*.js', 'lib/**/*-after.js'],
```

eslint.config.js
```diff
@@ -0,0 +1,193 @@
+const prettier = require('eslint-config-prettier');
+const globals = require('globals');
+const mochaNoOnly = require('eslint-plugin-mocha-no-only');
+
+module.exports = [
+  prettier,
+  {
+    rules: {
```
straker, line 8: Should leave a comment [as to why we don't configure our code to allow `globals.browser`](https://github.com/dequelabs/axe-core/blob/develop/.eslintrc.js#L104C9-L104C79)

```js
    // lib files should not access global window properties without going through window (i.e. do not add globals.browser)
```

straker (dismissed):

WilcoFiers (approved):

Issue #4460, opened 2024-05-14
landmark-unique does not reflect latest HTML-role mappings


gabalafou:
### Product

axe Extension

### Product Version

4.79.3

### Latest Version

- [X] I have tested the issue with the latest version of the product

### Issue Description

Note: I wrote up this issue based on the browser extension, but I also checked that this issue holds with the newer axe-core 4.9.1 using [@axe-core/cli](https://github.com/dequelabs/axe-core-npm/blob/develop/packages/cli/README.md).

#### Expectation

The following URL should have zero [landmark-unique](https://pydata-sphinx-theme.readthedocs.io/en/latest/examples/kitchen-sink/generic.html) violations:
https://pydata-sphinx-theme.readthedocs.io/en/latest/examples/kitchen-sink/generic.html

#### Actual

Axe finds several landmark-unique violations at the above URL. 

These landmark-unique violations are for several `<aside>` tags on that page that are actually not represented by the latest browsers as role=complementary landmarks.

#### How to Reproduce

Using:

- Chromium 124
- axe DevTools v4.79.3 (axe-core 4.8.4)

Scan the following URL:
https://pydata-sphinx-theme.readthedocs.io/en/latest/examples/kitchen-sink/generic.html

Observe that the tool reports several instances of the landmark-unique violation.

#### Additional context

The latest browsers follow the [HTML-role mapping specification](https://w3c.github.io/html-aam) and do not assign `role=complementary` to unnamed `<aside>` tags that are scoped to a sectioning content element ([section 3.4.9](https://w3c.github.io/html-aam/#el-aside)). 

But apparently Axe-core treats all `<aside>` tags in the document as landmarks, which triggers the landmark-unique violation.

For more context:

- This was the motivating issue on my end: https://github.com/pydata/pydata-sphinx-theme/issues/1479
- Here are links to issues showing the problem fixed recently in Firefox and Safari (it was fixed even longer ago in Chrome):
  - Firefox: [Conditional role mapping for aside element (WPT html-aam/roles-contextual.html el-aside-in-section-without-name)
](https://bugzilla.mozilla.org/show_bug.cgi?id=1749505)
  - Safari: [Assign generic role to aside tag within the sectioning content elements #20013
](https://github.com/WebKit/WebKit/pull/20013)

WilcoFiers:
[...]

PR #4469, opened 2024-05-21
fix(landmark-unique): follow spec, aside -> landmark

gabalafou:
Update the landmark-unique rule matcher for aside elements so that they are treated as landmarks using the same criteria specified in [Sections 3.4.8 and 3.4.9 of the HTML Accessibility API Mappings 1.0](https://w3c.github.io/html-aam/#el-aside-ancestorbodymain). 

Closes: #4460

### DIFF
--- a/lib/commons/standards/implicit-html-roles.js
+++ b/lib/commons/standards/implicit-html-roles.js
@@ -13,14 +13,29 @@ import { closest } from '../../core/utils';
 import cache from '../../core/base/cache';
 import getExplicitRole from '../aria/get-explicit-role';
 
-const getSectioningElementSelector = () => {
-  return cache.get('sectioningElementSelector', () => {
+// Sectioning content elements: article, aside, nav, section
+// https://html.spec.whatwg.org/multipage/dom.html#sectioning-content
+const getSectioningContentSelector = () => {
+  return cache.get('sectioningContentSelector', () => {
     return (
       getElementsByContentType('sectioning')
         .map(nodeName => `${nodeName}:not([role])`)
         .join(', ') +
-      ' , main:not([role]), [role=article], [role=complementary], [role=main], [role=navigation], [role=region]'
+      ' , [role=article], [role=complementary], [role=navigation], [role=region]'
     );
+  });
+};
+
+const getSectioningContentPlusMainSelector = () => {
+  // Why is there this similar but slightly different selector?
+  // ->
+  // To be mapped to landmark roles, asides can be scoped to body or main. But
+  // headers and footers must be scoped **only** to body.
+  //   - Header: https://w3c.github.io/html-aam/#el-header-ancestorbody
+  //   - Footer: https://w3c.github.io/html-aam/#el-footer-ancestorbody
+  //   - Aside: https://w3c.github.io/html-aam/#el-aside-ancestorbodymain
+  return cache.get('sectioningContentPlusMainSelector', () => {
+    return getSectioningContentSelector() + ' , main:not([role]), [role=main]';
   });
 };
 
@@ -40,7 +55,7 @@ function hasAccessibleName(vNode) {
   // testing for when browsers give a <section> a region role:
   // chrome - always a region role
   // firefox - if non-empty aria-labelledby, aria-label, or title
-  // safari - if non-empty aria-lablledby or aria-label
+  // safari - if non-empty aria-labelledby or aria-label
   //
   // we will go with safaris implantation as it is the least common
   // denominator
@@ -58,7 +73,18 @@ const implicitHtmlRoles = {
     return vNode.hasAttr('href') ? 'link' : null;
   },
   article: 'article',
-  aside: 'complementary',
+  aside: vNode => {
+    if (
+      closest(vNode.parent, getSectioningContentSelector()) &&
+      // An aside within sectioning content can still be mapped to
+      // role=complementary if it has an accessible name
+      !hasAccessibleName(vNode)
+    ) {
+      return null;
+    }
+
+    return 'complementary';
+  },
   body: 'document',
   button: 'button',
   datalist: 'listbox',
@@ -70,7 +96,10 @@ const implicitHtmlRoles = {
   fieldset: 'group',
   figure: 'figure',
   footer: vNode => {
-    const sectioningElement = closest(vNode, getSectioningElementSelector());
+    const sectioningElement = closest(
+      vNode,
+      getSectioningContentPlusMainSelector()
+    );
 
     return !sectioningElement ? 'contentinfo' : null;
   },
@@ -84,7 +113,10 @@ const implicitHtmlRoles = {
   h5: 'heading',
   h6: 'heading',
   header: vNode => {
-    const sectioningElement = closest(vNode, getSectioningElementSelector());
+    const sectioningElement = closest(
+      vNode,
+      getSectioningContentPlusMainSelector()
+    );
 
     return !sectioningElement ? 'banner' : null;
   },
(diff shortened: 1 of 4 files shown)


lib/rules/landmark-unique-matches.js
```diff
@@ -4,17 +4,18 @@ import { getRole } from '../commons/aria';
...
 import { accessibleTextVirtual } from '../commons/text';
 
+// Sectioning content -
+// https://html.spec.whatwg.org/multipage/dom.html#sectioning-content
+const sectioningContentElements = ['article', 'aside', 'nav', 'section'];
+
 /*
- * Since this is a best-practice rule, we are filtering elements as dictated by ARIA 1.1 Practices regardless of treatment by browser/AT combinations.
- *
- * Info: https://www.w3.org/TR/wai-aria-practices-1.1/#aria_landmark
```
gabalafou, line 10: This comment didn't make that much sense to me and the link was outdated, which is why I changed it.

test/rule-matches/landmark-unique-matches.js
```diff
@@ -128,6 +124,51 @@ describe('landmark-unique-matches', function () {
     });
   });
 
+  describe('aside should not match when scoped to a sectioning content element unless it has an accessible name', function () {
```
gabalafou, line 127: If you look at the rest of the test file, this newly added code is nearly copy-paste of the header and footer tests. Is that preferred, or should I attempt reduce the repetition?

lib/rules/landmark-unique-matches.js
```diff
@@ -46,3 +51,20 @@ function isLandmarkVirtual(vNode) {
...
+ * `main`. If it is within sectioning content, it is not a landmark unless it
+ * has an accessible name.
+ *
+ * https://w3c.github.io/html-aam/#el-aside-ancestorbodymain
+ */
+const sectioningContentSelector = sectioningContentElements.join(',');
+function isAsideLandmark(asideElement) {
+  return (
+    !closest(
+      asideElement.parent, // closest() can match self, which we do not want, so start at parent element
```
gabalafou, line 66: Wouldn't have caught this without tests

lib/rules/landmark-unique-matches.js
```diff
@@ -46,3 +51,20 @@ function isLandmarkVirtual(vNode) {
...
+      sectioningContentSelector
+    ) || !!accessibleTextVirtual(asideElement)
```
gabalafou, line 68: I'm not 100% sure that this function, accessibleTextVirtual(), is 100% equivalent to the "accessible name" requirement of the specification ([Section 3.4.9](https://w3c.github.io/html-aam/#el-aside))

straker:
Thanks for being willing to take on a pr for this. Unfortunately I don't think this is the correct place for the change. 

We have a file [implicit-html-roles](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/standards/implicit-html-roles.js) that dictates what the role of an element is when it has no `role` attribute. In there [we assign `aside` the role of `complimentary`](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/standards/implicit-html-roles.js#L61). Instead what we need to do is figure is two things: 

1. [Is the `aside` part of a sectioning content](https://w3c.github.io/html-aam/#el-aside-ancestorbodymain). We can determine this using the code from [footer and header](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/standards/implicit-html-roles.js#L72-L75). If it's not part of a sectioning content then the function should return `complementary`
2. [Does the `aside` have an accessible name](https://w3c.github.io/html-aam/#el-aside). Again, we can determine this by using the local [hasAccessibleName function](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/standards/implicit-html-roles.js#L39) similar to how the [from role](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/standards/implicit-html-roles.js#L78) works. If it does then return `complementary`, otherwise return `null` to signify no implicit role.

Fixing that should then fix the landmark unique issue as the `aside` will no longer be determined to have a `complementary` role.

gabalafou:
Oh that's great. Putting this aside-complementary logic in the landmark-unique matcher felt off to me, but I patterning off of what was already there. Given what you wrote, I think the isHeaderFooterLandmark() function is redundant. In fact, I just checked and if I comment it out, then run `npm run build && npm run test`, all the tests pass.

Would you like me to update this pull request?

straker:
Yep, that'd be great.

lib/rules/landmark-unique-matches.js
```diff
@@ -42,7 +25,3 @@ function isLandmarkVirtual(vNode) {
 
   return landmarkRoles.indexOf(role) >= 0 || role === 'region';
 }
-
-function isHeaderFooterLandmark(headerFooterElement) {
```
gabalafou, line 46: As mentioned in a previous comment, this function has apparently been redundant ever since the implicit role mapping was introduced/updated.

lib/commons/standards/implicit-html-roles.js
```diff
@@ -58,7 +73,18 @@ const implicitHtmlRoles = {
     return vNode.hasAttr('href') ? 'link' : null;
   },
   article: 'article',
-  aside: 'complementary',
+  aside: vNode => {
+    if (
+      closest(vNode.parent, getSectioningContentSelector()) &&
+      // An aside within sectioning content can still be mapped to
+      // role=complementary if it has an accessible name
+      !hasAccessibleName(vNode)
```
gabalafou, line 81: [Scott O'Hara's test page for aside mapping](https://scottaohara.github.io/testing/aside/) has an explicit test for the title attribute, so this code would fail his test page.

The motivating problem behind this pull request is the landmark-unique check. Ultimately we want good landmarks (distinct, not too many) for screen reader users. In that case, it would be better to flag nested asides with titles because in some browsers (Chrome) those will be given role=complementary, and therefore probably treated as landmarks by screen readers, potentially creating landmark ambiguity and noise.

For some other scenarios or checks, it might be more appropriate to only return the implicit role that the majority of browsers agree upon. 

This is why I did not modify the hasAccessibleName() function in my latest push to this pull request because I'm not sure it makes sense to do so for other accessibility checks. I only know it makes sense to do so for the landmark-unique check.

Should I open a PR that adds a check for title to the function? Then it could be discussed there.

gabalafou (commented):
self review

lib/commons/standards/implicit-html-roles.js
straker: Thanks for looking into that. I tested out `title` attributes on `aside` and it looks like it works in all assistive technologies. So we'll probably want to allow `title` attribute to contribute to accessible name in these cases.

However I also tested out `section` with a `title` and VoiceOver on Safari still continues to not treat it as a region, so we'll not want to update the `hasAccessibleName` function but instead add another check to the if statement in the aside function to check if it has content in the `title` attribute (with an appropriate test)

gabalafou: Thanks! That makes sense.

Just thinking out loud...would it make sense to add an options object to hasAccessibleName()? 

Then we could call `hasAccessibleName({ checkTitle: true })`...

straker: Sounds good to me

test/rule-matches/landmark-unique-matches.js
straker: Repetition is fine, we can clean it up later if we decide to.

straker (approved):

straker:
Awesome work! Thanks for the pr.

Reviewed for security.

PR #4493, opened 2024-06-06
style(lib/rules): var -> const & let

gaiety-deque:
Just like https://github.com/dequelabs/axe-core/pull/4451 but for `lib/rules`. This is the last of `lib/*`

Smaller part of the full vision: https://github.com/dequelabs/axe-core/pull/4444

### DIFF
--- a/lib/rules/landmark-unique-matches.js
+++ b/lib/rules/landmark-unique-matches.js
@@ -1,21 +1,7 @@
 import { isVisibleToScreenReaders } from '../commons/dom';
-import { closest } from '../core/utils';
 import { getRole } from '../commons/aria';
 import { getAriaRolesByType } from '../commons/standards';
 import { accessibleTextVirtual } from '../commons/text';
-
-/*
- * Since this is a best-practice rule, we are filtering elements as dictated by ARIA 1.1 Practices regardless of treatment by browser/AT combinations.
- *
- * Info: https://www.w3.org/TR/wai-aria-practices-1.1/#aria_landmark
- */
-const excludedParentsForHeaderFooterLandmarks = [
-  'article',
-  'aside',
-  'main',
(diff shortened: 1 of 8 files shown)


lib/rules/landmark-unique-matches.js
```diff
@@ -1,8 +1,22 @@
...
+ *
+ * Info: https://www.w3.org/TR/wai-aria-practices-1.1/#aria_landmark
+ */
+const excludedParentsForHeaderFooterLandmarks = [
+  'article',
+  'aside',
+  'main',
+  'nav',
+  'section'
+].join(',');
```
WilcoFiers, line 18: What's this doing here?

lib/rules/svg-namespace-matches.js
```diff
@@ -10,7 +10,8 @@ function svgNamespaceMatches(node, virtualNode) {
 
     // element is svg namespace if its parent is an svg element
     return !!closest(virtualNode, 'svg');
-  } catch {
+  } catch (e) {
+    console.debug(e);
```
WilcoFiers, line 14: Can't have this. Axe-core can't just put things into the console as the console may be piped out to a place that isn't expecting us to inject things into it.

```suggestion
  } catch {
```

WilcoFiers (changes requested):

lib/rules/landmark-unique-matches.js
gaiety-deque: Seems my local `develop` was out of date, fixed in [74e1cdc](https://github.com/dequelabs/axe-core/pull/4493/commits/74e1cdc1a5a23126f7cfc3d7ef7933e0088a9968)

WilcoFiers (approved):

Issue #4385, opened 2024-03-26
Make rule descriptions more consistent


sknep:
### Product

axe-core

### Description

In the rule description tables at https://github.com/dequelabs/axe-core/blob/develop/doc/rule-descriptions.md#rule-descriptions , many start with "Ensures...", but some start with "Ensure..." or "Ensure that...".

Is the description field meant to be an explanation of what the rule is for, or what the reader should do to remediate? The second seems more appropriate, but the implied subject of the sentence (rule vs remediation) changes between rules.

Also, "Ensure touch target have sufficient size and space" -- target should be plural, regardless.

straker:
Thanks for the issue. I've brought this up with the team and we've agreed that we'd like to go with "Ensure" for each of the rule descriptions. If you're willing, we'd love a PR fixing all the rule descriptions to be consistent, as well as fixing the target size rule as you've mentioned.

PR #4499, opened 2024-06-11
chore(text): update text content

marktnoonan:
Hi! I noticed some minor typos and inconsistent text capitalization, as well as what looks like flipped `help` and `description` text in one spot for the `landmark-unique` rule. This PR fixes the ones I was able to spot.

I noticed some overlap with https://github.com/dequelabs/axe-core/issues/4385, I'd be happy to grab the text changes mentioned there in this PR as well, ("Ensures" -> "Ensure" in rule descriptions), if that's not stepping on anybody's toes.


WilcoFiers (changes requested):
Thank you very much @marktnoonan. Would you mind making these changes in their source files instead of in the generated file though?

Sorry about that, we normally put a comment at the top of generated files. Can't put comments in JSON unfortunately.

locales/_template.json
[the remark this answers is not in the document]
marktnoonan: Oh that makes so much sense, yes I'll move changes to the proper place, thanks!

doc/rule-descriptions.md
```diff
@@ -86,9 +86,9 @@
```
marktnoonan, line 86: @WilcoFiers This file is also generated - should I remove these changes from the PR also?

WilcoFiers: No, we try to keep this in sync with the .json files in the repo. We actually need the _template.json file updated too. Would you mind running `npm run build` to update that file as well?

marktnoonan: Ah understood - we want the generated files updated and committed, but they are not edited directly, and so they stay current with the values used elsewhere in code. Updated the PR with the generated _template file.

marktnoonan: @WilcoFiers I believe everything is committed and good to go at this point.

WilcoFiers (approved):
LGTM. Thank you for the kind contribution @marktnoonan!

PR #4500, opened 2024-06-11
fix(autocomplete-valid): incomplete for invalid but safe values

### LINKED ISSUE #4492: autocomplete-valid (SC 1.3.5) triggers error for autocomplete="text"
### Product

axe Extension

### Product Version

_No response_

### Latest Version

#### Expectation
`autocomplete="text"` is not flagged as a violation of SC 1.3.5.

#### Actual
Axe Devtools flags `<input class="form-text" id="term" name="term" type="text" aria-required="false" autocomplete="text" placeholder="Zoek op titel, trefwoord" value="">` as a WCAG violation.

#### How to Reproduce
Create a page with something like `<input class="form-text" id="term" name="term" type="text" aria-required="false" autocomplete="text" value="">`  (basically a generic search field) or go to [Zoek een tool](https://www.mediawijs.be/nl/zoek-een-tool/zoeken). Run Axe Devtools and notice that it reports the following issue: [autocomplete attribute must be used correctly](https://dequeuniversity.com/rules/axe/4.9/autocomplete-valid). 

#### Additional context
While "text" is not among the [values listed in the HTML specificiation](https://html.spec.whatwg.org/#autofilling-form-controls:-the-autocomplete-attribute), the field does not collect information about the user, so SC 1.3.5 is not even applicable.

gaiety-deque:
Resolves #4492

`incomplete`'s some invalid, but safe to use, values for `autocomplete`. The common mistake in the original issue was `autocomplete="text"` and I've added a few more I'd expect to see.

Open to suggestions for more, or removing some of the ones below.

---

- text https://github.com/search?q=autocomplete%3D%22text%22&type=code
- pronouns https://github.com/search?q=autocomplete%3D%22pronouns%22&type=code
- gender https://github.com/search?q=autocomplete%3D%22gender%22&type=code
- message https://github.com/search?q=autocomplete%3D%22message%22&type=code
- content https://github.com/search?q=autocomplete%3D%22content%22&type=code

---

There are other ways to determine usage, GitHub is only one platform code is hosted for example. But it's a way to get a quick view into what's out there in production.


lib/commons/text/is-valid-autocomplete.js
```diff
@@ -59,7 +59,8 @@ export const autocomplete = {
     'email',
     'impp'
   ],
-  locations: ['billing', 'shipping']
+  locations: ['billing', 'shipping'],
+  invalidButSafe: ['text', 'pronouns', 'gender', 'message', 'content']
```
WilcoFiers, line 63: I think I'd prefer for these to come from the check option. And maybe rename this to `ignoredValues`? Safe doesn't feel quite right here. We incomplete these because we don't know if they're safe, but neither are we certain that they're problematic.


Have a look at checks/aria/aria-allowed-attr.json for an example of check options. The option then needs to be documented in doc/check-options.md.

lib/checks/forms/autocomplete-valid.json
```diff
@@ -5,7 +5,8 @@
     "impact": "serious",
     "messages": {
       "pass": "the autocomplete attribute is correctly formatted",
-      "fail": "the autocomplete attribute is incorrectly formatted"
+      "fail": "the autocomplete attribute is incorrectly formatted",
+      "incomplete": "the autocomplete attribute is incorrect but safe to use"
```
WilcoFiers, line 9: How about something like this:

```suggestion
      "incomplete": "check that this invalid autocomplete attribute does not necessary for accessibility"
```

WilcoFiers (dismissed):
Great work. Couple small suggestions for improvement.

lib/checks/forms/autocomplete-valid.json
```diff
@@ -4,8 +4,9 @@
   "metadata": {
     "impact": "serious",
     "messages": {
-      "pass": "The autocomplete attribute is correctly formatted",
-      "fail": "The autocomplete attribute is incorrectly formatted"
+      "pass": "the autocomplete attribute is correctly formatted",
+      "fail": "the autocomplete attribute is incorrectly formatted",
+      "incomplete": "the autocomplete attribute has a non-standard value. Check whether any standard value could be used instead."
```
straker, line 9: I struggled to understand what the meaning of this message was from the recent suggestion. I hope this clarifies it but we can discuss if you want this changed @WilcoFiers

WilcoFiers (approved):

PR #4501, opened 2024-06-12
perf: cache selector creation

WilcoFiers:
Closes: https://github.com/dequelabs/axe-pro-ml-service/issues/490

Selector generation is slow. We should avoid doing it more than once.


WilcoFiers:
Tested this out. I'm not seeing any perf benefit to this whatsoever.

Issue #4515, opened 2024-06-26
Should role=slider allow aria-valuetext as an alternative to aria-valuenow?


WilcoFiers:
According to WAI-ARIA, aria-valuenow is required for role=slider, even if it has aria-valuetext. From NVDA, JAWS and VO testing it looks like aria-valuetext works well as an alternative. While slider is intended as a value input mechanic, it is often used in video players. Their time value is better expressed with text than it is with a number. Having a number there at all may be more confusing than helpful.

PR #4518, opened 2024-07-01
fix(required-attr): allow aria-valuetext on slider instead of valuenow

WilcoFiers:
Closes: #4515

### DIFF
--- a/lib/checks/aria/aria-required-attr-evaluate.js
+++ b/lib/checks/aria/aria-required-attr-evaluate.js
@@ -54,6 +54,11 @@ export default function ariaRequiredAttrEvaluate(
   ) {
     return true;
   }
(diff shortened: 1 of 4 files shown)


lib/checks/aria/aria-required-attr-evaluate.js
```diff
@@ -54,6 +54,11 @@ export default function ariaRequiredAttrEvaluate(
   ) {
     return true;
   }
+  // Non-normative exception for things like media player seek slider.
+  // Tested to work in various screen readers.
+  if (role === 'slider' && virtualNode.attr('aria-valuetext')?.trim()) {
```
straker, line 59: Should this use `commons.text.sanitize` instead?

PR #4524, opened 2024-07-03
refactor: put isContext methods on axe.utils

WilcoFiers:
Had a couple scenarios where this code was getting copied over to other repos. For backward compat reasons that's going to need to stay, but at least if we have these methods on axe.utils we can prevent problems in the future if we change how context works. 

No QA needed on this one.

### DIFF
--- a/test/core/utils/is-context.js
+++ b/test/core/utils/is-context.js
@@ -124,6 +124,14 @@ describe('axe.utils isContext* methods', () => {
     it('is true for an object with both `include` and `exclude`', () => {
       assert.isTrue(isContextObject({ include: 'string', exclude: 'string' }));
     });
+
+    it('is true for an object with a valid `include` and invalid `exclude`', () => {
+      assert.isTrue(isContextObject({ include: [], exclude: 1 }));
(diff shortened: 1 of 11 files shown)


test/core/utils/is-context.js
```diff
@@ -0,0 +1,149 @@
...
+
+    it('is false if `include` is on the prototype', () => {
+      assert.isFalse(isContextObject(Object.create({ include: 'string' })));
+    });
+
+    it('is false if `exclude` is on the prototype', () => {
+      assert.isFalse(isContextObject(Object.create({ exclude: 'string' })));
+    });
+
+    it('is true for an object with both `include` and `exclude`', () => {
```
straker, line 124: Is it a bug or a feature that you can have a context object with invalid values for either `include` or `exclude` as long as one is valid? For example `{include: [], exclude: 1}`. Testing it looks like it works without issues.

It might be nice to have a test for it, but I don't think it's a blocker to the pr.

straker (approved):

straker (approved):

Issue #4523, opened 2024-07-02
Deprecated rules shouldn't run if their WCAG tag is configured


WilcoFiers:
The `aria-roledescription` rule was deprecated a while ago. In axe-core this was disabled and tagged as deprecated. That should have turned it off in all tools that use axe-core. It has done so in axe extension, but because of the way tags work in axe-core, it didn't turn the rule of the various axe SDKs (as well as anything built on top of them).

We'll want the following to be true:

```js
axe.run({ runOnly: ['wcag2a'] }) // does **not** run aria-roledescription
axe.run({ runOnly: ['wcag2a', 'deprecated' }) // Runs aria-roledescription
axe.run({ rules: { rules: { "aria-roledescription": { enabled: true } } } }) // Runs aria-roledescription
```

This should make deprecated rules work the same as experimental rules. And since it'll work the same way, we can use the same mechanism. We should be able to do this by adding "deprecated" to the [tagExclude option](https://github.com/dequelabs/axe-core/blob/develop/lib/core/base/audit.js#L194).

PR #4526, opened 2024-07-05
fix(aria-roledescription): keep disabled with { runOnly: 'wcag2a' }

WilcoFiers:
Deprecated rules are disabled by default, but because they still have WCAG / best-practice tags, using tags can unintentionally turn them back on. This PR makes it so that rules with the `deprecated` tag do not run unless they are explicitly enabled.

Closes: #4523

### DIFF
--- a/test/integration/full/configuration/tag-exclude.js
+++ b/test/integration/full/configuration/tag-exclude.js
@@ -1,51 +1,82 @@
 describe('all rules test', () => {
-  const experimentalRuleId = 'label-content-name-mismatch';
-  const deprecatedRuleId = 'aria-roledescription';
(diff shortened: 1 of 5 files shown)


test/integration/full/configuration/tag-exclude.js
```diff
@@ -0,0 +1,102 @@
+describe('all rules test', () => {
+  const experimentalRuleId = 'label-content-name-mismatch';
```
straker, line 2: This relies on the current state of axe and rules that could be made non-experimental or removed eventually. It would be better to configure axe with custom rules that are experimental and deprecated and test those ones instead.

straker (dismissed):

straker (approved):

Issue #2953, opened 2021-05-21
Should aria-prohibited-attr account for widget parents?


WilcoFiers:
The `aria-prohibited-attr` check flags use of `aria-label` on elements where it will be ignored. In building that, we did not account for scenarios like the following, where because of the way the accessible name calculation works, the `aria-label` attribute is used in the name of the parent widget.

```html
<button>
  <span aria-label="hello world"></span>
</button>
```

We need to do testing on this. Figure out if this works consistently across the supported AT, and if there are any differences between different browsers. For example, do links behave differently than buttons, because links don't have presentational children, but buttons (in some browsers) do. Does it matter if this is a native element, or an element with a role attribute. Does it matter if the element is focusable or not?

I'd like to see someone put together some test cases and try them with JAWS, NVDA & VO.

straker:
Tested with the following HTML:

```html
<h2>button</h2>
<button>
  <span aria-label="native button"></span>
</button>

<h2>link</h2>
<a href="#">
  <span aria-label="native link"></span>
</a>

<h2>role=button</h2>
<div role="button" tabindex="0">
  <span aria-label="role button"></span>
</div>

<h2>role=link</h2>
<div role="link" tabindex="0">
  <span aria-label="role link"></span>
</div>
```

* VO / Safari - Only the native button was focusable (using tab), announced as "native button, button." I wasn't able to focus the others nor navigate to them using line-by-line. Rotor menu was able to bring up both buttons in the Form Controls menu, both listed with `aria-label` as the name. Ditto for the Links menu.
* JAWS / IE11 - Can navigate to each one with either tab or line-by-line, names are announced with the `aria-label.`
* JAWS / Chrome - Same as JAWS / IE11
* NVDA / Firefox - Same as JAWS / IE11

So it looks like only VO / Safari has issues with it.

WilcoFiers:
Thanks for the testing. That sort of confirms what I was suspecting. So then there isn't really an issue in doing using aria-label on descendants of widgets. Lets add in an exception that for such cases.

straker:
This gets a bit more interesting as the following HTML also reads the name of the `aria-label` in all 3 screen readers / browsers.

```html
<button aira-labelledby="foo"></button>
<span id="foo" aria-label="foo"></span>
```

PR #4541, opened 2024-07-16
fix(aria-prohibited-attr): allow aria-label/ledby on decendants of widget

straker:
Closes: https://github.com/dequelabs/axe-core/issues/2953


lib/checks/aria/aria-prohibited-attr-evaluate.js
```diff
@@ -64,13 +66,32 @@ export default function ariaProhibitedAttrEvaluate(
...
   return ['aria-label', 'aria-labelledby'];
 }
+
+const getClosestAncestorRoleType = memoize(
+  function getClosestAncestorRoleTypeMemoized(vNode) {
+    if (!vNode) {
+      return;
+    }
+
+    const role = getRole(vNode, { noPresentational: true, chromium: true });
```
WilcoFiers, line 90: Lets add a test for `chromium: true`.

test/checks/aria/aria-prohibited-attr.js
```diff
@@ -138,4 +138,85 @@ describe('aria-prohibited-attr', function () {
...
+      assert.isFalse(checkEvaluate.apply(checkContext, params));
+    });
+
+    it('should skip "role=presentation" roles in between ancestor', () => {
+      var params = checkSetup(`
+        <button>
+          <h1 role="presentation">
+            <span id="target" aria-label="hello world"></span>
+          </h1>
+        </button>
```
WilcoFiers, line 183: Since button has presentational children, it'd be good to have an example of this with a widget that doesn't have this. I think link is the only one.

```suggestion
        <a href="#">
          <h1 role="presentation">
            <span id="target" aria-label="hello world"></span>
          </h1>
        </a>
```

WilcoFiers (changes requested):

WilcoFiers (approved):

PR #4551, opened 2024-08-01
feat(axe.d.ts): add nodeSerializer API

hamax:
This API was added in pr #4093, but TS definitions were never added.

For simplicity I'm using SerialDqElement in the API. We could introduce a generic for the custom serialized type (T extends SerialDqElement), but it's hard to consistently use it everywhere (AxeReporter, NodeSerializer.dqElmToSpec).

I also fixed DqElement.mergeSpecs which is needed to implement a custom node serializer: it exists on the constructor and not on the individual instances


axe.d.ts
```diff
@@ -405,6 +408,27 @@ declare namespace axe {
     boundingClientRect: DOMRect;
   }
 
+  interface CustomNodeSerializer {
```
straker, line 411: I think it would be more useful to have this be a type parameter and have all functions take and return that type. It would ensure that the functions take and return the same thing.

```jsx
interface CustomNodeSerializer<T = SerialDqElement> {
  toSpec: (dqElm: DqElement) => T
  mergeSpecs(nodeSpec: T, parentFrameSpec: T) => T
}
```

straker (changes requested):
Sorry for the late response and thanks for putting this together. I have a minor suggestion.

axe.d.ts
hamax: I also needed to make the nodeSerializaer.update generic to be able to take any CustomNodeSerializer (if it only takes CustomNodeSerializer<SerialDqElement> then making CustomNodeSerializer generic wouldn't really change anything.

There are still a lot of other functions that are not properly using this new type T here, like dqElmToSpec, RawNodeResult etc. Without changing the API, a lot of these functions would need to be return only generics, which is not great.

The way we use this is our code is by making our custom type extend SerialDqElement and only add optional additional properties which is not perfect but works around the type issue.

hamax:
@straker Hey, can you take another look? See inline comment above.

straker:
Apologies, didn't see the reply you gave earlier. I'm currently working on fixing our broken unit tests and then I can take another look at this.

dbjorge (approved):
Thanks for your patience and for the contribution!

dbjorge:
Approved for security and validated formatting

PR #4582, opened 2024-09-19
fix(rules): Change "alternate text" to "alternative text"

umut-gultekin:
It should be alternative text, not alternate text. 

Closes:

### DIFF
--- a/lib/rules/area-alt.json
+++ b/lib/rules/area-alt.json
@@ -19,8 +19,8 @@
   ],
   "actIds": ["c487ae"],
   "metadata": {
-    "description": "Ensure <area> elements of image maps have alternate text",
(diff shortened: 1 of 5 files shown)


straker:
Thanks for the pr. It seems we've used the wording "alternate text" in several different rules. We agree that updating the wording to "alternative text" to match the HTML and ARIA specs, but we'd like to do so consistently.

Would you mind updating all use-cases of the wording "alternate text" to "alternative text"? It looks like these rules are affected:

* area-alt
* image-alt
* input-image-alt
* object-alt
* role-img-alt

straker:
Thanks for the updates.

Approved for security.

straker (approved):

PR #4586, opened 2024-09-25
test: improve heading-matches test cases

HuanWuCode:
- In the beforeEach, it's fetching the empty-heading rule using axe.utils.getRule('empty-heading'). Ensure that this rule actually includes the matches function. If not, the first test (assert.isFunction(rule.matches)) will fail.
- some test case descriptions are misleading so it clarifies that


straker:
Thanks for the pr, but I don't think the changes improve the tests as the rule is guaranteed to exist so no need to test that it does and is a function.

PR #4588, opened 2024-09-26
fix(aria-allowed-role): add form to allowed roles of form element

matuzo:
The ARIA in HTML spec [allows a form element to have role=form](https://www.w3.org/TR/html-aria/#allowed-aria-roles-states-and-properties#el-form), but [axe does not allow it](https://github.com/dequelabs/axe-core/blob/develop/lib/standards/html-elms.js#L264). This PR adds form to the allowedRoles of form.

### DIFF
--- a/lib/standards/html-elms.js
+++ b/lib/standards/html-elms.js
@@ -261,7 +261,7 @@ const htmlElms = {
   },
   form: {
     contentTypes: ['flow'],
-    allowedRoles: ['search', 'none', 'presentation']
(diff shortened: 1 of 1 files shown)


straker:
Thanks for the pr. However this fix is different than the bug where axe-core does not require an accessible name on the `role=form` element, so I removed the `closes` part of your comment.

straker:
Approved for security

straker (approved):

Issue #3987, opened 2023-04-13
Report headers attribute referencing other <td> elements as unsupported


WilcoFiers:
The `headers` attribute on `td` elements should not be allowed to reference other `td` elements by the `td-headers-attr` rule.

The following should fail this rule:

```html
<table>
  <tr>
    <td id="hdr1">Hello</td>
    <td id="hdr2">World</td>
  </tr>
  <tr>
    <td headers="hdr1">A</td>
    <td headers="hdr2">1</td>
  </tr> 
  <tr>
    <td headers="hdr1">B</td>
    <td headers="hdr2">Wor2ld</td>
  </tr>
</table>
```

While HTML doesn't explicitly prohibit this, this does not seem to work with any screen reader running against a Chromium browser.

WilcoFiers:
The way to solve this is by adding code to test the node name in `lib/checks/tables/td-headers-attr-evaluate.js`. We'll need both unit tests for the check (`test/checks`) and integration tests (`test/integration/rules/`)

jlin95:
Hi @WilcoFiers , I'd like to take this if it's not assigned yet.

PR #4589, opened 2024-09-26
feat(td-headers-attr): report headers attribute referencing other <td> elements as unsupported 

engineerklimov:
Fix for the header attribute check. The check will report cells that references other `<td>` elements. Added unit and integrations tests as requested in corresponding issue.

Closes: [#3987](https://github.com/dequelabs/axe-core/issues/3987)


engineerklimov:
Please, let me know if this needs any changes. Thanks!

WilcoFiers:
@engineerklimov Heyya, apologies about the delays. Lots to do. I'll try to get to it within the week.

engineerklimov:
Appreciate your response, Sir
Looking forward to the review

lib/checks/tables/td-headers-attr-evaluate.js
```diff
@@ -14,9 +14,17 @@ export default function tdHeadersAttrEvaluate(node) {
...
+  // Get a list all the values of the headers attributes
+  // Find header cells inside the table
+  // Filter td elements and map to ids
+  const tableHeaders = cells
+    .filter(cell => cell.getAttribute('headers'))
+    .flatMap(cell => tokenList(cell.getAttribute('headers').trim()))
+    .map(headerId => cells.find(cell => cell.getAttribute('id') === headerId))
+    .filter(headerCell => headerCell)
+    .filter(headerCell => headerCell.nodeName.toLowerCase() !== 'td')
+    .map(headerCell => headerCell.getAttribute('id'));
```
WilcoFiers, line 27: i think there is an unnecessary amount of looping in here. It's probably better to stick more closely to what this was before and add the node name to it. I also think this is going to need to consider roles. An td with role=columnheader/rowheader should probably still pass.

lib/checks/tables/td-headers-attr-evaluate.js
```diff
@@ -40,8 +48,8 @@ export default function tdHeadersAttrEvaluate(node) {
         isSelf = headers.indexOf(cell.getAttribute('id').trim()) !== -1;
       }
 
-      // Check if the headers are of cells inside the table
-      notOfTable = headers.some(header => !ids.includes(header));
+      // Check if the headers are of headers inside the table
+      notOfTable = headers.some(header => !tableHeaders.includes(header));
```
WilcoFiers, line 52: The way this is done now axe will use the same fail message "The headers attribute is not exclusively used to refer to other cells in the table". That message needs to be updated, it should at least say "header cell". I'm not so sure that's enough though. I think it would be better if this check had a separate remediation message for this scenario. Something like "The header attribute refers a data cell, this must be a header cell so that all screen readers announce it".

To do that I think you need a separate array, something like `tdInTable`, and if it's not empty, use `this.data({ messageKey })` to set which type of message. You can look at `non-empty-alt` for an example of how message keys are used.

test/checks/tables/td-headers-attr.js
```diff
@@ -82,6 +82,15 @@ describe('td-headers-attr', function () {
...
     assert.isFalse(check.call(checkContext, node));
 
+    fixtureSetup(
+      '<table>' +
+        '  <tr> <td id="hi">hello</td> </tr>' +
+        '  <tr> <td headers="hi">goodbye</td> </tr>' +
+        '</table>'
+    );
+    node = fixture.querySelector('table');
+    assert.isFalse(check.call(checkContext, node));
```
WilcoFiers, line 92: Probably good to put this in a separate test. This one test already has more asserts than it should.

test/integration/rules/td-headers-attr/td-headers-attr.html
```diff
@@ -32,6 +32,11 @@
   <td id="self" headers="self">World</td>
 </table>
 
+<table id="fail4">
+  <td id="hdr1">Hello</td>
+  <td headers="hdr1">World</td>
+</table>
```
WilcoFiers, line 38: I think we need to consider what the role attribute does, not just the node name. We'll need to do screen reader testing to learn what works and what doesn't. Here are the scenarios I'm thinking of:

```html
<table>
  <td role="rowheader" id="hdr1">Hello</td>
  <td headers="hdr1">World</td>
</table>
```

```html
<table>
  <th role="cell" id="hdr1">Hello</td>
  <td headers="hdr1">World</td>
</table>
```

```html
<table>
  <th role="button" id="hdr1">Hello</td>
  <td headers="hdr1">World</td>
</table>
```

WilcoFiers (changes requested):
There are some scenarios the original ticket didn't consider that I feel we need to test before we can start failing this. Let me know if you have the necessary software to do this testing, otherwise we'll see what we can do on our end. At the very least we'll need to test this with NVDA + Firefox and JAWS + Chrome. Normally we'd also want VO + Safari but since it doesn't support the headers attribute anyway there's no point in that.

lib/checks/tables/td-headers-attr.json
engineerklimov, line 1: @WilcoFiers updated a messages as you requested. Can you please check if new messages are fine? I will update all locales once you confirm.

engineerklimov:
I have NVDA + Firefox and JAWS + Chrome, can you describe what tests need to be done?

lib/checks/tables/td-headers-attr.json
WilcoFiers: I think these are the scenarios that need to be tested:

```html
<h2>Scenario 1: td with role="rowheader"</h2>
<p>Check what headers are reported from "TEST":</p>
<table>
  <tr><td role="rowheader" id="hdr1">Hello</td> <td header="hdr2">ignored</td></tr>
  <tr><th id="hdr3">world</th> <td headers="hdr1 hdr2 hdr3">TEST</td></tr>
</table>

<h2>Scenario 2: th with role="cell"</h2>
<p>Check what headers are reported from "TEST":</p>
<table>
  <tr><th role="cell" id="hdr4">Hello</th> <td id="hdr5">ignored</td></tr>
  <tr><th id="hdr6">world</th> <td headers="hdr4 hdr5 hdr6">TEST</td></tr>
</table>

<h2>Scenario 3: th with role="button"</h2>
<p>Check what headers are reported from "TEST":</p>
<table>
  <tr><th role="button" id="hdr7">Hello</th> <td id="hdr8">ignored</td></tr>
  <tr><th id="hdr9">world</th> <td headers="hdr7 hdr8 hdr9">TEST</td></tr>
</table>
```

lib/checks/tables/td-headers-attr-evaluate.js
```diff
@@ -3,25 +3,32 @@ import { isVisibleToScreenReaders } from '../../commons/dom';
 
 export default function tdHeadersAttrEvaluate(node) {
```
WilcoFiers, line 4: I took a stab at refactoring this function. This was already looping over headers more than it needed to, and with the proposed change it would be twice. Let me know if you like this direction:

```js
import { tokenList } from '../../core/utils';
import { isVisibleToScreenReaders } from '../../commons/dom';
import { getRole } from '../../commons/aria';

// Order determines the priority of reporting
// Only if 0 of higher issues exists will the next be reported
const messageKeys = [
  'cell-header-not-in-table',
  'cell-header-not-th',
  'header-refs-self',
  'empty-hdrs', // incomplete
]
const [notInTable, notTh, selfRef, emptyHdrs] = messageKeys;

export default function tdHeadersAttrEvaluate(node) {
  const cells = [];
  const cellRoleById = {}
  for (let rowIndex = 0; rowIndex < node.rows.length; rowIndex++) {
    const row = node.rows[rowIndex];

    for (let cellIndex = 0; cellIndex < row.cells.length; cellIndex++) {
      const cell = row.cells[cellIndex];
      cells.push(cell);

      // Save header id to set if it's th or td with roles columnheader/rowheader
      const cellId = cell.getAttribute('id');
      if (cellId) {
        cellRoleById[cellId] = getRole(cell);
      }
    }
  }

  const badCells = { [selfRef]: [], [notInTable]: [], [notTh]: [], [emptyHdrs]: [] }
  cells.forEach(cell => {
    if (!cell.hasAttribute('headers') || !isVisibleToScreenReaders(cell)) {
      return;
    }
    const headersAttr = cell.getAttribute('headers').trim();
    if (!headersAttr) {
      badCells[emptyHdrs].push(cell);
      return;
    }

    const cellId = cell.getAttribute('id');
    // Get a list all the values of the headers attribute
    const headers = tokenList(headersAttr);
    headers.forEach(headerId => {
      if (cellId && headerId === cellId) {
        // Header references its own cell
        badCells[selfRef].push(cell);
      } else if (!cellRoleById[headerId]) {
        // Header references a cell that is not in the table
        badCells[notInTable].push(cell);
      } else if (['columnheader', 'rowheader'].includes(cellRoleById[headerId])) {
        // Header references a cell that is not a row or column header
        badCells[notTh].push(cell);
      }
    })
  });

  for (const messageKey of messageKeys) {
    if (badCells[messageKey].length > 0) {
      this.relatedNodes(badCells[messageKey]);
      if (messageKey === emptyHdrs) {
        return undefined;
      }
      this.data({ messageKey });
      return false;
    }
  }
  return true;
}
```

WilcoFiers (dismissed):

lib/checks/tables/td-headers-attr-evaluate.js
engineerklimov: Looks better than mine :sweat_smile: I will  take your impl. The only thing in mind i have is better to show all errors if there may so that user will not go back and forward and fix errors one by one, not sure if it is possible?

lib/checks/tables/td-headers-attr.json
straker: @engineerklimov I can do the tests. Here are the results of using table navigation on the tables:

## Table 1

* NVDA / Firefox
    * when starting from row 1, col 2 (ignored) and navigation down to row 2, col 2 (TEST): " Hello world row 2 TEST"
    * when starting from row 2, col 1 (world) and navigation over to row 2, col 2 (TEST): "column 2 TEST"
    * when reading the cell with headers: "Cell Hello world row 2 column 2"
* JAWS / Chrome
    * when starting from row 1, col 2 (ignored) and navigation down to row 2, col 2 (TEST): " world TEST row 2"
    * when starting from row 2, col 1 (world) and navigation over to row 2, col 2 (TEST): "Hello TEST column 2"
    * when reading the cell with headers: "column 2 row 2 Hello world TEST"

## table 2

* NVDA / Firefox
    * when starting from row 1, col 2 (ignored) and navigation down to row 2, col 2 (TEST): " world row 2 TEST"
    * when starting from row 2, col 1 (world) and navigation over to row 2, col 2 (TEST): "ignored column 2 TEST"
    * when reading the cell with headers: "Cell world row 2 column 2"
* JAWS / Chrome
    * when starting from row 1, col 2 (ignored) and navigation down to row 2, col 2 (TEST): " world TEST row 2"
    * when starting from row 2, col 1 (world) and navigation over to row 2, col 2 (TEST): "Hello ignored TEST column 2"
    * when reading the cell with headers: "column 2 row 2 Hello ignored world TEST"

## table 3

* NVDA / Firefox
    * when starting from row 1, col 2 (ignored) and navigation down to row 2, col 2 (TEST): " world row 2 TEST"
    * when starting from row 2, col 1 (world) and navigation over to row 2, col 2 (TEST): "ignored column 2 TEST"
    * when reading the cell with headers: "Cell world row 2 column 2"
* JAWS / Chrome
    * JAWS doesn't understand this table well. It thinks the column "ignored" is row 1, col 1 and so navigation down from that cell results in navigating to "world" as row 2, col 2. So it is unable to navigate to "TEST" from "ignored"
    * when starting from row 2, col 1 (world) and navigation over to row 2, col 2 (TEST): "ignored TEST column 2"
    * when reading the cell with headers: "column 2 row 2 ignored world TEST"

WilcoFiers: Interesting! That's more intuitive than I expected it would be. So then what this rule should do is report an issue when `headers` references a cell that doesn't have a role of `rowheader` or `columheader`. I.e. `th` and `tc[role=rowheader]` is allowed, but `td`, `th[role=cell]`, `th[role=button]` are not.

So then the code I suggested ought to work. It uses `getRole()` and only allows cells that are in the same table and have either `rowheader` or `columnheader` as a role, either because they're an unmodified `th` or because they have an explicit role.

WilcoFiers:
I'm going to change this PR title to "feat". This PR makes the rule stricter, we can't put that in a patch release.

lib/checks/tables/td-headers-attr-evaluate.js
WilcoFiers: Unfortunately there is no good way for a check to report multiple issues. You're right that would be better. Axe-core's architecture doesn't really allow that.

lib/checks/tables/td-headers-attr.json
engineerklimov: Just to double check i performed tests too. Here is steps i followed:
I added markup to test and saved to html file which i opened by browser. 
```
<!DOCTYPE html>
<html>
<style>
table, th, td {
  border:1px solid black;
}
</style>
<body>

<h2>Scenario 1: td with role="rowheader"</h2>
<p>Check what headers are reported from "TEST":</p>
<table>
  <tr><td role="rowheader" id="hdr1">Hello</td> <td header="hdr2">ignored</td></tr>
  <tr><th id="hdr3">world</th> <td headers="hdr1 hdr2 hdr3">TEST</td></tr>
</table>

<h2>Scenario 2: th with role="cell"</h2>
<p>Check what headers are reported from "TEST":</p>
<table>
  <tr><th role="cell" id="hdr4">Hello</th> <td id="hdr5">ignored</td></tr>
  <tr><th id="hdr6">world</th> <td headers="hdr4 hdr5 hdr6">TEST</td></tr>
</table>

<h2>Scenario 3: th with role="button"</h2>
<p>Check what headers are reported from "TEST":</p>
<table>
  <tr><th role="button" id="hdr7">Hello</th> <td id="hdr8">ignored</td></tr>
  <tr><th id="hdr9">world</th> <td headers="hdr7 hdr8 hdr9">TEST</td></tr>
</table>

</body>
</html>
```
I took text outputs from logs. 
NVDA + Firefox output:
```
heading    level 2  Scenario 1: td with role="rowheader"
Check what headers are reported from "TEST":
table  with 2 rows and 2 columns  row 1  column 1  Hello
column 2  ignored
row 2  column 1  world
column 2  TEST  out of table

heading    level 2  Scenario 2: th with role="cell"
Check what headers are reported from "TEST":
table  with 2 rows and 2 columns  row 1  column 1  Hello
column 2  ignored
row 2  column 1  world
ignored  column 2  TEST  out of table

heading    level 2  Scenario 3: th with role="button"
Check what headers are reported from "TEST":
table  with 2 rows and 2 columns  button    Hello
row 1  column 2  ignored
row 2  column 1  world
ignored  column 2  TEST
```

JAWS + chrome output:
```
Header Level 2 Scenario 1: td with role="rowheader"
The page contains 3 headings and zero links
4 visual highlights on this page.
Header Level 2 Scenario 1: td with role="rowheader"
Check what headers are reported from "TEST":
table, 2 columns and 2 rows
Hello
ignored
world
TEST
end of table
Header Level 2 Scenario 2: th with role="cell"
Check what headers are reported from "TEST":
table, 2 columns and 2 rows
Hello
ignored
world
TEST
end of table
Header Level 2 Scenario 3: th with role="button"
Check what headers are reported from "TEST":
table, 2 columns and 2 rows
Hello
Button
Scenario
JAWS Professional
JAWS Context Menu
```

engineerklimov:
@WilcoFiers I did all changes, the only thing left is locales updates. Can you check error messages? If they are fine i will update all translations.

lib/checks/tables/td-headers-attr-evaluate.js
```diff
@@ -1,63 +1,79 @@
...
 
+    const cellId = cell.getAttribute('id');
     // Get a list all the values of the headers attribute
     const headers = tokenList(headersAttr);
-
-    if (headers.length !== 0) {
-      // Check if the cell's id is in this list
-      if (cell.getAttribute('id')) {
-        isSelf = headers.indexOf(cell.getAttribute('id').trim()) !== -1;
+    headers.forEach(headerId => {
```
dbjorge, line 52: I think this is going to result in an issue where `cell` will be included multiple times in `relatedNodes` if its `headers` references multiple IDs with the same type of issue, ie

```ts
    fixtureSetup(`
      <table>
        <tr> <td id="hi1">hello</td> </tr>
        <tr> <td id="hi2">hello</td> </tr>
        <tr> <td id="bye" headers="hi1 hi2">goodbye</td> </tr>
      </table>'
    `);

    var node = fixture.querySelector('table');

    // assertion fails, _relatedNodes has length 2
    assert.deepEqual(checkContext._relatedNodes, [fixture.querySelector('#bye')]);
```

It'd be nice to improve the tests to exercise `relatedNodes` and to broaden the coverage of multiple-id cases such that they would catch this

lib/checks/tables/td-headers-attr.json
```diff
@@ -4,9 +4,13 @@
   "metadata": {
     "impact": "serious",
     "messages": {
-      "pass": "The headers attribute is exclusively used to refer to other cells in the table",
+      "pass": "The headers attribute is exclusively used to refer to other header cells in the table",
       "incomplete": "The headers attribute is empty",
-      "fail": "The headers attribute is not exclusively used to refer to other cells in the table"
+      "fail": {
+        "cell-header-not-in-table": "The headers attribute is not exclusively used to refer to other header cells in the table",
+        "cell-header-not-th": "The header attribute refers a data cell, this must be a header cell so that all screen readers announce it",
```
dbjorge, line 11: This should refer to "the headers attribute", not "the header attribute"

nit: I think it'd also be fine to be a bit more brief, maybe something like:

```suggestion
        "cell-header-not-th": "The headers attribute must refer to header cells, not data cells",
```

lib/checks/tables/td-headers-attr.json
```diff
@@ -4,9 +4,13 @@
   "metadata": {
     "impact": "serious",
     "messages": {
-      "pass": "The headers attribute is exclusively used to refer to other cells in the table",
+      "pass": "The headers attribute is exclusively used to refer to other header cells in the table",
```
dbjorge, line 7: It'd be good to make a similar update to the description and help metadata of the `td-headers-attr` rule (not just the check)

dbjorge (changes requested):
Thanks for the contribution! This is looking good overall, just noticed a few more bits to clean up

lib/checks/tables/td-headers-attr-evaluate.js
engineerklimov: I reviewed this and came to the conclusion that the node deduplication logic should be located in [check-helper.js](https://github.com/engineerklimov/axe-core/blob/develop/lib/core/utils/check-helper.js#L28). I believe it’s better to keep the check functions slim, and this approach will also help avoid code duplication.

Changing a core utility function, however, seems out of scope for this request, especially considering that this function is used in multiple places, which increases the risk of potential bugs. Additionally, if I were to add code to handle duplicated nodes here, and later someone enhances the relatedNodes function to handle it more generally, there’s a high chance that this logic would end up duplicated in both specific checks and the general utility.

In summary, I recommend handling duplicates directly in relatedNodes and making that enhancement in a separate request.

locales/ru.json
```diff
@@ -1098,7 +1098,11 @@
     "td-headers-attr": {
       "pass": "Атрибут headers используется исключительно для ссылки на другие ячейки таблицы",
       "incomplete": "Атрибут headers пуст",
-      "fail": "Атрибут headers не используется исключительно для ссылки на другие ячейки таблицы"
+      "fail": {
```
engineerklimov, line 1101: I tried to update locales. I found that all locales are quite outdated. Not sure that this change is good place to bring all locales up to date. So i updated only "ru" which i contributed recently. Not sure what is contribution policy here.

lib/rules/td-headers-attr.json
```diff
@@ -16,8 +16,8 @@
   ],
   "actIds": ["a25f45"],
   "metadata": {
-    "description": "Ensure that each cell in a table that uses the headers attribute refers only to other cells in that table",
-    "help": "Table cells that use the headers attribute must only refer to cells in the same table"
+    "description": "Ensure that each cell in a table using the headers attribute refers only to header cells within that table",
```
engineerklimov, line 19: I updated this upon @dbjorge request. This caused to changes in rule-descriptions.md should i commit it too?

dbjorge: Yes, please include it in the same PR. We do have [a workflow](https://github.com/dequelabs/axe-core/blob/develop/.github/workflows/update-generated-files.yaml) that will auto-update it as a backup for cases where it's missed, but it's easier to read the file change history if it's updated in the same PR with the rest of the context for the change.

lib/checks/tables/td-headers-attr-evaluate.js
dbjorge: I see where you're coming from, but I strongly agree that it'd be out of scope for this PR; besides being a core utility function, it's part of the published API for custom checks, so at the very least if we made such a change we'd want it to be its own changelog line.

I don't think we should allow this PR to regress the behavior in the meantime, though. I think a straightforward option would be to have `badCells` store `Set`s instead of arrays of cells and normalize it back to an array at the end with `this.relatedNodes([...badCells[messageKey]]);` (we've already established precedent of being willing to use `Set` in #3423)

engineerklimov: :dart: Set looks a good compromise, added your suggestion.

lib/rules/td-headers-attr.json
```diff
@@ -16,8 +16,8 @@
...
+    "help": "Table cells using the headers attribute must refer only to header cells or data cells with the role=columnheader/rowheader within the same table."
```
dbjorge, line 20: I appreciate the attempt to make this more concrete for folks that might not know exactly what a "header cell" formally means. But for consistency with other rules, it's actually preferable for "help" to be the shorter one rather than "description" (the fields are poorly named; "help" is really more like "suggested tooltip", so we try to keep it more concise than this). For both of them, we try to keep the overall lengths consistent between rules and fairly concise; the old help text was already towards the wordier end of what we prefer.

But I do agree that it's more understandable to use precise element/attribute names. I think we can actually get away with not mentioning the "role=" option in the text here; we only allow headers to point to a non-`<th>` element with header roles because it happens to be accessibility supported (ie, not a WCAG violation), but it's not actually compliant with the HTML standard ([which requires pointing specifically to `<th>` elements](https://html.spec.whatwg.org/multipage/tables.html#attributes-common-to-td-and-th-elements)), so I don't think we need the help text to encourage it as an option. In general, rule descriptions/help can and should omit mention of edge cases that we don't encourage but technically allow.

```suggestion
    "description": "Ensure that each cell in a table that uses the headers attribute refers only to other <th> elements in that table",
    "help": "Table cell headers attributes must refer to other <th> elements in the same table."
```

engineerklimov: Got your point. When I updated this text used following rule as example https://github.com/engineerklimov/axe-core/blob/develop/lib/rules/th-has-data-cells.json. You correctly undestood my intention behind message text, wanted to be 100% explicit about pass conditions for that check.

dbjorge (changes requested):
It looks like `fmt_check` is still showing 1 error to resolve, + it'd be good to do an `npm run build` to re-generate `/locales/_template.json` to match the most recent rule text update.

Other than those, this looks good to me!

engineerklimov:
> It looks like `fmt_check` is still showing 1 error to resolve, + it'd be good to do an `npm run build` to re-generate `/locales/_template.json` to match the most recent rule text update.
> 
> Other than those, this looks good to me!

I did updates you requested, there is problem with @WilcoFiers test case with wrong closing tag. Prettier shows error here somehow line ignore not working.

image

So i updated .prettierignore, let me know if there is better way to fix it.

test/integration/rules/td-headers-attr/td-headers-attr.html
```diff
@@ -32,6 +37,21 @@
...
+
+<table id="fail5">
+  <th role="cell" id="th-role-cell-hdr">Hello</td>
```
dbjorge, line 46: > So i updated .prettierignore, let me know if there is better way to fix it.

I'm pretty sure this was just a typo in his suggestion; this is intended to be a case exercising that the rule correctly fails if a `headers` attribute. points to a `th` element with a non-heading role, so I think it's correct for it to just be a `th` element. Shouldn't need to add it to `.prettierignore`, this is just prettier correctly catching a typo.

```suggestion
  <th role="cell" id="th-role-cell-hdr">Hello</th>
```

dbjorge (changes requested):

test/integration/rules/td-headers-attr/td-headers-attr.html
engineerklimov: Ok, fyi @WilcoFiers

engineerklimov:
I pursue your approvals guys @WilcoFiers @dbjorge @straker :slightly_smiling_face:

dbjorge:
(@WilcoFiers , once you've done the security review for my editorial commit, feel free to merge)

WilcoFiers:
For the bookx: Reviewed for security.

PR #4590, opened 2024-09-29
Accessibility: update a11y.js

HuanWuCode:
Enhance test readability by renaming tests for clarity


straker:
Thanks for the pr. These are just examples, but that same naming schema is used for other examples as well: 

* jasmine
* jsdom
* mocha

I'm not opposed to renaming the tests, but we should probably do it for all of them. If you'd like to update the pr to update all 4 tests with the same naming then I'll approve.

straker:
Closing due to inactivity

PR #4591, opened 2024-09-29
Update axe-core-tests.ts

HuanWuCode:
The 'rawEnv' value for options.reporter is being overwritten immediately by 'custom'. Remove the duplicate assignment.


straker:
Thanks for the pr. This file is a typescript test file and is there to execute code to ensure it is type compliant. So even though the value is overridden, it is there to ensure we're allowed to set the value according to type safety.

PR #4600, opened 2024-10-05
chore: avoid auto-formatter workflow failing on fork PRs

dbjorge:
It bothered me that every PR from a fork always had a red check from the autoformatter PR. This sets up the autoformatter to only run on PRs from non-forks, and adds a new `fmt:check` step to the tests workflow that will work even in forks. This lets our team still use the autoformatter without causing noise for external contributors, and also enforces our formatting requirements for PRs from forks where previously we would typically admin-merge past the check (which causes noise like #4599).

Once this PR merges, we'll want to update the repo's branch protection rules to add the new fmt-check job as a required check.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -1,10 +1,18 @@
 name: Formatter
 
-on: [pull_request]
+on:
+  pull_request:
+    branches:
+      - develop
 
 jobs:
   prettier:
+    # This conditional prevents running the job on PRs from forks; won't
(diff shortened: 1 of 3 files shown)


dbjorge:
(the broken ru localization gives a good demonstration of this working, where commit aefd015 showed a fmt_check failure and the autoformatter did its job)

stephenmathieson (approved):

WilcoFiers (approved):

Issue #4563, opened 2024-08-20
Ancestry selector of shadow root child not guaranteed unique


WilcoFiers:
Working on some shadow DOM stuff, I noticed that in order to determine whether there are multiple children in the shadow root, we're checking parentElement. For children of the shadow root this is null, which results in an ancestry that just uses the node name. You can reproduce it in the following:

```html
<script src="/axe.js"></script>
<script>
  window.onload = () => {
    axe.run({ runOnly: 'button-name', ancestry: true }, (err, results) => {
      console.log(err || results);
    });
  };
</script>
<custom-elm>
  <template shadowrootmode="open">
    <button></button>
    <button></button>
  </template>
</custom-elm>
```

This will give `['html > body > custom-elm', 'button']` as the ancestry for. BOTH buttons. This doesn't seem to happen with the regular `selector` axe-core has, which is probably why this has gone unnoticed for so long.

PR #4606, opened 2024-10-10
fix(get-ancestry): add nth-child selector for multiple siblings of shadow root

straker:
Closes: #4563

### DIFF
--- a/lib/core/utils/get-ancestry.js
+++ b/lib/core/utils/get-ancestry.js
@@ -1,6 +1,6 @@
 import getShadowSelector from './get-shadow-selector';
 
-function generateAncestry(node, options, doc) {
+function generateAncestry(node) {
   const nodeName = node.nodeName.toLowerCase();
   const parentElement = node.parentElement;
   const parentNode = node.parentNode;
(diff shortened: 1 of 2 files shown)


WilcoFiers (approved):

lib/core/utils/get-ancestry.js
```diff
@@ -2,22 +2,24 @@ import getShadowSelector from './get-shadow-selector';
 
 function generateAncestry(node) {
   const nodeName = node.nodeName.toLowerCase();
-  const parent = node.parentElement;
-  if (!parent) {
```
amakhrov, line 6: this check has now been removed.
so after this PR (axe-core 4.10.1) we started seeing 
```
TypeError: Cannot read properties of null (reading 'children')
    at generateAncestry (eval at evaluate (:234:30), <anonymous>:7617:70)
```

straker: Oh dang. I thought I had added a conditional operator to the children call, but looks like I removed it in a refactor. I'll get a fix in. Thanks for letting us know.

straker: https://github.com/dequelabs/axe-core/pull/4617

amakhrov: It happens as a part of an accessibility check in a Playwright test.

It doesn't fail every time though. perhaps it happens during some transition / animation state.

After looking closer at the test in question, looks like we omitted `await` by mistake there - so the test continues running during the acessibility check. I assume it might be causing the issue to in some cases

PR #4607, opened 2024-10-10
fix(button-name,input-button-name,input-img-alt): allow label to give accessible name

### LINKED ISSUE #3696: Unexpected error and guidance raised for button with accName from label element
### Product

axe-core

### Product Version

4.4.2

### Latest Version

### Issue Description

The following code snippet (reduced test case) was flagged as a critical impact error under the button-name rule.  However, the button _does_ have an accessible name as it is provided by its parent `<label>` element.

```
<label>
  <button role=switch aria-checked=false></button>
  Enable something	
</label>
```

Example codepen: https://codepen.io/scottohara/pen/vYjrboB
Additional use case for associating a `<label>` with a `<button>` provided in pen.

It seems that maybe the rule for checking if a button has an accessible name doesn't take into account that a button can be provided its name via a `<label>` element association. 

===

As a related note to this rule, the "to solve this problem, you need to..." section that accompanies this error ends with
>Element's default semantics were not overridden with `role="none"` or `role="presentation"`

That seems out of place or incomplete guidance with this rule, as those roles would not be respected on a `<button>` due to the presentation role conflict resolutions and these roles are ignored outright so long as the button remains focusable.  

So that bullet seems like it should either be removed, or it needs to explicitly indicate that the `<button>` needs to also have the `disabled` attribute so those roles won't be ignored on the button.

Related to the documentation piece of this issue, I've filed an issue against [ARIA in HTML](https://github.com/w3c/html-aria/issues/433) to better expose when `role=none | presentation` would be allowed on `<button>` and `<input>` elements.

### LINKED ISSUE #4472: False failure: button with role of combobox fails for missing discernible text
### Product

axe-core

### Product Version

4.9.1

### Latest Version

### Issue Description

**Note:** I was only allowed to select one product, but this was observed both in `axe-core` running in our Ember integration tests and in the extension in Chrome. I thought that a fix to `axe-core` may also make its way into the extension, but wanted to call this out to be sure.

#### Expectation
The expectation is that the inner text of a `button` with `role="combobox"` would still be considered when checking for "discernible text".

#### Actual
Currently when a `button` has `role="combobox"` axe scans fail with the error "Buttons must have discernible text". The list of potential issues states in the first bullet point "Element does not have inner text that is visible to screen readers". This is not actually true; the `button` has text contents.

#### How to Reproduce
The issue can be observed here: https://codepen.io/rgallagherab/pen/JjqKKpE

This is a non-functional example, but the markup includes all required ARIA attributes and elements for `combobox` role, and no other failures relevant to the example markup (although there are 2 issues observed in the Codepen layout elements).

#### Additional context
From what I can gather it appears that while the element should be assessed as a `combobox` role two things are true:
1. It is still being considered a `button` to some extent considering the "Buttons must have discernible text" error is presented.
2. More importantly: `axe-core` is not considering the inner text of the element as described under the "fix at least (1) of the following" list. I assume this is likely some sort of conflict in how the combination of `button` and `combobox` is being handled.

As far as I can tell `button` should be a valid element to use `role="combobox"` on. If this was not the case I'd expect to see a failure for `combobox` not being an allowed role.

straker:
I'm not sure if this should close #4472 as that also deals with buttons with a different role running `button-name`. For sure this closes #3696

Closes: #3696


lib/commons/text/native-text-methods.js
```diff
@@ -16,15 +16,14 @@ const nativeTextMethods = {
...
+  valueText: vNode => vNode.props.value || '',
 
   /**
    * Return default value of a button
    * @param {VirtualNode} element
    * @return {String} default button text
    */
-  buttonDefaultText: ({ actualNode }) =>
-    defaultButtonValues[actualNode.type] || '',
+  buttonDefaultText: vNode => defaultButtonValues[vNode.attr('type')] || '',
```
WilcoFiers, line 26: .attr is case sensitive:

```suggestion
  buttonDefaultText: vNode => defaultButtonValues[vNode.props.type] || '',
```

Let's make sure we have a test for this too.

WilcoFiers (changes requested):

dbjorge (dismissed):

WilcoFiers (approved):

PR #4609, opened 2024-10-11
fix(core): measure perf for async checks

WilcoFiers:
Performance of async checks isn't measured correctly. No tests, since we don't generally test perf timer stuff.


lib/core/base/rule.js
```diff
@@ -288,11 +288,17 @@ Rule.prototype.run = function run(context, options = {}, resolve, reject) {
   // See https://github.com/dequelabs/axe-core/pull/1172 for discussion and details.
   q.defer(res => setTimeout(res, 0));
 
-  if (options.performanceTimer) {
-    this._logRulePerformance();
-  }
-
-  q.then(() => resolve(ruleResult)).catch(error => reject(error));
+  q.then(() => {
```
dbjorge, line 291: I think it'd be preferable to put this before L289, no? If there's no other event-loop-blocking work pending then it won't make a difference, but if that `setTimeout` ends up being what causes other event-loop-blocking work to have a chance to run, I don't think it'd give us a better picture of rule performance to log independently of that other work.

dbjorge (dismissed):

lib/core/base/rule.js
WilcoFiers: Fair enough. But your solve doesn't work. If I leave it in the function where it is now then it's called synchronous. If I put it in the defer with the setTimeout then its called as soon as the async check starts to await. Either way it doesn't test the time the async check took to complete. The only way to be sure the check finished is to put the time end in the resolve. Even that's a little problematic because while we're awaiting check A, axe will run other checks, so the complete time of A will include the complete time of other checks too.

So I'm a little conflicted about how to best do this.

dbjorge: Per discussion in standup, we decided we're fine with doing a version that will only be accurate in a `{ runOnly: 'single-rule' }` scenario. It'd be nice to leave a comment explaining the limitation before merging, though.

dbjorge (approved):

PR #4543, opened 2024-07-23
fix(axe.d.ts): add typings for preload options object

blutorange:
Adds the TypeScript typings for passing an object to the `preload` setting for `axe.run`, as documented here: https://www.deque.com/axe/core-documentation/api-documentation/#preload-configuration-details

> Specifying an object

```js
axe.run(
  {
    preload: { assets: ['cssom'], timeout: 50000 }
  },
  (err, results) => {
    // ...
  }
);
 ```

PR #4615, opened 2024-10-16
chore(release): v4.10.1

### LINKED ISSUE #4472: False failure: button with role of combobox fails for missing discernible text

### LINKED ISSUE #4563: Ancestry selector of shadow root child not guaranteed unique

straker:
### Bug Fixes

- **aria-allowed-role:** add form to allowed roles of form element ([#4588](https://github.com/dequelabs/axe-core/issues/4588)) ([d462d67](https://github.com/dequelabs/axe-core/commit/d462d674bb7de0848ce2695f80b95d677c5016e0)), closes [/github.com/dequelabs/axe-core/blob/develop/lib/standards/html-elms.js#L264](https://github.com/dequelabs//github.com/dequelabs/axe-core/blob/develop/lib/standards/html-elms.js/issues/L264)
- **axe.d.ts:** add typings for preload options object ([#4543](https://github.com/dequelabs/axe-core/issues/4543)) ([72e269f](https://github.com/dequelabs/axe-core/commit/72e269f1e6d6039e70e614005f04ebfd3fe5aca5))
- **button-name,input-button-name,input-img-alt:** allow label to give accessible name ([#4607](https://github.com/dequelabs/axe-core/issues/4607)) ([364eb72](https://github.com/dequelabs/axe-core/commit/364eb72bb8f20b0ffc31be24cc96cbd349c301cb)), closes [#4472](https://github.com/dequelabs/axe-core/issues/4472) [#3696](https://github.com/dequelabs/axe-core/issues/3696) [#3696](https://github.com/dequelabs/axe-core/issues/3696)
- **get-ancestry:** add nth-child selector for multiple siblings of shadow root ([#4606](https://github.com/dequelabs/axe-core/issues/4606)) ([bdd94a2](https://github.com/dequelabs/axe-core/commit/bdd94a227a95cd5b9f8e2a1e0fd259ddd702810c)), closes [#4563](https://github.com/dequelabs/axe-core/issues/4563)
- **rules:** Change "alternate text" to "alternative text" ([#4582](https://github.com/dequelabs/axe-core/issues/4582)) ([31e0f61](https://github.com/dequelabs/axe-core/commit/31e0f61ca871b3df86468577c449a02c8ece12f0))


WilcoFiers (approved):
## Release Audit

- [x] https://github.com/dequelabs/axe-core/pull/4603
- [x] https://github.com/dequelabs/axe-core/pull/4587
- [x] https://github.com/dequelabs/axe-core/pull/4571

PR #4617, opened 2024-10-17
fix(get-ancestry): don't error when there is no parent

straker:
Seems I removed the conditional check on the children lookup during a refactor. Adding it back in and ensuring there's a test so this doesn't happen again.

### DIFF
--- a/lib/core/utils/get-ancestry.js
+++ b/lib/core/utils/get-ancestry.js
@@ -9,7 +9,7 @@ function generateAncestry(node) {
   if (
     nodeName !== 'head' &&
     nodeName !== 'body' &&
-    parentNode.children.length > 1
+    parentNode?.children.length > 1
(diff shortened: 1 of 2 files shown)


dbjorge (dismissed):

lib/core/utils/get-ancestry.js
```diff
@@ -9,7 +9,7 @@ function generateAncestry(node) {
   if (
     nodeName !== 'head' &&
     nodeName !== 'body' &&
-    parentNode.children.length > 1
+    parentNode?.children.length > 1
```
dbjorge, line 12: I didn't realize that the optional chaining operation short circuited like this; I got worried that this would need to be `parentNode?.children?.length`, but it turns out this is fine

dbjorge (approved):

PR #4621, opened 2024-10-22
docs: update link for WebAims Screen Reader User Survey

anticultist:
Updates the link to WebAims Screen Reader User Survey to point to the current survey.


straker:
Thanks for the updated link.

Issue #4626, opened 2024-10-24
aria-required-children should never fail if aria-busy=true is set


WilcoFiers:
The following shouldn't fail. This bug was introduced with a change in 4.9.0, where we updated how the remediation message of aria-busy is created.

```html
<div aria-busy="true" role="tablist">
  <div role="tabpanel">
    <a href="#" role="tab">Fizzbuzz</a>
  </div>
</div>
```

PR #4627, opened 2024-10-24
fix(aria-required-children): never fail if given aria-busy=true

WilcoFiers:
I think we made a mistake in https://github.com/dequelabs/axe-core/pull/4347/, where we allowed an element with aria-busy=true to fail if it has children it is not allowed to have. While that's a likely failure, I don't think axe can say it fails for certain.

Closes: #4626


straker (approved):

WilcoFiers:
Closing, we can reopen if we ever pick this up.

PR #4637, opened 2024-11-11
fix: consistently parse tabindex, following HTML 5 spec

### LINKED ISSUE #4632: isFocusable is too strict on tabindex
The way axe-core checks whether tabindex is valid is by calling `parseInt()`. That's different from how [HTML]()https://html.spec.whatwg.org/#rules-for-parsing-integers says to do it. Basically, axe should use this to match, and then parseInt the captured group: `/^\s*([-+]?\d+)/`

The issue is here: `lib/commons/dom/is-focusable.js`
The tests for it need to be here: `test/commons/dom/is-focusable.js`

ahmedhalac:
Ensures tabindex is correctly parsed as an integer using regex instead of parseInt to comply with HTML standards.

Closes #4632


WilcoFiers (changes requested):
This is a good start, but I think we can do a little better. Going over the code base, it looks like axe has an awful lot of places where it's parsing tabindex. I think it would be better if we put the code for this in one place. Normally we'd do this in a `commons` function, but because one of the places is in core, we'll need to do this in `core/utils`. So how about this:

In `lib/core/utils/` create a file `parse-tabindex.js` with a `parseTabindex` function that accepts either a string or null. And which returns an integer if the string is a valid tabindex value, or null if not valid or if the value is null. This would need tests in a new `test/core/utils/parse-tabindex.js`. 

This function could then be used in the following files, instead of the current ad-hoc methods:
- `lib/checks/keyboard/focusable-no-name-evaluate.js`
- `lib/checks/keyboard/no-focusable-content-evaluate.js`
- `lib/checks/keyboard/tabindex-evaluate.js`
- `lib/commons/dom/get-tabbable-elements.js`
- `lib/commons/dom/inserted-into-focus-order.js`
- `lib/commons/dom/is-focusable.js`
- `lib/commons/dom/is-in-tab-order.js`
- `lib/core/base/context/create-frame-context.js`
- `lib/rules/autocomplete-matches.js`
- `lib/rules/no-negative-tabindex-matches.js`

lib/checks/keyboard/focusable-no-name-evaluate.js
```diff
@@ -1,9 +1,11 @@
...
 import { accessibleTextVirtual } from '../../commons/text';
+import { parseTabindex } from '../../core/utils';
 
 function focusableNoNameEvaluate(node, options, virtualNode) {
-  const tabIndex = virtualNode.attr('tabindex');
-  const inFocusOrder = isFocusable(virtualNode) && tabIndex > -1;
+  const tabIndex = parseTabindex(virtualNode.attr('tabindex'));
+  const inFocusOrder =
+    isFocusable(virtualNode) && tabIndex !== null && tabIndex > -1;
   if (!inFocusOrder) {
```
WilcoFiers, line 9: This is essentially just `isInTabOrder`. Let's use that.

```suggestion
  if (!isInTabOrder(virtualNode)) {
```

lib/checks/keyboard/focusable-no-name-evaluate.js
```diff
@@ -1,9 +1,11 @@
 import { isFocusable } from '../../commons/dom';
 import { accessibleTextVirtual } from '../../commons/text';
+import { parseTabindex } from '../../core/utils';
```
WilcoFiers, line 3:
```suggestion
import { isInTabOrder } from '../../commons/dom';
import { accessibleTextVirtual } from '../../commons/text';
```

WilcoFiers (changes requested):
This is looking great! The change required a few additional updates, this is looking really good.

lib/rules/autocomplete-matches.js
```diff
@@ -34,8 +35,8 @@ function autocompleteMatches(node, virtualNode) {
   // The element has `tabindex="-1"` and has a [[semantic role]] that is
   //   not a [widget](https://www.w3.org/TR/wai-aria-1.1/#widget_roles)
   const role = virtualNode.attr('role');
-  const tabIndex = virtualNode.attr('tabindex');
-  if (tabIndex === '-1' && role) {
+  const tabIndex = parseTabindex(virtualNode.attr('tabindex'));
+  if (tabIndex < 0 && role) {
```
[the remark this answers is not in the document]
ahmedhalac: Fixed. Please review now.

WilcoFiers (approved):
This is looking great. @ahmedhalac thank you very much for the contribution!

For the books: I reviewed this PR for security.

PR #4654, opened 2024-12-12
chore: fix comma to period

eddie0329:
Other rule description ends with period, but "landmark-one-main" doesn't.(ends with comma)
To align with the context, I changed comma to period


straker (approved):
Thanks for the pr. 

Reviewed for security.

Issue #4657, opened 2024-12-16
ARIA in HTML update - allow role=math on img element


scottaohara:
ARIA in HTML will be updating to allow `role=math` on `img` elements - https://github.com/w3c/html-aria/pull/525/

This is being updated per an expressed need for this allowance in epub, and aligning with [example guidance the ARIA spec has been providing](https://www.w3.org/TR/wai-aria-1.2/#plain-html-or-polyfill-dom-result-of-the-mathml-quadratic-formula), indicating use of `role=math` on an `img` for platforms that don't support MathML.

WilcoFiers:
Thank you again Scott!

PR #4658, opened 2024-12-17
fix(aria-allowed-role): Add math to allowed roles for img element

notoriousmango:
img element can have 'math' as its role

Closes: #4657


straker (approved):
Thanks for the pr.

Reviewed for security.

PR #4668, opened 2025-01-08
chore: don't upgrade glob@11 due to node 18 support

straker:
Prevent dependabot upgrade to glob@11 as it drops support for node 18, which we still support in our tests. 

Ref: https://github.com/dequelabs/axe-core/pull/4549

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -42,6 +42,9 @@ updates:
       # @see https://github.com/dequelabs/axe-core/issues/4428
       - dependency-name: 'colorjs.io'
         versions: ['>0.4.3']
+      # Still need to support node 18
+      - dependency-name: 'glob'
+        versions: ['>11.0.0']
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -42,6 +42,9 @@ updates:
       # @see https://github.com/dequelabs/axe-core/issues/4428
       - dependency-name: 'colorjs.io'
         versions: ['>0.4.3']
+      # Still need to support node 18
+      - dependency-name: 'glob'
+        versions: ['>11.0.0']
```
dbjorge, line 47: 11.0.0 is the first version that requires node 20+ and not the last version that supports node 18, right?

```suggestion
        versions: ['>=11.0.0']
```

dbjorge (changes requested):

dbjorge (approved):

Issue #4665, opened 2025-01-07
preload-media should avoid blocking on uninteresting media elements


dbjorge:
### Product

axe-core

### Product Version

4.10.2

### Latest Version

- [x] I have tested the issue with the latest version of the product

### Issue Description

The only axe-core rule that actually uses `preload-media`, `no-autoplay-audio`, only cares about a specific subset of media elements (`autoplay` elements that are neither `paused` nor `muted`). However, axe-core's `preload-media` step blocks until *all* `src`-ful media elements have reached a nonzero `readyState`, even if those elements won't be relevant to any rule. This means that if a lazy-loaded media element exists on a page (and the page isn't in a state that will cause it to try to load), axe-core under default settings will block for 10 seconds waiting for preloading to timeout, even if the lazy-loaded media is irrelevant to any rule.

#### Expectation

We should avoid blocking axe-core on preload steps that might be elided and that no rule cares about.

#### Actual

A page with a `<video muted paused preload="none">` element will cause axe-core to take 10s+ to run under default settings, even though preloading its metadata isn't be relevant to any existing rule.

#### How to Reproduce

Minimal repro:

```html
<video preload="none">
  <source src="https://www.w3schools.com/html/mov_bbb.mp4" type="video/mp4">
</video>
```

In the motivating repro, the real video appears below the fold of the page and the page uses a scroll handler to implement a script-based lazy loading setup, and the delay is occuring within the context of `@axe-core/watcher` scanning such that it causes substantial delay to the containing test suite.

#### Additional context

There are a few options I can think of for improving this:

1. I think the least controversial improvement here would be to make preload-media ignore cases that we have no reason to expect will actually preload just by us waiting (ie, `elm.preload === 'none' && elm.networkState !== NETWORK_LOADING`)

2. We might also want a change to `no-autoplay-audio-evaluate.js` to make it more willing to deal with non-preloaded cases (ie, instead of immediately warning and incompleting, we could first say that if it has controls, return `true` regardless of whether duration is known, and if it doesn't have controls and does loop, return `false` regardless of whether duration is known) 

3. We might consider making preload-media ignore cases that will be irrelevant to no-autoplay-audio (non-`autoplay`, `paused`, `muted`). This might reduce unnecessary preload time under default settings, but would technically be a breaking change (if a custom rule cares about broader preload behavior, or if someone's test depends on the timing of the current preload behavior). We might have to undo it if we add a new rule in the future that needs the broader preload behavior.

[...]

PR #4684, opened 2025-01-28
fix(no-autoplay-audio): don't timeout for preload=none media elements

straker:
Closes: https://github.com/dequelabs/axe-core/issues/4665

### DIFF
--- a/lib/core/utils/preload-media.js
+++ b/lib/core/utils/preload-media.js
@@ -15,6 +15,27 @@ function preloadMedia({ treeRoot = axe._tree[0] }) {
     'video, audio',
     ({ actualNode }) => {
       /**
+       * Ignore media that won't load no matter how long we wait
+       * @see https://github.com/dequelabs/axe-core/issues/4665
+       */
+      if (
+        actualNode.preload === 'none' &&
+        actualNode.networkState !== actualNode.NETWORK_LOADING
+      ) {
+        return false;
+      }
+
+      /**
+       * Ignore media nodes which are `paused` or `muted`
+       */
+      if (
+        actualNode.hasAttribute('paused') ||
+        actualNode.hasAttribute('muted')
+      ) {
+        return false;
+      }
+
+      /**
        * this is to safe-gaurd against empty `src` values which can get resolved `window.location`, thus never preloading as the URL is not a media asset
        */
       if (actualNode.hasAttribute('src')) {
--- a/test/checks/media/no-autoplay-audio.js
+++ b/test/checks/media/no-autoplay-audio.js
@@ -1,119 +1,126 @@
-describe('no-autoplay-audio', function () {
-  'use strict';
+describe('no-autoplay-audio', () => {
+  const check = checks['no-autoplay-audio'];
+  const checkSetup = axe.testUtils.checkSetup;
+  const checkContext = axe.testUtils.MockCheckContext();
+  const preloadOptions = { preload: { assets: ['media'] } };
 
-  var check;
-  var fixture = document.getElementById('fixture');
-  var checkSetup = axe.testUtils.checkSetup;
-  var checkContext = axe.testUtils.MockCheckContext();
-  var preloadOptions = { preload: { assets: ['media'] } };
-
-  before(function () {
-    check = checks['no-autoplay-audio'];
-  });
-
-  afterEach(function () {
-    fixture.innerHTML = '';
-    axe._tree = undefined;
+  afterEach(() => {
     checkContext.reset();
   });
 
-  it('returns undefined when <audio> has no source (duration cannot be interpreted)', function (done) {
-    var checkArgs = checkSetup('<audio id="target"></audio>');
-    axe.utils.preload(preloadOptions).then(function () {
-      assert.isUndefined(check.evaluate.apply(checkContext, checkArgs));
-      done();
-    });
+  it('returns undefined when <audio> has no source (duration cannot be interpreted)', async () => {
+    const checkArgs = checkSetup('<audio id="target"></audio>');
+    await axe.utils.preload(preloadOptions);
+    assert.isUndefined(check.evaluate.apply(checkContext, checkArgs));
   });
 
-  it('returns undefined when <video> has no source (duration cannot be interpreted)', function (done) {
-    var checkArgs = checkSetup('<video id="target"><source src=""/></video>');
-    axe.utils.preload(preloadOptions).then(function () {
-      assert.isUndefined(check.evaluate.apply(checkContext, checkArgs));
-      done();
-    });
+  it('returns undefined when <video> has no source (duration cannot be interpreted)', async () => {
+    const checkArgs = checkSetup('<video id="target"><source src=""/></video>');
+    await axe.utils.preload(preloadOptions);
+    assert.isUndefined(check.evaluate.apply(checkContext, checkArgs));
   });
 
-  it('returns false when <audio> can autoplay and has no controls mechanism', function (done) {
-    var checkArgs = checkSetup(
+  it('returns false when <audio> can autoplay and has no controls mechanism', async () => {
+    const checkArgs = checkSetup(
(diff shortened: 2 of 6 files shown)


lib/core/utils/preload-media.js
```diff
@@ -14,6 +14,27 @@ function preloadMedia({ treeRoot = axe._tree[0] }) {
     treeRoot,
     'video, audio',
     ({ actualNode }) => {
+      /**
+       * Ignore media that won't load no matter how long we wait
+       * @see https://github.com/dequelabs/axe-core/issues/4665
+       */
+      if (
+        actualNode.preload === 'none' &&
+        actualNode.networkState !== actualNode.NETWORK_LOADING
```
zlayaAvocado, line 23: should we ignore the `preload="metadata"` too? it seems like it doesn't put the media in a ready to play state either

test/checks/media/no-autoplay-audio.js
```diff
@@ -1,119 +1,126 @@
...
-        '<source src="/test/assets/video.mp4" type="video/mp4" />' +
-        '</video>'
-    );
-    axe.utils.preload(preloadOptions).then(function () {
-      assert.isFalse(check.evaluate.apply(checkContext, checkArgs));
-      done();
-    });
+  it('returns false when <video> can autoplay and has no controls mechanism', async () => {
+    const checkArgs = checkSetup(`
+      <video id="target" autoplay="true">
```
zlayaAvocado, line 33: why are we explicitly specifying the value of the autoplay? it's it just a boolean attribute? E.g. `autoplay = "false"` is equal to `autoplay="true"` and will still autoplay.

zlayaAvocado (changes requested):
should this PR be marked as a breaking change? Since point 3 in [the ticket](https://github.com/dequelabs/axe-core/issues/4665) suggests that it should

lib/core/utils/preload-media.js
dbjorge: We don't need to wait for the media to be playable, we just need the metadata so we can identify the duration (that's why `isMediaElementReady` waits for any non-zero ready state, rather than specifically waiting for a contentful ready state)

lib/core/utils/preload-media.js
```diff
@@ -14,6 +14,27 @@ function preloadMedia({ treeRoot = axe._tree[0] }) {
...
+      ) {
+        return false;
+      }
+
+      /**
+       * Ignore media nodes which are `paused` or `muted`
+       */
+      if (
+        actualNode.hasAttribute('paused') ||
```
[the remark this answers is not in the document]
dbjorge: I'm just realizing now on a fresh reading of the spec that if we *do* restrict to only waiting for autoplay cases, we actually might want to *not* add the new logic for `preload="none"` because of this note in [the spec](https://html.spec.whatwg.org/multipage/media.html#attr-media-preload):

> Note: The [autoplay](https://html.spec.whatwg.org/multipage/media.html#attr-media-autoplay) attribute can override the [preload](https://html.spec.whatwg.org/multipage/media.html#attr-media-preload) attribute (since if the media plays, it naturally has to buffer first, regardless of the hint given by the [preload](https://html.spec.whatwg.org/multipage/media.html#attr-media-preload) attribute). Including both is not an error, however.

dbjorge (dismissed):

test/checks/media/no-autoplay-audio.js
```diff
@@ -1,119 +1,126 @@
...
   });
 
-  it('returns true when <video> can autoplay and duration is below allowed duration (by passing options)', function (done) {
-    var checkArgs = checkSetup(
-      '<video id="target" autoplay="true">' +
-        '<source src="/test/assets/video.webm" type="video/webm" />' +
-        '<source src="/test/assets/video.mp4" type="video/mp4" />' +
-        '</video>',
-      { allowedDuration: 15 }
+  it('returns false when <video> loops and has no controls mechanism when duration is unknown', () => {
```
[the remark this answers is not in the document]
straker: The reason the duration is unknown is because we don't call `await axe.utils.preload(preloadOptions);`

test/checks/media/no-autoplay-audio.js
straker: No reason other than just copy paste from previous tests.

lib/core/utils/preload-media.js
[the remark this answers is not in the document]
straker: Should remove the if block checking for preload and network state and instead add a check to this if block to check if the element is missing the `autoplay` attribute (we should return false if the element does not have it).

Additional should check in our browsers to make sure `autoplay` and `preload=none` respect the spec so we don't have the issue still.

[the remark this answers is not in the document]
straker: Testing out `autoplay` & `preload=none` on different browsers / systems. So far Chrome / Firefox / Safari on Mac and Chrome / Firefox / Edge on Windows will still autoplay with preload=none (have to specifically enable the autoplay settings though). However on Android, Chrome / Firefox will not autoplay the media and instead treat it as `preload=none` (even with the setting enabled).

As such, we should probably still flag media that is autoplayed and has `preload=none`. But since the `preload=none` may still prevent the media from loading in some situations, we'll need to figure out a way to detect if the media is loading or not that won't trigger the timeout.

test/core/utils/preload-media.js
```diff
@@ -1,73 +1,84 @@
...
-    axe.utils.preloadMedia({ treeRoot: axe._tree[0] }).then(function (result) {
-      assert.equal(result.length, 0);
-      done();
-    });
+    const result = await axe.utils.preloadMedia({ treeRoot: axe._tree[0] });
+    assert.equal(result.length, 0);
   });
 
-  it('returns empty array when <video> has no source', function (done) {
+  it('returns empty array when <video> has no source', async () => {
```
dbjorge, line 18: Should be updated to set autoplay, otherwise the test would pass regardless of the source being missing

test/core/utils/preload-media.js
```diff
@@ -1,73 +1,84 @@
...
-      done();
-    });
+    const result = await axe.utils.preloadMedia({ treeRoot: axe._tree[0] });
+    assert.equal(result.length, 0);
   });
 
-  it('returns media node (audio) after their metadata has been preloaded', function (done) {
+  it('returns empty array when media node does not preload', async () => {
+    fixtureSetup(`
+      <video id="target" preload="none">
```
dbjorge, line 26: Should be updated to set autoplay, otherwise the test would pass regardless of preload none.

(that's assuming the test browser we're using actually has the necessary autoplay behavior disabled; might need to compromise on this test if we can't work out a way to actually have the test browser skip preloading in that scenario)

test/core/utils/preload-media.js
```diff
@@ -1,73 +1,84 @@
...
+        <source src="/test/assets/video.mp4" type="video/mp4" />
+      </video>
+    `);
+    const result = await axe.utils.preloadMedia({ treeRoot: axe._tree[0] });
+    assert.equal(result.length, 0);
+  });
+
+  it('returns empty array when media node is muted', async () => {
+    fixtureSetup(`
+      <video id="target" muted>
```
dbjorge, line 36: Should be updated to set autoplay, otherwise the test would pass regardless of the muted attribute

test/core/utils/preload-media.js
```diff
@@ -1,73 +1,84 @@
...
+        <source src="/test/assets/video.mp4" type="video/mp4" />
+      </video>
+    `);
+    const result = await axe.utils.preloadMedia({ treeRoot: axe._tree[0] });
+    assert.equal(result.length, 0);
+  });
+
+  it('returns empty array when media node is paused', async () => {
+    fixtureSetup(`
+      <video id="target" paused>
```
dbjorge, line 46: Should be updated to set autoplay, otherwise the test would pass regardless of the paused attribute

dbjorge (changes requested):
Implementation looks good, just some test comments

dbjorge (approved):

Issue #4708, opened 2025-02-11
autocomplete rule should ignore readonly fields


WilcoFiers:
Axe-core's autocomplete-valid rule should not fail on readonly fields. If a field is readonly it cannot be used for user input.

```html
<input readonly autocomplete="some invalid value" />
```

PR #4721, opened 2025-03-06
fix(autocomplete-valid ): Ignore readonly autocomplete field

eddie0329:
<< Describe the changes >>

Closes: #4708

adding feature that autocomplete-valid rule should not fail on readonly fields.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -29,6 +29,6 @@ jobs:
       - run: npm run fmt
       # Prevent the prettierignore change from being committed.
       - run: git checkout .prettierignore
-      - uses: stefanzweifel/git-auto-commit-action@8621497c8c39c72f3e2a999a26b4ca1b5058a842 # tag=v5
+      - uses: stefanzweifel/git-auto-commit-action@e348103e9026cc0eee72ae06630dbe30c8bf7a79 # tag=v5
         with:
           commit_message: ':robot: Automated formatting fixes'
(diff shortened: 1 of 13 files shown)


WilcoFiers (changes requested):
Hey @eddie0329 Thank you so much for the contribution. While This addresses the issue, I don't think this is quite the right way to do this. The way you've done this will result in the rule passing. What I think we'd want here is for the rule to be inapplicable on readonly input elements. The way to do that is to move the check to the matches function in `lib/rules/autocomplete-matches.js`.

You'd then need to put a test for it in `test/rule-matches/autocomplete-matches.js`. We'll also want at least one E2E test to validate the result, which you can add to `test/integration/rules/autocomplete-valid/autocomplete-valid.html`.

WilcoFiers (dismissed):
LGTM, just need a couple integration tests in test/integration/rules/autocomplete-valid/autocomplete-valid.html.

eddie0329:
@WilcoFiers Thanks for review. 
I was trying to add some integration test,, but have little problem. 

Command , `npm run test:integration`, leads me error below:

```
SessionNotCreatedError: session not created: This version of ChromeDriver only supports Chrome version 131
```

My current chrome version is 134. Instead of down garde my original chrome version, is there any way using like virtual environment?

straker:
@eddie0329 if you run the command `npx browser-driver-manager install chromedriver` from the root of the axe-core directory, that should install a compatible version of chromedriver for you so you can run the integration tests.

straker (approved):
Thanks for the pr

PR #4725, opened 2025-03-11
feat(i18n): Add Portugal Portuguese translation

Finote2003:
Create pt_PT.json

### DIFF
--- a/locales/pt_PT.json
+++ b/locales/pt_PT.json
@@ -1,0 +1,1123 @@
+{
+  "lang": "pt_PT",
+  "rules": {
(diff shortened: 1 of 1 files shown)


straker:
@Finote2003 Before we can merge this in, we'll need you to sign the CLA.

straker:
@Finote2003 I see that you responded to the query about signing the CLA, were you able to do so? The CLA checker still says you haven't, but just wanted to make sure that was the case.

Finote2003:
> @straker 
I have sign praticaly every day, I don't now why, but every day I can sign.

Issue #3768, opened 2022-11-02
Support fallback roles


WilcoFiers:
In axe-core 4.5 we stopped failing elements that use fallback roles in the role attribute. However many places in axe-core still assume fallback roles are not allowed. For example many rule selectors do something like `[role="button"]` to select elements with an explicit button role. That wouldn't work for elements with for example `role="button img"`, which will not match.

I think we need to do things for this ticket;

1. Go through the code base, and replace all uses of `.getAttribute('role')` / `.attr('role')` with `aria.getExplicitRole()`

2. The `fallback` flag of `aria.getRole()` and its related methods needs to be set to true, and tests need to be added in relevant places that are impacted by this.

3. Replace rule selectors that use `[role="??"]` with something that uses the explicit role. There are lots of ways we could do this, so we need to figure out what's the best way to do this first. Some ideas include allowing a custom `::axe/explicitRole('button')` selector, allowing the matches property to use `axe.commons.matches()` objects, or maybe adding an `option` object for matcher methods, and add an `explicit-role-matches` method.

mjawili:
Any idea when this will be fixed? We noticed this on the [Adobe React Spectrum meter component](https://react-spectrum.adobe.com/react-spectrum/Meter.html).

WilcoFiers:
@mjawili thanks for raising that. We don't have a timeline for this at the moment. Resolving this will be part of the axe-core 5.0 release, which I'm expecting somewhere next year. That's as much as I know at the moment.

I haven't seen this result in a false positive before, but this is a very clear reason to prioritize this sooner rather than leter. I'll definitely see what we can do about this.

dannify:
Is there any update on this? It would be great to get this fix so that it doesn't continue to come up on Axe and Deque audits

PR #4756, opened 2025-04-22
chore: refactor away from looking at role attribute

straker:
This work is in preparation to support fallback roles. Any part of the code that used `.getAttribute('role')` or `.attr('role')` was migrated to use `getEplicitRole()` instead. There are a few places that still look at the role attribute directly, but those cases are validating the attributes value for different things so it's a valid use case (e.g. looking for abstract roles, invalid roles, fallback roles, unallowed roles, etc.).

Also migrates some of the older test code to use arrow functions and `const` rather than `var` (which is a tech debt strategy to slowly move updating the code as we touch the files).

Part of https://github.com/dequelabs/axe-core/issues/3768

### DIFF
--- a/lib/commons/table/is-data-cell.js
+++ b/lib/commons/table/is-data-cell.js
@@ -1,4 +1,4 @@
-import isValidRole from '../aria/is-valid-role';
+import getExplicitRole from '../aria/get-explicit-role';
 
 /**
  * Determine if a `HTMLTableCellElement` is a data cell
@@ -13,8 +13,8 @@ function isDataCell(cell) {
   if (!cell.children.length && !cell.textContent.trim()) {
     return false;
   }
-  const role = cell.getAttribute('role');
-  if (isValidRole(role)) {
+  const role = getExplicitRole(cell);
+  if (role) {
     return ['cell', 'gridcell'].includes(role);
   } else {
     return cell.nodeName.toUpperCase() === 'TD';
--- a/lib/rules/autocomplete-matches.js
+++ b/lib/rules/autocomplete-matches.js
@@ -2,6 +2,7 @@ import { sanitize } from '../commons/text';
 import standards from '../standards';
 import { isVisibleToScreenReaders, isVisibleOnScreen } from '../commons/dom';
 import { parseTabindex } from '../core/utils';
+import { getExplicitRole } from '../commons/aria';
 
 function autocompleteMatches(node, virtualNode) {
   const autocomplete = virtualNode.attr('autocomplete');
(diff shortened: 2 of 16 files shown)


lib/commons/table/is-data-cell.js
```diff
@@ -13,8 +13,8 @@ function isDataCell(cell) {
   if (!cell.children.length && !cell.textContent.trim()) {
     return false;
   }
-  const role = cell.getAttribute('role');
-  if (isValidRole(role)) {
+  const role = getExplicitRole(cell);
+  if (role) {
```
straker, line 17: `getExplicitRole` already validates the role if it returns, otherwise it returns `null` for an invalid role

lib/rules/autocomplete-matches.js
```diff
@@ -43,9 +44,10 @@ function autocompleteMatches(node, virtualNode) {
 
   // The element has `tabindex="-1"` and has a [[semantic role]] that is
   //   not a [widget](https://www.w3.org/TR/wai-aria-1.1/#widget_roles)
-  const role = virtualNode.attr('role');
+  const role = getExplicitRole(virtualNode);
+  console.log({ role });
```
WilcoFiers:
```suggestion
```

lib/rules/autocomplete-matches.js
```diff
@@ -43,9 +44,10 @@ function autocompleteMatches(node, virtualNode) {
...
   const tabIndex = parseTabindex(virtualNode.attr('tabindex'));
-  if (tabIndex < 0 && role) {
+  if (tabIndex < 0 && virtualNode.hasAttr('role')) {
     const roleDef = standards.ariaRoles[role];
     if (roleDef === undefined || roleDef.type !== 'widget') {
```
WilcoFiers, line 51: Should this use getRoleType?

lib/rules/link-in-text-block-matches.js
```diff
@@ -1,9 +1,10 @@
 import { sanitize } from '../commons/text';
 import { isVisibleOnScreen, isInTextBlock } from '../commons/dom';
+import { getExplicitRole } from '../commons/aria';
 
 function linkInTextBlockMatches(node) {
   const text = sanitize(node.innerText);
-  const role = node.getAttribute('role');
+  const role = getExplicitRole(node);
```
WilcoFiers, line 7: You may want to update the selector to `"a[href], [role^=link]"` to allow fallback roles when axe is configured to support that.

WilcoFiers (dismissed):

lib/rules/link-in-text-block-matches.js
straker: We can do that in another pr when we move towards fully supporting fallback roles

lib/rules/autocomplete-matches.js
straker: Not sure. This is using the explicit role of the element to determine the role, but `getRoleType` uses the semantic role. Which role should it be using?

WilcoFiers: I don't see how that can result in a change in behavior. It seems a little cleaner to use that. Up to you.

straker: Turns out we need the explicit role as the test wants to ignore `tabindex="-1" role="fakerole"` from an input element. If we were to use the semantic role we would get "textbox" for the role (which is a widget type) instead of empty.

PR #4758, opened 2025-04-24
docs(contributing): add info about structure, import, and philosophy of axe-core development

straker:
Just wanted to add some information about developing in axe-core.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,7 +25,7 @@ npm run eslint
 
 Axe-core uses an internal copy of the HTML page for most things internal to axe-core (called the Virtual Tree). Each HTML element on the page has an equivalent [Virtual Node](./lib/core/base/virtual-node) element, which allows us to cache or normalize information about the HTML element without mutating the actual DOM node.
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -21,6 +21,46 @@ The files in this project are formatted by [Prettier](https://prettier.io/) and
 npm run eslint
 ```
 
+### When to use HTMLElement vs Virtual Node
+
+Axe-core uses an internal copy of the HTML page for most things internal to axe-core (called the Virtual Tree). Each HTML element on the page has an equivalent [Virtual Node](./lib/core/base/virtual-node) element, which allows us to cache or normalize information about the HTML element without mutating the actual DOM node.
+
+Typically we use the Virtual Node when possible, but understand that's not always possible (such as accessing DOM only APIs). Furthermore, any function within the [utils directory](./lib/core/utils/) should not use Virtual Nodes. The reason for this is that using Virtual Nodes requires that axe first be [setup](./lib/core/public/setup.js) and create the Virtual Tree. For the most part, util functions are functions that can be run when no Virtual Tree exists.
```
scottmries, line 28: > (such as accessing DOM only APIs)

Is there an example you could link to for this? Understanding this tripped me up a little.

dbjorge (approved):
This is a good improvement, thanks!

It'd be great if a followup PR could enforce some of these conventions rather than just documenting them. In particular:

- The "no Virtual Nodes in utils" rule could probably be enforced (at least in a best-effort manner) by `no-restricted-syntax`
- The import requirements could probably be enforced by eslint rules `no-restricted-imports` and/oror `import/no-restricted-paths`

straker:
I like those ideas!

PR #4760, opened 2025-04-28
chore: add eslint rules to enact import rules

straker:
Adds eslint rules to restrict import paths from #4758, as well as virtual node use in utils.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -123,6 +123,182 @@ module.exports = [
     }
   },
   {
+    // disallow imports from node modules
+    ignores: ['lib/core/imports/**/*.js'],
+    rules: {
+      'no-restricted-imports': [
+        'error',
+        {
+          patterns: [
+            {
+              regex: '^[^.]',
+              message: 'Only core/imports files should import from node modules'
+            }
+          ]
+        }
+      ]
+    }
+  },
+  {
+    // disallow imports in standards
+    files: ['lib/standards/**/*.js'],
+    // index file can import other standards and from utils
+    ignores: ['lib/standards/index.js'],
+    rules: {
+      'no-restricted-imports': [
+        'error',
+        {
+          patterns: [
+            {
+              group: ['*'],
+              message:
+                "Standard files shouldn't use imports as they are just hard coded data objects"
+            }
+          ]
+        }
+      ]
+    }
+  },
+  {
+    // restrict imports to core/utils files to other core/utils, core, core/base, standards, or reporters/helpers
+    files: ['lib/core/utils/**/*.js'],
+    rules: {
+      'no-restricted-imports': [
+        'error',
+        {
+          patterns: [
+            {
+              // e.g. "../commons/aria/" or "../public/"
+              regex:
+                '.*\\.\\.\\/(commons|imports|public|checks|rules)\\/|.*\\.\\.\\/reporters\\/.*?\\.js',
+              message:
+                'Util files should only import from other utils, core, or standard files'
+            },
+            // disallow imports from node modules
+            // seems only 1 regex pattern is allowed to match as not having this allows node module imports even while having the general rule above for all files)
+            {
+              regex: '^[^.]',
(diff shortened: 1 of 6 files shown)


eslint.config.js
```diff
@@ -122,6 +122,182 @@ module.exports = [
...
+          patterns: [
+            {
+              regex: '^[^.]',
+              message: 'Only core/imports files should import from node modules'
+            }
+          ]
+        }
+      ]
+    }
+  },
```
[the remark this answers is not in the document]
dbjorge: https://eslint.org/docs/latest/rules/no-restricted-imports#regex

eslint.config.js
```diff
@@ -122,6 +122,182 @@ module.exports = [
...
+    // disallow imports from node modules
+    ignores: ['lib/core/imports/**/*.js'],
+    rules: {
+      'no-restricted-imports': [
+        'error',
+        {
+          patterns: [
+            {
+              regex: '^[^.]',
+              message: 'Only core/imports files should import from node modules'
```
dbjorge, line 135: It would be nice if this could summarize why we use this pattern (either directly or with a reference to an appropriate readme section/issue); future maintainers are more likely to respect the rule if it's cleary why it exists.

(I think this is more important for this rule and for the "no vNode in utils" thing (see comment below) than the other cases; the other ones are more obviously enforcing a non-circular import structure, but those two rules aren't so obvious about why they're here)

eslint.config.js
```diff
@@ -122,6 +122,182 @@ module.exports = [
...
+    files: ['lib/core/utils/**/*.js'],
+    rules: {
+      'no-restricted-imports': [
+        'error',
+        {
+          patterns: [
+            {
+              // e.g. "../commons/aria/" or "../public/"
+              regex:
+                '.*\\.\\.\\/(commons|imports|public|checks|rules)\\/|.*\\.\\.\\/reporters\\/.*?\\.js',
```
dbjorge: This should probably also forbid importing from the index files of the forbidden directories (eg, to catch a normal usage of importing `from '../imports'`)

```suggestion
                '.*\\.\\.\\/(commons|imports|public|checks|rules)(\\/|$)|.*\\.\\.\\/reporters\\/.*?\\.js',
```

(same comment for the other similar rules)

eslint.config.js
```diff
@@ -122,6 +122,182 @@ module.exports = [
...
+              regex: '^[^.]',
+              message: 'Only core/imports files should import from node modules'
+            }
+          ]
+        }
+      ]
+    }
+  },
+  {
+    // generally don't use virtual node in utils
```
dbjorge: What does "generally" mean? How will a future maintainer know whether to apply the rule in a specific case? This should either justify why not, or if there is no articulable justification and there are several exceptions already, it probably shouldn't be a rule.

straker: I think the goal was that `utils` should be functions you can use without setting up the virtual tree, as opposed to `commons` which require the virtual tree. However there is legacy code (`qsa`) and newer code (`closest`) which break that convention. I'm not sure how we want to resolve that, thus the "generally"

eslint.config.js
straker: Ya, I'm not sure if this restriction is actually needed anymore with how we build things now. I think it was more important before we moved over to es6 imports and esbuild. @WilcoFiers do you think this distinction is needed anymore, or could any file import from node modules?

eslint.config.js
dbjorge: I think just including that information in a comment here (or even better, in the error message) would be a fine resolution, it explains what the intended difference is and makes it clear what the future maintainer should do instead. "utils is meant for utility functions that work independently of axe's state; utilities that require the virtual tree to be set up should go in commons, not utils."

eslint.config.js
WilcoFiers: It's probably good we continue to put external dependencies on axe.imports so that we can use them in tests. Can't use imports in our tests yet.

PR #4769, opened 2025-05-08
test: fix debug mode failing tests

straker:
Trying to debug Firefox nightly test failures and I kept running into a problem with debug mode freezing or having tests fail. I'm still working on fixing all the issues, but here's what I've found so far:

* the `#mocha` div element would break a lot of our tests that looked at element selectors (especially `empty-heading`) since it would add non-unique elements to our tests (e.g. other `h1`s). This caused one of the tests to freeze and even crash the page as it would try to compare the result object and tried to output the entire contents of the DOM which would exceed the max string length
* the `#mocha` div element would break our grid stack, rect, and target size tests as it would add elements to the stack our tests weren't expecting
* tests that scrolled the page and didn't reset scroll position would break color contrast and grid tests when the fixture would end up in a negative position
* the size of the `#mocha` div would cause some our tests that used very large heights to move things off the page to fail when the entire test suite was run (but not individually) since the result list would end up being larger than the large height used in the test
* all tests that override `computedStyle` cause the code to freeze (not sure why yet, still investigating)
* setting `axe._memoizedFns` length to 0 meant our `teardown` code didn't reset memoized results, causing test failures across test directories (utils -> commons). Our normal test setup runs all directories in isolation

There are some other random test failures in run and run-rules that I'm also still investigating. 

My fixes include:

* moving the `#mocha` div to a closed shadow DOM in order to remove it from the root DOM and hiding it from our selector generation (this still allows us to use the HTML ui and get all the nice output in the browser)
* hiding the shadow DOM element from the grid so it's not included in element stacks and contexts
* fixing the height of the shadow DOM element so the max screen size isn't affected by the test output
* fixed the memoized length override

### DIFF
--- a/test/commons/color/get-background-color.js
+++ b/test/commons/color/get-background-color.js
@@ -1063,7 +1063,7 @@ describe('color.getBackgroundColor', function () {
 
     it('returns the html background when body does not cover the element', function () {
       fixture.innerHTML =
-        '<div id="target" style="position: absolute; top: 1000px;">elm<input></div>';
+        '<div id="target" style="position: absolute; top: 10000px;">elm<input></div>';
       document.documentElement.style.background = '#0F0';
       document.body.style.background = '#00F';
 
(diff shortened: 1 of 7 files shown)


straker:
This is no longer a problem since we moved off karma and to the new test framework.

Issue #4786, opened 2025-06-03
Upgrade Node.js to v22


stephenmathieson:
See https://github.com/dequelabs/axe-core/pull/4782#issuecomment-2931846131

PR #4787, opened 2025-06-03
build: Upgrade Node.js version to v22

stephenmathieson:
This patch upgrades the version of Node.js we use to develop and build axe-core.

Closes https://github.com/dequelabs/axe-core/issues/4786

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ defaults: &defaults
 
 unix_box: &unix_box
   docker:
-    - image: cimg/node:22-browsers
+    - image: cimg/node:22.16-browsers
 
 unix_nightly_box: &unix_nightly_box
   docker:


build/check-node-version.js
```diff
@@ -1,8 +1,13 @@
 #! /usr/bin/env node
 
-const currentVersion = process.version.replace('v', '');
+const fs = require('fs');
+const path = require('path');
+
+const PATH_TO_NVMRC = path.join(__dirname, '..', '.nvmrc');
+const nvmrc = fs.readFileSync(PATH_TO_NVMRC, 'utf8');
```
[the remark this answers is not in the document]
stephenmathieson: If someone removes the `.nvmrc` file, an error should be expected. I don't think this is necessary but am open to others feedback/thoughts.

.nvmrc
```diff
@@ -0,0 +1 @@
+22
```
straker, line 1: We had [a previous discussion about this](https://github.com/dequelabs/axe-core/pull/3580#discussion_r934793403) where having an `nvmrc` file set to a minimum version of node would force people to use this version rather than any newer version if they use an automatic script to set their node version. Is that something to consider still?

stephenmathieson: This does not _force_ a developer to use a different version. It allows developers to do `nvm use` in order to ensure they're using the expected version. If a developer working on axe-core wants to use Node.js v4, they can.

However, the [`build/check-node-version.js` file](https://github.com/dequelabs/axe-core/blob/develop/build/check-node-version.js) _does_ force the developer to use a more recent version of Node. This PR does not change that.

Additionally, we use `.nvmrc` files in nearly all of our other (maintained) Node.js-based repositories. No issues have been raised about the file internally.

TL;DR it'll [continue to Just Work™](https://github.com/dequelabs/axe-core/pull/3580#discussion_r936790403).

straker: We don't have `.nvmrc` files in a lot of our open source repos (https://github.com/dequelabs/axe-core-npm, etc.), which is where the question arrises more so than our internal repos. It's just a question about automatic scripts that use the `.nvmrc` file to switch node versions and if we are still on node 22 as a minimum but node 24 is released, an automatic script would force the user to use node 22 instead of 24. I'm just making sure that's not a concern since a user said [it would be problematic](https://github.com/dequelabs/axe-core/pull/3580#discussion_r938945769), and this would be our first major open source package to use the `.nvmrc` file strategy.

stephenmathieson: That person is wrong though. The issue with the `.nvmrc` file they tried to add is that it specified minor and patch versions, rather than only setting the major version.

If you don't believe me, please try it out.

PR #4790, opened 2025-06-05
refactor: add axe.commons.dom.getNodeGrid method

WilcoFiers:
We should avoid using that private `_grid` prop as though it was public.

### DIFF
--- a/axe.d.ts
+++ b/axe.d.ts
@@ -412,7 +412,7 @@ declare namespace axe {
 
   interface Grid {
     container: VirtualNode | null;
-    cells: GridCell[];
+    cells: unknown; // opaque implementation detail
     boundaries?: DOMRect;
     toGridIndex(num: number): number;
     getCellFromPoint(point: { x: number; y: number }): GridCell;


axe.d.ts
```diff
@@ -408,6 +408,24 @@ declare namespace axe {
     boundingClientRect: DOMRect;
   }
 
+  type GridCell = VirtualNode[];
+
+  interface Grid {
+    container: VirtualNode | null;
+    cells: GridCell[] & { _negativeIndex?: number };
```
dbjorge: `cells` is a list of rows, not a list of cells, no?

```suggestion
  type GridCell = VirtualNode[];
  type GridCellRow = GridCell[] & { _negativeIndex?: number };
  type GridCellMatrix = GridCellRow[] & { _negativeIndex?: number };

  interface Grid {
    container: VirtualNode | null;
    cells: GridCellMatrix;
```

dbjorge: Or if you prefer,

```suggestion
  type SparseMatrixRow<T> = T[] & { _negativeIndex?: number };
  type SparseMatrix<T> = SparseMatrixRow<T>[] & { _negativeIndex?: number };

  type GridCell = VirtualNode[];
  interface Grid {
    container: VirtualNode | null;
    cells: SparseMatrix<GridCell>;
```

WilcoFiers: I'm just going to ditch the private pro. Shouldn't be using that anyway. The whole reason to add `getNodeGrid` is to avoid private props.

dbjorge: I agree with not exposing `_negativeIndex`, but without it there isn't any meaningful way for a consumer to directly access the data in the `cells` property, so I'd lean towards just not exposing `cells` at all to avoid the appearance of it being safe to do so - I'd be equally fine with omitting the property from the `Grid` type or giving it a `cells: unknown // opaque implementation detail` type).

If you do prefer to expose it with a non-unknown type, I think you'd want it to be `GridCell[][]` rather than `GridCell[]` - it's a matrix of cells, not a row of cells (it's easiest to verify this by looking at the implementation of `loopGridPosition` - it's a bit confusing to read in other parts since a "cell" is a third layer of list)

dbjorge (changes requested):
[1 comment still outstanding](https://github.com/dequelabs/axe-core/pull/4790#discussion_r2140510591), otherwise lgtm

axe.d.ts
WilcoFiers: Unknown's fine

Issue #4544, opened 2024-07-25
Option to truncate excessively large `.html` in results


stephenmathieson:
### Product

axe-core

### Feature Description

When analyzing a axe Watcher results in prod, we noticed we have _a few_ excessively large datums (>1MB). This is preventing us from migrating result data to a different (_much faster_ data store).

After digging further, it was noticed that this is due to axe-core including the element in violation's `outerHTML` (eg `.violations[x].nodes[y].html`) and its related node's `outerHTML` (`.violations[x].nodes[y].relatedNodes[z].html`). Consider the following snippet:

```html
<img class="foo" src="data:png,base64, EXTREMELY LARGE IMAGE CONVERTED TO BASE64" />
```

It appears [axe-core is already truncating _some_ parts (children) of the element's HTML](https://github.com/dequelabs/axe-core/blob/v4.9.1/lib/core/utils/dq-element.js#L10-L19), but not the element itself.

An option similar to `axe.configure({ noHtml: true })` which causes axe-core to truncate *ALL* HTML snippets would be extremely helpful.

WilcoFiers:
@stephenmathieson Thanks for the suggestion. That's definitely a problem we should find a solution to. I'm not sure whether it should be a separate option. Axe-core truncates HTML snippets to prevent this problem. I think what we might want to do also put limits on attributes, both on the length of an attribute value, and of the number of attributes.

I'll bring this up with the team to come up with a solution.

WilcoFiers:
Talked about this, we agree this should happen. Will be in 4.11 / 5.0, won't do this in a patch.

We think the best way to do this is to:
1. Look at each attribute, if any are too long replace the value with "..."
2. If this results in a string that is overall still too long, start removing attributes and ad "..." at the end of the tag.

namansancheti:
Hi @WilcoFiers,
I'm new to open-source contributions and this would be my first contribution to `axe-core`, but I would like to work on this - can you please confirm no one else has already started on this ?

padmavemulapati:
This feature is released already and the testable html content file is not in QA-scope (didnot find a visible change in any axe-core supported rules)

PR #4796, opened 2025-06-10
feat(DqElement): Truncate large `html` strings when the element has a large outerHTML string

GeneratorX16:
feat(DqElement): Fixed truncation logic for large outerHTML for the the element itself

Changed truncation logic to also truncate the element of DqElement's source, while preserving the earlier truncation logic.

Closes: Issue #4544

### DIFF
--- a/lib/core/utils/dq-element.js
+++ b/lib/core/utils/dq-element.js
@@ -43,8 +43,13 @@ export function truncateElement(element) {
   let str = getOuterHtml(shallowNode);
 
   if (str.length < maxLen) {
-    const endingTagIndex = str.lastIndexOf('<');
-    str = str.substring(0, endingTagIndex);
+    let attrString = '';
+    for (const { name, value } of elementNodeMap) {
+      const attr = { name, value };
+      attrString += ` ${attr.name}="${attr.value}"`;
+    }
+
+    str = `<${nodeName}${attrString}>`;
     return str;
   }
   let strLen = `<${nodeName}>`.length;
@@ -55,8 +60,8 @@ export function truncateElement(element) {
     }
 
     const attr = { name, value };
-    let attrName = attr.name,
-      attrValue = attr.value;
+    let attrName = attr.name;
+    let attrValue = attr.value;
 
     attrName =
       attrName.length > maxAttrNameOrValueLen
--- a/lib/core/utils/dq-element.js
+++ b/lib/core/utils/dq-element.js
@@ -5,6 +5,7 @@ import getNodeFromTree from './get-node-from-tree';
 import AbstractVirtualNode from '../base/virtual-node/abstract-virtual-node';
 import cache from '../base/cache';
 import memoize from './memoize';
+import getNodeAttributes from './get-node-attributes';
 
 const CACHE_KEY = 'DqElm.RunOptions';
 
@@ -18,67 +19,60 @@ function getOuterHtml(element) {
   return source || '';
 }
 
-function truncateAttributes(element, truncatedAttributeSet) {
-  if (!element) {
-    return '';
+/**
+ * Truncates the outerHTML property of an element
+ * @param  {Node}   element the element node which needs to be truncated
+ */
+
+export function truncateElement(element) {
+  const maxLen = 300;
+  const maxAttrNameOrValueLen = 20;
+
+  const deepStr = getOuterHtml(element);
+
+  if (deepStr.length < maxLen) {
+    return deepStr;
   }
 
-  let str = getOuterHtml(element);
-  if (!truncatedAttributeSet || truncatedAttributeSet.size === 0) {
-    return str;
+  const attributeStrList = [];
+  const shallowNode = element.cloneNode(false);
+  const elementNodeMap = getNodeAttributes(shallowNode);
+
+  let str = getOuterHtml(shallowNode);
+  let strLen = `<${shallowNode.nodeName.toLowerCase()}>`.length;
+
+  for (const { name, value } of elementNodeMap) {
+    if (strLen > maxLen) {break;}
+
+    const attr = { name, value };
+    let attrName = attr.name,
+      attrValue = attr.value;
+
+    attrName =
+      attrName.length > maxAttrNameOrValueLen
+        ? attrName.substring(0, maxAttrNameOrValueLen) + '...'
+        : attrName;
+    attrValue =
+      attrValue.length > maxAttrNameOrValueLen
+        ? attrValue.substring(0, maxAttrNameOrValueLen) + '...'
+        : attrValue;
+
+    const strAttr = `${attrName}="${attrValue}"`;
+    strLen += (' ' + strAttr).length;
+    attributeStrList.push(strAttr);
   }
 
-  let intermediateAttrString = '';
-  const openingTagEndIndex = str.indexOf('>');
-
-  for (const truncatedAttr of truncatedAttributeSet) {
-    intermediateAttrString += ` ${truncatedAttr.substring(0, 20)}...`;
+  if (attributeStrList.length < elementNodeMap.length) {
+    str = `<${shallowNode.nodeName.toLowerCase()} ${attributeStrList.join(' ')} ...>`;
+  } else {
+    str = `<${shallowNode.nodeName.toLowerCase()} ${attributeStrList.join(' ')}>`;
   }
 
-  str =
-    str.substring(0, openingTagEndIndex) +
-    intermediateAttrString +
-    ' ' +
-    str.substring(openingTagEndIndex);
+  if (str.length > maxLen) {
+    str = str.substring(0, maxLen) + '... >';
+  }
+
   return str;
-}
-
-function truncateElement(element, maxLength, maxAttributeLength) {
-  maxLength = maxLength || 100;
-  maxAttributeLength = maxAttributeLength || 20;
-
-  const copyNode = element.cloneNode(true);
-  const attributesToTruncate = new Set();
-
-  const elementAttributes = element.attributes;
-  const lastIndex = elementAttributes.length - 1;
-  let currElementStr = truncateAttributes(copyNode, attributesToTruncate);
-
-  for (let i = lastIndex; i >= 0; i--) {
-    if (currElementStr.indexOf('>') < maxLength) {
-      break;
-    }
-
-    const attribute = elementAttributes[i];
-    const attributeLength = attribute.value.length + attribute.name.length;
-
-    if (attributeLength > maxAttributeLength) {
-      attributesToTruncate.add(attribute.name);
-      copyNode.attributes.removeNamedItem(attribute.name);
-      currElementStr = truncateAttributes(copyNode, attributesToTruncate);
-    }
-  }
-
-  if (currElementStr.indexOf('>') > maxLength) {
-    currElementStr = currElementStr.substring(0, maxLength) + '...>';
-  }
-
-  if (currElementStr.length > 3 * maxLength) {
-    const endIndex = currElementStr.indexOf('>');
-    currElementStr = currElementStr.substring(0, endIndex + 1);
-  }
-
-  return currElementStr;
 }
 
 function getSource(element) {
--- a/lib/core/utils/dq-element.js
+++ b/lib/core/utils/dq-element.js
@@ -9,10 +9,6 @@ import memoize from './memoize';
 const CACHE_KEY = 'DqElm.RunOptions';
 
 function getOuterHtml(element) {
-  if (!element?.outerHTML) {
-    return '';
-  }
-
   let source = element.outerHTML;
 
   if (!source && typeof window.XMLSerializer === 'function') {
@@ -22,27 +18,29 @@ function getOuterHtml(element) {
   return source || '';
 }
 
-function createTruncatedAttributeOuterHtml(element, truncatedAttributeSet) {
-  if (!element) {return;}
+function truncateAttributes(element, truncatedAttributeSet) {
+  if (!element) {
+    return '';
+  }
 
-  const copyNode = element.cloneNode(true);
-
-  let str = getOuterHtml(copyNode);
-  if (!truncatedAttributeSet || truncatedAttributeSet.size === 0) {return str;}
+  let str = getOuterHtml(element);
+  if (!truncatedAttributeSet || truncatedAttributeSet.size === 0) {
+    return str;
+  }
 
   let intermediateAttrString = '';
-  const tagNameEnd = str.indexOf('>');
+  const openingTagEndIndex = str.indexOf('>');
 
   for (const truncatedAttr of truncatedAttributeSet) {
     intermediateAttrString += `${truncatedAttr.substring(0, 20)}... `;
   }
 
   str =
-    str.substring(0, tagNameEnd) +
+    str.substring(0, openingTagEndIndex) +
     ' ' +
     intermediateAttrString +
     ' ' +
-    str.substring(tagNameEnd);
+    str.substring(openingTagEndIndex);
   return str;
 }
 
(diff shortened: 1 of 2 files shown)


test/core/utils/dq-element.js
stephenmathieson, line 1: Can you please add a test for excessively long attribute names? Something like:

```js
const el = document.createElement('div')
let attributeName = 'data-'
for (let i = 0; i < 10000000; i++) {
  attributeName += 'foo'
}
el.setAttribute(attributeName, 'bar')
```

I don't know what the expected behavior will be here, but I'm sure this is a case we'll want to handle.

GeneratorX16: Sure. For excessively long attribute names, it will truncate it to the max attribute length, which here would be 20 characters. So the result for this would be 
`<div data-foofoofoofoofoo...  ></div>`
Is this acceptable ?

stephenmathieson: That makes sense to me 👍

I'll bring this up internally and see what the rest of the team thinks.

WilcoFiers: Looks good to me. @straker can you check to make sure a change like this doesn't break anyone's deduplication code? Otherwise this may need to go behind an option.

lib/core/utils/dq-element.js
```diff
@@ -8,26 +8,85 @@ import memoize from './memoize';
...
+    return '';
+  }
+
+  let str = getOuterHtml(element);
+  if (!truncatedAttributeSet || truncatedAttributeSet.size === 0) {
+    return str;
+  }
+
+  let intermediateAttrString = '';
+  const openingTagEndIndex = str.indexOf('>');
```
straker: This makes an assumption that the first use of the `>` character is the end of the tag, but that isn't going to be true as it could be used in an attribute.

lib/core/utils/dq-element.js
```diff
@@ -8,26 +8,85 @@ import memoize from './memoize';
...
+
+  const copyNode = element.cloneNode(true);
+  const attributesToTruncate = new Set();
+
+  const elementAttributes = element.attributes;
+  const lastIndex = elementAttributes.length - 1;
+  let currElementStr = truncateAttributes(copyNode, attributesToTruncate);
+
+  for (let i = lastIndex; i >= 0; i--) {
+    if (currElementStr.indexOf('>') < maxLength) {
```
straker: `>` could be used in an attribute and not the closing of the opening tag.

lib/core/utils/dq-element.js
```diff
@@ -8,26 +8,85 @@ import memoize from './memoize';
...
+    str.substring(0, openingTagEndIndex) +
+    intermediateAttrString +
+    ' ' +
+    str.substring(openingTagEndIndex);
   return str;
 }
 
+function truncateElement(element, maxLength, maxAttributeLength) {
+  maxLength = maxLength || 100;
+  maxAttributeLength = maxAttributeLength || 20;
```
straker: This function isn't exported so only this file will have access to it. Since we don't change the max options anywhere, could remove them as arguments and instead create them as constants.

lib/core/utils/dq-element.js
```diff
@@ -8,26 +8,85 @@ import memoize from './memoize';
...
+
+  const copyNode = element.cloneNode(true);
+  const attributesToTruncate = new Set();
+
+  const elementAttributes = element.attributes;
```
straker: This should have thrown an eslint error warning of not using `element.attributes`. This can cause problems if the node has been clobbered, so should use [getNodeAttributes](https://github.com/dequelabs/axe-core/blob/develop/lib/core/utils/get-node-attributes.js) instead.

I'll have to update our eslint config to look for `element` and not just `node`.

lib/core/utils/dq-element.js
```diff
@@ -8,26 +8,85 @@ import memoize from './memoize';
...
+  const lastIndex = elementAttributes.length - 1;
+  let currElementStr = truncateAttributes(copyNode, attributesToTruncate);
```
straker: Why call `truncateAttributes` here if the `attributesToTruncate` is an empty set? Won't that just give you back the outer html of the element?

lib/core/utils/dq-element.js
```diff
@@ -8,26 +8,85 @@ import memoize from './memoize';
...
+    if (currElementStr.indexOf('>') < maxLength) {
+      break;
+    }
+
+    const attribute = elementAttributes[i];
+    const attributeLength = attribute.value.length + attribute.name.length;
+
+    if (attributeLength > maxAttributeLength) {
+      attributesToTruncate.add(attribute.name);
+      copyNode.attributes.removeNamedItem(attribute.name);
```
straker: This will be incompatible with the `getNodeAttributes` function as a clobbered element returns a copy of the attributes and not the attributes attached to the DOM node. The code will need to avoid using properties and functions off of a nodes `attributes`.

lib/core/utils/dq-element.js
```diff
@@ -8,26 +8,85 @@ import memoize from './memoize';
...
+      currElementStr = truncateAttributes(copyNode, attributesToTruncate);
```
straker: I worry that repeatedly calling `truncateAttributes`, specifically getting the elements outer HTML repeatedly, will be slow for large elements with many DOM nodes (think of doing this on `body` for a page with 100,000+ nodes when body has a long list of attributes). 

I think it would be faster if we either isolated the current element by doing `cloneNode(false)` when working on truncating the attributes, or possibly keeping a list of attributes as a string and using that to build up the element string instead.

lib/core/utils/dq-element.js
```diff
@@ -8,26 +8,85 @@ import memoize from './memoize';
...
+    }
+  }
+
+  if (currElementStr.indexOf('>') > maxLength) {
```
straker: `>` could be used in an attribute and not the closing of the opening tag.

straker (changes requested):
Thanks for the pr! I'll go confirm with other teams about their duplication strategy and make sure `source` isn't one of the properties they use. Overall the code and is good. Below I've left some comments and suggestions, mostly focused around using `element.attributes` clobbering and the use of `>` to detect the end of the opening tag.

lib/core/utils/dq-element.js
GeneratorX16: Yes. This gave me the idea to treat the innerHTML content and the node itself (using shallow node copy)  separately and then handle them and show the result by "merging" them. Is this a better approach ? 

If we do the string method, is it always the case that attribute values will be enclosed in double quotes (") for all elements ? (ex. `<div class="abcd"> ` ? 

Also, would it be better to remove all attributes that are greater than the given length in one go ? Or should I check the string after each truncation to keep it from truncating long attributes even if the string is within the given bounds (current behaviour) ?

lib/core/utils/dq-element.js
GeneratorX16: Would it be better to have a literal of void elements to check if it is a void element or a container element ? Or will calculating it be better (through logic, not through the tag name)

lib/core/utils/dq-element.js
straker: For the array method you wouldn't want to parse the outer html of the string itself, but instead use the `getNodeAttributes` and work off of that. Something like:

```js
let str = node.outerHTML(false);

// exit early if length is fine
if (str.length < maxLength) {
  return str;  // I can't remember if this includes children or not, so may need to do a cloneNode(true) here instead
}

const attributes = [];
// first pass to reduce size of any long attributes
for ({ name, value } of getNodeAttributes(node)) {
  const attr = { name, value }  // break it from being a live value which would modify the DOM value
  if (attribute.value > maxAttributeLength) {
    attr.value = '...'
  }

  attributes.push(attr);
}

str = '<' + node.nodeName.toLowerCase() + attributes.map({ name, value } => `${name}="${value}"`).join(' ') + '>'
while (str.length > maxLength) {
  attributes.pop();
  str = '<' + node.nodeName.toLowerCase() + attributes.map({ name, value } => `${name}="${value}"`).join(' ') + '...>'
}

return str;  // again, might need to make sure the return type is correct here
```

GeneratorX16: I have taken notes from this to build the latest method. It also includes for truncation of long attribute name and long tag name, along with the attribute value too.


Also, made the entire logic have a run time of linear order, so large elements would not considerably slow down the truncation logic.

lib/core/utils/dq-element.js
GeneratorX16: It is not needed now, as we are constructing the outerHTML string ourselves.

lib/core/utils/dq-element.js
GeneratorX16: It is not needed now, as we are constructing the outerHTML string ourselves.

lib/core/utils/dq-element.js
GeneratorX16: It is not needed now, as we are constructing the outerHTML string ourselves.

lib/core/utils/dq-element.js
GeneratorX16: I have replaced the attributes property with the result of getNodeAttributes.

lib/core/utils/dq-element.js
GeneratorX16: It is not needed now, as we are constructing the outerHTML string ourselves.

lib/core/utils/dq-element.js
GeneratorX16: It is not needed now, as we are constructing the outerHTML string ourselves.

straker:
Thanks for the changes. I'm about to leave for a trip so I won't be able to look at this till next week.

straker:
Thanks for the changes. I'm unable to look at this this week but will try to look at it next week.

lib/core/utils/dq-element.js
```diff
@@ -5,29 +5,82 @@ import getNodeFromTree from './get-node-from-tree';
...
+    str = `<${shallowNode.nodeName.toLowerCase()} ${attributeStrList.join(' ')} ...>`;
+  } else {
+    str = `<${shallowNode.nodeName.toLowerCase()} ${attributeStrList.join(' ')}>`;
+  }
+
+  if (str.length > maxLen) {
+    str = str.substring(0, maxLen) + '... >';
   }
 
   return str;
```
straker, line 91: This brings up an interesting question. What should happen if the element itself has too many attributes that it becomes too long, but the children length is fine? Currently something like `<button aria-label="Wednesday the 14th">Wed the 14th</button>` would be fine and show the full outerHTML for the `source`, but if we add a single attribute that makes the string >300, after truncating the string down we would end up with a length that would allow the children and end tag to fit with the 300 max length. @dbjorge @WilcoFiers  Should we try to add the children back in that case or just leave them out?

Example, should the following HTML

```html
<button aria-label="Wednesday the 14th" data-test="Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nam interdum lorem nisi, quis volutpat dolor efficitur sed. In sollicitudin nibh at nibh viverra, et scelerisque elit sodales. Quisque pellentesque laoreet turpis ac porttitor. Quisque a tristique felis. Curabitur enim neque, accumsan vel malesuada ac, maximus vel lectus. Nam auctor felis ac ullamcorper lobortis. Ut accumsan nulla et mauris varius posuere. Suspendisse placerat sapien metus, nec venenatis arcu porttitor quis. Aenean mollis venenatis velit sit amet bibendum. Nulla porta neque augue, eu condimentum orci malesuada vitae. Aliquam iaculis, felis ac fermentum mattis, tellus orci vehicula dui, malesuada fermentum enim mi sed enim. Donec tempor faucibus accumsan. Nunc id erat purus. Donec eget ipsum accumsan, euismod mauris vel, lacinia nisi.">Wed the 14th</button>
```

result in

```html
<button aria-label="Wednesday the 14th" data-test="Lorem ipsum dolor si...">
```

or

```html
<button aria-label="Wednesday the 14th" data-test="Lorem ipsum dolor si...">Wed the 14th</button>
```

I think personally I'm fine with the first option as it's the simplest.

lib/core/utils/dq-element.js
```diff
@@ -5,29 +5,82 @@ import getNodeFromTree from './get-node-from-tree';
...
+        ? attrValue.substring(0, maxAttrNameOrValueLen) + '...'
+        : attrValue;
+
+    const strAttr = `${attrName}="${attrValue}"`;
+    strLen += (' ' + strAttr).length;
+    attributeStrList.push(strAttr);
+  }
+
+  if (attributeStrList.length < elementNodeMap.length) {
+    str = `<${shallowNode.nodeName.toLowerCase()} ${attributeStrList.join(' ')} ...>`;
```
straker: If the document is in XHTML we shouldn't lowercase the node name (see [virtual node class](https://github.com/dequelabs/axe-core/blob/develop/lib/core/base/virtual-node/virtual-node.js#L58)).

lib/core/utils/dq-element.js
```diff
@@ -5,29 +5,82 @@ import getNodeFromTree from './get-node-from-tree';
...
+  } else {
+    str = `<${shallowNode.nodeName.toLowerCase()} ${attributeStrList.join(' ')}>`;
+  }
+
+  if (str.length > maxLen) {
+    str = str.substring(0, maxLen) + '... >';
```
straker: This leaves an odd edge case where if there are less `attributeStr` attributes than `elementNodeMap` attributes, you end the string with `...`. However if that final string is then longer than the `maxLen` you also add 3 more dots. So if the first string with the 3 dots is say 2 characters over the `maxLen`, you add more dots and end up with more than 3 dots at the end.

E.g.

```html
<button aria-label="Wednesday the 14th" data-test="Lorem ipsum dolor sit amet, consectetur adipiscing elit." data-test1="Lorem ipsum dolor sit amet, consectetur adipiscing elit." data-test2="Lorem ipsum dolor sit amet, consectetur adipiscing elit." data-test3="Lorem ipsum dolor sit amet, consectetur adipiscing elit." data-test4="Lorem ipsum dolor sit amet, consectetur adipiscing elit." data-test5="Lorem ipsum dolor sit amet, consectetur adipiscing elit." data-test678910="Lorem ipsum dolor sit amet, consectetur adipiscing elit." data-test7="Lorem ipsum dolor sit amet, consectetur adipiscing elit.">Wed the 14th</button>
```

```html
<button aria-label=\"Wednesday the 14th\" data-test=\"Lorem ipsum dolor si...\" data-test1=\"Lorem ipsum dolor si...\" data-test2=\"Lorem ipsum dolor si...\" data-test3=\"Lorem ipsum dolor si...\" data-test4=\"Lorem ipsum dolor si...\" data-test5=\"Lorem ipsum dolor si...\" data-test678910=\"Lorem ipsum dolor si..... >
```

lib/core/utils/dq-element.js
```diff
@@ -5,29 +5,82 @@ import getNodeFromTree from './get-node-from-tree';
...
+ * @param  {Node}   element the element node which needs to be truncated
+ */
+
+export function truncateElement(element) {
+  const maxLen = 300;
+  const maxAttrNameOrValueLen = 20;
+
+  const deepStr = getOuterHtml(element);
+
+  if (deepStr.length < maxLen) {
```
straker, line 39: This code assumes that the reason the outerHTML was long was because of the attributes, and not due to the children. If the children were longer than the max the attributes should probably be left unchanged if they fit with the 300 length. @dbjorge @WilcoFiers what do you think?

E.g.

```html
<button aria-label="Wednesday the 14th" data-test="Lorem ipsum dolor sit amet, consectetur adipiscing elit.">Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer et ante ligula. Suspendisse blandit, quam tincidunt hendrerit pulvinar, nisi lacus rutrum nibh, condimentum bibendum ipsum mi vitae neque. Morbi ac interdum lacus. Vivamus id quam in tellus dignissim convallis. Duis ultricies, eros quis aliquet cursus, lectus enim tempor tellus, eget fermentum ante augue sit amet lectus. Aliquam vehicula cursus molestie. Etiam ultricies sapien felis, vitae lacinia nisl sagittis quis.</button>
```

Should probably result in 

```html
<button aria-label="Wednesday the 14th" data-test="Lorem ipsum dolor sit amet, consectetur adipiscing elit.">
```

and not

```html
<button aria-label="Wednesday the 14th" data-test="Lorem ipsum dolor si...">
````

**Review by straker — DISMISSED** [review]
Thanks for the changes. Seeing the changes had me think about a few edge cases that we should make sure we all agree on so pining my colleagues about it.

File: lib/core/utils/dq-element.js (hunk 9 above)
Comment (WilcoFiers) [reply lib/core/utils/dq-element.js:91]: I'd lean towards leaving children off, so that your "..." is near the end of the snippet. That seems like it'd be a clearer truncation. I can go either way on that though.

File: lib/core/utils/dq-element.js (hunk 11 above)
Comment (GeneratorX16) [reply lib/core/utils/dq-element.js:72]: I will add an extra whitespace for the final truncation.
So if it is longer, it will show that the attributes are also truncated, and attributes have been removed too.
```html
<button aria-label=\"Wednesday the 14th\" data-test=\"Lorem ipsum dolor si...\" data-test1=\"Lorem ipsum dolor si...\" data-test2=\"Lorem ipsum dolor si...\" data-test3=\"Lorem ipsum dolor si...\" data-test4=\"Lorem ipsum dolor si...\" data-test5=\"Lorem ipsum dolor si...\" data-test678910=\"Lorem ipsum dolor si..... >
``` 

will be shown as:

```html
`<button aria-label=\"Wednesday the 14th\" data-test=\"Lorem ipsum dolor si...\" data-test1=\"Lorem ipsum dolor si...\" data-test2=\"Lorem ipsum dolor si...\" data-test3=\"Lorem ipsum dolor si...\" data-test4=\"Lorem ipsum dolor si...\" data-test5=\"Lorem ipsum dolor si...\" data-test678910=\"Lorem ipsum dolor si... ... >`
```

lib/core/utils/dq-element.js
GeneratorX16: I have handled this case now. If element is within bounds but the children are causing it to be larger, I do not truncate the element , but discard the children (resulting in the result expected in the comment)

GeneratorX16: Also, utilized Virtual Node in order to be compatible with Xhtml

lib/core/utils/dq-element.js
GeneratorX16: I have left the children off in this case.

straker:
Thanks for the changes. Unfortunately something in those changes broke a bunch of the tests. Double unfortunately since you forced pushed I'm unable to see the newest changes compared to prior changes, so I don't know exactly what you changed that caused the problem, nor can I revert to the point where the code was passing.

GeneratorX16:
I tried running the tests from the develop branch locally, and they were failing over there too. Can there be an issue in that ?
I will check and submit again.

straker:
Develop is passing on my end. The way our tests are set up you have to build first before you test, otherwise the changes won't be tested.

GeneratorX16:
My mistake, I was looking at the wrong place 😅. I have debugged the issue now.

straker:
This all looks good, thanks for the updates. I'm trying to finalize feedback from other teams about changing the `html` property, but once I get the all clear I should be able to merge.

straker (approved):
Sorry this took so long, took a while to get approval from all the other teams involved. They confirmed this change shouldn't break deduplication so can be merged.

sierra88-en (approved):
__

lib/core/utils/dq-element.js
sierra88-en: __

sierra88-en: H

test/core/utils/dq-element.js
sierra88-en: __

PR #4812, opened 2025-06-27
test: run list rule for act aria-required-owned failed example 10

straker:
[Nightly tests started failing](https://app.circleci.com/pipelines/github/dequelabs/axe-core/7177/workflows/03b4a72d-027e-4199-90cc-76821ff4cb9c/jobs/77330) due to a change to [ACT aria-required-owned](https://github.com/w3c/wcag-act-rules/commit/3fe6a8cb7ed8f0fdd9fcf0326a4577c77052fc5f) which added a failed example of a list with generic elements

```html
<ul>
  <div></div>
  <div></div>
</ul>
```

We fail the example using the `list` rule and not `aria-required-children`, so we can skip the test.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -77,7 +77,7 @@
         "typedarray": "^0.0.7",
         "typescript": "^5.2.2",
         "uglify-js": "^3.17.4",
-        "wcag-act-rules": "github:w3c/wcag-act-rules#dc90495a5533d326b300ee5a9487afdfc6d493c0",
+        "wcag-act-rules": "github:w3c/wcag-act-rules#5adb55d19eb2cd7fb23213b2d1acedf9004dc063",
         "weakmap-polyfill": "^2.0.4"
       },
       "engines": {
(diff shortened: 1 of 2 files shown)


test/act-rules/aria-required-owned-element-bc4a75.spec.js
```diff
@@ -1,5 +1,7 @@
 require('./act-runner.js')({
   id: 'bc4a75',
   title: 'ARIA required owned elements',
-  axeRules: ['aria-required-children']
+  axeRules: ['aria-required-children'],
+  // we fail failed example 10 with the list rule and not aria-required-children
+  skipTests: ['8b65672c9aefc4957b09a338eb85ad7dff6e53de']
```
WilcoFiers: Isn't this the fix then?

```suggestion
  axeRules: ['list', 'aria-required-children'],
```

WilcoFiers (dismissed):

test/act-rules/aria-required-owned-element-bc4a75.spec.js
straker: Oh right, forgot that we can run multiple rules at a time. My random sampling of act rule files didn't include ones that had multiple runnings. I think I'll update the readme to include info about the act object structure.

straker: https://github.com/dequelabs/axe-core/pull/4813

dbjorge (approved):

PR #4822, opened 2025-07-09
fix(color-contrast): do not run on elements with font-size: 0

### LINKED ISSUE #4820: False positive color-contrast: `font-size: 0` elements should be ignored
This came up recently where an element with `font-size: 0` is throwing a color-contrast violation when the text is not visible. Normally an element with `font-size: 0` would have a width and height of 0 (which should be ignored), but explicitly setting the width and height of the element overrides that while still visually hidding the font.

```html
<ol>
  <li>Visible text</li>
  <li style="font-size: 0; width: 16px; height: 16px; background: red; color: #282828;">Not visible</li>
</ol>
```

straker:
Closes #4820

### DIFF
--- a/lib/rules/color-contrast-matches.js
+++ b/lib/rules/color-contrast-matches.js
@@ -137,7 +137,7 @@ function colorContrastMatches(node, virtualNode) {
     return false;
   }
 
-  if (virtualNode.getComputedStylePropertyValue('font-size') === '0px') {
+  if (!parseFloat(virtualNode.getComputedStylePropertyValue('font-size'))) {
     return false;
   }
 
--- a/test/integration/rules/color-contrast/color-contrast.html
+++ b/test/integration/rules/color-contrast/color-contrast.html
@@ -465,3 +465,10 @@
     <div id="ignore13" style="padding-left: 65px">Hello World</div>
   </div>
 </div>
+
+<div
+  id="ignore13"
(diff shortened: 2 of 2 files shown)


lib/rules/color-contrast-matches.js
```diff
@@ -137,6 +137,10 @@ function colorContrastMatches(node, virtualNode) {
     return false;
   }
 
+  if (virtualNode.getComputedStylePropertyValue('font-size') === '0px') {
```
nickbrit: I’d like to suggest a small update for the font‑size check:

```
// Skip elements with zero or invalid font size
if (!parseFloat(virtualNode.getComputedStylePropertyValue('font-size'))) {
    return false;
}
```

Hope this helps!

WilcoFiers: I agree, that seems safer.

test/rule-matches/color-contrast-matches.js
```diff
@@ -431,6 +431,13 @@ describe('color-contrast-matches', function () {
     assert.isFalse(rule.matches(target, axe.utils.getNodeFromTree(target)));
   });
 
+  it('should not match text with font-size: 0', () => {
```
WilcoFiers, line 434: Let's add an integration test too.

WilcoFiers (dismissed):

WilcoFiers (approved):

PR #4834, opened 2025-07-22
fix(performanceTimer): work in frames

WilcoFiers:
Turns out our performanceTimer errors when running in iframes. I also added some detail to gather so we could see what rule is being gathered.

### DIFF
--- a/lib/core/utils/performance-timer.js
+++ b/lib/core/utils/performance-timer.js
@@ -90,11 +90,11 @@ const performanceTimer = (() => {
       ) {
         // only output measures that were started after axe started, otherwise
         // we get measures made by the page before axe ran (which is confusing)
-        const axeStart =
-          window.performance.getEntriesByName('mark_axe_start')[0];
+        const auditStart =
+          window.performance.getEntriesByName('mark_audit_start')[0];
         const measures = window.performance
           .getEntriesByType('measure')
-          .filter(measure => measure.startTime >= axeStart.startTime);
+          .filter(measure => measure.startTime >= auditStart.startTime);
         for (let i = 0; i < measures.length; ++i) {
           const req = measures[i];
           if (req.name === measureName) {
--- a/lib/core/base/rule.js
+++ b/lib/core/base/rule.js
@@ -380,7 +380,9 @@ Rule.prototype._trackPerformance = function _trackPerformance() {
  * @param {Array} nodes Result of rule.gather
  */
 Rule.prototype._logGatherPerformance = function _logGatherPerformance(nodes) {
-  log('gather (', nodes.length, '):', performanceTimer.timeElapsed() + 'ms');
+  log(
+    `gather for ${this.id} (${nodes.length} nodes): ${performanceTimer.timeElapsed()}ms`
+  );
   performanceTimer.mark(this._markChecksStart);
 };
 


lib/core/utils/performance-timer.js
```diff
@@ -90,11 +90,12 @@ const performanceTimer = (() => {
       ) {
         // only output measures that were started after axe started, otherwise
         // we get measures made by the page before axe ran (which is confusing)
-        const axeStart =
-          window.performance.getEntriesByName('mark_axe_start')[0];
+        const { startTime } =
+          window.performance.getEntriesByName('mark_axe_start')[0] ||
+          window.performance.getEntriesByName('mark_audit_start')[0];
```
WilcoFiers: This error is kinda making me wonder if we should rethink not having tests for this. WDYT?

dbjorge: I agree; I think tests would be a good idea (especially since `performanceTimer` is a documented part of our API).

lib/core/utils/performance-timer.js
```diff
@@ -90,11 +90,12 @@ const performanceTimer = (() => {
       ) {
         // only output measures that were started after axe started, otherwise
         // we get measures made by the page before axe ran (which is confusing)
-        const axeStart =
-          window.performance.getEntriesByName('mark_axe_start')[0];
+        const { startTime } =
```
dbjorge: nit: A few lines up, this call is defensive about the possibility of `window.performance` not existing at all; should it also be a bit more defensive about the possibility of the `mark_*_start` events not being available (eg, `assert` with a meaningful message that `start` or `auditStart` must be called first)?

lib/core/utils/performance-timer.js
```diff
@@ -90,11 +90,12 @@ const performanceTimer = (() => {
       ) {
```
dbjorge, line 115: nit: The actual functionality here feels unclear from the name/comment. For L81:

    /**
     * @member {Function} logMeasures Logs previously captured measures in chronological order.
     * Starts from the most recent start()/auditStart() call.
     * @param {String|undefined} measureName If provided, will only log up to the first matching measure.
     */

lib/core/utils/performance-timer.js
```diff
@@ -90,11 +90,12 @@ const performanceTimer = (() => {
...
         // only output measures that were started after axe started, otherwise
         // we get measures made by the page before axe ran (which is confusing)
-        const axeStart =
-          window.performance.getEntriesByName('mark_axe_start')[0];
+        const { startTime } =
+          window.performance.getEntriesByName('mark_axe_start')[0] ||
```
dbjorge: I don't think this actually does the right thing in the event that you're in an environment with more than one mark, since we never seem to clear marks/measures out from previous runs. I think what you'd actually want here is "grab the *most recent* (not 0th) of each type of mark, and if both exist, use the one with later `start_time`)".

dbjorge (dismissed):

test/core/utils/performance-timer.js
```diff
@@ -0,0 +1,148 @@
...
+      assert.match(messages[0], /Measure foo took [0-9.]+ms/);
+      assert.match(messages[1], /Measure bar took [0-9.]+ms/);
+    });
+  });
+
+  describe('timeElapsed', () => {
+    it('returns the time elapsed since axe started', async () => {
+      performanceTimer.start();
+      await sleep(100);
+      assert.closeTo(performanceTimer.timeElapsed(), 100, 10);
```
dbjorge, line 131: I was a bit worried when first reading this about how reliable a 10ms epsilon would be, but I verified in a few circleci runs + 20 trials locally that it does seem like it passes reliably. :+1:

lib/core/utils/performance-timer.js
```diff
@@ -65,61 +81,98 @@ const performanceTimer = (() => {
...
+      return currentTime - axeStartTime;
     },
+
+    /**
+     * @member {Function} clearMark Clears a mark
+     * @param {String} markName String name of mark to clear
+     */
+    clearMark(...markNames) {
+      for (const markName of markNames) {
+        window.performance.clearMarks(markName);
```
dbjorge, line 161: This should probably be equally as defensive about `window.performance.clearMarks` not existing as `mark`/`measure`/`logMeasures` are (who knows what existing jsdom environment breaks if we assume it's available...)

dbjorge (approved):

PR #4839, opened 2025-07-25
Docs: Update CHANGELOG.md

coliff:
HTML 5 to HTML5 (no space)

REF: https://en.wikipedia.org/wiki/HTML5


straker:
I'm not entirely sure. I have the belief that the release process just adds onto the README and doesn't try to correct anything from before it, but I can't say for certain that is the case. Either way we can merge this in and see what happens.

straker:
@coliff We'll just need you to sign the CLA before we can merge.

PR #4844, opened 2025-07-30
fix(idref): fallback to qsa

straker:


lib/commons/dom/idrefs.js
```diff
@@ -18,24 +23,29 @@ import { tokenList } from '../../core/utils';
...
-    const result = [];
-    let attrValue = node.getAttribute(attr);
+  if (attrValue) {
+    for (const token of tokenList(attrValue)) {
+      try {
+        doc ??= getRootNode(domNode);
+        result.push(doc.getElementById(token));
+      } catch {
+        // don't run QSA on detached nodes or partial trees
+        const root = closest(vNode, 'html');
```
dbjorge: I'm not so sure this special case belongs here specifically; there are lots of similar cases where `getRootNode` is being used similarly. I wonder if it'd be better to implement a vNode-based `getRootNode` (probably one that uses the existing `getRootNode` for a vNode with an actualNode and which otherwise walks up parents looking for a null/undefined/#document/#document-fragment case and taking the child of that, rather than looking for `html` specifically), and then doing a more broad audit of places where the non-vNode-based `getRootNode` is currently used to see where else it would make sense to apply something similar (even if we want those to be separate PRs)

straker: I tested the performance of doing something like this. I was worried that `getRootNode` and `getElementById` would be way more performant that us looking up the parent tree for root manually and then calling qsa on it. However with a page of 2000 nested nodes deep (was all I could create before crashing the page each time), doing the manual lookup and qsa was only 5ms while a `getRootNode` and `getElementById` was 0ms (basically instant). So while there is a difference, it's not something I would except to be problematic on an actual page (which when testing was only 1.5ms).

```js
let parent = fragment = document.createDocumentFragment();
for (let i = 0; i < 2000; i++) {
  const div = document.createElement('div');
  div.textContent = i
  div.id = 't' + i
  parent.append(div);
  parent = div;
}

document.body.append(fragment);

axe.setup();
const node = t1999 // the deepest nested node

window.performance.clearMarks();
window.performance.clearMeasures();

window.performance.mark('start');
const root = axe.utils.closest(axe.utils.getNodeFromTree(node), 'html')
const vNode = axe.utils.querySelectorAllFilter(root, '#t988', root.shadowId)[0]
window.performance.mark('end');

window.performance.measure('time', 'start', 'end')
console.log(window.performance.getEntriesByName('time')[0].duration)
```

As a side note, doing an id lookup on the root vNode is an O(1) lookup since we cache the `_selectorMap` on the root node, so the 5ms is entirely parent lookup calls and matching it against the selector. We could probably save some time if we didn't need to match against a selector when finding the root node.

lib/commons/text/accessible-text.js
```diff
@@ -12,8 +12,8 @@ import { getNodeFromTree } from '../../core/utils';
  * @return {string}
  */
 function accessibleText(element, context) {
-  const virtualNode = getNodeFromTree(element); // throws an exception on purpose if axe._tree not correct
```
straker, line 15: This comment is not true as `getNodeFromTree` does not throw, just returns `null` if the node isn't found.

lib/commons/dom/idrefs.js
```diff
@@ -18,24 +23,41 @@ import { tokenList } from '../../core/utils';
...
-      }
+    const root = getRootNode(domNode);
+    for (const token of tokenList(attrValue)) {
+      results.push(root.getElementById(token));
     }
-
-    return result;
   } catch {
-    throw new TypeError('Cannot resolve id references for non-DOM nodes');
+    const rootVNodes = getRootVNodes(vNode);
```
dbjorge: I think this is closer than previous attempts, but I think it's going to do the wrong thing with slotted elements; those appear in the virtual tree under the shadowId of the thing they're slotted into, but for idrefs purposes their id counts as being scoped to the root in which they're defined, not the one they're slotted into:

```html
  <!-- Behavior is consistent in current chrome/firefox/safari -->
  <label for="slotted-input-a">out-of-shadow label</label>
  <my-custom-elm>
    <template shadowrootmode="open">
      <label for="slotted-input-a">in-shadow label competing with out-of-shadow label</label>
      <label for="slotted-input-b">in-shadow-only label</label>
      <slot></slot>
    </template>
    <!-- Accessible name is "out-of-shadow label" -->
    <input id="slotted-input-a"></input>
    <!-- Accessible name is empty -->
    <input id="slotted-input-b"></input>
  </my-custom-elm>
```

I'm not sure there's enough information in the virtual tree right now to distinguish that case correctly; it might be necessary for us to start tracking information about the "source shadow ID" or something for slotted elements.

lib/commons/dom/get-root-vnodes.js
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Return the vNode(s)
+ * @method getRootVNodes
+ * @memberof axe.commons.dom
+ * @instance
+ * @param {Element|VirtualNode} node
+ * @returns {VirtualNode[]}
+ */
+export default function getRootVNodes(node) {
```
dbjorge: Naming-wise, I'd suggest something like `getRootChildren` instead - getRootVNodes makes me think it'd be returning a virtual document fragment node, since that's the "root" as the DOM uses that term, but like you commented down on L28, we don't track vnodes for those document fragments :disappointed:

lib/commons/dom/get-root-vnodes.js
```diff
@@ -0,0 +1,38 @@
...
+
+  // since the virtual tree does not have a #shadowRoot element the root virtual
+  // node is the shadow host element. however the shadow host element is not inside
+  // the shadow DOM tree so we return the children of the shadow host element in
+  // order to not cross shadow DOM boundaries
+  if (vNode.shadowId !== vNode.parent.shadowId) {
+    return [...vNode.parent.children];
+  }
+
+  vNode._rootNodes = getRootVNodes(vNode.parent);
```
dbjorge: Rather than needing to create new properties on every vnode, would it be lower overhead cache a map from shadowId to root children?

lib/commons/dom/idrefs.js
```diff
@@ -18,24 +23,41 @@ import { tokenList } from '../../core/utils';
...
+    if (!rootVNodes) {
+      throw new TypeError('Cannot resolve id references for non-DOM nodes');
+    }
+
+    for (const token of tokenList(attrValue)) {
+      let result = null;
+
+      for (const root of rootVNodes) {
+        const foundNode = querySelectorAll(root, `#${token}`)[0];
```
dbjorge, line 49: For shadow root cases, this is going to essentially bypass QSA's selector-cache behavior that would otherwise make the ID lookup fast; we might want to consider making it smarter about that (`findMatchingNodes` *looks* like it's meant to understand how to do it, but `getNodesMatchingExpression` is going to early exit from the cached path anytime `root` is not the document root)

dbjorge (changes requested):

lib/commons/dom/idrefs.js
dbjorge: When thinking what to track, consider also this test case:

```html
  <label for="slotted-input">out-of-shadow label</label>
  <level-a-custom-elm>
    <template shadowrootmode="open">
      <label for="slotted-input">shadow level A label</label>
      <level-b-custom-elm>
        <template shadowrootmode="open">
          <label for="slotted-input">shadow level B label</label>
          <slot></slot>
        </template>
        <slot></slot>
      </level-b-custom-elm>
    </template>
    <!-- Accessible name is "out-of-shadow label" -->
    <input id="slotted-input"></input>
  </level-a-custom-elm>
```

lib/commons/dom/get-root-vnodes.js
straker: I like that idea

lib/commons/dom/get-root-vnodes.js
straker: I'm not sure `getRootChildren` works either though as it sounds like I would return the children of the `html` element rather than the `html` element itself. Maybe that isn't a bad thing though as it would make it way more consistent with shadow DOM

lib/commons/dom/idrefs.js
straker: Ya. We originally did the cache on the root to improve rule node selection. We cached the information on the root node since it was the only part that would have a complete picture of the entire tree. We could work around that limit for just id lookups, but it would require somehow passing the desired `shadowId` for the `getNodesMatchingExpression` query instead of grabbing it from the root node. Maybe we create a `getElementById` like function that allows you do this this? Something like:

```js
function getById(id, shadowId) {
  const root = axe._tree;
  return getNodesMatchingExpression(root, id, () => {}, shadowId)[0]
}


// selector-cache.js
export function getNodesMatchingExpression(domTree, expressions, filter, shadowId) {
  shadowId = shadowId ?? domTree[0].shadowId;
}
```

PR #4846, opened 2025-07-31
fix(get-xpath): return proper relative selector for id

straker:
Closes: #xpath

### DIFF
--- a/test/core/utils/get-xpath.js
+++ b/test/core/utils/get-xpath.js
@@ -1,46 +1,60 @@
-describe('axe.utils.getXpath', function () {
+describe('axe.utils.getXpath', () => {
   'use strict';
 
-  var fixture = document.getElementById('fixture');
+  const fixture = document.getElementById('fixture');
 
-  afterEach(function () {
-    fixture.innerHTML = '';
-  });
-
-  it('should be a function', function () {
+  it('should be a function', () => {
     assert.isFunction(axe.utils.getXpath);
   });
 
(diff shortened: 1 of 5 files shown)


test/core/utils/get-xpath.js
```diff
@@ -1,60 +1,74 @@
...
-    var node = document.createElement('div');
+  it('should generate an XPath selector', () => {
+    const node = document.createElement('div');
     fixture.appendChild(node);
 
-    var sel = axe.utils.getXpath(node);
+    const sel = axe.utils.getXpath(node);
 
-    assert.equal(sel, "/div[@id='fixture']/div");
+    assert.equal(sel, "//div[@id='fixture']/div");
```
dbjorge, line 27: nit: it'd be nice to update these test cases so that in addition to pinning the exact values, they also verify that the selectors actually result in selecting the correct node when used with `document.evaluate`.

dbjorge (approved):
This looks right to me (modulo a nit about improving the tests that you can choose whether to prioritize or not), but asked @WilcoFiers to double check my thinking on the implementation change.

I could see an argument that this should be treated as a breaking change, but I think I'd probably lean towards treating it as a bugfix.

WilcoFiers (approved):
This is potentially disruptive to deduplication algorithms. While this is a bug fix, I don't think we should release this as a patch. We may want to prioritize a 4.11 release for this.

dbjorge (approved):

Issue #4506, opened 2024-06-20
Support `display: contents`


straker:
Currently axe-core does not take into account `display: contents` in any of our rules. We'll need to investigate which rules it affects and how to handle it for each case.

PR #4854, opened 2025-08-12
refactor: refactor flatten-tree to use return early style

straker:
Was making changes to this file and always found it hard to read through with the multiple nesting of if statements and the reduces. refactored to use early return style to prevent the nesting and moved away from array.reduce to just use a forEach over the array to add them (as well as a function to do this since we do it multiple times).

### DIFF
--- a/lib/core/utils/get-flattened-tree.js
+++ b/lib/core/utils/get-flattened-tree.js
@@ -86,7 +86,27 @@ function createNode(node, parent, shadowId) {
 }
 
 /**
- * Recursvely returns an array of the virtual DOM nodes at this level
+ * Add children to the parent virtual node
+ * @param {Node[]} childNodes the children of the parent
+ * @param {VirtualNode} parent the parent VirtualNode
+ * @param {String} shadowId, optional ID of the shadow DOM that is the closest shadow
+ *                           ancestor of the node
+ */
+function createChildren(childNodes, parent, shadowId) {
+  const children = [];
+
+  childNodes.forEach(childNode => {
+    const child = flattenTree(childNode, shadowId, parent);
+    if (child) {
+      children.push(...child);
+    }
+  });
+
+  return children;
+}
+
+/**
+ * Recursively returns an array of the virtual DOM nodes at this level
  * excluding comment nodes and the shadow DOM nodes <content> and <slot>
  *
  * @param {Node} node the current node
@@ -95,15 +115,7 @@ function createNode(node, parent, shadowId) {
  * @param {VirtualNode} parent the parent VirtualNode
  */
 function flattenTree(node, shadowId, parent) {
-  // using a closure here and therefore cannot easily refactor toreduce the statements
   let retVal, realArray;
-  function reduceShadowDOM(res, child, parentVNode) {
-    const replacements = flattenTree(child, shadowId, parentVNode);
-    if (replacements) {
-      res = res.concat(replacements);
-    }
-    return res;
-  }
 
   if (node.documentElement) {
     // document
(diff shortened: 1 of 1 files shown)


lib/core/utils/get-flattened-tree.js
```diff
@@ -86,7 +86,27 @@ function createNode(node, parent, shadowId) {
 }
 
 /**
- * Recursvely returns an array of the virtual DOM nodes at this level
+ * Add children to the parent virtual node
+ * @param {Node[]} childNodes the children of the parent
+ * @param {VirtualNode} parent the parent VirtualNode
+ * @param {String} shadowId, optional ID of the shadow DOM that is the closest shadow
+ *                           ancestor of the node
```
stephenmathieson: Strange formatting - is this whitespace intentional?

lib/core/utils/get-flattened-tree.js
```diff
@@ -119,59 +131,52 @@ function flattenTree(node, shadowId, parent) {
...
+  if (nodeName === 'slot' && typeof node.assignedNodes === 'function') {
+    realArray = Array.from(node.assignedNodes());
+    if (!realArray.length) {
+      // fallback content
+      realArray = getSlotChildren(node);
+    }
+
+    const styl = window.getComputedStyle(node);
+
+    // check the display property. intentionally does not run, see notable information above
```
stephenmathieson: What notable information?

lib/core/utils/get-flattened-tree.js
```diff
@@ -119,59 +131,52 @@ function flattenTree(node, shadowId, parent) {
...
+  if (node.nodeType === 1) {
+    retVal = createNode(node, parent, shadowId);
+    realArray = Array.from(node.childNodes);
+    retVal.children = createChildren(realArray, retVal, shadowId);
+
+    return [retVal];
+  }
+
+  if (node.nodeType === 3) {
+    // text
```
WilcoFiers: If you're worried about the understandability you could do this:

```suggestion
  if (node.nodeType === document.TEXT_NODE) {
```

lib/core/utils/get-flattened-tree.js
```diff
@@ -95,15 +115,7 @@ function createNode(node, parent, shadowId) {
  * @param {VirtualNode} parent the parent VirtualNode
  */
 function flattenTree(node, shadowId, parent) {
-  // using a closure here and therefore cannot easily refactor toreduce the statements
   let retVal, realArray;
```
WilcoFiers: since we're refactoring, I propose renaming these. Maybe:

```suggestion
  let newVNode, childNodes;
```

lib/core/utils/get-flattened-tree.js
```diff
@@ -119,59 +131,52 @@ function flattenTree(node, shadowId, parent) {
...
+  ) {
+    realArray = Array.from(node.getDistributedNodes());
+    return createChildren(realArray, parent, shadowId);
+  }
+
+  if (nodeName === 'slot' && typeof node.assignedNodes === 'function') {
+    realArray = Array.from(node.assignedNodes());
+    if (!realArray.length) {
+      // fallback content
+      realArray = getSlotChildren(node);
```
WilcoFiers: Oh weird! I didn't know slot elements used its children for fallback content.

lib/core/utils/get-flattened-tree.js
```diff
@@ -119,59 +131,52 @@ function flattenTree(node, shadowId, parent) {
...
+    }
+
+    const styl = window.getComputedStyle(node);
+
+    // check the display property. intentionally does not run, see notable information above
+    if (false && styl.display !== 'contents') {
```
WilcoFiers, line 155: I think we can enable this now. IIRC we did this because Chromium at the time couldn't manage `display: content`, and so instead just opted to ignore the slot elements regardless of its display. That was a known bug for them that has been fixed for quite a while. 

Alternatively we can open an issue for axe-core to properly support `display: content`. Essentially ever node that has display:content is pulled from the accessibility tree. But we may want axe-core to have it in its virtual tree so we can report them as being a problem - that has wider implications though, since rules like aria-required-children should know to skip them.

```suggestion
    if (styl.display !== 'contents') {
```

lib/core/utils/get-flattened-tree.js
```diff
@@ -119,59 +131,52 @@ function flattenTree(node, shadowId, parent) {
...
-        return [createNode(node, parent)];
-      }
-      return undefined;
+  }
+
+  if (
+    nodeName === 'content' &&
+    typeof node.getDistributedNodes === 'function'
+  ) {
+    realArray = Array.from(node.getDistributedNodes());
```
WilcoFiers: I don't think we need to do these array conversions. NodeList also has `forEach`. We could also use a for loop, which might be slightly faster.

lib/core/utils/get-flattened-tree.js
```diff
@@ -86,7 +86,27 @@ function createNode(node, parent, shadowId) {
...
+ * @param {VirtualNode} parent the parent VirtualNode
+ * @param {String} shadowId, optional ID of the shadow DOM that is the closest shadow
+ *                           ancestor of the node
+ */
+function createChildren(childNodes, parent, shadowId) {
+  const children = [];
+
+  childNodes.forEach(childNode => {
+    const child = flattenTree(childNode, shadowId, parent);
+    if (child) {
```
WilcoFiers, line 99: Is this ever not true? Every code path I see in flattenTree returns an array, so its always truthy, no?

WilcoFiers (dismissed):
Couple nitpicks I think are worth considering. I much prefer this version of it. Thank you!

lib/core/utils/get-flattened-tree.js
straker: I tried to do a map over it and it would add undefined to the list during tests. The very last statement of the code returns undefined so it's possible.

lib/core/utils/get-flattened-tree.js
straker: I don't think IE11 supports `forEach` on a NodeList, so kept it in since we still have to support it.

lib/core/utils/get-flattened-tree.js
straker: I'd rather leave it off for now and come back to it in a separate pr rather than in this refactor.

lib/core/utils/get-flattened-tree.js
straker: Just copy pasted from a pervious codeblock, so I guess it must be? I can move it to a single line.

lib/core/utils/get-flattened-tree.js
straker: At the top of the file is a large comment about why we turned this off. I had to go back in the code timeline to figure out why the comment `intentionally commented out` meant, so just wanted to update the comment to make it more clear.

lib/core/utils/get-flattened-tree.js
straker: Used `vNode` instead of `newVNode`

lib/core/utils/get-flattened-tree.js
dbjorge: Agree with treating it out of scope. I think https://github.com/dequelabs/axe-core/issues/4506 already exists as the suggested tracking issue.

PR #4858, opened 2025-08-21
docs: add trademarks information to the projects page

dylanb:
Added trademark information to the projects page

### DIFF
--- a/doc/projects.md
+++ b/doc/projects.md
@@ -61,4 +61,4 @@ Add your project/integration to this file and submit a pull request.
 1. [Webatool](https://github.com/balajihambeere/webatool)
 1. [A11y Audit Elixir](https://github.com/angelikatyborska/a11y-audit-elixir)
 
-AXE® and AXE_CORE® are registered trademarks of Deque Systems Inc.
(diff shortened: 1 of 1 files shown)


doc/projects.md
```diff
@@ -1,10 +1,10 @@
-# Projects that use axe-core
+# Projects that use axe-core®
 
 Add your project/integration to this file and submit a pull request.
 
 ## Deque Projects
 
-1. [Axe DevTools](https://www.deque.com/axe/devtools/)
+1. [Axe® DevTools](https://www.deque.com/axe/devtools/)
 1. [Axe Auditor](https://www.deque.com/axe/auditor/)
```
Garbee, line 8: Why do we not put the registered superscript after these other Axe mentions along the list?

dylanb: because legally we only have to annotate the first occurrence

Garbee: Cool, so just like `<abbr>` usage in web pages. Thank you for the verification on that. I didn't know that was actually a legal thing as well. Which, I guess makes sense. Otherwise so many documents would get littered with the symbol.

Issue #4860, opened 2025-08-22
Improved error handling [axe-core]


WilcoFiers:
**Currently**: When an error occurs inside a rule, axe will set an `error` and `errorNode` property. This is initially set to incomplete, but because `nodes` is empty. Because there are no nodes, `finalizeResults` sets this to `inapplicable`. The `v1` reporter than restores that to `incomplete` by detecting the `error` property, but leaving `nodes` empty.

Exceptions: 
- axe's async callback in evaluate accepts errors, but does not assign `errorNode`
- matches functions do not assign `errorNode`
- if an after function throws, the entire run throws

Additionally, if an error occurs inside a frame, the error is bubbled up and will throw the run.

## Goal

1. Axe-core should be able to recover from errors as much as possible, so that rules that can complete won't be blocked by those that errored out.

2. Axe should be transparent about what parts of it failed to complete, so that users can understand what was not tested.

3. Recoverable errors should be reported in the same format as other axe results so they can be given to users without special design considerations

## What to change

1. Instead of using an `error` and `errorNode` property, when an error occurs, the node list is replaced with an array with a single node, where the error occurs. A new `none` check is associated with this with the `error-occurred` id. The `data` object is made the serialized error, so that the `data` property will have `message`, `stack`, and `type`, along with optionally `cause`, which should also be serialized.

  Having a NodeResult for each error ensures that tools that store axe results by node can record display errors. By using axe's existing data format for errors we allow the error to be displayed. By serializing the error's properties we ensure it can be transferred, both across frames, as well as between tools. For backward compatibility, the `error` and `errorNode` properties should continue to be set. 

2. If we do nothing, the above change will result in error nodes getting passed to `after` methods. To ensure backward compatibility we'll need to remove error nodes before padding results to `after`, and add them back in when `after` is done.

3. When an after method throws, axe-core should catch the error and report that rule as `incomplete`, instead of failing completely as it does today. To avoid an overly lengthy number of incompletes, only the first node should be kept, the rest should be set as related nodes, up to a maximum of 10 nodes.

4. When a `matches` function throws, the node that was being matched should be recorded, Currently only the error is.

5. When an `Error` is passed to an `async` evaluate's callback, the node that was being evaluated should be recorded. Currently only the error is.

[...]

PR #4863, opened 2025-08-28
feat: incomplete with node on which an error occurred

WilcoFiers:
Part of https://github.com/dequelabs/axe-core/issues/4860

If a check or matches method throws, catch the error and add a node with an `error-occurred` none check. This way incomplete results are always reported.

This only happens for axe.run / axe.runPartial. `rule.runSync()` still throws if an occurs in a rule or check.


lib/core/base/audit.js
```diff
@@ -732,36 +739,31 @@ function getDefferedRule(rule, context, options) {
...
-        if (!options.debug) {
-          const errResult = Object.assign(new RuleResult(rule), {
-            result: constants.CANTTELL,
-            description: 'An error occured while running this rule',
-            message: err.message,
-            stack: err.stack,
-            error: err,
-            // Add a serialized reference to the node the rule failed on for easier debugging.
-            // See https://github.com/dequelabs/axe-core/issues/1317.
-            errorNode: err.errorNode
```
WilcoFiers, line 752: Note that I got rid of most of these, and updated how `error` works:

- error is now serialized so that it can be carried across frame boundaries
- description is clobbered by the reporter. It never even left axe (plus it has a typo)
- message and stack are duplicated on error
- errorNode wasn't getting pulled through nodeSerializer

ErrorNode in particular was a problem. Because it wasn't serialized, it A. doesn't have the frame selector, and B. leaves the `_element` property on it, which doesn't play well with serialization (axe Extension would completely error out trying to serialize this).

lib/core/utils/support-error.js
```diff
@@ -0,0 +1,23 @@
+import serializeError from './serialize-error';
+
+export default function SupportError({ error, ruleId, method, errorNode }) {
```
stephenmathieson: Any reason this is not a `class` that `extends Error`? If we're going to rewrite it, we may as well use more modern JS features 🤷‍♂️

axe.d.ts
```diff
@@ -167,6 +167,9 @@ declare namespace axe {
     tags: TagValue[];
     nodes: NodeResult[];
   }
+  interface IncompleteResult extends Result {
+    error: Omit<SupportError, 'errorNode'>;
```
dbjorge:
```suggestion
    error?: Omit<SupportError, 'errorNode'>;
```

test/core/utils/support-error.js
```diff
@@ -0,0 +1,45 @@
...
+    const error = new Error('test');
+    const supportError = new SupportError({ error, ruleId: 'test' });
+    assert.equal(supportError.ruleId, 'test');
+    assert.include(supportError.message, 'Skipping test rule.');
+  });
+
+  it('includes the method if provided', () => {
+    const error = new Error('test');
+    const supportError = new SupportError({ error, method: 'test' });
+    assert.equal(supportError.method, 'test');
```
straker: This is the same as the name so want to just make sure it's different

```suggestion
    const supportError = new SupportError({ error, method: '#matches' });
    assert.equal(supportError.method, '#matches');
```

test/core/utils/support-error.js
```diff
@@ -0,0 +1,45 @@
...
+    const error = new Error('test');
+    const supportError = new SupportError({ error });
+    assert.instanceOf(supportError, Error);
+  });
+
+  it('includes the ruleId if provided', () => {
+    const error = new Error('test');
+    const supportError = new SupportError({ error, ruleId: 'test' });
+    assert.equal(supportError.ruleId, 'test');
+    assert.include(supportError.message, 'Skipping test rule.');
```
straker:
```suggestion
    const supportError = new SupportError({ error, ruleId: 'aria' });
    assert.equal(supportError.ruleId, 'aria');
    assert.include(supportError.message, 'Skipping aria rule.');
```

test/core/utils/support-error.js
```diff
@@ -0,0 +1,45 @@
...
+  it('includes the method if provided', () => {
+    const error = new Error('test');
+    const supportError = new SupportError({ error, method: 'test' });
+    assert.equal(supportError.method, 'test');
+  });
+
+  it('includes the errorNode if provided', () => {
+    const error = new Error('test');
+    const supportError = new SupportError({ error, errorNode: 'test' });
+    assert.equal(supportError.errorNode, 'test');
```
straker:
```suggestion
    const supportError = new SupportError({ error, errorNode: 'err' });
    assert.equal(supportError.errorNode, 'err');
```

lib/core/utils/merge-results.js
```diff
@@ -95,6 +95,10 @@ function mergeResults(frameResults, options) {
         if (ruleResult.nodes.length) {
           spliceNodes(res.nodes, ruleResult.nodes);
         }
+        if (ruleResult.error) {
+          // TODO: Test me
```
dbjorge: It looks like this TODO was done, remove the comment?

lib/core/utils/serialize-error.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * Serializes an error to a JSON object
+ * @param e - The error to serialize
+ * @returns A JSON object representing the error
+ */
+export default function serializeError(err, iteration = 0) {
+  if (typeof err !== 'object' || err === null) {
+    return { message: String(err) };
+  }
+  const serial = { ...err }; // Copy all "own" properties
```
dbjorge: This `...err` is going to mean that any (potentially non-serializable) properties of the original error will stay as-is in the returned object, no? Doesn't this defeat the purpose? I'd have expected that whole point of this function would be to make sure the following test case passes:

```ts
  it('should not include unserializable errorNode properties', () => {
    const baseError = new Error('test');
    const errorNode = new axe.utils.DqElement(document.documentElement);
    const supportError = new axe.utils.SupportError({ error: baseError, errorNode })
    const serialized = serializeError(supportError);

    assert.isUndefined(serialized.errorNode?._element);
  })
```

dbjorge: (presumably by only passing through properties known to be serializable and specifically handling `errorNode` by passing it through `nodeSerializer.dqElmToSpec`)

lib/core/utils/support-error.js
```diff
@@ -0,0 +1,23 @@
+import serializeError from './serialize-error';
+
+export default class SupportError extends Error {
```
dbjorge: The name "SupportError" feels off to me; it implies to me "this indicates that axe-core doesn't support something you/the page tried to do", but really we're using this to generically mean a "an error that represents any unexpected failure while processing a rule". Perhaps `RuleError` would be more clear?

lib/core/base/rule.js
```diff
@@ -542,12 +553,20 @@ function sanitizeNodes(result) {
...
-    const afterResults = check.after(beforeResults, checkOption.options);
+    const checkOption = getCheckOption(check, this.id, options);
+    let afterResults;
+    try {
+      afterResults = check.after(beforeResults, checkOption.options);
+    } catch (error) {
+      throw new SupportError({
+        ruleId: this.id,
+        method: `${check.id}#after`,
+        errorNode: result.nodes?.[0].node,
```
dbjorge: There should be a test that would have caught this:

```suggestion
        errorNode: result.nodes?.[0]?.node,
```

lib/core/base/rule.js
```diff
@@ -451,7 +451,18 @@ Rule.prototype.gatherAndMatchNodes = function gatherAndMatchNodes(
...
 
-  nodes = nodes.filter(node => this.matches(node.actualNode, node, context));
+  nodes = nodes.filter(node => {
+    try {
+      return this.matches(node.actualNode, node, context);
+    } catch (error) {
+      throw new SupportError({
+        ruleId: this.id,
+        method: `#matches`,
+        errorNode: new DqElement(node.actualNode),
```
dbjorge: nit: new DqElement can be passed a virtualNode directly

```suggestion
        errorNode: new DqElement(node),
```

test/integration/full/error-occurred/error-ruleset.js
```diff
@@ -0,0 +1,55 @@
...
+  assert.include(result.error.message, message);
+
+  assert.lengthOf(result.nodes, 1);
+  const node = result.nodes[0];
+  assert.lengthOf(node.any, 0);
+  assert.lengthOf(node.all, 0);
+  assert.lengthOf(node.none, 1);
+  assert.equal(node.none[0].id, 'error-occurred');
+  assert.include(node.none[0].message, 'Axe encountered an error');
+  assert.deepEqual(node.target, target);
```
dbjorge, line 18: This is missing a few key checks that are currently failing:

```suggestion
  assert.isDefined(result);

  assert.isDefined(result.error);
  assert.include(result.error.message, message);
  assert.isDefined(result.error.method);
  assert.isUndefined(result.error.errorNode); // would be redundant with result.errorNode

  assert.isDefined(result.errorNode);
  assert.isUndefined(result.errorNode._element);
  assert.deepEqual(result.errorNode.target, target);
  assert.isDefined(result.errorNode.html);

  assert.lengthOf(result.nodes, 1);
  const node = result.nodes[0];
  assert.lengthOf(node.any, 0);
  assert.lengthOf(node.all, 0);
  assert.lengthOf(node.none, 1);
  assert.equal(node.none[0].id, 'error-occurred');
  assert.include(node.none[0].message, 'Axe encountered an error');
  assert.deepEqual(node.none[0].data, result.error);
  assert.deepEqual(node.target, target);
  assert.isDefined(node.html);
```

lib/core/base/audit.js
```diff
@@ -732,36 +743,37 @@ function getDefferedRule(rule, context, options) {
...
+    {
+      id: 'error-occurred',
+      result: undefined,
+      data: serialError,
+      relatedNodes: []
+    }
+  ];
+  const node = errorNode || new DqElement(document.documentElement);
+  return Object.assign(new RuleResult(rule), {
+    error: serialError,
```
dbjorge, line 771: The issue suggests that this should also include `errorNode` here (for compatibility with existing error handling behavior). Note that if we do this, we'd need to pass it through nodeSerializer at frame boundaries.

dbjorge (changes requested):

lib/core/base/audit.js
WilcoFiers: Yeah, that's why I chose not to do that. It's a pain to properly serialize, with little benefit. See PR description.

dbjorge: I don't think "it's a pain to do" is a good enough reason to break backcompat with existing users' error handlers. The specification in the #4860 (in "What to change" 1) mentioned "In addition to using an `error` and `errorNode` property" and "the `error` and `errorNode` properties should continue to be set." I think it'd be better to keep it for backcompat.

I'm not sure what you're referring to by "See PR description", I don't see anything relevant there - did you mean [this comment](https://github.com/dequelabs/axe-core/pull/4863#discussion_r2316240506)? That comment mentions that it's a problem that `errorNode` isn't going through `nodeSerializer` like it should, which I agree with, but I don't think it justifies why breaking the existing property is a preferable solution to "update it to go through nodeSerializer". There aren't that many places where we need to do `nodeSerializer.mergeSpecs` at frame boundaries, I'd expect it to be a reasonably contained change to update those places to make sure they also `mergeSpecs` on error nodes.

Looking at the current code (before this PR) that creates `errorNodes`, I'm actually a bit surprised/confused that we see any current scenarios where they contain `_element` properties that confuse serialization; the only places that appear to do an initial set of `errorNode` are in `check.js` where they come from a `nodeSerializer.toSpec` output. It makes sense that the original values would be missing ancestor frame information, but it surprises me that they'd end up with `_element` properties.

lib/core/utils/merge-results.js
WilcoFiers:
```suggestion
```

lib/core/base/audit.js
WilcoFiers: I don't agree it's a breaking change:
1. It's already an optional prop: not every method sets it, so not having it can't possibly break things.  
2. It's an undocumented property, changing those isn't a breaking change. 
3. If this was breaking, than so would removing `message` and `stack` be.

`errorNode` is also much buggier than I realized it was when I wrote the issue. It doesn't work across frames, and it doesn't run through NodeSerializer. Everything `errorNode` does can now be done in a better way. `errorNode` is a "feature" that has never worked, and that we keep making mistakes with. Better to drop it IMO.

dbjorge: I think 1 is a bit of a sketchy argument - I think if someone had a UI for errors that attempted to display which node an error occurred at, and that UI worked fine for errors in checks in <4.11 but stopped working for errors in checks in >=4.11, I think they could reasonably argue that the change broke them.

But points 2, 3, and that it's already really buggy anyway are all fair enough. I verified that I don't think we have any products with such an error UI already (we do have one product that *parses* `errorNode` if it's present, but it doesn't do anything with it). I can live with dropping it since it sounds like you feel strongly about it, it's not a hill I'm going to fight further on.

lib/core/utils/serialize-error.js
WilcoFiers: Yeah it's not perfect. I'm not sure there's an easy way to ensure this. Considering this is internal, this seemed sufficient to me.

dbjorge: if you leave an unserializable property on the "serialized" error, doesn't that defeat the purpose of the PR and still leave you in the same place as you were originally, where you'll get errors when trying to pass an axe result containing the partially-unserializable error across a serialization boundary?

If we're satisfied with not having `errorNode` and instead relying on the node in the synthetic check result, I think the "easy way to ensure this" is to replace `{ ...err }` with `{}` here and add the non-errorNode properties of `RuleError` to the list of props on L12. (maybe also add `code` to the list, it's common enough in things that could be error `cause`s)

PR #4864, opened 2025-08-30
Revert 1 develop

sierra88-en:
<< Describe the c
G hanges >>

Closes:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,6 +14,8 @@ updates:
         update-types:
           - 'minor'
           - 'patch'
+    cooldown:
+      default-days: 7
 
   - package-ecosystem: 'npm'
     directory: '/'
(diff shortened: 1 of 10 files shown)


lib/rules/color-contrast-matches.js
```diff
@@ -137,10 +137,6 @@ function colorContrastMatches(node, virtualNode) {
     return false;
   }
 
-  if (!parseFloat(virtualNode.getComputedStylePropertyValue('font-size'))) {
```
stephenmathieson, line 140: Why is this being removed?

stephenmathieson (changes requested):
I do not understand what the intent of this patch is. Changes to axe-core will not be made without a discussion and/or accompanying ticket. If there are changes you want to make, please create an issue first.

Issue #4129, opened 2023-08-21
Deprecate DqElement.fromFrame


straker:
With https://github.com/dequelabs/axe-core/pull/409 we should deprecate `DqElement.fromFrame`.

dbjorge:
To clarify, I think this is referring to the static `DqElement.fromFrame` function (at [`dq-element.js`](https://github.com/dequelabs/axe-core/blob/develop/lib/core/utils/dq-element.js) L215), not the `someDqElementInstance.fromFrame` instance property (at L130). The latter is still used in many places, but I think the former is not used anymore in axe-core.

PR #4881, opened 2025-09-16
feat(DqElement): deprecate fromFrame function

straker:
The function is no longer used in the code base and was replaced in #4093 

Closes: https://github.com/dequelabs/axe-core/issues/4129

### DIFF
--- a/lib/core/utils/dq-element.js
+++ b/lib/core/utils/dq-element.js
@@ -212,7 +212,7 @@ DqElement.prototype = {
   }
 };
 
-// @deprecated
+/** @deprecated */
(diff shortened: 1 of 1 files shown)


lib/core/utils/dq-element.js
```diff
@@ -212,6 +212,7 @@ DqElement.prototype = {
   }
 };
 
+// @deprecated
```
dbjorge:
```suggestion
/** @deprecated */
```

dbjorge (dismissed):

stephenmathieson (approved):

dbjorge (approved):

PR #4883, opened 2025-09-16
ci: fix isolated-env test

straker:
This test in firefox started to fail all of a sudden, and [almost consistently](https://github.com/dequelabs/axe-core/pull/4881). I couldn't figure out why the test kept failing, so after playing around with the order of operations I figured out that the iframe is loading axe and calling the `axe-loaded` event [_before_ mocha runs the `before`](https://app.circleci.com/pipelines/github/dequelabs/axe-core/7369/workflows/8fa125f3-2a32-43d9-8ec9-d5a9ccd172df/jobs/79258) where we look for the event. So to fix it I changed up the loading order so the script runs first and adds the event tracking, and now the event tracking now happens outside the `before` function and keeps track of the logs. Then inside the `before` we see if the message has already been fired, otherwise we listen to the event as we were doing before.

### DIFF
--- a/test/integration/full/isolated-env/isolated-env.js
+++ b/test/integration/full/isolated-env/isolated-env.js
@@ -1,4 +1,8 @@
 /* global chai */
+var messages = [];
+window.addEventListener('message', function (msg) {
+  messages.push(msg.data);
+});
 
 describe('isolated-env test', function () {
   'use strict';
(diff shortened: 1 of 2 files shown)


test/integration/full/isolated-env/isolated-env.js
```diff
@@ -40,11 +44,16 @@ describe('isolated-env test', function () {
...
     var isloadedPromise = new Promise(function (resolve, reject) {
-      window.addEventListener('message', function (msg) {
-        if (msg.data === 'axe-loaded') {
-          resolve();
-        }
-      });
+      if (messages.includes('axe-loaded')) {
+        resolve();
+      } else {
+        window.addEventListener('message', function (msg) {
```
Garbee, line 50: Minor thing and since this in our tests I don't think it'll matter _too_ much. But should we not remove this listener after it has resolved? That way there are no extraneous listeners during test execution. And we can do this through a signal to cancel listening instead of doing a named function then `removeEventListener`. Bit cleaner these days.

straker: Nah, this page runs in isolation so can't affect other pages and is closed once it's run so nothing will remain in memory or anything

Issue #4535, opened 2024-07-10
scrollable-region-focusable should consider an element's content real content, not its content box


WilcoFiers:
The following shouldn't fail the `scrollable-region-focusable` rule:

```html
<style>
  #container {
    width: 300px;
    background: #aff;
    overflow-y: auto;
  }
  p {
    width: 600px;
  }
</style>
<div id="container">
  <p>Contents</p>
</div>
```

The text in this paragraph is fully visible. Axe-core is failing this because its detecting a `p` element who's content box ix much larger than it scrollable parent. Axe shouldn't use the content box here though, it should use the actual size and position of the element's content.

padmavemulapati:
Verified with develop version of axe-core, not seeing `scrollable-regio-focusable` rule failing. 
"The text in this paragraph is fully visible.  axe-core is detecting a `p` element who's content box is much larger than scrollable parent. 

https://github.com/user-attachments/assets/fecaf936-567b-4310-acbf-10f8ee841a6a

PR #4884, opened 2025-09-18
fix(scrollable-region-focusable): do not fail elements whose contents are fully visible without scrolling

straker:
Doing this required a bit of change to current tests as they were failing exactly what this ticket tries to solve, that is creating a larger scroll within the scrolling parent that was just empty scroll area. To fix that I either had to add more content that would scroll off the screen and/or update the size of the scrolling parent to make sure the content would scroll.

I tried to use `getVisibleTextRects` to determine the rect of the contents, but it also filters out rects either outside the bounding rect of the element or creates a smaller subrect that falls completely within the bounding rect if it's too large. Both cases were not what we needed for the check since we want to know the exact dimensions without modification. So I instead copied the main part of the function and put it in the matches directly.

I also made sure that the tests took into account non-text content (such as images) and to fail / ignore the scrolling parent if the content needed to scroll.

Closes: #4535

### DIFF
--- a/lib/rules/scrollable-region-focusable-matches.js
+++ b/lib/rules/scrollable-region-focusable-matches.js
@@ -1,21 +1,56 @@
 import hasContentVirtual from '../commons/dom/has-content-virtual';
 import isComboboxPopup from '../commons/aria/is-combobox-popup';
+import sanitize from '../commons/text/sanitize';
 import { querySelectorAll, getScroll } from '../core/utils';
 
+const buffer = 13;
+
 export default function scrollableRegionFocusableMatches(node, virtualNode) {
+  const boundingRect = virtualNode.boundingClientRect;
   return (
     // The element scrolls
-    getScroll(node, 13) !== undefined &&
+    getScroll(node, buffer) !== undefined &&
     // It's not a combobox popup, which commonly has keyboard focus added
     isComboboxPopup(virtualNode) === false &&
     // And there's something actually worth scrolling to
-    isNoneEmptyElement(virtualNode)
+    hasScrollableContent(node, virtualNode, boundingRect)
   );
 }
 
-function isNoneEmptyElement(vNode) {
-  return querySelectorAll(vNode, '*').some(elm =>
-    // (elm, noRecursion, ignoreAria)
-    hasContentVirtual(elm, true, true)
-  );
+function hasScrollableContent(node, virtualNode, boundingRect) {
+  return querySelectorAll(virtualNode, '*').some(vNode => {
+    const hasContent = hasContentVirtual(vNode, true, true);
+    if (!hasContent) {
+      return false;
+    }
+
+    return getChildTextRects(vNode).some(
+      rect =>
+        // part or all of the element is outside the scroll area
+        rect.left - boundingRect.left + rect.width >
+          node.clientWidth + buffer ||
+        rect.top - boundingRect.top + rect.height > node.clientHeight + buffer
+    );
+  });
+}
+
+function getChildTextRects(vNode) {
+  const boundingRect = vNode.boundingClientRect;
+  const clientRects = [];
+
+  vNode.actualNode.childNodes.forEach(textNode => {
+    if (textNode.nodeType !== 3 || sanitize(textNode.nodeValue) === '') {
+      return;
+    }
+
+    clientRects.push(...getContentRects(textNode));
+  });
+
+  return clientRects.length ? clientRects : [boundingRect];
+}
+
+function getContentRects(node) {
+  const range = document.createRange();
+  range.selectNodeContents(node);
+  return Array.from(range.getClientRects());
 }
--- a/test/integration/rules/scrollable-region-focusable/scrollable-region-focusable.html
+++ b/test/integration/rules/scrollable-region-focusable/scrollable-region-focusable.html
@@ -3,9 +3,12 @@
   <input type="text" />
 </div>
 
-<div id="pass2" style="height: 200px; overflow-y: auto" tabindex="0">
-  <div style="height: 2000px">
-    <p>Content</p>
+<div id="pass2" style="width: 200px; overflow-y: auto" tabindex="0">
+  <div>
+    <p style="width: 2000px">
+      Sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium
+      doloremque laudantium.
+    </p>
   </div>
 </div>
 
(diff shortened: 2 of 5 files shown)


lib/rules/scrollable-region-focusable-matches.js
```diff
@@ -1,21 +1,56 @@
 import hasContentVirtual from '../commons/dom/has-content-virtual';
 import isComboboxPopup from '../commons/aria/is-combobox-popup';
+import sanitize from '../commons/text/sanitize';
 import { querySelectorAll, getScroll } from '../core/utils';
 
+const buffer = 13;
```
stephenmathieson, line 8: What is this magic number?

stephenmathieson (changes requested):
The added magic number should be documented, something like:

```js
/**
 * This is a magic number that is magical and important. It is
 * responsible for many things in the universe.
 */
const SOME_MAGIC_NUMBER = 13
```

Note - I did not review the rest of this PR, but feel strongly enough that this needs docs to justify requesting changes.

lib/rules/scrollable-region-focusable-matches.js
straker: Good question. had to go back in git history [to the pr it was added](https://github.com/dequelabs/axe-core/pull/1396) to see why it was 13. Added a comment.

lib/rules/scrollable-region-focusable-matches.js
```diff
@@ -1,21 +1,58 @@
...
+
+  vNode.actualNode.childNodes.forEach(textNode => {
+    if (textNode.nodeType !== 3 || sanitize(textNode.nodeValue) === '') {
+      return;
+    }
+
+    clientRects.push(...getContentRects(textNode));
+  });
+
+  return clientRects.length ? clientRects : [boundingRect];
```
dbjorge, line 51: I don't think this is quite right:

- I think there are situations where you'd want to include the whole bounding client rect even if text nodes are present (eg, visual content elements, and maybe elements that have CSS properties like `background-image` that might give meaning to the whole client rect)
- I think there are situations where either a text node rect or the whole element might be hidden for reasons other than the scroll container, where we'd probably want to ignore it for the purposes of this matches

lib/rules/scrollable-region-focusable-matches.js
```diff
@@ -1,21 +1,58 @@
...
+    if (!hasContent) {
+      return false;
+    }
+
+    return getChildTextRects(vNode).some(
+      rect =>
+        // part or all of the element is outside the scroll area
+        rect.left - boundingRect.left + rect.width >
+          node.clientWidth + buffer ||
+        rect.top - boundingRect.top + rect.height > node.clientHeight + buffer
```
dbjorge, line 34: This is only checking the right and bottom sides, which might be ok if we knew the scroll container was in its leftmost/topmost scroll position, but we aren't checking that. I think this should either:

- check all 4 directions, or
- identify which direction(s) the user could still scroll further in, and check only those directions

I could live with either. The latter might be better at ignoring the case where someone uses an element-hiding technique based on a negative x position, but it'd be more complicated and I could imagine it missing cases with nested different-direction scroll containers that we'd prefer it to match.

lib/rules/scrollable-region-focusable-matches.js
```diff
@@ -1,21 +1,58 @@
...
   );
 }
 
-function isNoneEmptyElement(vNode) {
-  return querySelectorAll(vNode, '*').some(elm =>
-    // (elm, noRecursion, ignoreAria)
-    hasContentVirtual(elm, true, true)
-  );
+function hasScrollableContent(node, virtualNode, boundingRect) {
+  return querySelectorAll(virtualNode, '*').some(vNode => {
```
WilcoFiers, line 23: Is this going to create a performance problem? I think its worth considering if there are other ways to do this. This may end up running a `*` query hundreds of times on a page.

lib/rules/scrollable-region-focusable-matches.js
```diff
@@ -1,21 +1,58 @@
...
+  const boundingRect = virtualNode.boundingClientRect;
   return (
     // The element scrolls
-    getScroll(node, 13) !== undefined &&
+    getScroll(node, buffer) !== undefined &&
     // It's not a combobox popup, which commonly has keyboard focus added
     isComboboxPopup(virtualNode) === false &&
     // And there's something actually worth scrolling to
-    isNoneEmptyElement(virtualNode)
+    hasScrollableContent(node, virtualNode, boundingRect)
```
WilcoFiers, line 18: Why are you passing boundingRect in? You're also giving it virtualNode, which has that as a prop. Doesn't seem necessary.

lib/rules/scrollable-region-focusable-matches.js
WilcoFiers: Agreed, only looking at text here is going to miss problems.

lib/rules/scrollable-region-focusable-matches.js
```diff
@@ -1,21 +1,58 @@
...
   );
 }
 
-function isNoneEmptyElement(vNode) {
-  return querySelectorAll(vNode, '*').some(elm =>
-    // (elm, noRecursion, ignoreAria)
-    hasContentVirtual(elm, true, true)
-  );
+function hasScrollableContent(node, virtualNode, boundingRect) {
```
WilcoFiers, line 22: I wonder if we can use the grid to do this. That seems like it'd be more efficient. Each scrollable element should have a subgrid, right? Why not look at the subgrid, find each element that actually extends beyond the visible boundary, and then check if A. that element has text that extends the boundary, or B. that element is a graphical object with a non-empty accessible text (i.e. it's not decorative).

lib/rules/scrollable-region-focusable-matches.js
straker: Since it's the last part of the `if` statement, I think this should be fine. It'll only run if the other conditions are met first, plus this isn't doing anything it wasn't doing before.

lib/rules/scrollable-region-focusable-matches.js
straker: I'm not sure this would work easily. The grid was designed to do bottom-up lookups from a cell of a grid really easily, but it doesn't do top-down from a container very well. To do this we'd have to loop over every element within the grid, which requires looping over every cell of the grid, and then check if that element is itself a scroll container (checking `elm._subgrid`), and then start again and loop over every cell / element within the subgrid in order to find all elements to see if they have a text node or is a graphical object.

straker:
Superseded by https://github.com/dequelabs/axe-core/pull/4993

PR #4893, opened 2025-09-30
fix(type): correct RuleError type

WilcoFiers:
When changing SupportError to RuleError I missed the type file.

### DIFF
--- a/axe.d.ts
+++ b/axe.d.ts
@@ -168,7 +168,7 @@ declare namespace axe {
     nodes: NodeResult[];
   }
   interface IncompleteResult extends Result {
-    error?: Omit<SupportError, 'errorNode'>;
+    error?: Omit<RuleError, 'errorNode'>;
   }
   interface NodeResult {
     html: string;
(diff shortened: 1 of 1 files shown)


axe.d.ts
```diff
@@ -479,12 +479,12 @@ declare namespace axe {
...
       selector: unknown
     ) => selector is LabelledShadowDomSelector;
-    SupportError: (
+    RuleError: (
       error: Error,
       ruleId?: string,
       method?: string,
       errorNode?: DqElement
-    ) => SupportError;
+    ) => RuleError;
```
straker: This type is not correct. It's a class so needs `new` and it should be an object for the param not a list of params. Otherwise this passes `tsc` when it shouldn't.

```js
const ruleError = axe.utils.RuleError(new Error('foo'));
```

Instead it should be (github won't let me suggest)

```ts
   RuleError: new (options: {
      error: Error,
      ruleId?: string,
      method?: string,
      errorNode?: DqElement
    }) => RuleError
```

Issue #4912, opened 2025-10-24
Update deployment workflow to use OIDC


Garbee:
The publishing needs to be moved to Github Actions and made to use OIDC tokens in the process.

Garbee:
Workflow is failing. Needs triage.

Garbee:
Confirmed with nightly builds. Per the epic discussion during planning, closing this as done since axe-core may not see a patch update for a bit to fully confirm this.

PR #4918, opened 2025-10-31
ci(nightly): port testing to github actions

Garbee:
This patch begins moving the workflows from CircleCI to Github. In order to minimize the reviews, the first chunk is only converting the nightly tests.

# Key Changes

## Removal of Circle CI Config for Nightly Tests

Since this patch adds support for running nightly tests in GitHub Actions, we no longer need it in Circle CI. It would be wasteful of resources to do the same work in multiple places.

## Install Deps Composite Action

This file reduces duplication of steps between multiple jobs and workflows. It is called after a repository is checked out in CI in order to install core OS dependencies, browsers, package dependencies, and starts (when needed) a virtual X server for running headed browsers against for testing.

## Nightly Tests

The nightly workflow runs the current beta releases for Firefox and Chrome, WCAG ACT's current version, and the latest aria practices tests.

## Webdriver Configuration

Previously, `browser-driver-manager` was used to force install the correct browser, driver, and updated selenium to match. This caused specific quirks in ordering of steps that had to be adhered to. With this conversion, the sources are now setup to take in the needed paths by an environment variable. This allows the `install-deps` action to install the browsers as they are an OS level dependency to run. Then the paths are passed to the test runs and utilized. Bypassing the need to run certain installs in a specific order.

# Verification

The original code in the PR was running on every push to get a test run going. [This run](https://github.com/dequelabs/axe-core/actions/runs/18976324783) was from the commit before the final push which turned off the `on: push` execution. This shows the workflow successfully running with all nightly tests passing.

Refs: #4912

### DIFF
--- a/.github/workflows/nightly-tests.yml
+++ b/.github/workflows/nightly-tests.yml
@@ -7,8 +7,6 @@ on:
     # to avoid high contention times on GitHub runners.
     - cron: '17 2 * * *'
   workflow_dispatch:
-  # For testing
-  push:
 
 env:
   CHROME_DEVEL_SANDBOX: /opt/google/chrome/chrome-sandbox


test/act-rules/act-runner.js
```diff
@@ -28,11 +28,8 @@ module.exports = ({ id, title, axeRules, skipTests = [] }) => {
     this.timeout(50000);
     this.retries(3);
 
-    before(async () => {
-      driver = getWebdriver();
-    });
-
     before(done => {
+      driver = getWebdriver();
```
Garbee, line 32: At one point I thought I had a race condition on things. Ended up looking at this and realized we had duplicate `before` blocks that aren't necessary. So got them together for better clarity in the processing order.

.github/workflows/nightly-tests.yml
```diff
@@ -0,0 +1,77 @@
...
+
+env:
+  CHROME_DEVEL_SANDBOX: /opt/google/chrome/chrome-sandbox
+
+permissions:
+  contents: read
+
+jobs:
+  browsers:
+    runs-on: ubuntu-24.04
```
Garbee, line 19: I'm going to start locking our ubuntu version as well. Just since some packages we may need to install might change names between versions, especially with the unix timestamp things going on.

Extra layer of pinning where feasible to help make sure changes under us don't break us later.

.github/workflows/nightly-tests.yml
```diff
@@ -0,0 +1,77 @@
...
+      - &build
+        name: Build
+        run: |
+          npm run prepare
+          npm run build
+      - name: Run Firefox Nightly Browser Tests
+        env:
+          FIREFOX_NIGHTLY_BIN: ${{ steps.install-deps.outputs.firefox-path }}
+        run: npm run test -- --browsers FirefoxNightly
+      - name: Run Chrome Beta Browser Tests
```
straker, line 42: So before we would run chrome and firefox in kinda parallel with the command `run: npm run test -- --browsers Chrome,FirefoxNightly` letting Karma handle how to run it, but this makes it run serially. Any reason why you separated it out like that?

test/act-rules/act-runner.js
straker, line 32: I think this is incorrect. `getWebdriver` returns a promise as the function returns the `new Builder().build()` but doesn't await it. I think the prior before being async meant it would return the promise and wait for it to resolve before completing. Now I believe `driver` is the promise and not the resolved promise value.

.github/workflows/nightly-tests.yml
Garbee: Eventual reporting to Slack when failures happen. To have it more concisely tell us which browser failed. Since we can check the step that failed when the time comes. At least for right now, it is more clear if things do fail from a nightly which it is from the log. Instead of parsing the log contents to figure out the browser in question.

test/act-rules/act-runner.js
Garbee: That is not what the prior behavior did. It would put the `before` function on the microtask queue. Then it would spin up, run the thing and put the promise to the variable. It would not await it, since there was no `await` in the function body itself.

If we _need_ to wait on this for safety, then there are two ways to handle it. I'll look first into making sure it is done here where both those are the one function since we have the `done` callback.

straker: Hu weird. Testing this out it looks like webdriver handles oddities like this. It returns a `thenable` but doesn't resolve the driver build until much later. When `driver.get` is called it'll resolve it right there, but if I comment out the `driver.get` it will resolve much later. So it seems webdriver is ok handling a driver that is still waiting for the promise when we call `get` later. I didn't know that's how it worked.

Garbee: I just double checked the type because I remember looking into this and doing this for a reason... and yes, this `build` method is *NOT* asynchronous. Ref: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/selenium-webdriver/index.d.ts#L534-L547

It returns a *thenable* object so the result can be used as such. But the main call itself is synchronous. So this is doing the exact same behavior as before. The extra async hook was not providing any value.

straker: Webdriver is so weird

PR #4919, opened 2025-10-31
ci(tests): port to github actions

Garbee:
This patch ports the main test workflow over to GHA. This is not removing Circle CI configuration as we need to convert required checks over. As well as only removing that after the publishing is ported over to GitHub.

One key change in the patterns from the CircleCI system, is the browser tests and the examples have to do the build internally instead of using the build artifact. I have not had the time to fully investigate what is going on with the differences. I _believe_ it fails in re-use now because of some slight difference in the way the caches were done before vs how they are done in GitHub Actions.

As an aside, I realized in the previous nightly PR I did mess up one aspect. Since the browsers are individual steps now, if the first running fails it was blocking the second from proceeding. For that situation, I added an instruction to always run the second browser. This is good enough for now and we can further optimize later when we setup better failure reporting.

Refs: #4912

### DIFF
--- a/test/integration/full/contrast/prototype.html
+++ b/test/integration/full/contrast/prototype.html
@@ -24,7 +24,7 @@
   </head>
 
   <body>
-    <div id="fixture" style="background-color: darkred">Text</div>
+    <div id="fixture" style="background-color: darkred; color: black">Text</div>
     <script src="/test/testutils.js"></script>
     <script src="prototype.js"></script>
     <script src="/test/integration/adapter.js"></script>
--- a/test/integration/full/preload-cssom/preload-cssom.js
+++ b/test/integration/full/preload-cssom/preload-cssom.js
@@ -40,6 +40,10 @@ describe('preload cssom integration test', function () {
     styleTagWithCyclicCrossOriginImports: {
       id: 'styleTagWithCyclicCrossOriginImports',
       text: '@import "cyclic-cross-origin-import-1.css";'
+    },
+    styleTagWithNonExistentImport: {
+      id: 'styleTagWithNonExistentImport',
+      text: '@import "non-existent-import.css";'
     }
   };
   let stylesForPage;
@@ -141,12 +145,7 @@ describe('preload cssom integration test', function () {
     });
 
     it('throws if cross-origin stylesheet fail to load', done => {
-      stylesForPage = [
-        {
-          id: 'nonExistingStylesheet',
-          text: '@import "import-non-existing-cross-origin.css";'
-        }
-      ];
+      stylesForPage = [styleSheets.styleTagWithNonExistentImport];
       attachStylesheets(
         {
           root: root,
@@ -157,8 +156,12 @@ describe('preload cssom integration test', function () {
             done(err);
           }
           getPreloadCssom(root)
-            .then(() => {
-              done(new Error('Expected getPreload to reject.'));
+            .then(sheets => {
+              done(
+                new Error(
+                  `Expected getPreload to reject, but it resolved with ${JSON.stringify(sheets)}.`
+                )
+              );
             })
             .catch(e => {
               assert.isDefined(e);
--- a/test/testutils.js
+++ b/test/testutils.js
@@ -357,7 +357,7 @@ var commons;
         doc.head.appendChild(link);
       });
     } else {
-      q.defer(function (resolve) {
+      q.defer(function (resolve, reject) {
         const style = doc.createElement('style');
         if (data.id) {
           style.id = data.id;
@@ -365,9 +365,25 @@ var commons;
         style.type = 'text/css';
         style.appendChild(doc.createTextNode(data.text));
         doc.head.appendChild(style);
-        setTimeout(function () {
-          resolve();
-        }, 100); // -> note: gives firefox to load (document.stylesheets), other browsers are fine.
+        // In Firefox, there is a delay between adding the element and it appearing in
+        // document.styleSheets, so we poll until we see it there.
+        const timeoutAt = Date.now() + 500;
+        const interval = setInterval(function () {
+          const isLoaded = Array.from(doc.styleSheets).some(
+            sheet => sheet.ownerNode === style
+          );
+          if (isLoaded) {
+            clearInterval(interval);
+            resolve();
+          } else if (Date.now() > timeoutAt) {
+            clearInterval(interval);
+            reject(
+              new Error(
+                'Added <style> element was not reflected in doc.styleSheets within timeout'
+              )
+            );
+          }
+        }, 5);
       });
     }
     return q;


.github/workflows/test.yml
```diff
@@ -7,50 +7,214 @@ on:
       - master
       - develop
 
+concurrency:
```
Garbee, line 14: Since we run on PRs and on pushes to key branches, the concurrency setup will run each `push` event one at a time in series. So everything will fully finish there. Meanwhile for PRs, if something is rapidly pushed to a PR branch, it will cancel the previous commits run since it is no longer useful to the PR and continue with running the latest commit on those branches.

This helps save resources for the project.

.github/workflows/test.yml
```diff
@@ -7,50 +7,213 @@ on:
...
+  group: axe-core Tests ${{ github.ref_name }}
+  cancel-in-progress: ${{ github.event_name != 'push' }}
+
+permissions: {}
```
Garbee, line 18: Nifty trick for disabling all permissions in one sweep. So we refine permissions needed (if any) at the job level.

Ref: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#defining-access-for-the-github_token-scopes

test/integration/full/test-webdriver.js
```diff
@@ -142,7 +142,7 @@ function buildWebDriver(browser) {
     const options = new firefox.Options().addArguments('--headless');
 
     if (process.env.FIREFOX_BIN) {
-      options.setBinaryPath(process.env.FIREFOX_BIN);
+      options.setBinary(process.env.FIREFOX_BIN);
```
Garbee, line 145: I got confused somewhere and had the wrong method in here. It just wasn't hit in the nightly testing where I was setting this env change up since it used a different variable.

Why does Webdriver not have uniform method naming for doing the same action between browsers? No clue. But we also can't use `// @ts-check` in this file since so many other things are broken. That will need a bigger effort to go fix on its own to help avoid these issues.

.github/workflows/test.yml
```diff
@@ -7,50 +7,213 @@ on:
...
+  test_chrome:
+    runs-on: ubuntu-24.04
+    timeout-minutes: 10
+    env:
+      DISPLAY: :99
     steps:
-      - uses: actions/checkout@v5
-      - uses: actions/setup-node@v6
+      - *checkout
+      - &install-deps-with-xvfb
```
straker, line 77: So if I understand everything correctly, calling `&install-deps-with-xvfb` will reinstall the node modules for every step and not cache it, whereas in circle we found a way to cache the build and the node modules so it only happened once for the entire run. This isn't so much a problem for now as as you said for the build we can worry about optimizing after. Looking at the runs, each of the tests take ~2mins longer than their circleCI counterpart, but since it all runs in parallel it doesn't add too much time in the long run.

straker (dismissed):
All LGTM, but I would like @dbjorge or @WilcoFiers approval so we don't miss anything. 

I did notice we didn't port over the `test_rule_help_version` step which should be fine as that should only run on a pr to `master` but we were running it on `develop`, which I don't think we need to do since a pr to `develop` will never increase the version number before it goes to `master`.

.github/workflows/test.yml
WilcoFiers: That feels like a step back. I don't care if we do that in this PR or a separate one but I don't think we should call this done until we've sorted this.

Garbee: So, cacheing of node modules is handled by `actions/setup-node`. It caches the local cache of node_modules and not the modules in the repo after installation. Reason being, just restoring the cache in the repo may miss running install scripts that make things work. (This is probably why some of the tests broke after the move.)

The build time is like 30 seconds of effort when we do need to re-run it. Not enough time to run up a bill too much.

Ideally, we'd ultimately end up building docker images we run against that have all the OS deps pre-installed (aside from nightly stuff which of course we need to install at runtime.) This would reduce the bulk of that extra ~2m time. Since most of it is spent downloading and installing the same OS packages on every runner.

No one that I'm aware of in the org has done Docker Images for CI yet in GHA. This "install deps" shared action is how every other project has done this that I've seen. So it's a trade-off the org has accepted for the initial conversion of these projects to GHA. If we want to prioritize setting up the infrastructure for making docker images and refreshing them regularly for CI to run against, I'm all for it. But that's a bigger task that will need to be planned out with PO buy-in so it has accounted time to work on it and get it done.

.github/workflows/nightly-tests.yml
```diff
@@ -40,6 +40,7 @@ jobs:
           FIREFOX_NIGHTLY_BIN: ${{ steps.install-deps.outputs.firefox-path }}
         run: npm run test -- --browsers FirefoxNightly
       - name: Run Chrome Beta Browser Tests
+        if: ${{ always() }}
```
dbjorge: I don't think this is quite right; I understand what you mentioned in the PR description about wanting this to run even if the firefox tests failed, but we still want this to be cancelled if the workflow as a whole is cancelled or if deps/build/etc fail. I think you'd want something like:

```suggestion
        if: ${{ !cancelled() && steps.build.conclusion == 'success' }}
```

(this will require giving `id: build` to the Build step)

.github/workflows/test.yml
dbjorge: I think this behavior is ok for PRs, but unnecessarily serializes tests for different commits that are merged to develop near each other. Consider instead using a `group` identifier that groups together runs from the same PR but doesn't group together runs from the same push-target (like [this example from GitHub's docs](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency#example-using-a-fallback-value)):

```
  # head_ref is only defined for PRs, not pushes, so this will cause all pull_request
  # events for a single PR to be grouped, but will leave different push events ungrouped
  # (even for the same branch).
  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
  cancel-in-progress: true
```

.github/workflows/test.yml
dbjorge: Garbee is correct, `actions/setup-node` provides caching that is more robust than what we were doing in circleci and I agree that we should use that rather than caching the whole `node_modules` folder - I've had to debug issues caused by over-caching of whole `node_modules` folders before, it's a rough debugging experience I don't think caching `node_modules` instead is worth 20s of wall time.

This PR *is* caching the build between most jobs already (that's what `&restore-axe-build` does). But I agree with Garbee's choice to *not* use the cached build in the long-pole jobs; because `build` is faster than `install_deps` (even cached), the total wall-time of "wait for build job -> install_deps_with_xvfb -> download build" is less than the wall-time of "start immediately -> install_deps_with_xvfb -> redundant build". Unfortunately, GitHub Actions' `needs` dependencies only operate at the job level, not the step level; I'm not aware of a great way to represent "let the `test_chrome` job start doing its dep installation in parallel with the build job, and don't block on the build job until you actually get to the 'download artifact' step", which is the behavior you'd really want here.

I agree with Garbee that the part of dep installation that is actually most expensive in the long-pole `test_chrome` and `test_firefox` jobs is the xvfb/browser installation , and that the way you'd want to optimize that is to run builds against docker images that already have those pre-installed. We have an issue at https://github.com/dequelabs/on-prem-services/issues/8 tracking that we'd like to implement a centralized setup for automatically-refreshed common docker images for different teams to use, but haven't found time to implement it yet. I've updated that ticket to note that we'd like an image for this use case.

Playwright publishes docker images that come with xvfb and browser deps pre-installed that we might consider using here even without otherwise using Playwright ([docs](https://playwright.dev/docs/ci#via-containers)). But I think we should leave experimenting with that for a separate PR/issue, the jobs this PR sets up are already faster than circle as-is ([the total runtime for GHA here is 5m59s](https://github.com/dequelabs/axe-core/actions/runs/19103518006?pr=4919), vs [6m49s for circle](https://app.circleci.com/pipelines/github/dequelabs/axe-core/7563/workflows/a4aa3e7c-284c-41ce-b1d1-0d76b000b992))

dbjorge (dismissed):
Looks good to me overall, only two suggested changes are both minor.

.github/workflows/test.yml
Garbee: Nifty trick. That's a great upgrade to keep the core branches running smoothly. I'll definitely get it added.

Garbee: Just added, and left an extra bit on the comment to explain why this is done.

.github/workflows/nightly-tests.yml
Garbee: Good catch.

dbjorge (approved):
I debugged the flaky test and created a fix in https://github.com/dequelabs/axe-core/pull/4938. The fix is independent of this PR.

scottmries (dismissed):
Looks ok to me, given the above the discussion.

PR #4929, opened 2025-11-06
ci(deploy): move to gha for oidc publishing

Garbee:
# Background

In order to continue publishing to npm, the org is moving to OIDC tokens. This is setup through GitHub Actions. Which has prompted a conversion from CircleCI to GHA for this repository. In this the two previous PRs, #4918 and #4919, have slowly ported chunks of the dependency workflows over. This PR now provides the deployment support to ship changes from GHA.

# Key Changes

## Addition of Deployment Workflow

**Concurrency**: One queue per branch. Does not cancel previous runs.

The deployment workflow is triggered only on `develop` and `master` commits once the tests have completed. First, the status of the tests is checked before proceeding with any work. If any failures happened, this workflow does not do anything.

First, the "Tests" workflow for the associated commit is waited on. About 12 minutes (overhead for network slowness) is allocated, 14 total runtime minutes on the job before it is force killed. If the workflow was successful, the deployment for the branch in question begins. If any other conclusion is found, or none at the end of the timer, the job exits with an error so no deployment occurs.

If the branch is `develop`, execution runs autonomously to build and test the package before publishing. Once all is successful, a publish of the next tag is conducted.

If the branch is `master`, execution starts with the `prod-hold` job. This is a job that accesses an environment that is defined with "Required Reviewers". Meaning it will not allow the job to continue until someone on that reviewer list has approved it to happen.

Once permission is granted, the `prod-deploy` job kicks off. Conducting the same actions as the next release does, but no special tags on the version which makes it `latest` and stable. Once the publish is out, some key information is retrieved as output for the job.

Once a successful `prod-deploy` has happened, two jobs kick off. One creates the GitHub Release, which uses the same script as was used in CircleCI. The other sets up NodeJS, waits for the package to be visible on npm if it isn't already, then installs the package globally. After global installation finishes, a few tests of the package that was deployed are run to ensure it is functional.

A key part of both workflows, is ensuring the package contents look correct before publishing. To support more expansive testing of the required behavior, a new node script is introduced. Which helps separate the pre-publish and post-publish testing. As well as expands the capability of that test to be more comprehensive.

### New Environment

A new environment is introduced, `production-deploy`. This is restricted to only being accessible from the `master` branch. It is configured with required reviewers. So *before* the environment can be accessed in a workflow run, someone from the required reviewer group must manually approve the access in the GitHub Actions UX.

This secondary environment is needed since we auto-deploy `next` tags off the `develop` branch. But for `master` with stable deploys, we want to have manual approval before it goes out. Since we can't add the required reviewers to the main environment as that would impact `develop`, this is made to target just stable releases.

## New pre-publish validation script

CircleCI is configured with a shell script that did some pre and post deploy validation of the contents. While it is effective for the basics, it was fairly minimal in what it did as a whole. This now has a far more comprehensive suite of checks that runs for this step.

The new script pulls key data from the package to build a report. It checks that _all_ contents of the `files` definition exist on the filesystem. It then checks to ensure that the package can be imported in CommonJS. After that, ensuring it is importable (along with all importable files shipped) are capable of that in ESM. Finally, it validates the SRI hashes defined in the sri history are what are computed from the current version. The SRI check only occurs on `master` and `release-` branches. As on `develop` the hash is not updated with every change.

This script works by running as a Node ESM script. It pulls key information, does the filesystem check, then links the package to itself. This makes resolving the package pull the linked version on the filesystem instead of the version installed by `@axe-core/webdriverjs` for integration testing.

The output of this is logged to the console and compiled into a more easily viewable step summary. The successful output of which can be seen in [a comment on this PR](https://github.com/dequelabs/axe-core/pull/4929#issuecomment-3497869310).

## Addition of final tests before deployment of stable

One test not added previously, `test_rule_help_version`, which checks for the docs to be live; is now added. As this only impacted stable releases going out, it made sense to reduce the initial scope and introduce this here.

The next test added was an explicit SRI validation. This runs on all commits to `master` and `release-*` branches as well as all PRs targeting them. It uses the old validation pathway for quickly getting the check back in place. We can look into what to do with the SRI stuff itself later.

## Removal of CircleCI Publishing Configuration

Within this patch, the CircleCI configuration to deploy to NPM is removed since we now want to run only from OIDC on GitHub Actions. The full test suite is not removed with this. It will be removed shortly after this lands when we convert the required checks over to GHA for merging.

### Deploy Workflow

<figure><legend>Workflow Control Color Coding</legend>

| Color | Element |
|-------|---------|
| Light Blue - `#e1f5ff` | Start (Trigger) |
| Light Red - `#ffe1e1` | End (No Deploy) |
| Light Green - `#e1ffe1` | End (Success states) |
| Light Yellow - `#fff4e1` | Decision gates & Manual approval |

</figure>

<figure><legend>Step Color Coding</legend>

| Color | Step Name |
|-------|-----------|
| Blue - `#b3d9ff` | Checkout Code |
| Purple - `#d9b3ff` | Install Dependencies |
| Pink - `#ffb3d9` | Build Package |
| Green - `#b3ffb3` | Validate Package |
| Orange - `#ffd9b3` | Publish to NPM |

</figure>

<details><summary>Click to open diagram</summary>

```mermaid
flowchart TD
    Start([Push to master/develop]) --> WaitTests[wait-for-tests Job]
    
    WaitTests --> WT1[Checkout Code]
    WT1 --> WT2[Wait for Tests Workflow]
    WT2 --> TestSuccess{Tests Successful?}
    
    TestSuccess -->|No| End([End - Tests Failed])
    TestSuccess -->|Yes| BranchCheck{Which Branch?}
    
    BranchCheck -->|develop| DeployNext[deploy-next Job]
    BranchCheck -->|master| ProdHold[prod-hold Job]
    
    DeployNext --> DN1[Checkout Code]
    DN1 --> DN2[Install Dependencies]
    DN2 --> DN3[Build Package<br/>npm prepare & build]
    DN3 --> DN4[Determine Prerelease Version]
    DN4 --> DN5[Bump Version]
    DN5 --> DN6[Validate Package]
    DN6 --> DN7[Publish to NPM<br/>--tag=next]
    
    DN7 --> ValidateNextDeploy[validate-next-deploy Job]
    
    ValidateNextDeploy --> VND1[Checkout Code]
    VND1 --> VND2[Setup Node.js]
    VND2 --> VND3[Wait for Package on NPM]
    VND3 --> VND4[Validate Installation of next]
    VND4 --> EndNext([End - Next Version Validated])
    
    ProdHold --> PH1[Manual Approval Gate]
    PH1 --> ProdDeploy[prod-deploy Job]
    
    ProdDeploy --> PD1[Checkout Code]
    PD1 --> PD2[Install Dependencies]
    PD2 --> PD3[Build Package<br/>npm prepare & build]
    PD3 --> PD4[Validate Package]
    PD4 --> PD5[Publish to NPM<br/>stable version]
    PD5 --> PD6[Get Package Data<br/>version & name]
    
    PD6 --> CreateRelease[create-github-release Job]
    PD6 --> ValidateDeploy[validate-deploy Job]
    
    CreateRelease --> CR1[Checkout Code]
    CR1 --> CR2[Install Release Helper<br/>github-release tool]
    CR2 --> CR3[Download Release Script]
    CR3 --> CR4[Make Script Executable]
    CR4 --> CR5[Create GitHub Release]
    CR5 --> EndRelease([GitHub Release Created])
    
    ValidateDeploy --> VD1[Checkout Code]
    VD1 --> VD2[Setup Node.js]
    VD2 --> VD3[Wait for Package on NPM]
    VD3 --> VD4[Validate Installation of Stable]
    VD4 --> EndValidate([End - Deploy Validated])

    %% Consistent step colors with accessible text
    style DN1 fill:#b3d9ff,color:#001a33,stroke:#0066cc
    style PD1 fill:#b3d9ff,color:#001a33,stroke:#0066cc
    style CR1 fill:#b3d9ff,color:#001a33,stroke:#0066cc
    style WT1 fill:#b3d9ff,color:#001a33,stroke:#0066cc
    style VND1 fill:#b3d9ff,color:#001a33,stroke:#0066cc
    style VD1 fill:#b3d9ff,color:#001a33,stroke:#0066cc
    
    style DN2 fill:#d9b3ff,color:#1a0033,stroke:#6600cc
    style PD2 fill:#d9b3ff,color:#1a0033,stroke:#6600cc
    
    style DN3 fill:#ffb3d9,color:#330011,stroke:#cc0066
    style PD3 fill:#ffb3d9,color:#330011,stroke:#cc0066
    
    style DN6 fill:#b3ffb3,color:#002200,stroke:#00aa00
    style PD4 fill:#b3ffb3,color:#002200,stroke:#00aa00
    
    style DN7 fill:#ffd9b3,color:#331100,stroke:#cc6600
    style PD5 fill:#ffd9b3,color:#331100,stroke:#cc6600
    
    %% Decision/Gate styling
    style Start fill:#e1f5ff,color:#001a33
    style End fill:#ffe1e1,color:#330000
    style EndNext fill:#e1ffe1,color:#002200
    style EndRelease fill:#e1ffe1,color:#002200
    style EndValidate fill:#e1ffe1,color:#002200
    style PH1 fill:#fff4e1,color:#331100
    style TestSuccess fill:#fff4e1,color:#331100
    style BranchCheck fill:#fff4e1,color:#331100
```

</details>

### Validation Script

<figure><legend>Color Coding of Scopes</legend>

| Color | Scope |
|-------|-------|
| Dark Gray - `#1a1a1a` | Setup/Teardown |
| Blue - `#004d99` | File Existence Check |
| Orange/Brown - `#664d00` | CommonJS Compatibility Check |
| Purple - `#4d004d` | Importable Check |
| Teal - `#004d4d` | SRI Hash Validation |
| Gray - `#5c5c5c` | Skipped operations |

</figure>

<figure><legend>Color Coding of States</legend>

| Color | State |
|-------|-------|
| Green - `#2d5016` | Success |
| Red - `#7d1007` | Error/Failure |
| Gray - `#5c5c5c` | Skipped |

</figure>

<details><summary>Click to open diagram</summary>

```mermaid
flowchart TD
    Start([Start Script]) --> Init[Initialize Variables]
    Init --> FileCheck[File Existence Check]
    
    FileCheck --> FileLoop[For each file in pkg.files array]
    FileLoop --> FileExists{File exists?}
    FileExists -->|yes| FilePass[✓ Mark as Found]
    FileExists -->|no| FileFail[✗ Mark as Missing<br/>exitCode++]
    FilePass --> LinkSetup
    FileFail --> LinkSetup
    
    LinkSetup[Create npm link] -->|success| CJS[CommonJS Compatibility Check]
    LinkSetup -->|fail| LinkError[Log error and skip remaining checks]
    
    CJS --> CJSRequire{Require package successful?}
    CJSRequire -->|yes| CJSValidate{Validate export type and version exists?}
    CJSRequire -->|no| CJSFail[✗ CommonJS Failed<br/>exitCode++]
    CJSValidate -->|yes| CJSPass[✓ CommonJS Compatible]
    CJSValidate -->|no| CJSFail
    
    CJSPass --> Import
    CJSFail --> Import
    
    Import[Importable Check] --> ImportMain{Import main package?}
    ImportMain -->|success| ValidateMainVersion{Version property exists?}
    ImportMain -->|fail| ImportMainFail[✗ Not Importable<br/>anyCaught = true]
    
    ValidateMainVersion -->|yes| ImportMainPass[✓ Mark as Importable]
    ValidateMainVersion -->|no| ImportMainFail
    ImportMainPass --> ImportLoop
    ImportMainFail --> ImportLoop
    
    ImportLoop[For each file in pkg.files] --> FileType{File type?}
    FileType -->|skip .txt, .d.ts, folders| CheckAnyCaught
    FileType -->|check resolve| NodeModCheck{Resolves to node_modules?}
    
    NodeModCheck -->|yes| ImportNodeFail[✗ Resolves to node_modules<br/>exitCode++]
    NodeModCheck -->|no| ImportTry{Import successful?}
    
    ImportTry -->|yes| CheckVersion{Has version property?}
    ImportTry -->|no| ImportFileFail[✗ Not Importable<br/>anyCaught = true]
    CheckVersion -->|yes| ImportPass[✓ Importable]
    CheckVersion -->|no| ImportFileFail
    
    ImportPass --> CheckAnyCaught
    ImportFileFail --> CheckAnyCaught
    ImportNodeFail --> CheckAnyCaught
    
    CheckAnyCaught{anyCaught true?} -->|yes| ImportExit[exitCode++]
    CheckAnyCaught -->|no| SRICheck
    ImportExit --> SRICheck
    
    SRICheck[SRI Hash Validation] --> BranchCheck{Branch is master or release-*?}
    BranchCheck -->|no| SkipSRI[Skip SRI validation]
    BranchCheck -->|yes| SRILoad[Load sri-history.json<br/>Calculate hashes for axe.js and axe.min.js]
    
    SRILoad --> SRICompare{Hash matches expected?}
    SRICompare -->|yes| SRIPass[✓ Valid SRI]
    SRICompare -->|no| SRIFail[✗ Invalid SRI<br/>exitCode++]
    
    SRIPass --> Cleanup
    SRIFail --> Cleanup
    SkipSRI --> Cleanup
    
    Cleanup[Unlink npm package] --> End([Exit with exitCode])
    LinkError --> CleanupError[Attempt unlink if needed]
    CleanupError --> End
    
    style Start fill:#1a1a1a,stroke:#000,stroke-width:2px,color:#fff
    style Init fill:#1a1a1a,stroke:#000,stroke-width:2px,color:#fff
    style End fill:#1a1a1a,stroke:#000,stroke-width:2px,color:#fff
    
    style FileCheck fill:#004d99,stroke:#000,stroke-width:2px,color:#fff
    style FileLoop fill:#004d99,stroke:#000,stroke-width:2px,color:#fff
    style FileExists fill:#004d99,stroke:#000,stroke-width:2px,color:#fff
    style FilePass fill:#2d5016,stroke:#000,stroke-width:2px,color:#fff
    style FileFail fill:#7d1007,stroke:#000,stroke-width:2px,color:#fff
    
    style LinkSetup fill:#1a1a1a,stroke:#000,stroke-width:2px,color:#fff
    style LinkError fill:#7d1007,stroke:#000,stroke-width:2px,color:#fff
    
    style CJS fill:#664d00,stroke:#000,stroke-width:2px,color:#fff
    style CJSRequire fill:#664d00,stroke:#000,stroke-width:2px,color:#fff
    style CJSValidate fill:#664d00,stroke:#000,stroke-width:2px,color:#fff
    style CJSPass fill:#2d5016,stroke:#000,stroke-width:2px,color:#fff
    style CJSFail fill:#7d1007,stroke:#000,stroke-width:2px,color:#fff
    
    style Import fill:#4d004d,stroke:#000,stroke-width:2px,color:#fff
    style ImportMain fill:#4d004d,stroke:#000,stroke-width:2px,color:#fff
    style ValidateMainVersion fill:#4d004d,stroke:#000,stroke-width:2px,color:#fff
    style ImportMainPass fill:#2d5016,stroke:#000,stroke-width:2px,color:#fff
    style ImportLoop fill:#4d004d,stroke:#000,stroke-width:2px,color:#fff
    style FileType fill:#4d004d,stroke:#000,stroke-width:2px,color:#fff
    style NodeModCheck fill:#4d004d,stroke:#000,stroke-width:2px,color:#fff
    style ImportTry fill:#4d004d,stroke:#000,stroke-width:2px,color:#fff
    style CheckVersion fill:#4d004d,stroke:#000,stroke-width:2px,color:#fff
    style CheckAnyCaught fill:#4d004d,stroke:#000,stroke-width:2px,color:#fff
    style ImportExit fill:#4d004d,stroke:#000,stroke-width:2px,color:#fff
    style ImportPass fill:#2d5016,stroke:#000,stroke-width:2px,color:#fff
    style ImportMainFail fill:#7d1007,stroke:#000,stroke-width:2px,color:#fff
    style ImportFileFail fill:#7d1007,stroke:#000,stroke-width:2px,color:#fff
    style ImportNodeFail fill:#7d1007,stroke:#000,stroke-width:2px,color:#fff
    
    style SRICheck fill:#004d4d,stroke:#000,stroke-width:2px,color:#fff
    style BranchCheck fill:#004d4d,stroke:#000,stroke-width:2px,color:#fff
    style SRILoad fill:#004d4d,stroke:#000,stroke-width:2px,color:#fff
    style SRICompare fill:#004d4d,stroke:#000,stroke-width:2px,color:#fff
    style SRIPass fill:#2d5016,stroke:#000,stroke-width:2px,color:#fff
    style SRIFail fill:#7d1007,stroke:#000,stroke-width:2px,color:#fff
    style SkipSRI fill:#5c5c5c,stroke:#000,stroke-width:2px,color:#fff
    
    style Cleanup fill:#1a1a1a,stroke:#000,stroke-width:2px,color:#fff
    style CleanupError fill:#1a1a1a,stroke:#000,stroke-width:2px,color:#fff
```

</details> 

Fixes: #4912

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -80,7 +80,7 @@ jobs:
       - name: Bump version
         env:
           NEW_VERSION: ${{ steps.determine-version.outputs.version }}
-        run: npm version "$NEW_VERSION" --no-git-tag-version
+        run: npm version "$NEW_VERSION" --no-git-tag-version --ignore-scripts
       - &validate-package
         name: Validate package is consumable
         env:


.eslintignore
Garbee, line 1: This is no longer used with flat config files. Had console warnings while I was debugging some issues in the eslint setup. So removed this to quiet things down and kept it since it doesn't hurt anything to remove now.

eslint.config.js
```diff
@@ -363,6 +363,21 @@ module.exports = [
...
     }
   },
+  {
+    files: ['.github/bin/*.mjs'],
+    languageOptions: {
+      ecmaVersion: 'latest',
+      sourceType: 'module',
+      globals: {
+        ...globals.node,
+        ...globals.es2022
```
Garbee, line 373: Just being conservative here since we run on node LTS builds. We can freely bump as we hit a need later for a new global.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,165 @@
...
+  create-github-release:
+    needs: prod-deploy
+    if: ${{ needs.prod-deploy.result == 'success' }}
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: write
+    steps:
+      - *checkout
+      - name: Install Release Helper
+        run: go get gopkg.in/aktau/github-release.v0
```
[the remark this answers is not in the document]
Garbee: Just confirmed this update is still installing correctly and named right so the pulled script should pick it up.

.github/workflows/nightly-tests.yml
```diff
@@ -32,6 +34,7 @@ jobs:
           start-xvfb: ${{ env.DISPLAY }}
       - &build
         name: Build
+        id: build
```
Garbee, line 37: I forgot to include this id when making an update in the last PR. My editor only started showing me the warning when trying to do the merge conflict resolution.

kelvin7891:
> @copilot review

.github/bin/validate-package.mjs
```diff
@@ -0,0 +1,445 @@
...
+      exitCode++;
+    }
+
+    if (axe.version !== pkg.version) {
+      console.error(
+        `✗ ${pkg.name} version mismatch: expected ${pkg.version}, got ${axe.version}`
+      );
+      summary += `| \`${pkg.name}\` version | ✗ Version Mismatch |\n`;
+      exitCode++;
+    }
```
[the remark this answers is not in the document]
Garbee: Solved this with a new check to throw an error when the version doesn't exist. So the catch clause will do the appropriate work.

.github/bin/validate-package.mjs
```diff
@@ -0,0 +1,445 @@
...
+
+  const importTargets = [...pkg.files.map(file => `${pkg.name}/${file}`)];
+  let anyCaught = false;
+
+  console.log('Validating package files are importable:');
+
+  try {
+    const axe = await import(pkg.name);
+    console.info(`✓ ${pkg.name}`);
+    summary += `| \`${pkg.name}\` | ✓ Importable | ${axe.default.version} |\n`;
```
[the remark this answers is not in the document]
Garbee: Solved with a guarding thrown error. Like the previous recommendation in this area.

.github/bin/validate-package.mjs
```diff
@@ -0,0 +1,445 @@
...
+    }
+
+    try {
+      let version = '';
+      if (target.endsWith('.json')) {
+        const data = await import(target, { with: { type: 'json' } });
+        // Ensure version is empty since it doesn't apply to the json file.
+        // This keeps the table aligned and prevents it from showing a version
+        // when it might be set from a previous loop iteration.
+        version = Object.keys(data.default).at(-1);
```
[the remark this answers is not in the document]
Garbee: The comment was confusing. Just removed it. The code works by pulling the last JSON version known to the file.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,263 @@
...
+      - &validate-package
+        name: Validate package is consumable
+        env:
+          # Ref: https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#runner-context
+          # Linting shows this context might be invalid, but it shouldn't be per docs.
+          # Probably something missing in the schema.
+          DEBUG: ${{ runner.debug == '1' }}
+        run: node .github/bin/validate-package.mjs
+      - name: Publish "next" version to npm
+        run: npm publish --tag=next
```
straker, line 93: The job should run the `next-release` package script before publishing in order to bump the version in package.json

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,263 @@
...
+      - *checkout
+      - name: Install Release Helper
+        run: go install gopkg.in/aktau/github-release.v0@latest
+      - name: Download Release Script
+        run: curl https://raw.githubusercontent.com/dequelabs/attest-release-scripts/develop/src/node-github-release.sh -s -o ./node-github-release.sh
+      - name: Make Release Script Executable
+        run: chmod +x ./node-github-release.sh
+      - name: Create GitHub Release
+        run: ./node-github-release.sh
+  validate-deploy:
```
straker, line 172: Can this step somehow use a shared step to do the validation? Right now it's defined twice so it'd be nice if we could only define it once.

.github/workflows/nightly-tests.yml
straker, line 37: We probably should have done this in a separate pr

.github/bin/validate-package.mjs
```diff
@@ -0,0 +1,456 @@
...
+  summary += `\n## Subresource Integrity Check
+
+This check validates the current build against the SRI hash
+for the version defined in \`sri-history.json\`.
+
+| File | Status |
+|------|--------|
+`;
+
+  const sriHistory = await import(`${pkg.name}/sri-history.json`, {
```
straker, line 321: Is there a reason you're doing all this manually instead of using the npm script `sri-validate`?

.github/workflows/deploy.yml
Garbee: Ah, good catch on that being missed.

.github/workflows/deploy.yml
Garbee: It's done twice explicitly to avoid cloning the repo and causing more work on the target drive. Since once we clone it, we start to add complexity on making sure the scripts run without a local install causing issues.

.github/bin/validate-package.mjs
Garbee: The implementation here just helps us get better output in the summary log. Without calling another subprocesses and updating the previous script, which itself was marked as deprecated (one reason I did a fresh implementation to begin with.)

Ultimately, it looks like since the SRI history is deprecated we may want to remove it entirely some day. So ultimately the original `sri-update.js` file and these checking scripts would get removed.

.github/workflows/deploy.yml
Garbee: Looking at it, that `next-version.js` script is hard coded into Circle CI. I can fix this in a quick way just from the shell using the data we have from CI. Almost like how I did the version checks in Watcher for continually releasing that.

.github/workflows/nightly-tests.yml
Garbee: Normally yes, but since CircleCI is still operating while we get this transition done, this wasn't as critical to rush in and cause more CI time and resources just over it.

.github/workflows/deploy.yml
Garbee: Fixed this by adding a step before to determine and bump the version, instead of pulling it after. Then that is used to bump the versions directly, since we don't use the changelog from `standard-version` with this anyways.

The old script is removed since they are no longer used as this PR removed the CircleCI flow that would call it.

.github/bin/validate-package.mjs
Garbee: Oh, and that old script is only called from Circle CI now. So once we remove the circle ci config it will get removed as well.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,276 @@
...
+      name: registry.npmjs.org
+    permissions:
+      contents: read
+      id-token: write # Required for OIDC
+    outputs:
+      version: ${{ steps.get-data.outputs.version }}
+      packageName: ${{ steps.get-data.outputs.name }}
+    runs-on: ubuntu-24.04
+    steps:
+      - *checkout
```
dbjorge, line 143: Because you're using `workflow_run` and not `push` as the trigger, I don't think this is going to do what you expect; I think `workflow_run` is going to use the most recent commit on the default branch at the time this workflow is started, not "the same commit that the triggering Tests workflow used" like I think you're intending here. I think this is going to pick up the wrong branch in the prod deploy, but more subtly, it also risks picking up the wrong commit from the right branch in the event that PR A merges to develop, and then while its tests are running, PR B merges to develop.

You could pick up the commit to use from the triggering workflow, but that substantially increases the risks associated with using a `workflow_run` trigger type, since it makes it much more likely that some bug in the checks we use could result in a bogus trigger that points to a maliciously-crafted commit. I think what might be preferable would be to switch to using a `push` trigger, and handle "waiting for tests before deployment" by using a job that polls until a successful Tests workflow finishes for the same commit. That will waste a few build minutes compared to only triggering after the workflow_run is triggered, but we don't pay for build minutes in this repo anyway since it's public; I think wasting a few build-minutes per develop commit is a better tradeoff than risking the insecure `workflow_run` trigger.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,276 @@
...
+          # Linting shows this context might be invalid, but it shouldn't be per docs.
+          # Probably something missing in the schema.
+          DEBUG: ${{ runner.debug == '1' }}
+        run: node .github/bin/validate-package.mjs
+      - name: Publish "next" version to npm
+        run: npm publish --tag=next
+  validate-next-deploy:
+    name: Validate Next Deployment
+    needs: deploy-next
+    if: ${{ needs.deploy-next.result == 'success' }}
```
dbjorge: This is already the default behavior of needs relationships between jobs, you don't need to specify this (here or in most of the similar syntax throughout the file)

(the exception is the similar check on L30 against the workflow_run event, which *is* necessary)

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,276 @@
...
+          npm run build
+      - name: Determine prerelease version
+        id: determine-version
+        run: |
+          echo "::group::Determining new prerelease version"
+          NAME=$(npm pkg get name | tr -d '"')
+          LATEST_VERSION=$(npm pkg get version | tr -d '"')
+          echo "Latest version in package.json: $LATEST_VERSION"
+
+          # add 8 characters from the current SHA as prerelease identifier
```
dbjorge: The circleci version used 7 chars, not 8. We shouldn't change the format without a good reason

.github/workflows/deploy.yml
dbjorge: I agree with Steve, and generally would prefer to keep scripts like this in their own files rather than embedded in GHA yaml; CI systems come and go every few years, I'd rather have the actual work/tests be in CI-system-independent scripts where feasible.

> It's done twice explicitly to avoid cloning the repo and causing more work on the target drive.

"causing more work on the target drive" represents one build-machine-second, it's completely negligible compared to the developer overhead of maintaining 2 copies of the script and risking forgetting to update both if someone needs to fix a bug.

> once we clone it, we start to add complexity on making sure the scripts run without a local install causing issues.

I don't think this really adds a problematic amount of complexity, just run the script from a different `cwd` than the place the repo is cloned to.

.github/bin/validate-package.mjs
```diff
@@ -0,0 +1,456 @@
...
+      console.error(`✗ ${pkg.name} exported invalid value`);
+      summary += `| \`${pkg.name}\` export | ✗ Invalid export |\n`;
+      exitCode++;
+    }
+
+    if (!axe.version) {
+      throw new Error('Missing version property');
+    }
+
+    if (axe.version !== pkg.version) {
```
dbjorge: In develop, this script looks like it's going to run *after* pkg.version is updated with the canary SHA suffix, but axe.version is still going to be the built version without that suffix, so I'd expect this to fail, no? I think axe.version still reflecting a non-canary version identifier matches the old circleci behavior, so I think we should accept the old behavior as-is for the purposes of this PR.

dbjorge (changes requested):

.github/workflows/deploy.yml
Garbee: Ah, good catch on the `workflow_run` difference. I didn't realize from my reading the checkout acted that way.

The thing is, `on: push` we can't restrict to after the tests finish and pass. I might need to do a `worklow_call` instead so the tests can call this when everything is ready. Let me look into the impact of that change-up and how it might interfere with OIDC. I remember some people in one of the conversion threads saying `workflow_call` introduced problems, but it could have also been their specific setup was an issue. I don't know if that event inherently changes something about the generated token that makes NPM not trust it.

.github/workflows/deploy.yml
Garbee: I was just being extra cautious with the series of expectations given how critical this file is and the nature of `workflow_run` being in play before. But, if the triggering event change-up goes well, then most of these conditionals might be moot. So I'll drop them if that's the case.

.github/workflows/deploy.yml
Garbee: It was 0 to 7, meaning 8 characters. I'll double check what is on NPM once the Cloudflare outage is resolved and if it was pulling 7 to submit then it's an easy update to make.

.github/bin/validate-package.mjs
Garbee: I can move the version update *before* the build step. I didn't realize the build would bundle in the version as a hardcoded string. I thought we'd resolve up and query the version dynamically from the package.json.

Now, whether we want this to hard fail validation right now, that's as simple as removing the exitCode increment and related materials.

But, that does introduce complexities for anyone running the next builds for testing. So we don't want to introduce a change like that to the versioning.

So, I'm going to leave the ordering how it is, where the main version is fixed in and pull out the error state from this branch of logic. Instead just leaving it as a report of which NPM version is sending which axe version, no failure state. That way we can visibly see the information in the report if needed, but it isn't blocking anything since we do have this condition going on (which, we can solve later to at least verify the primary version is matching in the two areas.)

.github/workflows/deploy.yml
dbjorge: > The thing is, on: push we can't restrict to after the tests finish and pass.

What I'm suggesting is that you use on:push, but begin with a step that polls for the "Tests" workflow to have separately passed against the same commit. I think "workflow_call" has most of the same security concerns as "workflow_run" would; fundamentally, we want the workflow to trigger based on an event that GitHub itself triggers based on actions that require contributor access to the repo, not events that could feasibly be triggered based on activity in unprotected branches (like the ones dependabot uses).

.github/workflows/deploy.yml
dbjorge: The old version was `CIRCLE_SHA1.substr(0, 7)` - `substr` takes `startIndex, length`, not `startIndex, endIndexInclusive`. You can confirm this at https://www.npmjs.com/package/axe-core?activeTab=versions

Garbee: Yup, now that NPM is back up I see that. I'll get it fixed when I'm back in the code. Once I've figured out how to tackle the workflow trigger.

.github/workflows/deploy.yml
Garbee: Yea, looks like polling is the only way. `workflow_call` (regardless of even the potential security differences) can't work because the calling workflow would not be seen as `deploy.yml` in the OIDC token. It is coded to the *caller*. Which we absolutely don't want the test workflow to be allowed to deploy directly.

`workflow_run` is the pertinent way (with some massaging on context) to have it run "independently" as the previous finishes. No timing sequence issues. But, there is complexity to this.

`push` is going to need a starting step to just wait for the GH CLI to report the target sequence of tests and E2E tests are finished and succeed before starting. So we waste time just waiting on the box, but it is what it is given the situation that everyone is stuck in.

.github/workflows/deploy.yml
Garbee: So, I figured out the clean solution while doing the update just now... Don't install the package deps. Duh. Then you don't have any "local" install to cause issues at all.   We just have the scripts and they can move out of the repo space no problem.

Garbee:
So, I have updates ready locally but since GitHub is having issues with git operations, I can't push right now. Only thing left for tomorrow to implement is the conversion from `workflow_run` to `push`. I don't think that should take _too_ long. So I'll give it 30 minutes in the morning and then either push all the needed changes or just push what I have while I finish up that triggering event conversion.

.github/workflows/deploy.yml
```diff
@@ -25,9 +16,31 @@ permissions:
   contents: read
 
 jobs:
+  wait-for-tests:
```
dbjorge, line 26: I do agree that the `workflow_run` version looked like a more obvious solution, so I think it'd be a good idea to leave a comment explaining to a future maintainer why we aren't using it even though it seems more natural:

```suggestion
  # It's important that we use a on: push trigger and poll for Tests to pass,
  # and *not* a workflow_run trigger against the Tests workflow, to enable
  # GitHub to guarantee that *only* protected branch commits can trigger
  # this workflow and to enforce environment protection rules accordingly.
  wait-for-tests:
```

.github/bin/validate-package.mjs
```diff
@@ -0,0 +1,441 @@
...
+      console.error(`✗ ${target}`);
+      summary += `| \`${target}\` | ✗ Not Importable | Not Found |\n`;
+      summary += `\n\`\`\`\n${error.message}\n\`\`\`\n`;
+      anyCaught = true;
+    }
+  }
+
+  if (anyCaught) {
+    exitCode++;
+  }
```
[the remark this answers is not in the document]
Garbee: When any are caught we only want one increment. This is fine.

.github/bin/validate-package.mjs
```diff
@@ -0,0 +1,441 @@
...
+
+        version = axe.default.version;
+      }
+      console.info(`✓ ${target}`);
+      summary += `| \`${target}\` | ✓ Importable | ${version} |\n`;
+    } catch (error) {
+      console.error(`✗ ${target}`);
+      summary += `| \`${target}\` | ✗ Not Importable | Not Found |\n`;
+      summary += `\n\`\`\`\n${error.message}\n\`\`\`\n`;
+      anyCaught = true;
```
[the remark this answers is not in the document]
Garbee: Newlines are in place. So it wont' be immediately after.

.github/bin/wait-for-workflow-success.sh
```diff
@@ -0,0 +1,58 @@
...
+# This gives up to 12 minutes for the workflow to complete.
+
+SLEEP_SECONDS=${SLEEP_SECONDS:-30}
+MAX_ATTEMPTS=${MAX_ATTEMPTS:-24}
+
+attempt=0
+
+while [ $attempt -lt $MAX_ATTEMPTS ]; do
+  # Get the most recent workflow run for this SHA and workflow name
+  # Workflow runs are returned in descending order by created_at (most recent first)
```
dbjorge: GitHub's API docs don't document any guarantee of the sort order; if the API owner doesn't document the behavior, we shouldn't assume this is guaranteed just because it happened to use that order when we tried it.

.github/bin/wait-for-workflow-success.sh
```diff
@@ -0,0 +1,58 @@
...
+  workflow_data=$(gh api "repos/$REPOSITORY/actions/runs?head_sha=$SHA" \
```
dbjorge: We should filter this on the branch it tried to run on, too, since there are jobs in the Tests workflow that only run on only *some* branches; if a fast-forward merge happens such that a PR's last commit and the commit in the target branch have the same SHA, we want this to wait for the Tests workflow that gets triggered against the target branch, not to pass immediately based on the previous Tests workflow that ran in the PR. (you'll need to pipe in the $BRANCH as an env var)

```suggestion
  workflow_data=$(gh api "repos/$REPOSITORY/actions/runs?head_sha=$SHA&branch=$BRANCH&exclude_pull_requests=true" \
```

.github/workflows/deploy.yml
dbjorge: No, this should use .nvmrc. We don't want "while a prod deploy job is running" to ever be the first time we run a test against a new node version; in the deploy job, we want repeatability and a guarantee that what we already tested against is what we're deploying against.

Testing against a floating node_version *is* a good idea, just not here in deploy.yml; this would be better suited to a `nightly-tests` job that tests against a matrix of floating lts and latest versions, similar to how we do nightly tests against floating "latest" browser versions.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,179 @@
...
+          NEW_VERSION: ${{ steps.determine-version.outputs.version }}
+        run: npm version "$NEW_VERSION" --no-git-tag-version
+      - &validate-package
+        name: Validate package is consumable
+        env:
+          # Ref: https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#runner-context
+          # Linting shows this context might be invalid, but it shouldn't be per docs.
+          # Probably something missing in the schema.
+          DEBUG: ${{ runner.debug == '1' }}
+        run: node .github/bin/validate-package.mjs
```
dbjorge, line 91: I'm fine with these tests running here after bumping the version number, but a lot of what it's testing for would be more likely to be introduced as a regression in a normal PR; perhaps it would make sense to also add it as a job in `test.yml`?

.github/bin/validate-package.mjs
```diff
@@ -0,0 +1,441 @@
...
+try {
+  // Link the package globally, then update the package
+  // internally to use the linked version.
+  // This is needed because we don't have `exports` defined
+  // yet, so self referencing imports won't work.
+  // We also have a circular dependency on the package.
+  // That means if we try to resolve the import without
+  // linking, it will resolve the version in `node_modules`
+  // from npm.
+  execSync('npm link', execOptions);
```
dbjorge, line 402: This is acceptable for the purposes of not blocking this PR, but I think these tests would be more valuable if they used a strategy of "`npm pack` + set up a test consumer package that refers to the packed tarball", instead of "self-referencing the directory that we built the package from (either via `npm link` or a self-referencing-import)".

The big difference between those strategies is that the `npm pack` approach will result in your test only "seeing" the files within the package that we *actually* publish, and *not* any of the extra source files/etc which we don't pack into the published package. That creates a more realistic testing environment:

- If we miss an entry in `files` that's necessary for importing the package to work, the `pack` version would catch it but the `npm link` version would miss it
- if we accidentally depend on a `devDependency` at runtime, the `pack` version would catch it but the `npm link` version would pass (because we installed dev dependencies in the source directory in order to build the package)

.github/workflows/deploy.yml
dbjorge: (that would also address the issue where the SRI hash validation ought to be running against `release-*` candidate branches, but isn't going to do so after this change)

.github/bin/determine-version.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
```
dbjorge, line 1: All of these scripts should start with `set -eo pipefail` and be shellcheck-clean.

dbjorge (dismissed):

.github/workflows/deploy.yml
Garbee: Do you mind if we try to document this externally for the org as a whole? I want to make a doc with our "General Guidelines" and why we use these patterns. Maybe since this repo is public we do it here as well as in an internal document.

.github/bin/validate-package.mjs
dbjorge: (these are real classes of issues we've been bitten by before, not hypotheticals; in early versions of the axe-watcher package, we really did have both of those categories of issues that we only discovered after publishing when we set up example repos outside of the main watcher repo)

.github/bin/wait-for-workflow-success.sh
Garbee: I tried branch but it failed. Let me double check what I was testing locally later to see if it was a syntax issue with my commands.

Good find on exclude pull requests. This probably negates the situation I thought `head -n 1` was fixing to pop the first off.

.github/bin/wait-for-workflow-success.sh
Garbee: Double checking the docs, yea this endpoint isn't. Combined with the other bit to scope down what we're looking for at the query level, I can probably use `jq` after to sort it internally by the created_at field. Then that will get us what we're looking for.

.github/workflows/deploy.yml
Garbee: Ok, I was just trying to save some seconds by not needing to download a node bundle since this uses something local to the machine.

.github/workflows/deploy.yml
Garbee: This is only here replicating the original workflow, but expanded on the checks being done to get more in the weeds. If we want to make this a normal PR test, I'm fine with that. But should that block the OIDC conversion? Also, if we do make this a regular PR check. I'd rather investigate tools like https://publint.dev but that run against things locally to validate. That way there is a community tool with more knowledge behind it (and hopefully more concise output too.)

.github/bin/validate-package.mjs
Garbee: Yea, I'd much rather do testing against an `npm pack` build. It's only this way now because I was trying replicate the original setup, but do more expansive testing as well while I was in it.

I also want to pilot `npm pack` based testing in other internal repositories first to make sure we have proper patterns down before doing it in this repository.

.github/bin/determine-version.sh
Garbee: `set -eo pipefail`, good catch on that... Shellcheck *should* have been running in my editor. I'll double check that when I'm back in it to setup the pipefail.

.github/workflows/deploy.yml
dbjorge: I agree with putting it in both places

Garbee: Ok. I'll work on the comment here a tiny bit. I think the key thing I missed from the GH Docs that we need to call out is `workflow_run` ONLY runs in the context of the default branch. Regardless of which branch it is running _against_. That context piece is what matters and what the GH docs don't make entirely clear.

.github/workflows/deploy.yml
dbjorge: I mentioned it mostly because it's *not* currently quite replicating the original workflow - the part of the validation script which *was* in the original workflow (the SRI validation) ran on some branches that this new setup isn't going to run on (`release-` candidate branches). I think we should aim for at-least-parity in the release-process-only checks we used to run.

Publint is a cool find, I agree that it'd be a good idea for us to investigate (in a separate PR).

.github/workflows/deploy.yml
Garbee: Split this into two comment sections. One on the trigger and another here to specify why we're doing this part.

.github/workflows/deploy.yml
Garbee: Ah, yea I saw the `release-*` filter, but I couldn't trace how it actually executed against that branch set on CircleCI. Since I didn't see any inlet to let it run. I'll double check the CircleCI config after the wait for workflow success updates and see if I can find it. Now that the bulk of the work is out of the way. Maybe it'll be clearer how the trigger occurred on that branch going back to focus on that.

.github/bin/wait-for-workflow-success.sh
Garbee: So, the problem from yesterday when I tried filtering locally. I didn't have the right parameters for *this* endpoint. When I was searching for branch filtering, I ended up in the rest endpoints for PRs so I was using those queries locally and did not catch that I was in the wrong area of docs. Left a reference link to the correct doc section for later while I was in it this morning.

.github/workflows/deploy.yml
Garbee: Pushed that to the tests file. Re-used the old script for now to quickly match parity in the workflow itself without needing to adjust this script to run on PRs yet.

It runs against all PRs targeting the RC and `master` branches, as well as all commits to them. That way if for any reason someone is patching an RC it helps make sure the hash is updated right before the change lands. (Should this flow ever come up for us.)

Issue #4933, opened 2025-11-07
Correct axe-core luminance number


WilcoFiers:
Axe-core uses `0.03928` for its luminance calculation. This number came from WCAG, however WCAG's number was wrong. This was corrected a while back to be `0.04045`. The change is theoretical, the difference is so small that the contrast ratio won't change. But it would be good to use update axe to use the same number that's in WCAG 2.2.

This change needs to happen in lib/commons/color/get-contrast.js

PR #4934, opened 2025-11-07
fix: replaced luminance threshold constant 0.03928 with 0.04045

AdminAdi:
Closes: #4933 

Updates the luminance calculation constant from 0.03928 to 0.04045 in lib/commons/color/get-contrast.js to align with the corrected WCAG 2.2 standard. While the practical impact on contrast ratios is minimal, this change ensures standards compliance with the current WCAG guidelines.


WilcoFiers (approved):
Looks good to me. Thank you for the contribution @AdminAdi 

For the books: I reviewed this PR for security.

PR #4952, opened 2025-12-01
fix(core): stop ignoring shadow roots in context

### LINKED ISSUE #4941: Should it be possible to use a `ShadowRoot` as context?
### Product

axe-core

### Product Version

4.11.0

### Latest Version

#### Expectation + Actual

At the moment, passing a [`ShadowRoot`](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot) instance as `context` to `axe.run()` results in an error: "No elements found for include in page Context".

It's either a bug, or an issue with TypeScript types (see below).

Consider either:
* Supporting a `ShadowRoot` as context (the TS types suggest that context may be of type `Node`, and `ShadowRoot` is a descendant of `Node`).
* Aligning the types with the [documentation](https://www.deque.com/axe/core-documentation/api-documentation/#context-parameter), which mentions that contest should be an "element" (not a "node" — [`Element`](https://developer.mozilla.org/en-US/docs/Web/API/Element) is also a descendant of `Node`).

#### How to Reproduce

CodePen: https://codepen.io/pavelp/pen/QwNEpwv

#### Additional context

* I ran into this due to noticing a bug in [Accented](https://accented.dev), a library that I maintain. There's a `MutationObserver` there, and the nodes that the mutation observer passes to its callback get passed as context to `axe.run()`. The mutation observer may pass a `ShadowRoot` instance there, causing the bug. It will be easy enough to fix it downstream by passing the shadow root's [`host`](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot/host) to `axe.run()` instead of passing the shadow root. However, axe would be a little more robust if it accepted the shadow root directly.
* It's not clear whether such treatment of shadow roots was intentional or not, and how easy that would be to change. I poked at the code a little, and it seems that axe maintains [a cache of all the elements on the page](https://github.com/dequelabs/axe-core/blob/7691bab84afc21a8ec0be07b22e703cd5c69c62b/lib/core/utils/get-node-from-tree.js#L11), and the code that populates it [omits the shadow roots themselves, jumping straight to their children](https://github.com/dequelabs/axe-core/blob/7691bab84afc21a8ec0be07b22e703cd5c69c62b/lib/core/utils/get-flattened-tree.js#L131).

pomerantsev:
Before: if a shadow root (not the shadow host element, but an actual [`ShadowRoot`](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot) instance) is passed as part of context, it's silently ignored.

After: the shadow root node is substituted with its ~host element~ children for the rest of Axe algorithms to work with.

Closes: #4941


straker (changes requested):
Thanks for the pr, it's a great start. For handling shadow roots, instead of adding the host element we should instead add all the children of the shadow root. This way if there are accessibility violations on the host node we do not report them since the shadow root was passed in and not the host node itself. 

We'll also want a test added to https://github.com/dequelabs/axe-core/blob/develop/test/core/base/context.js under the `describe('include')` block that tests that passing a shadow root to the context will add all it's children. You can use `createNestedShadowDom` (and tests that use it) to make creating the shadow DOM element and children easier.

test/core/base/context.js
```diff
@@ -16,15 +16,15 @@ describe('Context', () => {
   it('should not mutate exclude in input', () => {
     fixture.innerHTML = '<div id="foo"></div>';
     const context = { exclude: [['iframe', '#foo']] };
-    // eslint-disable-next-line no-new
```
pomerantsev, line 19: Removed by a pre-commit hook.

pomerantsev:
Fixed format.

straker (changes requested):
Looking good. Thanks for making that change. Last thing we'll want to do is add an integration test to ensure axe-core runs against a shadow root as expected. Just add a test or two to https://github.com/dequelabs/axe-core/blob/develop/test/integration/full/context/context.js, passing in a shadow root and making sure the host doesn't throw any accessibility violations and that the children do. Even though the tests uses `shadowSupported` we don't need to do that anymore as all browsers we test now support shadow DOM, so you can write a test without checking that first.

pomerantsev:
@straker thanks, I added the tests. Let me know if there's anything else left to address.

straker (approved):
Sorry for the delay in reviewing this, was out on vacation. Everything's looking good

Issue #4894, opened 2025-09-30
Graceful handling of color contrast check fails for oklch and oklab with `none`


pattonwebz:
### Product

axe-core

### Product Version

4.10.3

### Latest Version

- [x] I have tested the issue with the latest version of the product

#### Expectation
Color contrast checks return results for issues on the page, or partial results in event of an error with a specific color.

#### Actual
Color contrast checks return an error with no results.

#### How to Reproduce
A page with CSS like this produces the issue: `oklch(0.961073 0.000047911 none / 0.2)`

#### Additional context
It seems the source of this is ColorJS treats `none` as `NaN` in the version used in axe-core. They have a beta that resolves this in the v6 line that handles none and NaN properly so those colors don't fail.

Related issue:
#4269

namansancheti:
@straker 
I see, thanks for the detailed walkthrough on the required changes, in any case 👍

chutchins25:
Should have posted in this sooner but I picked this up this morning and have begun looking into it.

PR #4959, opened 2025-12-05
fix: color contrast fails for oklch and oklab with none

chutchins25:
Adds error handling for color-contrast checks. This now will report an 'incomplete' error message when the foreground and/or background color strings were not able to be parsed into an actual color.

Tests also verify that text-shadow color is handled should that occur.

fixes: [4894](https://github.com/dequelabs/axe-core/issues/4894)

---

Developer Notes:
I was not able to have a test case trigger the colorParse error on text-shadow colorParse issues. This case was getting caught already but in the form of the 'complexTextShadows' error.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -44,12 +44,16 @@ updates:
       # @see https://github.com/dequelabs/axe-core/issues/4428
       - dependency-name: 'colorjs.io'
         versions: ['>0.4.3']
-      # Still need to support node 18
+      # Still need to support node 18 in our tests
       - dependency-name: 'glob'
         versions: ['>=11.0.0']
       # Use node 4 types for backward compatibility
       - dependency-name: '@types/node'
         versions: ['>=5.0.0']
+      # Breaking change that we need to handle in its own pr first
+      # @see https://github.com/dequelabs/axe-core/pull/4264
+      - dependency-name: 'css-selector-parser'
+        versions: ['>=2.0.0']
     groups:
       # Any updates not caught by the group config will get individual PRs
       npm-low-risk:
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
         "browser-driver-manager": "1.0.4",
         "chai": "^4.3.7",
         "chalk": "^4.x",
-        "chromedriver": "*",
+        "chromedriver": "latest",
         "clean-jsdoc-theme": "^4.2.17",
         "clone": "^2.1.2",
         "colorjs.io": "0.4.3",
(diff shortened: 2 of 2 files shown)


lib/commons/color/color.js
```diff
@@ -172,7 +172,9 @@ export default class Color {
       // color.alpha is a Number object so convert it to a number
       this.alpha = +color.alpha;
     } catch {
-      throw new Error(`Unable to parse color "${colorString}"`);
+      throw new Error(`Unable to parse color "${colorString}"`, {
+        cause: { code: 'COLOR_STRING', value: colorString }
```
straker: Unfortunately axe-core has to still "support" IE11 (even though we deprecated it and stopped testing with it) and I don't believe the Error object supports `cause` in IE11. I tested it out and it doesn't seem to come through so we may not be able to do it this way.

lib/commons/color/get-background-color.js
```diff
@@ -75,9 +75,18 @@ function _getBackgroundColor(elm, bgElms, shadowOutlineEmMax) {
...
-    if (bgColor.alpha === 0) {
-      continue;
+    let bgColor;
+    try {
+      bgColor = getOwnBackgroundColor(bgElmStyle);
+      if (bgColor.alpha === 0) {
+        continue;
+      }
+    } catch (error) {
+      if (error.cause) {
```
straker: We shouldn't assume that any use of the `cause` object means the `colorParse` error. Should probably test for the specific error we expect and then set the `colorParse` data if it matches. That way if we throw a different error in the future that uses `cause` it won't alway sbe `colorParse` message.

lib/commons/color/get-foreground-color.js
```diff
@@ -32,17 +32,25 @@ export default function getForegroundColor(node, _, bgColor, options = {}) {
...
 
-    if (color.alpha === 1) {
-      break;
+      if (color.alpha === 1) {
+        break;
+      }
+    }
+  } catch (error) {
+    if (error.cause) {
+      incompleteData.set('colorParse', error.cause.value);
```
straker: Ditto for here (don't assume `cause` === `colorParse`)

test/checks/color/color-contrast.js
```diff
@@ -15,6 +15,45 @@ describe('color-contrast', function () {
     axe._tree = undefined;
   });
 
+  it('should return undefined if cannot handle color', function () {
```
straker, line 18: Great tests!

lib/commons/color/get-background-color.js
chutchins25: I was using the error.cause object to pass along the color string that caused the error. In color.js, it is given to the original error string but it becomes part of a full string. At the time, I did not see a way to get the color to have displayed in the colorParse data.colorParse message. I will see if there is another way I can capture the original color string.

lib/commons/color/color.js
chutchins25: Changed this to set and send the colorString from color.js using the incompleteData object. This involved introducing the incompleteData object to color.js via import.

chutchins25: It now throws the regular Error object, like it used to.

lib/commons/color/get-background-color.js
chutchins25: Changed this to set and send the colorString from color.js using the incompleteData object. This involved introducing the incompleteData object to color.js via import.

lib/commons/color/get-foreground-color.js
chutchins25: Changed this to set and send the colorString from color.js using the incompleteData object. This involved introducing the incompleteData object to color.js via import.

test/checks/color/color-contrast.js
chutchins25: Added corresponding integration tests!

PR #4960, opened 2025-12-11
chore: update node to 24

Garbee:
This patch updates the configured node version to use 24 for development and CI purposes. Which installs npm@11 by default to fix the deployment process. The developer guide has the new version as well as the eslint config for the newer GitHub scripts to have the right context.

Closes: #4912


straker (changes requested):
Please also update the [developer guide](https://github.com/dequelabs/axe-core/blob/develop/doc/developer-guide.md#environment-pre-requisites) with the node change.

straker (approved):

Issue #4963, opened 2025-12-12
Add support for `isv` as valid language tag


carlosjeurissen:
### Product

axe-core

### Feature Description

Interslavic (with code `isv`), is currently regarded as an invalid language code. Even though it is a valid BCP47 language tag.

https://github.com/dequelabs/axe-core/blob/develop/lib/core/utils/valid-langs.js

carlosjeurissen:
@straker thanks for the update!

PR #4966, opened 2025-12-15
fix(valid-lang): update valid-langs for newer language codes

straker:
Just ran the script in the comment of the file on the latest version of the page to update the data. 

Note: the removal of `eslint-disable-next-line` seems to happen on commit and wasn't something I did.

Closes: #4963


Garbee (approved):
LGTM. Nice comment in this valid-langs to detail what is going on.

Issue #4947, opened 2025-11-26
Contrast Error “Background Color Cannot Be Determined” on Textarea


Karthikeyan-mallow:
### Product

axe-core

### Latest Version

- [x] I have tested the issue with the latest version of the product

#### Expectation

The accessibility scanner should correctly detect the background color of the <textarea> element and validate color contrast according to WCAG 2.1/2.2 guidelines. No contrast error should be raised when the textarea has a valid text and background color.

#### Actual

The tool reports a false positive error:

“Element's background color could not be determined because it's partially obscured by another element.”

This happens even though:

There are no overlapping elements in the layout.

The issue occurs even when testing a standalone <textarea> with no other HTML elements.

Manual WCAG contrast testing shows the textarea meets the required contrast ratio.

#### How to Reproduce

Use the following minimal code snippet:

<body>
    <textarea style="color: #000; background-color: #fff;"></textarea>
</body>

Run any automated accessibility scanner (e.g., Axe, WAVE, Chrome Lighthouse).
The tool incorrectly flags the textarea with the error:
“Background color could not be determined because it's partially obscured.”

#### Additional Context

This appears to be a known false positive in certain WCAG scanning tools when analyzing form controls, especially <textarea>.

Multiple tests confirmed consistent false positives regardless of styling.

Manual inspection validates that the component passes WCAG contrast requirements, so this may be a scanner limitation rather than an actual issue.

WilcoFiers:
@Karthikeyan-mallow Thank you reporting this. This is what axe-core calls a "needs review" result. This is not an accessibility issue, rather it is axe informing you that it wasn't able to test an element, and that it will require manual review. We do this because we don't want people to overlook accessibility issues, assuming axe tested a thing that it did not test. A false positive occurs when something is failed when it shouldn't have been, which this is not.


Nevertheless I would like to look into this. We are always looking for ways to refine and improve these rules, and I appreciate that if you see a lot of these it can get frustrating. This is also why axe Extension has further tools to resolve these color contrast needs reviews, as well as the option to hide all needs review results.

If you can share the page on which you encountered this issue we can investigate the cause and look to see if there is a way we can avoid it.

PR #4968, opened 2025-12-15
fix(color-contrast): do not incomplete on textarea

straker:
Closes: #4947


Garbee:
The description here has no content. We should probably at least have what the patch is doing and how it addresses the problem at a minimum for content. That way as we merge the commit history starts to contain some useful information when going through the history and bisecting problems.

straker:
The description isn't needed if the issue it's closing details all the work, in this case https://github.com/dequelabs/axe-core/issues/4947#issuecomment-3657910693.

WilcoFiers (approved):

Issue #4798, opened 2025-06-12
html-has-lang false positive with Browser MCP


WilcoFiers:
I started noticing this false positive when I had the Browser MCP extension installed:

<img width="382" alt="browser-mcp-container custom element with html in the shadow root" src="https://github.com/user-attachments/assets/cdfbf240-615b-43a1-914a-dadac98a4c28" />

Apparently this injects a custom `browser-mcp-container` element which has its own HTML element. The browser's parser prevents adding multiple HTML elements to the page, and so axe seems to assume that when it queries the `html` element the only thing it could ever possibly match is the root. That's not correct. Instead of selecting the HTML element `html-has-lang` should select `:root`.

I don't think this is just a shadow DOM issue. I suspect you can also `document.createElement('html')` and then append that to the DOM.

WilcoFiers:
This one's even weirder: I somehow got a `frame-tested` needs review against that element, even though it isn't in a frame.

<img width="840" alt="frame-tested error on 'browser-mcp-container,html' element" src="https://github.com/user-attachments/assets/ad364248-2c17-413a-82e1-ef8fcc999864" />

PR #4969, opened 2025-12-16
fix: don't runs rules that select `html` on nested `html` elements

straker:
Our version of `css-selector-parser` doesn't support the `:root` pseudo selector, so had to get a bit creative with the selector. Also updated all rules that used the `html` selector and not just the `frame-tested` and `html-valid-lang` rules as described in the ticket.

Closes: https://github.com/dequelabs/axe-core/issues/4798

### DIFF
--- a/test/integration/full/bypass/header2.html
+++ b/test/integration/full/bypass/header2.html
@@ -24,13 +24,6 @@
     <h1>This header will make the test pass.</h1>
     <a href="http://www.deque.com">stuff</a>
     <div id="mocha"></div>
-    <div id="additional"></div>
-    <script>
-      // only way to get a 2nd html element on the page is to create it with createElement
-      // if you try to do it through the html parser it won't work
-      var root = document.createElement('html');
-      document.getElementById('additional').append(root);
-    </script>
     <script src="/test/testutils.js"></script>
     <script src="pass-tests.js"></script>
     <script src="/test/integration/adapter.js"></script>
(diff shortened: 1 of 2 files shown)


lib/rules/bypass.json
```diff
@@ -1,7 +1,7 @@
 {
   "id": "bypass",
   "impact": "serious",
-  "selector": "html",
+  "selector": "html:not(html *)",
```
Garbee, line 4: I don't think this will work in IE 11. In that the not pseudo-selector only supports basic entry like `:not(html)`, it doesn't support complex or list selectors that I'm aware of. This should be validated before merge, as it could break existing checks being ran in the wild.

WilcoFiers: Doesn't matter. These don't actually get executed. Axe parses them and walks the tree with them.

@straker did you perf test this? This seems like an awfully slow selector.

straker: This isn't slow at all. The way the QSA code works is it matches [each expression in turn](https://github.com/dequelabs/axe-core/blob/develop/lib/core/utils/selector-cache.js#L42). So it starts by matching all `html` elements on the page (an O(1) operation), then it runs the `:not` operator which is also a quick match using the [matchesExpression logic](https://github.com/dequelabs/axe-core/blob/develop/lib/core/utils/matches.js) on any found nodes. This barely adds any time to the selection, and testing this out didn't result in any difference in timings.

On https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Your_first_WebExtension (size of the page does not matter for matching tag selectors), current axe-core average of five runs is 235.76ms. Using this code average of five runs is 220.52ms. This isn't a time savings it's just the variability of the runs. 15ms of time difference is not significant.

WilcoFiers (approved):

Garbee (approved):
Given the context that we don't run the selectors directly in the browser engine, this LGTM.

Issue #4957, opened 2025-12-01
`aria-valid-attr-value` does not correctly handle multiple id tokens for `aria-errormessage` with `aria-describedby`


dbarnespaychex:
As of the latest Axe Core (v4.11.0), combining `aria-errormessage` and `aria-describedby` when using multiple IDREFs fails:

```html
<fieldset>
    <legend>Phone Number</legend>
    <div>
        <input
            type="text"
            name="area-code"
            placeholder="123"
            aria-label="area code" />
    </div>
    <div>
        <input
            type="text"
            name="telephone"
            placeholder="456-7890"
            aria-invalid="true"
            aria-label="telephone"
            aria-describedby="invalid-phonenumber invalid-phonenumber-field"
            aria-errormessage="invalid-phonenumber invalid-phonenumber-field" />
        <div id="invalid-phonenumber">
            <span>Invalid phone number.</span>
        </div>
    </div>
    <div id="invalid-phonenumber-field">
        <span>The field is invalid.</span>
    </div>
</fieldset>
```

Reporting this result:

```json
{
  "id": "aria-valid-attr-value",
  "impact": "critical",
  "tags": [
    "cat.aria",
    "wcag2a",
    "wcag412",
    "EN-301-549",
    "EN-9.4.1.2",
    "RGAAv4",
    "RGAA-7.1.1"
  ],
  "description": "Ensure all ARIA attributes have valid values",
  "help": "ARIA attributes must conform to valid values",
  "helpUrl": "https://dequeuniversity.com/rules/axe/4.11/aria-valid-attr-value?application=axeAPI",
  "nodes": [
    {
      "any": [],
      "all": [
        {
          "id": "aria-errormessage",
          "data": [
            "invalid-phonenumber",
            "invalid-phonenumber-field"
          ],
          "relatedNodes": [],
          "impact": "critical",
          "message": "aria-errormessage values `invalid-phonenumber, invalid-phonenumber-field` must use a technique to announce the message (e.g., aria-live, aria-describedby, role=alert, etc.)"
        }
      ],
      "none": [],
      "impact": "critical",
      "html": "<input type=\"text\" name=\"telephone\" placeholder=\"456-7890\" aria-invalid=\"true\" aria-label=\"telephone\" aria-describedby=\"invalid-phonenumber invalid-phonenumber-field\" aria-errormessage=\"invalid-phonenumber invalid-phonenumber-field\">",
      "target": [
[...]

PR #4973, opened 2025-12-27
fix(aria-valid-attr-value): handle multiple aria-errormessage IDs

JustasMonkev:
When aria-errormessage contains multiple space-separated IDs, the check now correctly verifies that all IDs are present in aria-describedby instead of failing to match the entire string against tokenized values.

Closes: https://github.com/dequelabs/axe-core/issues/4957

### DIFF
--- a/test/integration/rules/aria-valid-attr-value/aria-valid-attr-value.html
+++ b/test/integration/rules/aria-valid-attr-value/aria-valid-attr-value.html
@@ -371,6 +371,18 @@
     aria-valuetext=""
   ></div>
 
+  <div>
+    <input
+      type="text"
+      id="pass192"
+      aria-invalid="true"
+      aria-errormessage="pass192-ref1"
+      aria-describedby="pass192-ref1 pass192-ref2"
+    />
+    <div id="pass192-ref1">Error message 1</div>
+    <div id="pass192-ref2">Error message 2</div>
+  </div>
+
   <div id="ref">Hi</div>
   <div id="ref2">Hi2</div>
 
--- a/test/integration/rules/aria-valid-attr-value/aria-valid-attr-value.json
+++ b/test/integration/rules/aria-valid-attr-value/aria-valid-attr-value.json
@@ -228,7 +228,8 @@
     ["#pass188"],
     ["#pass189"],
     ["#pass190"],
-    ["#pass191"]
+    ["#pass191"],
+    ["#pass192"]
   ],
   "incomplete": [
     ["#incomplete1"],


JustasMonkev:
@straker could you please check again?

straker:
Sorry for the delay. I've been out of office and so haven't had time to look at this. After discussing this with the team here is what we are thinking.

Testing multiple `aria-errormessage` ids with various screen readers and browsers, only JAWS/Chrome supports this. NVDA/Firefox only announces the first id, while VoiceOver/Safari doesn't announce any. This means that even if multiple ids are supported by the spec, only JAWS/Chrome actually support it. When cases such as this happen with ARIA attributes, we [typically report this as "unsupported"](https://github.com/dequelabs/axe-core/blob/develop/lib/checks/aria/aria-unsupported-attr.json#L8) and that it's not widely supported by screen readers.

Taking that into consideration, as well as the various edge cases that would need to be handled to support multiple ids (what happens if the first id isn't on the page but all others are, what happens if the last id is the only one on the page, etc.), we feel it's best to report multiple ids on `aria-errormessage` as unsupported. 

What this would mean is that we would update the `aria-errormessage` check to provide a new incomplete message of "Multiple IDs in aria-errormessage is not widely supported in assistive technologies" and in the evaluate function check if the `tokenList` of `aria-errormessage` contains more than 2 entires, and if so set the `this.data({ message: 'unsupported' })` and return `undefined`.

JustasMonkev:
> Sorry for the delay. I've been out of office and so haven't had time to look at this. After discussing this with the team here is what we are thinking.
> 
> Testing multiple `aria-errormessage` ids with various screen readers and browsers, only JAWS/Chrome supports this. NVDA/Firefox only announces the first id, while VoiceOver/Safari doesn't announce any. This means that even if multiple ids are supported by the spec, only JAWS/Chrome actually support it. When cases such as this happen with ARIA attributes, we [typically report this as "unsupported"](https://github.com/dequelabs/axe-core/blob/develop/lib/checks/aria/aria-unsupported-attr.json#L8) and that it's not widely supported by screen readers.
> 
> Taking that into consideration, as well as the various edge cases that would need to be handled to support multiple ids (what happens if the first id isn't on the page but all others are, what happens if the last id is the only one on the page, etc.), we feel it's best to report multiple ids on `aria-errormessage` as unsupported.
> 
> What this would mean is that we would update the `aria-errormessage` check to provide a new failure message of "Multiple IDs in aria-errormessage is not widely supported in assistive technologies" and in the evaluate function check if the `tokenList` of `aria-errormessage` contains more than 2 entires, and if so set the `this.data({ message: 'unsupported' })` and return `false`.

Thanks for the comment, updated it

test/integration/rules/aria-valid-attr-value/aria-valid-attr-value.html
```diff
@@ -90,6 +90,27 @@ <h2>Violations</h2>
...
     <div id="violation46-ref" aria-hidden="true">Error message 1</div>
   </div>
+
+  <!-- Multiple IDs in aria-errormessage are not supported -->
+  <div>
+    <input
+      type="text"
+      id="violation47"
+      aria-invalid="true"
+      aria-errormessage="violation47-ref1 violation47-ref2"
```
straker, line 100: Thanks for adding the tests, just a small change should bring this across the finish line. Thanks for taking the time to work on this!

Since `aria-errormessage` can fail if the IDs are not part of `aria-describedby` or other techniques, we'll want to make sure the code would normally pass without multiple IDs so that it's only the addition of the multiple ids that causes the failure.

PR #4974, opened 2025-12-30
fix(aria-allowed-attr): restrict br and wbr elements to aria-hidden only

nami8824:
Closes: https://github.com/dequelabs/axe-core/issues/3177

### DIFF
--- a/lib/checks/aria/aria-allowed-attr-elm-evaluate.js
+++ b/lib/checks/aria/aria-allowed-attr-elm-evaluate.js
@@ -1,5 +1,5 @@
-import { validateAttr, getExplicitRole } from '../../commons/aria';
-import { getElementSpec } from '../../commons/standards';
+import { getExplicitRole } from '../../commons/aria';
+import { getElementSpec, getGlobalAriaAttrs } from '../../commons/standards';
 
 export default function ariaAllowedAttrElmEvaluate(node, options, virtualNode) {
   const elmSpec = getElementSpec(virtualNode);
@@ -16,10 +16,14 @@ export default function ariaAllowedAttrElmEvaluate(node, options, virtualNode) {
   }
 
   const { allowedAriaAttrs } = elmSpec;
+  const globalAriaAttrs = getGlobalAriaAttrs();
   const invalid = [];
 
   for (const attrName of virtualNode.attrNames) {
-    if (validateAttr(attrName) && !allowedAriaAttrs.includes(attrName)) {
+    if (
+      globalAriaAttrs.includes(attrName) &&
+      !allowedAriaAttrs.includes(attrName)
+    ) {
       invalid.push(attrName);
     }
   }
--- a/test/checks/aria/aria-allowed-attr-elm.js
+++ b/test/checks/aria/aria-allowed-attr-elm.js
@@ -58,6 +58,16 @@ describe('aria-allowed-attr-elm', () => {
     });
   });
 
+  it('should pass for br with non-global aria attribute', () => {
+    const vNode = queryFixture('<br aria-expanded="true" id="target">');
+
+    assert.isTrue(
+      axe.testUtils
+        .getCheckEvaluate('aria-allowed-attr-elm')
+        .call(checkContext, null, null, vNode)
+    );
+  });
+
   it('should pass for br with explicit role', () => {
     const vNode = queryFixture(
       '<br role="heading" aria-busy="true" id="target">'


lib/checks/aria/aria-allowed-attr-evaluate.js
```diff
@@ -29,7 +29,7 @@ import { isFocusable } from '../../commons/dom';
 export default function ariaAllowedAttrEvaluate(node, options, virtualNode) {
   const invalid = [];
   const role = getRole(virtualNode);
-  let allowed = allowedAttr(role);
+  let allowed = allowedAttr(role, { vNode: virtualNode });
```
WilcoFiers: I'm a bit reluctant about doing it this way for two reasons.

1. I think this is mixing node constraints with role constraints. I don't think this function should give both the allowed attributes based on roles, and the allowed attrs based on node names. I think we should probably have a separate method for something like this. 
 
2. I think we should have role restrictions like this available on the standards object, so that it is configurable and findable, instead of having an inline exception for this case.

The way you've implemented this will ignore the role. I don't think that's the right thing to do here. A `<br>` with role=heading should allow all ARIA properties allowed on headings. Axe has a separate rule that reports using role=heading on br's as an issue. That way we keep the issues more clearly separated.

I appreciate this is going to make this work a little more complicated. @straker would you mind weighing in here too? I think it might be good to maybe propose a structure for the standards object before asking for more changes.

WilcoFiers (dismissed):
@nami8824 Thank you very much for this contribution! It is greatly appreciated.

lib/checks/aria/aria-allowed-attr-evaluate.js
nami8824: Thank you for the feedback.

**Regarding my original approach:** the issue I was addressing was specifically about restricting ARIA attributes on certain node types (`br` / `wbr`), rather than on roles. Because of that, I initially thought it was necessary to pass the node information into `allowedAttr` in order to enforce node-level constraints.

**About the standards object:** when you mention moving these restrictions there, are you referring to `lib/standards/html-elms.js`? If so, would supporting this require changing the data structure in a way that affects other elements as well, or is it expected to be handled only for `br` / `wbr`?

straker: I agree we could have the aria restrictions in the `htmlElms` standards, but that logic becomes a bit tricky in `aria-allowed-attr`. The way I see it working would be in the `allowed-attr` function if the element doesn't have a role, we check if the `htmlElms` standard has any restrictions on attrs and then remove them from the array before we return it. But I don't think that's the correct way to go as it makes the final failure message of not allowing it muddy since the element does not have a role (and in fact has nothing to do with the role).

We could probably move the logic to the `aria-prohibited-attr` check, similar to our `aria-label` and `aria-labelledby` checks on nodes with no role, but again that is a little difficult since we need to list all attributes as prohibited except `aria-hidden`. @dbjorge what do you think?

dbjorge: I don't like the idea of listing all attributes as prohibited, I'd rather keep the standards object closer to the actual standard (the wording of the relevant table header in ARIA in HTML is "ARIA role, state and property allowances"). I agree that it'd be good to keep the failure message clear about whether the issue is coming from a role-based or element-based restriction. I think based on the rule naming and our existing rule docs, it'd make the most sense for this new type of violation to appear under `aria-allowed-attr` and not `aria-prohibited-attr`.

I'd lean towards something like:

1. Update the `html-elms` shape to add a new property for `allowedAriaAttrs` - array(optional). If specified, restricts which ARIA attributes may be used with this element (*in addition to* any role-based restrictions).`
2. Keep using the `aria-allowed-attr` rule, but split the `aria-allowed-attr` check into `aria-allowed-attr-role` and `aria-allowed-attr-elm` (both in the rule's `all` checks)
3. The role one functions like the current check. The elm one checks if `allowedAriaAttrs` is set for the elm type, and if so, verifies against that list. It uses a different failure message that clarifies that the source of the failure is a restriction from ARIA in HTML's allowances for that element type.

We might consider whether we should deprecate `noAriaAttrs` in favor of `allowedAriaAttrs: []`. That makes conceptual sense to me, but would make this a much broader change; it doesn't look like we actually *use* `noAriaAttrs` anywhere in axe-core right now, so that would start enforcing restrictions on quite a few cases that we don't currently enforce.

nami8824: Thanks for the guidance! I updated the implementation based on the proposal.
Does this approach look okay to you?
I limited the change to `hr` as a first step. Supporting other elements might require additional updates.

lib/checks/aria/aria-allowed-attr-elm-evaluate.js
```diff
@@ -0,0 +1,41 @@
...
+    if (validateAttr(attrName) && !allowedAriaAttrs.includes(attrName)) {
+      invalid.push(attrName);
+    }
+  }
+
+  if (!invalid.length) {
+    return true;
+  }
+
+  const messageKey = invalid.length > 1 ? 'plural' : 'singular';
```
straker, line 35: So this is handled automatically when the string is interpolated based on the number of items in the `values` array so you don't need to set it here.

lib/checks/aria/aria-allowed-attr-role.json
```diff
@@ -1,6 +1,6 @@
 {
-  "id": "aria-allowed-attr",
-  "evaluate": "aria-allowed-attr-evaluate",
+  "id": "aria-allowed-attr-role",
```
straker: Renaming a check id is considered a breaking change so we'll want to leave this as just `aria-allowed-attr`

lib/checks/aria/aria-allowed-attr-elm.json
```diff
@@ -0,0 +1,14 @@
+{
+  "id": "aria-allowed-attr-elm",
+  "evaluate": "aria-allowed-attr-elm-evaluate",
+  "metadata": {
+    "impact": "critical",
```
straker: `impact` on checks is deprecated and the rule itself defines the impact so this isn't needed anymore.

lib/checks/aria/aria-allowed-attr-elm-evaluate.js
nami8824: Since this check needs to include `nodeName` in the failure message, I'm passing an object to `this.data()`rather than an array. As I understand it, the automatic singular/plural handling in `process-message.js` only applies when data is an array (`Array.isArray(data)`), so it doesn't seem to work in this case.

I followed the existing pattern in `aria-prohibited-attr-evaluate.js`, which also passes an object containing `nodeName` and explicitly sets `messageKey`.

lib/checks/aria/aria-allowed-attr-role.json
nami8824: I've reverted the check id back to `aria-allowed-attr` and also renamed the files back to their original names

lib/checks/aria/aria-allowed-attr-elm.json
nami8824: Fixed.

nami8824:
I've addressed the review feedback. Could you take another look?

lib/checks/aria/aria-allowed-attr-elm-evaluate.js
straker: I see where the confusion came from. `aria-prohibited-attr-evaluate` is a special case as we have specific message keys called `hasRolePlural/Singular` and `noRolePlural/Singular` so we are creating the message string dynamically. However the normal `singular` and `plural` messages [are automatically calculated for the message key](https://github.com/dequelabs/axe-core/blob/develop/lib/core/utils/process-message.js#L40-L50) so don't need to be done by hand.

lib/checks/aria/aria-allowed-attr-elm-evaluate.js
```diff
@@ -0,0 +1,41 @@
...
+  this.data({
+    messageKey,
+    nodeName: virtualNode.props.nodeName,
+    values: invalid
```
straker, line 39: You don't have to call `.join` here as the code will handle that as well. Not doing `join` will keep it as an array and let the singular/plural keys get handle automatically

straker:
Wait... I see what you mean by the `nodeName`. Yes because the `nodeName` attribute is there you can't just set `data(values)` and that's why you are doing the values and message key yourself. Ignore everything I've said :D

straker (changes requested):
Sorry for the back and forth review. I found one minor thing that we should change. When the element has both global and non-global ARIA attributes, the resulting fix message is:

```
<br aria-busy="true" id="fail9" aria-expanded=false />

"Fix all of the following:
  ARIA attribute is not allowed: aria-expanded=\"false\"
  ARIA attributes are not allowed on br elements: aria-busy=\"true\", aria-expanded=\"false\""
```

This is because the `aria-allowed-attr` check fails for the non-global ARIA attribute `aria-expanded`, and then the new `aria-allowed-attr-elm` check also sees 2 attributes not allowed. I think the fix is to ignore non-global ARIA attributes in the new aria-allowed-elm-check that way the message doesn't have duplication.

```
"Fix all of the following:
  ARIA attribute is not allowed: aria-expanded=\"false\"
  ARIA attribute is not allowed on br elements: aria-busy=\"true\"
```

You can use the function [get-global-aria-attrs](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/standards/get-global-aria-attrs.js) to figure out which ones are global or not.

lib/checks/aria/aria-allowed-attr-elm-evaluate.js
```diff
@@ -16,10 +16,14 @@ export default function ariaAllowedAttrElmEvaluate(node, options, virtualNode) {
   }
 
   const { allowedAriaAttrs } = elmSpec;
+  const globalAriaAttrs = getGlobalAriaAttrs();
   const invalid = [];
 
   for (const attrName of virtualNode.attrNames) {
-    if (validateAttr(attrName) && !allowedAriaAttrs.includes(attrName)) {
```
nami8824, line 22: I removed `validateAttr()` because any attribute returned by `getGlobalAriaAttrs()` is already guaranteed to exist in `standards.ariaAttrs`.

nami8824:
Thanks for the review. I've updated `aria-allowed-attr-elm-evaluate.js`.

Issue #4840, opened 2025-07-25
aria-required-children can double-report related nodes through aria-owns


WilcoFiers:
In the following example `#foo` is reported twice as a related element, once for being a child, and once for aria-owns.

```html
<div role="tablist" aria-owns="foo">
  <div id="foo" role="menuitem">foo</div>
</div>
```

PR #4987, opened 2026-01-11
fix(aria): prevent getOwnedVirtual from returning duplicate nodes

camiha:
When a child element is also referenced via aria-owns, getOwnedVirtual returned the same node twice. Filter aria-owns references that already exist in children using Set.

Closes: #4840


lib/commons/aria/get-owned-virtual.js
```diff
@@ -13,12 +13,15 @@ function getOwnedVirtual(virtualNode) {
...
   if (virtualNode.hasAttr('aria-owns')) {
     const owns = idrefs(actualNode, 'aria-owns')
       .filter(element => !!element)
       .map(element => axe.utils.getNodeFromTree(element));
-    return [...children, ...owns];
+
+    const childNodeSet = new Set(children.map(child => child.actualNode));
+    const uniqueOwns = owns.filter(own => !childNodeSet.has(own.actualNode));
+
+    return [...children, ...uniqueOwns];
```
WilcoFiers: I'm not sure the order is correct this way. This needs to be tested, but I'd wager that if you have child A and B, and you then use aria-owns on A, it'll change the owned elements order to B, A.

lib/commons/aria/get-owned-virtual.js
```diff
@@ -13,12 +13,15 @@ function getOwnedVirtual(virtualNode) {
...
   // TODO: Descend into children with role=presentation|none
-  // TODO: Exclude descendents owned by other elements
   if (virtualNode.hasAttr('aria-owns')) {
     const owns = idrefs(actualNode, 'aria-owns')
       .filter(element => !!element)
       .map(element => axe.utils.getNodeFromTree(element));
-    return [...children, ...owns];
+
+    const childNodeSet = new Set(children.map(child => child.actualNode));
+    const uniqueOwns = owns.filter(own => !childNodeSet.has(own.actualNode));
```
WilcoFiers: We don't need actualNode or Set here. This can be done with children.includes(owned)

lib/commons/aria/get-owned-virtual.js
```diff
@@ -13,12 +13,15 @@ function getOwnedVirtual(virtualNode) {
   }
   // TODO: Check that the element has a role
   // TODO: Descend into children with role=presentation|none
-  // TODO: Exclude descendents owned by other elements
   if (virtualNode.hasAttr('aria-owns')) {
     const owns = idrefs(actualNode, 'aria-owns')
       .filter(element => !!element)
       .map(element => axe.utils.getNodeFromTree(element));
```
WilcoFiers, line 20: I don't think this fully solves the problem of ensuring this function doesn't return the same node twice. You've solved it for returning a child that's also aria-owns, but not for the same node getting picked by aria-owns twice. I think to solve that you need a different solution.

What I don't know is if I do aria-owns="A B A" whether the order I get in the accessibility tree is "A B" or "B A". This should be tested before implementing this, and we'll probably want to leave a comment about it in the code to have a record of that.

lib/commons/aria/get-owned-virtual.js
camiha: I've verified with Safari + VoiceOver and updated accordingly. I also added a test case for this behavior (with `aria-owns="c a"` and children `a, b, c`, the order becomes `b → c → a`).

lib/commons/aria/get-owned-virtual.js
camiha: Simplified to use `includes()` directly.

lib/commons/aria/get-owned-virtual.js
camiha: Updated to deduplicate by first occurrence and added a comment documenting this behavior.

camiha:
@WilcoFiers 
Thank you for the review. This is my first time submitting a code change to an open source project, so I really appreciate the thorough feedback.

I tested the edge cases you mentioned with Safari (26.2) + VoiceOver.

- Child also in aria-owns
	- with `aria-owns="a"` and children `a, b`, I confirmed that VoiceOver reads them in the order `b → a`.
- Duplicate IDs in aria-owns
	- with `aria-owns="a b a"`, element A is only announced once, with the first occurrence taking precedence. (`a → b`)

I've updated the implementation accordingly. I don't have access to JAWS or NVDA testing environments, so if the behavior differs, please let me know and I'll make adjustments.

Issue #4837, opened 2025-07-23
ARIA in HTML updates re: popovertarget and command attributes


scottaohara:
[ARIA in HTML](https://github.com/w3c/html-aria/pull/514) is updating to identify misuse of `aria-expanded` when used on a button element that also have a `popovertarget` or `command` attribute specified.

These updates mirror other similar rules such as input type=checkbox discouraging use of `aria-checked` since the browser already handles these states and the use of author specified ARIA attributes is redundant at best, and can lead to potential misalignments that browsers _should_ correct for - though there have been issues with this in the past.

This [reduced test case](https://w3c.github.io/html-aria/tests/popover-and-command.html) helps illustrate the updates to the spec.

straker:
@chutchins25 Fixing this should be updating [aria-allowed-attr-evaluate's ignoredAttrs fucntion](https://github.com/dequelabs/axe-core/blob/develop/lib/checks/aria/aria-allowed-attr-evaluate.js#L65) to look at if the `vNode` is a button element and has the `aria-expanded` attribute set to `'true'` and does not have the `popovertarget` or `command` attributes then return `true`. Adding a test case similar to the one Scott provided.

PR #4988, opened 2026-01-12
fix: 4837 aria expanded attr

chutchins25:
Using the `aria-expanded` attribute is valid on `button` elements. This adds handling of invalid cases where a `button` element with `aria-expanded` attribute also contains either the `command` attribute or the `popovertarget` attribute or both. Also includes the cases where the additional attributes were being added via a script after the fact (IDL attribute cases).

fixes: [4837](https://github.com/dequelabs/axe-core/issues/4837)

### DIFF
--- a/lib/checks/aria/aria-allowed-attr-evaluate.js
+++ b/lib/checks/aria/aria-allowed-attr-evaluate.js
@@ -45,10 +45,10 @@ export default function ariaAllowedAttrEvaluate(node, options, virtualNode) {
     // popovertarget and command attributes are not present
     // @see https://github.com/dequelabs/axe-core/issues/4837
     if (attrName === 'aria-expanded' && role === 'button') {
-      const isButtonWithStandaloneAriaExpanded = !(
+      const hasNoPopoverOrCommand = !(
         virtualNode.hasAttr('popovertarget') || virtualNode.hasAttr('command')
       );
-      isAllowed = isAllowed && isButtonWithStandaloneAriaExpanded;
+      isAllowed = isAllowed && hasNoPopoverOrCommand;
     }
 
     if (


straker (commented):
Alright, after some testing I think we need to be a bit more strict here than we are being. 

So the accessibility violation should come into play when an element has both the `aria-expanded` attribute and the `popovertarget` attribute. This is because the `popovertarget` attribute [automatically creates the expanded link](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/button#popovertarget) for the element so the addition of the `aria-expanded` attribute is both redundant and can cause problems when it is out of sync to the actual state.

So I tested out what is announced when the `popovertarget` attribute has a value and the associated element is on the page, when the value is empty, and when the associated element is missing. For every screen reader / browser we test (JAWS/Chrome, NVDA/Firefox, and VoiceOver/Mac), the results were the same:

* `popovertarget` attribute has a value and the associated element is on the page - screen readers announce the button in a collapsed state
* `popovertarget` attribute has an empty value - screen readers treat it as a regular button, no collapsed state
*  `popovertarget` associated element is missing - screen readers treat it as a regular button, no collapsed state

So this means the accessibility violation should only happen when 1) the `popovertarget` has a value and 2) when the associated element for the `popovertarget` is also found on the page. This becomes a little tricker when we set the IDL attribute since the attribute value is empty in this case but the IDL value is not. 

Now interestingly when I tested this for the `command` and `commandFor` attribute no screen reader / browser combo added the `aria-expanded` link, even when the `commandFor` associated element was on the page.

@WilcoFiers and @dbjorge should weigh in on how we should treat this. From me, I would say we only throw the error when either attribute (`popovertarget` or `commandFor` [not the `command` attribute]) has a value and the associated element is on the page. Otherwise the link between the button and the element it operates doesn't exist and the button is treated as a normal button.

lib/checks/aria/aria-allowed-attr-evaluate.js
```diff
@@ -39,9 +39,21 @@ export default function ariaAllowedAttrEvaluate(node, options, virtualNode) {
...
   // Unknown ARIA attributes are tested in aria-valid-attr
   for (const attrName of virtualNode.attrNames) {
+    let isAllowed = allowed.includes(attrName);
+
+    // allow aria-expanded for a button element but only if
+    // popovertarget and command attributes are not present
+    // @see https://github.com/dequelabs/axe-core/issues/4837
+    if (attrName === 'aria-expanded' && role === 'button') {
+      const hasNoPopoverOrCommand = !(
+        virtualNode.hasAttr('popovertarget') || virtualNode.hasAttr('command')
```
WilcoFiers, line 49: I think this is a little problematic. Firstly, the original issue doesn't say this, but if you click through to the PR you'll notice that this isn't just any command attribute, it's command attributes with a limited number of values:

> Authors MUST NOT use the `aria-expanded` attribute on any element which also has a `command` attribute in the <strong>Toggle popover</strong>, <strong>Show popover</strong> or <strong>Hide popover</strong> state. E.g., `command=toggle-popover`, `command=show-popover` or `command=hide-popover`.

Even then, did we test that this actually results in a problem? I can well imagine people maybe adding a script to explicitly set aria-expanded as a fallback option to ensure the state is correctly communicate. E.g. if the popover is open and aria-expanded=true, I don't think there is a problem. Even if these two don't agree, as long as browsers consistently pick the implicit state over the explicit one, there might not be a problem either.

The only reason we fail `<input type=checkbox checked aria-checked="false">` is because screen readers announced this inconsistently. You are guaranteed to have an accessibility problem whatever the state is.

straker:
When `aria-expanded=true` is set on a button with `popovertarget` and the popover is closed, here are the results of screen readers:

* JAWS/Chrome - button collapsed (ignores `aria-expanded` value)
* NVDA/Firefox - button expanded (uses the `aria-expanded` value)
* VoiceOver/Safari - button collapsed (ignores `aria-expanded` value)

So I would say screen readers also read the attributes inconsistently and we should treat them like `<input type=checkbox checked aria-checked="false"`

straker:
Closing as this was not a quick fix as I had originally hoped. Ported the comments to the issue.

PR #4998, opened 2026-01-27
feat(core): expose normalizeRunOptions

qazwsxedcrfvtgb1111:
Move normalizeOptions from axe._audit to axe.utils.normalizeRunOptions

Related to: https://github.com/dequelabs/axe-pro-ml-service/issues/1296

### DIFF
--- a/lib/core/base/audit.js
+++ b/lib/core/base/audit.js
@@ -10,11 +10,11 @@ import {
   findBy,
   ruleShouldRun,
   performanceTimer,
-  serializeError
+  serializeError,
+  normalizeRunOptions
 } from '../utils';
 import { doT } from '../imports';
 import constants from '../constants';
-import { utils } from '../../commons';
 
 const dotRegex = /\{\{.+?\}\}/g;
 
@@ -272,7 +272,7 @@ export default class Audit {
    * @param  {Function}  reject  Callback function to fire when audit experiences an error
    */
   run(context, options, resolve, reject) {
-    utils.normalizeRunOptions(options);
+    normalizeRunOptions(options);
     DqElement.setRunOptions(options);
 
     // TODO: es-modules_selectCache


lib/core/base/audit.js
```diff
@@ -271,7 +271,7 @@ export default class Audit {
    * @param  {Function}  reject  Callback function to fire when audit experiences an error
    */
   run(context, options, resolve, reject) {
-    this.normalizeOptions(options);
+    axe.utils.normalizeRunOptions(options);
```
straker: This should be imported from `utils` and not used off the global axe object.

lib/core/public/finish-run.js
```diff
@@ -12,7 +12,7 @@ export default function finishRun(partialResults, options = {}) {
   const { environmentData } = partialResults.find(r => r.environmentData) || {};
 
   // normalize the runOnly option for the output of reporters toolOptions
-  axe._audit.normalizeOptions(options);
+  axe.utils.normalizeRunOptions(options);
```
straker: Also imported instead of directly used

lib/core/utils/normalize-run-options.js
```diff
@@ -0,0 +1,79 @@
+/**
+ * Ensure all rules that are expected to run exist
+ * @throws {Error} If any tag or rule specified in options is unknown
+ * @param  {Object} options  Options object
+ * @return {Object}          Validated options object
+ */
+export default function normalizeRunOptions(options) {
```
straker: The Typescript definition states `options` is optional but this code requires it to at least be an object and not `undefined`. Normally this function was only called after `axe.run/Partial` args were normalized and would set `options` [to an empty object](https://github.com/dequelabs/axe-core/blob/develop/lib/core/public/run/normalize-run-params.js#L33). This function should do the same. Setting a default should suffice.


```suggestion
export default function normalizeRunOptions(options = {}) {
```

test/core/base/audit.js
```diff
@@ -1185,16 +1190,18 @@ describe('Audit', () => {
...
         '<div id="monkeys">bananas</div>' +
         '<input aria-labelledby="monkeys" type="text">' +
         '<blink>FAIL ME</blink>';
       let checked = 'options not validated';
-
-      audit.normalizeOptions = () => {
-        checked = 'options validated';
+      axe.utils = {
+        ...axe.utils,
+        normalizeRunOptions: () => {
```
straker: Checking to see if audit called normalize run options directly is probably not a great test in and of itself. Instead of trying to override `normalizeRunOptions`, we could pass some options that are not correct and check for the error:

```js
assert.throws(() => {
  audit.run(
    { include: [axe.utils.getFlattenedTree(fixture)[0]] },
    { runOnly: { type: 'rule', values: ['foo'] } },
    noop,
    isNotCalled
  )
}, 'unknown rule `foo` in options.runOnly');
```

test/core/utils/normalize-run-options.js
```diff
@@ -0,0 +1,268 @@
...
+    const opt = {
+      runOnly: {
+        type: 'rule',
+        values: ['positive1', 'positive2']
+      },
+      rules: {
+        negative1: { enabled: false }
+      }
+    };
+    assert(axe.utils.normalizeRunOptions(opt), opt);
```
straker: I agree with the Copilot suggestions that having these just use `assert` probably isn't what we should do. I know this is just copy paste code from the original audit tests, but since this is it's now test file now it wouldn't hurt to update the tests to use `assert.deepEqual` and `assert.equal`. However this is just a nit so up to you.

lib/core/base/audit.js
```diff
@@ -14,6 +14,7 @@ import {
 } from '../utils';
 import { doT } from '../imports';
 import constants from '../constants';
+import { utils } from '../../commons';
```
straker: Just a few lines above this are a bunch of imports from utils, so you should grab `normalizeRunOptions` from that  rather than brining in commons here.

PR #5011, opened 2026-02-12
feat(existing-rule): aria-busy now does not allow for unallowed children

chutchins25:
Updated existing rule for `aria-required-children` for cases where the `aria-busy` attribute is used. This now no longer allows for unallowed children and will show the user a specific error message regarding `aria-busy` instead of the one for unallowed children.

UPDATE: This has been changed to be labeled as a feat. It knowingly will introduce an expected failing assertion with the ACT test cases. We identified using aria-busy="true" with unallowed children as a failure but the ACT tests have a case they deem 'inapplicable' but we have agree it should be treated as a fail case.

See the ACT test case [here](https://www.w3.org/WAI/standards-guidelines/act/rules/bc4a75/proposed/#inapplicable-example-4) that will fail.

Closes: [#4626](https://github.com/dequelabs/axe-core/issues/4626)

### DIFF
--- a/test/act-rules/aria-required-owned-element-bc4a75.spec.js
+++ b/test/act-rules/aria-required-owned-element-bc4a75.spec.js
@@ -1,5 +1,7 @@
 require('./act-runner.js')({
   id: 'bc4a75',
   title: 'ARIA required owned elements',
-  axeRules: ['list', 'aria-required-children']
+  axeRules: ['list', 'aria-required-children'],
+  // See https://github.com/act-rules/act-rules.github.io/pull/2387
+  skipTests: ['837f92d0ac41c14e55782991cbab75975b492702']
 });


locales/_template.json
```diff
@@ -503,13 +503,13 @@
...
-        "default": "Required ARIA children are present",
-        "aria-busy": "Element has an aria-busy attribute, so it is allowed to omit required children"
+        "default": "Required ARIA children are present"
       },
       "fail": {
         "singular": "Required ARIA child role not present: ${data.values}",
         "plural": "Required ARIA children role not present: ${data.values}",
-        "unallowed": "Element has children which are not allowed: ${data.values}"
+        "unallowed": "Element has children which are not allowed: ${data.values}",
+        "aria-busy-fail": "Element has an aria-busy attribute, however having it does not allow unallowed children"
```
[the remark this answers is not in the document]
chutchins25: Updated error message based on this feedback but modified the message slightly.

lib/checks/aria/aria-required-children-evaluate.js
```diff
@@ -40,6 +40,11 @@ export default function ariaRequiredChildrenEvaluate(
...
 
   if (unallowed.length) {
+    if (virtualNode.attr('aria-busy') === 'true') {
+      this.data({ messageKey: 'aria-busy-fail' });
+      return false;
+    }
+
     this.relatedNodes(unallowed.map(({ vNode }) => vNode));
     this.data({
       messageKey: 'unallowed',
```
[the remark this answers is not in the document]
chutchins25: To my knowledge, this was intended to act as a short circuit to inform the user they are trying to use `aria-body` and have unallowed children. It feels like this is achieved with existing code changes introduced with this PR.

lib/checks/aria/aria-required-children-evaluate.js
```diff
@@ -55,11 +60,6 @@ export default function ariaRequiredChildrenEvaluate(
     return true;
   }
```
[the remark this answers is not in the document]
chutchins25: Based on discussions with the axe-core team prior to this card, it was understood that we actaully intend for us to REMOVE the existing check, that Copilot is suggesting to bring back, and introduce a whole new check for the unallowed children specific case. The PR code as is achieves this.

lib/checks/aria/aria-required-children.json
```diff
@@ -21,13 +21,13 @@
...
-        "default": "Required ARIA children are present",
-        "aria-busy": "Element has an aria-busy attribute, so it is allowed to omit required children"
+        "default": "Required ARIA children are present"
       },
       "fail": {
         "singular": "Required ARIA child role not present: ${data.values}",
         "plural": "Required ARIA children role not present: ${data.values}",
-        "unallowed": "Element has children which are not allowed: ${data.values}"
+        "unallowed": "Element has children which are not allowed: ${data.values}",
+        "aria-busy-fail": "Element has aria-busy=\"true\", but this does not allow children with roles that are not allowed"
```
WilcoFiers:
```suggestion
        "aria-busy-fail": "Element has children which are not allowed: ${data.value}; aria-busy=\"true\" is not a valid exception"
```

lib/checks/aria/aria-required-children-evaluate.js
```diff
@@ -40,6 +40,11 @@ export default function ariaRequiredChildrenEvaluate(
   );
 
   if (unallowed.length) {
+    if (virtualNode.attr('aria-busy') === 'true') {
+      this.data({ messageKey: 'aria-busy-fail' });
```
WilcoFiers: I think. you still need to report which attributes fail.

lib/checks/aria/aria-required-children.json
chutchins25: Updated text for more clarity and included the list of children that were not allowed.

lib/checks/aria/aria-required-children-evaluate.js
chutchins25: Will add the data.values object to house the list of attributes that fail with aria-busy set to true.

Issue #4928, opened 2025-11-05
Target-size issue with reflowing links


WilcoFiers:
Here's a writeup of the problem in codepen. The TLDR is that axe-core shouldn't always use the bounding box when computing the size / offset of a target. I think the best way to address this is to use the bounding box for the target, and the client rects for its neighbors. We have a little more investigating to do to make sure we didn't miss any edge cases.

https://codepen.io/wilcofiers/pen/NPxJoMP?editors=1100

Amandeque:
Triage with Wilco:
Estimation: 3 - 5 SP 
Owner: Steve
Notes: Needs more grooming with Wilco, Steve and Chris

padmavemulapati:
Verified with the latest axe-core dev build, 
earlier for the code snippet `target-offset` those were failing now only target size failing on obscured target.


https://github.com/user-attachments/assets/acc5cd9d-e4ff-4b19-a7df-155090a926db


### code snippet:

````
<!--target-offset-->
 <div style="font-size: 18px; margin: 1em auto; width: 6em; line-height: 1.3;">
        <a id="target" href="/foo" class="A"> Hello hello hello</a>
        <a href="/bar" class="B">Hello hello hello</a>
      </div>

      <div style="font-size: 18px; margin: 1em auto; width: 6em; line-height: 1.1;">
          <a id="target" href="/foo" class="A"> Hello hello hello</a>
          <a href="/bar" class="B">Hello hello hello</a>
        </div>

      <div style="font-size: 18px; margin: 1em auto; width: 6em; line-height: 1.3;">
          <a id="target" href="/foo" class="A"> Hello hello hello</a>
          <a style="margin-left: -30px" href="/bar" class="B">Hello hello hello</a>
        </div>
        
   <!--target-size-->     
        <div style="position: relative;">
            <button id="target" style="width: 80px; height: 40px;">X</button>
            <button id="obscurer" style="width: 80px; height: 40px; position: absolute; left: 30px; top: 20px;">x</button>
          </div>

          <div style="font-size: 18px; margin: 1em auto; width: 6em; line-height: 1.3">
            <a id="not-obscurer" href="/foo" class="A"> Hello hello</a>
            <a id="target" href="/bar" class="B">Hello hello hello</a>
            <a id="obscurer" href="/bar" class="C">Hello hello hello</a>
          </div>

          <div style="width: 20px">
        <a href="#" id="target">hello world</a>
      </div>

```

PR #5012, opened 2026-02-12
fix(target-size): determine offset using clientRects if target is display:inline

straker:
This pr does two things:

1. For target size we decided to calculate the visible unobscured rects by taking the bounding rect of the target and subtracting the client rects of the obscurer if the obscurer is inline. This is because an inline element only accepts pointer events on the client rects and not the bounding rect. For the target we still take the bounding rect and not the client rects if the target is inline because we both agreed that the goal of the target size rule is to prevent clicking on the wrong target and not that the target itself is large enough to click on. So an inline element with multiple lines would have dead pointer zones within it, but accidentally clicking on a dead zone is not a failure of target size.
2. This also applies using client rects for inline elements for the spacing exception for the same reasons as above. What this does is allow checking multiple target rects for inline nodes and makes sure each rect passes the spacing exception. If at least one rect does not pass the rule fails for that element.

Closes: https://github.com/dequelabs/axe-core/issues/4928

### DIFF
--- a/test/playground.html
+++ b/test/playground.html
@@ -4,87 +4,18 @@
   <!-- The playground is used for quick local host axe-core during development -->
 
   <main>
-    <style>
-      .target-rect {
-        position: absolute;
-        width: 24px;
-        height: 24px;
-        background: rgba(255, 0, 0, 0.4);
-        border-radius: 50%;
-      }
-
-      .neighbor-rect {
-        position: absolute;
-        background: rgba(0, 255, 0, 0.4);
-      }
-    </style>
     <h1>Hello World</h1>
-    <div style="position: relative">
-      <button id="target" style="width: 80px; height: 40px">X</button>
-      <button
-        id="obscurer"
-        style="
-          width: 80px;
-          height: 40px;
-          position: absolute;
-          left: 30px;
-          top: 20px;
-        "
-      >
-        x
-      </button>
-    </div>
-
-    <div
-      style="font-size: 18px; margin: 1em auto; width: 6em; line-height: 1.3"
-    >
-      <a id="pass20" href="/foo" class="A"> Hello hello</a>
-      <a id="pass21" href="/bar" class="B">Hello hello hello</a>
-      <a id="pass22" href="/bar" class="C">Hello hello hello</a>
-    </div>
   </main>
 
   <script src="/axe.js"></script>
   <script src="/test/testutils.js"></script>
   <script>
-    window.showOffset = function (targetRects, neighborRects) {
-      document
-        .querySelectorAll('.target-rect,.neighbor-rect')
-        .forEach(node => node.remove());
-      targetRects.forEach(rect => {
-        const div = document.createElement('div');
-        div.setAttribute('class', 'target-rect');
-        div.setAttribute(
-          'style',
-          `left: ${rect.left + rect.width / 2 - 12}px; top: ${rect.top + rect.height / 2 - 12}px;`
-        );
-        document.body.append(div);
-      });
-      neighborRects.forEach(rect => {
-        const div = document.createElement('div');
-        div.setAttribute('class', 'neighbor-rect');
-        div.setAttribute(
-          'style',
-          `left: ${rect.left}px; top: ${rect.top}px; width: ${rect.width}px; height: ${rect.height}px;`
-        );
-        document.body.append(div);
-      });
-    };
-    window.highlightRect = function (rect, color) {
-      const div = document.createElement('div');
-      div.setAttribute('class', 'neighbor-rect');
-      div.setAttribute(
-        'style',
-        `left: ${rect.left}px; top: ${rect.top}px; width: ${rect.width}px; height: ${rect.height}px; ${color ? `background: ${color};` : ''}`
-      );
-      document.body.append(div);
-    };
     axe.testUtils.awaitNestedLoad(
       window,
       () => {
         axe.run(
           {
-            runOnly: 'target-size',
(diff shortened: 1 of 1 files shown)


lib/commons/dom/get-target-rects.js
```diff
@@ -13,7 +13,9 @@ export default memoize(getTargetRects);
  * @return {DOMRect[]}
  */
 function getTargetRects(vNode) {
-  const nodeRect = vNode.boundingClientRect;
+  const display = vNode.getComputedStylePropertyValue('display');
+  const nodeRects =
+    display === 'inline' ? vNode.clientRects : [vNode.boundingClientRect];
```
WilcoFiers, line 18: I THINK this should still use bounding rect... looking into it

lib/commons/math/get-offset.js
```diff
@@ -21,33 +21,33 @@ export default function getOffset(vTarget, vNeighbor, minRadiusNeighbour = 12) {
     return null;
   }
 
-  const targetBoundingBox = targetRects.reduce(getBoundingRect);
-  const targetCenter = getRectCenter(targetBoundingBox);
+  // calculate distance of each target rect to the closest edge of each neighbor rect
+  return targetRects.reduce((minDistance, targetRect) => {
```
WilcoFiers, line 25: I thought you were allergic to reduce functions Steve? 😅

test/integration/rules/target-size/target-size.html
```diff
@@ -145,6 +145,11 @@
   >
 </p>
 
+<div style="font-size: 18px; margin: 1em auto; width: 6em; line-height: 1.3">
+  <a id="pass18" href="/foo" class="A"> Hello hello hello</a>
+  <a id="pass19" href="/bar" class="B">Hello hello hello</a>
+</div>
```
WilcoFiers, line 151: This passes for the wrong reason. You fixed the offset, which I do think is improved by the change, but with a slight tweak to this example the original problem returns:

```html
<div style="font-size: 18px; margin: 1em auto; width: 6em; line-height: 1.3">
  <a id="pass20" href="/foo" class="A"> Hello hello</a>
  <a id="pass21" href="/bar" class="B">Hello hello hello</a>
  <a id="pass22" href="/bar" class="C">Hello hello hello</a>
</div>
``

The reason this should pass is because B has enough target space. You updated the offset algorithm, but not the target space one.

File: lib/checks/mobile/target-size-evaluate.js [hunk 4]
```diff
@@ -66,7 +66,11 @@ export default function targetSizeEvaluate(node, options, vNode) {
     return true;
   }
 
-  const largestInnerRect = getLargestUnobscuredArea(vNode, obscuredWidgets);
+  const largestInnerRect = getLargestUnobscuredArea(
+    vNode,
+    obscuredWidgets,
+    minSize
```
straker, line 72: This was a bug were we didn't pass `minSize` into the largest rect calculation so no rect would pass min size check and it would default to picking the largest area rect.

lib/checks/mobile/target-size-evaluate.js
```diff
@@ -127,11 +131,16 @@ function filterByElmsOverlap(vNode, nearbyElms) {
...
-    ({ boundingClientRect: rect }) => rect
-  );
+  const obscuringRects = obscuredNodes
+    .map(obscuredNode => {
+      const display = obscuredNode.getComputedStylePropertyValue('display');
+      return display === 'inline'
+        ? obscuredNode.clientRects
+        : obscuredNode.boundingClientRect;
+    })
+    .flat(Infinity);
```
WilcoFiers, line 143: What is this for? Why not just 2?

lib/commons/math/get-offset.js
```diff
@@ -21,33 +21,33 @@ export default function getOffset(vTarget, vNeighbor, minRadiusNeighbour = 12) {
...
+    if (rectHasMinimumSize(minRadiusNeighbour * 2, neighborTargetSize)) {
+      return minDistance;
+    }
 
-  const neighborBoundingBox = neighborRects.reduce(getBoundingRect);
-  const neighborCenter = getRectCenter(neighborBoundingBox);
-  // subtract the radius of the circle from the distance to center to get distance to edge of the neighbor circle
-  const centerDistance =
-    pointDistance(targetCenter, neighborCenter) - minRadiusNeighbour;
+    const neighborBoundingBox = neighborRects.reduce(getBoundingRect);
```
chutchins25, line 43: Curious. There are several const variables in this .reduce() loop. Does each pass through the loop alter these in any way? Might be a minor gain in performance but I was just curious if there was a way to move them out of the targetRects.reduce() loop.

test/playground.html
WilcoFiers, line 1: Please pull this change. The rest looks right.

lib/commons/dom/get-target-rects.js
```diff
@@ -23,13 +25,19 @@ function getTargetRects(vNode) {
...
-  return splitRects(nodeRect, obscuringRects);
+  const obscuringRects = overlappingVNodes
+    .map(overlappingVNode => {
+      const overlappingDisplay =
+        overlappingVNode.getComputedStylePropertyValue('display');
+      return overlappingDisplay === 'inline'
+        ? overlappingVNode.clientRects
+        : overlappingVNode.boundingClientRect;
+    })
+    .flat(Infinity);
```
WilcoFiers, line 39: Same here, this is 2 layers deep. Why give it Infinity?

WilcoFiers (changes requested):

Issue #4996, opened 2026-01-26
element.cloneNode can cause problems with custom elements


WilcoFiers:
Axe-core uses element.cloneNode in a couple places. Essentially for two reasons:

1. to get the element's HTML without its content a shallow clone is used
2. To get the attributes where element.attributes is set for silly reasons `<input name=attribute>`

Calling `element.cloneNode` can invoke the constructor of custom elements. I don't think for scenario 2 that's avoidable, but for scenario 1 it is. Instead of cloning, we can do the following:

```js
elmOnlyHtml = elm.outerHTML.replace(elm.innerHTML, '')
```

## UPDATE

The solve here is just to rewrite `truncateElement` to in a way that doesn't require cloneNode. I would also refactor the code. This `getSource` code should be its own function with its own tests. It'd be nice to make the params configurable. This is a nice function to have:

```
axe.utils.getSource(elm, { maxLength: 300, attrLimit: 20 })
```

PR #5013, opened 2026-02-13
fix(DqElement): avoid calling constructors with cloneNode

WilcoFiers:
Closes: https://github.com/dequelabs/axe-core/issues/4996

- **Avoid calling cloneNode when retrieving source**
- Add a new utils.getElementSource (which works even if the tree is not constructed
- Made sure getElementSource works with namespaces and non-node elements
- Check each attribute if it fits in the truncated source, instead of stopping at the first that doesn't fit

### DIFF
--- a/lib/core/utils/get-element-source.js
+++ b/lib/core/utils/get-element-source.js
@@ -44,7 +44,8 @@ function getOuterHtml(element) {
 }
 
 /**
- * Builds a truncated HTML representation of an element when outerHTML exceeds maxLength
+ * Builds a truncated HTML representation of an element when outerHTML exceeds maxLength.
+ * Note: attribute order may differ from the original source as node.attributes order is not guaranteed.
  * @param  {Element} elm the DOM element
  * @param  {Object} options truncation options
  * @param  {Number} options.maxLength maximum length of the output
@@ -61,9 +62,11 @@ function getTruncatedElementSource(elm, { maxLength, attrLimit }) {
     ({ name, value }) => ({ name, value })
   );
   const attrsLength = nodeAttrs.reduce((acc, { name, value }) => {
+    // 4 = space before name + equals sign + opening quote + closing quote
     return acc + name.length + value.length + 4;
   }, 0);
 
+  // 2 = opening "<" + space before first attribute
   if (2 + nodeName.length + attrsLength > maxLength) {
     nodeAttrs.forEach(attr => {
       attr.name = truncate(attr.name, attrLimit);
@@ -74,7 +77,7 @@ function getTruncatedElementSource(elm, { maxLength, attrLimit }) {
   let source = `<${nodeName}`;
   let tagEnd = '>';
   const truncateEnd = ' ...>';
-  // Append any attribute that fits within maxLength
+  // Must check every attribute: an attr that doesn't fit may be followed by one that does
   for (const attr of nodeAttrs) {
     const attrStr = ` ${attr.name}="${attr.value}"`;
     if (source.length + attrStr.length > maxLength - truncateEnd.length) {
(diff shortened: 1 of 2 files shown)


lib/core/utils/get-element-source.js
```diff
@@ -0,0 +1,98 @@
...
+  const nodeName = isXHTML(elm.ownerDocument || document)
+    ? elm.nodeName
+    : elm.nodeName.toLowerCase();
+
+  // Get a mutable attribute map, and work out their rendered length
+  const nodeAttrs = Array.from(getNodeAttributes(elm)).map(
+    ({ name, value }) => ({ name, value })
+  );
+  const attrsLength = nodeAttrs.reduce((acc, { name, value }) => {
+    return acc + name.length + value.length + 4;
```
straker, line 66: I'm guessing what the 4 represents but we should comment the number

```suggestion
    // 4 = whitespace + equal sign + start and end quote
    return acc + name.length + value.length + 4;
```

lib/core/utils/get-element-source.js
```diff
@@ -0,0 +1,98 @@
...
+  }, 0);
+
+  if (2 + nodeName.length + attrsLength > maxLength) {
```
straker, line 70: Same here for the 2


```suggestion
  // 2 = whitespace before attrs + ?
  if (2 + nodeName.length + attrsLength > maxLength) {
```

lib/core/utils/get-element-source.js
```diff
@@ -0,0 +1,98 @@
...
+
+  let source = `<${nodeName}`;
+  let tagEnd = '>';
+  const truncateEnd = ' ...>';
+  // Append any attribute that fits within maxLength
+  for (const attr of nodeAttrs) {
+    const attrStr = ` ${attr.name}="${attr.value}"`;
+    if (source.length + attrStr.length > maxLength - truncateEnd.length) {
+      tagEnd = truncateEnd;
+      continue;
```
straker, line 85: This should probably `break` since we've reached the max length. Otherwise this will continue to process attrs we don't need to. 

*Edit*: I see the last test tests that the function can add later attrs that do fit when priors don't. This just means we will always loop over every attribute even if none could fit. Not sure if that's a performance problem, but something to keep in mind.

lib/core/utils/get-element-source.js
```diff
@@ -0,0 +1,98 @@
...
+    }
+    source += attrStr;
+  }
+
+  return source + tagEnd;
```
straker, line 90: One thing to note with this approach is the case where the full `outerHTML` is too long but the `outerHTML` of just the element without children is not too long. As far as I am aware `node.attributes` is not guaranteed to be the same order as the element source. E.g.

```html
<a id="link" href="#" style="color: blue" class="thing"></a>
<script>
link.attributes // { 0: style, 1: href, 2: class, 3: id }
</script>
```

Not having the same order would mean that combining them back into a string could change the attribute order from what the DOM is showing. Not sure if that will be a problem or not, but something to keep in mind. Probably worth the performance gain over doing what we thought before.

test/core/utils/get-element-source.js
```diff
@@ -0,0 +1,233 @@
...
+    it('should truncate SVG with namespaced attributes correctly', () => {
+      const vNode = queryFixture(
+        '<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">' +
+          '<a id="target" xlink:href="#section" class="link">' +
+          'x'.repeat(400) +
+          '</a>' +
+          '</svg>'
+      );
+      const result = getElementSource(vNode.actualNode, { maxLength: 50 });
+      assert.include(result, 'a');
```
straker: Just looking for the letter `a` may not be exactly what we want to test as `id="target"` also includes the letter `a`. So not sure if this is testing that the `a` node is in the string or something else.

lib/core/utils/get-element-source.js
WilcoFiers: Fair, but that's also the case for the current implementation. Even if we detect that it it'll fit we use node.attributes to generate the string.

straker (approved):

Issue #5018, opened 2026-02-27
Memory leak in ACT test runner: `process` `exit` listeners accumulate across test suites


Garbee:
### Summary

Running the ACT rules test suite triggers Node's `MaxListenersExceededWarning` because each spec file leaves a dangling `exit` listener on `process` that is never cleaned up.

### Call chain

Each of the ~35 spec files calls `act-runner.js`, which in its `before()` hook calls `getWebdriver()` ([act-runner.js#L32](https://github.com/dequelabs/axe-core/blob/a4b809183f43c4296a3ec57cd80d8a8f34743361/test/act-rules/act-runner.js#L32)):

```js
driver = getWebdriver();
```

`getWebdriver()` calls `Builder.build()`, which spawns a ChromeDriver subprocess via selenium-webdriver's internal `exec()` function. On every spawn, `exec()` registers a cleanup listener ([selenium-webdriver/io/exec.js#L134](https://github.com/dequelabs/axe-core/blob/a4b809183f43c4296a3ec57cd80d8a8f34743361/node_modules/selenium-webdriver/io/exec.js#L134)):

```js
process.once('exit', onProcessExit)
```

This listener is designed to self-clean when the child process exits ([exec.js#L139](https://github.com/dequelabs/axe-core/blob/a4b809183f43c4296a3ec57cd80d8a8f34743361/node_modules/selenium-webdriver/io/exec.js#L139)):

```js
process.removeListener('exit', onProcessExit)
```

### Root cause

The `after()` hook tears down the driver with `driver.close()` ([act-runner.js#L50](https://github.com/dequelabs/axe-core/blob/a4b809183f43c4296a3ec57cd80d8a8f34743361/test/act-rules/act-runner.js#L50)):

```js
after(async () => {
  await driver.close();
  await new Promise(r => server.close(r));
});
```

`driver.close()` closes the browser window but **does not stop the ChromeDriver service process**. Because the subprocess never exits during the test run, the `onProcessExit` listener on `process` is never removed. With ~35 spec files each creating their own driver, **35 dangling `exit` listeners accumulate** on `process`.

### Fix

Change `driver.close()` to `driver.quit()` at [act-runner.js#L50](https://github.com/dequelabs/axe-core/blob/a4b809183f43c4296a3ec57cd80d8a8f34743361/test/act-rules/act-runner.js#L50):

```diff
- await driver.close();
+ await driver.quit();
```

[...]

PR #5025, opened 2026-03-04
test: Memory leak fix in act runner

pswitchy:
I resolved the `MaxListenersExceededWarning` by identifying that `process.on('exit')` listeners were being leaked in each test file. By switching from `driver.close()` to `driver.quit()`, I ensured that the **ChromeDriver child process** is fully terminated rather than just closing the browser window.

#### **1. Robust Teardown Logic**

To prevent `EADDRINUSE` errors and handle cases where the driver might crash, I wrapped the teardown in a `try/catch/finally` block. This guarantees the server closes even if the driver fails to quit.

```javascript
after(async () => {
  try {
    // I switched to .quit() to kill the process and clear listeners
    if (driver) await driver.quit();
  } catch (err) {
    console.error('Error quitting driver:', err.message);
  } finally {
    // I use finally to ensure the server always releases the port
    if (server) await new Promise(r => server.close(r));
  }
});

```

#### **2. My CI Stress Test Configuration**

To prove the stability of my fix, I modified the GitHub Actions workflow. I added a strategy matrix to force the suite to run 10 times concurrently on every PR.

```yaml
test_act:
    runs-on: ubuntu-24.04
    strategy:
      fail-fast: false # I disabled fail-fast so all 10 runs must complete
      matrix:
        run_number: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
    steps:
      - name: Run ACT test suite
        run: npm run test:act

```

Closes:
#5018

### DIFF
--- a/test/get-webdriver.js
+++ b/test/get-webdriver.js
@@ -5,7 +5,12 @@ const chromedriverPath =
 
 const getWebdriver = () => {
   const service = new chrome.ServiceBuilder(chromedriverPath);
-  const options = new chrome.Options().addArguments('--headless');
+  const options = new chrome.Options().addArguments([
+    '--headless',
+    '--no-sandbox',
+    '--disable-extensions',
+    '--disable-dev-shm-usage'
+  ]);
 
   if (process.env.CHROME_BIN) {
     options.setBinaryPath(process.env.CHROME_BIN);


pswitchy:
> Thanks for pr. Once the tests runs the 10x and we've shown it doesn't error, we'll need you to remove the 10x runner before we merge.

chrome is crashing in these tests, do you need me to do this right now, or do i try to fix the issue first?

.github/workflows/test.yml
```diff
@@ -131,6 +131,10 @@ jobs:
     runs-on: ubuntu-24.04
     timeout-minutes: 10
     needs: build
+    strategy:
+      fail-fast: false
+      matrix:
+        run_number: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
```
stephenmathieson, line 137: This needs to be removed prior to merging.

test/act-rules/act-runner.js
```diff
@@ -47,8 +47,13 @@ module.exports = ({ id, title, axeRules, skipTests = [] }) => {
     });
 
     after(async () => {
-      await driver.close();
-      await new Promise(r => server.close(r));
+      try {
+        if (driver) await driver.quit();
```
stephenmathieson, line 51: This is a good fix, but this PR is incomplete. [`test/aria-practices/apg.spec.js`](https://github.com/dequelabs/axe-core/blob/develop/test/aria-practices/apg.spec.js#L31) has the same problem. Could you please update it too?

test/get-webdriver.js
```diff
@@ -5,7 +5,12 @@ const chromedriverPath =
 
 const getWebdriver = () => {
   const service = new chrome.ServiceBuilder(chromedriverPath);
-  const options = new chrome.Options().addArguments('--headless');
+  const options = new chrome.Options().addArguments([
+    '--headless',
+    '--no-sandbox',
+    '--disable-extensions',
+    '--disable-dev-shm-usage'
+  ]);
```
stephenmathieson, line 13: These are reasonable additions, but adding them seems unrelated to this PR.

Issue #501, opened 2017-08-28
Make results better for calculating metrics based on passes and failures


dylanb:
The passes for rules that do not filter with a matches function contain "passes" that are not that meaningful. For example, the aria-required-children should only evaluate aria roles that actually require children and should not mark things as passed if they do not have any required children.

Other rules probably suffer from similar issues.

So first order of business is to review all the rules to see where we could benefit from filtering and review all the current filters to see which ones could be made better

straker:
Going through all the rules these are the ones I found that I think could benefit:

- `aria-required-children` - matches to see if role requires children
- `aria-required-parent` - matches to see if role requires parent
- `avoid-inline-spacing` - matches to see if element has the CSS properties we look for (right now all `[style]` elements will pass even if they don't use any inline-spacing properties)

straker:
Issue has been addressed. We will always run `inline-spacing` rule on `[style]` nodes.

PR #5026, opened 2026-03-05
test: fix chromedriver 146 failing to create session

straker:
Similar to our fix in [axe-core-npm](https://github.com/dequelabs/axe-core-npm/pull/1291) this pins chrome/driver to v145 in order to fix the failing tests due to chromedriver crashing or failing to create a session. It also fixes the memory issue in https://github.com/dequelabs/axe-core/issues/5018 and adds some chrome options that are good for ci usage. I also created a tech debt ticket to track the pin.

See https://github.com/dequelabs/axe-core/actions/runs/22737133182?pr=5026 that it shows 10 successful runs.

Closes: #501

### DIFF
--- a/.github/actions/install-deps/action.yml
+++ b/.github/actions/install-deps/action.yml
@@ -56,6 +56,7 @@ runs:
       id: setup-chrome
       uses: browser-actions/setup-chrome@b94431e051d1c52dcbe9a7092a4f10f827795416 # v2.1.0
       with:
+        # @see https://github.com/dequelabs/axe-core/issues/5027
         chrome-version: ${{ inputs.nightly == 'true' && 'beta' || 145 }}
         install-chromedriver: true
         install-dependencies: true


.github/actions/install-deps/action.yml
```diff
@@ -56,7 +56,7 @@ runs:
       id: setup-chrome
       uses: browser-actions/setup-chrome@b94431e051d1c52dcbe9a7092a4f10f827795416 # v2.1.0
       with:
-        chrome-version: ${{ inputs.nightly == 'true' && 'beta' || 'stable' }}
+        chrome-version: ${{ inputs.nightly == 'true' && 'beta' || 145 }}
```
stephenmathieson, line 60: Hard coding this seems like a pain to maintain. Can we call this out in contributing documentation?

stephenmathieson (changes requested):

.github/actions/install-deps/action.yml
Garbee: This is temporary and we have other solutions that we can apply later. Like using a variable in the repository settings and use that if available, otherwise default to `'stable'`. This isn't going to be hard-coded long-term (or probably even medium-term.) As soon as we can validate Chromedriver is fixed, we go back to stable as the default.

stephenmathieson: >  This isn't going to be hard-coded long-term

Famous last words

---

Can we at least add a comment to the YAML saying this is temporary? Maybe a `# TODO: Replace the hard-coded version with ____`?

straker: I created a ticket to do that https://github.com/dequelabs/axe-core/issues/5027

stephenmathieson: Any reason not to link to the ticket in code? Doing so is a pretty common practice in most other Deque repos

Issue #5028, opened 2026-03-05
Shadow DOM list false positive


dylanb:
### Product

axe-core

### Product Version

4.11.1

### Latest Version

- [x] I have tested the issue with the latest version of the product

#### Expectation
When a `<ul>` or `<ol>` element uses a `<slot>` to distribute custom elements whose shadow DOM renders as `<li>`, the list rule should pass. The structure is semantically valid — each custom element is a web component wrapper around an `<li>`.

#### Actual
The list rule fails with a violation on the `<ul>`, reporting that it contains non-`<li>` children. This is a false positive caused by axe-core evaluating the slotted custom elements (`<my-list-item>`) as the direct children of `<ul>` in the flat tree, without looking through their shadow DOM to see that they render as `<li>`.

#### How to Reproduce
https://codepen.io/lupecamacho/pen/pvEgQPj

  Or the following minimal example:

  ```html
<!-- Light DOM -->
  <my-list>
    <my-list-item>One</my-list-item>
    <my-list-item>Two</my-list-item>
    <my-list-item>Three</my-list-item>
  </my-list>
```

```js
  import { LitElement, html } from "https://cdn.skypack.dev/lit";

  class MyList extends LitElement {
    render() {
      return html`<ul><slot></slot></ul>`;
    }
  }

  class MyListItem extends LitElement {
    render() {
      return html`<li><slot></slot></li>`;
    }
  }

  customElements.define("my-list", MyList);
  customElements.define("my-list-item", MyListItem);
```
Running `axe.run()` reports a list violation on the `<ul>` element inside `<my-list>`'s shadow DOM.

#### Additional context

The issue is in the only-listitems check (lib/checks/lists/only-listitems-evaluate.js). axe-core's flat tree algorithm correctly resolves the `<slot>` in `<ul>`, making the `<my-list-item>` elements the virtual children of `<ul>`. However, the check only tests whether each child is literally an `<li>` element or has an explicit listitem role — it does not account for the case where the child is a shadow host whose shadow DOM root element is an `<li>`.

The same pattern applies to any web component that encapsulates list item semantics in a shadow DOM (e.g. design system components).

PR #5029, opened 2026-03-06
fix: list item in shadow DOM false positive

dylanb:
When a `<ul>` or `<ol>` element uses a `<slot>` to distribute custom elements whose shadow DOM renders as `<li>`, the list rule should pass. The structure is semantically valid — each custom element is a web component wrapper around an `<li>`.

closes https://github.com/dequelabs/axe-core/issues/5028


lib/checks/lists/structured-dlitems-evaluate.js
```diff
@@ -23,4 +28,25 @@ function structuredDlitemsEvaluate(node, options, virtualNode) {
   return hasDt || hasDd;
 }
 
+/**
```
dylanb, line 30: should we make this a util?

dylanb: Seems like CoPilot thinks I should 🤣

PR #5035, opened 2026-03-19
chore: add CLAUDE.md and pull request checklist

dylanb:
Adds a CLAUDE.md file for contributors

### DIFF
--- a/doc/pull-request-checklist.md
+++ b/doc/pull-request-checklist.md
@@ -31,7 +31,7 @@ Complete all applicable items before opening a PR. For items that do not apply t
 - [ ] All commits follow Angular format: `<type>(<scope>): <subject>`
 - [ ] Subject is imperative, lowercase, no trailing period, ≤100 chars total
 - [ ] Commit body explains motivation (not just what changed)
-- [ ] Footer includes issue reference: `Closes issue #123` or full URL
+- [ ] Footer includes issue reference: `Closes issue: #123` or full URL
 - [ ] Breaking changes avoided — prefer supporting both old and new formats simultaneously. If unavoidable, document in footer: `BREAKING CHANGE: description`
 
 ## Docs


doc/examples/code-patterns.md
```diff
@@ -0,0 +1,153 @@
...
+  const rect = node.getBoundingClientRect();
+  const rootNode = node.getRootNode();
+}
+
+// Convert ambiguous input to Virtual Node
+import { getNodeFromTree } from '../../core/utils';
+
+function myFunction(nodeOrVirtual) {
+  const vNode =
+    nodeOrVirtual instanceof AbstractVirtuaNode
```
straker: This should use `nodeLookup` instead to resolve the `vNode`.

```js
const { vNode } = nodeLookup(nodeOrVirtual)
```

doc/examples/pr-review-patterns.md
```diff
@@ -0,0 +1,69 @@
...
+2. Clear, detailed commit messages with motivation
+3. JSDoc comments that explain "why" not just "what"
+4. Performance-conscious code that caches appropriately
+5. Following existing patterns in similar files
+6. Integration tests that cover both pass and fail cases
+
+## Common PR Mistakes
+
+1. Not running `npm test` locally before pushing
+2. Committing generated files (`axe.js`, `axe.min.js`, `locales/_template.json`) in a separate commit — always include them with their source changes
```
straker: `axe.js` and `axe.min.js` should never be committed

doc/examples/pr-review-patterns.md
```diff
@@ -0,0 +1,69 @@
...
+3. Changing multiple unrelated things in one PR — split refactoring from feature work
+4. Not updating integration tests when changing rule behavior
+5. Using `console.log` instead of `import log from '../../core/log'`
```
straker: This is rarely a thing. Most of the time `console.log` statements should not be committed.

doc/examples/rule-check-templates.md
```diff
@@ -0,0 +1,111 @@
...
+    }
+  }
+}
+```
+
+**Message templates:**
+
+- Use `${data.property}` for dynamic values set via `this.data()`
+- Support singular/plural variants via object with `singular`/`plural` keys
+- Always update `locales/_template.json` when changing messages
```
straker: This file is auto generated with `npm run build` so should not be manually changed

doc/examples/rule-check-templates.md
```diff
@@ -0,0 +1,111 @@
...
+
+- `true` — check passed
+- `false` — check failed
+- `undefined` — incomplete (cannot determine result)
+
+**`this.data()` usage:**
+
+- `messageKey` — selects which message variant to use
+- `values` — dynamic content for `${data.values}` in message templates
+- `missingData` — reason for incomplete results
```
straker: `values` isn't a property you set on `this.data` but instead is only used in the message string to resolve an array of values set as `this.data`.

`missingData` also isn't as set property name but instead should be any key name

doc/examples/test-patterns.md
```diff
@@ -0,0 +1,101 @@
...
+
+```javascript
+describe('aria-allowed-attr', function () {
+  'use strict';
+
+  var fixture = document.getElementById('fixture');
+  var checkContext = axe.testUtils.MockCheckContext();
+
+  afterEach(function () {
+    fixture.innerHTML = '';
```
straker: `fixture.innerHTML` is cleared automatically and shouldn't be needed in an `afterEach`

doc/examples/test-patterns.md
```diff
@@ -0,0 +1,101 @@
...
+    checkContext.reset();
+    axe._tree = null;
```
straker: The tree is also cleared after each test already

doc/examples/test-patterns.md
```diff
@@ -0,0 +1,101 @@
...
+    assert.equal(axe.commons.text.sanitize(null), '');
+  });
+});
+```
+
+## Unit Test — Check Evaluate
+
+```javascript
+describe('aria-allowed-attr', function () {
+  'use strict';
```
straker: `use strict` shouldn't be used anymore

doc/examples/test-patterns.md
```diff
@@ -0,0 +1,101 @@
...
+
+  var fixture = document.getElementById('fixture');
```
straker: Should use `const`

doc/examples/test-patterns.md
```diff
@@ -0,0 +1,101 @@
...
+    checkContext.reset();
+    axe._tree = null;
+  });
+
+  it('should return true if all ARIA attributes are allowed', function () {
+    fixture.innerHTML =
+      '<div role="textbox" aria-placeholder="foo" id="target"></div>';
+    axe.testUtils.flatTreeSetup(fixture);
+    var node = fixture.querySelector('#target');
+    var virtualNode = axe.utils.getNodeFromTree(node);
```
straker: This should use `queryFixture` to get the vNode 

```js
const queryFixture = axe.testUtils.queryFixture

const vNode = queryFixture('<div role="textbox" aria-placeholder="foo" id="target"></div>');
```

doc/examples/test-patterns.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
+## Shadow DOM Test
+
+```javascript
+it('should work with Shadow DOM', function () {
+  var shadowSupport = axe.testUtils.shadowSupport.v1;
+  if (!shadowSupport) {
+    this.skip();
+  }
```
straker: We fully support shadow dom in our test browsers and don't need to check for support anymore so this can be removed.

doc/examples/test-patterns.md
```diff
@@ -0,0 +1,101 @@
...
+
+  fixture.innerHTML = '<div id="host"></div>';
+  var host = fixture.querySelector('#host');
+  var shadowRoot = host.attachShadow({ mode: 'open' });
+  shadowRoot.innerHTML = '<div role="button" id="target">Test</div>';
+
+  axe.testUtils.flatTreeSetup(fixture);
```
straker: This should use `queryShadowFixture` to setup the shadow host and dom

```js
const vNode = queryShadowFixture(
  '<div id="host"></div>',
  '<div role="button" id="target">Test</div>'
)
```

doc/examples/test-patterns.md
```diff
@@ -0,0 +1,101 @@
...
+  // Test your function against the shadow DOM content
+});
+```
+
+## Integration Test — Rule
+
+Each rule change requires an HTML + JSON pair in `test/integration/rules/<rule-name>/`.
```
straker: This is somewhat correct. we have `integration/rules` and `integration/full` and where the integration tests go depends on if the rule requires a full HTML page that we control (`full`) rather than use the one created from mocha (`rules`). We also have `virtual-rules` that will also need to be updated or created for appropriate rules

doc/pull-request-checklist.md
```diff
@@ -0,0 +1,40 @@
+# axe-core — PR Checklist
+
+Complete all applicable items before opening a PR. For items that do not apply to this PR (for example, tests for a `chore`-only change), mark them as "N/A" and follow `doc/code-submission-guidelines.md`. Reviewers will not merge until all required boxes are checked.
+
+## Code
+
+- [ ] Default export is at the top of the file, immediately after imports
+- [ ] Return early pattern used — no nested conditionals where an early return works
+- [ ] DocBlock comments where appropriate (especially for public/exported APIs), per `doc/code-submission-guidelines.md`
+- [ ] `console.log` replaced with the `lib/core/log.js` wrapper (imported via a relative path)
```
straker: No console logs

doc/pull-request-checklist.md
```diff
@@ -0,0 +1,40 @@
...
+- [ ] No hardcoded ARIA/HTML lists — queried from `standards/` via `commons/standards`
+- [ ] Imports follow directory restrictions — especially no `commons` from `core/utils`
+
+## Tests
+
+- [ ] Unit tests cover all code paths including `null`/`undefined` inputs
```
straker: If this means function inputs then `null` or `undefined` will never be passed to the standard matches or evaluate functions. This should only apply to functions on the public api.

doc/pull-request-checklist.md
```diff
@@ -0,0 +1,40 @@
...
+- [ ] Integration test HTML + JSON pair added/updated for any rule changes
+- [ ] Shadow DOM test case added where relevant
+- [ ] `npm test` passes locally
```
straker: Should also pass the following tests locally (`npm test` only covers the unit tests):

- `test:act`
- `test:apg`
- `test:virtual-rules`
- `test:examples`
- `test:integration`
- `test:node`
- `test:jsdom`

Also the code will need to be built before any testing happens

doc/pull-request-checklist.md
```diff
@@ -0,0 +1,40 @@
...
+- [ ] `npm run eslint` passes
+- [ ] `npm run build` run — generated artifacts that are tracked in git (for example `locales/_template.json`) are committed in the same commit as source changes
+
+## Commits
+
+- [ ] All commits follow Angular format: `<type>(<scope>): <subject>`
+- [ ] Subject is imperative, lowercase, no trailing period, ≤100 chars total
+- [ ] Commit body explains motivation (not just what changed)
+- [ ] Footer includes issue reference: `Closes issue #123` or full URL
+- [ ] Breaking changes documented in footer: `BREAKING CHANGE: description`
```
straker: While true, we avoid making any breaking changes and instead opt to support backwards compatibility while introducing new changes. If we want to move towards a new format we support both at the same time without breaking.

doc/pull-request-checklist.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Docs
+
+- [ ] `doc/rule-descriptions.md` updated for new rules
```
straker: This file is automatically generated with the build and shouldn't be manually edited.

doc/pull-request-checklist.md
```diff
@@ -0,0 +1,42 @@
...
+- [ ] `npm run fmt` passes (Prettier)
+- [ ] `npm run eslint` passes
+- [ ] `npm run build` run — generated artifacts that are tracked in git (for example `locales/_template.json`) are committed in the same commit as source changes
+
+## Commits
+
+- [ ] All commits follow Angular format: `<type>(<scope>): <subject>`
+- [ ] Subject is imperative, lowercase, no trailing period, ≤100 chars total
+- [ ] Commit body explains motivation (not just what changed)
+- [ ] Footer includes issue reference: `Closes issue #123` or full URL
```
Garbee:
```suggestion
- [ ] Footer includes issue reference: `Closes issue: #123` or full URL
```

We should probably encourage proper trailer syntax since this is automated. Help keep things as consistent as we can for the machines.

thomasloke-web (commented):
doc/pull-request-checklist.md

PR #5059, opened 2026-04-05
fix(utils): update isShadowRoot to use spec-compliant custom element regex

### LINKED ISSUE #5030: Regex for allowed shadow hosts is inaccurate
### Product

axe-core

### Product Version

4.11.1

### Latest Version

### Issue Description

According to the [latest specs](https://html.spec.whatwg.org/multipage/custom-elements.html#valid-custom-element-name), our regex is slightly incorrect

Examples that fail that should not fail: `<math-π>`, `<emotion-😍>` or `<café-menu>`

Also, we are not testing for the reserved names

Here is a proposed updated isShadowRoot

```js
function isShadowRoot(node) {
  if (!node || !node.shadowRoot) return false;

  const nodeName = node.nodeName.toLowerCase();

  // 1. Valid Built-in Host Elements (W3C Standard)
  const possibleShadowRoots = [
    'article', 'aside', 'blockquote', 'body', 'div', 
    'footer', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6', 
    'header', 'main', 'nav', 'p', 'section', 'span'
  ];

  if (possibleShadowRoots.indexOf(nodeName) !== -1) {
    return true;
  }

  // 2. Reserved Custom Element Names (Must return false even if they match the regex)
  const reservedNames = [
    'annotation-xml', 'color-profile', 'font-face', 'font-face-src', 
    'font-face-uri', 'font-face-format', 'font-face-name', 'missing-glyph'
  ];

  if (reservedNames.indexOf(nodeName) !== -1) {
    return false;
  }

  // 3. Spec-Compliant PCEN (Potential Custom Element Name) Regex
  // This captures Unicode characters and is safe for legacy environments/JSDOM.
  const pcenRegex = /^[a-z](?:[a-z0-9._-]|[\u00B7\u00C0-\u00D6\u00D8-\u00F6\u00F8-\u037D\u037F-\u1FFF\u200C-\u200D\u203F-\u2040\u2070-\u218F\u2C00-\u2FEF\u3001-\uD7FF\uF900-\uFDCF\uFDF0-\uFFFD])*-(?:[a-z0-9._-]|[\u00B7\u00C0-\u00D6\u00D8-\u00F6\u00F8-\u037D\u037F-\u1FFF\u200C-\u200D\u203F-\u2040\u2070-\u218F\u2C00-\u2FEF\u3001-\uD7FF\uF900-\uFDCF\uFDF0-\uFFFD])*$/;

  return pcenRegex.test(nodeName);
}
```

bwyard:
The custom element name regex in `isShadowRoot` only matches ASCII characters. The HTML spec allows a broader set of Unicode characters in Potential Custom Element Names (PCEN), so valid elements like `<café-menu>` and `<math-π>` were being rejected as shadow root candidates.

Two changes:
- Replace the ASCII-only regex with the spec-compliant PCEN character ranges
- Add an explicit reserved names check (`annotation-xml`, `color-profile`, `font-face`, etc.): these match the pattern but are explicitly excluded by the spec

Tests added for both cases.

Closes #5030

### DIFF
--- a/lib/core/utils/is-shadow-root.js
+++ b/lib/core/utils/is-shadow-root.js
@@ -18,6 +18,26 @@ const possibleShadowRoots = [
   'section',
   'span'
 ];
+
+// Reserved names that match the custom element regex but are not valid custom elements.
+// https://html.spec.whatwg.org/multipage/custom-elements.html#valid-custom-element-name
+const reservedNames = [
+  'annotation-xml',
+  'color-profile',
+  'font-face',
+  'font-face-src',
+  'font-face-uri',
+  'font-face-format',
+  'font-face-name',
+  'missing-glyph'
+];
+
+// Spec-compliant PCEN (Potential Custom Element Name) regex.
+// Extends the ASCII-only regex to include Unicode characters permitted by the spec.
+// https://html.spec.whatwg.org/multipage/custom-elements.html#valid-custom-element-name
+const customElementRegex =
+  /^[a-z](?:[a-z0-9._-]|[\u00B7\u00C0-\u00D6\u00D8-\u00F6\u00F8-\u037D\u037F-\u1FFF\u200C-\u200D\u203F-\u2040\u2070-\u218F\u2C00-\u2FEF\u3001-\uD7FF\uF900-\uFDCF\uFDF0-\uFFFD])*-(?:[a-z0-9._-]|[\u00B7\u00C0-\u00D6\u00D8-\u00F6\u00F8-\u037D\u037F-\u1FFF\u200C-\u200D\u203F-\u2040\u2070-\u218F\u2C00-\u2FEF\u3001-\uD7FF\uF900-\uFDCF\uFDF0-\uFFFD])*$/;
+
 /**
  * Test a node to see if it has a spec-conforming shadow root
  *
@@ -29,7 +49,7 @@ function isShadowRoot(node) {
     const nodeName = node.nodeName.toLowerCase();
     if (
       possibleShadowRoots.includes(nodeName) ||
-      /^[a-z][a-z0-9_.-]*-[a-z0-9_.-]*$/.test(nodeName)
+      (!reservedNames.includes(nodeName) && customElementRegex.test(nodeName))
     ) {
       return true;
     }
--- a/lib/core/utils/is-shadow-root.js
+++ b/lib/core/utils/is-shadow-root.js
@@ -1,3 +1,5 @@
+import isValidCustomElementName from './is-valid-custom-element-name';
+
 const possibleShadowRoots = [
   'article',
   'aside',
@@ -18,6 +20,7 @@ const possibleShadowRoots = [
   'section',
   'span'
 ];
+
 /**
  * Test a node to see if it has a spec-conforming shadow root
  *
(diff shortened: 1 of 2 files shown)


lib/core/utils/is-shadow-root.js
```diff
@@ -18,6 +18,26 @@ const possibleShadowRoots = [
...
+  'font-face-format',
+  'font-face-name',
+  'missing-glyph'
+];
+
+// Spec-compliant PCEN (Potential Custom Element Name) regex.
+// Extends the ASCII-only regex to include Unicode characters permitted by the spec.
+// https://html.spec.whatwg.org/multipage/custom-elements.html#valid-custom-element-name
+const customElementRegex =
+  /^[a-z](?:[a-z0-9._-]|[\u00B7\u00C0-\u00D6\u00D8-\u00F6\u00F8-\u037D\u037F-\u1FFF\u200C-\u200D\u203F-\u2040\u2070-\u218F\u2C00-\u2FEF\u3001-\uD7FF\uF900-\uFDCF\uFDF0-\uFFFD])*-(?:[a-z0-9._-]|[\u00B7\u00C0-\u00D6\u00D8-\u00F6\u00F8-\u037D\u037F-\u1FFF\u200C-\u200D\u203F-\u2040\u2070-\u218F\u2C00-\u2FEF\u3001-\uD7FF\uF900-\uFDCF\uFDF0-\uFFFD])*$/;
```
straker, line 39: According to the [custom element name spec](https://html.spec.whatwg.org/#valid-custom-element-name), the name must be a valid element local name, which the [spec gives a regex](https://dom.spec.whatwg.org/#valid-element-local-name) as `/^(?:[A-Za-z][^\0\t\n\f\r\u0020/>]*|[:_\u0080-\u{10FFFF}][A-Za-z0-9-.:_\u0080-\u{10FFFF}]*)$/u`. I assume we could use that with a modification to match the custom element spec of requiring an alpha lower for the 0th char, no alpha upper chars, and a hyphen. Wouldn't this regex work? `/^[a-z][A-Za-z0-9-.:_\u0080-\u{10FFFF}]*-[A-Za-z0-9-.:_\u0080-\u{10FFFF}]*$/u`

Tested that `customElements.define('c-')` is a valid custom element name which means we don't need anything after the hyphen either.

straker: I'd also add a link to the valid element local name spec as well

straker (commented):
Thanks for the pr. A suggestions about the regex.

lib/core/utils/is-shadow-root.js
bwyard: Thanks yeah I'll take a look a little later this evening and read through your suggestions and update the pr to match accordingly

bwyard: Got real sick I'll take a look this weekend

WilcoFiers: @bwyard Hope you feel better. Checking in just to see this PR is still alive.

bwyard:
Yes it is. Im starting to turn the corner think I'll be able to do it this coming weekend. Had a Really bad health spat

straker:
@bwyard hope you get better! Just a heads up, I created https://github.com/dequelabs/axe-core/pull/5097 as I needed it in another piece of code we've been working on. So you should be able to call that function in place of the regex we have in `isShadowRoot`.

lib/core/utils/is-shadow-root.js
bwyard: Sorry this took a while, chronic illness can be rough and I'm starting to feel much better. Thanks for being patient. Read through the thread. Went with `isValidCustomElementName` from #5097 per your suggestion -- `isShadowRoot` now just calls `isValidCustomElementName(nodeName)`. The inline regex and `reservedNames` list are removed. Rebased onto develop and pushed.

straker (approved):
Thanks for updating that!

Reviewed for security.

PR #5060, opened 2026-04-07
fix(core): prevent ping timeout during long gather steps

### LINKED ISSUE #4831: Axe:ping will time out for long gather steps
Axe-core uses setTimeout(..., 0) between rules to give the browser an opportunity to do other things while axe is running. This is necessary as without it the page would become completely unresponsive. Firefox even raises an error on scripts that take too long to run (or used to, it has been a few years since I looked).

We **only** do this after completing a rule. But there are other parts of axe-core that can take multiple seconds on especially large pages. I ran into this problem with the gather step in particular. A very large page can take multiple seconds to gather nodes. This blocks the vent loop for multiple seconds, which is a problem for cross-frame testing.

What happens is this:

1. Axe sends a message to frames, starting with a ping to check if axe is even in that frame, this times out after 500ms by default
2. Axe starts gathering nodes in the top-window
3. In the frame, the ping message is received, but because gather is running in the same thread, the browser has to wait executing the handler
4. Several seconds later gather completes, now unblocking other scripts from executing
5. In the frame, the event listener finally gets the ping and sends a message back
6. The ping timeout runs, logging the frame has timed out
7. The response to the ping is received, but too late

## Steps to reproduce

1. Go to https://qateam.dequecloud.com/demos/dynamic/index.html?profiles=5000
2. Run the axe extension
3. Check the console, it will have a "Uncaught (in promise) No response received from frame within timeout of 5000ms." error

This is 5 seconds, not 500ms. Axe extension already overrides axe-core's ping, but that isn't enough on this page.

Krishnachaitanyakc:
## Summary

Fixes the cross-frame ping timeout that occurs on large pages when the synchronous gather step blocks the event loop for multiple seconds, preventing iframe ping responses from being processed.

Two complementary changes:

- **Time-budgeted queue yielding**: Enhanced `queue()` to accept an optional `{ timeBudget }` parameter. When cumulative synchronous execution time exceeds the budget (50ms), the queue yields to the event loop via `setTimeout(0)` before running the next task. Applied to rule execution queues in `audit.run()`, this allows pending `postMessage` events (including iframe ping responses) to be processed between rule gather steps.

- **Yield before `audit.run()` when iframes are present**: Wraps the `audit.run()` call in `setTimeout(0)` when cross-frame testing is active, giving the event loop a tick to process pending ping responses before heavy gather work begins. No yield overhead when iframes are absent.

### Implementation details

- `queue.js`: Added `timeBudget` option, `pendingYield` flag to prevent duplicate `pop()` calls during yield, `aborted` flag to prevent yielded tasks from running after `abort()`. Uses `Date.now()` for universal compatibility.
- `run-rules.js`: Conditional `setTimeout(0)` only when `context.frames.length > 0`.
- `audit.js`: `nowRulesQueue` and `laterRulesQueue` use `{ timeBudget: 50 }`.
- 6 new tests covering: default behavior preservation, async completion, synchronous completion when budget not exceeded, error handling during yield, and abort during yield.

Closes: https://github.com/dequelabs/axe-core/issues/4831

### DIFF
--- a/lib/core/base/audit.js
+++ b/lib/core/base/audit.js
@@ -281,8 +281,11 @@ export default class Audit {
     const allRulesToRun = getRulesToRun(this.rules, context, options);
     const runNowRules = allRulesToRun.now;
     const runLaterRules = allRulesToRun.later;
-    // init a NOW queue for rules to run immediately
-    const nowRulesQueue = queue();
+    // init a NOW queue for rules to run immediately.
+    // Use a time budget to yield to the event loop periodically,
+    // preventing long gather steps from blocking iframe communication.
+    // See: #4831
+    const nowRulesQueue = queue({ timeBudget: 50 });
     // construct can run NOW rules into NOW queue
     runNowRules.forEach(rule => {
       nowRulesQueue.defer(getDefferedRule(rule, context, options));
@@ -341,7 +344,7 @@ export default class Audit {
           return;
         }
         // init a LATER queue for rules that are dependant on preloaded assets
-        const laterRulesQueue = queue();
+        const laterRulesQueue = queue({ timeBudget: 50 });
         runLaterRules.forEach(rule => {
           const deferredRule = getDefferedRule(rule, context, options);
           laterRulesQueue.defer(deferredRule);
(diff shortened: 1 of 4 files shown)


straker:
Thanks for the pr. Unfortunately I don't think this is a problem we need to address in axe-core per say. I'll get with @WilcoFiers and see what we should do about it as it only happens in the extension.

PR #5062, opened 2026-04-08
refactor(frame-messenger): Guard against inherited properties as topics and channelIds

RinZ27:
Updated `frame-messenger` and `respondable` modules to use `Object.create(null)` for message and topic handler stores. 

### Description
The current implementation uses plain objects (`{}`) which are susceptible to prototype pollution if untrusted strings like `"__proto__"` are passed as `channelId` or `topic`. By using `Object.create(null)`, these stores become "pure" maps without a prototype, ensuring security and robustness in cross-frame communication.

This change is backward compatible and adheres to standard security practices for high-performance JavaScript libraries.

Closes: #5062

### DIFF
--- a/lib/core/utils/frame-messenger/channel-store.js
+++ b/lib/core/utils/frame-messenger/channel-store.js
@@ -8,14 +8,16 @@ export function storeReplyHandler(
   sendToParent = true
 ) {
   assert(
-    !channels[channelId],
+    !Object.prototype.hasOwnProperty.call(channels, channelId),
     `A replyHandler already exists for this message channel.`
   );
   channels[channelId] = { replyHandler, sendToParent };
 }
 
 export function getReplyHandler(channelId) {
-  return channels[channelId];
+  return Object.prototype.hasOwnProperty.call(channels, channelId)
+    ? channels[channelId]
+    : undefined;
 }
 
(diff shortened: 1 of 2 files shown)


straker (commented):
Thanks for the pr. After looking at the code I don't think the way the code is currently written there is any danger of prototype pollution in the frame messenger code. In both cases no untrusted string is ever used as a write key on either object, and the read-only lookups that could be passed on inherited properties (`constructor`, `toString,` etc.) are either harmless or filtered out by existing guards (`!topicHandler`). 

If we wanted to guard against inherited properties as topics just to be on the safe side, then we could do a `hasOwnProperty` lookup.

```js
const topicHandler = Object.prototype.hasOwnProperty.call(topicHandlers, topic)
    ? topicHandlers[topic]
    : undefined;
```

RinZ27:
@straker Appreciate the feedback. I've updated the PR to use the `hasOwnProperty` guard instead of `Object.create(null)` for both `channels` and `topicHandlers`. This keeps the implementation simpler and consistent with the existing codebase while still addressing the potential prototype pollution concerns.

straker (approved):
Thanks for the pr.

Reviewed for security

PR #5066, opened 2026-04-09
fix(target-size): ignore position: fixed elements that are offscreen when page is scrolled

### LINKED ISSUE #5065: target-size should ignore `position: fixed` elements outside the viewport when the page is scrolled
When an element is `position: fixed` and positioned above the viewport of the page, and the page is scrolled, target-size no longer ignores them and reports violations on them. However these elements do not report violations when the page is not scrolled.

Currently a `position: fixed` element is ignored for target-size if it's above the viewport. This is because when creating the grid we check if the element `isVisibleOnScreen` before adding it to the grid. `isVisibleOnScreen` determines that the element is above the viewport so does not add it to the grid. But when the page is scrolled the `getElementCoordinates` function (called by `isOffscreen`) adds the page scroll to the elements position. This causes the element to have a position within the viewport and thus be added to the grid. 

If you take the following code and run `target-size` at the top of the page, both buttons report as passing (the button above the viewport is correctly ignored when calculating `target-offset`). However if you scroll to the bottom and then run `target-size` both buttons return as violations as the Offscreen button is now considered to be inside the viewport.

```html
<!doctype html>
<html lang="en">
  <body style="height: 5000px">
    <p><button style="position: fixed; top: -15px; height: 10px">Offscreen button</button></p>
    <p><button style="position: fixed; top: 0; height: 10px">Too-small visible button</button></p>
  </body>
</html>
```

straker:
closes #5065


lib/commons/dom/is-fixed-position.js
```diff
@@ -0,0 +1,45 @@
...
+ * @param {Boolean} [options.skipAncestors] If the ancestor tree should not be used
+ * @return {Boolean} The element's position state
+ */
+export default function isFixedPosition(node, { skipAncestors } = {}) {
+  const { vNode } = nodeLookup(node);
+
+  // detached element
+  if (!vNode) {
+    return false;
+  }
```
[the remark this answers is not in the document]
WilcoFiers: Not necessary. We can assume setup was run in commons.

test/integration/full/target-size/fixed-scroll.js
```diff
@@ -0,0 +1,46 @@
...
+      const context = {
+        // ignore the mocha links
+        exclude: '#mocha'
+      };
+      axe.run(context, options, (err, r) => {
+        if (err) {
+          return done(err);
+        }
+        results = r;
+        console.log(results);
```
Garbee: Should we leave this log in? It is only a test, but still adds cruft to the output data.

lib/commons/dom/is-offscreen.js
```diff
@@ -37,42 +38,48 @@ function isOffscreen(element, { isAncestor } = {}) {
...
     (noParentScrolled(domNode, coords.bottom) || styl.position === 'absolute')
   ) {
     return true;
   }
 
-  if (coords.left === 0 && coords.right === 0) {
-    //This is an edge case, an empty (zero-width) element that isn't positioned 'off screen'.
-    return false;
+  if (isFixed && coords.top >= window.innerHeight) {
+    return true; // Fixed above the viewport
```
Garbee: Is `coords.top >= window.innerHeight` not determining if the element is *below* the Viewport? Is the comment wrong, the code, or my understanding of it?

lib/commons/dom/is-offscreen.js
```diff
@@ -37,42 +38,48 @@ function isOffscreen(element, { isAncestor } = {}) {
...
+  }
+
+  if (isFixed && coords.left >= window.innerWidth) {
```
Garbee: Was RTL mode considered here? This feels like in that context things might slip through.

test/integration/full/target-size/fixed-scroll.js
WilcoFiers: Definitely not. 🤨 I saw that and didn't even question it.

lib/commons/dom/is-offscreen.js
WilcoFiers: Yeah, comment is wrong.

lib/commons/dom/is-offscreen.js
WilcoFiers: I had, twice. But thinking about it a third time there is a little edge case here we missed. Cheers!

test/integration/full/target-size/fixed-scroll.js
WilcoFiers: Looks like Steve copied it. We have it in a couple more places. I'll just clean those up as well.

lib/commons/dom/is-offscreen.js
```diff
@@ -37,42 +38,48 @@ function isOffscreen(element, { isAncestor } = {}) {
...
+  }
+  if (coords.left === 0 && coords.right === 0) {
+    return false;
+  }
 
-  // bottom edge beyond
   if (
-    coords.bottom < 0 &&
+    coords.bottom <= 0 &&
     (noParentScrolled(domNode, coords.bottom) || styl.position === 'absolute')
```
chutchins25, line 61: Do we need to bypass this for fixed elements?

chutchins25 (commented):
One question for you. Everything else was already fixed or found by Garbee.

lib/commons/dom/is-offscreen.js
Garbee: What situation would propose that we should?

WilcoFiers: Yeah this code is buggy. It's not a new problem. I looked at fixing it while I was in here but it's too complex. I don't want to rush into it so I opened an issue for it instead https://github.com/dequelabs/axe-core/issues/5069

chutchins25 (approved):
Looks good to me.

WilcoFiers (approved):

PR #5068, opened 2026-04-10
chore: code mod to refactor test dir

WilcoFiers:
We need to clean up that old test dir. Figured I'd have Cursor generate a code mod for us to do it.

Here's what I did for this:

1. Had Cursor build a code mod for me that does:
    - Swap var with let/const
    - Replace anonymous function with arrow functions where possible
    - Replace multi-line single quote strings with backtick strings
    - Added html template tag
    - Pulled out shadow dom test toggles
2. I committed that code
3. Ran it, and committed the results of that in https://github.com/dequelabs/axe-core/pull/5068/changes/4005683b19fb72c14e4eee2d1c081fe7440e01d1
4. Went through issues that that caused and corrected them in subsequent commits

## For review

Focus on the code mod and the manual corrections I did afterwards. A 40k+ change is far too much to review manually. But the code mod and the changes after should be. The code mod isn't perfect, but it's one-off code so I didn't think that mattered.


test/checks/aria/required-parent.js
```diff
@@ -1,19 +1,18 @@
...
-  var shadowSupported = axe.testUtils.shadowSupport.v1;
-  var checkContext = axe.testUtils.MockCheckContext();
-  var checkSetup = axe.testUtils.checkSetup;
+  const fixture = document.getElementById('fixture');
+  const checkContext = axe.testUtils.MockCheckContext();
+  const checkSetup = axe.testUtils.checkSetup;
 
-  afterEach(function () {
+  afterEach(() => {
     fixture.innerHTML = '';
```
straker, line 9: `fixture.innerHTML = ''` and `axe._tree = undefined` should be removed from the `after/beforeEach` as the main `afterEach` now does this.

test/checks/aria/required-parent.js
```diff
@@ -24,47 +23,45 @@ describe('aria-required-parent', function () {
     assert.deepEqual(checkContext._data, ['list']);
   });
 
-  (shadowSupported ? it : xit)(
-    'should detect missing required parent across shadow boundary',
-    function () {
-      fixture.innerHTML = '<div id="target"></div>';
+  it('should detect missing required parent across shadow boundary', () => {
+    fixture.innerHTML = '<div id="target"></div>';
```
straker, line 27: I would like to see this use the new `shadowCheckSetup` rather than set up the tree itself.

test/checks/aria/unsupportedrole.js
```diff
@@ -1,15 +1,13 @@
...
-  var checkSetup = axe.testUtils.checkSetup;
-  var check = checks.unsupportedrole;
-  afterEach(function () {
+describe('unsupportedrole', () => {
+  const checkContext = axe.testUtils.MockCheckContext();
+  const checkSetup = axe.testUtils.checkSetup;
+  const check = checks.unsupportedrole;
+  afterEach(() => {
     checkContext.reset();
     axe.reset();
```
straker, line 7: I wonder if we should call `axe.reset` in the main `afterEach` so individual tests don't have to.

test/checks/keyboard/page-has-elm.js
```diff
@@ -106,103 +113,103 @@ describe('page-has-*', function () {
     });
   });
 
-  describe('page-has-main', function () {
-    var check = checks['page-has-main'];
+  describe('page-has-main', () => {
+    const check = checks['page-has-main'];
```
straker, line 117: This should use `axe.testUtils.getCheckEvaluate` instead of accessing the global `checks`. Also means all the `chec.evaluate.apply` should update as well

test/checks/label/hidden-explicit-label.js
```diff
@@ -82,15 +73,15 @@ describe('hidden-explicit-label', function () {
         html += '  <input id="target" title="Hi">';
         html += '</div>';
```
straker, line 74: This should also use the `html` and string template

test/core/utils/get-selector.js
```diff
@@ -1,112 +1,105 @@
...
-  var shadowSupported = axe.testUtils.shadowSupport.v1;
-  var fixtureSetup = axe.testUtils.fixtureSetup;
+  const fixture = document.getElementById('fixture');
+  const fixtureSetup = axe.testUtils.fixtureSetup;
 
-  afterEach(function () {
+  afterEach(() => {
     fixture.innerHTML = '';
     axe._tree = undefined;
     axe._selectorData = undefined;
```
straker, line 44: `axe._selectorData = undefined` can be removed from the `after/beforeEach` as well

test/core/utils/pollyfills.elements-from-point.js
```diff
@@ -64,89 +64,127 @@ describe('document.elementsFromPoint pollyfills', function () {
...
+      fixture.innerHTML = html`
+        <div
+          id="container"
+          style="position: absolute; top: 0px; left: 0px; height: 100px; 
+        width: 90px; background-color: rgba(0, 128, 0, 0.5);"
+        >
+          <div
+            id="pos"
+            style="position: absolute; top: 50px; left: 40px; height: 40px; 
+        width: 30px; background-color: rgba(0, 128, 0, 0.5);"
```
straker, line 82: This style multiline split makes this harder to read than if it was all one line

test/core/utils/qsa.js
```diff
@@ -53,87 +51,92 @@ function getTestDom() {
   return tree;
 }
 
-describe('axe.utils.querySelectorAllFilter', function () {
-  'use strict';
-  var dom;
-  afterEach(function () {});
+describe('axe.utils.querySelectorAllFilter', () => {
+  let dom;
+  afterEach(() => {});
```
straker, line 56: This could be removed

straker (dismissed):
Reviewed by tanking a randomly sampling of the changes to verify their correctness (a la release audit), looking for things that I would normally look at when doing a conversion. I don't think we need to commit the codemode file

test/checks/aria/required-parent.js
WilcoFiers: I think this PR is large enough as it is. Happy to do a follow-up.

test/checks/aria/required-parent.js
WilcoFiers: Another day.

test/checks/aria/unsupportedrole.js
WilcoFiers: Fine with me, but separate PR please.

test/checks/keyboard/page-has-elm.js
WilcoFiers: Makes sense, but separate PR plz.

test/core/utils/pollyfills.elements-from-point.js
WilcoFiers: Agreed, but this was on two lines before the codemod too. We can tweak this kind of stuff later.

test/core/utils/get-selector.js
WilcoFiers: Separate PR plz.

test/checks/label/hidden-explicit-label.js
WilcoFiers: Separate PR plz

test/core/utils/qsa.js
WilcoFiers: Separate PR plz.

straker (approved):

PR #5070, opened 2026-04-13
chore(release): v4.11.3

### LINKED ISSUE #5065: target-size should ignore `position: fixed` elements outside the viewport when the page is scrolled

straker:
### Bug Fixes

- **aria-allowed-attr:** restrict br and wbr elements to aria-hidden only ([#4974](https://github.com/dequelabs/axe-core/issues/4974)) ([1d80163](https://github.com/dequelabs/axe-core/commit/1d801636f058f2abd885c488baff954872b13846))
- **target-size:** ignore position: fixed elements that are offscreen when page is scrolled ([#5066](https://github.com/dequelabs/axe-core/issues/5066)) ([5906273](https://github.com/dequelabs/axe-core/commit/5906273841cbd7ac9e08af730dffc244cf42b39b)), closes [#5065](https://github.com/dequelabs/axe-core/issues/5065)


WilcoFiers (approved):
## Release audit

- [x] https://github.com/dequelabs/axe-core/pull/5066
- [x] https://github.com/dequelabs/axe-core/pull/5062

dylanb:
reviewed

PR #5075, opened 2026-04-14
fix(utils): Add null check to parseCrossOriginStylesheet, closes #5074

### LINKED ISSUE #5074: Error when encountering cross origin style sheet with null href
### Product

axe-core

### Product Version

_No response_

### Latest Version

#### Expectation
I am using Storybook with the addon-a11y, which uses axe.
For some reason, Storybook is inserting a style sheet which appears to be cross origin, but which is empty and has no href.
`<style type="text/css" id="dark-mode-custom-style"></style>`
I would generally expect Axe to ignore this style sheet, since it is empty and has no url.

#### Actual
Axe tries to load `sheet.href`, which is null, and that throws an error in the console. 

Image

#### How to Reproduce
Generally, the above. It seems to occur in any basic Storybook setup that includes their `addon-a11y`. 

#### Additional context
The issue seems to warrant a change to `parseStylesheet` or `parseCrossOriginStylesheet`. Those methods return a promise, but generally it seems like they should return an empty promise or return early if encountering a cross-origin stylesheet that has no href.

I may open a PR that tries to address this issue by returning `Promise.resolve()` in this case, but I do not know for certain if this is a safe fix.

Stack Trace from the browser network tab:

Image

dcbroad3:
If a null href is passed to `parseCrossOriginStylesheet`, the function will currently convert that to the string "null" and attempt to load that href. This PR adds a null check so that the function instead returns early if a nullish href is passed, similar to how `parseSameOriginStylesheet` returns early if a sheet includes no rules.

Closes:
#5074


lib/core/utils/parse-crossorigin-stylesheet.js
```diff
@@ -20,6 +20,10 @@ function parseCrossOriginStylesheet(
   importedUrls,
   isCrossOrigin
 ) {
+  if (url == null) {
+    return Promise.resolve();
```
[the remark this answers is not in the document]
dcbroad3: I am following the pattern of the early return in `parseSameOriginStylesheet` if `rules` is falsy, but let me know if a different result should be returned (or if the function should do an early reject instead?)

straker (approved):

lib/core/utils/parse-crossorigin-stylesheet.js
```diff
@@ -20,6 +20,10 @@ function parseCrossOriginStylesheet(
   importedUrls,
   isCrossOrigin
 ) {
+  if (url == null) {
```
straker: Actually thinking about this I wonder if we should reject the promise here instead of returning. If the cross origin stylesheet somehow has no href but has styles that are applying to the page, we should reject to signify we don't have the full styles. Let me ask my team about what they think.

straker (dismissed):
We determined that this will probably only happen with extensions adding styles to the page, so we are fine resolving the promise since we want to test the page itself not the styles added by the extension.

lib/core/utils/parse-crossorigin-stylesheet.js
straker: Linting is failing as this should use `===` instead of `==`

dcbroad3: You guys don't allow nullish checks? That makes me sad. Will update.

WilcoFiers: @dcbroad3 Any updates here? Were going to feature freeze for axe-core 4.12 today. If you can get this addressed we can get it in.

dcbroad3: Ah, apologies, this slipped my mind. I'll try to get this addressed this morning.

dcbroad3: @straker @WilcoFiers Updated.. I hope this is what you meant. (`x == null` is functionally equivalent to `x === null || x === undefined`)

straker (approved):
Thanks for the pr!

Reviewed for security.

PR #5076, opened 2026-04-15
fix(commons/text): exclude natively hidden elements from aria-labelledby accessible name

### LINKED ISSUE #4704: accessibleText incorrectly handles aria-labelledby pointing towards <style> elements
### Product

axe-core

### Product Version

4.10.2

### Latest Version

### Issue Description

The `accessibleText` utility doesn't correctly handle cases where the target element is `aria-labelledby` an `aria-hidden` element that contains a `<style>` element. We noticed this in practice against some of the icon buttons at the top-right of https://github.com, which use a pattern along these lines:

```html
<a href="..." aria-labelledby="tooltip-1"><img ...></a>
<tool-tip id="tooltip-1" aria-hidden="true">
  #shadow-root (open)
    <style>.stuff {}</style>
    <slot></slot>
  Tooltip content
</tool-tip>
```

...and the accessible text comes back like `".stuff {} Tooltip content"` instead of the expected `"Tooltip content"`.

#### Expectation
The 2 new `/test/commons/text/accessible-text.js` test cases below should pass

#### Actual
They fail

#### How to Reproduce
```js
  it('should not use <style> contents as a label, even if explicitly referenced with labelledby', () => {
    fixture.innerHTML = html`
      <div id="t1" aria-labelledby="style-id"></div>
      <style id="style-id">.shouldnt-be-part-of-accessible-text {}</style>
    `

    const target = axe.utils.querySelectorAll(axe._tree, '#t1')[0];
    assert.equal(axe.commons.text.accessibleTextVirtual(target), '');
  });

  it('should not use <style> contents as a label, even if labelledby explicitly references a hidden container', () => {
    fixture.innerHTML = html`
      <div id="t1" aria-labelledby="hidden-container"></div>
      <div id="hidden-container" aria-hidden="true">
        <style>.shouldnt-be-part-of-accessible-text {}</style>
        Expected label
      </div>
    `

    const target = axe.utils.querySelectorAll(axe._tree, '#t1')[0];
    assert.equal(axe.commons.text.accessibleTextVirtual(target), 'Expected label');
  });
```

#### Additional context
I already double checked in Chrome/Firefox/Safari that the browsers don't include the `<style>` contents in their calculations here.

nami8824:
Fixes accessible name calculation to correctly exclude `<style>` contents, as well as `<script>`, `<noscript>`, and `<template>`, when referenced via `aria-labelledby`.

Closes #4704

### DIFF
--- a/test/commons/text/accessible-text.js
+++ b/test/commons/text/accessible-text.js
@@ -1649,6 +1649,38 @@ describe('text.accessibleTextVirtual', () => {
     });
   });
 
+  describe('natively hidden elements', () => {
+    const tags = ['style', 'script', 'noscript', 'template'];
+
+    it('should not use content as accessible name when directly referenced by aria-labelledby', () => {
+      tags.forEach(tag => {
+        fixture.innerHTML = `<div id="t1" aria-labelledby="el-id"></div><${tag} id="el-id">content</${tag}>`;
+        axe.testUtils.flatTreeSetup(fixture);
+        const target = axe.utils.querySelectorAll(axe._tree, '#t1')[0];
+        assert.equal(axe.commons.text.accessibleTextVirtual(target), '', tag);
+      });
+    });
+
+    it('should not use content when inside a hidden container referenced by aria-labelledby', () => {
+      tags.forEach(tag => {
+        fixture.innerHTML = `
+          <div id="t1" aria-labelledby="hidden-container"></div>
+          <div id="hidden-container" aria-hidden="true">
+            <${tag}>content</${tag}>
+            Expected label
+          </div>
+        `;
+        axe.testUtils.flatTreeSetup(fixture);
+        const target = axe.utils.querySelectorAll(axe._tree, '#t1')[0];
+        assert.equal(
+          axe.commons.text.accessibleTextVirtual(target),
+          'Expected label',
+          tag
+        );
+      });
+    });
+  });
+
   describe('text.accessibleText acceptance tests', () => {
     'use strict';
     // Tests borrowed from the AccName 1.1 testing docs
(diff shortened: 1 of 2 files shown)


test/commons/text/accessible-text.js
```diff
@@ -1649,6 +1649,38 @@ describe('text.accessibleTextVirtual', () => {
...
+
+    it('should not use content as accessible name when directly referenced by aria-labelledby', () => {
+      tags.forEach(tag => {
+        fixture.innerHTML = `<div id="t1" aria-labelledby="el-id"></div><${tag} id="el-id">content</${tag}>`;
+        axe.testUtils.flatTreeSetup(fixture);
+        const target = axe.utils.querySelectorAll(axe._tree, '#t1')[0];
+        assert.equal(axe.commons.text.accessibleTextVirtual(target), '', tag);
+      });
+    });
+
```
WilcoFiers, line 1663: I think this is worth having a test for too. This isn't about hiding content. The element itself should be ignored. There are elements with hidden content that aren't themselves hidden (canvas, iframe). The fix here is specifically about hiding the element.

```suggestion

    it('should ignore aria-label as accessible name when directly referenced by aria-labelledby', () => {
      tags.forEach(tag => {
        fixture.innerHTML = `<div id="t1" aria-labelledby="el-id"></div><${tag} id="el-id" aria-label="aria-label"></${tag}>`;
        axe.testUtils.flatTreeSetup(fixture);
        const target = axe.utils.querySelectorAll(axe._tree, '#t1')[0];
        assert.equal(axe.commons.text.accessibleTextVirtual(target), '', tag);
      });
    });

```

WilcoFiers (changes requested):
Looks solid. I'd like to see one test case added.

test/commons/text/accessible-text.js
nami8824: Added the test case.

WilcoFiers (approved):
Excellent. Thank you so much @nami8824!


For the books; PR reviewed for security.

PR #5077, opened 2026-04-20
feat(utils): add `getElementInternals` function

### LINKED ISSUE #5038: ElementInternals: Add `getElementInternals` function and tests
Add the `getElementInternals` function which looks at an HTMLElement and gathers all element internal properties. Add tests for all use cases.

Since element internals is private and there is no API to get the internals object, we've decided to use a https://github.com/dequelabs/axe-core/issues/4259#issuecomment-2665835385. In essence this is a set of documented techniques that the community would have to implement on their custom elements in order for axe-core to access the element's internals.

```js
const propNames = ['_internals', 'internals', 'internals_'];
const symbolNames = ['internals', 'privateInternals'];

function getElementInternals(node) {
  // we could break early if we detect node is not a custom element

  if (globalThis.axeInternalsMap?.get(node)) {
    return globalThis.axeInternalsMap.get(node);
  }

  for (propName of propNames) {
    if (Object.getOwnPropertyDescriptor(node, propName).get) {
      return;
    }
    if (node[propName] instanceof window.ElementInternals) {
      return node[propName];
    }
  }

  const ownSymbols = Object.getOwnPropertySymbols(node);
  if (!ownSymbols.length) {
    return;
  }

  for (symbolName of symbolNames) {
    const symbol = ownSymbols.find(s => s.description === symbolName);

    if (symbol) {
      if (Object.getOwnPropertyDescriptor(node, symbol).get) {
        return;
      }
      if (node[symbol] instanceof window.ElementInternals) {
        return node[symbol];
      }
    }
  }
}
```

straker:
Closes: https://github.com/dequelabs/axe-core/issues/5038

### DIFF
--- a/lib/core/utils/get-element-internals.js
+++ b/lib/core/utils/get-element-internals.js
@@ -38,7 +38,7 @@ export default function getElementInternals(node) {
 
   for (const propName of propNames) {
     if (Object.getOwnPropertyDescriptor(node, propName)?.get) {
-      return;
+      continue;
     }
     if (node[propName] instanceof window.ElementInternals) {
       return node[propName];
@@ -55,7 +55,7 @@ export default function getElementInternals(node) {
 
     if (symbol) {
       if (Object.getOwnPropertyDescriptor(node, symbol)?.get) {
-        return;
+        continue;
       }
       if (node[symbol] instanceof window.ElementInternals) {
         return node[symbol];
--- a/lib/core/utils/get-element-internals.js
+++ b/lib/core/utils/get-element-internals.js
@@ -13,7 +13,7 @@ const symbolNames = ['internals', 'privateInternals'];
 /**
  * Get the ElementInternals object for a custom element.
  * @param  {HTMLElement} node
- * @return {ElementInternal|undefined}
+ * @return {ElementInternals|undefined}
  */
 export default function getElementInternals(node) {
   // support finding internals from an optional global map. we assume that if a node is set here the value will be an ElementInternals object
--- a/lib/core/utils/get-element-internals.js
+++ b/lib/core/utils/get-element-internals.js
@@ -1,0 +1,46 @@
+/**
+ * community protocols that axe-core supports to find element internals:
+ * - globalThis.axeInternalsMap.get(node)
+ * - node._internals
+ * - node.internals
+ * - node.internals_
+ * - node[Symbol('internals')]
+ * - node[Symbol('privateInternals')]
+ **/
+const propNames = ['_internals', 'internals', 'internals_'];
+const symbolNames = ['internals', 'privateInternals'];
+
+export default function getElementInternals(node) {
+  // support finding internals from an optional global map
+  // e.g. globalThis.axeInternalsMap.set(node, internals)
+  if (globalThis.axeInternalsMap?.get(node)) {
+    return globalThis.axeInternalsMap.get(node);
+  }
+
+  for (const propName of propNames) {
+    if (Object.getOwnPropertyDescriptor(node, propName)?.get) {
+      return;
+    }
+    if (node[propName] instanceof window.ElementInternals) {
+      return node[propName];
+    }
+  }
+
+  const ownSymbols = Object.getOwnPropertySymbols(node);
+  if (!ownSymbols.length) {
+    return;
+  }
+
+  for (const symbolName of symbolNames) {
+    const symbol = ownSymbols.find(s => s.description === symbolName);
+
+    if (symbol) {
+      if (Object.getOwnPropertyDescriptor(node, symbol)?.get) {
+        return;
+      }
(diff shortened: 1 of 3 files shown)


lib/core/utils/get-element-internals.js
```diff
@@ -0,0 +1,55 @@
...
+  // ie11 guard
+  if (!('ElementInternals' in window)) {
+    return;
+  }
+
+  for (const propName of propNames) {
+    if (Object.getOwnPropertyDescriptor(node, propName)?.get) {
+      return;
+    }
+    if (node[propName] instanceof window.ElementInternals) {
```
chutchins25, line 43: Question: is the "bail out of the entire lookup on any getter" behavior intentional, or should a getter only disqualify the specific protocol slot it appears on?

As written, if `node.internals` happens to be defined as a getter (via a mixin, base class, or unrelated library), `_internals`, `internals_`, and both `Symbol(...)` protocols are never checked — even when one of them holds a valid `ElementInternals`. The same applies in the symbol loop at lines 44–47 (a getter on `Symbol('internals')` skips `Symbol('privateInternals')`).

If it's meant to be a hard bail-out, a short comment explaining why would help future readers (defense against side-effectful accessors?). If not, swapping `return;` for `continue;` in both loops keeps the search going against the remaining protocols. Either way, a test covering the cross-protocol overlap case (e.g. `internals` getter + valid `_internals` prop) would lock the intended behavior in.

lib/core/utils/get-element-internals.js
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Get the ElementInternals object for a custom element.
+ * @param  {HTMLElement} node
+ * @return {ElementInternals|undefined}
+ */
+export default function getElementInternals(node) {
+  // support finding internals from an optional global map. we assume that if a node is set here the value will be an ElementInternals object
+  if (globalThis.axeInternalsMap?.get(node)) {
+    return globalThis.axeInternalsMap.get(node);
+  }
```
chutchins25: Small tweak suggestion, not a show-stopper: `axeInternalsMap.get(node)` is called twice here. Capturing the result in a local reads a bit cleaner and avoids the double lookup:

```js
const fromMap = globalThis.axeInternalsMap?.get(node);
if (fromMap) return fromMap;
```

test/core/utils/get-element-internals.js
```diff
@@ -0,0 +1,124 @@
+describe('utils.getElementInternals', () => {
```
chutchins25, line 1: Two coverage gaps worth adding while you're here:

1. **Protocol precedence** — if a node has both `_internals` (prop) and `Symbol('internals')` set, which wins? The code implies prop-first, but it isn't asserted.
2. **Invalid input** — `getElementInternals(null)` / `undefined` will throw at `Object.getOwnPropertyDescriptor(null, ...)` once past the global-map branch. Worth deciding whether to guard or document the precondition, and locking it in with a test either way.

lib/core/utils/get-element-internals.js
```diff
@@ -0,0 +1,55 @@
+/**
+ * community protocols that axe-core supports to find element internals:
+ * - globalThis.axeInternalsMap.get(node)
+ * - node._internals
+ * - node.internals
+ * - node.internals_
+ * - node[Symbol('internals')]
+ * - node[Symbol('privateInternals')]
+ **/
```
chutchins25, line 9: Since `axeInternalsMap` is the one part of this API that requires explicit consumer setup, consider expanding the doc block to cover:

- **Recommend `WeakMap`** — the tests use `WeakMap`, and that matters. Consumers who reach for a regular `Map` will leak detached custom-element nodes.
- **Show the setup** — a one-line example makes the contract clear:

  ```js
  globalThis.axeInternalsMap ??= new WeakMap();
  globalThis.axeInternalsMap.set(this, this.attachInternals());
  ```

This is the first introduction of the protocol, so it's a natural place for the guidance to live.

chutchins25 (commented):
Nice work setting up the community-protocol foundation — coverage is solid and the code is small and focused. A few questions and polish suggestions on the implementation and test surface below. No blockers.

test/core/utils/get-element-internals.js
straker: 1. unlike the global map precedence, the props don't really have a precedence and we wouldn't expect multiple props / symbols to be set at a time
2. axe-core rarely does null / undefined input checks. we always just work under the assumption that the input is valid when called

lib/core/utils/get-element-internals.js
straker: We definitely don't want to access a getter for a prop, and since we wouldn't expect multiple props / symbols to be set at a time I think a bail is fine.

WilcoFiers (approved):

PR #5080, opened 2026-04-22
feat(aria-allowed/prohibited-attr, aria-required-parent/children): partially support element internals role

### LINKED ISSUE #4259: Custom Elements using `ElementInternals` to set `role` are flagged when `aria-label` is provided
### Product

axe-core

### Product Version

4.8.2

### Latest Version

### Issue Description

We are currently migrating our Web Component library to use the `ElementInternals` API to set default ARIA attributes, such as `role`, on all of our semantic host elements instead of sprouting these `role` attributes when the element is connected. This allows for setting default semantics, while still allowing for consumers to override these semantics if necessary. Upon doing so, axe is flagging _some_ of these custom elements as critical issues that appear to be false positives.

It's worth noting that axe will not flag these elements if any non-empty text nodes appear in the shadow DOM of the element, or if the `role` attribute is set directly on the element.

#### Expectation
Custom Elements that use `ElementInternals` to set their default `role` should not be flagged by axe-core.

#### Actual
The extension produced the following critical error:
> aria-label attribute cannot be used on a custom-progress-internals with no valid role attribute. 

#### How to Reproduce
The following codepen is a reproduction in the simplest form that shows a custom element with a default `role` of `progressbar` set via `ElementInternals`, along with an `aria-label` set in the light DOM on the host:
https://codepen.io/drifty17/full/WNPgvoM

Open the axe extension in DevTools and scan the page to see the error:
image

Upon viewing the accessibility tree you can also see that the semantics appear to be set properly (tested with VoiceOver as well and it interpreted it correctly):
image

#### Additional context
I added an additional (working) example of a separate custom element for comparison that is identical except that it will sprout the `role` attribute when it connects to the DOM. When the `role` attribute is present on the element, axe does not flag the element, but when the same `role` is set via `ElementInternals` instead it does get flagged.

### LINKED ISSUE #5039: ElementInternals: update `implicitRole` to use `getElementInternals`
Update the `implicitRole` function to use the `getElementInternals` function to get the internal role of the element, adding new tests to ensure it's working. The tests would also need to be updated in rules that look for required children/parent roles or allowed/required attributes for roles (e.g. aria-required-parent, aria-required, aria-allowed-attr and aria-required-attr) to add ElementInternal tests (similar to our virtual node tests)

straker:
Note that this can only support rules that don't select for `[role]` on the element itself, which is why we can't support `aria-required-attr` yet, or any rules that look for specific roles (e.g. `aria-command-name`). Additionally we can only test the supported rules partially as those rules look for an element with `[role]`. For example, for `aria-required-parent` we can test that the parent of an element with `[role]` has a required internal role, but we cannot test a child with an internal role has a required parent. 

Fully supporting rules can only be supported once we've figured out how to update our selectors to look at more than just CSS selectors (planned but not fully spec'd out yet).

Closes: #5039
Closes: #4259


test/testutils.js
```diff
@@ -32,6 +32,25 @@ var commons;
     fixture.setAttribute('id', 'fixture');
     document.body.insertBefore(fixture, document.body.firstChild);
   }
+  testUtils.fixture = fixture;
```
straker, line 35: Adding fixture to the testutils objects makes getting it a bit easier instead of having to do `const fixture = document.querySelector('#fixture')` in every test that needs it.

test/commons/aria/implicit-role.js
```diff
@@ -1,562 +1,484 @@
-describe('aria.implicitRole', function () {
-  'use strict';
-  var implicitRole = axe.commons.aria.implicitRole;
-  var flatTreeSetup = axe.testUtils.flatTreeSetup;
-  var fixture = document.querySelector('#fixture');
-
-  afterEach(function () {
-    fixture.innerHTML = '';
-  });
+describe('aria.implicitRole', () => {
```
straker, line 1: Most of this file is cleaning it up to match the new es6 style. Bottom of the file has the newest tests.

test/checks/aria/required-parent.js
```diff
@@ -1,170 +1,115 @@
-describe('aria-required-parent', function () {
-  'use strict';
+describe('aria-required-parent', () => {
```
straker, line 1: Most of this file is cleaning it up to match the new es6 style. Bottom of the file has the newest tests.

lib/checks/aria/aria-required-children-evaluate.js
```diff
@@ -103,7 +103,8 @@ function getOwnedRoles(virtualNode, required) {
     ) {
       ownedVirtual.push(...vNode.children);
     } else if (role || hasGlobalAriaOrFocusable) {
-      const attr = globalAriaAttr || 'tabindex';
+      const attr =
+        globalAriaAttr || (vNode.hasAttr('tabindex') ? 'tabindex' : undefined);
```
straker, line 107: This code assumed that if an element had a role it also had a global aria attr available, which wasn't true. This lead to a custom element with an internal role to always have `[tabindex]` added to the selector even though that attr didn't exist.

test/checks/aria/aria-allowed-attr.js
```diff
@@ -282,4 +282,31 @@ describe('aria-allowed-attr', () => {
       );
     });
   });
+
+  describe('ElementInternals', () => {
+    it('should detect incorrectly used attributes', () => {
+      const vNode = queryFixture(
+        '<testutils-element with-role="link" id="target" tabindex="1" aria-selected="true"></testutils-element>'
```
straker: See the changes to the `test/testutils.js` file for how this test custom element works.

lib/core/utils/get-element-internals.js
```diff
@@ -25,6 +27,12 @@ const symbolNames = ['internals', 'privateInternals'];
  * @return {ElementInternals|undefined}
  */
 export default function getElementInternals(node) {
+  // internals can only be attached to custom-elements
+  // trying to do otherwise results in an error HTMLElement': Unable to attach ElementInternals to non-custom elements
+  if (isHtmlElement(node)) {
```
chutchins25: Two concerns with placing this guard before the `axeInternalsMap` lookup:

1. The `axeInternalsMap` protocol is documented as a way for consumers to hand axe internals for any node — if the guard runs first, storing a built-in in the map silently no-ops.
2. Customized built-ins (`<a is="my-link">`) can call `attachInternals()` per spec, but hit the guard because their `nodeName` is in `standards.htmlElms`.

Would something like this work — map lookup first, and the guard allows customized built-ins through?

```js
  // support finding internals from an optional global map...
  const mapInternals = globalThis.axeInternalsMap?.get(node);
  if (mapInternals) {
    return mapInternals;
  }

  // internals can only be attached to custom elements or customized built-ins;
  // plain built-ins throw "Unable to attach ElementInternals to non-custom elements"
  if (isHtmlElement(node) && !node.getAttribute?.('is')) {
    return;
  }
```

Or is the intent to deliberately restrict both cases and the tradeoff was considered?

test/checks/aria/required-parent.js
```diff
@@ -1,170 +1,115 @@
...
-      var params = [shadowContent, undefined, virtualTarget];
-      assert.isFalse(
-        axe.testUtils
-          .getCheckEvaluate('aria-required-parent')
-          .apply(checkContext, params)
-      );
-      assert.deepEqual(checkContext._data, ['list']);
-    }
-  );
+  it('should detect missing required parent across shadow boundary', () => {
```
chutchins25, line 26: Not a blocker, just a consistency question — these two files lost the `(shadowSupported ? it : xit)` gating, but `errormessage.js` and `valid-attr-value.js` still have it, and `testutils.js` still defines `shadowSupport.v1`. Is there a follow-up ticket to sweep the gate out of the whole suite (and maybe retire `shadowSupport.v1`), or should the gate stay on these tests for consistency with the rest of the repo?

test/core/utils/get-element-internals.js
```diff
@@ -149,4 +149,13 @@ describe('utils.getElementInternals', () => {
 
     assert.isUndefined(getElementInternals(node));
   });
+
+  it('returns undefined for native element', () => {
+    const node = document.createElement('div');
+    // verify we don't return internals for non-native element
```
chutchins25:
```suggestion
    // verify we don't return internals for native element
```
Tiny one: the comment says "non-native" but the test creates a `<div>` (native) and asserts `isUndefined`. Was the "non-" intended here?

chutchins25 (commented):
Nice work! A few questions for you:

1. Are we ok bundling large refactors like this in with the feature work? Would it be worth splitting the ES6 cleanup into its own PR next time to keep the feature diff focused? CLAUDE.md §3 leans that way, but curious if there's a reason the bundle made sense here.

2. On naming — since `internals` lands on `VirtualNode` as public-ish surface (consumers can reach it through any VirtualNode instance), renaming it later would be a breaking change. Was `elementInternals` considered as a more explicit name, or is the short form intentional?

lib/core/utils/get-element-internals.js
straker: Good catch. I'll add some tests for built-ins

test/checks/aria/required-parent.js
straker: It's just something we'll convert over with time as edit the files that use them. All browsers we test under now fully support it so it always returns true.

straker:
> 1. Are we ok bundling large refactors like this in with the feature work? Would it be worth splitting the ES6 cleanup into its own PR next time to keep the feature diff focused? CLAUDE.md §3 leans that way, but curious if there's a reason the bundle made sense here.

Yep. Any time we edit files that use the old syntax we convert the file first to the new syntax and then add the changes. It allows us to slowly convert the files over time without having to take the time to do it on it's own.

> 2. On naming — since internals lands on VirtualNode as public-ish surface (consumers can reach it through any VirtualNode instance), renaming it later would be a breaking change. Was elementInternals considered as a more explicit name, or is the short form intentional?

I can rename it

lib/core/utils/get-element-internals.js
straker: Apparently you can't attach internals to customized built-in either:

```html
<button is="customized-button"></button>
<script>
customElements.define(
  'customized-button',
  class CustomizedButton extends HTMLButtonElement {
    constructor() {
      super();
      this._internals = this.attachInternals();
    }
  },
  { extends: 'button' }
); 

// Uncaught NotSupportedError: Failed to execute 'attachInternals' on 'HTMLElement': Unable to attach ElementInternals to a customized built-in element.
</script>
```

straker: According to spec, this is expected as the first step of the [`attachInternals` algorithm](https://html.spec.whatwg.org/multipage/custom-elements.html#element-internals) says 

> If this's `is` value is not null, then throw a `"NotSupportedError" DOMException.`

test/checks/aria/aria-prohibited-attr.js
```diff
@@ -267,4 +267,40 @@ describe('aria-prohibited-attr', () => {
       assert.isTrue(checkEvaluate.apply(checkContext, params));
     });
   });
+
+  describe('ElementInternals', () => {
+    it('should return true for prohibited attributes and no content', () => {
+      const params = checkSetup(
+        '<testutils-element with-role="code" id="target" aria-hidden="false" aria-label="foo"></div>'
```
chutchins25: Not a blocker, just a consistency question — was `</div>` intended here? The element opens with `<testutils-element>`, so the closing tag mismatches. Browsers auto-close it, but the fixture is clearer with a matching tag:

```suggestion
        '<testutils-element with-role="code" id="target" aria-hidden="false" aria-label="foo"></testutils-element>'
```

(L287 has the same pattern — also double-spaced between `aria-hidden` and `aria-label`.)

chutchins25 (commented):
Took another pass. Naming and bundling questions from the earlier review are resolved on my end — thanks for the rename. Two small fixture tag-mismatch nits noted inline; not blockers.

test/checks/aria/aria-prohibited-attr.js
chutchins25: Same `</div>` mismatch as L274 — and a double space between `aria-hidden="false"` and `aria-label="foo"`:

```suggestion
        '<testutils-element with-role="code" id="target" aria-hidden="false" aria-label="foo" aria-labelledby="foo"></testutils-element>'
```

chutchins25 (commented):
Line Numbers got messed up on previous review for suggested changes.

chutchins25 (dismissed):
Looks good to me.

lib/core/utils/get-element-internals.js
```diff
@@ -25,6 +27,12 @@ const symbolNames = ['internals', 'privateInternals'];
  * @return {ElementInternals|undefined}
  */
 export default function getElementInternals(node) {
+  // internals can only be attached to custom-elements
+  // trying to do otherwise results in an error "Cannot attach ElementInternals to a customized built-in or non-custom element"
+  if (isHtmlElement(node)) {
```
WilcoFiers: What about SVG or MathML?

lib/core/base/virtual-node/virtual-node.js
```diff
@@ -189,6 +189,16 @@ class VirtualNode extends AbstractVirtualNode {
...
     return this._cache.boundingClientRect;
   }
-}
 
-export default VirtualNode;
+  /**
+   * Return the element internals for this element and cache the result.
+   * @return {ElementInternals}
+   */
+  get elementInternals() {
```
WilcoFiers, line 197: Slight preference for your proposal's `internals` over `elementInternals`. But I can live with this.

Since we're going to need to set this prop when axe runs in an isolated world I think you need to add a setter. JIT compilers don't like it when you change the shape of an object. 

That said, I'm not sure how useful it even is doing this with a getter. Role calculation is everywhere in axe-core. We're going to end up needing to do this for just about every element. So adding a cache layer may very well just make this a bit slower.

lib/commons/aria/implicit-role.js
```diff
@@ -12,20 +11,22 @@ import AbstractVirtuaNode from '../../core/base/virtual-node/abstract-virtual-no
...
+  // such as if a user used `getFlattenedTree` manually on a
+  // subset of the DOM tree
   if (!vNode) {
     throw new ReferenceError(
       'Cannot get implicit role of a node outside the current scope.'
     );
   }
 
+  if (vNode.elementInternals && vNode.elementInternals.role) {
+    return vNode.elementInternals.role;
```
WilcoFiers:
```suggestion
  if (vNode.elementInternals?.role) {
    return vNode.elementInternals.role;
```

test/commons/aria/implicit-role.js
```diff
@@ -1,562 +1,484 @@
...
+        '<testutils-element id="target"></testutils-element>'
+      );
+      assert.equal(implicitRole(vNode), 'button');
+    });
+
+    it('returns null for element without internals', () => {
+      const vNode = queryFixture(
+        '<testutils-element id="target"></testutils-element>'
+      );
+      delete vNode.actualNode._internals;
```
WilcoFiers, line 567: Could just create a different element ;)

lib/core/base/virtual-node/virtual-node.js
```diff
@@ -189,6 +189,16 @@ class VirtualNode extends AbstractVirtualNode {
...
+    if (!this._cache.hasOwnProperty('elementInternals')) {
+      this._cache.elementInternals = getElementInternals(this.actualNode);
+    }
+
+    return this._cache.elementInternals;
```
WilcoFiers, line 208: This is going to put the codebase into an unreleasable state. We can't release this feature until we're able to make it work in isolated worlds, otherwise we're going to introduce differences in how axe behaves in different contexts. We need to lock this code behind some kind of config / feature flag until we're ready to roll this out everywhere. Maybe something like axe.configure({ features })?

Something like that should probably go into it's own PR. We don't have any plans to do a minor release at this point so I don't mind this going in now, as long as you pick that feature flag thing up next.

test/checks/aria/aria-allowed-attr.js
```diff
@@ -282,4 +282,31 @@ describe('aria-allowed-attr', () => {
       );
     });
   });
+
+  describe('ElementInternals', () => {
```
WilcoFiers: We probably didn't need these. We have unit tests for implicitRole, and integration tests for the rules affected. Unit tests for the rules is probably overkill unless you updated the rule itself (like aria-required-children).

WilcoFiers (dismissed):
Looks solid. Couple nitpicky things. Leaving it to you if you want to pick those up.

lib/core/utils/get-element-internals.js
straker: Both apparently don't define the function in any browser (Chrome, Safari, Firefox). Trying to attach internals reports it's not a function. I'm not sure how we would short circuit svg or math child elements though since we don't maintain a list of those anywhere.

WilcoFiers (approved):

PR #5082, opened 2026-04-27
fix(utils): improve robustness of frame messenger and string processing

RinZ27:
Strengthening the robustness of internal utilities ensures better stability across different environments. While reviewing the core library, I noticed several areas where string processing and frame communication could be more resilient against unexpected input.

Addressing potential logic errors in `process-message.js`, this update introduces proper regex escaping for property names. I encountered a case where special characters in metadata could interfere with the template substitution, so centralizing the escaping logic was a priority.

Updating the frame messenger's `channel-store.js` adds a necessary layer of validation for `channelId`. Ensuring only string identifiers are processed helps maintain the integrity of communication channels between frames. Additionally, I modified `uuid.js` to safely detect the `crypto` API, preventing crashes in environments where `window` is unavailable, such as Node.js or Web Workers.

Refining `get-friendly-uri-end.js` extends the ignored schemes to include `vbscript:` and `file:`. These changes collectively improve the engine's correctness when handling diverse URI formats in a security-conscious manner.

Validation of these fixes was performed locally to confirm that the existing test suite remains stable and the reported edge cases are now correctly handled.

### DIFF
--- a/lib/core/utils/escape-reg-exp.js
+++ b/lib/core/utils/escape-reg-exp.js
@@ -1,0 +1,12 @@
+/**
+ * Escape a string for use in a regular expression
+ * @private
+ * @param {String} string String to escape
+ * @return {String} Escaped string
+ */
+export default function escapeRegExp(string) {
+  /*! Credit: XRegExp 0.6.1 (c) 2007-2008 Steven Levithan <http://stevenlevithan.com/regex/xregexp/> MIT License */
+  const from = /(?=[\-\[\]{}()*+?.\\\^$|,#\s])/g;
+  const to = '\\';
+  return string.replace(from, to);
+}
--- a/lib/core/utils/frame-messenger/channel-store.js
+++ b/lib/core/utils/frame-messenger/channel-store.js
@@ -1,12 +1,16 @@
 import assert from '../assert';
 
-const channels = {};
+const channels = Object.create(null);
+const restrictedKeys = ['__proto__', 'constructor', 'prototype'];
 
 export function storeReplyHandler(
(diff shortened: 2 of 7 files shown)


straker:
Thanks for the pr. Unfortunately I don't think any of the changes improve the robustness any further than https://github.com/dequelabs/axe-core/pull/5062. If `channelId` isn't a string it's going to fail, and what we put on data can't be anything that'd need escaping so we don't need a regex escape. 

As for the `uuid` change, have you encountered this as a problem? We already run in axe in tests in node and jsdom, and we normalize different global objects to all be window. 

Unless you've encountered problems that necessitated these changes I'm inclined to close the PR as won't implement.

RinZ27:
Appreciate the feedback. These changes are intended to address specific security and environment-related edge cases:

- **Prototype Pollution:** Using `Object.create(null)` and filtering keys like `__proto__` in `channel-store.js` is a defensive measure. Since `channelId` can come from external `postMessage` sources, this prevents potential pollution of the global object prototype.
- **Web Workers:** The update to `uuid.js` allows the utility to function in Web Workers where `window` is not defined but `self.crypto` is available. This ensures we can still generate secure UUIDs in background threads.
- **Regex Injection:** Escaping the property name before passing it to `new RegExp` prevents malformed or malicious strings from breaking the regex logic or causing ReDoS, even if we currently trust the inputs.

I believe these harden the utility functions for broader use cases and prevent subtle bugs in non-DOM or high-risk environments.

straker:
Going to close as won't implement.

Issue #5064, opened 2026-04-09
`landmark-unique` does not match `<form role="search">` without an accessible name


nami8824:
#### Expectation
Two `<form role="search">` elements without an accessible name should be flagged as duplicate landmarks by the `landmark-unique` rule, in the same way that two `<div role="search">` elements are.      

#### Actual

```
<!-- Correctly flagged -->                                                                                                                                                             
<div role="search">div search1</div>
<div role="search">div search2</div>

<!-- Not flagged -->
<form role="search">form search1</form>
<form role="search">form search2</form>                                                                                                                                    
```                                                                                                                                                                                

#### How to Reproduce
https://codepen.io/sign_8824/pen/MYjqZmQ

#### Additional context
- `lib/rules/landmark-unique-matches.js` appears to skip `form` and `section` elements that do not have an accessible name.
- The `landmark-unique` rule currently flags duplicate landmarks even when they do not have an accessible name.

straker:
Fixed in #5085

PR #5085, opened 2026-04-29
fix(landmark-unique): exclude section/form with non-landmark roles from landmark match

### LINKED ISSUE #4722: Mis-identification of landmark role, false positive for unique AccName
### Product

axe-core

### Product Version

4.10.2

### Latest Version

#### Expectation
I expect landmark roles to be identified correctly. 
I expect values for aria-labelledby to be correctly differentiated.

#### Actual
Screenshot:
Image

I have a `section` element with the role of `tabpanel`. There are several tabpanels, and they are all given a unique accessible name through the `aria-labelledby` attribute. Visible in the screenshot are those values. 

My first issue is that an element with the role of `tabpanel` is not a landmark element. 
My second issue is that I did not expect to receive an error about unique AccNames, when the values for each `aria-labelledby` attribute are unique ID values. 

#### How to Reproduce
Hopefully the attached screenshot is enough information to reproduce the issue. If not, please let me know.

nami8824:
Closes #4722, #5064

Updated `landmark-unique-matches` to change the matching logic for `section` / `form` elements as follows:
                                                                                                                                                                                                
before
                                     
 |                         | No `aria-label` | Has `aria-label` |
 | ----------------------- | :-------------: | :--------------: |                                                                                                                              
 | No explicit `role`      | ❌              | ✅               |
 | Explicit landmark role  | ❌              | ✅               |                                                                                                                              
 | Non-landmark role       | ❌              | ✅ ← bug        | 
                                                                                                                                                                                                
after
                                                                                                                                                                                                
 |                         | No `aria-label` | Has `aria-label` |
 | ----------------------- | :-------------: | :--------------: |                                                                                                                              
 | No explicit `role`      | ❌              | ✅               |
 | Explicit landmark role  | ✅               | ✅               |                                                                                                                              
 | Non-landmark role       | ❌              | ❌               |

~~Unrelated to this issue, but I think `section` / `form` with an explicit landmark role and no accessible name might also need to match, opened a separate issue for it: #5064~~


lib/rules/landmark-unique-matches.js
```diff
@@ -17,11 +17,11 @@ function isLandmarkVirtual(vNode) {
   }
 
   const { nodeName } = vNode.props;
+  const isLandmarkRole = landmarkRoles.includes(role);
 
   if (nodeName === 'section' || nodeName === 'form') {
-    const accessibleText = accessibleTextVirtual(vNode);
-    return !!accessibleText;
+    return isLandmarkRole && !!accessibleTextVirtual(vNode);
```
WilcoFiers: I think the real bug here is that section / form elements are always treated as landmarks if they have an accessible name, even if they have a role that doesn't map to your name. We should add a test for that:

```html
<form role="tabpanel" aria-label="foo" id="pass-form-tabpanel1"></form>
<form role="tabpanel" aria-label="foo" id="pass-form-tabpanel1"></form>
```

lib/rules/landmark-unique-matches.js
WilcoFiers: Axe-core already has code for checking which element is a landmark. The real bug here is that this matches function wasn't updated to use `getRoleType`. You can remove `isLandmarkVirtual` from this file.

```suggestion
    getRoleType(virtualNode) === 'landmark' &&
    isVisibleToScreenReaders(virtualNode)
```

This fix also addresses the other issue you raised.

WilcoFiers (changes requested):
Thank you for raising this. There's already a solution built into axe-core for doing the thing isLandmarkVirtual is trying to do. We should use that instead.

test/integration/rules/landmark-unique/landmark-unique-fail.html
```diff
@@ -87,8 +98,8 @@
 <div id="violation-role-region" role="region"></div>
 <div id="violation-role-region-2" role="region"></div>
 
-<div id="violation-role-search" role="search"></div>
-<div id="violation-role-search-2" role="search"></div>
+<form id="violation-role-search" role="search"></form>
+<form id="violation-role-search-2" role="search"></form>
```
nami8824, line 103: Ideally, I wanted to add new test cases for `<section>/<form>` with a landmark role but no `aria-label`. However, all landmark roles were already used in the no-label section of this file, so there was no available role to use for new elements. Instead, I replaced the existing `<div role="search">` with `<form role="search">`.

nami8824:
I've made the requested changes. Could you take another look?

https://github.com/dequelabs/axe-core/pull/5085/changes/0afa4bd1741124b12f91ef63ced20dd6520224a9

test/integration/rules/landmark-unique/landmark-unique-pass.html
```diff
@@ -5,6 +5,17 @@
...
+<form
+  id="pass-form-search-aria-label-1"
+  role="search"
+  aria-label="form-search-label-1"
+></form>
+<form
+  id="pass-form-search-aria-label-2"
+  role="search"
+  aria-label="form-search-label-2"
+></form>
```
WilcoFiers, line 17: This is the false positive of the original issue. Good to have that in the E2E suite.

```suggestion
<section
  id="pass-section-tabpanel-aria-label-repeat-1"
  role="tabpanel"
  aria-label="My tab panel"
></section>
<section
  id="pass-section-tabpanel-aria-label-repeat-2"
  role="tabpanel"
  aria-label="My tab panel"
></section>
```

test/integration/rules/landmark-unique/landmark-unique-pass.json
```diff
@@ -6,6 +6,8 @@
     ["#pass-header"],
     ["#pass-form-aria-label-1"],
     ["#pass-form-aria-label-2"],
+    ["#pass-form-search-aria-label-1"],
+    ["#pass-form-search-aria-label-2"],
```
WilcoFiers, line 10:
```suggestion
    ["#pass-section-tabpanel-aria-label-repeat-1"],
    ["#pass-section-tabpanel-aria-label-repeat-2"],
```

WilcoFiers (dismissed):
Fantastic. One more suggestion.

test/integration/rules/landmark-unique/landmark-unique-pass.json
nami8824: Thanks for the suggestion. I tried applying the suggested JSON change by adding these selectors, but the integration test fails with `"Element not found"`.
If I add the suggested elements only to `landmark-unique-pass.html`, without listing them in `landmark-unique-pass.json`, the test passes. Would you prefer that approach?
If the rule-matches test already verifies that these elements are not matched, I think we may not need additional integration coverage for this case.

Also, I originally added the `form` + `role="search"` + `aria-label` cases because I did not see existing integration coverage for form/section + accessible name + explicit landmark role passing. I also added similar coverage in `test/rule-matches/landmark-unique-matches.js`. If that coverage is not useful here, I can remove those cases.

straker (approved):

straker:
Approved for security

PR #5086, opened 2026-04-29
fix(landmark-unique): correct form/section landmark calculation

### LINKED ISSUE #4722: Mis-identification of landmark role, false positive for unique AccName

### LINKED ISSUE #5064: `landmark-unique` does not match `<form role="search">` without an accessible name

WilcoFiers:
The landmark-unique-matches function used an outdated approach to finding landmarks. I noticed that looking at https://github.com/dequelabs/axe-core/pull/5085, which tried to patch the problem. The better solve is to use the existing code for finding larndmarks. 

Closes: https://github.com/dequelabs/axe-core/issues/4722, closes https://github.com/dequelabs/axe-core/issues/5064

### DIFF
--- a/lib/rules/landmark-unique-matches.js
+++ b/lib/rules/landmark-unique-matches.js
@@ -1,27 +1,9 @@
 import { isVisibleToScreenReaders } from '../commons/dom';
-import { getRole } from '../commons/aria';
-import { getAriaRolesByType } from '../commons/standards';
-import { accessibleTextVirtual } from '../commons/text';
+import { getRoleType } from '../commons/aria';
 
 export default function landmarkUniqueMatches(node, virtualNode) {
   return (
-    isLandmarkVirtual(virtualNode) && isVisibleToScreenReaders(virtualNode)
+    getRoleType(virtualNode) === 'landmark' &&
+    isVisibleToScreenReaders(virtualNode)
   );
-}
-
-function isLandmarkVirtual(vNode) {
-  const landmarkRoles = getAriaRolesByType('landmark');
-  const role = getRole(vNode);
-  if (!role) {
-    return false;
-  }
-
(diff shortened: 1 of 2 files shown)


WilcoFiers:
Didn't mean to open this as. PR. I was just testing.

PR #5087, opened 2026-04-30
selectors.ignoredAttributes

WestonThayer:
npx mocha test/node/test-selector-config*.js to run tests

### DIFF
--- a/axe.d.ts
+++ b/axe.d.ts
@@ -136,7 +136,7 @@ declare namespace axe {
     rules?: RuleObject;
     reporter?: ReporterVersion | string;
     resultTypes?: resultGroups[];
-    selectors?: boolean;
+    selectors?: boolean | SelectorConfig;
     ancestry?: boolean;
     xpath?: boolean;
     absolutePaths?: boolean;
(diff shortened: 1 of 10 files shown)


WilcoFiers (commented):
Solid idea! Would you mind opening an issue first so we can discuss the purpose of such a feature and what the best design for it would be if we decided to add it? For one, I'm wondering whether this should be in axe's configuration rather then it being an option that is always passed in. We should also think if we want this to apply to the HTML snippets - at least if the idea here is to avoid outputting sensitive information.

WestonThayer:
Ah sorry Wilco didn't realize this would ping you. This was an experiment to see if there was a middle ground solution to shake randomness out of `target` (React random IDs, CSS-in-JS fun times) without swinging all the way over to `ancestry: true`'s `nth-child` brittleness.

Testing these changes out, I'd say no, not really, forbidding id,class just forces `target` to use `nth-child` in many cases because there's not much else to latch onto (depending on the site).

I also played with XPath's ability to match innerText, but that doesn't help icons, etc.

WilcoFiers:
Stale. I'm going to close this for now. Feel free to open that issue if you want to discuss further.

PR #5094, opened 2026-05-01
ci(deploy): use gh cli to create github release

### LINKED ISSUE #5052: Convert release creation to GitHub CLI tool
## Summary

The script for creating a release is failing to run. It is right now using a go tool that uses the GitHub API to make a release. During the conversion to GitHub Actions, this internal script was maintained for consistency with CircleCI to reduce the number of changes going on at once. Now though, we are debugging something that we don't even need since the GitHub CLI provides this functionality directly.

## Acceptance Criteria

- [ ] GitHub Release creation job is modified
- [ ] It no longer installs the 3rd party CLI tool for creating a release
- [ ] It utilizes a new script, kept in this repository at `.github/scripts/deploy` for creating the release with the `gh` command
- [ ] The new script still creates the description content just like the current script.

## Useful Information

Current script for getting the changelog:

https://github.com/dequelabs/attest-release-scripts/blob/8e4afc470af0a627cf4564006ce1d70c5511851b/src/node-github-release.sh#L10-L32

Garbee:
Replaces the Go-based `github-release` tool and the externally fetched `node-github-release.sh` script in the `create-github-release` job with a single step that uses the `gh` CLI (pre-installed on GitHub-hosted runners).

### What changed

- **Dropped:** `go install gopkg.in/aktau/github-release.v0@latest`, the `curl` of `dequelabs/attest-release-scripts/.../node-github-release.sh`, the `chmod +x` step.
- **Added:** an `Extract release notes` step that runs the same changelog-section `awk` as the original script and writes the result to a multiline `notes` step output.
- **Added:** a `Create GitHub Release` step that calls `gh release create "v${VERSION}" --title "Release ${VERSION}" --notes "$NOTES"`, taking the version from `prod-deploy`'s job output and the notes from the previous step's output (passed via `NOTES` env var so shell metacharacters in the changelog aren't re-evaluated).
- Switched the auth env var from `GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}` to `GH_TOKEN: ${{ github.token }}` to match the rest of the workflow and the `gh` CLI's preferred env var.

### Why

- Removes two external dependencies (a Go binary install and a remote shell script fetched at job runtime) in favor of a tool already on the runner image.
- Keeps the entire release-creation flow visible in `deploy.yml` instead of being hidden in a fetched script from another repo.

<details>
<summary>How the <code>awk</code> changelog extraction works</summary>

```
awk '/# \[?[0-9]/{if(found) exit; found=1; next} found{print}' CHANGELOG.md
```

awk walks `CHANGELOG.md` line by line and runs two `pattern { action }` blocks:

1. `/# \[?[0-9]/` matches a version-header line — `#`, a space, an optional `[`, a digit. In our CHANGELOG that hits lines like `### [4.11.4](...) (2026-04-23)` (the pattern isn't anchored, so it matches the `# [4` substring inside `###`).
    - On the **first** match: set `found=1` and `next` past the line (so the header itself isn't printed).
    - On the **next** match (the previous release's header): `exit`.
2. `found{print}` — for every line where `found` is non-zero, print it.

Net effect: prints everything between the latest version header and the previous one — i.e. the body of the most recent release section, which becomes the GitHub release notes.

Walking through the current `CHANGELOG.md`:

| Line | Matches v-header? | `found` | Action |
| --- | --- | --- | --- |
| `# Changelog` | no | 0 | nothing (`found` is 0) |
| (blank) | no | 0 | nothing |
| `All notable changes...` | no | 0 | nothing |
| (blank) | no | 0 | nothing |
| `### [4.11.4](...) (2026-04-23)` | **yes** | 0 → 1 | set `found=1`, skip line |
| (blank) | no | 1 | print |
| `### Bug Fixes` | no | 1 | print |
| (blank) | no | 1 | print |
| `- **commons/text:** ...` | no | 1 | print |
| `- **utils/getAncestry:** ...` | no | 1 | print |
| (blank) | no | 1 | print |
| `### [4.11.3](...) (2026-04-13)` | **yes** | 1 | `exit` |

</details>

Closes: #5052

### DIFF
--- a/.github/bin/extract-release-notes.sh
+++ b/.github/bin/extract-release-notes.sh
@@ -1,0 +1,37 @@
+#!/usr/bin/env bash
+
+set -eu -o pipefail
+
+# Extracts the body of the most recent release section from CHANGELOG.md —
+# everything between the latest version header and the previous one. The
+# version header itself is not included. The result is written as a
+# multiline `notes` step output to $GITHUB_OUTPUT when set (GitHub
+# Actions), otherwise to stdout.
+#
+# The awk pattern /# \[?[0-9]/ matches version-header lines (`#`, space,
+# optional `[`, digit), e.g. `### [4.11.4](...) (2026-04-23)`. The pattern
+# is not anchored, so it matches the `# [4` substring inside `###`. On the
+# first match it sets `found=1` and skips the line (so the header itself
+# is not printed); on the next match (the previous release's header) it
+# exits. `found{print}` prints every line in between.
+#
+# Walking through a typical CHANGELOG.md:
+#
+#   Line                              v-header?  found    Action
+#   --------------------------------  ---------  ------   -------
+#   # Changelog                       no         0        nothing
+#   All notable changes...            no         0        nothing
+#   ### [4.11.4](...) (2026-04-23)    yes        0 -> 1   skip
+#   (blank)                           no         1        print
+#   ### Bug Fixes                     no         1        print
+#   - **commons/text:** ...           no         1        print
+#   - **utils/getAncestry:** ...      no         1        print
+#   ### [4.11.3](...) (2026-04-13)    yes        1        exit
+
+repo_root="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+
+{
+  echo 'notes<<EOF'
+  awk '/# \[?[0-9]/{if(found) exit; found=1; next} found{print}' "$repo_root/CHANGELOG.md"
+  echo 'EOF'
+} >> "${GITHUB_OUTPUT:-/dev/stdout}"
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -168,12 +168,7 @@ jobs:
       - *checkout
       - name: Extract release notes
         id: release-notes
-        run: |
-          {
-            echo 'notes<<EOF'
-            awk '/# \[?[0-9]/{if(found) exit; found=1; next} found{print}' CHANGELOG.md
-            echo 'EOF'
-          } >> "$GITHUB_OUTPUT"
+        run: ./.github/bin/extract-release-notes.sh
       - name: Create GitHub Release
         env:
           GH_TOKEN: ${{ github.token }}
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -177,6 +177,7 @@ jobs:
         run: |
           gh release create "v${VERSION}" \
             --title "Release ${VERSION}" \
+            --target ${{ github.sha }} \
             --notes "$NOTES"
   validate-deploy:
     name: Validate Deployment
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -166,16 +166,23 @@ jobs:
       contents: write # Required to create releases
     steps:
       - *checkout
-      - name: Install Release Helper
-        run: go install gopkg.in/aktau/github-release.v0@latest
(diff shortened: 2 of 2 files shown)


straker (commented):
In our [release script](https://github.com/dequelabs/axe-core/blob/develop/.github/workflows/release.yml#L33-L40), we use the following to generate the release notes for the pr body:

```
# Get the additions to the changelog as the commit body and generate the PR url
          ReleaseNotes=$(
            git show \
              --no-color \
              --no-prefix \
              --output-indicator-new=! CHANGELOG.md | egrep '^!' | awk -F'^[!]' '{print $2}' | sed -e 's/\n/$0A/g'
          )
```

We should generate the changelog consistently

Garbee:
> We should generate the changelog consistently

Can you prove the new method is inconsistent with the previous behavior? What situation have you tested that shows the differential behavior?

A few reasons the current proposed change is better than the old method:

1. **Decoupled from commit shape.** The git-show approach only works when HEAD is exactly the release commit. A follow-up typo fix to `CHANGELOG.md`, an unrelated commit that touches it, or a merge commit resolving changelog conflicts all silently produce wrong or partial notes. This coupling also makes the script harder to test and debug — validating behavior requires mutating git state (staging a fake release commit, resetting, etc.) just to exercise the extraction logic. The awk version is a pure function of the file on disk, so we can run it at any point in the working tree and get the same answer CI would.
1. **Preserves formatting.** `git show | grep '^!'` only captures added lines, dropping the blank lines that separate sub-sections. The awk version returns the section as-authored, which renders correctly in GitHub releases.
1. **No newline-encoding bug.** The old snippet's `sed 's/\n/$0A/g'` is broken — `sed` doesn't see `\n` across line boundaries in a basic `s///`, and the intended escape was `%0A` anyway. The awk output just flows into a heredoc `$GITHUB_OUTPUT` write with no encoding gymnastics.
1. **Simpler to reason about.** One command, no pipeline of four tools, no flag (`--output-indicator-new=!`) chosen specifically to dodge false positives from `+` characters in the diff.

straker:
I didn't say one was better than the other, I just said we should generate it the same in both situations. I'd be fine if we changed the release script code to use the `awk` code.

Garbee:
> I didn't say one was better than the other, I just said we should generate it the same in both situations. I'd be fine if we changed the release script code to use the `awk` code.

The entire release script code in another repo is going away. We are re-doing the workflows for better devops all around. So changing that is out of scope on fixing the problem in axe-core.

straker:
The release script I pointed to lives in axe-core's action `release.yml` https://github.com/dequelabs/axe-core/blob/develop/.github/workflows/release.yml#L34C1-L40

Garbee:
> The release script I pointed to lives in axe-core's action `release.yml` https://github.com/dequelabs/axe-core/blob/develop/.github/workflows/release.yml#L34C1-L40

Ah, sorry with the original script to this being called a release script I was confused.

So, I don't disagree that the other area _should_ be changed, but right now they are functionally doing basically the same thing. And in the context of that release workflow it is at least fine there as-is since it is in a script guaranteed to have the last commit be the one with the content. Since `npm run release` runs before it in the script.

I think doing anything with the release script, which things _should_ be done is an auxiliary task and unrelated to this change to fix the specific bug. Feel free to backlog a work item and get it prioritized. But we don't need to scope creep this fix. That script probably needs more work than just a tweak to creating the PR description.

Also, that script is only using that content in the PR description. So it's a very low priority thing to even be consistent with. Since ultimately we're talking about a few empty lines of content difference potentially.

straker (approved):

PR #5096, opened 2026-05-05
fix(label-content-name-mismatch): match visible text with aria-label and exclude invisible text

### LINKED ISSUE #5063: img[alt] should not be treated as text in label-content-name-mismatch
#### Expectation

Given the following HTML, axe should **not** report a `label-content-name-mismatch` violation:

```html
<a aria-label="Deque" href="/">
  <img alt="Logo" width="100" src="logo.svg" />
  <span>Deque</span>
</a>
```

The visible text is "Deque" and the `aria-label` is "Deque" — these match. The image alt text "Logo" is not visible on screen.

#### Actual

axe reports a `label-content-name-mismatch` violation: *"Text inside the element is not included in the accessible name."*

#### Root cause

The evaluate function in `lib/checks/label/label-content-name-mismatch-evaluate.js` uses `subtreeText()` to determine "visible text." However, `subtreeText` computes the **accessible text from content** (per the AccName algorithm), not the **visually rendered text**. For each child, it calls `accessibleTextVirtual()`, which includes:

- Image alt text (via `nativeTextAlternative`)
- SVG titles (see #4065)
- Visually hidden text (see #4678)

In this case, `subtreeText` returns `"Logo Deque"` (including the img alt), which does not match the `aria-label` `"Deque"`.

By contrast, the matches function correctly uses `visibleVirtual()`, which only collects actual text nodes rendered on screen.

#### Additional context

Related issues sharing the same root cause (`subtreeText` including non-visual text in the "visible text" computation):

- #4065 (SVG titles)
- #4678 (visually hidden text)
- #4561 (switch value text)

### Potential solution

Replace subtreeText() with visibleVirtual() in the evaluate function to compare only on-screen text against the accessible name. This fixes false positives where image alt text, SVG titles, or visually-hidden text inflate the "visible text" string. Icon ligature filtering can be handled as a separate step using isIconLigature, following the existing pattern in color-contrast-matches.js. Both visibleVirtual and isIconLigature are already available in commons/text. This should also resolve the related issues #4065 and #4678.

mehm8128:
Using `visibleVirtual` instead of `subtreeText`, I resolved issue.
Because if I simple replace it and use `isIconLigature` existing two tests failed (`'returns true when visible text excluding ligature icon is part of accessible name'` and `'returns true when text contains <br/>'`), I changed `visible-virtual.js`, but I'm not sure this is proper changes (this is the first PR for me to axe-core).

Closes: https://github.com/dequelabs/axe-core/issues/5063

### DIFF
--- a/lib/commons/text/visible-virtual.js
+++ b/lib/commons/text/visible-virtual.js
@@ -1,7 +1,8 @@
-import sanitize from './sanitize';
+import { nodeLookup } from '../../core/utils';
 import isVisibleOnScreen from '../dom/is-visible-on-screen';
 import isVisibleToScreenReaders from '../dom/is-visible-to-screenreader';
-import { nodeLookup } from '../../core/utils';
+import isIconLigature from './is-icon-ligature';
+import sanitize from './sanitize';
 
 /**
  * Returns the visible text of the virtual node
@@ -15,9 +16,13 @@ import { nodeLookup } from '../../core/utils';
  * @param  {Boolean} screenReader When provided, will evaluate visibility from the perspective of a screen reader
  * @param  {Boolean} noRecursing When False, the result will contain text from the element and it's children.
  * When True, the result will only contain text from the element
+ * @param  {Object} [options]
+ * @param  {Boolean} [options.ignoreIconLigature] When true, icon ligature text nodes are excluded
+ * @param  {number} [options.pixelThreshold] Pixel threshold for icon ligature detection
+ * @param  {number} [options.occurrenceThreshold] Occurrence threshold for icon ligature detection
  * @return {String}
  */
-function visibleVirtual(element, screenReader, noRecursing) {
+function visibleVirtual(element, screenReader, noRecursing, options = {}) {
   const { vNode } = nodeLookup(element);
   const visibleMethod = screenReader
     ? isVisibleToScreenReaders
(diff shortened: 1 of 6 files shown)


lib/commons/text/visible-virtual.js
```diff
@@ -28,16 +33,26 @@ function visibleVirtual(element, screenReader, noRecursing) {
...
       if (nodeType === 3) {
         // filter on text nodes
-        if (nodeValue && visible) {
-          return nodeValue;
+        if (!nodeValue || !visible) {
+          return '';
+        }
+        if (
+          ignoreIconLigature &&
+          isIconLigature(child, pixelThreshold, occurrenceThreshold)
```
mehm8128, line 48: we should judge `isIconLigature` here because it needs `child` node as its first param

lib/commons/text/visible-virtual.js
```diff
@@ -28,16 +33,29 @@ function visibleVirtual(element, screenReader, noRecursing) {
...
+        ) {
+          return '';
         }
-      } else if (!noRecursing) {
-        return visibleVirtual(child, screenReader);
+        return nodeValue;
+      }
+      if (nodeName === 'br') {
+        return ' ';
+      }
```
mehm8128, line 56: we should replace `<br>` with `' '` to pass `'returns true when text contains <br/>'` test

test/checks/label/label-content-name-mismatch.js
```diff
@@ -186,4 +186,12 @@ describe('label-content-name-mismatch tests', function () {
     var actual = check.evaluate(vNode.actualNode, options, vNode);
     assert.isTrue(actual);
   });
+
```
WilcoFiers, line 189: Would you mind adding a ligature font unit test here as well?

WilcoFiers (commented):
Looks good. I think this is missing one test.

WilcoFiers (changes requested):
Should have made that a change request.

test/checks/label/label-content-name-mismatch.js
mehm8128: added more test cases!

WilcoFiers (dismissed):

WilcoFiers:
@copilot resolve the merge conflicts in this pull request

mehm8128:
Copilot didn't seem to be working, so I resolved the conflict manually

WilcoFiers:
@mehm8128 would you mind running Prettier? the fmt_check is failing because of it.

mehm8128:
@WilcoFiers done

WilcoFiers (approved):
Looks good to me. Thank you @mehm8128!

For the books, I reviewed this PR for security.

PR #5099, opened 2026-05-07
feat: add gather-internals.js external script

### LINKED ISSUE #5041: ElementInternals: create the internals gather script
Axe-core will have a new exported file (separate from the main `axe.js` and `axe.min.js` files which will gather the internals of the object for the extension. The file should be thoroughly tested.

```js
// gather-internals.js
(() => {

  function generateAncestry(node) {
    // copy of the axe-core utils function since axe does not exist in the main context
  }

  function getElementInternals(node) {
    // copy of the function taken directly from axe-core element-internals proposal
  }

  const elementInternalsMap = {};
  // The final implementation will be conceptually similar to this, but will need to do a
  // shadow-DOM-piercing walk instead of solely relying on document.createTreeWalker
  const treeWalker = document.createTreeWalker(
    document.body,
    window.NodeFilter.SHOW_ELEMENT,
    null,
    false
  );
  
  let node = treeWalker.currentNode;
  while (node) {
    
    const elementInternals = getElementInternals(node);
    if (elementInternals) {
      
      const ancestry = generateAncestry(node);
      const internals = {};  

      // can't spread internals so have to loop over the props instead
      for (const prop in elementInternals) {
        
        // trying to read internals.form will throw if the custom element isn't a form associated element :(
        try {
          if (elementInternals[prop] !== null) {
            internals[prop] = elementInternals[prop];

            // convert idref(s) to node ancestry
            if (prop.endsWith('Element')) {
              internals[prop] = generateAncestry(elementInternals[prop])
            }
            else if (prop.endsWith('Elements')) {
              internals[prop] = elementInternals[prop].map(node => generateAncestry(node));
            }
          }
        } catch {}
      }
      elementInternalsMap[ancestry] = internals;
    }

    node = treeWalker.nextNode();
  }

  return elementInternalsMap;
})();
```

straker:
Testing this was interesting as it doesn't get added to the `axe` object. I needed to test the logic of the code but the way we have to build this for the extension script means we only get the returned object when the file is injected, so I split out the logic into it's own file that is built separately and tested. This uses the `testutils` changes from #5080, so will fail until that is merged into develop

Closes: https://github.com/dequelabs/axe-core/issues/5041


pwsmaster (approved):

lib/gather-internals/walk-tree.js
```diff
@@ -0,0 +1,60 @@
...
+    if (elementInternals) {
+      const ancestry = getAncestry(node);
+      const internals = {};
+
+      // can't spread internals so have to loop over the props instead
+      for (const prop in elementInternals) {
+        // trying to read internals.form will throw if the custom element isn't a form associated element :(
+        try {
+          if (elementInternals[prop] !== null) {
+            internals[prop] = elementInternals[prop];
```
WilcoFiers: We may end up dragging in a lot more then we actually need through this. I think we should limit this to aria props, and the few props that don't start with `aria` that we know we need. On top of that we should check the types of what we're handling. The result of this function must be serializable without errors. We should make sure that's the case.

lib/gather-internals/walk-tree.js
```diff
@@ -0,0 +1,60 @@
+/* eslint-disable max-depth, guard-for-in */
```
WilcoFiers: I don't think we should disable max-depth here. This doesn't need to be a single deep function.

lib/gather-internals/walk-tree.js
```diff
@@ -0,0 +1,60 @@
...
+        // trying to read internals.form will throw if the custom element isn't a form associated element :(
+        try {
+          if (elementInternals[prop] !== null) {
+            internals[prop] = elementInternals[prop];
+
+            // convert idref(s) to node ancestry
+            if (prop.endsWith('Element')) {
+              internals[prop] = getAncestry(elementInternals[prop]);
+            } else if (prop.endsWith('Elements')) {
+              internals[prop] = elementInternals[prop].map(n => getAncestry(n));
```
WilcoFiers: Doing this by type rather then by how the string ends seems safer. There's no guarantee every future prop will continue to follow this convention.

lib/gather-internals/main.js
```diff
@@ -0,0 +1,4 @@
+import { walkTree, elementInternalsMap } from './walk-tree';
+
+walkTree();
+module.exports = elementInternalsMap;
```
WilcoFiers, line 4: This isn't an index file. I think we should rename this to `main.js`

lib/gather-internals/walk-tree.js
```diff
@@ -0,0 +1,60 @@
+/* eslint-disable max-depth, guard-for-in */
+
+// importing from the index file results in esbuild processing all utils files and not tree shaking them correctly
```
WilcoFiers, line 1: I don't much like this. We have no safeguards in place that prevent one of these files from importing an index, and through it pulling all of axe-core in. Or worse, maybe trying to use the scoped `axe` variable in some code path and breaking because in `gather-internals.js` that variable doesn't exist.

Can you come up with some way to guard against that?

WilcoFiers (dismissed):
Couple thoughts

chutchins25 (commented):
A couple of questions on the gather path itself — no blockers from my side, just things worth thinking about alongside Wilco's feedback.

lib/gather-internals/walk-tree.js
```diff
@@ -0,0 +1,60 @@
...
+              internals[prop] = getAncestry(elementInternals[prop]);
+            } else if (prop.endsWith('Elements')) {
+              internals[prop] = elementInternals[prop].map(n => getAncestry(n));
+            }
+          }
+        } catch {
+          // do nothing
+        }
+      }
+      elementInternalsMap[ancestry] = internals;
```
chutchins25: The map key here is the raw return of `getAncestry(node)`, which is `Array<String>` for shadow-DOM nodes (per `getShadowSelector`). JS coerces that array to a comma-joined string when used as an object key — but the same `getAncestry()` output stored as a property value on lines 34/36 keeps its array form. Was the asymmetry (string keys / array-shaped idref values) intentional? Consumers can't reuse an idref value as a lookup key without re-stringifying.

lib/gather-internals/walk-tree.js
```diff
@@ -0,0 +1,60 @@
+/* eslint-disable max-depth, guard-for-in */
+
+// importing from the index file results in esbuild processing all utils files and not tree shaking them correctly
+import getElementInternals from '../core/utils/get-element-internals';
+import getAncestry from '../core/utils/get-ancestry';
+import isShadowRoot from '../core/utils/is-shadow-root';
+
+export const elementInternalsMap = {};
+
+export function walkTree(tree = document.body) {
```
chutchins25, line 10: Is it possible for the injection at document_start to be with no body yet? `document.createTreeWalker(null, ...)` will throw. Is this the behavior we desire?

straker: This should only run after the page is settled, so I don't see this as a problem

lib/gather-internals/walk-tree.js
straker: Great catch!

chutchins25: Fix has been applied.

chutchins25 (dismissed):
Looks good to me after changes applied.

lib/gather-internals/walk-tree.js
```diff
@@ -0,0 +1,80 @@
...
+          else if (
+            Array.isArray(value) ||
+            value instanceof globalThis.NodeList
+          ) {
+            const array = Array.from(value).filter(n => n.isConnected);
+            internals[prop] = array.length
+              ? array.map(n => getAncestry(n))
+              : undefined;
+          } else {
+            internals[prop] = value;
```
WilcoFiers, line 55: At this point, these should either be a string, number, or boolean correct? Should we test that it is so that we don't end up exporting something that can't be stringified?

straker: According to the docs I think it's only a string. Testing shows that if you try to set the property to a non-string value it stringifies it. Since we're filtering by known prop names I think we're fine not testing the value.

lib/gather-internals/walk-tree.js
```diff
@@ -0,0 +1,80 @@
...
+          else if (
+            Array.isArray(value) ||
+            value instanceof globalThis.NodeList
+          ) {
+            const array = Array.from(value).filter(n => n.isConnected);
+            internals[prop] = array.length
+              ? { type: 'NodeList', value: array.map(n => getAncestry(n)) }
+              : undefined;
+          } else {
+            internals[prop] = value;
```
WilcoFiers: OK in that case I think we should do this. We need to make sure that what we're passing in can be serialized and transferred.

```suggestion
          } else if (typeof value === 'string') {
            // Arbitrary string limit to avoid cross-world message size limits.
            // Axe-core is unlikely to need the full string
            if (value.length >= 1000) {
              value = value.substring(0, 999) + '…'
            }
            internals[prop] = value;
```

WilcoFiers (dismissed):

lib/gather-internals/walk-tree.js
straker: I'm  not sure about that. Rules like `label-content-name-mismatch` definitely will need the full string for the `ariaLabel` prop otherwise we'll only work with a subset of the value when testing. That seems like a false-positive can easily occur.

WilcoFiers: That's a good point. Alright lets leave it at just checking for strings.

WilcoFiers (approved):

PR #5100, opened 2026-05-12
fix(aria-conditional-attr): add support for radio

### LINKED ISSUE #4416: aria-conditional-attr correct for checkbox, but silent on radios
The aria-conditional-attr  correctly identifies when the aria-checked attribute is used on a native checkbox - and while that's the primary example that had been discussed per the creation of this rule, the same failure should be called out if someone does `<input type=radio aria-checked=...>` since it results in the same situation of the ARIA attribute not necessarily matching the state of the html input.

The bit we added to ARIA in HTML about this was worded to be for both input types:
>Authors MUST NOT use the [aria-checked](https://www.w3.org/TR/wai-aria-1.2/#aria-checked) attribute on any element where the [checkedness](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#concept-fe-checked), or the indeterminate checked value of the element can be in opposition to the current value of the aria-checked attribute.

mehm8128:
added support for radio for `aria-conditional-attr` as well as checkbox.

Closes: https://github.com/dequelabs/axe-core/issues/4416

### DIFF
--- a/lib/checks/aria/aria-conditional-attr-evaluate.js
+++ b/lib/checks/aria/aria-conditional-attr-evaluate.js
@@ -1,10 +1,12 @@
 import getRole from '../../commons/aria/get-role';
 import ariaConditionalCheckboxAttr from './aria-conditional-checkbox-attr-evaluate';
+import ariaConditionalRadioAttr from './aria-conditional-radio-attr-evaluate';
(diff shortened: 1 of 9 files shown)


straker (approved):
Thanks for the pr! Looks great.

Reviewed for security

PR #5104, opened 2026-05-14
chore: replace grunt with custom build scripts

WilcoFiers:
- Replaces all uses of grunt with build scripts.
- Migrate build dir to .mjs
- Remove grunt from qunit test example


build/shared/format.mjs
Garbee, line 1: Can we not just call the prettier CLI?

WilcoFiers: I guess, so inline an execSync? Are there any advantages to doing it that way?

Garbee: I was thinking this was being called from a different context yesterday. Looking deeper into it this morning, there is no advantage given the context, only drawbacks. Primarily if we spawned a process it is slower since we're waiting for other OS processes to spin up in the `run-build` loop.

My thinking was if we're replacing grunt then we should have a system that is built to try and leverage the more native capabilities of the platform, without so much manual wiring.

I still think the overall maintenance with this is, massive. There is a lot of mental context to manually wire up. But, we probably aren't escaping that in any way.

Fewer older dependencies the better IMO, so I'm good with the architectural direction this is going in. Even though I'd rather figure out a more modern orchestration system, I don't want to delay moving off grunt based on that alone. We can iterate on it as it becomes necessary.

build/run-build.mjs
```diff
@@ -0,0 +1,772 @@
...
+  const langs = deriveLangSuffixes(parsed);
+  fs.mkdirSync(path.join(ctx.root, 'tmp/core'), { recursive: true });
+  await esbuildBuild({
+    entryPoints: [path.join(ctx.root, 'lib/core/core.js')],
+    outfile: path.join(ctx.root, 'tmp/core/core.js'),
+    minify: false,
+    bundle: true
+  });
+  await runConfigureAll(ctx, langs, parsed.tags);
+}
```
chutchins25: Tracing the configure path, `runConfigureAll` → `buildRules` → `buildManual` only reads `lib/{rules,checks,misc}/*.json` — nothing downstream consumes `tmp/core/core.js`, and the next `npm run build` wipes `tmp/core/` via `cleanOutputs` before re-bundling. The old `grunt configure` task didn't bundle the core either, so it looks like leftover from copy-pasting the full-build flow. If there's no consumer I'm missing, dropping the `mkdirSync` + `esbuildBuild` block would shave ~1s off every `husky` pre-commit fire.

package.json
```diff
@@ -146,13 +148,7 @@
...
     "glob": "^10.3.10",
     "globals": "^17.1.0",
-    "grunt": "^1.5.3",
-    "grunt-babel": "^8.0.0",
-    "grunt-bytesize": "^0.2.0",
-    "grunt-contrib-clean": "^2.0.1",
-    "grunt-contrib-concat": "^2.1.0",
-    "grunt-contrib-uglify": "^5.2.2",
-    "grunt-contrib-watch": "^1.1.0",
+    "chokidar": "^4.0.3",
```
chutchins25: Consistency suggestion — the `devDependencies` block is alphabetical except for this entry. `chokidar` should sit between `chalk` (line 134) and `chromedriver` (line 135). Recommend deleting this line and re-inserting `    "chokidar": "^4.0.3",` after `"chalk": "^4.x",`.

build/aria-supported-build.mjs
```diff
@@ -0,0 +1,154 @@
...
+ */
+export async function runAriaSupported(
+  root,
+  langSuffixes,
+  listType = 'unsupported'
+) {
+  const require = createRequire(import.meta.url);
+  const suffix = langSuffixes && langSuffixes.length ? langSuffixes[0] : '';
+  const axePath = path.join(root, `axe${suffix}.js`);
+  const axe = require(axePath);
```
chutchins25, line 22: Pre-existing question, not a blocker — is this a good opportunity to clean up if it's actually an issue? `require(axePath)` is cached, so on the 2nd+ rebuild in `npm run develop`, this regenerates `doc/aria-supported.md` from a stale `axe` snapshot until the cache happens to re-align. The old `grunt aria-supported` task had the same behavior under `spawn: false`. A one-liner `delete require.cache[require.resolve(axePath)]` before the `require` would fix it; happy to drop this if you've already considered it.

chutchins25 (commented):
A few low-severity suggestions below, none blocking. Also flagging that 3 of the 5 Copilot inline comments are now stale (already fixed in the latest commits): the eslint `.mjs` pattern, the metadata-function-map banner, and the README "or equivalently" wording. Copilot's QUnit page-error claim is a false positive — `hadPageRuntimeError` is tracked and propagated to `process.exitCode` at lines 60-62.

build/run-build.mjs
```diff
@@ -0,0 +1,772 @@
...
+      clearTimeout(timer);
+    }
+    timer = setTimeout(() => {
+      timer = null;
+      run(p);
+    }, 150);
+  };
+
+  chokidar
+    .watch(['lib', 'build'], {
```
straker: esbuild provides a `watch` api that we could use to do this instead of using `chokidar`

build/rule-generator/get-files-metadata.mjs
```diff
@@ -1,5 +1,5 @@
-const directories = require('./directories');
-const outdent = require('outdent');
+import outdent from 'outdent';
```
straker, line 1: Question: do we really need the rule generator any more? If not, we could delete it for simplicity

straker (commented):
I'm not quite sold on creating a manual task / build pipeline over using something else that is a task runner already (make, gulp, etc.).  I wonder if using a makefile would be simpler than trying to come up with our own large js file

doc/examples/qunit/run-qunit.mjs
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env node
```
straker: Since this is an example, do we really need to do all this to run the test file in a browser? The qunit docs seem to make this a lot simpler and doesn't need all this extra setup. I'd like to make this as simple as possible for our docs rather than try to replace `grunt-contrib-qunit` whole sale.

build/run-build.mjs
```diff
@@ -0,0 +1,836 @@
+#!/usr/bin/env node
```
straker, line 1: Instead of a very complicated and large js file, would this be simpler as a make file?

If not, I think splitting this file up would help (e.g. `runEsbuildCore` is it's own file). Then this file could just be a run file that imports all the other functions and directs what gets run and when.

build/run-build.mjs
```diff
@@ -0,0 +1,836 @@
...
+
+async function runValidateCommand() {
+  const ctx = createBuildContext(root);
+  if (!validateAll(ctx)) {
+    process.exitCode = 1;
+  }
+}
+
+/**
+ * File watch + optional http-server + rebuild / Karma (see `npm run develop`).
```
straker: Develop before didn't try to start a server, just rebuilt on changes. Did you intend for it to take over `npm start` as well?

straker (dismissed):
There's a babel message that should be addressed or removed from the build:

> "[BABEL] Note: The code generator has deoptimised the styling of /tmp/core/core.js as it exceeds the max of 500KB. created file at /locales/_template.json"

If you are in `develop` and save multiple files at once (e.g. running `npm run fmt` or switching branches) the watch mode detects each file changed individually and builds once, then runs each file's test file in turn. This locks the develop step and it won't be able to rebuild on changes until all test files have finished. __Even worse is that you can't exit the program (ctrl+c doesn't work) so you can't stop the process at all. It also leaves zombie processes hanging around.__ I'm not sure how karma handled it, but we should probably have some sort of debounce on the script that waits for all files to settle before running the build, and then skip tests if there are multiple changed files. 

Additionally, a small nit, but I think logs would be helpful during that build. Otherwise you run the build and see nothing for about 10 seconds. Some additional newlines in the output would also be helpful so in `develop` mode you can more easily see where one build ends and the next begins.

build/run-build.mjs
WilcoFiers: Yeah, I did that on purpose.

WilcoFiers:
Thanks for the thorough review. Addressed the blocking feedback from the June 2 review:

**Watch mode**
- Debounce/coalesce multi-file changes (400ms); skip per-file Karma runs when multiple `lib/` or rule-spec files change in one batch
- Reentrancy guard so builds do not stack while Karma is running
- Replaced `execSync` with `spawn` and kill active Karma/http-server children on SIGINT/SIGTERM
- Debounced the `test/` watcher as well

**Build output**
- Step-by-step build logging is available via `--log` (off by default), e.g. `npm run build -- --log` or `npm run develop -- --log`

**Other fixes**
- Babel deoptimisation warning suppressed via `compact: true`
- Removed unused esbuild step from `configure` (pre-commit speedup)
- Cleared `require.cache` before loading `axe*.js` in aria-supported build
- Split `run-build.mjs` into `build/run-build/` modules

**Docs**
- Updated `CONTRIBUTING.md` and `doc/developer-guide.md` for current `develop` behavior (embedded http-server on port 9876, batching, `--log`)

Re: **develop taking over `npm start`** — intentional (replied inline).

**Not changed in this pass** (happy to follow up separately):
- QUnit example simplification (`run-qunit.mjs`) — still a headless Puppeteer runner for CI; open to simplifying if we want a lighter doc example
- esbuild `watch` API vs chokidar — possible future optimization
- Rule generator removal — still in use via `npm run rule-gen`; can discuss separately

Please take another look when you have a moment.

build/run-build/full-build.mjs
```diff
@@ -0,0 +1,71 @@
...
+  step('uglify');
+  runUglify(ctx, langs);
+  step('aria supported docs');
+  await runAriaSupported(ctx.root, langs, 'unsupported');
+  step('locale template');
+  runAddLocaleTemplate(ctx);
+  step('postbuild');
+  runPostbuild(ctx);
+  step('bytesize');
+  runBytesize(ctx, langs);
```
straker, line 65: This is a lot easier to parse, thanks!

straker:
Changes are looking good. I tried the swapping of branches and it almost works out great, but still has a few hiccups (though not nearly as problematic as before). At this point it's just refinement of the system so could be fixed here or in a separate pr, up to you.

* when running in `develop` and a file changes, you cannot quit the program during the build (`ctrl + c` doesn't work). the build has to finish to completion before the program allows sigterm
* when running in `develop` and a large group of files changes, the build no longer runs tests in series, but it tries to build multiple times in a row even while trying to run tests so the subsequent tests don't have `axe` and break

build/run-build/watch-helpers.mjs
```diff
@@ -0,0 +1,293 @@
...
+  const paths = [].concat(testAbsPaths);
+  const testFilesValue = paths
+    .map(p => unitTestFilesArg(projectRoot, p))
+    .join(',');
+
+  return new Promise((resolve, reject) => {
+    killActiveTest();
+    const child = spawn(
+      'npm',
+      ['run', 'test:unit', '--', `testFiles=${testFilesValue}`],
```
chutchins25: This passes the Karma-era `testFiles=` convention, but `develop` migrated to Web Test Runner (#5121) — every `test:unit:*` script in this same `package.json` targets files with `--files '<glob>'`, and `wtr.config.mjs` has no `testFiles=` handling. WTR treats the bare `testFiles=…` arg as a file glob, matches nothing, and exits non-zero, so watch mode's targeted test runs don't actually execute. The `testFiles`/"karma" comments at lines 154 and 215 look like leftovers from before the WTR migration. Could you switch this to WTR's `--files` (one flag per path) and refresh those comments? Worth a quick `npm run develop` + save-a-file check to confirm the targeted run fires.

chutchins25 (commented):
Nice work — splitting `run-build` into `build/run-build/` modules and the QUnit simplification both read much cleaner, and my four earlier comments are all addressed. Two questions on the current state, both on the dev tooling rather than the shipped bundle (CI is unaffected):

1. Watch mode still uses the Karma `testFiles=` convention against the new Web Test Runner — targeted auto-test runs look broken.
2. The new `test:build` suites don't appear to run in `npm test` or CI.

package.json
```diff
@@ -110,16 +111,18 @@
...
     "codemod:modernize-tests": "node build/codemods/modernize-test-files.mjs",
     "fmt": "prettier --write .",
     "fmt:check": "prettier --check .",
     "prepare": "husky && npm run patch",
-    "prebuild": "node ./build/check-node-version.js",
-    "pretest": "node ./build/check-node-version.js",
+    "prebuild": "node ./build/check-node-version.mjs",
+    "pretest": "node ./build/check-node-version.mjs",
     "pretest:unit:integration": "npm run build:integration-tests",
+    "test:build": "node --test build/run-build/*.test.mjs",
```
chutchins25, line 124: These new `node --test` suites are a nice addition, but `test:build` doesn't seem to be invoked anywhere — `npm test` is `test:tsc && run-s "test:unit:* -- {@}"` (the glob only matches `test:unit:*`, line 84) and there are no references under `.github/`. As written, the watch/build-helper tests only run on a manual `npm run test:build`, so CI won't catch regressions in them. Was standalone-only intentional, or should `test:build` join the `test` aggregate / a CI step?

WilcoFiers: That's intentional, see: https://github.com/dequelabs/axe-core/pull/5104/changes#diff-70a5b47038db68ea7be6ddb573f300bee9bf63a35f1ddc874ab9f882d9c02b56R48

I'm open to discuss, but testing watch mode doesn't seem like something we need in CI.

build/run-build/watch-helpers.mjs
WilcoFiers: Darnit :/ I had a feeling I was going to miss something like that. I couldn't find it, and neither did Cursor. Thanks!

chutchins25: Confirmed — `4d2bd3f` switches to repeatable `--files` and cleans up the `Karma`/`testFiles` comments. The targeted run resolves correctly now. Thanks!

package.json
chutchins25: Just making sure we didn't miss something was all. We can leave it out of CI. Thanks for the link too.

chutchins25 (dismissed):
Looks good to me!

straker (dismissed):
Almost there. The recent changes to use `--files` for running web test runner also need to account for when the file changes is in the `integration/rules` dir and run the build script, then run tests on the resulting build. Otherwise the develop freezes as it tries to run `web-test-runner --config test/wtr.config.mjs --files test/integration/rules/accesskeys/accesskeys.html`

See https://github.com/dequelabs/axe-core/pull/5121/changes#diff-270c437420821159935d15b9235c3c1b3e23c613914bb3f2070498d3013faaca

straker (approved):

PR #5105, opened 2026-05-15
feat(axe.externalAPIs): add public api for setting elementInternal data

### LINKED ISSUE #5040: ElementInternals: Create extension interface to pass internal data from extension
Create the `axe.externalAPIs` function which accepts how to handle passing information from the extension to the axe-core context. Add necessary tests.

```js
axe.external = {};
axe.externalAPIs = ({ elementInternalsTimeout = 1000, elementInternals }) => {
  axe.external.elementInternals = (res, rej) => {
    let timeoutCalled = false;
    const timeout = setTimeout(() => {
      timeoutCalled = true;
      res();
    }, elementInternalsTimeout);

    return elementInternals()
      .then(results => {
        clearTimeout(timeout);

        // timeout already called so don't process the response
        if (timeoutCalled) {
          return;
        }

        // resolve ancestry strings to nodes
        for (const ancestry in results) {
          const node = document.querySelector(ancestry);
          const vNode = axe.utils.getNodeFromTree(node);
          const internals = results[ancestry];

          // convert idref(s) ancestries back to nodes
          for (const prop in internals) {
            if (prop.endsWith('Element')) {
              internals[prop] = document.querySelector(internals[prop])
            }
            else if (prop.endsWith('Elements')) {
              internals[prop] = internals[prop].map(node => document.querySelector(node));
            }
          }

          // set internals directly onto the vNode
          vNode.internals = internals;
        }

        res();
      })
      .catch(rej);
  };
}

class Audit {
  run(context, options, resolve, reject) {
    this.normalizeOptions(options);
    DqElement.setRunOptions(options);

    const internalsQueue = queue();
    if (axe.external.elementInternals) {
      internalsQueue.defer(axe.external.elementInternals);
    }

    // if there is no external api function the queue resolves immediately
    internalsQueue.then(() => {
      // rest of audit.run code
    })
  });
}

class VirtualNode {
  // move to getter so it's not automatically run when the node is created
  // and then cache the results similar to vNode.props
  get internals() {
    if (!this._cache.hasOwnProperty('internals')) {
      this._cache.internals = getElementInternals(this.actualNode);
    }

    return this._cache.internals;
  }

  set internals(value) {
    this._cache.internals = value;

1.   }

}
```

straker:
Closes: https://github.com/dequelabs/axe-core/issues/5040

### DIFF
--- a/axe.d.ts
+++ b/axe.d.ts
@@ -681,7 +681,7 @@ declare namespace axe {
    * @returns {Promise<PartialResult>}  Partial result, for use in axe.finishRun.
    */
   function externalAPIs(params?: {
-    getElementInternalsTimeout?: number;
+    elementInternalsTimeout?: number;
     getElementInternals?: () => Promise<ElementInternalsMap>;
   }): void;
 
--- a/axe.d.ts
+++ b/axe.d.ts
@@ -674,6 +674,31 @@ declare namespace axe {
     isDefault?: boolean
   ): void;
 
+  /**
+   * Run axe in the current window only
+   * @param   {ElementContext} context  Optional The `Context` specification object @see Context
+   * @param   {RunOptions}     options  Optional Options passed into rules or checks, temporarily modifying them.
+   * @returns {Promise<PartialResult>}  Partial result, for use in axe.finishRun.
+   */
+  function externalAPIs(params?: {
+    getElementInternalsTimeout?: number;
+    getElementInternals?: () => Promise<ElementInternalsMap>;
+  }): void;
+
+  type ElementInternalsMap = Array<{
+    ancestry: CrossTreeSelector;
+    internals: Record<string, InternalsData>;
+  }>;
+  type InternalsData = string | InternalsDataIdref | InternalsDataIdrefs;
+  type InternalsDataIdref = {
+    type: 'HTMLElement';
+    value: CrossTreeSelector;
+  };
+  type InternalsDataIdrefs = {
+    type: 'NodeList';
+    value: CrossTreeSelector[];
+  };
+
   // axe.frameMessenger
   type FrameMessenger = {
     open: (topicHandler: TopicHandler) => Close | void;
--- a/doc/element-internals.md
+++ b/doc/element-internals.md
@@ -1,0 +1,45 @@
+# ElementInternals
+
+Axe-core supports CustomElements with attached ElementInternals or ARIA Properties. However, JavaScript does not provide an API for accessing ElementInternals information on a node so axe-core must rely on developers implementing a [community protocol](https://github.com/webcomponents-cg/community-protocols/blob/main/proposals/element-internals-declaration.md) on their CustomElements in order to find this information.
+
+> [!Note]
+> At this time, axe-core only supports the ElementInternals `role` property and no others (e.g. `ariaLabel`), though support for other properties is planned. Additionally axe-core will not validate the value of the ElementInternals `role`, and many rules will not run against the element (e.g. `aria-required-attr`). Rules that do run may only be partial supported (e.g. `aria-required-children`).
+
+```js
+CustomElements.define(
+  'my-custom-button',
+  class MyCustomButton extends HTMLElement {
+    constructor() {
+      super();
+
+      const internals = this.attachInternals();
+      internals.role = 'button';
+
+      globalThis._elementInternals ??= new WeakMap();
+      globalThis._elementInternals.set(this, internals);
+    }
+  }
+);
+```
+
+In addition to the global WeakMap `globalThis._elementInternals`, axe-core also supports the following public properties or Symbols on the CustomElement:
+
+- `_internals`
+- `internals`
+- `internals_`
+- `Symbol('internals')`
+- `Symbol('privateInternals')`
+
+```js
+CustomElements.define(
+  'my-custom-button',
+  class MyCustomButton extends HTMLElement {
+    constructor() {
+      super();
+
+      this._internals = this.attachInternals();
+      this._internals.role = 'button';
+    }
+  }
+);
+```
--- a/doc/element-internals.md
+++ b/doc/element-internals.md
@@ -3,7 +3,7 @@
 Axe-core supports CustomElements with attached ElementInternals or ARIA Properties. However, JavaScript does not provide an API for accessing ElementInternals information on a node so axe-core must rely on developers implementing a [community protocol](https://github.com/webcomponents-cg/community-protocols/blob/main/proposals/element-internals-declaration.md) on their CustomElements in order to find this information.
 
 > [!Note]
-> At this time, axe-core only supports the ElementInternals `role` property and no others (e.g. `ariaLabel`), though support for other properties is planned. Additionally axe-core will not validate the value of the ElementInternals `role`, and many rules will not run against the element (e.g. `aria-required-attr`). Rules that do run may only be partial supported (e.g. `aria-required-children`).
+> At this time, axe-core only supports the ElementInternals `role` property and no others (e.g. `ariaLabel`), though support for other properties is planned. Additionally axe-core will not validate the value of the ElementInternals `role`, and many rules will not run against the element (e.g. `aria-required-attr`). Rules that do run may only be partially supported (e.g. `aria-required-children`).
 >
 > Lastly, support for ElementInternals is behind a feature flag `axe._enableElementInternals`, which must manually be set to `true` before axe runs.
 
--- a/doc/element-internals.md
+++ b/doc/element-internals.md
@@ -4,6 +4,8 @@ Axe-core supports CustomElements with attached ElementInternals or ARIA Properti
 
 > [!Note]
 > At this time, axe-core only supports the ElementInternals `role` property and no others (e.g. `ariaLabel`), though support for other properties is planned. Additionally axe-core will not validate the value of the ElementInternals `role`, and many rules will not run against the element (e.g. `aria-required-attr`). Rules that do run may only be partial supported (e.g. `aria-required-children`).
+>
+> Lastly, support for ElementInternals is behind a feature flag `axe._enableElementInternals`, which must manually be set to `true` before axe runs.
 
 ```js
 CustomElements.define(
--- a/doc/external-apis.md
+++ b/doc/external-apis.md
@@ -53,7 +53,7 @@ const internals = await chrome.scripting.executeScript({
   target: {
     tabId: tab.id
   },
-  files: ['/node_moduels/axe-core/gather-internals.js'],
+  files: ['/node_modules/axe-core/gather-internals.js'],
   world: 'MAIN'
 });
 ```
--- a/doc/external-apis.md
+++ b/doc/external-apis.md
@@ -1,0 +1,63 @@
+# External APIs
+
+Axe externalAPIs can be used to configure to pass information to axe-core that it would be otherwise unable to obtain on it's own (such as when it is running in an isolated JavaScript context of an extension). By default axe-core will try to gather the information if the external API for it is not set. If the external API is set axe-core will only rely on the passed in data and will not gather it itself.
+
+```js
+axe.externalAPIs({
+  // Pass ElementInternal data to axe-core
+  elementInternals() {
+    return Promise.resolve(internalData);
+  }
+});
+```
+
+## axe.externalAPIS({ elementInternals })
+
+`elementInternals` is a function that can be used to pass ElementInternal data for the elements on the page. It should return a Promise with the ElementInternal data array. The data inside the ElementInternal array must have the following properties:
+
+```js
+axe.externalAPIs({
+  elementInternals() {
+    return Promise.resolve([
+      {
+        // The CSS ancestry selector, which is either a string or an array of strings if the element is inside a ShadowDom tree. Generated using `axe.utils.getAncestry(node)`
+        ancestry: 'html > body > main > my-custom-button',
+
+        // Object of each ElementInternals property name and value. The properties must be serialized so some serialization work is needed for element reference properties (idref(s))
+        internals: {
+          // A simple String value
+          role: 'button',
+
+          // An idref value. The type will always be "HTMLElement" and the value is a CSS ancestry selector
+          ariaActiveDescendantElement: {
+            type: 'HTMLElement',
+            value: 'html > body > main > my-custom-button > div:nth-child(1)'
+          },
+
+          // An idrefs value. The type will always be "NodeList" and the value is an array of CSS ancestry selectors for each node
+          ariaLabelledbyElements: {
+            type: 'NodeList',
+            value: ['html > body > div:nth-child(4)']
+          }
+        }
+      }
+    ]);
+  }
+});
+```
+
+Axe-core provides a package script `gather-internals.js` that can used to inject into the main context when working in an extension. The returned object from the script can be used directly as the returned value of the Promise for `elementInternals`.
+
+```js
+const internals = await chrome.scripting.executeScript({
+  target: {
+    tabId: tab.id
+  },
+  files: ['/node_moduels/axe-core/gather-internals.js'],
+  world: 'MAIN'
+});
+```
+
+## axe.externalAPIS({ elementInternalsTimeout })
+
+Since gathering ElementInternals data is an async operation, you can configure how long axe-core will wait for `elementInternals` promise to resolve. By default the timeout is set to 1 second. If the timeout occurs axe-core will not run and will throw an error.
--- a/doc/external-apis.md
+++ b/doc/external-apis.md
@@ -46,7 +46,7 @@ axe.externalAPIs({
 });
 ```
 
-Axe-core provides a package script `gather-internals.js` that can used to inject into the main context when working in an extension. The returned object from the script can be used directly as the returned value of the Promise for `elementInternals`.
+Axe-core provides a package script `gather-internals.js` that can be used to inject into the main context when working in an extension. The returned object from the script can be used directly as the returned value of the Promise for `elementInternals`.
 
 ```js
 const internals = await chrome.scripting.executeScript({
--- a/doc/external-apis.md
+++ b/doc/external-apis.md
@@ -58,6 +58,11 @@ const internals = await chrome.scripting.executeScript({
 });
 ```
 
+See [element-internals.md](element-internals.md) for more information on what ElementInternal properties are supported.
+
+> [!Note]
+> Support for ElementInternals is behind a feature flag `axe._enableElementInternals`, which must manually be set to `true` before axe runs, even when passing in `elementInternals` data.
+
 ## axe.externalAPIs({ elementInternalsTimeout })
 
 Since gathering ElementInternals data is an async operation, you can configure how long axe-core will wait for `elementInternals` promise to resolve. By default the timeout is set to 1 second. If the timeout occurs axe-core will not run and will throw an error.
--- a/doc/external-apis.md
+++ b/doc/external-apis.md
@@ -11,7 +11,7 @@ axe.externalAPIs({
 });
 ```
 
-## axe.externalAPIS({ elementInternals })
+## axe.externalAPIs({ elementInternals })
 
 `elementInternals` is a function that can be used to pass ElementInternal data for the elements on the page. It should return a Promise with the ElementInternal data array. The data inside the ElementInternal array must have the following properties:
 
--- a/doc/external-apis.md
+++ b/doc/external-apis.md
@@ -1,6 +1,6 @@
 # External APIs
 
-Axe externalAPIs can be used to configure to pass information to axe-core that it would be otherwise unable to obtain on it's own (such as when it is running in an isolated JavaScript context of an extension). By default axe-core will try to gather the information if the external API for it is not set. If the external API is set axe-core will only rely on the passed in data and will not gather it itself.
+Axe externalAPIs can be used to pass information to axe-core that it would be otherwise unable to obtain on its own (such as when it is running in an isolated JavaScript context of an extension). By default axe-core will try to gather the information if the external API for it is not set. If the external API is set axe-core will only rely on the passed in data and will not gather it itself.
 
 ```js
 axe.externalAPIs({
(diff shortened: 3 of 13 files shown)


lib/core/public/external-apis.js
```diff
@@ -0,0 +1,89 @@
+import { shadowSelect, getNodeFromTree } from '../utils';
+
+export const external = {};
+
+export default function externalAPIs({
+  elementInternalsTimeout = 1000,
+  elementInternals
+}) {
+  if (typeof elementInternals !== undefined) {
```
chutchins25: `typeof` returns a string, so this compares `"undefined"` to the `undefined` value and is always `true`. Calling `axe.externalAPIs({ elementInternalsTimeout: 500 })` (or even `axe.externalAPIs({})`) currently throws `TypeError` because we fall through to the inner `typeof !== 'function'` check.

```suggestion
  if (typeof elementInternals !== 'undefined') {
```

Worth adding a test like `assert.doesNotThrow(() => externalAPIs({ elementInternalsTimeout: 500 }))` to lock this in.

lib/core/public/external-apis.js
```diff
@@ -0,0 +1,89 @@
...
+            continue;
+          }
+
+          // convert idref(s) ancestries back to nodes
+          for (const [key, val] of Object.entries(internals)) {
+            if (typeof val === 'string') {
+              continue;
+            }
+
+            const { type, value } = val;
```
chutchins25: If any property in `internals` is `null`, this destructure throws (`typeof null === 'object'`, so the string-skip above misses it), and the whole audit run rejects via the outer `.catch(rej)`. Real ElementInternals payloads can legitimately have null-valued ARIA properties (e.g. an unset `ariaPressed`), and the docs don't say nulls are disallowed.

Cheapest fix is to widen the skip to cover non-objects:

```suggestion
            if (typeof val !== 'object' || val === null) {
              continue;
            }

            const { type, value } = val;
```

Worth a test like `Promise.resolve([{ ancestry: '…', internals: { role: 'button', ariaPressed: null } }])` to lock it in.

doc/external-apis.md
```diff
@@ -0,0 +1,63 @@
+# External APIs
+
+Axe externalAPIs can be used to configure to pass information to axe-core that it would be otherwise unable to obtain on it's own (such as when it is running in an isolated JavaScript context of an extension). By default axe-core will try to gather the information if the external API for it is not set. If the external API is set axe-core will only rely on the passed in data and will not gather it itself.
```
chutchins25: Worth a cross-link to `doc/element-internals.md` from this file — a reader following the `axe.externalAPIs({ elementInternals })` setup here can do everything correctly and still see no behavior change because `axe._enableElementInternals` defaults to off. The integration test in this PR demonstrates exactly that gate.

chutchins25: Not a blocker, just a consistency question — two small grammar nits on this line: "configure to pass" reads as a double-verb, and "it's" should be "its" (possessive).

```suggestion
Axe externalAPIs can be used to pass information to axe-core that it would be otherwise unable to obtain on its own (such as when it is running in an isolated JavaScript context of an extension). By default axe-core will try to gather the information if the external API for it is not set. If the external API is set axe-core will only rely on the passed in data and will not gather it itself.
```

chutchins25 (commented):
Nice work on landing a clean public surface for this — the audit-time queueing is well-isolated and the test coverage is broad. A few things to flag before merge:

**One critical correctness bug** in the option validation (`typeof x !== undefined` always evaluates `true` — see inline on `lib/core/public/external-apis.js:9`). With the current code, `axe.externalAPIs({ elementInternalsTimeout: 500 })` or any call that omits `elementInternals` throws an unhelpful `TypeError`. None of the existing tests catch this because every test passes a value.

**One additional correctness gap** — `null` values inside `internals` crash the destructure (`typeof null === 'object'`, so the string-skip misses them) and reject the whole audit run. Real ElementInternals payloads can legitimately have null ARIA properties.

A handful of doc fixes (the API heading is spelled `axe.externalAPIS` in two places, a `node_moduels` typo, references to a `gather-internals.js` script that isn't in this PR, missing cross-link to the `_enableElementInternals` flag prerequisite) and three sync-`done()` tests that don't actually verify the async path.

## Missing artifacts before merge

A few things commonly expected for a new public API (per the repo guidelines) don't appear in this PR:

- **`axe.d.ts`** — no TypeScript definition for `axe.externalAPIs`. TS consumers won't get completion or compile-time checking on release.
- **`doc/API.md`** — the new `axe.externalAPIs` entry point isn't linked or referenced from the API reference; the new `doc/external-apis.md` and `doc/element-internals.md` are good but don't connect into the existing API surface map.
- **`CHANGELOG.md`** — no entry for what is a meaningful new public API + behavior gate (`_enableElementInternals`).

Happy to be wrong on any of these if they're tracked in a follow-up — just call it out so reviewers know it's intentional.

doc/external-apis.md
```diff
@@ -0,0 +1,63 @@
...
+```js
+axe.externalAPIs({
+  // Pass ElementInternal data to axe-core
+  elementInternals() {
+    return Promise.resolve(internalData);
+  }
+});
+```
+
+## axe.externalAPIS({ elementInternals })
```
chutchins25: Heading capitalizes the `S` (`externalAPIS`) — the actual API is `axe.externalAPIs`.

```suggestion
## axe.externalAPIs({ elementInternals })
```

doc/external-apis.md
```diff
@@ -0,0 +1,63 @@
...
+            value: ['html > body > div:nth-child(4)']
+          }
+        }
+      }
+    ]);
+  }
+});
+```
+
+Axe-core provides a package script `gather-internals.js` that can used to inject into the main context when working in an extension. The returned object from the script can be used directly as the returned value of the Promise for `elementInternals`.
```
chutchins25: Couldn't find a `gather-internals.js` script anywhere in this PR or in the repo at HEAD. Is it landing in a follow-up, or should this section be cut for now? As-is, a reader following the extension example will hit a 404 with no path forward.

If it's coming separately, worth at minimum calling that out here (e.g. "shipping in #NNNN") so the example isn't a dead end on release.

chutchins25: Not a blocker — "can used" → "can be used".

```suggestion
Axe-core provides a package script `gather-internals.js` that can be used to inject into the main context when working in an extension. The returned object from the script can be used directly as the returned value of the Promise for `elementInternals`.
```

doc/external-apis.md
```diff
@@ -0,0 +1,63 @@
...
+
+```js
+const internals = await chrome.scripting.executeScript({
+  target: {
+    tabId: tab.id
+  },
+  files: ['/node_moduels/axe-core/gather-internals.js'],
```
chutchins25: Typo — `node_moduels` should be `node_modules`. Copy-paste of this example would silently fail.

```suggestion
  files: ['/node_modules/axe-core/gather-internals.js'],
```

doc/external-apis.md
```diff
@@ -0,0 +1,63 @@
...
+  world: 'MAIN'
+});
+```
+
+## axe.externalAPIS({ elementInternalsTimeout })
```
chutchins25: Same `externalAPIS` → `externalAPIs` typo as on line 14.

```suggestion
## axe.externalAPIs({ elementInternalsTimeout })
```

doc/external-apis.md
```diff
@@ -0,0 +1,63 @@
...
+
+Since gathering ElementInternals data is an async operation, you can configure how long axe-core will wait for `elementInternals` promise to resolve. By default the timeout is set to 1 second. If the timeout occurs axe-core will not run and will throw an error.
```
chutchins25, line 68: "will throw an error" — the implementation rejects the Promise rather than throwing synchronously. Consumers reading this would wire `try/catch` around the `axe.run` call site, which won't catch a Promise rejection.

```suggestion
Since gathering ElementInternals data is an async operation, you can configure how long axe-core will wait for `elementInternals` promise to resolve. By default the timeout is set to 1 second. If the timeout occurs `axe.run` will reject with an error.
```

doc/element-internals.md
```diff
@@ -0,0 +1,47 @@
+# ElementInternals
+
+Axe-core supports CustomElements with attached ElementInternals or ARIA Properties. However, JavaScript does not provide an API for accessing ElementInternals information on a node so axe-core must rely on developers implementing a [community protocol](https://github.com/webcomponents-cg/community-protocols/blob/main/proposals/element-internals-declaration.md) on their CustomElements in order to find this information.
+
+> [!Note]
+> At this time, axe-core only supports the ElementInternals `role` property and no others (e.g. `ariaLabel`), though support for other properties is planned. Additionally axe-core will not validate the value of the ElementInternals `role`, and many rules will not run against the element (e.g. `aria-required-attr`). Rules that do run may only be partial supported (e.g. `aria-required-children`).
```
chutchins25: Not a blocker — "only partial supported" → "only partially supported".

```suggestion
> At this time, axe-core only supports the ElementInternals `role` property and no others (e.g. `ariaLabel`), though support for other properties is planned. Additionally axe-core will not validate the value of the ElementInternals `role`, and many rules will not run against the element (e.g. `aria-required-attr`). Rules that do run may only be partially supported (e.g. `aria-required-children`).
```

test/core/public/external-apis.js
```diff
@@ -0,0 +1,297 @@
...
+        externalAPIs({
+          elementInternals() {
+            return Promise.resolve(type);
+          }
+        });
+
+        external.elementInternals(() => {}, isNotCalled);
+      }
+
+      done();
```
chutchins25: `done()` fires synchronously after the loop, but `external.elementInternals(...)` runs its success/reject callbacks in a microtask. So this passes as long as nothing throws synchronously — it doesn't verify the async path actually completed cleanly, and any post-`done()` `isNotCalled` error would just leak into the next test.

Same pattern in the two tests below at L256 and L275. Could either gather the promises and `Promise.all` them, or call `done()` from inside the resolve callback of the last iteration.

test/core/public/external-apis.js
```diff
@@ -0,0 +1,297 @@
...
+                internals: type
+              }
+            ]);
+          }
+        });
+
+        external.elementInternals(() => {}, isNotCalled);
+      }
+
+      done();
```
chutchins25: Same sync-`done()` issue as noted on L221 — the async path in `external.elementInternals` runs after this `done()` returns.

test/core/public/external-apis.js
```diff
@@ -0,0 +1,297 @@
...
+                internals: {}
+              }
+            ]);
+          }
+        });
+
+        external.elementInternals(() => {}, isNotCalled);
+      }
+
+      done();
```
chutchins25: Same sync-`done()` issue as noted on L221.

lib/core/public/external-apis.js
straker: The `gather-internals` script that this relies on will not support passing `null` values across, but rather omit the value entirely, so I'm ok not supporting null.

doc/external-apis.md
straker: https://github.com/dequelabs/axe-core/pull/5099, which is required and will be in when we release

doc/external-apis.md
straker: A rejected promise during the audit bubbles up to a throw in axe-core using a callback function, but rejects if using a promise. Not sure which would be the best wording in this case.

lib/core/public/external-apis.js
```diff
@@ -0,0 +1,89 @@
+import { shadowSelect, getNodeFromTree } from '../utils';
+
+export const external = {};
+
+export default function externalAPIs({
```
WilcoFiers, line 9: Can you add a type definition for this plz?

lib/core/public/external-apis.js
```diff
@@ -0,0 +1,89 @@
...
+      throw new TypeError(
+        'elementInternals must be a function that returns a Promise'
+      );
+    }
+
+    setElementInternals(elementInternalsTimeout, elementInternals);
+  }
+}
+
+function setElementInternals(elementInternalsTimeout, elementInternals) {
```
WilcoFiers: Any reason this isn't an async function?

lib/core/public/external-apis.js
```diff
@@ -0,0 +1,89 @@
+import { shadowSelect, getNodeFromTree } from '../utils';
+
+export const external = {};
```
WilcoFiers: This feels a little odd. Why don't you put a `getElementInternals` on here, or maybe just export it directly, and have that function use `externalElementInternals` if provided, or return null if not.

lib/core/public/external-apis.js
```diff
@@ -0,0 +1,89 @@
...
+        'elementInternals must be a function that returns a Promise'
+      );
+    }
+
+    setElementInternals(elementInternalsTimeout, elementInternals);
+  }
+}
+
+function setElementInternals(elementInternalsTimeout, elementInternals) {
+  external.elementInternals = (res, rej) => {
```
WilcoFiers: I think we need better variable names here. You have two things named `elementInternals`, neither of which is actually the elementInternals data, which you end up calling `results`. Maybe `externalElementInternals` and `getElementInternals`, that way you can rename `results` to `elementInternalsMap`.

lib/core/public/external-apis.js
```diff
@@ -0,0 +1,89 @@
...
+        if (timeoutCalled) {
+          return;
+        }
+
+        if (!results || !Array.isArray(results)) {
+          return;
+        }
+
+        // resolve ancestry strings to nodes
+        for (const { internals, ancestry } of results) {
```
WilcoFiers: We didn't test that results is a destructable object. This can throw if we get something odd here. Either we should fail silently (like you're doing above) or we should throw with a clear message of what's going on. This function is also a bit of a mix of validation and functionality. Consider creating a separate normalize function here. Maybe even one that converts this to a Map with actual elements like we get from `globalThis._elementInternals` so that the result of this is in a shape axe can consume.

lib/core/public/external-apis.js
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        res();
+      })
+      .catch(rej);
+  };
+}
+
+// expose for testing
+export function _reset() {
```
WilcoFiers: I think we should allow resetting with `axe.externalAPIs(null)`. It'll make testing easier, and while I don't see an immediate use case for needing to reset this, having it part of the public API won't hurt.

lib/core/public/external-apis.js
WilcoFiers, line 9: Different PR, but what would you think about adding `nodeSerializer` and `frameMessenger` here, and deprecating how we're doing that now. That way we have one API for connecting axe-core to external resources.

lib/core/public/external-apis.js
```diff
@@ -0,0 +1,89 @@
...
+      .then(results => {
+        clearTimeout(timeout);
+
+        // timeout already called so don't process the response
+        if (timeoutCalled) {
+          return;
+        }
+
+        if (!results || !Array.isArray(results)) {
+          return;
```
WilcoFiers: I don't think this should fail silently. I don't think we should throw here, but we may want to console.log an error at least. Might be even better if we raised a separate incomplete issue, similar to `frame-tested`. That's out of scope for this work though.

test/core/base/audit.js
```diff
@@ -1243,6 +1246,89 @@ describe('Audit', () => {
...
+        });
+
+        audit.run(
+          { include: [axe.utils.getFlattenedTree()[0]] },
+          {},
+          results => {
+            assert.isTrue(stub.called);
+            done();
+          },
+          isNotCalled
```
WilcoFiers, line 1267: This doesn't work does it? This is an async error. How would that be caught by the test? I think this should be:

```suggestion
          (e) => done(e)
```

lib/core/public/external-apis.js
```diff
@@ -0,0 +1,89 @@
...
+            const { type, value } = val;
+            if (type === 'HTMLElement') {
+              internals[key] = shadowSelect(value);
+            } else if (type === 'NodeList') {
+              internals[key] = value.map(a => shadowSelect(a));
+            }
+          }
+
+          // set internals directly onto the vNode
+          vNode.elementInternals = internals;
```
WilcoFiers: I don't think we should pass an external object directly into axe. Safer to create a new object and assign properties that we know are safe to it. So only strings and things we looked up

lib/core/public/external-apis.js
```diff
@@ -0,0 +1,89 @@
...
+
+          // convert idref(s) ancestries back to nodes
+          for (const [key, val] of Object.entries(internals)) {
+            if (typeof val === 'string') {
+              continue;
+            }
+
+            const { type, value } = val;
+            if (type === 'HTMLElement') {
+              internals[key] = shadowSelect(value);
```
WilcoFiers: We need to think about what happens when this returns null. I'm not entirely sure, lets discuss.

WilcoFiers (dismissed):

lib/core/public/external-apis.js
straker: Possibly? That sounds like a breaking change though to remove it, but a feature to allow it to configure those things.

lib/core/public/external-apis.js
straker: Im not sure about the mapping to a structure similar to `globalThis._elementInternals`. The whole reason we're adding the internals directly to the vNode is to ensure that the external data isn't saved globally anywhere and only lives for the current run. Axe-core doesn't interact with any other place to look at internals other than the `globalThis. _elementInternals` (which we wouldn't want to override as it would live after the run) and the vNode itself.

lib/core/public/external-apis.js
```diff
@@ -0,0 +1,171 @@
...
+      );
+    }
+
+    getElementInternals = getInternals;
+  }
+  // reset if not provided
+  else {
+    getElementInternals = null;
+  }
+}
```
WilcoFiers, line 36: This function has an odd gotcha in it I think. I wouldn't expect things to unset if they are left undefined:

```
axe.externalAPI({ getInternals })
axe.externalAPI({ getInternalsTimeout: 300 }) // I wouldn't expect this to unset getInternals
```

Ideally that doesn't happen, but the extension code base being what it is, it might. Best to be prepared. I think I'd have it work like this:

```suggestion
export default function externalAPIs({
  getElementInternalsTimeout: internalsTimeout,
  getElementInternals: getInternals
} = {}) {
  if (typeof getElementInternalsTimeout === 'number') {
    getElementInternalsTimeout = internalsTimeout;
  } else if (getElementInternalsTimeout === null) {
    getElementInternalsTimeout = GET_ELEMENT_INTERNALS_TIMEOUT_DEFAULT
  } else if (getElementInternalsTimeout !== undefined) {
    throw ...
  }

  if (getInternals === null || typeof getInternals === 'function') {
    getElementInternals = getInternals;
  } else if (getInternals !== undefined) {
    throw new TypeError('getElementInternals must be a function that returns a Promise');
  }
}
```

test/core/base/audit.js
```diff
@@ -102,6 +103,7 @@ describe('Audit', () => {
   });
 
   afterEach(() => {
+    externalAPIs();
```
WilcoFiers:
```suggestion
    // reset elementInternals
    externalAPIs({ elementInternals: null, elementInternalsTimeout: null });
```

axe.d.ts
```diff
@@ -674,6 +674,31 @@ declare namespace axe {
...
   ): void;
 
+  /**
+   * Run axe in the current window only
+   * @param   {ElementContext} context  Optional The `Context` specification object @see Context
+   * @param   {RunOptions}     options  Optional Options passed into rules or checks, temporarily modifying them.
+   * @returns {Promise<PartialResult>}  Partial result, for use in axe.finishRun.
+   */
+  function externalAPIs(params?: {
+    getElementInternalsTimeout?: number;
```
WilcoFiers: Hmm, should this be `get`? This is just a number, we're not passing in a function.

lib/core/public/external-apis.js
```diff
@@ -1,89 +1,171 @@
...
+      }
+    }
+
+    // set internals directly onto the vNode
+    vNode.elementInternals = internals;
+  }
+}
+
+export const external = {
+  setElementInternals
```
WilcoFiers: Sry should have caught this in my last review. Is `set` the right verb here? Set suggests I need to pass something in that then gets set. That's not what this does. This is more like "load" or maybe "fetch" or something.

test/core/public/external-apis.js
```diff
@@ -306,9 +307,69 @@ describe('externalAPIs', () => {
...
+      await external.setElementInternals(subLogger);
+
+      assert.isTrue(subLogger.called);
+      assert.include(
+        subLogger.firstCall.args[0],
+        'Unable to locate node using selector'
+      );
+    });
+
+    it("sets a getter that throws if it can't find an idref element", async () => {
```
WilcoFiers, line 395: I think an integration test would be good for this too to show this escalates to an incomplete when needed.

WilcoFiers (dismissed):

lib/core/public/external-apis.js
straker: I'd be ok with `load`, but not `fetch` as it does more than just get the data. The external function does set the element internals on the vnodes so that's why I used `set`.

axe.d.ts
straker: I named it `get` just to match the name of the function it times out. Otherwise it'd be `elementInternalsTimeout` and `getElementInternals` function.

test/core/base/audit.js
```diff
@@ -103,7 +103,11 @@ describe('Audit', () => {
   });
 
   afterEach(() => {
-    externalAPIs();
+    // reset elementInternals
+    externalAPIs({
+      elementInternals: null,
+      elementInternalsTimeout: null
```
WilcoFiers:
```suggestion
    externalAPIs({
      getElementInternals: null,
      getElementInternalsTimeout: null
```

WilcoFiers (approved):

axe.d.ts
WilcoFiers: Yeah, that's what I had in mind. That seemed more appropriate. I can go with this if you prefer it.

WilcoFiers (dismissed):
Uh, no the other one. :P Just the typo.

WilcoFiers (approved):

PR #5108, opened 2026-05-18
feat: expose axe.resetLocale() to restore the default locale

### LINKED ISSUE #5107: feat: expose axe.resetLocale() to restore the default English locale without full axe.reset()
### Product

axe-core

### Feature Description

**Problem**

`axe.configure({ locale })` has no inverse. Once a non-English locale has been applied, there's no way to revert to the default English messages short of `axe.reset()` — which also clears branding, rule enable/disable overrides, `frameMessenger`, `noHtml`, `allowedOrigins`, and anything else `Audit._init()` resets.

Consumers that let end users switch languages at runtime hit this: after applying e.g. Japanese via `axe.configure({ locale })`, there's no clean way to roll back to the default English messages without also destroying every other `configure()` call they made during setup. The only workarounds are heavy (re-load axe-core entirely) or fragile (snapshot internals manually).

**What already exists internally**

`lib/core/base/audit.js` already has the exact machinery:

- `Audit._setDefaultLocale()` (line 41) — snapshots the default English locale on the first `applyLocale` call, and noops on subsequent calls so the snapshot remains the original.
- `Audit._resetLocale()` (line 85) — applies the saved default.

Both are private (underscore prefix) and only reachable via `axe.reset()` → `audit.resetRulesAndChecks()` → `_resetLocale()`, which also re-`_init()`s everything else.

**Proposed API**

Add a public `axe.resetLocale()` that maps directly to `audit._resetLocale()` and nothing else:

```js
// lib/core/public/reset-locale.js
function resetLocale() {
  const audit = axe._audit;
  if (!audit) {
    throw new Error('No audit configured');
  }
  audit._resetLocale();
}
export default resetLocale;
```

Wired into the public exports the same way `axe.reset` is.

**Semantics**

- No-op if no non-default locale has ever been applied (matches existing `_resetLocale` early-return when `_defaultLocale` is null).
- Does not affect branding, rule enable/disable overrides, `tagExclude`, `noHtml`, `allowedOrigins`, `frameMessenger`, or anything else `axe.reset()` touches.
- Safe to call repeatedly.

**Why a new method rather than overloading `configure`**

Alternatives would be `axe.configure({ locale: null })` or a `resetLocale: true` flag. A dedicated method is more discoverable and avoids overloading the existing `locale` param semantics, which today only accepts a `Locale` object.

zlayaAvocado:
Adds a public `axe.resetLocale()` that reverts the locale back to the default English snapshot taken on the first `applyLocale` call, without touching any other configuration.

Closes #5107

## Motivation

`axe.configure({ locale })` has no inverse, and `axe.reset()` also clears branding, rule enable/disable overrides, `frameMessenger`, `noHtml`, `allowedOrigins`, and anything else `Audit._init()` resets. Consumers that let end users switch languages at runtime have no clean way to roll back to the default English messages without destroying the rest of their setup. The internal machinery for this already exists (`Audit._setDefaultLocale` + `Audit._resetLocale`) — this PR just exposes it as a public method.

## Implementation

- New `lib/core/public/reset-locale.js`, wired into `lib/core/core.js` next to `axe.reset`.
- Delegates to the existing private `audit._resetLocale()`. Throws `No audit configured` if called before `_load`, matching `axe.reset`.
- No-op when no non-default locale has ever been applied (existing `_resetLocale` early-return when `_defaultLocale` is null).
- Does not touch branding, rule enable/disable, `tagExclude`, `noHtml`, `allowedOrigins`, `frameMessenger`, or standards.
- Safe to call repeatedly.
- Type declaration added to `axe.d.ts`.
- API documentation added to `doc/API.md`.

## Tests

`test/core/public/reset-locale.js` covers:
- throws when no audit is configured
- no-op when no locale has been applied
- restores original check messages after `axe.configure({ locale })`
- restores original `lang`
- safe to call repeatedly
- does not affect other configuration (verified by registering a custom rule and confirming it survives)

All 6 pass; existing `reset` and `configure` tests still pass (61/61).


doc/API.md
```diff
@@ -296,6 +297,26 @@ axe.reset();
...
 
+### API Name: axe.resetLocale
+
+#### Purpose
+
+Restore the default locale that was active before any `axe.configure({ locale })` call, without touching the rest of the configuration.
+
+#### Description
+
+`axe.configure({ locale })` has no inverse, and `axe.reset()` also clears branding, rule enable/disable overrides, `frameMessenger`, and other configuration. `axe.resetLocale()` reverts only the locale (rule descriptions, check messages, failure summaries, `lang`) back to the default that was in effect before the first `applyLocale` call. It is a no-op if no non-default locale has ever been applied, and safe to call repeatedly.
```
straker: This describes more of why the api exists instead of what it's doing. A simple description of what the api does and an example should suffice.

straker (dismissed):
Just a minor nit on the docs, but not something that really blocks the pr. Up to you if you'd like to chagne it.

straker (approved):

PR #5109, opened 2026-05-18
feat(aria/get-aria-value): new function to get aria values of a node

straker:
The `getAriaValue` function gets the value from the 3 different locations for aria data (attribute, property, element internals). The function will always return am object of `{ value: String, source: String }` or null. However this is a slight problem when an idref(s) property is used as they return DOM nodes or arrays of such. In order to still report what the value is as a string, the function massages the data into a string by taking the node name of the DOM node as that's probably the easiest way to report it to the user. This would result in something like `node.ariaLabelledbyElements = [div, div];` having a value of `'[div,div]'` instead of `''` if we only used the attribute value.

In a separate PR we'll create two functions for resolving idref(s) values into vNodes that should be used 99% of the time instead of `getAriaValue`. `getAriaValue` for idref(s) should only be used when reporting the value to the user (such as in `aria-validate-attr-value`).

Closes: #5042


chutchins25 (commented):
Nice work — clean three-source resolution and thorough tests. A few questions/suggestions, none blocking. (Verified the two moderate ones against the code: `SerialVirtualNode` has no `actualNode`, and the single-node branch only matches `HTMLElement`.)

lib/commons/aria/get-aria-value.js
```diff
@@ -0,0 +1,100 @@
...
+    return value;
+  }
+
+  // return null for the attribute if the value is empty but the idref(s) prop is not
+  return prop && vNode.actualNode[prop] ? null : value;
+}
+
+function getPropertyValue(vNode, attrStandard) {
+  const { prop } = attrStandard;
+  return prop ? vNode.actualNode[prop] : null;
```
chutchins25: `getPropertyValue` (and the idref branch at line 89) read `vNode.actualNode[prop]`. The JSDoc accepts a `VirtualNode`, but `SerialVirtualNode` (virtual-rules) has no `actualNode`, so for any *absent* attribute the property source would hit `undefined[prop]` and throw. `getInternalValue` happens to be safe because `vNode.elementInternals` is `undefined` on serial nodes. Is `getAriaValue` intended to be real-DOM-only? If so, worth saying so in the JSDoc; if not, a guard would cover it:

```suggestion
  return prop ? vNode.actualNode?.[prop] : null;
```

(and the same `?.` on line 89's `vNode.actualNode[prop]`).

lib/commons/aria/get-aria-value.js
```diff
@@ -0,0 +1,100 @@
...
+  const { vNode } = nodeLookup(node);
+  const { lowercase } = options;
+
+  for (const { source, getValue } of sources) {
+    let value = getValue(vNode, attrStandard, attrName);
+    if (value === null) {
+      continue;
+    }
+
+    if (value instanceof window.HTMLElement) {
```
chutchins25: idref targets can be SVG elements (`SVGElement`, not `HTMLElement`). For a single SVG node resolved via the AOM property this branch is skipped, the array branch is skipped, and line 61's `value.trim()` then runs on the Element and throws. `window.Node` would cover both HTML and SVG:

```suggestion
    if (value instanceof window.Node) {
```

Could be worth a test with an SVG idref target (e.g. `aria-activedescendant` pointing at an SVG element) to lock this in.

lib/commons/aria/get-aria-value.js
```diff
@@ -0,0 +1,100 @@
...
+    getValue: getPropertyValue
+  },
+  {
+    source: 'internals',
+    getValue: getInternalValue
+  }
+];
+
+/**
+ * Get the value of a non idref(s) ARIA attribute, property, or ElementInternal
```
chutchins25: The function handles idref(s) (it's a core feature here), so "non" reads as leftover:

```suggestion
 * Get the value of an ARIA attribute, property, or ElementInternal
```

lib/commons/aria/get-aria-value.js
```diff
@@ -0,0 +1,100 @@
...
+ * @method getAriaValue
+ * @memberof axe.commons.aria
+ * @instance
+ * @param {Node|Element|VirtualNode} node
+ * @param {String} attrName The ARIA attribute name to get the value of. Use the ARIA attr name even if getting values from properties.
+ * @param {Object} [options]
+ * @param {Boolean} [options.lowercase=false] toLowerCase the value
+ * @return {String|null} value
```
chutchins25: Return tag doesn't match the actual return shape (`{ value, source }` or `null`):

```suggestion
 * @return {{value: String, source: String}|null} value
```

lib/commons/aria/get-aria-value.js
```diff
@@ -0,0 +1,100 @@
...
+    return value;
+  }
+
+  // return the idref(s) attribute if it is not empty
+  if (!!value) {
+    return value;
+  }
+
+  // return null for the attribute if the value is empty but the idref(s) prop is not
+  return prop && vNode.actualNode[prop] ? null : value;
```
chutchins25: Reflected element-array props (`ariaLabelledByElements`, etc.) return an empty array `[]` when nothing is associated, and `[]` is truthy — so a genuinely empty `aria-labelledby=""` would fall through to the property source and report `'[]'` rather than `''`. Is that the intended reported value for an empty idrefs attribute, or should the empty-array case be treated as "prop not set"?

lib/commons/aria/get-aria-value.js
straker: This is a good point, and one we'll need to replicate in `walk-tree` as well as it uses `HTMLElement`

chutchins25 (approved):
Thanks for adding those additional tests! Looks good to me.

WilcoFiers (approved):

PR #5110, opened 2026-05-19
ci: ignore gather-internals.js from import deploy validation

straker:
`@next` [failed to deploy](https://github.com/dequelabs/axe-core/actions/runs/26101303110/job/76753975655#step:8:32) with the new `gather-internals.js` export script as it's not an importable file. Fixed to ignore that file from the check as it is not suppose to be imported.

### DIFF
--- a/.github/bin/validate-package.mjs
+++ b/.github/bin/validate-package.mjs
@@ -210,7 +210,13 @@ defined files in the \`files\` array of \`package.json\`.
 | File | Status | Version |\n|------|--------|--------|
 `;
 
-  const importTargets = [...pkg.files.map(file => `${pkg.name}/${file}`)];
+  // these files are not part of the importable api
+  const nonImportableFiles = ['gather-internals.js'];
+  const importTargets = [
+    ...pkg.files
+      .filter(file => !nonImportableFiles.includes(file))
+      .map(file => `${pkg.name}/${file}`)
+  ];
   let anyCaught = false;
 
   console.log('Validating package files are importable:');
--- a/.github/bin/validate-package.mjs
+++ b/.github/bin/validate-package.mjs
@@ -290,6 +290,27 @@ defined files in the \`files\` array of \`package.json\`.
     }
   }
 
+  // check that the non-importable files are parsable
+  for (const file of nonImportableFiles) {
+    const target = `${pkg.name}/${file}`;
+    try {
+      const resolved = import.meta.resolve(target);
+      if (resolved.includes('node_modules')) {
+        console.error(`✗ ${target} resolves to node_modules`);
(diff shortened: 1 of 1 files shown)


chutchins25 (dismissed):
Looks good to me.

Garbee (dismissed):
We should probably do something like this:

```javascript
import { execSync } from 'node:child_process';

const browserOnlyFiles = ['gather-internals.js'];

for (const file of browserOnlyFiles) {
  const target = `${pkg.name}/${file}`;
  try {
    const resolved = import.meta.resolve(target);
    if (resolved.includes('node_modules')) {
      console.error(`✗ ${target} resolves to node_modules`);
      summary += `| \`${target}\` | ✗ Resolves to node_modules |\n`;
      exitCode++;
      continue;
    }
    execSync(`node --check "${fileURLToPath(resolved)}"`, { stdio: 'pipe' });
    console.info(`✓ ${target} (parse-only, browser-only)`);
    summary += `| \`${target}\` | ✓ Parse OK (browser-only) | n/a |\n`;
  } catch (error) {
    console.error(`✗ ${target}: ${error.message}`);
    summary += `| \`${target}\` | ✗ Parse failed (browser-only) | n/a |\n`;
    exitCode++;
  }
}
```

Reason being, this package *is* something that is expected to go out of the built file. Which is what this entire file exists to ensure continues to work. This change still checks the resolved path exists and is packaged well, without trying to execute the contents which is what is causing the problem.

This is in fact an importable file. It just is only importable in browser scope only. Although, maybe that indicates a potential issue with the implementation? Not sure if there is a rule that what JS we export should import fine in node and browser.

pwsmaster (approved):

pwsmaster (approved):

straker:
> This is in fact an importable file

It is not though. The file is a standalone file that is meant to be loaded via a browser using the `execScript` api. For that you pass it the file path and never import it. It's structured specifically in such a way that it doesn't work as an ESM file but instead runs an IIFE that returns the data to the extension.

```js
const internals = await chrome.scripting.executeScript({
  target: {
    tabId: tab.id
  },
  files: ['node_modules/axe-core/gather-internals.js'],
  world: 'MAIN'
});
```

Garbee:
Screenshot 2026-05-21 at 12 25 24 PM

The `node --check` is enough, since it validates the file can be imported and has correct syntax. It is not checking anything deeper.

The point is to validate what we ship can be utilized. Famous Linux mantra, "Once someone relies on it, it isn't a bug it is a feature." We have no export isolation right now, so everything is importable. Thus, we should validate that behavior until we can do a breaking change release and do export controls.

Garbee (approved):

WilcoFiers (approved):

PR #5118, opened 2026-05-26
feat(getRules): include rule enabled state in returned objects

### LINKED ISSUE #5116: Expose a rule's `enabled` property in the official API
### Product

axe-core

### Feature Description

To my knowledge, there's currently no official way to get the list of rules that are enabled when `axe.run()` is called with no options.

There's an [`axe.getRules()`](https://github.com/dequelabs/axe-core/blob/c4f60ffcd47eb64514e8cbafbc68ad357ce60e77/doc/API.md#api-name-axegetrules) method, but that returns _all_ rules, including the ones that are disabled by default.

There is an _undocumented_ way to get such a list today (`axe._audit.rules.filter(rule => rule.enabled)`).

It would be great to expose the ability to filter rules by whether they're `enabled` as part of the official API, for example by adding the `enabled` key to the rule objects returned by `axe.getRules()`, or by introducing a new method.

pomerantsev:
Adds an `enabled` boolean to each rule object returned by `axe.getRules()`, exposing whether the rule runs by default when `axe.run()` is called without options. Previously the only way to determine this was via the undocumented `axe._audit.rules` field.

Closes: #5116

### DIFF
--- a/axe.d.ts
+++ b/axe.d.ts
@@ -390,8 +390,10 @@ declare namespace axe {
     frameContext: FrameContextObject;
   }
 
-  interface RawCheckResult
-    extends Omit<CheckResult, 'relatedNodes' | 'impact'> {
+  interface RawCheckResult extends Omit<
+    CheckResult,
+    'relatedNodes' | 'impact'
+  > {
     relatedNodes?: Array<SerialDqElement | DqElement>;
     impact?: ImpactValue;
   }
--- a/axe.d.ts
+++ b/axe.d.ts
@@ -334,7 +334,7 @@ declare namespace axe {
     matches?: string | ((node: Element, virtualNode: VirtualNode) => boolean);
     reviewOnFail?: boolean;
     actIds?: string[];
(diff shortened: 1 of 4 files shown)


pomerantsev:
I've made two contributions previously, not sure why the CLA bot is complaining again 🤷‍♂️

straker:
@pomerantsev prettier is failing, do you mind running `npx prettier axe.d.ts doc/API.md --write` and pushing the changes. That should get the build to pass again.

WilcoFiers (approved):
Fantastic! We're planning a 4.12 release next week. Expect to see this in it.

WilcoFiers:
For the books. I reviewed this PR for security.

WilcoFiers:
@pomerantsev Would you mind signing the CLA again? This might have been updated since you last signed it. We need to feature freeze for 4.12 today and can't merge this in until the CLA is signed.

pomerantsev:
@WilcoFiers thanks! I just signed. Feel free to merge if this can still make it into 4.12. Not a big deal if not.

straker (approved):
Thanks for the update.

Reviewed for security.

PR #5119, opened 2026-05-26
feat(list,listitem): support element internals role

straker:
Noticed while testing out some things that we could also support element internals for `list` and `listitem` rules. Just needed to have them look at the semantic role as well.


WilcoFiers (approved):

Garbee (approved):
Looks reasonable to me.

chutchins25 (approved):
Looks good to me.

PR #5121, opened 2026-05-28
chore: migrate Karma to Web Test Runner

### LINKED ISSUE #5115: Nightly tests are failing; Move off Karma
For the past few weeks our nightly firefox tests have been failing. At first glance which tests are failing is flakey and not consistent, but the cause is always a test timeout. I dug into what was going on and from my findings I discovered that running karma and firefox is the cause. 

For some reason, the karma integration is adding a 2-3 second delay to our tests. I have logged performance metrics everywhere in our code and no where causes a 2 second delay. Furthermore, opening the tests in an actual Firefox browser using `npm run test:debug` removes the 2 second delay. Additionally, when I tested moving to a different test runner such as playwright and Web Test Runner in Firefox the 2 second delay was gone as well.

The test failure problem is that Nightly Firefox has increased that delay to 4-6 seconds, which exceeds our test timeout. 

To fix this we're going to have to migrate off Karma. [Karma itself is deprecated](https://github.com/karma-runner/karma#karma-is-deprecated-and-is-not-accepting-new-features-or-general-bug-fixes) and recommends Web Test Runner to run tests in an actual browser (as opposed to jest which run in JSDOM). A quick test shows it could work for us. However, one problem I foresee is that WTR uses Playwright as the default runner, and [Playwright does not support Nightly Firefox](https://playwright.dev/docs/browsers#firefox). We'll probably need to use their Webdriverio/js runner instead just so we can run that.

straker:
Migrating Karma to use Web Test Runner (WTR). Out of the box it has some problems that needed to be addressed, and some things we'll need to get use to using a new test runner.

* WTR uses Playwright by default, but Playwright doesn't support Firefox Nightly. We'll use the Selenium driver in order to support that
* However the Selenium driver automatically runs each test inside an iframe, which we do not want (and also breaks our tests). The Selenium driver doesn't have an option to disable this, so had to hack a way around this to ensure each test ran in the root document and no iframe
* WTR runs each test file in an isolated Mocha context, so using `.only` no longer works across test files, just within the file it's used in. A bit annoying, but we can get use to this (tried to hack a way around this but it became way too complicated and broke a bunch of other stuff). `develop` still will watch files and run the changed test file for us
* WTR `--manual` is used to replace `test:debug`. This didn't have a UI at first, so added the `html` style mocha reporter that we are use to. The only difference now is that the browser doesn't open automatically (have to press 'D') and that the landing page is a list of links to each test file
* WTR doesn't have a preprocess step like Karma had, so we had to move it to it's own file and do the preprocessing for integration tests before we run them

Additionally:

* Noticed that our axe.log stubbing in various tests weren't working since we moved to ES6 modules (the code directly calls `log` and doesn't go through `axe.log`), this meant the asserts testing the messages weren't being run at all. So I fixed that by adding a `axe._setLogger` helper that would let us designate the logger stub. This allows us to both stub the logger or suppress it (so we don't have to see performance timer logs in tests, etc.)
* Puppeteer was [failing a lot recently](https://github.com/dequelabs/axe-core/actions/runs/26545263689/job/78398503218) so fixed that while I was here
* WTR can support both our current test loading (relying on global axe and properties) and es6 imports to the `lib` directory. I updated `performance-timer` to show this being possible. This means we should be able to remove the `_thisWillBeDeletedDoNotUse` property that was added just to get things to the test and we can import those directly in the tests that need them. However, this isn't something we can do right away as the imported file is a different context from the same file being imported in axe (since we bundle the axe files together). This means file scope/state is different so we can't import the new `setLogger` function directly in tests and use it as it doesn't set the axe.js bundled `logger` variable scope. 

Closes #5115


lib/core/log.js
```diff
@@ -1,13 +1,16 @@
...
  */
-function log() {
-  if (typeof console === 'object' && console.log) {
-    // IE does not support console.log.apply
-    Function.prototype.apply.call(console.log, console, arguments);
+export default function log(...args) {
+  if (logger) {
+    logger(...args);
+  } else if (typeof console === 'object' && console.log) {
+    console.log(...args);
```
straker, line 10: Who needs `Function.prototype.apply` when we can spread now? :D

test/core/base/audit.js
```diff
@@ -1248,7 +1248,9 @@ describe('Audit', () => {
         externalAPIs({
           getElementInternals: () => {
             elementInternalsStartTime = performance.now();
-            return Promise.resolve([]);
+            return new Promise(res => {
+              setTimeout(() => res([]), 100);
```
straker, line 1252: Firefox would resolve this almost instantly so the two times were equal. Adding a timeout to ensure we can track one starting after the other

test/gather-internals/index.js
```diff
@@ -328,9 +328,7 @@ describe('gather-internals.walkTree', () => {
     fixture.innerHTML = html`<form id="form">
       <gather-internals-element></gather-internals-element>
     </form>`;
-    const node = fixture.querySelector('gather-internals-element');
     const form = fixture.querySelector('#form');
-    node._internals.form = form;
```
straker, line 333: This would fail in strict mode. It's automatically set by the browser and is a read only property

Garbee:
Doing some re-runs to just check for flakiness. If they pass all clear, then this all looks good to me.

Garbee (dismissed):
A few re-runs of the tests and no failures. It looks good to me.

WilcoFiers (commented):
## Summary

Solid migration from Karma to Web Test Runner. The custom `DirectSeleniumLauncher` is a reasonable fix for WTR’s iframe wrapper, and the `axe._setLogger` / integration-test generation changes address real gaps exposed by the move. Garbee’s re-runs give me confidence in stability.

A few follow-ups would help contributors and reduce confusion—mostly docs and small cleanups, nothing that blocks merge from my side.

---

## Important

### 1. Update contributor docs for WTR

[`doc/developer-guide.md`](https://github.com/dequelabs/axe-core/blob/e4fee1ae54dad87c529b8b0fde5b1c61271ea862/doc/developer-guide.md#L46-L89), [`CONTRIBUTING.md`](https://github.com/dequelabs/axe-core/blob/e4fee1ae54dad87c529b8b0fde5b1c61271ea862/CONTRIBUTING.md#L199-L212), and [`test/integration/rules/README.md`](https://github.com/dequelabs/axe-core/blob/e4fee1ae54dad87c529b8b0fde5b1c61271ea862/test/integration/rules/README.md) still describe Karma and the old `testDirs` debug API.

Please update them to match this PR:

- WTR + `--manual` for `test:debug` (press `D`, file list landing page)
- `--files 'test/core/**/*.js'` instead of `testDirs=core`
- Rule integration tests need `npm run build:integration-tests` first
- WTR behavior: `.only` only within a single test file

Also worth replacing “Mocha / Karma” in the watch section with WTR.

---

### 2. Remove unused `@web/test-runner-selenium`

[`package.json`](https://github.com/dequelabs/axe-core/blob/e4fee1ae54dad87c529b8b0fde5b1c61271ea862/package.json#L133) lists `@web/test-runner-selenium`, but [`test/wtr.config.mjs`](https://github.com/dequelabs/axe-core/blob/e4fee1ae54dad87c529b8b0fde5b1c61271ea862/test/wtr.config.mjs) uses `selenium-webdriver` directly with `DirectSeleniumLauncher`. Consider dropping the unused dependency to avoid confusion about which Selenium integration is canonical.

---

### 3. Fix misleading comment in `wtr.debug.config.mjs`

[`test/wtr.debug.config.mjs`](https://github.com/dequelabs/axe-core/blob/e4fee1ae54dad87c529b8b0fde5b1c61271ea862/test/wtr.debug.config.mjs#L7-L9) says “The browser opens automatically,” but the PR description notes that with `--manual` you press `D` and land on a file list. Aligning the header comment with that flow would save future debug sessions.

---

## Suggestions

### 4. Update `Gruntfile` clean path for integration tests

[`Gruntfile.js`](https://github.com/dequelabs/axe-core/blob/e4fee1ae54dad87c529b8b0fde5b1c61271ea862/Gruntfile.js#L48) still cleans `tmp/integration-tests.js`, while generation now writes to `tmp/integration-tests/`. Updating `clean.tests` to the directory (or relying on the existing `tmp` ignore) keeps `grunt clean` accurate.

---

### 5. Document `axe._setLogger` as test-only

[`lib/core/log.js`](https://github.com/dequelabs/axe-core/blob/e4fee1ae54dad87c529b8b0fde5b1c61271ea862/lib/core/log.js) / [`lib/core/core.js`](https://github.com/dequelabs/axe-core/blob/e4fee1ae54dad87c529b8b0fde5b1c61271ea862/lib/core/core.js#L44) — a short `@private` / “for tests only” JSDoc on `setLogger` and `axe._setLogger` would clarify this isn’t supported public API (no `axe.d.ts` change needed if it stays internal).

---

### 6. Consider surfacing errors in the Selenium session chain

[`test/wtr.config.mjs`](https://github.com/dequelabs/axe-core/blob/e4fee1ae54dad87c529b8b0fde5b1c61271ea862/test/wtr.config.mjs#L93):

```js
this._chain = navigate.then(() => sessionDone).catch(() => {});
```

Swallowing navigation failures may make Selenium/WTR issues harder to diagnose. Either rethrow/log, or add a one-line comment explaining why the chain must continue after failure.

WilcoFiers:
That was Cursor. I checked all of those, and agree. Nothing for myself to add.

Garbee:
I don't disagree with most of what Wilco's AI review pointed out. However, I would say we just ignore the grunt-related bit since we're already talking about removing that? It's basically a race condition of which lands first and how to handle it. But once one lands the other has to update and account for things anyways. So, seems moot to spend cycles on it. (Although, it seems like it could be a one string change. In which case, just do it.)

WilcoFiers (approved):

lib/core/log.js
Garbee: If there is ever the concern of potentially exceeding the maximum call stack size...

PR #5129, opened 2026-06-01
fix(axe.d.ts): make enabled property of RuleMetadata optional

straker:
This was an accidental breaking change if anyone uses `RuleMetadata` directly.


typings/axe-core/axe-core-tests.ts
```diff
@@ -512,3 +512,13 @@ axe.commons.dom.isFocusable(document.body);
 axe.commons.dom.isNativelyFocusable(document.body);
 axe.commons.dom.getNodeGrid(document.body);
 axe.commons.text.accessibleText(document.body);
+
+// Types
+let metadata: axe.RuleMetadata;
```
Garbee, line 522:
```suggestion
/**
* Validate the original structure for metadata persists in support.
* Do not amend this structure with new properties. As all new properteis
* should be optional as to not break existing implementations.
*/
let metadata: axe.RuleMetadata;
```

Just leave a clear note as to why we test this type structure.

axe.d.ts
```diff
@@ -352,7 +352,7 @@ declare namespace axe {
     helpUrl: string;
     tags: string[];
     actIds?: string[];
-    enabled: boolean;
+    enabled?: boolean;
```
WilcoFiers, line 355: The return type of `getRules()` is now wrong. While this is optional in the config, it'll always be returned.

On top of that, there's a bug in that experimental rules are returned as enabled. We need to update the logic of getRules()[0].enabled

WilcoFiers (dismissed):

axe.d.ts
```diff
@@ -623,7 +623,7 @@ declare namespace axe {
    * @param  {Array}  tags  Optional array of tags
    * @return {Array}  Array of rules
    */
-  function getRules(tags?: string[]): RuleMetadata[];
+  function getRules(tags?: string[]): Required<RuleMetadata>[];
```
chutchins25: `Required<RuleMetadata>[]` also forces `actIds` to be non-optional, but `getRules()` returns `actIds: undefined` for the ~49 rules with no `actIds` (`Rule` sets `this.actIds = spec.actIds`). So `getRules()[0].actIds.length` would typecheck and then throw at runtime — swapping one type bug for another. Could we make only `enabled` required and leave `actIds` optional?

```suggestion
  function getRules(
    tags?: string[]
  ): (Omit<RuleMetadata, 'enabled'> & { enabled: boolean })[];
```

lib/core/public/get-rules.js
```diff
@@ -5,15 +5,17 @@
  */
 function getRules(tags) {
   tags = tags || [];
+  const { rules, data, tagExclude } = this._audit;
```
chutchins25: Was the switch from `axe._audit` to `this._audit` intended? Every other method in `core/public/` reads the global `axe._audit`, and since `axe.getRules` is an unbound assignment, a destructured call (`const { getRules } = axe; getRules()`) would now throw on `this` being undefined — which worked before. Given this PR is about avoiding accidental breakage, matching the existing pattern seems safer:

```suggestion
  const { rules, data, tagExclude } = axe._audit;
```

chutchins25 (commented):
Nice work — the `RuleMetadata.enabled` revert is the right call. A couple of questions on the `getRules()` re-typing and audit access; neither is a hard blocker.

**PR checklist (repo conventions):**
- ✅ **Unit tests** — added `experimental` / `deprecated` cases in `test/core/public/get-rules.js`
- ✅ **Type tests** — added `RuleMetadata` structure assertion in `typings/axe-core/axe-core-tests.ts`
- ✅ **Commit format** — `fix(axe.d.ts): …` conforms to Angular convention
- ➖ **i18n / `locales/_template.json`** — N/A, no message changes
- ➖ **Integration / Shadow DOM tests** — N/A, public-API/type change, not a rule
- ⚠️ **Type-test coverage gap** — the new `Required<>` return type isn't exercised against a rule *without* `actIds` (see inline); a `.actIds` assertion would lock the contract
- 🔸 **`doc/API.md`** (optional) — `getRules()` now reports `enabled: false` for default-excluded tags; a one-line note there would document the behavior

lib/core/public/get-rules.js
straker: Totally should be `axe._audit`.

axe.d.ts
WilcoFiers: Jup, I spotted that too.

WilcoFiers (dismissed):
See @chutchins25 's comment

WilcoFiers (approved):

pomerantsev:
Thanks all for both the runtime and type-level fixes, and sorry about the regression.

PR #5135, opened 2026-06-08
fix(aria/get-aria-value): allow getting role value

straker:
While reviewing code in preparation for https://github.com/dequelabs/axe-core/issues/5044, I realized that we'll need to be able to look at the `role` value (and not use `getExplicitRole` like I was thinking). So fixed the code in order to allow that. Alternatively, if we wanted to not have to special case `role` we could add `role` into the `ariaAttrs` standard since we'll also need to do this for the [`hasAttrValue` function](https://github.com/dequelabs/axe-core/pull/5136) as well.

### DIFF
--- a/lib/commons/aria/get-aria-value.js
+++ b/lib/commons/aria/get-aria-value.js
@@ -33,7 +33,7 @@ export default function getAriaValue(node, attrName, options = {}) {
   const attrStandard = standards.ariaAttrs[attrName]
     ? standards.ariaAttrs[attrName]
     : attrName === 'role'
-      ? { type: 'string', prop: 'roll' }
+      ? { type: 'string', prop: 'role' }
       : null;
   if (!attrStandard) {
     return null;
--- a/lib/commons/aria/get-aria-value.js
+++ b/lib/commons/aria/get-aria-value.js
@@ -30,7 +30,11 @@ const sources = [
  * @return {{value: String, source: String}|null} value
  */
 export default function getAriaValue(node, attrName, options = {}) {
-  const attrStandard = standards.ariaAttrs[attrName];
+  const attrStandard = standards.ariaAttrs[attrName]
+    ? standards.ariaAttrs[attrName]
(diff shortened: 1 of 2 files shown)


lib/commons/aria/get-aria-value.js
```diff
@@ -30,7 +30,11 @@ const sources = [
  * @return {{value: String, source: String}|null} value
  */
 export default function getAriaValue(node, attrName, options = {}) {
-  const attrStandard = standards.ariaAttrs[attrName];
+  const attrStandard = standards.ariaAttrs[attrName]
+    ? standards.ariaAttrs[attrName]
+    : attrName === 'role'
+      ? { type: 'string', prop: 'role' }
+      : null;
```
chutchins25, line 37: **On the design question you raised:** since you've already flagged that #5136 (`hasAttrValue`) needs the same `role` handling, inlining `{ type: 'string', prop: 'role' }` here means two copies that can drift. A small shared resolver — e.g. `getAttrStandard(attrName)` that both call — keeps it in one place without overloading the `ariaAttrs` data with a non-aria entry (which reads oddly given the name). Adding `role` to `ariaAttrs` would also work and avoids the special-case entirely, but the helper feels cleaner semantically. Either way, deciding before #5136 lands avoids the duplication.

**Readability nit (independent of the above):** the nested ternary looks up `ariaAttrs[attrName]` twice; `??` collapses it:
```suggestion
  const attrStandard =
    standards.ariaAttrs[attrName] ??
    (attrName === 'role' ? { type: 'string', prop: 'role' } : null);
```

chutchins25 (commented):
Nice work — clean, well-targeted fix, and the two source-path tests (attribute + internals) cover it well. One inline note below: mostly a heads-up on the `role` special-case you flagged in the description (dedup vs #5136), plus a small readability tweak. Both non-blocking.

straker:
We decided to hold off on doing this until we're ready to start validating the role values

PR #5136, opened 2026-06-08
feat(aria/has-attr-value): new function to check if node has aria value

straker:
Companion to #5109 but replaces `vNode.hasAttr`.

This resulted in finding a bug in Firefox which crashes the browser when trying to access an empty array set to an idrefs property in element internals. This is why the Firefox tests are failing (technically crashing). Filled here https://bugzilla.mozilla.org/show_bug.cgi?id=2045887.

### DIFF
--- a/lib/commons/aria/has-aria-value.js
+++ b/lib/commons/aria/has-aria-value.js
@@ -11,13 +11,9 @@ import standards from '../../standards';
  * @return {Boolean}
  */
 export default function hasAriaValue(node, attrName) {
-  const attrStandard = standards.ariaAttrs[attrName]
-    ? standards.ariaAttrs[attrName]
-    : attrName === 'role'
-      ? { prop: 'role' }
-      : null;
+  const attrStandard = standards.ariaAttrs[attrName];
   if (!attrStandard) {
-    return false;
+    throw new TypeError(`Attribute ${attrName} is not an ARIA attribute`);
   }
 
   const { vNode } = nodeLookup(node);
--- a/lib/commons/aria/has-aria-value.js
+++ b/lib/commons/aria/has-aria-value.js
@@ -1,0 +1,49 @@
+import { nodeLookup } from '../../core/utils';
+import standards from '../../standards';
+
+/**
+ * Test if the element has an ARIA attribute, property, or ElementInternal key
+ * @method hasAriaValue
+ * @memberof axe.commons.aria
+ * @instance
+ * @param {Node|Element|VirtualNode} node
+ * @param {String} attrName The ARIA attribute name to get the value of. Use the ARIA attr name even if getting values from properties.
+ * @return {Boolean}
+ */
+export default function hasAriaValue(node, attrName) {
+  const attrStandard = standards.ariaAttrs[attrName]
+    ? standards.ariaAttrs[attrName]
+    : attrName === 'role'
+      ? { prop: 'role' }
+      : null;
+  if (!attrStandard) {
+    return false;
+  }
+
+  const { vNode } = nodeLookup(node);
+  return (
+    hasAttributeValue(vNode, attrName) ||
+    hasPropertyValue(vNode, attrStandard) ||
+    hasInternalValue(vNode, attrStandard)
+  );
+}
+
+function hasAttributeValue(vNode, attrName) {
+  return vNode.hasAttr(attrName);
+}
+
+function hasPropertyValue(vNode, attrStandard) {
+  const { prop } = attrStandard;
+  if (prop && vNode.actualNode) {
+    return vNode.actualNode[prop] !== null;
+  }
+  return false;
+}
+
+function hasInternalValue(vNode, attrStandard) {
+  const { prop } = attrStandard;
+  if (prop && vNode.elementInternals) {
+    return vNode.elementInternals[prop] !== null;
+  }
+  return false;
+}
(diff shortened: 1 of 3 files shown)


lib/commons/aria/has-aria-value.js
```diff
@@ -0,0 +1,49 @@
...
+}
+
+function hasAttributeValue(vNode, attrName) {
+  return vNode.hasAttr(attrName);
+}
+
+function hasPropertyValue(vNode, attrStandard) {
+  const { prop } = attrStandard;
+  if (prop && vNode.actualNode) {
+    return vNode.actualNode[prop] !== null;
```
chutchins25, line 34: Reflected ARIA props default to `null` when unset, but a prop the browser *doesn't support* returns `undefined`, and `undefined !== null` is `true`. So for the element-reference props (`ariaControlsElements`, `ariaActiveDescendantElement`, …) that aren't universally supported (notably Firefox), this reports a false-positive "has value" for an element that has no such attribute. Loose `!= null` treats `undefined` as absent while still keeping the empty-string-is-present behavior your tests assert. `hasInternalValue` (line 46) has the identical check. Worth noting the companion `getAriaValue` has the same `undefined` leak via its `=== null` loop check — probably worth fixing both together.
```suggestion
    return vNode.actualNode[prop] != null;
```

lib/commons/aria/has-aria-value.js
```diff
@@ -0,0 +1,49 @@
...
+  }
+  return false;
+}
+
+function hasInternalValue(vNode, attrStandard) {
+  const { prop } = attrStandard;
+  if (prop && vNode.elementInternals) {
+    return vNode.elementInternals[prop] !== null;
```
chutchins25, line 42: Same `undefined`-vs-`null` fix as line 38.
```suggestion
    return vNode.elementInternals[prop] != null;
```

lib/commons/aria/has-aria-value.js
```diff
@@ -0,0 +1,49 @@
...
+ * @param {Node|Element|VirtualNode} node
+ * @param {String} attrName The ARIA attribute name to get the value of. Use the ARIA attr name even if getting values from properties.
+ * @return {Boolean}
+ */
+export default function hasAriaValue(node, attrName) {
+  const attrStandard = standards.ariaAttrs[attrName]
+    ? standards.ariaAttrs[attrName]
+    : attrName === 'role'
+      ? { prop: 'role' }
+      : null;
```
chutchins25: This is the second copy of the `role` special-case from #5135, and the two have already diverged — `getAriaValue` uses `{ type: 'string', prop: 'role' }`, this one `{ prop: 'role' }`. That's exactly the drift the [dedup suggestion on #5135](https://github.com/dequelabs/axe-core/pull/5135#discussion_r3381143424) was about; a shared `getAttrStandard(attrName)` resolver used by both would collapse these into one source of truth. Separately, the nested ternary reads more cleanly with `??`:
```suggestion
  const attrStandard =
    standards.ariaAttrs[attrName] ?? (attrName === 'role' ? { prop: 'role' } : null);
```

test/commons/aria/has-aria-value.js
```diff
@@ -0,0 +1,106 @@
...
+
+  it('returns true if elementInternals idrefs is empty', () => {
+    const vNode = queryFixture(
+      html`<testutils-element id="target"></testutils-element>`
+    );
+    const node = vNode.actualNode;
+    node._internals.ariaLabelledByElements = [];
+
+    assert.isTrue(hasAriaValue(vNode, 'aria-labelledby'));
+  });
```
chutchins25: Per your PR description, this case crashes Firefox ([bugzilla 2045887](https://bugzilla.mozilla.org/show_bug.cgi?id=2045887)), so Firefox CI is red and the PR can't merge as-is. What's the plan — skip/guard this test on Firefox with the bugzilla link inline so CI goes green and the crash is documented, or hold until Mozilla ships a fix? Also worth confirming the runtime exposure: `hasInternalValue` reads `elementInternals[prop]` unconditionally, so if the *access* is what crashes Firefox, a rule calling `hasAriaValue` could crash on a real page with an empty idrefs internals array — note this test happens to short-circuit on the empty `''` attribute via `hasAttr` before reaching internals, so the crashing runtime path may not be the one exercised here.

chutchins25 (commented):
Nice work — `hasAriaValue` is a clean boolean companion to `getAriaValue`, and the test matrix (attribute / property / internals / empty cases / SerialVirtualNode) is thorough. A few items below, two worth resolving before merge:

1. `!== null` lets `undefined` through for browser-unsupported reflected props — a false-positive "has value". The companion `getAriaValue` has the same leak (inline).
2. Firefox is crashing per your note — let's pin down the merge plan and whether the runtime read needs a guard (inline on the test).
3. Confirms the `role` dedup concern from #5135 — the two role standards have already diverged (inline).

lib/commons/aria/has-aria-value.js
straker: Eslint prevents `!=` and requires `!==`. All modern browsers support all AOM properties and element internals properties, including Firefox.

test/commons/aria/has-aria-value.js
```diff
@@ -0,0 +1,112 @@
...
+
+    assert.isFalse(hasAriaValue(vNode, 'aria-sort'));
+  });
+
+  it('returns false for non-aria value', () => {
+    const vNode = queryFixture(
+      html`<div id="target" aria-label="hello"></div>`
+    );
+
+    assert.isFalse(hasAriaValue(vNode, 'id'));
```
WilcoFiers: I wonder if we should have this throw. If this happens we've done something wrong. I think I'd prefer this to throw so we can catch it in testing, but I'll leave the decision to you.

WilcoFiers (dismissed):
Left a suggestion. I'm okay with it either way.

WilcoFiers (approved):

chutchins25 (approved):

PR #5137, opened 2026-06-08
test: support with-aria-* attributes on testutils-element

straker:
In preparation for https://github.com/dequelabs/axe-core/issues/5044, this will allow us to declaratively set aria prop values on the `testutils-element` using `with-aria-*` attributes (mimicking the `with-role` attribute). It handles converting idref(s) attributes to their DOM nodes.

### DIFF
--- a/test/testutils.js
+++ b/test/testutils.js
@@ -38,6 +38,9 @@ var helpers;
 
   // create a custom element that all tests can use
   if (!customElements.get('testutils-element')) {
+    const withAriaRegex = /^with-(aria-.+)$/;
+    const idrefTypes = ['idref', 'idrefs'];
+
     customElements.define(
       'testutils-element',
       class TestutilsElement extends HTMLElement {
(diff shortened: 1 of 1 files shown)


chutchins25 (approved):
Looks good — clean implementation that mirrors the existing `with-role` pattern, and reading the attributes in the constructor works fine for the `innerHTML`/`queryFixture` usage since the element upgrades after its attributes are appended. Nice prep for #5044.

One thing worth keeping in mind for later (not blocking): the singular-vs-array branch keys off `type`, so it covers the attributes #5044 needs but not `aria-details`/`aria-errormessage`, which are `type: 'idref'` in the standards data yet reflect to the plural `*Elements` array props. If those come into scope, deriving arity from the `prop` suffix (`Elements` → array) would handle them uniformly.

This is what Claude kept reporting during my review. I feel like we have had discussions about this on and off the last several days. Will defer to you on what, if anything, we should change about it.

straker:
aria-details/aria-errormessage being `idref` in standards kinda a "bug" (but not really). They should be updated to `idrefs` as of WAI-ARIA 1.3, especially since their AOM properties are plural. I don't think that will cause us any issues for now though, and not for this pr.

WilcoFiers (approved):

PR #5138, opened 2026-06-08
feat(dom/idrefs): support ARIA idrefs properties and element internals

straker:
Update `idrefs` function to understand handling ARIA idrefs properties and element internals properties. This was the leftover work from #5109 to handle idrefs being a string (attribute) or array of nodes (property). This should seemlessly allow all function calls to `idrefs` to handle element internals.


straker:
Decided to make this it's own function after all in order to not cause any problems with current use of `idrefs`

Issue #5139, opened 2026-06-09
ElementInternals: accessing empty internals idrefs property crashes Firefox


straker:
[Reported to bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=2045887). Tracking this as an issue so when we decided to flip the element internals feature flag we can see if the bug is fixed or not and decided what to do then.

Garbee:
I'm working with Firefox to patch this from our side.

Garbee:
Presented patch: https://phabricator.services.mozilla.com/D305689

Garbee:
Patch landed, this will go out with FF 153 on the 21st next month.

straker:
I'd like to keep this open until we remove the [`it.skip` part of the code](https://github.com/dequelabs/axe-core/blob/develop/test/commons/aria/has-aria-value.js#L40-L54) we had to add to handle this case

PR #5151, opened 2026-06-09
feat(dom/getResolvedRefs): new function to get the resolved virtual nodes of idrefs

straker:
A new function that is very similar to `commons.dom.idrefs` function, but understands handling ARIA idrefs properties and element internals properties and additionally returns virtual nodes rather than DOM nodes. This was the leftover work from #5109 to handle idrefs being a string (attribute) or array of nodes (property). 

I also update the `queryShadowFixture` function to allow both `#shadow` and `#target` ids to exist in the light DOM. I looked at all uses of `queryShadowFixture` and almost all of them use the `#shadow` in the light DOM and `#target` in the shadow DOM, though a few rely on `#target` in the light DOM (but no `#shadow`). The changes prioritize the order of find things so that `#shadow` is used first as the container, then `#target` (instead of the other way around) and the target selector is found in the shadow DOM first, then in the light DOM after (instead of the other way around). Lastly to support changing either selector I allowed the 3rd parameter to be an object that could define them (no current uses of `queryShadowFixture` pass a 3rd parameter, but I left the option to just pass the target selector as a string).

### DIFF
--- a/lib/commons/dom/get-resolved-refs.js
+++ b/lib/commons/dom/get-resolved-refs.js
@@ -30,8 +30,9 @@ export default function getResolvedRefs(node, attr, options = {}) {
       return [];
     }
 
-    // already resolved node refs
     let refs;
+
+    // already resolved node refs
     if (Array.isArray(attrValue) || attrValue instanceof window.NodeList) {
       refs = Array.from(attrValue);
     } else {
--- a/lib/commons/dom/get-resolved-refs.js
+++ b/lib/commons/dom/get-resolved-refs.js
@@ -1,0 +1,76 @@
+import getRootNode from './get-root-node';
+import { tokenList, nodeLookup, getNodeFromTree } from '../../core/utils';
+import standards from '../../standards';
+
+/**
+ * Get elements referenced via a space-separated token attribute or AOM property;
+ * it will insert `null` for any Element that is not found
+ * @method getResolvedRefs
+ * @memberof axe.commons.dom
+ * @instance
+ * @param  {HTMLElement|Node|VirtualNode} node
+ * @param  {String} attr The attribute name to get the value of. Use the ARIA attr name even if getting values from AOM properties.
+ * @return {Array<VirtualNodes|null>} Array of mixed values: Virtual Nodes or `null` if not found
+ *
+ * NOTE: When in a shadow DOM environment: ID refs (even for slotted content)
+ * refer to the document in which the element is considered to be in the
+ * "light DOM". Therefore, we use getElementById on the root node and not QSA
+ * on the flattened tree to dereference idrefs.
+ *
+ */
+export default function getResolvedRefs(node, attr) {
+  const { vNode, domNode } = nodeLookup(node);
+
+  try {
+    const attrValue = getIdrefsValue(vNode, attr);
+    if (!attrValue) {
+      return [];
+    }
+
+    // already resolved node refs
+    if (Array.isArray(attrValue) || attrValue instanceof window.NodeList) {
+      return Array.from(attrValue).map(n => {
+        return getNodeFromTree(n);
+      });
+    }
+
+    const doc = getRootNode(domNode);
+    return tokenList(attrValue).map(value => {
+      const n = doc.getElementById(value);
+      if (!n) {
+        return null;
+      }
+
+      return getNodeFromTree(n);
+    });
+  } catch {
+    throw new TypeError('Cannot resolve id references for non-DOM nodes');
+  }
+}
+
+function getIdrefsValue(vNode, attr) {
+  const { prop } = standards.ariaAttrs[attr] ?? {};
+
+  // get prop value first as setting idrefs can result in empty attribute values
+  // e.g. el.ariaLabelledByElements = [label]; el.getAttribute('aria-labelledby') === ''
+  if (prop && vNode.actualNode) {
+    const propValue = vNode.actualNode[prop];
+    if (propValue !== null) {
+      return propValue;
+    }
+  }
+
+  const attrValue = vNode.attr(attr);
+  if (attrValue !== null) {
+    return attrValue;
+  }
+
+  if (prop && vNode.elementInternals) {
+    const internalsValue = vNode.elementInternals[prop];
+    if (internalsValue !== null) {
+      return internalsValue;
+    }
+  }
+
+  return null;
+}
--- a/lib/commons/dom/get-resolved-refs.js
+++ b/lib/commons/dom/get-resolved-refs.js
@@ -10,6 +10,8 @@ import standards from '../../standards';
  * @instance
  * @param  {HTMLElement|Node|VirtualNode} node
  * @param  {String} attr The attribute name to get the value of. Use the ARIA attr name even if getting values from AOM properties.
+ * @param {Object} [options]
+ * @param {Boolean} [options.self=true] If the resolved refs should include the passed in node if the attr references its own id
  * @return {Array<VirtualNodes|null>} Array of mixed values: Virtual Nodes or `null` if not found
  *
  * NOTE: When in a shadow DOM environment: ID refs (even for slotted content)
@@ -18,7 +20,8 @@ import standards from '../../standards';
  * on the flattened tree to dereference idrefs.
  *
  */
-export default function getResolvedRefs(node, attr) {
+export default function getResolvedRefs(node, attr, options = {}) {
+  const { self = true } = options;
   const { vNode, domNode } = nodeLookup(node);
 
   try {
@@ -29,30 +32,26 @@ export default function getResolvedRefs(node, attr) {
 
     // already resolved node refs
     if (Array.isArray(attrValue) || attrValue instanceof window.NodeList) {
-      return (
-        Array.from(attrValue)
-          // remove self references
-          .filter(n => n !== domNode)
-          .map(n => {
-            return getNodeFromTree(n);
-          })
-      );
+      let refs = Array.from(attrValue);
+      if (!self) {
+        refs = refs.filter(n => n !== domNode);
+      }
+      return refs.map(n => getNodeFromTree(n));
     }
 
     const doc = getRootNode(domNode);
-    return (
-      tokenList(attrValue)
-        // remove self references
-        .filter(value => value !== vNode.props.id)
-        .map(value => {
-          const n = doc.getElementById(value);
-          if (!n) {
-            return null;
-          }
+    let refs = tokenList(attrValue);
+    if (!self) {
+      refs = refs.filter(value => value !== vNode.props.id);
+    }
+    return refs.map(value => {
+      const n = doc.getElementById(value);
+      if (!n) {
+        return null;
+      }
 
-          return getNodeFromTree(n);
-        })
-    );
+      return getNodeFromTree(n);
+    });
   } catch (cause) {
     throw new TypeError('Cannot resolve id references for non-DOM nodes', {
       cause
--- a/lib/commons/dom/get-resolved-refs.js
+++ b/lib/commons/dom/get-resolved-refs.js
@@ -29,20 +29,30 @@ export default function getResolvedRefs(node, attr) {
 
     // already resolved node refs
     if (Array.isArray(attrValue) || attrValue instanceof window.NodeList) {
-      return Array.from(attrValue).map(n => {
-        return getNodeFromTree(n);
-      });
+      return (
+        Array.from(attrValue)
+          // remove self references
+          .filter(n => n !== domNode)
+          .map(n => {
+            return getNodeFromTree(n);
+          })
+      );
     }
 
     const doc = getRootNode(domNode);
-    return tokenList(attrValue).map(value => {
-      const n = doc.getElementById(value);
-      if (!n) {
-        return null;
-      }
+    return (
+      tokenList(attrValue)
+        // remove self references
+        .filter(value => value !== vNode.props.id)
+        .map(value => {
+          const n = doc.getElementById(value);
+          if (!n) {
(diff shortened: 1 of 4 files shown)


lib/commons/dom/get-resolved-refs.js
```diff
@@ -0,0 +1,76 @@
...
+}
+
+function getIdrefsValue(vNode, attr) {
+  const { prop } = standards.ariaAttrs[attr] ?? {};
+
+  // get prop value first as setting idrefs can result in empty attribute values
+  // e.g. el.ariaLabelledByElements = [label]; el.getAttribute('aria-labelledby') === ''
+  if (prop && vNode.actualNode) {
+    const propValue = vNode.actualNode[prop];
+    if (propValue !== null) {
```
chutchins25: Same `!== null`/`undefined` pattern as #5136 (https://github.com/dequelabs/axe-core/pull/5136#discussion_r3381396423). For an element-reference prop the browser doesn't implement (`ariaControlsElements`/`ariaLabelledByElements` on regular elements in Firefox/Safari, or newer props), `actualNode[prop]` is `undefined`, and `undefined !== null` is `true` — so it returns `undefined`, and `getResolvedRefs` then hits `if (!attrValue) return []`. Result: a plain `<button aria-controls="menu">` (attribute, not set via IDL) resolves to `[]` instead of `[menu]` on those browsers. The tests don't catch it — they use a non-standard attr (`aria-cats`, no prop) or explicitly-set props (which become expandos when unsupported). Loose `!= null` lets `undefined` fall through to the attribute check, which is exactly what the prop-first ordering's own rationale wants (the empty-attribute case only arises when the prop *is* supported and set via IDL):
```suggestion
    if (propValue != null) {
```

test/testutils.js
```diff
@@ -218,33 +218,36 @@ var helpers;
...
+    shadowContent,
+    targetSelector = '#target'
+  ) => {
     // Normalize target, allow it to be the provided string or use '#target' to query composed tree
-    if (typeof targetSelector !== 'string') {
-      targetSelector = '#target';
+    let shadowSelector = '#shadow';
+    if (typeof targetSelector === 'object') {
+      targetSelector = targetSelector.target;
+      shadowSelector = targetSelector.shadow;
```
chutchins25: Order-of-operations bug in the new object-param branch: after `targetSelector = targetSelector.target`, the next line reads `targetSelector.shadow` off the *target string*, not the original object — so `shadowSelector` becomes `undefined` and the custom shadow selector is silently ignored. No current caller passes a 3rd param, but it's the feature this PR adds. (`typeof null === 'object'` would also throw on `null.target`.) Read `shadow` before reassigning, with defaults for partial objects:
```suggestion
    if (typeof targetSelector === 'object' && targetSelector !== null) {
      shadowSelector = targetSelector.shadow ?? '#shadow';
      targetSelector = targetSelector.target ?? '#target';
    }
```
Small heads-up too: with the new `targetSelector = '#target'` default, the later `if (!targetSelector && !targetCandidate) throw` is now near-unreachable — not worth changing, just noting the guard is mostly dead.

chutchins25 (commented):
Nice work — clean addition, the shadow-DOM scoping tests are a good touch, and the `content instanceof Node` → `shadowContent instanceof Node` typo fix is a nice catch along the way. Two suggestions below, both bugs with simple fixes: one cross-browser correctness issue in the lib, one in the new `queryShadowFixture` object-param path.

lib/commons/dom/get-resolved-refs.js
straker: This is incorrect. Both Safari and Firefox support all ARIA props and internals, and resolves any idrefs props properly.

```html
<button aria-controls="menu"></button>
<div id="menu"></div>
```

Returns `null` if the attribute is unset and `[<div id="menu"></div>]` for the `el.ariaControlsElements` in both Safari and Firefox.

lib/commons/dom/get-resolved-refs.js
```diff
@@ -0,0 +1,76 @@
...
+    return tokenList(attrValue).map(value => {
+      const n = doc.getElementById(value);
+      if (!n) {
+        return null;
+      }
+
+      return getNodeFromTree(n);
+    });
+  } catch {
+    throw new TypeError('Cannot resolve id references for non-DOM nodes');
```
WilcoFiers: For easier debugging later.

```suggestion
  } catch (cause) {
    throw new TypeError('Cannot resolve id references for non-DOM nodes', { cause });
```

lib/commons/dom/get-resolved-refs.js
```diff
@@ -0,0 +1,76 @@
...
+    const propValue = vNode.actualNode[prop];
+    if (propValue !== null) {
+      return propValue;
+    }
+  }
+
+  const attrValue = vNode.attr(attr);
+  if (attrValue !== null) {
+    return attrValue;
+  }
```
WilcoFiers, line 73: Should this be in an else block? Is there ever a case where we want to run this if we checked the prop?


```suggestion
  else {
    const attrValue = vNode.attr(attr);
    if (attrValue !== null) {
      return attrValue;
    }
  }
```

test/commons/dom/get-resolved-refs.js
```diff
@@ -0,0 +1,175 @@
+describe('dom.getResolvedRefs', () => {
+  const { html, queryFixture, queryShadowFixture, fixture } = axe.testUtils;
+  const getResolvedRefs = axe.commons.dom.getResolvedRefs;
+  const getNodeFromTree = axe.utils.getNodeFromTree;
+
+  it('should find referenced nodes by ID', () => {
+    const vNode = queryFixture(html`
+      <div aria-cats="target1 target2" id="target"></div>
```
WilcoFiers, line 8: If ARIA WG ever decides `aria-cats` needs to be an ARIA property we have a problem 🤣

lib/commons/dom/get-resolved-refs.js
```diff
@@ -0,0 +1,76 @@
...
+ * @param  {String} attr The attribute name to get the value of. Use the ARIA attr name even if getting values from AOM properties.
+ * @return {Array<VirtualNodes|null>} Array of mixed values: Virtual Nodes or `null` if not found
+ *
+ * NOTE: When in a shadow DOM environment: ID refs (even for slotted content)
+ * refer to the document in which the element is considered to be in the
+ * "light DOM". Therefore, we use getElementById on the root node and not QSA
+ * on the flattened tree to dereference idrefs.
+ *
+ */
+export default function getResolvedRefs(node, attr) {
```
WilcoFiers: For discussion: Should this filter duplicates and self-references?

test/testutils.js
```diff
@@ -218,33 +218,36 @@ var helpers;
    * @param String				Target selector for the check, can be inside or outside of Shadow DOM (optional, default: '#target')
    * @return VirtualNode
    */
-  testUtils.queryShadowFixture = (content, shadowContent, targetSelector) => {
+  testUtils.queryShadowFixture = (
+    content,
+    shadowContent,
+    targetSelector = '#target'
+  ) => {
```
WilcoFiers, line 225: If we're going to tweak this I prefer we start using declarative shadow DOM. I did that in advanced rules. It requires a bit of setup, but is much easier to read than what we've done here with queryShadowFixture. 

Not a blocker.

WilcoFiers (commented):
Some comments. Think we need to at least consider deduplication

lib/commons/dom/get-resolved-refs.js
straker: So testing the following html and here's the results

```html
<div id="label1">hello</div>
<div id="label2">dear</div>
<div id="label3">world</div>

<h2>Labelledby</h2>
<h3>Standard</h3>
<button aria-labelledby="label1 label2 label3"></button>

<h3>Duplicate</h3>
<button aria-labelledby="label1 label1 label2 label3"></button>

<h3>Self</h3>
<button id="self1" aria-labelledby="label1 self1 label2 label3"></button>

<h2>Owns</h2>
<h3>Standard</h3>
<button aria-owns="label1 label2 label3"></button>

<h3>Duplicate</h3>
<button aria-owns="label1 label1 label2 label3"></button>

<h3>Self</h3>
<button id="self2" aria-owns="label1 self2 label2 label3"></button>

<h2>Owns Unique</h2>
<div id="label4">hello</div>
<div id="label5">dear</div>
<div id="label6">world</div>
<h3>Standard</h3>
<button aria-owns="label4"></button>

<h3>Duplicate</h3>
<button aria-owns="label5 label5"></button>

<h3>Self</h3>
<button id="self3" aria-owns="label6 self3"></button>
```

__NVDA/Firefox__

* __labelledby->standard__: "hello dear world button"
* __labelledby->duplicate__: "hello dear world button"
* __labelledby->self__: "hello dear world button button"
* __owns->standard__: "hello dear world button"
* __owns->duplicate__: "button" 
* __owns->self__: "hello dear world button button" 
* __owns unique->standard__: "hello button"
* __owns unique->duplicate__: "dear button"
* __owns unique->self__: "world button"

__JAWS/Chrome__

* __labelledby->standard__: "hello dear world button"
* __labelledby->duplicate__: "hello hello dear world button"
* __labelledby->self__: "hello dear world button"
* __owns->standard__: "unlabeled 10 button"
* __owns->duplicate__: "unlabeled 12 button"
* __owns->self__: "hello dear world button"
* __owns unique->standard__: "hello button"
* __owns unique->duplicate__: "dear button"
* __owns unique->self__: "world button"

__VoiceOver/Safari__

* __labelledby->standard__: "hello dear world button"
* __labelledby->duplicate__: "hello dear world button"
* __labelledby->self__: "hello dear world button"
* __owns->standard__: "button"
* __owns->duplicate__: "button"
* __owns->self__: "button"
* __owns unique->standard__: "button"
* __owns unique->duplicate__: "button"
* __owns unique->self__: "button"

For the Owns cases of JAWs having unlabeled buttons, it looks like the accessibility tree thinks they have no children, and only the last "self" button has the children. So aria-owns stealing applies and the last references steal all the children from previous references.

straker: From the testing, it looks like we should be able to remove self references, but for the duplicate case it's a bit tricker. In JAWS/Chrome for `aria-labelledby` and `aria-owns` behave differently: `aria-labelledby` uses the duplication in the name but `aria-owns` does not duplicate the child for the name. In NVDA/Firefox the duplication is ignored in both cases.

lib/commons/dom/get-resolved-refs.js
straker: Maybe? With the current code, if we check the prop and it's `null` we fallback to the attribute, which we would assume would also be `null`. Is there ever a chance that it wouldn't be `null` if the prop is?

straker: I just remembered that we still support IE11 so we have to fallback to the attribute case as the prop won't exist.

lib/commons/dom/get-resolved-refs.js
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    // already resolved node refs
+    if (Array.isArray(attrValue) || attrValue instanceof window.NodeList) {
+      return (
+        Array.from(attrValue)
+          // remove self references
+          .filter(n => n !== domNode)
+          .map(n => {
+            return getNodeFromTree(n);
```
chutchins25: The two paths return different "not found" values: the string path below returns `null` for an unresolved id, but this array path maps through `getNodeFromTree` with no guard, so it yields `undefined` for any element not in the flat tree (disconnected, or outside the tested context). That contradicts the JSDoc (`Array<VirtualNode | null>`), and a consumer filtering on `!== null` would keep the `undefined` entries.

Also, the proposal's "Should handle #5" calls for filtering out unattached elements (and the #5139 Firefox empty-array case) — this array path doesn't. Suggest normalizing both branches to the documented contract (map out-of-tree/disconnected → `null`, or filter them consistently).

chutchins25 (commented):
Re-review — the earlier `queryShadowFixture` order-of-operations and `instanceof` issues look fixed, and the `!== undefined` guard is correct (needed for IE11). Two things left: a question on the self-reference filtering (replied on that thread), and the `null`-vs-`undefined` return inconsistency flagged inline.

lib/commons/dom/get-resolved-refs.js
chutchins25: Following up on the self-reference question: the AT testing above used empty `<button>`s, which masks one case. For `aria-labelledby`/`aria-describedby`, a self-reference is a valid accname pattern — the element contributes its own name-from-content (recursion is prevented by the accname "already-visited" guard, not by dropping the ref). So for `<button id="x" aria-labelledby="x label">Save</button>`, the name should include "Save", but filtering the self-ref here would drop it once an accname consumer (e.g. `arialabelledby-text`) adopts `getResolvedRefs`.

Worth confirming whether self-ref removal was validated against self-referencing elements that have their own text content. If it's only safe for some attributes (e.g. `aria-owns`/`aria-activedescendant`), consider scoping the filter to those rather than applying it to every reference attribute in a generic resolver — or leaving self/dup handling to the consumer.

test/testutils.js
straker: Created https://github.com/dequelabs/axe-core/pull/5158

lib/commons/dom/get-resolved-refs.js
straker: Good catch

chutchins25 (approved):
Everything seems resolved and taken care of now. Thanks for that. Looks good to me. Leaving Security Review for Wilco when he applies his review.

WilcoFiers (approved):

PR #5157, opened 2026-06-10
fix(standards): update aria-errormessage and aria-details to be idrefs

straker:
This is in the current [WAI-ARIA 1.3 draft](https://www.w3.org/TR/wai-aria-1.3/#aria-errormessage) and both attributes were changed to `idrefs` types. Since the AOM property is an idrefs type (returns an array) this allows us to be in sync with that for the Element Internals work. 

The `aria-errormessage` check [already accounts the attribute being an idrefs value](https://github.com/dequelabs/axe-core/pull/4973) and will incomplete if the value has more than one id, so no changes should be needed with this pr.


chutchins25 (commented):
Nice — clean alignment with the ARIA 1.3 idrefs change, and the `aria-errormessage` analysis checks out: the `aria-valid-attr-value` check skips it via `skipAttrs`, and the dedicated `aria-errormessage` check handles the multi-id case independently of the standards `type`. One question inline about test coverage on the `aria-details` side.

lib/standards/aria-attrs.js
```diff
@@ -78,7 +78,7 @@ const ariaAttrs = {
     global: true
   },
   'aria-details': {
-    type: 'idref',
+    type: 'idrefs',
```
chutchins25, line 87: Is there a test case for the `aria-details` side of this? Unlike `aria-errormessage` (skipped in `aria-valid-attr-value` + covered by its own check), `aria-details` flips behavior here — `aria-details="a b"` was a violation before and now passes if one id resolves — and I didn't see a multi-id case covering it.

WilcoFiers (dismissed):
Agreed. I wouldn't mind seeing a test here. Also, can you test these actually work? I think we should still make this change if it doesn't, but we may want to consider opening an issue to add a check with a dedicated issue message for it if people use multiple IDs in aria-errormessage when it doesn't work.

straker:
> add a check with a dedicated issue message for it if people use multiple IDs in aria-errormessage

We already have that. It was added in the pr I linked above.

As for adding a test for `aria-details`, we currently have no tests that test its value. Should we add one to something like `aria-valid-attr-value`?

WilcoFiers:
> We already have that. It was added in the pr I linked above.

Didn't realize that. Fair enough!

> As for adding a test for aria-details, we currently have no tests that test its value. Should we add one to something like aria-valid-attr-value?

Yes I think that'd be good.

chutchins25 (approved):
Re-reviewed on `c8c7a7e8`. My earlier question is resolved — the new `#pass194/195/196` cases cover the `aria-details` side, and they mirror the existing `aria-controls`/`aria-describedby` triples directly above them, so they slot into the established pattern.

I also traced every consumer of the standards `type` field to confirm the blast radius is what you describe: `validateAttrValue` (`validate-attr-value.js:57`) is the only place the change is observable, `getAriaValue`'s `idrefTypes` already listed both, and `getAccessibleRefs` matches on `/^idrefs?$/` — which means it was *already* tokenizing `aria-details`, so this removes a pre-existing internal inconsistency too. `aria-errormessage` stays in `skipAttrs` and its dedicated check hand-rolls `tokenList` + `getResolvedRefs`, so that half is inert today, as you said.

One thing worth surfacing for the next reviewer rather than a change request: after this change `aria-errormessage="a b"` fails with "Multiple IDs in aria-errormessage is not widely supported in assistive technologies", while `aria-details="a b"` now passes as long as one id resolves. If that asymmetry is deliberate — AT support genuinely differing between the two — no objection from me; just flagging that it's the half of @WilcoFiers' earlier question that isn't covered by the #4973 check.

WilcoFiers (approved):

PR #5158, opened 2026-06-10
test: allow declarative shadow dom

straker:
This allows `queryFixture` to now support declarative shadow DOM, replacing the need for `queryShadowFixture` and `createNestedShadowDom`. I wanted the detection to be conditional so we didn't need to do the expensive calls of `setHTMLUnsafe` and querying for every shadow on every test. So the code only does the expensive operations when it detects a declarative shadow DOM. Alternatively if we don't like the regex we could just have the `options.shadow` be passed into the `queryFixture`, but that would be a bit odd when the 2nd param is the query, but would be fine for the few times we actually use the functions: `queryFixture('<div>', '#target', { shadow: true });`

### DIFF
--- a/test/testutils.js
+++ b/test/testutils.js
@@ -576,7 +576,9 @@ const declarativeShadowDOMRegex =
     let ret = root.shadowRoot ? [root.shadowRoot] : [];
     ret.push(
       ...[...root.querySelectorAll('*')].flatMap(element => {
-        if (!element.shadowRoot) {return [];}
+        if (!element.shadowRoot) {
+          return [];
+        }
 
         let rets = [element.shadowRoot];
         if (recursive) {
--- a/test/testutils.js
+++ b/test/testutils.js
@@ -189,6 +189,8 @@ const declarativeShadowDOMRegex =
   /**
    * Method for injecting content into a fixture
    * @param {String|Node} content Stuff to go into the fixture (html or DOM node)
+   * @param {Object} [options]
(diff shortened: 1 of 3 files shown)


test/testutils.js
```diff
@@ -187,13 +191,18 @@ var helpers;
...
+  testUtils.injectIntoFixture = (content, options = {}) => {
     if (typeof content !== 'undefined') {
       fixture.innerHTML = '';
     }
 
     if (typeof content === 'string') {
-      fixture.innerHTML = content;
+      if (options.shadow) {
+        // allow declarative shadow DOM which requires using setUnsafeHTML to parse
+        fixture.setHTMLUnsafe(content);
```
WilcoFiers, line 204: Huh. I didn't know you could do that!

test/testutils.js
```diff
@@ -524,15 +533,63 @@ var helpers;
...
+        rootNode.actualNode,
+        ...getShadowRootsUnder(rootNode.actualNode)
+      ].reverse();
+      for (const root of roots) {
+        vNode = axe.utils.getNodeFromTree(root.querySelector(query));
+        if (vNode) {
+          break;
+        }
+      }
+    }
```
WilcoFiers: We should do this in checkSetup as well

WilcoFiers (dismissed):
The checkSetup thing is missing. I think it'd be nice to convert at least a couple tests to use this just to demonstrate it works. I had to do that locally now to learn that.

WilcoFiers (approved):

PR #5159, opened 2026-06-10
chore: update deprecated and old deps

straker:
This started because `chromedriver` wasn't the latest version in the package-lock so every time I switched between the grunt removal branch and any other branch I had to reinstall chromedriver to the latest. It then spiraled to looking at all the npm warnings and updating what made sense.

* `@babel/plugin-proposal-object-rest-spread` was deprecated as it was merged into the spec and we can now use the actual package `@babel/plugin-transform-object-rest-spread`
* `standard-version` was deprecated and they recommend using `commit-and-tag-version`
* `glob` was prevented from updating to v11 as it removed node version <20 and we still test under node 18. But looking at the code now we only use glob when we build, which we [already restore a cache version](https://github.com/dequelabs/axe-core/blob/develop/.github/workflows/test.yml#L236-L237) of before we test our code using different node versions. so we should be able to update it now

The remaining warns can't be fixed at the moment.

* `git-semver-tags@5.0.1` and `git-raw-commits@3.0.0` are both used by latest `commit-and-tag-version`. [an issue exists about it](https://github.com/absolute-version/commit-and-tag-version/issues/295)
* `whatwg-encoding@2.0.0` is used by latest `http-server`. [an issue exists about it](https://github.com/http-party/http-server/issues/973)
* `uuid@8.3.2` is used by `node-notifier`, which hasn't been updated in several years. we could move to https://github.com/Aetherinox/node-toasted-notifier which is a fork of it, but I wasn't sure if we wanted to or not. Alternatively we could just drop the package, it's only used 1 time in our build to notify when axe-core build is completed (a nice to have but not necessary).
* `glob@10.5.0` is used by latest `mocha`, but the next major mocha version looks like it updates to the latest glob

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -44,9 +44,6 @@ updates:
       # @see https://github.com/dequelabs/axe-core/issues/4428
       - dependency-name: 'colorjs.io'
         versions: ['>0.4.3']
-      # Still need to support node 18 in our tests
-      - dependency-name: 'glob'
-        versions: ['>=11.0.0']
       # Use node 4 types for backward compatibility
       - dependency-name: '@types/node'
         versions: ['>=5.0.0']
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,2 @@
 registry=https://registry.npmjs.org
+@deque:registry=https://registry.npmjs.org
--- a/.babelrc
+++ b/.babelrc
@@ -2,7 +2,7 @@
   "presets": ["@babel/preset-env"],
   "plugins": [
     [
-      "@babel/plugin-proposal-object-rest-spread",
+      "@babel/plugin-transform-object-rest-spread",
       {
         "loose": true
       }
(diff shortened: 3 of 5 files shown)


.npmrc
```diff
@@ -1 +1,2 @@
-registry=https://registry.npmjs.org
\ No newline at end of file
+registry=https://registry.npmjs.org
+@deque:registry=https://registry.npmjs.org
```
straker, line 2: Installing made `@deque/dot` not use the public npm so needed to add this to make sure it didn't happen again.

Garbee: Yea, because our user configs all have `@deque` pointing to the private registry.

.github/dependabot.yml
```diff
@@ -44,9 +44,6 @@ updates:
       # @see https://github.com/dequelabs/axe-core/issues/4428
       - dependency-name: 'colorjs.io'
         versions: ['>0.4.3']
-      # Still need to support node 18 in our tests
```
Garbee, line 47: So, we are not supporting Node 18 (or potentially earlier) testing? What exactly is the impact here?

Or is older stuff only tested after being compiled which negates the impact of this dep?

straker: The latter. We cache the built files using current node, then restore the built files when we test older node versions so we don't have to build them.

WilcoFiers (approved):

PR #5161, opened 2026-06-11
fix(gather-internals): handle non-HTMLElement nodes

straker:
Use `Node` instead of `HTMLElement` when checking the element internals prop value so we include all node types, such as SVGs. This is the follow up to a [similar problem in the `get-aria-value` pr](https://github.com/dequelabs/axe-core/pull/5109#discussion_r3357886786). 

I also added an eslint rule to prevent us from using `HTMLElements` in the future so we don't make the same mistake again.

### DIFF
--- a/lib/gather-internals/walk-tree.js
+++ b/lib/gather-internals/walk-tree.js
@@ -37,7 +37,7 @@ export function walkTree(tree = document.body) {
 
           // convert idref to node ancestry
           // aria props can use unattached nodes but those won't be used by the browser to calculate the prop value
-          if (value instanceof globalThis.HTMLElement) {
+          if (value instanceof globalThis.Node) {
             internals[prop] = value.isConnected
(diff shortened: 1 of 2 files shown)


lib/gather-internals/walk-tree.js
```diff
@@ -37,7 +37,7 @@ export function walkTree(tree = document.body) {
 
           // convert idref to node ancestry
           // aria props can use unattached nodes but those won't be used by the browser to calculate the prop value
-          if (value instanceof globalThis.HTMLElement) {
+          if (value instanceof globalThis.Node) {
```
WilcoFiers, line 40: This may need to be Element? I tested it out, you can't assign text nodes to ariaLabelledbyElement. So that's good. In practice this shouldn't matter, so I think it's fine.

WilcoFiers (approved):

PR #5163, opened 2026-06-11
feat(commons/dom): support ARIA element internals properties

straker:
Updates the `commons/dom` directory to use the new `getAriaValue` or `hasAriaValue` functions where it makes sense. Also update any uses of `node.getAttribute` or `node.hasAttribute` where it made sense as well.

* `commons/dom/focus-disabled` - hasAttr(disabled), disabled (not aria-disabled) is not an ARIA prop so remains hasAttr
* `commons/dom/get-element-by-reference` - looks up non-aria attrs, should still use .attr, but do we need to split the logic to handle non-aria vs aria attributes differently?
* `commons/dom/tabbable-elements` - looks at tabindex, should still use .attr
* `commons/dom/has-lang-text` - looks at lang attr, should still use .attr
* `commons/dom/idrefs` - superceded by `get-resolved-refs` so won't update uses
* `Commons/dom/inserted-into-focus-order` - looks at tabindex, should still use .attr
* `commons/dom/is-current-page-link` - looks at href, should still use .attr (or getAttribute in this case since it has to deal with html anchor only properties)
* `commons/dom/is-focusable` - tabindex, use .attr
* `commons/dom/is-in-tab-order` - tabindex, use .attr
* `commons/dom/is-inert` - inert, use .hasAttr
* `commons/dom/is-natively-focusable` - href, use .hasAttr
* `commons/dom/is-visibile (deprecated)` - no conversion
* `commons/dom/url-props-from-attribute` - needs node (`ownerSVGElement` prop), so no conversion from getAttribute
* `commons/dom/visibility-methods` - only the aria-hidden use should move to getAriaValue, others are non-aria attrs

Refs: https://github.com/dequelabs/axe-core/issues/5142
Refs: https://github.com/dequelabs/axe-core/issues/5043

### DIFF
--- a/lib/commons/aria/get-aria-value.js
+++ b/lib/commons/aria/get-aria-value.js
@@ -45,12 +45,16 @@ export default function getAriaValue(node, attrName, options = {}) {
       continue;
     }
 
-    if ('Node' in window && value instanceof window.Node) {
+    if (typeof value === 'string') {
+      // string types (like aria-label) should not be trimmed
+      if (type !== 'string') {
+        value = value.trim();
+      }
+      value = lowercase ? value.toLowerCase() : value;
+    } else if (value instanceof window.Node) {
       value = value.nodeName;
-    } else if (
-      Array.isArray(value) ||
-      ('NodeList' in window && value instanceof window.NodeList)
-    ) {
+    } else {
+      // already resolved node refs
       value =
         '[' +
         Array.from(value)
@@ -59,15 +63,7 @@ export default function getAriaValue(node, attrName, options = {}) {
         ']';
     }
 
-    // string types (like aria-label) should not be trimmed
-    if (type !== 'string') {
-      value = value.trim();
-    }
-
-    return {
-      value: lowercase ? value.toLowerCase() : value,
-      source
-    };
+    return { value, source };
   }
 
   return null;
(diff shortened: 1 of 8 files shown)


lib/commons/aria/get-aria-value.js
```diff
@@ -45,9 +45,12 @@ export default function getAriaValue(node, attrName, options = {}) {
       continue;
     }
 
-    if (value instanceof window.Node) {
+    if ('Node' in window && value instanceof window.Node) {
```
straker: This [throws an errorr when running virtual rules test in node](https://github.com/dequelabs/axe-core/actions/runs/27421675714/job/81048886062?pr=5163) as `window.Node` does not exist. Need to guard here and in `get-resolved-refs` before using for `instanceof`

lib/commons/aria/get-aria-value.js
```diff
@@ -45,9 +45,16 @@ export default function getAriaValue(node, attrName, options = {}) {
...
     }
 
-    if (value instanceof window.Node) {
+    if (typeof value === 'string') {
+      // string types (like aria-label) should not be trimmed
+      if (type !== 'string') {
+        value = value.trim();
+      }
+      value = lowercase ? value.toLowerCase() : value;
+    } else if (value instanceof window.Node) {
```
straker, line 54: Looking at `window.Node` or `window.NodeList` throws an error when running tests in pure node environements (e.g. `npm run test:virtual-rules`). Swapped the logic around here and in `get-resolved-refs` so that the string case is first which is the only case virtual rules should hit.

WilcoFiers (approved):
Assuming you confirmed aria-hidden is case insensitive everywhere, LGTM.

PR #5165, opened 2026-06-11
feat(commons/forms): support ARIA element internals properties

straker:
Updates the `commons/forms` directory to use the new `getAriaValue` or `hasAriaValue` functions where it makes sense. Also updated the test files to use `queryFixture` instead of appending to fixture directly and trying to find the node after.

* `commons/forms/is-native-textbox` - looks at `type` attribute, keep .attr

Refs: https://github.com/dequelabs/axe-core/issues/5142

### DIFF
--- a/lib/commons/forms/is-disabled.js
+++ b/lib/commons/forms/is-disabled.js
@@ -1,3 +1,5 @@
+import getAriaValue from '../aria/get-aria-value';
+
 const disabledNodeNames = ['fieldset', 'button', 'select', 'input', 'textarea'];
 
 /**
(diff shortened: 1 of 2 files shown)


lib/commons/forms/is-disabled.js
```diff
@@ -19,12 +21,14 @@ function isDisabled(virtualNode) {
   }
 
   const { nodeName } = virtualNode.props;
-  const ariaDisabled = virtualNode.attr('aria-disabled');
+  const ariaDisabled = getAriaValue(virtualNode, 'aria-disabled', {
+    lowercase: true
```
WilcoFiers, line 25: Are we sure this isn't a bug? Lets test it to be sure. Firefox + NVDA is the one I'm expecting problems from here.

WilcoFiers (approved):
LGTM. Please confirm aria-disabled is case insensitive in Firefox before merging.

lib/commons/forms/is-disabled.js
straker:
```html
<button aria-disabled="true">hello</button>
<button aria-disabled="TrUE">hello</button>
```

All 3 ATs/browsers report both as unavailable/dimmed, so seems like it's case insensitive.

Issue #5144, opened 2026-06-09
ElementInternals: convert `lib/checks/aria` to `getAriaValue`/`hasAriaValue`


Garbee:
Part of #5044 — incremental, per-subdirectory conversion of ARIA-attribute reads to the new internals-aware resolvers.

### Scope: `lib/checks/aria` (the largest sub-issue — consider splitting into 2 PRs if a single diff gets unwieldy)
Representative files/usages:
- `aria-allowed-attr`, `aria-prohibited-attr`
- `aria-required-attr` (`aria-valuetext`, `aria-expanded`, variable `requiredAttr`)
- `aria-required-children` (`aria-busy`)
- `aria-conditional-attr` / `aria-conditional-row-attr` / `aria-conditional-checkbox-attr`
- `aria-errormessage` (`aria-errormessage`, `aria-invalid`)
- `aria-level`, `aria-busy`
- `has-global-aria-attribute`
- `braille-label-equivalent` (`aria-braillelabel`)
- `braille-roledescription-equivalent` (`aria-brailleroledescription`, `aria-roledescription`)

### Conversion pattern (shared by every #5044 sub-issue)
1. Replace ARIA-attribute reads — `vNode.attr('aria-*')` / `vNode.hasAttr('aria-*')`, incl. variable-attr forms — with `getAriaValue` / `hasAriaValue`, **only where reading an internals-supplied value makes sense.**
2. Consume normalized/typed return values (booleans; resolved vNode arrays for `idref(s)`); filter unattached elements (see #5139).
3. Add `ElementInternals` test fixtures (`_internals` and `axeInternalsMap`) mirroring existing virtual-node tests.
4. Update messages that name an ARIA attribute/role to reflect value source where confusing (e.g. avoid "ARIA attribute not allowed: aria-label" when the value came from internals).

### Out of scope
`role`-attribute reads (#5039); the **validity** checks `aria-valid-attr`, `aria-valid-attr-value`, `aria-unsupported-attr` (they inspect the literal HTML attribute set on purpose — do not convert); rule `selector` matching; form-label work (#5045).

### Dependencies
Blocked by #5043; requires `getAriaValue`/`hasAriaValue` (#5042). Prefer landing after the `lib/commons/aria` sub-issue.

### Acceptance criteria
- Existing suite green with `elementInternals: false`.
- New internals fixtures pass with `elementInternals: true`.
- Messages updated where they name an attribute/role.

chutchins25:
[...]

PR #5168, opened 2026-06-15
feat(rules): support ARIA element internals properties

### LINKED ISSUE #5150: ElementInternals: convert `lib/rules/*-matches` to `getAriaValue`/`hasAriaValue`

### Scope: `lib/rules/*-matches` (public-facing rule matchers — land last)
Representative files/usages:
- `aria-allowed-attr-matches`, `aria-has-attr-matches`
- `aria-hidden-focus-matches` (`aria-hidden`)
- `color-contrast-matches` (`aria-labelledby`)
- `landmark-has-body-context-matches`

Safest place to finish, since by this point every helper the matchers depend on has been converted.

### Out of scope
`role`-attribute reads (#5039); rule `selector` matching changes (future proposal — only the matcher-function bodies are in scope here); validity checks; form-label work (#5045).

### Dependencies
Blocked by #5043; requires `getAriaValue`/`hasAriaValue` (#5042). Land after the commons, core/utils, and checks sub-issues.

### Acceptance criteria

straker:
Updates the `rules` directory to use the new `getAriaValue` function where it makes sense. Only the JS match files were reviewed; JSON rule definitions have no ARIA attr access.

Here's a report of the full directory (JS files only):

* `rules/autocomplete-matches` - aria-readonly and aria-disabled → getAriaValue with `{ lowercase: true }` (both are boolean type); removed the now-redundant `.toLowerCase()` calls on the comparison
* `rules/aria-allowed-attr-matches` - iterates over `attrNames`, no direct `.attr('aria-*')` call
* `rules/aria-has-attr-matches` - checks for presence of any aria attr via `attrNames`, not a value lookup
* `rules/bypass-matches` - checks for links and headings, no ARIA attr access
* `rules/color-contrast-matches` - checks visibility, no ARIA attr access
* `rules/identical-links-same-purpose-matches` - computes accessible names, no ARIA attr access
* `rules/inserted-into-focus-order-matches` - checks tabindex, not an ARIA prop
* `rules/label-matches` - checks element type and role, no direct aria attr value access
* `rules/landmark-has-body-context-matches` - checks role, not an ARIA prop
* `rules/no-autoplay-audio-matches` - checks media attributes, no ARIA attr access
* `rules/no-empty-role-matches` - checks role attr, not an ARIA prop
* `rules/no-negative-tabindex-matches` - checks tabindex, not an ARIA prop
* `rules/no-role-matches` - checks role attr, not an ARIA prop

Closes: https://github.com/dequelabs/axe-core/issues/5150


lib/rules/aria-hidden-focus-matches.js
```diff
@@ -9,7 +11,14 @@ function shouldMatchElement(el) {
...
     return true;
   }
-  if (el.getAttribute('aria-hidden') === 'true') {
+
+  const vNode = getNodeFromTree(el);
+
+  // the composed parent could or could not be included in the tree so we'll need to handle either case
+  const ariaHidden = vNode
+    ? getAriaValue(vNode, 'aria-hidden', { lowercase: true })?.value
+    : el.getAttribute('aria-hidden');
```
WilcoFiers: Assume we can drop this with https://github.com/dequelabs/axe-core/pull/5177/changes#r3435735513

WilcoFiers (dismissed):
LGTM, just needs to be updated to 5177

chutchins25:
Heads up: the audit report in the PR description looks out of sync with the actual diff.

The description lists `autocomplete-matches` as the converted file and marks both `color-contrast-matches` and (implicitly) `aria-hidden-focus-matches` as having no ARIA attr access. But the diff actually converts:

- `lib/rules/aria-hidden-focus-matches.js` — `aria-hidden` → `getAriaValue`
- `lib/rules/color-contrast-matches.js` — `aria-labelledby` → `getResolvedRefs` / `getAriaValue`

…and doesn't touch `autocomplete-matches`. Worth refreshing the report so it matches what shipped before merge.

(Also updated the issue trailer — this is the `rules/*-matches` conversion, so it was pointing at the wrong issue. For history: changed from `Refs: https://github.com/dequelabs/axe-core/issues/5142` to `Closes: https://github.com/dequelabs/axe-core/issues/5150`.)

chutchins25:
Following up on my earlier note about the `autocomplete-matches` audit/diff mismatch — I looked into whether it actually needs converting, and I think it's the one matches-file that doesn't.

`autocomplete-matches` early-returns on `['textarea', 'input', 'select'].includes(nodeName) === false`, so it only runs on native form controls. Custom elements — the only carriers of ElementInternals — are excluded by that check, and `aria-readonly`/`aria-disabled` are reflected (`ariaReadonly`/`ariaDisabled`), so on a native element the property already syncs to the attribute `attr()` reads. There's no internals/property value `getAriaValue` would catch here that `attr()` misses.

The only behavioral difference would be `getAriaValue`'s trimming — e.g. `aria-readonly=" true "` would flip from "not readonly" to "readonly". Arguably more correct, but unrelated to element internals.

This mirrors the `aria-conditional-*-attr` checks we just closed #5144 on (native-gated, no internals reachability). Since this one's a *consumer* rather than a validator, converting for uniformity is defensible — but it buys no internals coverage and adds a small behavior change. Your call: convert for consistency, or leave it literal and drop the `autocomplete-matches` line from the audit. The two files you did convert (`aria-hidden-focus`, `color-contrast`) *do* reach custom elements, so those are clearly worth it.

lib/rules/aria-hidden-focus-matches.js
straker: Yes, but this pr now depends on https://github.com/dequelabs/axe-core/pull/5219

WilcoFiers (approved):

chutchins25 (approved):

PR #5169, opened 2026-06-15
feat(commons/text): support ARIA element internals properties

### LINKED ISSUE #5141: ElementInternals: convert `lib/commons/text` to `getAriaValue`/`hasAriaValue`

### Scope: `lib/commons/text` (accessible-name computation)
Representative files/usages:
- `native-text-methods` (`vNode.attr(attr)`)
- `native-text-alternative`
- `subtree-text`
- `title-text`
- `form-control-value` (`aria-selected`, `aria-valuenow`)

### Out of scope

### Dependencies

### Acceptance criteria

straker:
Updates the `commons/text` directory to use the new `getAriaValue` and `hasAriaValue` functions where it makes sense.

* `commons/text/form-control-value` - aria-selected → getAriaValue; aria-valuenow → hasAriaValue + getAriaValue
* `commons/text/accessible-text` / `accessible-text-virtual` - computes accessible names via commons/aria functions; no direct ARIA attr access
* `commons/text/label-text` / `label-virtual` / `label` - delegates to commons/aria functions; no direct ARIA attr access
* `commons/text/native-text-alternative` / `native-text-methods` / `native-element-type` - checks native HTML semantics (type, value, placeholder, etc.), not ARIA props
* `commons/text/title-text` - checks title attribute, not an ARIA prop
* `commons/text/visible-virtual` / `visible` / `visible-text-nodes` / `subtree-text` - text visibility utilities, no ARIA attr access
* `commons/text/sanitize` / `unicode` / `has-unicode` / `remove-unicode` / `is-human-interpretable` / `is-icon-ligature` / `is-valid-autocomplete` - text processing utilities, no ARIA attr access

Closes: https://github.com/dequelabs/axe-core/issues/5141

### DIFF
--- a/lib/commons/text/form-control-value.js
+++ b/lib/commons/text/form-control-value.js
@@ -150,8 +150,7 @@ function ariaListboxValue(node, context) {
   const selected = getOwnedVirtual(vNode).filter(
     owned =>
       getRole(owned) === 'option' &&
-      getAriaValue(owned, 'aria-selected', { lowercase: true })?.value ===
-        'true'
+      getAriaValue(owned, 'aria-selected')?.value === 'true'
   );
 
   if (selected.length === 0) {
--- a/lib/commons/text/form-control-value.js
+++ b/lib/commons/text/form-control-value.js
@@ -12,6 +12,8 @@ import getOwnedVirtual from '../aria/get-owned-virtual';
 import isHiddenForEveryone from '../dom/is-hidden-for-everyone';
 import { nodeLookup, querySelectorAll } from '../../core/utils';
 import log from '../../core/log';
+import getAriaValue from '../aria/get-aria-value';
+import hasAriaValue from '../aria/has-aria-value';
 
 export const controlValueRoles = [
   'textbox',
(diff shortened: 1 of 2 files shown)


chutchins25 (approved):
Looks good to me.

lib/commons/text/form-control-value.js
```diff
@@ -147,7 +149,8 @@ function ariaListboxValue(node, context) {
 
   const selected = getOwnedVirtual(vNode).filter(
     owned =>
-      getRole(owned) === 'option' && owned.attr('aria-selected') === 'true'
+      getRole(owned) === 'option' &&
+      getAriaValue(owned, 'aria-selected')?.value === 'true'
```
WilcoFiers, line 153: Looks like this case sensitivity is intentional. Can you add a comment about which browser/AT+versions you saw that case sensitivity mattered here for future reference?

WilcoFiers (changes requested):
Left you a Q on case sensitivity.

lib/commons/text/form-control-value.js
WilcoFiers: Left a msg on Slack. I actually think we should put a prop on the standard object on which things are case insensitive and put a comment on where / why we deviate from the standard there. That's a separate PR, so I think this is fine.

WilcoFiers (approved):

PR #5170, opened 2026-06-15
feat(checks/label): support ARIA element internals properties

### LINKED ISSUE #5146: ElementInternals: convert `lib/checks/label` + `lib/checks/forms` to `getAriaValue`/`hasAriaValue`

### Scope: `lib/checks/label` + `lib/checks/forms`
Representative files/usages:
- `no-implicit-explicit-label`
- autocomplete handling (`aria-readonly`, `aria-disabled`)

Coordinate with #5045 (form-associated labels) so the two efforts don't collide on the same files.

### Out of scope

### Dependencies

### Acceptance criteria

straker:
Updates the `checks/label` directory to use the new `getAriaValue`, `hasAriaValue`, and `getResolvedRefs` functions where it makes sense.

* `checks/label/help-same-as-label` - aria-describedby: replaced `node.getAttribute` + `idrefs()` with `hasAriaValue` + `getResolvedRefs`; since `getResolvedRefs` returns VirtualNodes, updated `accessibleText(thing)` to `accessibleText(thing.actualNode)`
* `checks/label/multiple-label` - aria-labelledby: replaced `idrefs()` with `getResolvedRefs`; since VirtualNodes are returned, updated `.includes(domNode)` to `.some(ref => ref?.actualNode === domNode)`
* `checks/label/title-only` - aria-describedby: replaced `.attr()` truthy check with `hasAriaValue` since only presence is needed
* `checks/label/alt-space-value` - checks alt attribute, not an ARIA prop
* `checks/label/duplicate-img-label` - compares accessible names, no direct ARIA attr access
* `checks/label/explicit` - checks label[for] association, no ARIA attr access
* `checks/label/hidden-explicit-label` - checks label visibility, no ARIA attr access
* `checks/label/implicit` - checks implicit label wrapping, no ARIA attr access
* `checks/label/label-content-name-mismatch` - computes accessible names, no direct ARIA attr access
* `checks/label/no-implicit-explicit-label` - no direct ARIA attr access (in aria directory, not label)

Closes: https://github.com/dequelabs/axe-core/issues/5146

### DIFF
--- a/lib/checks/label/multiple-label-evaluate.js
+++ b/lib/checks/label/multiple-label-evaluate.js
@@ -2,9 +2,9 @@ import {
   getRootNode,
   isHiddenForEveryone,
   isVisibleToScreenReaders,
-  idrefs
+  getResolvedRefs
 } from '../../commons/dom';
-import { escapeSelector } from '../../core/utils';
+import { escapeSelector, getNodeFromTree } from '../../core/utils';
 
 function multipleLabelEvaluate(node) {
   const id = escapeSelector(node.getAttribute('id'));
@@ -40,8 +40,11 @@ function multipleLabelEvaluate(node) {
       return undefined;
     }
     // make sure the ONE AT visible label is in the list of idRefs of aria-labelledby
-    const labelledby = idrefs(node, 'aria-labelledby');
-    return !labelledby.includes(ATVisibleLabels[0]) ? undefined : false;
+    const vNode = getNodeFromTree(node);
+    const labelledby = getResolvedRefs(vNode, 'aria-labelledby');
+    return !labelledby.some(ref => ref?.actualNode === ATVisibleLabels[0])
+      ? undefined
+      : false;
   }
 
   // only 1 CSS visible label
--- a/lib/checks/label/multiple-label-evaluate.js
+++ b/lib/checks/label/multiple-label-evaluate.js
@@ -4,7 +4,7 @@ import {
   isVisibleToScreenReaders,
   getResolvedRefs
 } from '../../commons/dom';
-import { escapeSelector, getNodeFromTree } from '../../core/utils';
+import { escapeSelector } from '../../core/utils';
 
 function multipleLabelEvaluate(node) {
   const id = escapeSelector(node.getAttribute('id'));
@@ -40,8 +40,7 @@ function multipleLabelEvaluate(node) {
       return undefined;
     }
     // make sure the ONE AT visible label is in the list of idRefs of aria-labelledby
-    const vNode = getNodeFromTree(node);
-    const labelledby = getResolvedRefs(vNode, 'aria-labelledby');
+    const labelledby = getResolvedRefs(node, 'aria-labelledby');
     return !labelledby.some(ref => ref?.actualNode === ATVisibleLabels[0])
       ? undefined
       : false;
--- a/lib/checks/label/title-only-evaluate.js
+++ b/lib/checks/label/title-only-evaluate.js
@@ -1,12 +1,12 @@
 import { labelVirtual, titleText } from '../../commons/text';
-import { hasAriaValue } from '../../commons/aria';
+import { getAriaValue } from '../../commons/aria';
 
 function titleOnlyEvaluate(node, options, virtualNode) {
   const labelText = labelVirtual(virtualNode);
   const title = titleText(virtualNode);
-  const ariaDescribedBy = hasAriaValue(virtualNode, 'aria-describedby');
+  const ariaDescribedBy = getAriaValue(virtualNode, 'aria-describedby');
 
-  return !labelText && !!(title || ariaDescribedBy);
+  return !labelText && !!(title || ariaDescribedBy?.value);
 }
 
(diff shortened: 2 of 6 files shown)


lib/checks/label/title-only-evaluate.js
```diff
@@ -1,9 +1,10 @@
 import { labelVirtual, titleText } from '../../commons/text';
+import { hasAriaValue } from '../../commons/aria';
 
 function titleOnlyEvaluate(node, options, virtualNode) {
   const labelText = labelVirtual(virtualNode);
   const title = titleText(virtualNode);
-  const ariaDescribedBy = virtualNode.attr('aria-describedby');
+  const ariaDescribedBy = hasAriaValue(virtualNode, 'aria-describedby');
```
chutchins25: Not a blocker, just a consistency question — `hasAriaValue` is presence-based (it returns `true` for an empty `aria-describedby=""`), whereas the previous `.attr('aria-describedby')` truthy check treated an empty attribute as absent. So an element with an empty `aria-describedby` and no title/label would now flip `title-only` from `false` to `true`. Such an element has no accessible name anyway, so the practical impact is negligible — was the presence-based semantics (vs. value-truthy) intended here?

lib/checks/label/multiple-label-evaluate.js
```diff
@@ -40,8 +40,11 @@ function multipleLabelEvaluate(node) {
       return undefined;
     }
     // make sure the ONE AT visible label is in the list of idRefs of aria-labelledby
-    const labelledby = idrefs(node, 'aria-labelledby');
-    return !labelledby.includes(ATVisibleLabels[0]) ? undefined : false;
+    const vNode = getNodeFromTree(node);
```
chutchins25: Not a blocker, just a simplification — `getResolvedRefs` accepts a raw DOM node and calls `nodeLookup` internally, so you can pass `node` directly and drop the `getNodeFromTree` import/call (`const labelledby = getResolvedRefs(node, 'aria-labelledby')`). Up to you if the intermediate `vNode` reads more clearly.

chutchins25 (commented):
Nice work — clean migration to the element-internals helpers, and the new `testutils-element` tests cover the internals path for each check. Two low-severity questions below, neither blocking.

lib/checks/label/title-only-evaluate.js
WilcoFiers: Agreed. This isn't right.

WilcoFiers (dismissed):
See Chris's comment

lib/checks/label/title-only-evaluate.js
chutchins25: Resolved — switched to `getAriaValue` and check `.value`, so an empty `aria-describedby` is treated as absent again.

lib/checks/label/multiple-label-evaluate.js
chutchins25: Resolved — passing `node` directly to `getResolvedRefs` and dropped the `getNodeFromTree` call.

test/checks/label/multiple-label.js
```diff
@@ -243,6 +243,38 @@ describe('multiple-label', () => {
     );
   });
 
+  // ariaLabelledByElements (reflected AOM property) coverage adapted from
+  // @jcfranco's work in #5187 (issue #4943)
```
WilcoFiers, line 247:
```suggestion
```

We can do attribution in other ways. Doesn't belong in the code.

lib/checks/label/multiple-label-evaluate.js
WilcoFiers, line 38: I'm skeptical this is still true. Would you mind opening an issue for us to investigate?

lib/checks/label/title-only-evaluate.js
```diff
@@ -1,11 +1,12 @@
...
+import { getAriaValue } from '../../commons/aria';
 
 function titleOnlyEvaluate(node, options, virtualNode) {
   const labelText = labelVirtual(virtualNode);
   const title = titleText(virtualNode);
-  const ariaDescribedBy = virtualNode.attr('aria-describedby');
+  const ariaDescribedBy = getAriaValue(virtualNode, 'aria-describedby');
 
-  return !labelText && !!(title || ariaDescribedBy);
+  return !labelText && !!(title || ariaDescribedBy?.value);
```
WilcoFiers, line 9: This is buggy (not new). We should consider building a proper accessibleDescription method and use that. Would you mind opening an issue for this?

WilcoFiers (approved):
LGTM. I would suggest pulling the comment, but I don't feel strongly about that. Couple issues to open though.

lib/checks/label/multiple-label-evaluate.js
chutchins25: Opened #5211 to investigate this.

lib/checks/label/title-only-evaluate.js
chutchins25: Opened #5212 to track building a proper accessibleDescription method and switching this check to use it.

Issue #5162, opened 2026-06-11
ElementInternals ARIA role conflict resolution


straker:
When resolving the role of an element with an attached internals role, and the element defines an explicit none role and is either focusable or has a global ARIA attribute, axe-core currently resolves the internal role as the implicit role. 

```html
<my-button tabindex="0" role="none">btn none focusable</my-button>
<script>
  customElements.define(
    'my-button',
    class MyButton extends HTMLElement {
      constructor() {
        super();
        this._internals = this.attachInternals();
        this._internals.role = 'button';
      }
    }
  )
</script>
```

According to the spec this should be correct, however no browser currently supports this. Tracking here so we  can watch it. Filed bugs with:

- [ ] [Chrome](https://issues.chromium.org/issues/522529333)
- [ ] [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=2046886)
- [ ] [Safari](https://bugs.webkit.org/show_bug.cgi?id=316919)
- [ ] [WPT](https://github.com/web-platform-tests/wpt/issues/60563)

straker:
We discussed our ideas on how to handle conflict resolution with element internals:

__1) should setting an internal global ARIA value and role, and an explicit role=none (`<testutils-element with-aria-haspopup="true" role="none">`) trigger role conflict resolution?__

We weren't sure at first. Wilco thought that it shouldn't since the internal role is a default value for the ARIA value and changing the role would override the default value. So a spinbutton has implicit ARIA values for `aria-valuemin/max/value`. We looked at the spec and discussed that default value !== implicit values for the role. We tested this out for the `output` element as it has an implicit global ARIA attribute for `aria-live`. 

Taking the following HTML:

```html
<form id="example-form" onSubmit="return false;">
  <input type="range" id="b" name="b" value="50" /> +
  <input type="number" id="a" name="a" value="10" /> =
  <output name="result" for="a b"></output>

</form>
<script>
  const form = document.getElementById("example-form");
  const a = form.elements["a"];
  const b = form.elements["b"];
  const result = form.elements["result"];

  function updateResult() {
  const aValue = a.valueAsNumber;
  const bValue = b.valueAsNumber;
    result.value = aValue + bValue;
  }

  form.addEventListener("input", updateResult);
</script>
```

[...]

PR #5171, opened 2026-06-15
feat(commons/aria): support ARIA element internals properties

### LINKED ISSUE #5140: ElementInternals: convert `lib/commons/aria` to `getAriaValue`/`hasAriaValue`

### Scope: `lib/commons/aria` (foundational — highest blast radius, land first)
Convert ARIA-attribute reads in the name/role resolution helpers used throughout axe-core.
Representative files/usages:
- `arialabel-text` (`aria-label`)
- `arialabelledby-text` (`aria-labelledby`)
- `label-virtual` (`aria-labelledby`)
- `get-owned-virtual` (`aria-owns`)
- `named-from-contents`
- `get-role` / `implicit-html-roles` (`hasAttr(attr)` for ARIA attrs)

1. Replace ARIA-attribute reads — `vNode.attr('aria-*')` / `vNode.hasAttr('aria-*')`, incl. variable-attr forms — with `getAriaValue(node, attrName)` / `hasAriaValue(node, attrName)`, **only where reading an internals-supplied value makes sense.**
2. Consume the normalized/typed return values (booleans not `"true"`; resolved vNode arrays not id strings for `idref(s)`); stop re-parsing strings; filter unattached elements (see #5139).
3. Add `ElementInternals` test fixtures (both property-style `_internals` and the `axeInternalsMap` path) mirroring the existing virtual-node tests for each touched function.
4. Update any check/helper message that names an ARIA attribute/role to reflect the value source where it would otherwise confuse.

### Out of scope
`role`-attribute reads (#5039); validity checks (`aria-valid-attr`, `aria-valid-attr-value`, `aria-unsupported-attr`); rule `selector` matching (future proposal); form-label work (#5045).

### Dependencies
- Blocked by #5043 (`getAttribute`/`hasAttribute` → vNode).
- Requires `getAriaValue`/`hasAriaValue` in `lib/commons/aria` (#5042). **Note:** confirm `hasAriaValue` actually exists — only `getAriaValue` was observed exported; a companion may be needed first.

### Acceptance criteria
- Existing suite green with `elementInternals: false` (no behavior change when the flag is off).

straker:
Updates the `commons/aria` directory to use the new `getAriaValue`, `hasAriaValue`, and `getResolvedRefs` functions where it makes sense.

* `commons/aria/arialabel-text` - aria-label → getAriaValue
* `commons/aria/arialabelledby-text` - aria-labelledby: `.attr()` check → hasAriaValue; `idrefs()` → getResolvedRefs; since getResolvedRefs returns VirtualNodes, switched `accessibleText(elm)` to `accessibleTextVirtual(elm)` for the direct VirtualNode
* `commons/aria/get-owned-virtual` - aria-owns: `.hasAttr()` → hasAriaValue; `idrefs(actualNode, ...)` + `getNodeFromTree()` → getResolvedRefs (already returns VirtualNodes, so getNodeFromTree is no longer needed)
* `commons/aria/label-virtual` - aria-labelledby: `.attr()` check → hasAriaValue; `idrefs()` + `getNodeFromTree()` → getResolvedRefs; aria-label → getAriaValue
* `commons/aria/validate-attr-value` - validates attrs, no conversion at this time
* `commons/aria/get-accessible-refs` - uses `node.hasAttribute('for')` on raw DOM node for HTML `for` attribute, not an ARIA prop
* `commons/aria/get-aria-value` - this is the new function being adopted
* `commons/aria/get-element-unallowed-roles` - checks `role` attribute, not an ARIA prop
* `commons/aria/get-role` - matching browser support and not using internals global attrs to check for role resolution, and also not running implicit role for role resolution if it's a custom element
* `commons/aria/has-aria-value` - this is the new function being adopted
* `commons/aria/lookup-table` - deprecated
* `commons/aria/allowed-attr` / `get-explicit-role` / `get-role-type` / `get-roles-by-type` / `get-roles-with-name-from-contents` / `implicit-role` / `implicit-nodes` / `is-accessible-ref` / `is-aria-role-allowed-on-element` / `is-combobox-popup` / `is-unsupported-role` / `is-valid-role` / `label` / `named-from-contents` / `required-attr` / `required-context` / `required-owned` / `validate-attr` - no ARIA attr value access

Closes: https://github.com/dequelabs/axe-core/issues/5140

### DIFF
--- a/lib/commons/aria/arialabel-text.js
+++ b/lib/commons/aria/arialabel-text.js
@@ -1,4 +1,5 @@
 import { nodeLookup } from '../../core/utils';
+import getAriaValue from './get-aria-value';
 
 /**
  * Get the text value of aria-label, if any
@@ -13,5 +14,5 @@ export default function arialabelText(element) {
     return '';
   }
 
-  return vNode.attr('aria-label') || '';
+  return getAriaValue(vNode, 'aria-label')?.value || '';
 }
--- a/test/commons/aria/get-role.js
+++ b/test/commons/aria/get-role.js
@@ -249,8 +249,11 @@ describe('aria.getRole', () => {
     // Presentational role inheritance is keyed off the HTML tag name of the
     // required child (e.g. `li`, `td`), not its computed role. A custom element
     // exposing an internal `listitem` role is not an HTML `<li>`, so it does not
-    // inherit a presentational role from an ancestor `role="none"` list. Whether
-    // it should is an open spec/browser question — see issue #5181.
+    // inherit a presentational role from an ancestor `role="none"` list. This
+    // matches Chrome and Firefox, which both leave the element as `listitem`
+    // (while stripping a native `<li>` to `none`/`generic`). Per spec it should
+    // inherit, but no browser implements internal-role conflict resolution yet
+    // — tracked in #5162.
     it('does not inherit presentation onto an ElementInternals listitem role', () => {
       fixture.innerHTML =
         '<ul role="none"><testutils-element with-role="listitem" id="target"></testutils-element></ul>';
--- a/test/commons/aria/get-role.js
+++ b/test/commons/aria/get-role.js
@@ -91,6 +91,14 @@ describe('aria.getRole', () => {
     assert.equal(aria.getRole(node), 'button');
   });
 
+  it('does not run role resolution with presentation role on custom element with an internal role', () => {
+    fixture.innerHTML =
+      '<ul><testutils-element with-role="listitem" id="target" role="presentation" aria-label="foo"></testutils-element></ul>';
+    flatTreeSetup(fixture);
+    const node = fixture.querySelector('#target');
+    assert.isNull(aria.getRole(node));
+  });
+
   describe('presentational role inheritance', () => {
     it('handles presentation role inheritance for ul', () => {
       fixture.innerHTML =
--- a/test/commons/aria/get-role.js
+++ b/test/commons/aria/get-role.js
@@ -245,6 +245,27 @@ describe('aria.getRole', () => {
         aria.getRole(node);
       });
     });
+
+    // Presentational role inheritance is keyed off the HTML tag name of the
+    // required child (e.g. `li`, `td`), not its computed role. A custom element
+    // exposing an internal `listitem` role is not an HTML `<li>`, so it does not
+    // inherit a presentational role from an ancestor `role="none"` list. Whether
+    // it should is an open spec/browser question — see issue #5181.
+    it('does not inherit presentation onto an ElementInternals listitem role', () => {
+      fixture.innerHTML =
+        '<ul role="none"><testutils-element with-role="listitem" id="target"></testutils-element></ul>';
+      flatTreeSetup(fixture);
+      const node = fixture.querySelector('#target');
+      assert.equal(aria.getRole(node), 'listitem');
+    });
+
+    it('does not inherit presentation onto an ElementInternals non-child role', () => {
+      fixture.innerHTML =
+        '<ul role="none"><testutils-element with-role="button" id="target"></testutils-element></ul>';
+      flatTreeSetup(fixture);
+      const node = fixture.querySelector('#target');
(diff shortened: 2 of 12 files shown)


chutchins25 (dismissed):
Looks good to me.

lib/commons/aria/arialabel-text.js
```diff
@@ -13,5 +14,5 @@ export default function arialabelText(element) {
     return '';
   }
 
-  return vNode.attr('aria-label') || '';
+  return getAriaValue(vNode, 'aria-label')?.value || '';
```
WilcoFiers, line 17: I'm not sure I like that `getAriaValue` returns null anymore. Should we change that to be `{ value: null, source: null }` instead so that we can safely deconstruct it? Separate PR though.

test/commons/aria/get-role.js
WilcoFiers, line 125: There's another interesting scenario here we didn't talk about:

```suggestion
    });

  it('handles presentation role inheritance on an ElementInternal role', () => {
    fixture.innerHTML =
      '<ul role="none"><testutils-element with-role="listitem" id="target"></testutils-element></ul>';
    flatTreeSetup(fixture);
    const node = fixture.querySelector('#target');
    assert.equal(aria.getRole(node), 'presentation');
  });
  
  it('does not set presentation role when ElementInternals isn\'t an allowed child', () => {
    fixture.innerHTML =
      '<ul role="none"><testutils-element with-role="button" id="target"></testutils-element></ul>';
    flatTreeSetup(fixture);
    const node = fixture.querySelector('#target');
    assert.equal(aria.getRole(node), 'button');
  });
```

I'm not actually sure this is how it works in practice. But role=none on ul SHOULD set its child listitems to presentational as well. We should test this I think, and maybe raise a case for in WPT if they don't already have one.

WilcoFiers (dismissed):
We need to look at presentational inheritance.

test/commons/aria/get-role.js
chutchins25: Tried to verify the actual behavior: 
`getRole` returns `listitem` (and `button` stays `button`). 
Presentational inheritance is keyed off the HTML tag name (`li`, `td`, …) in `inheritsPresentationChain`, not the computed role, so a `testutils-element` never matches.

Added two tests documenting this current behavior. Re-keying inheritance off computed role is a broader change (also affects `<div role="listitem">` etc.) and a genuine spec/WPT question, so I've split it into a follow-up: #5181.

chutchins25: Verified actual browser behavior for this exact case. Both Chrome and Firefox leave the custom element as `listitem` — neither applies presentational inheritance to an internal role (Chrome strips a native `<li>` to `none`, Firefox to `generic`, but both keep the custom element as `listitem`). So the current tests document the behavior that matches browsers.

This is the same spec-ahead-of-browsers gap as #5162 (internal-role conflict resolution — Chrome/Firefox/Safari/WPT bugs already filed there), so I've folded the follow-up into #5162 and closed #5181 rather than tracking it separately. Let me know if you're good with matching browsers here.

lib/commons/aria/arialabel-text.js
chutchins25: Agreed this is worth doing but out of scope for this PR (as you noted). Opened #5188 to track changing `getAriaValue` to return `{ value: null, source: null }`. Resolving this thread against that issue.

test/commons/aria/get-role.js
pkra: @WilcoFiers 

> But role=none on ul SHOULD set its child listitems to presentational as well.

I think ARIA only requires inheritance for (host-language-)required child elements that do not have explicit roles. (Though I spotted some odd phrasing that seems to have been introduced accidentally - I've filed https://github.com/w3c/aria/issues/2832 to clarify.)

(Then again `ul > test-element` is invalid HTML and its repair is not specified anywhere, I think.)

chutchins25 (approved):
Looks good to me. Pending Wilco reviewing the changes that spun off his earlier comment.

WilcoFiers (approved):

PR #5172, opened 2026-06-15
feat(checks/aria): support ARIA element internals properties

straker:
Updates the `checks/aria` directory to use the new `getAriaValue`, `hasAriaValue`, and `getResolvedRefs` functions where it makes sense.

Here's a report of the full directory:

* `checks/aria/aria-busy` - aria-busy → getAriaValue with `{ lowercase: true }` (boolean type)
* `checks/aria/aria-conditional-checkbox-attr` - aria-checked → getAriaValue with `{ lowercase: true }` (normalizeAriaChecked calls toLowerCase)
* `checks/aria/aria-conditional-radio-attr` - aria-checked → getAriaValue with `{ lowercase: true }` (same reason as checkbox)
* `checks/aria/aria-conditional-row-attr` - checks aria-rowindex/aria-colspan, but only via `.attrNames` presence check, no value access
* `checks/aria/aria-errormessage` - aria-errormessage → getAriaValue + hasAriaValue; aria-invalid → getAriaValue + hasAriaValue; `idrefs()` → getResolvedRefs; since getResolvedRefs returns a VirtualNode, changed `idref.getAttribute('aria-live')` to `idref.attr('aria-live')`; aria-describedby → getAriaValue
* `checks/aria/aria-hidden-body` - aria-hidden → getAriaValue (nmtoken type, no toLowerCase since original didn't use it)
* `checks/aria/aria-level` - aria-level → getAriaValue (int type, no toLowerCase)
* `checks/aria/aria-required-attr` - aria-valuetext → getAriaValue; aria-expanded → getAriaValue (in isClosedCombobox); required attr presence check → hasAriaValue (so internals-set required attrs are recognized)
* `checks/aria/aria-required-children` - aria-busy → getAriaValue with `{ lowercase: true }` (boolean type, two uses)
* `checks/aria/aria-valid-attr-value` - preCheck functions for aria-haspopup, aria-controls, aria-expanded, aria-selected, aria-current, aria-owns, aria-describedby, aria-labelledby → getAriaValue; the main attrNames loop and `const attrValue = virtualNode.attr(attrName)` remain as-is since validation iterates over actual HTML attributes only
* `checks/aria/braille-label-equivalent` - aria-braillelabel → getAriaValue (string type)
* `checks/aria/braille-roledescription-equivalent` - aria-brailleroledescription and aria-roledescription → getAriaValue (both string type); defaulted roleDesc to null so `typeof roleDesc !== 'string'` check is preserved
[...]

PR #5173, opened 2026-06-16
feat(aria-errormessage): honor element internals values

chutchins25:
Part of #5044, sub-issue #5144 (`lib/checks/aria`). This converts the **`aria-errormessage`** check — the one `lib/checks/aria` check that consumes an idref — to honor `ElementInternals`, using the now-merged `getResolvedRefs` (#5151).

## What
- Resolve the reference via **`getResolvedRefs`** instead of `idrefs` (handles the attribute, AOM property, and `ElementInternals` element arrays).
- Read **`aria-errormessage`** and **`aria-invalid`** through **`getAriaValue`/`hasAriaValue`**, and the referenced node's `aria-live` through `getAriaValue`, so internals-supplied values are honored the same as HTML attributes.
- Source-aware messages: append a `${data.source}` clause noting when a value came from element internals rather than an HTML attribute (see the review note below).

## Scope
- This is the `getResolvedRefs`-consuming slice of #5144. The remaining `getAriaValue`/`hasAriaValue`-only checks (`aria-busy`, `aria-level`, `aria-required-attr`, braille checks, `has-global-aria-attribute`) are a planned follow-up PR → **Part of #5144**, not Closes.
- Value-validating checks (`aria-valid-attr-value`, etc.) intentionally left on literal attribute reads, per the proposal and Wilco's guidance.

## Behavior / safety
- **No behavior change with the `elementInternals` flag off** — the source clause is empty for HTML attributes.
- All 23 existing `aria-errormessage` tests pass unchanged; 6 new `ElementInternals` tests added (`_internals` property, global `axeInternalsMap`, and a source-message assertion).
- Full `checks/aria` (415) and `aria-valid-attr-value` integration (238) green; `test:locales`, `eslint`, `tsc` clean.

Closes nothing on its own. Part of #5144.

### DIFF
--- a/lib/checks/aria/aria-errormessage-evaluate.js
+++ b/lib/checks/aria/aria-errormessage-evaluate.js
@@ -1,7 +1,11 @@
 import standards from '../../standards';
-import { idrefs, isVisibleToScreenReaders } from '../../commons/dom';
+import { getResolvedRefs, isVisibleToScreenReaders } from '../../commons/dom';
+import {
+  getAriaValue,
+  hasAriaValue,
+  getExplicitRole
+} from '../../commons/aria';
 import { tokenList } from '../../core/utils';
-import { getExplicitRole } from '../../commons/aria';
 /**
  * Check if `aria-errormessage` references an element that also uses a technique to announce the message (aria-live, aria-describedby, etc.).
  *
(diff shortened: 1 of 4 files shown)


chutchins25:
**@straker @WilcoFiers** — this is the first `check/aria` conversion in #5044, so it sets precedent on two points worth your review. Note #5080 made several checks internals-aware (for roles) *without* touching messages, so there's no existing convention to follow:

1. **Message source-awareness** (per the proposal's *"if the check message refers to an ARIA attribute or role, it should define a new message [reflecting] where it's defined"*): I append `${data.source}` → *" The aria-errormessage value is set via element internals, not an HTML attribute."* when the value isn't from the HTML attribute (empty string for attributes, so no change to existing output). Wording and approach are **provisional** — please confirm the convention you want, since it'll be replicated across all `check/aria` conversions.
2. **Value rendering for internals**: when `aria-errormessage` comes from internals, `data.values` renders `getAriaValue`'s node-name string (e.g. `` `[DIV]` ``) since internals references have no id. Flagging in case you'd prefer a different representation (or to suppress the value entirely for the internals case).

chutchins25:
https://github.com/dequelabs/axe-core/pull/5172 covers the purpose of this PR making it obsolete. Closing it.

PR #5175, opened 2026-06-16
test(checks/shared): confirm element internals do not trigger role conflict resolution

### LINKED ISSUE #5145: ElementInternals: convert `lib/checks/shared` to `getAriaValue`/`hasAriaValue`

### Scope: `lib/checks/shared`
Representative files/usages:
- `aria-hidden-body` (`aria-hidden`)
- `help-same-as-label` (`aria-describedby`)
- `title-only` (`aria-describedby`)
- `attr-non-space-content` (`vNode.attr(options.attribute)`)
- `presentational-role` (`hasAttr(attr)`)

### Out of scope

### Dependencies

### Acceptance criteria

chutchins25:
Audited `lib/checks/shared` for the #5044 conversion — no `getAriaValue`/`hasAriaValue` conversion applies:

- `aria-label` / `aria-labelledby` checks delegate to the already-converted `commons/aria` helpers (`arialabelText` / `arialabelledbyText`), so they're internals-aware transitively.
- `presentational-role` uses one global-ARIA **presence** check for role conflict resolution, which stays `hasAttr` — element internals must **not** trigger conflict resolution (no browser supports it yet; see #5162). This matches the convention already set in #5172 (`has-global-aria-attribute`) and #5171 (`get-role`).

So this is a **test-only** change: it adds a test demonstrating that a global ARIA attribute supplied via element internals does **not** override a presentational role. No production code changes.

Closes #5145

### DIFF
--- a/lib/checks/shared/presentational-role-evaluate.js
+++ b/lib/checks/shared/presentational-role-evaluate.js
@@ -1,4 +1,4 @@
-import { getExplicitRole, getRole, hasAriaValue } from '../../commons/aria';
+import { getExplicitRole, getRole } from '../../commons/aria';
 import { getGlobalAriaAttrs } from '../../commons/standards';
 import { isFocusable } from '../../commons/dom';
 
@@ -34,10 +34,8 @@ export default function presentationalRoleEvaluate(node, options, virtualNode) {
 
   // user intended to make this presentational so inform them of
   // problems caused by role conflict resolution
-  // use hasAriaValue so a global ARIA attribute supplied via the reflected AOM
-  // property or element internals counts, not only the HTML attribute
   const hasGlobalAria = getGlobalAriaAttrs().some(attr =>
-    hasAriaValue(virtualNode, attr)
+    virtualNode.hasAttr(attr)
   );
   const focusable = isFocusable(virtualNode);
   let messageKey;
--- a/lib/checks/shared/presentational-role-evaluate.js
+++ b/lib/checks/shared/presentational-role-evaluate.js
@@ -1,4 +1,4 @@
-import { getExplicitRole, getRole } from '../../commons/aria';
+import { getExplicitRole, getRole, hasAriaValue } from '../../commons/aria';
 import { getGlobalAriaAttrs } from '../../commons/standards';
 import { isFocusable } from '../../commons/dom';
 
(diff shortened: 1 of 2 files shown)


test/checks/shared/presentational-role.js
```diff
@@ -62,6 +62,27 @@ describe('presentational-role', () => {
...
+
+    assert.isFalse(checkEvaluate.call(checkContext, null, null, vNode));
+    assert.deepEqual(checkContext._data.messageKey, 'both');
+  });
+
+  it('detects a global aria attribute set via element internals', () => {
+    const vNode = queryFixture(
+      '<testutils-element id="target" no-role role="none" tabindex="0">x</testutils-element>'
+    );
+    vNode.actualNode._internals.ariaLive = 'assertive';
```
WilcoFiers: You can use `with-aria-live`. Steve added that to testutils-element.

test/checks/shared/presentational-role.js
```diff
@@ -62,6 +62,27 @@ describe('presentational-role', () => {
...
+
+    assert.isFalse(checkEvaluate.call(checkContext, null, null, vNode));
```
WilcoFiers:
```suggestion
    assert.isTrue(checkEvaluate.call(checkContext, null, null, vNode));
```

lib/checks/shared/presentational-role-evaluate.js
```diff
@@ -34,8 +34,10 @@ export default function presentationalRoleEvaluate(node, options, virtualNode) {
 
   // user intended to make this presentational so inform them of
   // problems caused by role conflict resolution
+  // use hasAriaValue so a global ARIA attribute supplied via the reflected AOM
+  // property or element internals counts, not only the HTML attribute
   const hasGlobalAria = getGlobalAriaAttrs().some(attr =>
-    virtualNode.hasAttr(attr)
+    hasAriaValue(virtualNode, attr)
```
WilcoFiers: This is incorrect. Internals do not trigger conflict resolution. See https://github.com/dequelabs/axe-core/issues/5162

This file doesn't need to change. I do think we need a test to demonstrate that we don't want internals to trigger conflict res.

WilcoFiers (dismissed):

lib/checks/shared/presentational-role-evaluate.js
chutchins25: Reverted — agreed, internals shouldn't trigger conflict resolution (#5162). The file no longer changes; added a test demonstrating an internals-set global ARIA attribute doesn't override a presentational role.

test/checks/shared/presentational-role.js
chutchins25: Done — using `with-aria-live` now.

test/checks/shared/presentational-role.js
chutchins25: Done — switched to `assert.isTrue`.

WilcoFiers (approved):

PR #5177, opened 2026-06-16
chore: support elements not in tree for get/hasAriaValue

straker:
Found this when tying to update our [matches functions](https://github.com/dequelabs/axe-core/pull/5168). 

There's a few times when we traverse up the DOM tree (`getComposedParent`, just a loop, etc.) and we want to get an ARIA attribute from it (e.g. `aria-hidden-focus-matches.js` walks up the tree looking for any ancestor with `aria-hidden`). Currently it's a bit clunky to have to handle both elements that could or could not be part of the tree as `getAriaValue` assumes vNode, so you have to do something like this:

```js
const vNode = getNodeFromTree(el);
const ariaHidden = vNode
  ? getAriaValue(vNode, 'aria-hidden', { lowercase: true })?.value
  : (el.getAttribute('aria-hidden') || '').toLowerCase();
```

This is also a problem if the element we want to look at is a custom element outside the tree, as we won't be able to look at the element internals of it.

This fixes that by allowing the two get/has aria value functions to fallback to DOM nodes if the node is not in the tree.

### DIFF
--- a/lib/commons/aria/get-aria-value.js
+++ b/lib/commons/aria/get-aria-value.js
@@ -85,7 +85,7 @@ function getAttributeValue(node, attrStandard, attrName) {
   }
 
   // return null for the attribute if the value is empty but the idref(s) prop is not
-  const propValue = getPropertyValue(vNode, attrStandard);
+  const propValue = getPropertyValue(node, attrStandard);
   const propEmpty = Array.isArray(propValue) ? propValue.length : !!propValue;
   return propEmpty ? null : value;
 }
--- a/lib/commons/aria/get-aria-value.js
+++ b/lib/commons/aria/get-aria-value.js
@@ -35,11 +35,12 @@ export default function getAriaValue(node, attrName, options = {}) {
     return null;
   }
 
+  const { vNode, domNode } = nodeLookup(node);
   const { type } = attrStandard;
   const { lowercase } = options;
 
   for (const { source, getValue } of sources) {
-    let value = getValue(node, attrStandard, attrName);
+    let value = getValue(vNode, domNode, attrStandard, attrName);
     if (value === null || value === undefined) {
       continue;
     }
@@ -68,8 +69,7 @@ export default function getAriaValue(node, attrName, options = {}) {
   return null;
 }
 
-function getAttributeValue(node, attrStandard, attrName) {
-  const { vNode, domNode } = nodeLookup(node);
+function getAttributeValue(vNode, domNode, attrStandard, attrName) {
   const { type } = attrStandard;
   const value = vNode ? vNode.attr(attrName) : domNode.getAttribute(attrName);
 
@@ -85,21 +85,25 @@ function getAttributeValue(node, attrStandard, attrName) {
   }
 
   // return null for the attribute if the value is empty but the idref(s) prop is not
-  const propValue = getPropertyValue(node, attrStandard);
+  const propValue = getPropertyValue(vNode, domNode, attrStandard);
   const propEmpty = Array.isArray(propValue) ? propValue.length : !!propValue;
   return propEmpty ? null : value;
 }
 
-function getPropertyValue(node, attrStandard) {
-  const { domNode } = nodeLookup(node);
+function getPropertyValue(vNode, domNode, attrStandard) {
   const { prop } = attrStandard;
   return prop && domNode ? domNode[prop] : null;
 }
 
-function getInternalValue(node, attrStandard) {
-  const { vNode, domNode } = nodeLookup(node);
+function getInternalValue(vNode, domNode, attrStandard) {
   const { prop } = attrStandard;
   if (!prop) {
+    return null;
+  }
+
+  // feature flag to enable internals. uses globalThis.axe as it can be run outside of axe context
+  // TODO: remove when feature is fully enabled
+  if (!axe._enableElementInternals) {
     return null;
   }
 
--- a/lib/commons/aria/get-aria-value.js
+++ b/lib/commons/aria/get-aria-value.js
@@ -1,4 +1,4 @@
-import { nodeLookup } from '../../core/utils';
+import { nodeLookup, getElementInternals } from '../../core/utils';
 import standards from '../../standards';
 
 const idrefTypes = ['idref', 'idrefs'];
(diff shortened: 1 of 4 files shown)


lib/commons/aria/get-aria-value.js
```diff
@@ -69,9 +68,10 @@ export default function getAriaValue(node, attrName, options = {}) {
   return null;
 }
 
-function getAttributeValue(vNode, attrStandard, attrName) {
+function getAttributeValue(node, attrStandard, attrName) {
+  const { vNode, domNode } = nodeLookup(node);
   const { type } = attrStandard;
-  const value = vNode.attr(attrName);
+  const value = vNode ? vNode.attr(attrName) : domNode.getAttribute(attrName);
```
chutchins25, line 74: A few lines below (line 88), the idref(s) empty-attribute fallback still passes the old `vNode`:

```js
const propValue = getPropertyValue(vNode, attrStandard);
```

Was that meant to be `node`? `getPropertyValue` now runs its own `nodeLookup`, so for an out-of-tree element `vNode` is `undefined` here and `getPropertyValue(undefined, …)` returns `null`. The empty-attribute fallback then treats the property as empty and returns the empty `aria-*` attribute value instead of falling through to the property source — which is the case this PR is adding support for. In-tree nodes are unaffected (there `vNode` is truthy). Suggested fix:

```js
const propValue = getPropertyValue(node, attrStandard);
```

A regression test would guard this — an out-of-tree idref case (empty `aria-labelledby` attribute + a set `ariaLabelledByElements`, fetched via `fixture.querySelector`) — since the new tests cover the attribute and internals sources but not the idref path.

chutchins25 (dismissed):
Nice work — the `nodeLookup` normalization reads cleanly and the in-tree path stays behavior-preserving. One question on the idref(s) fallback in `getAttributeValue` (inline).

lib/commons/aria/has-aria-value.js
```diff
@@ -16,31 +16,37 @@ export default function hasAriaValue(node, attrName) {
     throw new TypeError(`Attribute ${attrName} is not an ARIA attribute`);
   }
 
-  const { vNode } = nodeLookup(node);
```
WilcoFiers, line 19: Feels a bit clunky to call nodeLookup several times. maybe you can pass domNode and vNode into these functions?

chutchins25 (dismissed):
Approving — this resolves the `getPropertyValue` issue I flagged on the prior commit: `getAttributeValue` now passes `node` (not the stale, out-of-tree-`undefined` `vNode`) into the idref(s) empty-attribute fallback, and the new out-of-tree `idref`/`idrefs` property tests guard that path. The `nodeLookup(node).domNode` swap is behavior-preserving for in-tree and SerialVirtualNode inputs, and out-of-tree support reads cleanly.

One optional, non-blocking polish nit (no action required): in `has-aria-value.js`, `hasInternalValue` computes `getElementInternals(domNode)` before the `prop` check, so for ARIA attributes with no associated `prop` it does the full internals lookup on out-of-tree nodes only to discard it. Its sibling `getInternalValue` in `get-aria-value.js` short-circuits with `if (!prop) return null;` first. Reordering `hasInternalValue` to check `prop` first would match that and skip the wasted work — purely a consistency/micro-perf thing, fine to leave.

WilcoFiers (approved):
Left a suggestion. I'm okay with it either way.

lib/commons/aria/get-aria-value.js
```diff
@@ -85,17 +85,27 @@ function getAttributeValue(vNode, attrStandard, attrName) {
...
 
-function getInternalValue(vNode, attrStandard) {
+function getInternalValue(node, attrStandard) {
+  const { vNode, domNode } = nodeLookup(node);
   const { prop } = attrStandard;
-  return prop && vNode.elementInternals ? vNode.elementInternals[prop] : null;
+  if (!prop) {
+    return null;
+  }
+
```
WilcoFiers, line 103:
```suggestion

if (!axe._enableElementInternals) {
  return null
}
```

lib/commons/aria/has-aria-value.js
```diff
@@ -16,31 +16,37 @@ export default function hasAriaValue(node, attrName) {
...
 
-function hasInternalValue(vNode, attrStandard) {
+function hasInternalValue(node, attrStandard) {
+  const { vNode, domNode } = nodeLookup(node);
   const { prop } = attrStandard;
-  if (prop && vNode.elementInternals) {
-    const internalsValue = vNode.elementInternals[prop];
+  const internals = vNode
+    ? vNode.elementInternals
+    : getElementInternals(domNode);
```
WilcoFiers, line 52: Same here, need to put the feature flag gate in. Or maybe put that into getElementInternals directly?

WilcoFiers (dismissed):

lib/commons/aria/has-aria-value.js
straker: We didn't want to put it directly into `getElementInternals` as it's used by the `gather-internals` script and that couldn't really use a global `axe` due to the extensions needing it.

test/commons/aria/get-aria-value.js
```diff
@@ -173,6 +177,7 @@ describe('aria.getAriaValue', () => {
   });
 
   it('returns element internals from an element not in the tree', () => {
+    axe._elemen;
```
WilcoFiers, line 180: This looks like it doesn't belong here.

WilcoFiers (changes requested):

chutchins25:
Hey @straker — flagging this one since it's a blocker for #5168 (the `lib/rules/*-matches` conversion, part of #5150). #5168 is otherwise pre-approved — WilcoFiers left *"LGTM, just needs to be updated to 5177"* — so once this lands I can rebase #5168, drop the `aria-hidden-focus-matches.js` change, and it's ready.

WilcoFiers flipped this from approved to changes-requested on Jul 13 with two open threads:

1. **`has-aria-value.js`** — gate the internals lookup behind the `elementInternals` run option (he suggested maybe doing it inside `getElementInternals` directly).
2. **`test/commons/aria/get-aria-value.js`** — a test he flagged as not belonging here; needs removing or relocating.

Both look small, and CI is green — it's ~12 commits behind `develop`, so it'll want a quick rebase too. Anything I can help with?

straker:
We have decided that axe-core should not handle nodes outside of the virtual tree as doing `axe.run()` will always populate the virtual tree with the entire DOM tree (excluding closed shadow DOM). This code was only needed as the tests do not set up the entire DOM tree as the virtual tree, which causes problems in just our tests. We'll update the test code to set up the virtual tree to better parity real use.

PR #5178, opened 2026-06-17
test(context): cover NodeList for include/exclude and fix doc examples

### LINKED ISSUE #3348: Update docs and tests to add NodeList support to include/exclude
We confirmed in #3338 that `NodeList` or any array-like object is supported for the include/exclude context object. We don't have docs or tests for this, so we just need to add a few tests and update the docs.

cycsmail:
The `context.md` examples labeled "Test a Node list" and "Test everything except these DOM nodes" used `document.querySelector(...)`, which only ever returns a single node, so they didn't actually show a NodeList. Switched both to `document.querySelectorAll(...)` and tweaked the comments so the examples demonstrate the NodeList support that already works. Also added two `Context` unit tests: one for a NodeList passed as the bare context, and one for NodeLists passed via `include`/`exclude`. (`doc/API.md` already lists NodeList as an accepted context, so it didn't need a change.)

Heads up: I couldn't get the browser test runner going in my environment, so the new assertions were checked against the built `axe.js` bundle under jsdom rather than web-test-runner — prettier and eslint are clean. Worth a CI run to confirm in-browser.

Closes: #3348

### DIFF
--- a/doc/context.md
+++ b/doc/context.md
@@ -43,8 +43,8 @@ const img = document.createElement('img');
 document.body.appendChild(img);
 await axe.run(img);
 
-// Test a Node list:
-const nodes = document.querySelector('main, header');
+// Test a NodeList, such as one returned by document.querySelectorAll:
+const nodes = document.querySelectorAll('main, header');
 await axe.run(nodes);
 
 // Test an array of nodes:
@@ -73,8 +73,8 @@ There are often areas of a page that as a developer you have no control over. Yo
 // Test everything except the ad banners:
 await axe.run({ exclude: '.ad-banner' });
 
-// Test everything except these DOM nodes:
-const youtubeVids = document.querySelector('iframe[src^="youtube.com"]');
+// Test everything except these DOM nodes (a NodeList works too):
+const youtubeVids = document.querySelectorAll('iframe[src^="youtube.com"]');
 await axe.run({ exclude: youtubeVids });
 ```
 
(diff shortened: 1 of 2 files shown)


straker:
>  I couldn't get the browser test runner going in my environment, so the new assertions were checked against the built axe.js bundle under jsdom rather than web-test-runner — prettier and eslint are clean

Interesting. What environment are you running where that didn't work? We just switched over to web test runner so I'd like to make sure it's usable for anyone who wants to work in axe-core.

cycsmail:
Chrome-for-Testing downloaded fine but the browser wouldn't launch in my container — needed `--no-sandbox` and was missing the usual system libs (libnss3, libgbm, libasound2…). So I ran the same assertions through the jsdom path against the built `axe.js` instead. The two tests are in the normal browser suite though, so they'll run fine in your CI.

If you want it smoother for container/CI folks, passing `--no-sandbox --disable-dev-shm-usage` in the launcher + a note on the apt deps would cover most headless setups.

cycsmail:
Update — sorted the browser in my container (system libs + `--no-sandbox`) and ran the full web-test-runner suite this time. Both new NodeList tests pass in real Chromium, whole suite green (8,304 across 124 files) — holds up in-browser, not just the jsdom check.

straker (approved):
Thanks for the pr! Apologies for it taking some time to get in.

Approved for security

cycsmail:
no worries at all, thanks for getting it in!

PR #5198, opened 2026-07-01
refactor: use virtual node attr/hasAttr in rules and checks

chutchins25:
Converts `getAttribute`/`hasAttribute` calls to the VirtualNode `.attr()`/`.hasAttr()` equivalents where it's a safe, behavior-preserving drop-in. `.attr()`/`.hasAttr()` return the raw attribute value (or `null`) exactly like `getAttribute`/`hasAttribute` — no trimming or normalization — so these swaps preserve behavior.

## Scope

A full sweep of `lib/` found **71 raw `getAttribute`/`hasAttribute` calls** (74 grep hits − 3 JSDoc comments). Of these, **13 are converted here**; the remaining **58 are intentionally left as-is** (52 must-stay + 6 judgment). #5043 is satisfied by converting where it makes sense, not by driving the count to zero.

### ✅ Converted (13)

**`lib/rules/*-matches` (6):**

| Site | Change |
|------|--------|
| `xml-lang-mismatch-matches` | `getAttribute('lang')`, `getAttribute('xml:lang')` → `.attr()` |
| `duplicate-id-active-matches` | `getAttribute('id').trim()` → `.attr('id').trim()` |
| `duplicate-id-misc-matches` | `getAttribute('id').trim()` → `.attr('id').trim()` |
| `frame-title-has-text-matches` | `getAttribute('title')` → `.attr('title')` |
| `landmark-has-body-context-matches` | `hasAttribute('role')` → `.hasAttr('role')` |

**`lib/checks/` (7):**

| Site | Change |
|------|--------|
| `label/help-same-as-label` | `getAttribute('title')`, `getAttribute('aria-describedby')` → `.attr()` |
| `label/hidden-explicit-label` | `getAttribute('id')` → `.attr('id')` |
| `label/multiple-label` | `getAttribute('id')` → `.attr('id')` |
| `parsing/duplicate-id` | `getAttribute('id').trim()` → `.attr('id').trim()` |
| `tables/scope-value` | `getAttribute('scope').toLowerCase()` → `.attr('scope').toLowerCase()` |
| `generic/page-no-duplicate` | `elm.actualNode.hasAttribute('role')` → `elm.hasAttr('role')` |

`matches` functions receive `virtualNode` as the 2nd argument and checks as the 3rd, so the virtual node is always available (a few functions had the parameter added). Affected `matches`/check tests were updated to pass a virtual node argument.

### ⏸️ Left as-is — must-stay (52)

These operate on real DOM by design; converting them would be incorrect or pointless.

| Area | Sites | Reason |
|------|-------|--------|
| `core/utils/is-hidden`, `preload-media`, `get-xpath`, `get-selector`, `preload-cssom` | 14 | `core/utils` works on real DOM by design |
| `core/base/context/create-frame-context` | 3 | Runs before the tree is built (raw frame element) |
| `core/base/virtual-node` | 3 | This *is* the `.attr()`/`.hasAttr()` implementation + constructor |
| `commons/table` (`to-grid`, `is-header`, `is-data-table`, `get-headers`) | 11 | Raw cells accessed via `.rows`/`.cells`; `is-data-table` takes a real DOM table |
| `commons/aria/get-accessible-refs` | 3 | Raw `childNodes` recursion |
| `commons/dom` (`idrefs`, `is-current-page-link`, `url-props-from-attribute`, `is-visible`) | 6 | `idrefs` left per #5138; live URL props; `is-visible:147` is an intentional vNode-null fallback |
| `checks/tables/th-has-data-cells`, `td-headers-attr` | 8 | Raw cells; table cells can't be custom elements |
| `checks/aria/aria-required-parent` | 2 | `parentElement` walk on the raw element |
| `rules/aria-hidden-focus-matches` | 1 | Real-DOM composed-parent walk (no vNode in the recursion) |
| `rules/color-contrast-matches` | 1 | `control` comes from `getAccessibleRefs` (raw refs); color-contrast is #5150 scope |

### ⏸️ Left as-is — judgment (6)

These mix attribute reads with live DOM/property reads, so converting only the attribute reads would be inconsistent.

| Site | Sites | Reason |
|------|-------|--------|
| `commons/aria/lookup-table` | 3 | `evaluateRoleForElement` handlers read live props (`node.type`, etc.) alongside the attribute |
| `checks/media/no-autoplay-audio-evaluate`, `rules/no-autoplay-audio-matches` | 3 | Mixed with live media props (`node.duration`, `node.currentSrc`) |

## Testing

- `matches`/check tests updated to pass a virtual node argument (some previously passed only the DOM node and would throw on the new `.attr()` reads).
- `npm run eslint`, `npm run test:tsc`, `npm run build`, and the affected rule-matches + check suites all pass.

Closes #5043

### DIFF
--- a/lib/checks/generic/page-no-duplicate-evaluate.js
+++ b/lib/checks/generic/page-no-duplicate-evaluate.js
@@ -28,8 +28,7 @@ function pageNoDuplicateEvaluate(node, options, virtualNode) {
   if (typeof options.nativeScopeFilter === 'string') {
     elms = elms.filter(elm => {
       return (
-        elm.actualNode.hasAttribute('role') ||
-        !findUpVirtual(elm, options.nativeScopeFilter)
+        elm.hasAttr('role') || !findUpVirtual(elm, options.nativeScopeFilter)
       );
     });
   }
--- a/lib/checks/label/help-same-as-label-evaluate.js
+++ b/lib/checks/label/help-same-as-label-evaluate.js
@@ -8,7 +8,7 @@ import { hasAriaValue } from '../../commons/aria';
 
 function helpSameAsLabelEvaluate(node, options, virtualNode) {
   const labelText = labelVirtual(virtualNode);
-  let check = node.getAttribute('title');
+  let check = virtualNode.attr('title');
 
   if (!labelText) {
     return false;
--- a/lib/checks/label/hidden-explicit-label-evaluate.js
+++ b/lib/checks/label/hidden-explicit-label-evaluate.js
@@ -9,7 +9,7 @@ function hiddenExplicitLabelEvaluate(node, options, virtualNode) {
     }
 
     const root = getRootNode(node);
-    const id = escapeSelector(node.getAttribute('id'));
+    const id = escapeSelector(virtualNode.attr('id'));
     const label = root.querySelector(`label[for="${id}"]`);
 
     if (label && !isVisibleToScreenReaders(label)) {
--- a/lib/checks/label/multiple-label-evaluate.js
+++ b/lib/checks/label/multiple-label-evaluate.js
@@ -6,8 +6,8 @@ import {
 } from '../../commons/dom';
 import { escapeSelector } from '../../core/utils';
 
-function multipleLabelEvaluate(node) {
-  const id = escapeSelector(node.getAttribute('id'));
+function multipleLabelEvaluate(node, options, virtualNode) {
+  const id = escapeSelector(virtualNode.attr('id'));
   let parent = node.parentNode;
   let root = getRootNode(node);
   root = root.documentElement || root;
--- a/lib/checks/parsing/duplicate-id-evaluate.js
+++ b/lib/checks/parsing/duplicate-id-evaluate.js
@@ -1,8 +1,8 @@
 import { getRootNode } from '../../commons/dom';
 import { escapeSelector } from '../../core/utils';
 
-function duplicateIdEvaluate(node) {
-  const id = node.getAttribute('id').trim();
+function duplicateIdEvaluate(node, options, virtualNode) {
+  const id = virtualNode.attr('id').trim();
   // Since empty ID's are not meaningful and are ignored by Edge, we'll
   // let those pass.
   if (!id) {
--- a/lib/checks/tables/scope-value-evaluate.js
+++ b/lib/checks/tables/scope-value-evaluate.js
@@ -1,5 +1,5 @@
-function scopeValueEvaluate(node, options) {
-  const value = node.getAttribute('scope').toLowerCase();
+function scopeValueEvaluate(node, options, virtualNode) {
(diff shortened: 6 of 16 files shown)


WilcoFiers (dismissed):
We should be converting these to `has/getAriaValue`.

chutchins25:
Thanks @WilcoFiers. Wanted to check scope before I make changes, because `getAriaValue`/`hasAriaValue` both key off `standards.ariaAttrs[attrName]`, so they only apply to ARIA attributes — and 12 of the 13 reads this PR converts are **non-ARIA**:

| Attribute | Sites | `getAriaValue`/`hasAriaValue` behavior |
|---|---|---|
| `lang`, `xml:lang`, `id`, `title`, `scope` | 10 | `getAriaValue(...)` returns `null` |
| `role` | landmark-has-body-context, page-no-duplicate | `hasAriaValue('role')` throws `TypeError: Attribute role is not an ARIA attribute` |

So a literal swap would either no-op to `null` or throw for those.

The one genuine ARIA read is `aria-describedby` in `help-same-as-label`. Even there it's not a clean `getAriaValue` swap: it's `type: idrefs`, used as a presence guard before `idrefs()` resolution — making it internals-aware really wants `getResolvedRefs` (#5151), which I'd put in #5150's scope rather than here.

So before I adjust: are you pointing at specific sites, or is the broader point that ARIA-semantic reads belong in the #5044 `getAriaValue` work and the plain `getAttribute → .attr()` conversions in #5043 aren't worth doing on their own? Happy to rescope either way.

cc @straker since this sits on the #5043-vs-#5044 boundary.

straker:
In the order of changes, this was technically suppose to be done before any of the migration work to use `get/hasAriaValue`, but we kind of did it all at once. So moving these to use the new functions where appropriate would be helpful so we don't have to do it twice.

chutchins25:
Pushed the one ARIA read here — `aria-describedby` in `help-same-as-label` — over to `hasAriaValue`, which also picks up the property/ElementInternals reflection.

The other 12 reads in this PR are non-ARIA (`lang`, `xml:lang`, `id`, `title`, `scope`, `role`) — `getAriaValue`/`hasAriaValue` key off `standards.ariaAttrs`, so they return `null` / throw for those, and stay on `.attr()`/`.hasAttr()`.

@WilcoFiers re-requesting review to clear the change request. @straker this takes the one ARIA site straight to the new function so it won't need a second pass in the `get/hasAriaValue` migration.

chutchins25:
Update after rebasing on develop: develop already migrated `help-same-as-label` to `hasAriaValue` + `getResolvedRefs`, so I dropped my redundant commit. This PR's remaining `help-same-as-label` change is just the `title` → `.attr()` conversion; the one ARIA read (`aria-describedby`) is already on `hasAriaValue` via develop. Net: all conversions in this PR are non-ARIA `.attr()`/`.hasAttr()` swaps, which is the correct end state.

straker (approved):

Issue #4584, opened 2024-09-24
Reenable ARIA practices test for tabs-action 


straker:
ARIA practices created [a new tabs page](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-actions/) that uses [a proposed `aria-actions` attribute](https://github.com/w3c/aria/issues/1440) for handling tabs with a primary and secondary action. Our nightly tests are failing on the page so we decided to disable testing on the page. 

For now we turned off testing the page, but we should look at reenabling it.

straker:
ARIA practices have now added [a new listbox page](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-actions/) that does the same, so we've disabled it as well.

padmavemulapati:
Valdiated using latest axe-core develop nranch code base,

Observations & Valdiations:

In `lib/standards/aria-attrs.js`, `aria-actions` is defined as type: 'idrefs', and `allowEmpty: true`, `global: true`. 
1. Because of it's global: true, allowed on any role/element (so no aria-allowed-attr violations), and 
2. because of allowEmpty: true, aria-actions="" is a valid value. 
3. The only violations come from aria-prohibited-attr: it's prohibited on name-prohibited roles. 
4. fixture uses role="button" as a pass and role="code" / role="mark" as fails. 
5. Also  APG tabs-actions and listbox-actions pages, axe recognizes the attribute but the authoring pattern still trips structural rules (aria-required-children, nested-interactive), which is why those pages disable specific rules rather than the attribute check
6. The merged fixture only actually exercises role="code" and role="mark" as aria-actions fails, but the full set of name-prohibited roles (caption, deletion, emphasis, etc.) behave identically, so  included them all as violations, and set runOnly: 'aria-prohibited-attr' since that's the rule that produces violations for aria-actions




code base:

```

<body>
  <div>
    <!-- aria-prohibited-attr: aria-actions is prohibited on these
         name-prohibited roles (mirrors fail35/fail36 in the merged fixture) -->
    <span role="caption" aria-actions="act1" id="violation1">caption</span>
    <span role="code" aria-actions="act1" id="violation2">code</span>
    <span role="deletion" aria-actions="act1" id="violation3">deletion</span>
    <span role="emphasis" aria-actions="act1" id="violation4">emphasis</span>
    <span role="insertion" aria-actions="act1" id="violation5">insertion</span>
[...]

PR #5209, opened 2026-07-14
test: disable aria-practices listbox-actions

straker:
ARIA practices added [a new listbox actions example](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-actions/) which uses the proposed `aria-actions` attribute and pattern. We don't support that yet so we'll skip it. We already had to do this for a tab pattern they added previously. See https://github.com/dequelabs/axe-core/issues/4584


WilcoFiers (changes requested):
I think we can just allow aria-action instead. There's not much that can go wrong with it.

PR #5214, opened 2026-07-15
fix(color-contrast): fix various stacking context bugs

### LINKED ISSUE #4540: Stacking order bug causes color-contrast to misidentify background
The following should pass:

```html
<style>
  #container {
    position: fixed;
    background-color: #333;
    padding: 0.5em 2em;
  }
  #target {
    filter: contrast(100%);
    color: #eee;
  }
</style>
<div id="container">
  <span id="target"> Hello world </span>
</div>
```

There's some bug in `createGrid()` which results in the stacking order of `#target` to be calculated incorrectly. This comes from the `filter` property. Removing that avoids the issue. This is going to require a very close read of the CSS spec to figure out.

### LINKED ISSUE #4542: Stacking-order z-index: 0 bug
There's a bug in our [getStackLevel](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/dom/create-grid.js#L313) code where we treat a declared `z-index: 0` as having a stacking context but not a z-index. 

For example, the correct stack order of the following code should be `[#parent, #child, body, html]` but instead it comes out to `[#parent, body, html, #child]`

```html
<div id="parent" style="position: relative: z-index: 0">
  <div id="child" style="position: relative; z-index: -1; background: red"></div>
</div>
```

Instead we need to allow `z-index: 0` to be treated as a valid `z-index` value and allow the stacking level to equal 0.  However doing so creates another problem though as we made the `ROOT_LEVEL` z-index equal 0 and then remove [root level elements when we encounter a real stack](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/dom/create-grid.js#L274). (a positioned stacking context that declares a non-auto z-index). 

We'll also need to update our `visuallySort` function as it treats the -1 z-index of a child context as being higher in the visual stack than the parent element, when in reality the parent should be on top.

### LINKED ISSUE #4629: color-contrast false negative with nested combination of stacking contexts
### Product

axe-core

### Product Version

4.10.2

### Latest Version

### Issue Description

We received a customer report recently of a color-contrast false-negative where it reported a `bgOverlap` `incomplete` for a case that it ought to have been able to detect as a violation. This happens due to axe's grid machinery calculating an incorrect stacking order for the elements in question; it incorrectly determines that one of the ancestor container elements is being rendered on top of the leaf descendant element, which is inconsistent with the actual browser paint order (and the order reported by `document.elementsFromPoint`).

The specific customer repro is private, but the test cases below represent minimal repros of the same underlying issue. In the motivating example, this pattern was part of a more complicated interactive grid control.

#### Expectation
The suggested new test cases below should all be satisfied

#### Actual
The test cases are not satisfied; the integration tests `incomplete` with `bgOverlap` and the unit test shows an incorrect stack order calculation:

```
1) should correctly order deeply nested stacking contexts
     dom.getElementStack stack order
     AssertionError: expected [ '2', 'target', '3', '1', 'fixture' ] to deeply equal [ 'target', '3', '2', '1', 'fixture' ]
    at Context.<anonymous> (test/commons/dom/get-element-stack.js:434:14)
```

##### Suggested new case for `test/commons/dom/get-element-stack.js`

```ts
    it('should correctly order nested combinations of positioning and stacking contexts', () => {
      fixture.innerHTML = `
        <div id="1" style="position: relative; z-index: 4; width: 200px; height: 200px;">
          <div id="2" style="position: absolute;">
            <div id="3" style="transform: translate3d(0px, 0px, 0px);">
              <span id="target">text</span>
            </div>
          </div>
        </div>
      `;
      const target = fixture.querySelector('#target');
      const stack = mapToIDs(getElementStack(target));
      // matches browsers' document.elementsFromPoint behavior
      assert.deepEqual(stack, ['target', '3', '2', '1', 'fixture']);
    });
```

##### Suggested new cases for `test/integration/rules/color-contrast/color-contrast.html`

Updating `color-contrast.json` with `#fail13`, `#fail14`, `#pass24`, `#pass25`:

```html
<div style="position: relative; z-index: 1; width: 200px; height: 200px;">
  <div style="position: absolute;">
    <div style="transform: translate3d(0px, 0px, 0px);">
      <span id="fail13" style="background-color: #111; color: #000">low contrast</span>
      <span id="pass24" style="background-color: #fff; color: #000">ok contrast</span>
    </div>
  </div>
</div>

<div style="position: relative; z-index: 1; width: 200px; height: 200px;">
  <div style="position: absolute;">
    <div style="opacity: 0.99;">
      <span id="fail14" style="background-color: #111; color: #000">low contrast</span>
      <span id="pass25" style="background-color: #fff; color: #000">ok contrast</span>
    </div>
  </div>
</div>
```

#### Additional context

We think this is probably the same root cause as #4350 (where it causes a target-size false positive instead of this color-contrast false negative). We started working on this in #4351 but didn't complete before hitting our timebox.

### LINKED ISSUE #5213: Stacking-order bug with multiple pseudo stacks
The following should pass color-contrast but fails:

```html
<body style="background: #00000073">
  <div id="1" style="position: fixed; z-index: 50;">
    <div id="2" style="position: fixed; background: #280071">
      <div id="3" style="position: relative">
        <div id="target" style="transform: matrix(1, 0, 0, 1, 0, 0); color: #fff">text</div>
      </div>
    </div>
  </div>
</body>
```

The issue is that the stack is incorrect and it places the target element below the elements 1 and 2 when it should be placed above it. This causes a bug when we call `reduceElementsBelowFloating` as it sees elements 1 and 2 as being above the target element and removes them from the color contrast check, so we don't get the background color of 2.

straker:
This should correctly fix various stacking context order bugs we have. The main issues was I was misinterpreting the [paint order spec](https://www.w3.org/Style/css2-updates/css2/zindex.html) and conflating "non-positioned" and "positioned" elements as being sorted in different layers. However the spec seems to imply that these should be sorted together. I also didn't realize that the pseudo stack replacement in step #8 applied to both stacking contexts _and_ positioned elements.

Closes: https://github.com/dequelabs/axe-core/issues/4542
Closes: https://github.com/dequelabs/axe-core/issues/4540
Closes: https://github.com/dequelabs/axe-core/issues/4629
Closes: #5213

### DIFF
--- a/lib/commons/dom/create-grid.js
+++ b/lib/commons/dom/create-grid.js
@@ -12,7 +12,6 @@ import { getIntersectionRect } from '../math';
 const ROOT_LEVEL = 0;
 const DEFAULT_LEVEL = 0.1;
 const FLOAT_LEVEL = 0.2;
-const POSITION_LEVEL = 0.3;
 let nodeIndex = 0;
 
 /**
@@ -42,7 +41,7 @@ export default function createGrid(
 
     nodeIndex = 0;
     vNode._stackingOrder = [
-      createStackingContext(ROOT_LEVEL, nodeIndex++, null)
+      createStackingContext(ROOT_LEVEL, true, nodeIndex++, null)
     ];
     rootGrid ??= new Grid();
     addNodeToGrid(rootGrid, vNode);
@@ -264,23 +263,27 @@ function isFlexOrGridContainer(vNode) {
 function createStackingOrder(vNode, parentVNode, treeOrder) {
   const stackingOrder = parentVNode._stackingOrder.slice();
 
-  // if an element creates a stacking context, find the first
-  // true stack (not a "fake" stack created from positioned or
-  // floated elements without a z-index) and create a new stack at
-  // that point (step #5 and step #8)
-  // @see https://www.w3.org/Style/css2-updates/css2/zindex.html
-  if (isStackingContext(vNode, parentVNode)) {
-    const index = stackingOrder.findIndex(({ stackLevel }) =>
-      [ROOT_LEVEL, FLOAT_LEVEL, POSITION_LEVEL].includes(stackLevel)
-    );
-    if (index !== -1) {
-      stackingOrder.splice(index, stackingOrder.length - index);
+  // if an element is positioned or creates a stacking context, find the first
+  // true stack (not a pseudo stack created from positioned or floated elements
+  // without a z-index) and create a new stack at that point (step #5 and step #8)
+  // @see https://www.w3.org/TR/CSS2/zindex.html
+  const isRealStack = isStackingContext(vNode, parentVNode);
+  const isPositioned =
+    vNode.getComputedStylePropertyValue('position') !== 'static';
+
+  if (isRealStack || isPositioned) {
+    const firstPseudo = stackingOrder.findIndex(({ pseudo }) => !!pseudo);
+    if (firstPseudo !== -1) {
+      stackingOrder.splice(firstPseudo);
     }
   }
 
   const stackLevel = getStackLevel(vNode, parentVNode);
   if (stackLevel !== null) {
-    stackingOrder.push(createStackingContext(stackLevel, treeOrder, vNode));
+    // pseudo only if this element does not create a real stacking context
+    stackingOrder.push(
+      createStackingContext(stackLevel, !isRealStack, treeOrder, vNode)
+    );
   }
   return stackingOrder;
 }
@@ -291,12 +294,14 @@ function createStackingOrder(vNode, parentVNode, treeOrder) {
  * @see https://www.w3.org/Style/css2-updates/css2/zindex.html
  * @see https://www.w3.org/Style/css2-updates/css2/visuren.html#layers
  * @param {Number} stackLevel - The stack level of the stacking context
+ * @param {Boolean} pseudo - If the stack is a pseudo stack or a true stack
  * @param {Number} treeOrder - The elements depth-first traversal order
  * @param {VirtualNode} vNode - The virtual node that is the container for the stacking context
  */
-function createStackingContext(stackLevel, treeOrder, vNode) {
+function createStackingContext(stackLevel, pseudo, treeOrder, vNode) {
   return {
     stackLevel,
+    pseudo,
     treeOrder,
     vNode
   };
@@ -314,24 +319,20 @@ function getStackLevel(vNode, parentVNode) {
     return parseInt(zIndex);
   }
 
-  // if a positioned element has z-index: auto or 0 (step #8), or if
-  // a non-positioned floating element (step #5), treat it as its
-  // own stacking context
-  // @see https://www.w3.org/Style/css2-updates/css2/zindex.html
+  // both real and pseudo stacks get sorted at the same layer (step #7)
+  const isRealStack = isStackingContext(vNode, parentVNode);
+  const isPseudoStack =
+    vNode.getComputedStylePropertyValue('position') !== 'static';
 
   // Put positioned elements above floated elements
-  if (vNode.getComputedStylePropertyValue('position') !== 'static') {
-    return POSITION_LEVEL;
+  if (isRealStack || isPseudoStack) {
+    return DEFAULT_LEVEL;
   }
 
   // Put floated elements above z-index: 0
   // (step #5 floating get sorted below step #8 positioned)
   if (vNode.getComputedStylePropertyValue('float') !== 'none') {
     return FLOAT_LEVEL;
-  }
-
-  if (isStackingContext(vNode, parentVNode)) {
-    return DEFAULT_LEVEL;
   }
 
   return null;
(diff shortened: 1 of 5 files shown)


chutchins25 (commented):
Nice work untangling this — the pseudo-stack model reads much more clearly than the old three-level scheme, and the four new test cases are solid coverage.

One thing I'd like to double-check before it lands, plus a small consistency nit — details inline:

1. **Possible regression** — collapsing positioned elements to `DEFAULT_LEVEL` (0.1), which sits below `FLOAT_LEVEL` (0.2), looks like it inverts float-vs-positioned paint order when they share a stacking-context ancestor. It fails on this branch but passes on `develop`; repro + trace inline.
2. **Mixed `@see` URLs** — minor consistency nit, inline.

lib/commons/dom/create-grid.js
```diff
@@ -314,26 +319,20 @@ function getStackLevel(vNode, parentVNode) {
...
+  // step #7 both real and pseudo stacks get sorted at the same layer
+  const isRealStack = isStackingContext(vNode, parentVNode);
+  const isPseudoStack =
+    vNode.getComputedStylePropertyValue('position') !== 'static';
 
-  // Put positioned elements above floated elements
-  if (vNode.getComputedStylePropertyValue('position') !== 'static') {
-    return POSITION_LEVEL;
+  if (isRealStack || isPseudoStack) {
+    return DEFAULT_LEVEL;
```
chutchins25, line 334: Is mapping positioned elements to `DEFAULT_LEVEL` (0.1) intended here, given that's *below* `FLOAT_LEVEL` (0.2)? I think it may be a regression.

Per CSS2 painting order, a positioned element with `z-index: auto` (step #6) paints above a non-positioned float (step #4). At the document root the positioned element splices out the root pseudo-stack, so its index-0 level (0.1) still beats a float sibling's root (0) and it sorts on top. But once a stacking-context ancestor has already replaced the root with a non-`pseudo` entry at 0.1, the positioned child has no `pseudo` entry left to splice — so the raw `0.1 < 0.2` comparison wins and the float paints on top.

Minimal repro:

```html
<div id="ctx" style="opacity: 0.99">
  <div id="float" style="float: left; width: 80px; height: 80px"></div>
  <div id="target" style="position: relative; width: 80px; height: 80px; margin-top: -40px"></div>
</div>
```

`getElementStack(target)` returns `['float', 'target', 'ctx', 'fixture']` on this branch, but `['target', 'float', 'ctx', 'fixture']` on `develop` (target on top, per spec). The same pair at the document root — no `opacity` wrapper — sorts correctly on both. For color-contrast this could select the wrong background element behind overlapping text.

Worth confirming — and maybe a test for the float-inside-a-stacking-context case, since the new nested test uses transform/absolute rather than a float.

lib/commons/dom/create-grid.js
```diff
@@ -264,23 +263,27 @@ function isFlexOrGridContainer(vNode) {
...
-  if (isStackingContext(vNode, parentVNode)) {
-    const index = stackingOrder.findIndex(({ stackLevel }) =>
-      [ROOT_LEVEL, FLOAT_LEVEL, POSITION_LEVEL].includes(stackLevel)
-    );
-    if (index !== -1) {
-      stackingOrder.splice(index, stackingOrder.length - index);
+  // if an element is positioned or creates a stacking context, find the first
+  // true stack (not a pseudo stack created from positioned or floated elements
+  // without a z-index) and create a new stack at that point (step #5 and step #8)
+  // @see https://www.w3.org/TR/CSS2/zindex.html
```
chutchins25, line 269: Not a blocker, just a consistency question — this updates the `@see` to the `TR/CSS2` URL, but `createStackingContext`'s JSDoc (lines 294–295) and other comments in the file still use the old `www.w3.org/Style/css2-updates/css2/…` form. Worth updating those in the same pass so the file points at the spec one way.

lib/commons/dom/create-grid.js
chutchins25: Retracting this — false alarm on my part. This PR swaps the constants (`FLOAT_LEVEL` is now `0.1`, `DEFAULT_LEVEL` `0.2`), so a positioned element maps to `DEFAULT_LEVEL` (`0.2`) and sorts *above* a float (`0.1`) — the spec-correct behavior. I re-ran my repro against this branch and `getElementStack(target)` returns `['target', 'float', 'ctx', 'fixture']` (target on top), not the inverted order I described. No regression here.

lib/commons/dom/create-grid.js
chutchins25: Correction — `create-grid.js` is consistent now; lines 294–295 were updated to `TR/CSS2` in this PR along with the rest of the file. The only remaining `css2-updates` URLs are in the sibling `visually-sort.js` (lines 6–7 and 105). Not a blocker — worth a small follow-up if we want the whole `dom/` stacking subsystem pointing at the spec one way.

chutchins25 (approved):
Looks good to me after the updates you made. The Firefox browser fix was very interesting.

WilcoFiers (approved):
LGTM.

PR #5219, opened 2026-07-20
test: setup the virtual tree using root rather than fixture

straker:
The tests only set up the virtual tree from the `fixture` down. This causes problems in our tests when code walks up the DOM tree (not the virtual tree) and tries to get virtual nodes from the walk. This causes tests to fail when it wouldn't fail in an actual run (as doing `axe.run` setups the virtual tree for the entire page).

This updates the testutils functions to set up the virtual tree using the entire DOM to parity an actual run.

### DIFF
--- a/test/commons/dom/create-grid.js
+++ b/test/commons/dom/create-grid.js
@@ -30,7 +30,7 @@ describe('create-grid', () => {
 
     createGrid();
     assert.isDefined(fixture._grid);
-    assert.equal(fixture._grid, fixture.children[0]._grid);
+    assert.isTrue(fixture._grid === fixture.children[0]._grid);
   });
 
   it('adds elements to the correct cell in the grid', () => {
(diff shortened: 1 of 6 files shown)


test/commons/dom/create-grid.js
```diff
@@ -30,7 +30,7 @@ describe('create-grid', () => {
 
     createGrid();
     assert.isDefined(fixture._grid);
-    assert.equal(fixture._grid, fixture.children[0]._grid);
+    assert.isTrue(fixture._grid === fixture.children[0]._grid);
```
straker, line 33: When these two aren't `equal`, the code freezes as it tries to output the stringified DOM nodes which crashes the browser.

WilcoFiers (approved):

chutchins25 (approved):
Looks good to me.

Issue #4618, opened 2024-10-17
Consistently handle case insensitivity


WilcoFiers:
Axe has various places where it is treating things as though they should be case sensitive when they shouldn't be. This for example happens in (parts of) selectors, and when using attr methods to get at roles. We should go over the code base to figure out where all of these different issues are.

PR #5224, opened 2026-07-20
feat(standards/ariaAttrs): add caseInsensitive property for attributes

straker:
Add the `caseInsensitive` property to the aria attrs standard. This is used to automatically lowercase the value for `commons.aria.getAriaValue` instead of having to manually pass `{ lowercase: true }` every time you wanted to access one of the attributes. This updates `aria-expanded`, `aria-hidden`, `aria-haspopup`, `aria-selected`, `aria-invalid`, `aria-disabled`, and `ara-live` to be case insensitive (which we have tested in prior prs).

Additionally, as a `chore` style change this also updates `getAriaValue` to not return `null` anymore but instead `{ value: null, source: null }` so that we don't have to conditionally access the `value` property of the return.


WilcoFiers (dismissed):
Thanks for tackling this — the `caseInsensitive` flag is a nice cleanup. A few things to address before merge, mostly around consistency and finishing the replacement.

### Important

**1. Leftover `{ lowercase: true }` for the now-`caseInsensitive` attrs.** Some call sites weren't updated:
- [`lib/rules/aria-hidden-focus-matches.js:10`](https://github.com/dequelabs/axe-core/blob/819add85fe9943a2f8dd9bccfa1a94155cec7e03/lib/rules/aria-hidden-focus-matches.js#L10) still passes `{ lowercase: true }` for `aria-hidden` (this file isn't in the PR at all).
- [`lib/checks/aria/aria-valid-attr-value-evaluate.js:54`](https://github.com/dequelabs/axe-core/blob/819add85fe9943a2f8dd9bccfa1a94155cec7e03/lib/checks/aria/aria-valid-attr-value-evaluate.js#L54) and [`:78`](https://github.com/dequelabs/axe-core/blob/819add85fe9943a2f8dd9bccfa1a94155cec7e03/lib/checks/aria/aria-valid-attr-value-evaluate.js#L78) still pass it for `aria-expanded`, which is inconsistent with `aria-required-attr-evaluate.js` where the PR removed the same pattern.

These are redundant now and undercut the stated goal of the PR. A repo-wide grep for `lowercase: true` would confirm none of the four flagged attrs still pass it.

**2. Doc/test typo `caseInsenstive`.** [`doc/standards-object.md:54`](https://github.com/dequelabs/axe-core/blob/819add85fe9943a2f8dd9bccfa1a94155cec7e03/doc/standards-object.md#L54) and the test name at [`test/commons/aria/get-aria-value.js:128`](https://github.com/dequelabs/axe-core/blob/819add85fe9943a2f8dd9bccfa1a94155cec7e03/test/commons/aria/get-aria-value.js#L128) spell it `caseInsenstive` — the actual property is `caseInsensitive`, so the docs describe a property that doesn't exist.

**3. `aria-selected` is treated case-insensitively but not flagged in the standard.** [`aria-valid-attr-value-evaluate.js:57`](https://github.com/dequelabs/axe-core/blob/819add85fe9943a2f8dd9bccfa1a94155cec7e03/lib/checks/aria/aria-valid-attr-value-evaluate.js#L57) lowercases `aria-selected` via `{ lowercase: true }`, but [`aria-selected` in the standard](https://github.com/dequelabs/axe-core/blob/819add85fe9943a2f8dd9bccfa1a94155cec7e03/lib/standards/aria-attrs.js#L249) is not marked `caseInsensitive`. So axe already treats it case-insensitively at this one call site while the standard says it's case-sensitive. WAI-ARIA treats enumerated token values as case-insensitive generally, so as part of this work we should test the remaining token props to confirm whether they're actually still case-sensitive, and mark them accordingly (or intentionally leave them and document why — see below). At minimum `aria-selected` should be reconciled so behavior doesn't depend on the call site.

### Suggestions

**4. Document why some token props are left case-sensitive.** Since the spec treats enumerated token values as case-insensitive but this PR flags only four attributes, add a note (in `lib/standards/aria-attrs.js` and/or the doc) explaining why the remaining token attributes are intentionally left case-sensitive in Axe. **Where we don't have test data yet, we'll need to run the tests**.

**5. Redundant optional chaining.** Now that `getAriaValue` always returns an object, remaining `?.value` calls (e.g. [`aria-valid-attr-value-evaluate.js:45`](https://github.com/dequelabs/axe-core/blob/819add85fe9943a2f8dd9bccfa1a94155cec7e03/lib/checks/aria/aria-valid-attr-value-evaluate.js#L45)) are dead defensiveness and mix styles within the file. Low priority.

test/commons/text/form-control-value.js
```diff
@@ -357,7 +357,6 @@ describe('text.formControlValue', () => {
       const target = queryFixture(html`
         <div id="target" role="listbox">
           <div role="option" aria-selected="false">foo</div>
-          <div role="option" aria-selected="TRUE">bar</div>
```
straker, line 360: Tested this and it's case insenstive

lib/standards/aria-attrs.js
WilcoFiers, line 21: Can you document why this isn't case insensitive?

WilcoFiers, line 46: Can you document why this isn't case insensitive?

WilcoFiers, line 74: Can you document why this isn't case insensitive?

lib/standards/aria-attrs.js
```diff
@@ -92,7 +92,8 @@ const ariaAttrs = {
   'aria-disabled': {
     type: 'boolean',
     prop: 'ariaDisabled',
-    global: true
+    global: true,
+    caseInsensitive: true
   },
   'aria-dropeffect': {
     type: 'nmtokens',
```
WilcoFiers, line 107: Can you document why this isn't case insensitive?

lib/standards/aria-attrs.js
WilcoFiers, line 133: Can you document why this isn't case insensitive?

WilcoFiers, line 209: Can you document why this isn't case insensitive?

WilcoFiers, line 291: Can you document why this isn't case insensitive?

WilcoFiers (dismissed):
I can live with leaving testing of those lesser used props for another time, but in that case please open an issue.

lib/standards/aria-attrs.js
straker: I'm not sure how to test this. The value is never announced by screen readers as far as I can tell. Doing `<input type="text" role="combobox">` and `<input type="text" role="combobox" aria-autocomplete="list">` produces the same output in all ATs I tested.

lib/standards/aria-attrs.js
straker: This attribute is deprecated and will report needs review if used, do we care to check case sensitivity for it?

lib/standards/aria-attrs.js
straker: Also deprecated

chutchins25 (commented):
Nice work — `58cb8a08` is a good deal more than was asked for. 20 of the 23 token/boolean attrs now carry an explicit `caseInsensitive` value, and the per-AT rationale comments on the nine `false` ones are genuinely useful to have in the standard.

A few things I checked that seem worth recording:

- Every removed `{ lowercase: true }` maps to an attribute now marked `caseInsensitive: true` — all 8 call sites (`aria-invalid`, `aria-hidden` ×3, `aria-expanded` ×3, `aria-haspopup`, `aria-selected`, `aria-live`, `aria-disabled`). No silent loss of lowercasing.
- The `getAriaValue` return-shape change isn't a breaking change: `lib/commons/aria/get-aria-value.js` is absent from `v4.11.4`, `v4.12.0` and `v4.12.1`, and `getAriaValue` appears nowhere in `doc/`. It's never shipped, so there's no public contract to break.
- Items 1, 3 and 5 from the first round all check out on this head — zero `lowercase: true` left in `lib/` (the three remaining hits are tests of the option itself on `aria-label`), `aria-selected` is now flagged in the standard, and zero `?.value` left in any changed lib file.

One process thing: @WilcoFiers asked for an issue if the lesser-used props were left untested, and I couldn't find one. #4618 is the only open case-sensitivity ticket and it's a broad 2024 "go over the code base" item that doesn't name these attributes, and there's no `Closes` line on this PR. Worth filing before merge so that condition is actually met.

Plus a question inline about the `aria-selected` test that lost its uppercase case, and a note on `aria-attrs.js` about the three unresolved attrs.

lib/standards/aria-attrs.js
```diff
@@ -14,7 +14,8 @@ const ariaAttrs = {
   'aria-atomic': {
     type: 'boolean',
     prop: 'ariaAtomic',
-    global: true
+    global: true,
+    caseInsensitive: true
   },
   'aria-autocomplete': {
```
chutchins25, line 20: Following on from your comments last round — after `58cb8a08`, `caseInsensitive` effectively has three states, and `doc/standards-object.md:54` only describes two:

- `true` — 11 attrs, AT-verified case-insensitive
- `false` — 9 attrs, AT-verified case-*sensitive*, each with a rationale comment
- absent — 3 attrs (`aria-autocomplete`, `aria-dropeffect`, `aria-grabbed`), never evaluated

The doc reads *"boolean(optional, default `false`). If the value of the attribute is case insensitive"*, under which absent and `false` are equivalent. So the nine deliberate `false` values lose the distinction that makes them worth recording, and someone reading `aria-autocomplete` here can't tell it was never tested rather than tested-and-case-sensitive. Your answers on those three are in the review threads but not in the code.

Would a short `// not yet tested, see #NNNN` on each of the three — or a line in `standards-object.md` spelling out that an explicit `false` means verified — be worth it?

Also minor: the "does not lowercase" test at `get-aria-value.js:100-125` is now anchored partly on `aria-grabbed`, one of the three unresolved ones — and it already had to be repointed once in `dccd3d32` when `58cb8a08` promoted `aria-modal` to `true`.

test/commons/text/form-control-value.js
```diff
@@ -357,7 +357,6 @@ describe('text.formControlValue', () => {
       const target = queryFixture(html`
         <div id="target" role="listbox">
           <div role="option" aria-selected="false">foo</div>
```
chutchins25, line 359: Dropping the `aria-selected="TRUE"` case from this test makes sense — with `aria-selected` now `caseInsensitive: true` it resolves to `'true'`, so "bar" becomes selected and this assertion would flip from `''` to `'bar'`.

The part I'd flag is that nothing replaced it. As far as I can tell it was the only test touching uppercase `aria-selected`, so the change to `ariaListboxValue` (`form-control-value.js:150-154`) now has no coverage at this call site. Your note that you'd tested it is on the thread attached to the deleted line, so the verification exists but isn't captured in code. Worth adding the positive counterpart?

```js
it('returns the option when aria-selected is not lowercase', () => {
  const target = queryFixture(html`
    <div id="target" role="listbox">
      <div role="option" aria-selected="false">foo</div>
      <div role="option" aria-selected="TRUE">bar</div>
    </div>
  `);
  assert.equal(ariaListboxValue(target), 'bar');
});
```

(Plain block rather than a suggestion since it needs to land after this test's closing brace.)

WilcoFiers (approved):

PR #5229, opened 2026-07-21
test(hasAriaValue): reenable skipped test due to firefox crash

### LINKED ISSUE #5139: ElementInternals: accessing empty internals idrefs property crashes Firefox

straker:
Tested in latest Firefox (153.0) and this looks like it doesn't crash the browser anymore. The other non-crash issues still exist (logging the warning, `null` for empty `ariaLabelledByElements `), but we should be able to reenable the test now.

Closes: https://github.com/dequelabs/axe-core/issues/5139

### DIFF
--- a/test/commons/aria/has-aria-value.js
+++ b/test/commons/aria/has-aria-value.js
@@ -37,21 +37,15 @@ describe('aria.hasAriaValue', () => {
     assert.isTrue(hasAriaValue(vNode, 'aria-label'));
   });
 
-  // TODO: remove when we resolve https://github.com/dequelabs/axe-core/issues/5139
-  // accessing empty idrefs element internal values crashes firefox
-  // @see https://bugzilla.mozilla.org/show_bug.cgi?id=2045887
-  (navigator.userAgent.indexOf('Firefox') > 0 ? it.skip : it)(
-    'returns true if elementInternals idrefs is empty',
-    () => {
-      const vNode = queryFixture(
-        html`<testutils-element id="target"></testutils-element>`
-      );
-      const node = vNode.actualNode;
-      node._internals.ariaLabelledByElements = [];
+  it('returns true if elementInternals idrefs is empty', () => {
+    const vNode = queryFixture(
+      html`<testutils-element id="target"></testutils-element>`
+    );
+    const node = vNode.actualNode;
+    node._internals.ariaLabelledByElements = [];
 
-      assert.isTrue(hasAriaValue(vNode, 'aria-labelledby'));
-    }
-  );
+    assert.isTrue(hasAriaValue(vNode, 'aria-labelledby'));
+  });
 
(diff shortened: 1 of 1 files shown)


Garbee (approved):

test/commons/aria/has-aria-value.js
```diff
@@ -37,21 +37,15 @@ describe('aria.hasAriaValue', () => {
...
-    }
-  );
+  it('returns true if elementInternals idrefs is empty', () => {
+    const vNode = queryFixture(
+      html`<testutils-element id="target"></testutils-element>`
+    );
+    const node = vNode.actualNode;
+    node._internals.ariaLabelledByElements = [];
+
+    assert.isTrue(hasAriaValue(vNode, 'aria-labelledby'));
```
chutchins25, line 47: The description mentions `null` for empty `ariaLabelledByElements` still happening — but that looks like the exact path this assertion covers? `vNode.elementInternals` hands back the live internals object, and `hasInternalValue` is a `!== null` check, so a `null` here would fail this test rather than pass it. Since `test_firefox` is green, is the null behavior fixed in 153 too, or does it show up somewhere else (`getAriaValue`, or a never-set prop)?

Asking mostly for issue hygiene: #5139 is the only place those remaining quirks are written down and this PR closes it. Worth a follow-up issue if anything is still outstanding?

Related, and the reason I'd push a little on that: the crash isn't test-only. It's the live `vNode.elementInternals[prop]` read in `hasInternalValue`, which checks reach on real pages via the `_internals` protocol — and a browser crash isn't catchable by the try/catch in `walk-tree.js`. Against README's stated Firefox 38+ support that seems worth keeping written down somewhere after #5139 closes.

chutchins25 (commented):
Nice work — good to see this one clear out. I confirmed the workaround was the only one: zero remaining references to `5139` or `2045887` anywhere in the head, and no other test in `test/` carries a Firefox guard for it. `Closes #5139` matches what you asked for on the issue, too.

One question inline about the remaining Firefox quirks.

Also worth a rebase before merge — the branch is 22 commits behind develop with a base predating the pnpm conversion (#5264), so the green checks ran on the old npm toolchain. No conflicts, just means the checkmark is a bit stale.

test/commons/aria/has-aria-value.js
straker: The `null` refers to when it's set via the HTML attribute but left empty `<div aria-labelledby>` and you access the idref property `div.ariaLabelledByElements`. It doesn't affect this particular test.

WilcoFiers (approved):

PR #5237, opened 2026-07-22
chore: remove packages

straker:
These should not have been added and were used only to create the [codemode to update our test files](https://github.com/dequelabs/axe-core/commit/57cb1efb8f81d46158551532d9dcfc9c7aa32ab1). Removing both `recast` and `parse5` which we don't use, and resetting our `@types/node` to latest node 4 version in order to maintain backwards compatibility with our [`engines` declaration](https://github.com/dequelabs/axe-core/blob/develop/package.json#L6-L8) and [dependabot config](https://github.com/dequelabs/axe-core/blob/develop/.github/dependabot.yml#L49-L51).


WilcoFiers (approved):

dylanb:
@straker how did this get merged without a security review?

PR #5238, opened 2026-07-22
feat(aria-roles): add sectionheader and sectionfooter roles

### LINKED ISSUE #4734: sectionhead role should be sectionheader
Spotted this when I was going over the standards. Axe has `sectionhead` listed as a role. This is a typo, it should be [`sectionheader`](https://w3c.github.io/aria/#sectionheader).

chutchins25:
Adds the new ARIA 1.3 `sectionheader` and `sectionfooter` roles to axe-core's standards data.

## What & why

Per [w3c/aria#1931](https://github.com/w3c/aria/pull/1931) (merged into the ARIA spec `main`), `sectionheader` and `sectionfooter` are new **concrete structural roles** with superclass `section`. axe-core did not recognize them, so `role="sectionheader"` / `role="sectionfooter"` were flagged as invalid roles.

Definitions were grounded directly against the spec source (`w3c/aria` `main`):
- `roleInfo.js` — `parentRoles: ["section"]`, no required / prohibited / local (non-global) properties.
- `index.html` characteristics table — **Is Abstract:** no (concrete), **Name From:** author (so no `nameFromContent`), **Accessible Name Required:** no.

They are modeled on the existing `blockquote` precedent (`type: 'structure', superclassRole: ['section']`).

## Note on scope — bundling `sectionfooter`

The issue (#4734) names only `sectionheader`, but the spec introduced `sectionheader` and `sectionfooter` together and there is no separate `sectionfooter` issue. They are bundled here as one cohesive "ARIA 1.3 section header/footer roles" change.

## Out of scope / follow-up

The updated HTML-AAM implicit-role mapping (`<header>` / `<footer>` scoped inside sectioning content → implicit `sectionheader` / `sectionfooter`) is **not** included here — it has a much larger blast radius (landmark/region computations, `aria-allowed-role`) and will be tracked as a separate issue.

## Tests
- `aria-roles` integration HTML/JSON — pass cases for both roles.
- `aria-roles` virtual-rule — pass cases for both roles.
- Verified `abstractrole` / `fallbackrole` / `unsupportedrole` / `aria-allowed-role` check suites still pass.

Closes #4734


test/integration/virtual-rules/aria-roles.js
```diff
@@ -73,4 +73,30 @@ describe('aria-roles virtual-rule', () => {
...
+      nodeName: 'div',
+      attributes: {
+        role: 'sectionfooter'
+      }
+    });
+
+    assert.lengthOf(results.passes, 1);
+    assert.lengthOf(results.violations, 0);
+    assert.lengthOf(results.incomplete, 0);
+  });
```
straker: These aren't needed

straker (dismissed):
Looks good, just the one thing below.

test/integration/virtual-rules/aria-roles.js
chutchins25: Removed — role validity is already covered by the `aria-roles` integration tests.

straker (approved):

Garbee (dismissed):
Talking about the commit reference internally before merging.

Issue #3911, opened 2023-02-13
ARIA in HTML update for summary element allowances


scottaohara:
ARIA in HTML is updating to clarify the ARIA attribute allowances for the summary element:
- https://github.com/w3c/html-aria/pull/435
- [preview spec change: summary element allowances](https://pr-preview.s3.amazonaws.com/w3c/html-aria/pull/435.html#el-summary)

The summary currently allows no role, and any global aria-* attributes, and any aria-* attributes applicable to the button role.

The update changes this to indicate that no role is allowed so long as the summary element serves as the summary for its parent details element.  However, Any role will be allowed on the element if it does not meet this requirement.

For instance, the following would continue to disallow any role from developers, as applying roles can cause AT to no longer correctly expose the current state of the disclosure widget:

```
<details>
  <summary>no role allowed</summary>
  ...
</details>
```

But the following examples demonstrate where any role may be applied to the summary element, as it does not meet HTML's conditions to be considered the summary for its parent details.

```
<details>
  <summary>
  	no role allowed on this summary, as it is the first instance of the summary
  	element within the details element, and thus browsers treat this as the 
  	summary / trigger to open/close the disclosure widget.
  </summary>

  <summary role=paragraph>
   any role is allowed on this summary element because it is ignored as a summary
   element by browsers, and instead returned as a generic element.
  </summary>
</details>

<summary role=group>
  this summary element also allows any role because it is not a child to a 
  details element, and thus is exposed as a generic by browsers.
</summary>
```

As the summary element is not consistently exposed as a button by all browsers, AND not all of the allowed ARIA attributes for button element even work on this element, the allowed attributes have been modified to reflect this.

If the summary element acts as the summary for its parent details, then global aria attributes and the **aria-haspopup and aria-disabled attribute may be specified**.  Note that aria-expanded and aria-pressed are the attributes no longer allowed, and per testing, these attributes were not respected anyway.

If the summary element is not the summary for its parent details, then any global aria attributes and any attribute applicable to the allowed specified role may be used.

Additional examples and test cases can be viewed here:
- https://codepen.io/scottohara/pen/ZEjZPoR
- https://w3c.github.io/html-aria/tests/summary-allowances.html

straker:
[...]

PR #5240, opened 2026-07-22
fix(aria-allowed-role): restrict figure roles with child figcaption

### LINKED ISSUE #3443: Update role allowance for figure with child figcaption
[ARIA in HTML](https://github.com/w3c/html-aria/pull/415) is updating the role allowances for a `figure` element with a child `figcaption` to allow for authors to use `doc-example`.

`doc-example` is a dpub variant of the `figure` role. This role being the same as `figure` but with the opportunity of being exposed as an "example" for its role description, needs to be an allowed role for the figure element - [further discussion on this topic via the original dpub aria issue](https://github.com/w3c/dpub-aria/issues/42#issuecomment-1092558918).

chutchins25:
Restricts the roles allowed on a `<figure>` element when it has a child `<figcaption>`, per the ARIA in HTML update.

## What & why

Per [ARIA in HTML](https://www.w3.org/TR/html-aria/) (`#el-figure`, [w3c/html-aria#415](https://github.com/w3c/html-aria/pull/415)):

- A `figure` **with** a valid `figcaption` descendant allows **no role other than `figure`**, plus the DPub role **`doc-example`**.
- A `figure` **without** a `figcaption` allows **any role** (though `figure` itself is NOT RECOMMENDED) — unchanged.

axe previously modeled `figure` with a flat `allowedRoles: true` and an inline comment admitting it couldn't match the figcaption case. This replaces that with a `variant` using a `condition` matcher.

## Implementation notes

- The implicit `figure` role still passes via the existing `allowImplicit` short-circuit, so the with-figcaption branch only needs `allowedRoles: ['doc-example']`.
- The matcher checks **direct children** (`vNode.children?.some(...)`), not a descendant walk: HTML's content model only permits `figcaption` as the first/last child of `figure`, and a direct-child check avoids a nested-`<figure>` false match (an inner figure's figcaption wrongly restricting the outer figure).

## Behavior change

A `<figure>` with a `<figcaption>` and an explicit role other than `figure`/`doc-example` (including `role="none"`/`role="presentation"`) now flags on `aria-allowed-role` (minor impact). This is spec-correct but will surface new findings for existing users.

## Tests
- `get-element-spec` unit — variant resolution with/without figcaption.
- `is-aria-role-allowed-on-element` and `aria-allowed-role` check units — incl. an open **Shadow DOM** case.
- `aria-allowed-role` virtual-rule — figcaption-child cases (built with `SerialVirtualNode`) plus the no-figcaption default.
- `aria-allowed-role` integration HTML/JSON — pass (`doc-example`, implicit `figure`) and violation (`group`) cases; existing no-figcaption pass retained as a regression anchor.

Closes #3443


lib/standards/html-elms.js
```diff
@@ -248,10 +248,25 @@ const htmlElms = {
     allowedRoles: ['group', 'none', 'presentation']
   },
   figure: {
+    variant: {
+      // A figure with a child figcaption is restricted to its implicit
+      // role (figure) and the DPub role doc-example.
+      // See https://github.com/w3c/html-aria/pull/415
+      figcaption: {
+        matches: {
+          condition: vNode =>
```
straker: So this is almost what we want to do, but not make it an inline function. We want the standard objects to be serializable so no functions should be added. Instead we want to do something similar to our proposal in https://github.com/dequelabs/axe-core/issues/3911#issuecomment-5036896185. So we'll want a new matches function name that can be triggered via the matches object that checks the condition we want.

lib/standards/html-elms.js
```diff
@@ -248,10 +248,25 @@ const htmlElms = {
...
+            !!vNode.children?.some(
+              child => child.props.nodeName === 'figcaption'
+            )
+        },
+        allowedRoles: ['doc-example']
```
straker, line 263: I'm not sure if this is correct. https://pr-preview.s3.amazonaws.com/w3c/html-aria/pull/415.html is a bit ambiguous here in whether this meant only allowed or additionally allowed. Since `figure` already allows any role I don't see why only allowing `doc-example` would be restricted just because it has a figcaption child. If we look at the `div` row in the table it shows an example of where use is restricted by saying "only allowed," which the figure callout does not. We should get @WilcoFiers interpretation of this. If this is only saying additionally allowed I think this is already covered by `allowedRoles: true` and we don't really have to do anything for this pr other than add an integration test to ensure it's covered.

straker (changes requested):

lib/standards/html-elms.js
chutchins25: Agreed on keeping the standards serializable. I'll move to a named matches function — though as noted on #5242, `firstOfType` alone isn't sufficient (it also matches a standalone summary), so I'll align both once we settle the matcher shape. Also gated on the `doc-example` interpretation below.

lib/standards/html-elms.js
chutchins25: Agreed this needs @WilcoFiers' read. The PR-preview table doesn't use the "only allowed" phrasing it uses for the `div` row, so "additionally allowed" is plausible — in which case `allowedRoles: true` already covers it and this reduces to just adding an integration test. Holding the code change until we get his interpretation.

lib/standards/html-elms.js
chutchins25: Done — moved to the serializable `hasChild: 'figcaption'` matcher from #5262 (no more inline function). This surfaced that `hasChild` threw on a `SerialVirtualNode` with undefined `children` (hit by the figure virtual-rule test), so I added a guard (`vNode.children?.some(...)`) plus a unit test — kept as a separate commit in case you'd rather split it out.

lib/standards/html-elms.js
chutchins25: @WilcoFiers — context for when you get to this. The open question: does a `<figure>` with a child `<figcaption>` restrict its roles to **only** `doc-example` (plus the implicit `figure` role), or is `doc-example` **additionally** allowed on top of figure's existing `allowedRoles: true`?

- The [html-aria PR](https://pr-preview.s3.amazonaws.com/w3c/html-aria/pull/415.html) figure callout does not use the "only allowed" phrasing that the `div` row uses, which suggests "additionally allowed."
- If **additionally allowed**: `figure` already allows any role, so no restriction is needed and this PR reduces to just an integration test.
- If **only allowed**: the current restriction stands.

The code currently implements the **only-allowed** interpretation (restricted to `doc-example`), now via the serializable `hasChild: 'figcaption'` matcher, as a placeholder. Once you confirm, we'll keep it or drop the restriction. Flagging so you have the full picture — no rush.

lib/commons/matches/has-child.js
```diff
@@ -17,5 +17,7 @@ import { matches } from '../../core/utils';
  * @returns {Boolean}
  */
 export default function hasChild(vNode, selector) {
-  return vNode.children.some(node => matches(node, selector));
+  // vNode.children can be undefined for a SerialVirtualNode (e.g. a node
+  // passed to runVirtualRule without children); treat that as no match.
+  return !!vNode.children?.some(node => matches(node, selector));
```
straker: Wilco and I discussed this and we decided to not guard against child lookups so we can treat this as always having children

lib/standards/html-elms.js
straker: From https://www.w3.org/TR/html-aria/#el-figure, Wilco and I interpret that to mean `doc-example` is always allowed. So if there is a figcaption it's the only allowed role, otherwise any role is allowed. So this code is correct for that interpretation.

lib/standards/html-elms.js
```diff
@@ -255,10 +255,22 @@ const htmlElms = {
     allowedRoles: ['group', 'none', 'presentation']
   },
   figure: {
+    variant: {
+      // A figure with a child figcaption is restricted to its implicit
+      // role (figure) and the DPub role doc-example.
+      // See https://github.com/w3c/html-aria/pull/415
```
straker: No comment needed as the spec defines all this

lib/standards/html-elms.js
```diff
@@ -255,10 +255,22 @@ const htmlElms = {
...
+      figcaption: {
+        matches: {
+          hasChild: 'figcaption'
+        },
+        allowedRoles: ['doc-example']
+      },
+      // Otherwise any role is allowed (figure itself is NOT RECOMMENDED)
```
straker: Same no comment needed

test/checks/aria/aria-allowed-role.js
```diff
@@ -426,4 +426,40 @@ describe('aria-allowed-role', () => {
...
+      axe.testUtils
+        .getCheckEvaluate('aria-allowed-role')
+        .call(checkContext, null, null, vNode)
+    );
+  });
+
+  it('returns false for a figure with a child figcaption and an unallowed role in shadow DOM', () => {
+    const vNode = axe.testUtils.queryShadowFixture(
+      '<div id="shadow"></div>',
+      '<figure id="target" role="group"><figcaption>caption</figcaption></figure>'
```
straker: Test not needed as this does not cross shadow DOM boundries

test/checks/aria/aria-allowed-role.js
```diff
@@ -426,4 +426,40 @@ describe('aria-allowed-role', () => {
         .call(checkContext, null, null, vNode)
     );
   });
+
+  it('returns false for a figure with a child figcaption and an unallowed role', () => {
```
straker: These aren't needed as the integration tests can test these cases

test/commons/matches/has-child.js
```diff
@@ -43,4 +43,11 @@ describe('matches.hasChild', () => {
     serialNode.children = [childNode];
     assert.isTrue(hasChild(serialNode, 'span'));
   });
+
+  it('returns false for a SerialVirtualNode without children', () => {
```
straker: Remove this as we expect children

test/commons/standards/get-element-spec.js
```diff
@@ -39,6 +39,18 @@ describe('standards.getElementSpec', () => {
...
   });
 
+  it('should restrict figure allowedRoles when it has a child figcaption', () => {
+    const vNode = queryFixture(
+      '<figure id="target"><figcaption>caption</figcaption></figure>'
+    );
+    assert.deepEqual(getElementSpec(vNode).allowedRoles, ['doc-example']);
+  });
+
+  it('should allow any role on a figure without a child figcaption', () => {
```
straker, line 49: Add test for a nested figcaption to ensure it doesn't pass (`<figure><figure><figcaption>`) that returns `allowedRoles: true`

straker (changes requested):

test/integration/virtual-rules/aria-allowed-role.js
```diff
@@ -90,4 +90,57 @@ describe('aria-allowed-role virtual-rule', () => {
     assert.lengthOf(results.violations, 0);
     assert.lengthOf(results.incomplete, 1);
   });
+
+  it('should pass doc-example on a figure with a child figcaption', () => {
```
straker: Specific tests shouldn't be added here as they are covered in the integration test

lib/standards/html-elms.js
chutchins25: Thanks — with the "only allowed" interpretation confirmed, the restriction stays (figcaption ⇒ `doc-example` only). Code is unchanged.

lib/commons/matches/has-child.js
chutchins25: Done — dropped the guard; `hasChild` now assumes children are present.

lib/standards/html-elms.js
chutchins25: Done — removed.

lib/standards/html-elms.js
chutchins25: Done — removed.

test/checks/aria/aria-allowed-role.js
chutchins25: Removed.

test/checks/aria/aria-allowed-role.js
chutchins25: Removed — the integration tests cover these.

test/commons/matches/has-child.js
chutchins25: Removed.

test/commons/standards/get-element-spec.js
chutchins25: Added — get-element-spec now asserts `<figure><figure><figcaption>` still returns `allowedRoles: true`.

test/integration/virtual-rules/aria-allowed-role.js
chutchins25: Removed — the integration tests cover these.

straker (approved):

Issue #5241, opened 2026-07-22
ARIA in HTML: disallow aria-expanded/aria-pressed on a summary that is a details summary


chutchins25:
Follow-up to #3911, which updated the allowed **roles** for the `summary` element but intentionally left the allowed **`aria-*` attribute** side out of scope.

## Background

Per [ARIA in HTML](https://www.w3.org/TR/html-aria/) (`#el-summary`), when a `<summary>` is the "summary for its parent `details` element" (i.e. the first `summary` child of a `details`), the only allowed ARIA attributes are:

> Global `aria-*` attributes, `aria-disabled`, and `aria-haspopup`.

This means **`aria-expanded` and `aria-pressed` are not allowed** on a details-summary (browsers don't honor them there anyway).

## Why it wasn't done in #3911

axe's `aria-allowed-attr` rule derives allowed attributes from the element's **resolved role**, not from `lib/standards/html-elms.js`. A `summary`'s implicit role is `button`, and `button`'s `allowedAttrs` includes `aria-expanded`/`aria-pressed`, so axe currently allows them on a summary.

Disallowing these attributes **only on a details-summary** — without affecting real `<button>` elements — requires either:

1. a new **element-level disallowed-attribute mechanism** wired into `aria-allowed-attr-evaluate` (axe has no such mechanism today), or
2. changing the summary's implicit/exposed role so it is not `button` when it is the disclosure trigger — which has a much wider blast radius (accessible-name computation, `aria-required-attr`, etc.).

Both are larger than a standards-data tweak and warrant their own PR.

## Scope

- Disallow `aria-expanded` and `aria-pressed` on a `<summary>` that is its parent `<details>`'s disclosure trigger, while continuing to allow them on other elements (including standalone summaries and real buttons).
- Keep `aria-disabled` / `aria-haspopup` allowed (already global).
- Tests: `aria-allowed-attr` check unit + Shadow DOM, integration HTML/JSON, virtual-rules.

Related: #3911

straker:
We can close this. The true issue is that `summary` in our code has an implicit role of `button`, which allows both those attributes. If we want to fix that we need to update `summary` to not have an implicit role.

PR #5242, opened 2026-07-22
fix(aria-allowed-role): allow roles on a non-details summary

### LINKED ISSUE #3911: ARIA in HTML update for summary element allowances

chutchins25:
Updates the roles allowed on a `<summary>` element based on whether it is its parent `<details>` element's disclosure trigger, per the ARIA in HTML update.

## What & why

Per [ARIA in HTML](https://www.w3.org/TR/html-aria/) (`#el-summary`, [w3c/html-aria#435](https://github.com/w3c/html-aria/pull/435)):

- A `summary` that **is** the "summary for its parent `details`" (the first `summary` child of a `details`) allows **no explicit role** — browsers expose it as the disclosure trigger.
- A `summary` that is **not** the details summary (standalone, or a 2nd+ `summary`) is exposed as generic and allows **any role**.

axe previously modeled `summary` with an unconditional `allowedRoles: false`. This converts it to a `variant` using a `condition` matcher (the same pattern introduced for `figure` in #3443).

## Scope — role side only

Issue #3911 covers both the allowed **roles** and the allowed **`aria-*` attributes** for `summary`. This PR implements the **role** side. The attribute side (the spec disallows `aria-expanded`/`aria-pressed` on a details-summary) is intentionally deferred — it requires an element-level disallowed-attribute mechanism axe does not have today, and is tracked in **#5241**.

## Behavior change

A standalone / non-details `<summary>` now permits **any** role, where previously only the implicit `button` role was allowed. This is a spec-correct loosening; a `<details>`-summary's role handling is unchanged in effect (still restricted to its implicit `button` role).

## Tests
- `get-element-spec` unit — variant resolution for the three cases (details summary, non-first summary, standalone).
- `aria-allowed-role` check unit — incl. an open **Shadow DOM** case.
- `aria-allowed-role` virtual-rule — details/summary structures built with `SerialVirtualNode`, incl. the no-parent default.
- `aria-allowed-role` integration HTML/JSON — first-summary fail, 2nd-summary pass, standalone pass, implicit-button pass.

Closes #3911


lib/standards/html-elms.js
```diff
@@ -872,7 +872,32 @@ const htmlElms = {
...
   summary: {
-    allowedRoles: false,
+    variant: {
+      // A summary is the "summary for its parent details" when it is the
+      // first summary child of a details element. Browsers expose it as the
+      // details disclosure trigger, so authors cannot override its role.
+      // See https://github.com/w3c/html-aria/pull/435
+      summaryForDetails: {
+        matches: {
+          condition: vNode => {
```
straker: Probably a new matches function called `firstOfType` which kinda mimics the CSS pseudo selector of the same name `:first-of-type`. So maybe something like 

```js
matches: {
  firstOfType: 'summary'
}
```

straker: Also we have code that does this lookup of if the summary is the first summary of a details parent in https://github.com/dequelabs/axe-core/blob/develop/lib/commons/dom/visibility-methods.js#L248-L256

lib/standards/html-elms.js
```diff
@@ -872,7 +872,32 @@ const htmlElms = {
...
+              return false;
+            }
+            return (
+              parent.children.find(
+                child => child.props.nodeName === 'summary'
+              ) === vNode
+            );
+          }
+        },
+        allowedRoles: false
```
straker, line 902: Might also need to add the [allowed attrs](https://www.w3.org/TR/html-aria/#el-summary):`allowedAriaAttrs: ['aria-disabled', 'aria-haspopup']`

test/checks/aria/aria-allowed-role.js
```diff
@@ -426,4 +426,62 @@ describe('aria-allowed-role', () => {
...
+    );
+    assert.isFalse(
+      axe.testUtils
+        .getCheckEvaluate('aria-allowed-role')
+        .call(checkContext, null, null, vNode)
+    );
+    assert.deepEqual(checkContext._data, ['tab']);
+  });
+
+  it('returns true for a details summary with its implicit button role', () => {
```
straker: `summary` doesn't seem to have an implicit role of `button` anymore? I'm not sure when this changed as the way back machine is down at the moment. We should look into when this change was made. At least in https://www.w3.org/TR/2023/REC-html-aria-20230306/ they updated it to say that it had no role but some assistive technologies would announce it as button. This implies we need to test it ourselves and make sure what is announced.

test/checks/aria/aria-allowed-role.js
```diff
@@ -426,4 +426,62 @@ describe('aria-allowed-role', () => {
...
+      '<summary id="target" role="tab">summary</summary>'
+    );
+    assert.isTrue(
+      axe.testUtils
+        .getCheckEvaluate('aria-allowed-role')
+        .call(checkContext, null, null, vNode)
+    );
+  });
+
+  it('returns false for a details summary with an unallowed role in shadow DOM', () => {
```
straker: This type of shadow dom test isn't needed. Only time this is needed is if we need to cross the shadow dom boundary or not. But ones that just test things inside the shadow DOM aren't needed.

test/commons/standards/get-element-spec.js
```diff
@@ -39,6 +39,25 @@ describe('standards.getElementSpec', () => {
     assert.deepEqual(getElementSpec(vNode), {});
   });
 
+  it('should not allow a role on a summary that is the summary for its details', () => {
```
straker: This file tests more generic style tests than this so aren't needed. The integration tests for allowed-role will cover this.

straker (changes requested):

lib/standards/html-elms.js
chutchins25: Built the `firstOfType` matcher as suggested, but `matches: { firstOfType: 'summary' }` also matches a standalone summary (first-of-type under a non-`details` parent), which the spec treats as "any role" — so it regresses the standalone cases. Keeping it correct needs an additional "parent is `details`" condition, which the `matches` object can't express today without the inline function we're removing. Options: (a) extend the matches mechanism with a parent/ancestor matcher, (b) a more specific named matcher (e.g. `summaryForDetails`), or (c) accept treating any first-of-type summary as the trigger (matches the existing `visibility-methods` assumption). Which do you prefer? I've reverted the half-built version pending your call.

chutchins25: Thanks for the pointer — I'll reuse that first-summary-of-details logic. Folding this into the matcher-shape question in the thread above.

lib/standards/html-elms.js
chutchins25: The attribute side is handled in the stacked follow-up #5241 via an element-level `disallowedAttrs: ['aria-expanded', 'aria-pressed']`. `allowedAriaAttrs` (`aria-allowed-attr-elm`) only restricts *global* attrs and defers when there's an explicit role — but `aria-expanded`/`aria-pressed` are non-global, so it wouldn't catch them, and all globals are in fact allowed on a details-summary. Mind taking a look at #5241 to see if that shape works for you?

test/checks/aria/aria-allowed-role.js
chutchins25: Good catch. axe maps `summary` → `button` implicitly (since #2375, 2019), but current html-aria says summary has "no corresponding role." That's a pre-existing mapping with its own blast radius, so I've dropped the implicit-button assertion here and think changing the mapping belongs in its own issue rather than this PR. Want me to file one?

test/checks/aria/aria-allowed-role.js
chutchins25: Removed — no boundary crossing here, so agreed it's not needed.

test/commons/standards/get-element-spec.js
chutchins25: Removed — the allowed-role integration tests cover this.

lib/standards/html-elms.js
chutchins25: Done — replaced the inline condition with the serializable `isSummaryForDetails` matcher from #5262 (standards object is function-free now).

One behavior note: `isSummaryForDetails` intentionally throws (`Cannot resolve parent for non-DOM nodes`) for a parentless node — its own test asserts this — so `runVirtualRule('aria-allowed-role', { nodeName: 'summary' })` on a *bare* summary now throws instead of returning a result. I updated the summary virtual-rule test to give it a non-details parent (the realistic case). Flagging in case you'd prefer the matcher guard the parentless case instead — your call.

lib/standards/html-elms.js
```diff
@@ -872,7 +872,32 @@ const htmlElms = {
...
+          condition: vNode => {
+            const parent = vNode.parent;
+            if (!parent || parent.props.nodeName !== 'details') {
+              return false;
+            }
+            return (
+              parent.children.find(
+                child => child.props.nodeName === 'summary'
+              ) === vNode
+            );
```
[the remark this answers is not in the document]
chutchins25: Superseded. The variant no longer has inline `parent.children` logic — it now uses the shared `isSummaryForDetails` matcher (#5262), so the flagged implementation is gone. Parent resolution now lives in that matcher.

lib/standards/html-elms.js
chutchins25: Update: applying the conventions from #5240, I've removed the summary virtual-rule tests entirely (the aria-allowed-role integration tests cover these) rather than working around the matcher's parentless throw — so the earlier "gave it a non-details parent" change is gone. The matcher's throw-on-parentless behavior is left as-is.

straker (dismissed):

chutchins25:
@straker rebased onto develop — the merge conflicts are resolved. They were additive fixture collisions with the figure fixtures from #5240 that landed on develop, so I kept both the figure and summary cases in the `aria-allowed-role` integration fixtures. All 8 fixtures pass. It's mergeable now and there are no open threads, so it should be ready for another look whenever you get a chance.

straker (approved):

PR #5243, opened 2026-07-23
fix(aria-prohibited-attr): allow aria-label/labelledby on section and aside

### LINKED ISSUE #5185: section / aside elements should allow empty aria-labelledby
## Expected

This should pass `aria-prohibited-attr`:

```html
<secton aria-labelledby="foo">
  <p id="foo"></p>
</section>
```

## The issue

This is a bit of quirk in the ARIA specs. `section` elements get a `region` role if they have a non-empty accessible name, and `generic` if they don't. And since `generic` prohibited `aria-labelledby`, if the section has an empty accessible name, it causes the generic role, makes aria-labelledy prohivited. HOWEVER, looking at the allowances in ARIA in HTML, the `section` element is explicitly listed as allowing all global attributes. This indirectly creates an exception, due to which this passes.

There are quite a few elements in HTML that change role based on if they have a name. We should go over the ARIA in HTML specification and make sure we have all those allowances built into axe-core.

chutchins25:
Allows `aria-label` / `aria-labelledby` on a `<section>` or `<aside>` that has no accessible name, per ARIA in HTML.

## What & why

Per [ARIA in HTML](https://www.w3.org/TR/html-aria/) (`#el-section`, `#el-aside`), both elements allow **Global `aria-*` attributes** regardless of the role they resolve to.

In axe, an unnamed `<section>` (or an unnamed `<aside>` nested in sectioning content) resolves to a **`null` role**. The `aria-prohibited-attr` check treats a null-role element as prohibiting `aria-label`/`aria-labelledby`, so `<section aria-labelledby="…">` whose reference resolves to an empty name was wrongly flagged.

## Fix

Add `section` and `aside` to the check's `elementsAllowedAriaLabel` option. Named `section`/`aside` already pass (they resolve to `region`/`complementary`, which have no prohibited label attrs), so this only changes the null-role/empty-name case.

## Tests
- `aria-prohibited-attr` check unit — section with empty `aria-labelledby`, aside nested in sectioning content, named section, and an open **Shadow DOM** case.
- `aria-prohibited-attr` integration HTML/JSON — pass cases for section/aside with empty `aria-labelledby` and a named section.
- No virtual-rule cases: `aria-labelledby` idref resolution throws on serial (non-DOM) nodes, so the empty-labelledby scenario can't be exercised there.

## Doc reconciliation

`doc/check-options.md` documented the `elementsAllowedAriaLabel` default as a 12-element list that no longer matched the shipped default (`["applet", "input"]`). Corrected to the accurate current default (`applet`, `input`, `section`, `aside`).

Closes #5185

### DIFF
--- a/doc/check-options.md
+++ b/doc/check-options.md
@@ -99,18 +99,10 @@ All checks allow these global options:
       </td>
       <td align="left">
         <pre lang=js><code>[
-  "audio",
   "applet",
-  "canvas",
-  "dl",
-  "embed",
-  "iframe",
   "input",
-  "label",
-  "meter",
-  "object",
-  "svg",
-  "video"
+  "section",
+  "aside"
 ]</code></pre>
         </td>
       <td align="left">List of element names that without a role, are allowed an `aria-label` and `aria-labelledby` attribute</td>
--- a/lib/checks/aria/aria-prohibited-attr.json
+++ b/lib/checks/aria/aria-prohibited-attr.json
@@ -2,7 +2,7 @@
   "id": "aria-prohibited-attr",
   "evaluate": "aria-prohibited-attr-evaluate",
   "options": {
-    "elementsAllowedAriaLabel": ["applet", "input"]
+    "elementsAllowedAriaLabel": ["applet", "input", "section", "aside"]
   },
   "metadata": {
     "impact": "serious",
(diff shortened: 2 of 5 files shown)


chutchins25:
Superseded by #5259, which handles `aria-prohibited-attr` more generally — resolving naming/prohibited attributes from the html-elm spec vs the ARIA-role spec based on role resolution — and closes #5185. Closing in favor of that approach. Thanks @straker!

Issue #4747, opened 2025-04-07
Support role=generic


WilcoFiers:
Axe-core currently include generic in its list of roles. Even though ARIA says authors should not use it, they can, and from some quick testing it seems Firefox and Chrome allow it, and VO + Safari does seem to know to ignore it. More testing is probably needed, but even if it isn't fully supported it should be reported as not widely supported, instead of reporting it as an invalid role. It is a valid role.

lilzesh:
Same may go for disallowing. E.g., in aria-roledescription-evaluate. You’d want to ensure the role was not generic in order to hit the false return, since it’s prohibited. Currently, one of the examples listed as a failure doesn’t actually fail as it returns undefined.

`<div aria-roledescription=“my div” id=“fail2”>div</div>`

PR #5245, opened 2026-07-23
fix(aria-prohibited-attr): prohibit aria-label/labelledby on body and label

### LINKED ISSUE #3410: Three additional elements that prohibit naming from authors
From my checks, axe-core largely identifies ARIA 1.2's [naming prohibited rule](https://www.w3.org/TR/wai-aria-1.2/#namefromprohibited)., and extends that rule to HTML elements with similar implicit roles.  However, there are a few elements which were not flagged for having an`aria-label` or `aria-labelledby` attribute per the [test case](https://w3c.github.io/html-aria/tests/prohibit-names.html) I checked against.

- `body` 
- `label`
- `rp`

The remaining 38 elements checked in the test case were all flagged by axe for not supporting `aria-label` or `aria-labelledby` unless a role which allows naming is also specified.

chutchins25:
Prohibits `aria-label` / `aria-labelledby` on the `body` and `label` elements, per ARIA in HTML "naming prohibited".

> **Stacked PR** — targets `chut/5185-section-aside-labelledby` (#5185). GitHub will auto-retarget this to `develop` once #5185 merges.

## What & why

Per [ARIA in HTML](https://www.w3.org/TR/html-aria/):
- **`body`** — role generic, **Naming Prohibited**.
- **`label`** — **Naming Prohibited** when exposed as generic / associated with a labelable element. The authoritative html-aria test fixture (`tests/prohibit-names.html`, authored by the spec editor) marks a bare `<label aria-label>` / `<label aria-labelledby>` as **fail**.

Both currently pass `aria-prohibited-attr` because they resolve to a non-null role (`body` → implicit `document`; `label` → chromium `Label`), so the check's "no role → prohibited" branch never runs.

## Fix

Adds a new `elementsProhibitedAriaLabel` check option (default `["body", "label"]`) and an element-level short-circuit in the evaluate: elements in this list (and not overridden via `elementsAllowedAriaLabel`) prohibit `aria-label`/`aria-labelledby` regardless of the role they expose. These report as "no role" since the prohibition is tied to the element, not the implicit role.

The `label` prohibition is unconditional (matches the html-aria test fixture and the issue), rather than attempting to detect labelable-element association.

Note: the issue originally listed a third element (`rp`), but the author struck it — not included.

## Behavior change

`<body aria-label>` / `<label aria-label>` (and `aria-labelledby`) now report on `aria-prohibited-attr`: a **violation** when the element has no visible text, or **needs-review (incomplete)** when it does. Element-level and gated on `nodeName`, so no other elements are affected.

## Tests
- `aria-prohibited-attr` check unit — label single / plural / text-incomplete / `elementsAllowedAriaLabel` opt-out, plus an open **Shadow DOM** case.
- `aria-prohibited-attr` virtual-rule — `body` and `label` violations (`body` can't live in the shared integration page).
- `aria-prohibited-attr` integration HTML/JSON — `label` violation + incomplete cases.
- `doc/check-options.md` — documents the new `elementsProhibitedAriaLabel` option.

Closes #3410

### DIFF
--- a/doc/check-options.md
+++ b/doc/check-options.md
@@ -107,6 +107,18 @@ All checks allow these global options:
         </td>
       <td align="left">List of element names that without a role, are allowed an `aria-label` and `aria-labelledby` attribute</td>
     </tr>
+    <tr>
+      <td>
+        <code>elementsProhibitedAriaLabel</code>
+      </td>
+      <td align="left">
+        <pre lang=js><code>[
+  "body",
+  "label"
+]</code></pre>
+        </td>
+      <td align="left">List of element names that are prohibited from having an `aria-label` or `aria-labelledby` attribute, regardless of the role they expose</td>
+    </tr>
   </tbody>
 </table>
 
--- a/lib/checks/aria/aria-prohibited-attr-evaluate.js
+++ b/lib/checks/aria/aria-prohibited-attr-evaluate.js
@@ -33,6 +33,8 @@ export default function ariaProhibitedAttrEvaluate(
   virtualNode
 ) {
   const elementsAllowedAriaLabel = options?.elementsAllowedAriaLabel || [];
+  const elementsProhibitedAriaLabel =
+    options?.elementsProhibitedAriaLabel || [];
   const { nodeName } = virtualNode.props;
   const role = getRole(virtualNode, {
     chromium: true,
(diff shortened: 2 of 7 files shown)


straker (changes requested):
While this solves the problem I don't think it does so in the way we want. The true issue is that axe-core [does not yet support the `generic` role](https://github.com/dequelabs/axe-core/issues/4747). Both `body` and `label` should be implicit role of `generic`, which would then prohibit `aria-label` and `aria-labelledby`. So the fix to this issue should be to have `generic` be fully supported in axe-core.

chutchins25:
Agree the root cause is the missing `generic` role (#4747) — `body`/`label` should map to implicit `generic`, which prohibits naming. That's a larger change (add the role, remap several elements' implicit roles, "not widely supported" reporting, broad blast radius) well beyond this PR's scope.

Options:
1. Close this PR and fold the fix into #4747.
2. Land the contained option-based fix now and supersede it when #4747 lands.

I lean toward (1) if #4747 is near-term. What's your preference?

straker:
Option 1 is good

chutchins25:
Superseded by #5259, which prohibits naming on `body`/`label` via the html-elm `namingProhibited` mechanism and closes #3410. Closing in favor of that approach.

PR #5246, opened 2026-07-23
feat(aria-allowed-attr): flag deprecated ARIA attributes as needs-review

### LINKED ISSUE #3341: Indicate that aria-grabbed and aria-dropeffect are deprecated
ARIA 1.1 deprecated [`aria-grabbed`](https://www.w3.org/TR/wai-aria-1.1/#aria-grabbed) and [`aria-dropeffect`](https://www.w3.org/TR/wai-aria-1.1/#aria-dropeffect).  While no feature yet exists to replace these attributes, their current support is not particularly great, and since they are deprecated, it is highly unlikely their support will improve.

ARIA in HTML is being updated to call out that authors should not rely on these features, and request that conformance checkers warn authors about the attributes' deprecated status / potential lack of support from browsers/at.

[`aria-grabbed` and `aria-dropeffect` test case](https://w3c.github.io/html-aria/tests/deprecated-attrs.html)

chutchins25:
Flags the deprecated ARIA attributes `aria-grabbed` and `aria-dropeffect` as "needs review".

## What & why

WAI-ARIA 1.1 deprecated `aria-grabbed` and `aria-dropeffect` globally, but axe had no signal for deprecated ARIA *attributes* (it already handles deprecated *roles* via `aria-deprecated-role`).

Following the approach in the issue thread ([@WilcoFiers](https://github.com/dequelabs/axe-core/issues/3341)): add a new **`aria-deprecated-attr` check to the existing `aria-allowed-attr` rule** (rather than a standalone rule), which **incompletes** an element using deprecated ARIA, or lets the rule **fail** when the element also uses an unsupported/unallowed attribute.

## Implementation

- `deprecated: true` on `aria-grabbed` / `aria-dropeffect` in `lib/standards/aria-attrs.js`.
- New data-driven `aria-deprecated-attr` check: returns **incomplete** when any attribute with `deprecated: true` is present, pass otherwise.
- Added to the `aria-allowed-attr` rule's `all` array. Because a failing sibling check (`aria-allowed-attr` / `aria-unsupported-attr`) outranks an incomplete, an element with a deprecated attribute is:
  - **needs review** when that's its only ARIA issue, or
  - a **violation** when it also has an unsupported/unallowed attribute.

## Behavior change

Any element using `aria-grabbed` or `aria-dropeffect` now reports under `aria-allowed-attr` (needs-review, or a violation when combined with an unsupported attribute). The `passes.html` integration fixture used these two as "global attribute" filler on 68 pass blocks; they've been removed from those blocks (each still tests global-attribute allowance via its other global attributes), with dedicated incomplete/failure coverage added.

## Tests
- New `aria-deprecated-attr` check unit — incomplete (single/multiple), data-driven via `axe.configure`, and an open **Shadow DOM** case.
- `aria-allowed-attr` virtual-rule — incomplete for a deprecated attr; violation when combined with an unallowed attr.
- `aria-allowed-attr` integration — `incomplete.html` (deprecated-only, singular + plural) and `failures.html` (#fail12: deprecated + unallowed → violation); 68 pass blocks updated.

## Follow-up

Per the issue discussion, a separate ticket will track a future standards mechanism to return "needs review" for not-recommended-but-valid roles/attributes generally.

Closes #3341


lib/rules/aria-allowed-attr.json
```diff
@@ -16,7 +16,7 @@
     "description": "Ensure an element's role supports its ARIA attributes",
     "help": "Elements must only use supported ARIA attributes"
   },
-  "all": ["aria-allowed-attr", "aria-allowed-attr-elm"],
+  "all": ["aria-allowed-attr", "aria-allowed-attr-elm", "aria-deprecated-attr"],
```
straker: Minor nit: Since this is used in the `all` it would be more correct to say this is `aria-no-deprecated-attr` to read as what the positive case is that should pass (not having a deprecated attr).

test/checks/aria/aria-deprecated-attr.js
```diff
@@ -0,0 +1,69 @@
...
+      }
+    });
+    const params = checkSetup(
+      '<div id="target" aria-fizz="buzz">Contents</div>'
+    );
+    assert.isUndefined(checkEvaluate.apply(checkContext, params));
+    assert.deepEqual(checkContext._data, ['aria-fizz']);
+  });
+
+  it('returns undefined for a deprecated attribute in shadow DOM', () => {
```
straker: Isn't needed.

straker (dismissed):

lib/rules/aria-allowed-attr.json
chutchins25: Done — renamed the check to `aria-no-deprecated-attr` so the `all` entry reads as the passing case. Propagated it through the rule, check JSON, locales, and tests, and rebased the stacked #5257 onto it.

test/checks/aria/aria-deprecated-attr.js
chutchins25: Removed — dropped the in-boundary Shadow DOM test since it doesn't cross a boundary.

lib/checks/aria/aria-no-deprecated-attr-evaluate.js
```diff
@@ -0,0 +1,42 @@
...
+ *       <th>Type</th>
+ *       <th>Description</th>
+ *     </tr>
+ *   </thead>
+ *   <tbody>
+ *     <tr>
+ *       <td><code>String[]</code></td>
+ *       <td>List of all deprecated attributes</td>
+ *     </tr>
+ *   </tbody>
```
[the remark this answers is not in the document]
chutchins25: Fixed. Reworded the data-table description to "List of the deprecated ARIA attributes used on the element" to match what the check reports (the deprecated attrs found on the element, not the full standards list).

straker (approved):

PR #5248, opened 2026-07-23
feat(aria): support role=image as equivalent to role=img

### LINKED ISSUE #4656: false positive for use of valid role=image
### Product

axe-core

### Product Version

4.10.2

### Latest Version

### Issue Description

per the following test
```
<div role=image>foo</div>
```

the checker reports a"ARIA roles used must conform to valid values".  `role=image`  is a [valid role](https://w3c.github.io/aria/#image) which is supported by all browsers ([wpt demonstrating role=image supported in all browsers](https://wpt.fyi/results/wai-aria/role/synonym-roles.html?label=pr_head&max-count=1&pr=42203).

ARIA in HTML is updating to indicate this is an allowed role for all the instances of where `role=img` was also allowed.

chutchins25:
Adds support for `role="image"` as a valid role equivalent to `role="img"` (ARIA 1.3, [w3c/aria#1370](https://github.com/w3c/aria/pull/1370)).

## What & why

ARIA 1.3 adds `image` as an alias of `img` with an identical mapping; axe previously treated `role="image"` as an invalid role.

Per the review discussion, this does **not** use a synonym-normalization mechanism (which would force every reporting path to remember the role the author used vs. the one our code resolved it to). Instead, `image` is a first-class role that shares `img`'s definition and is allowed wherever `img` is:

- `image` is added to `lib/standards/aria-roles.js`, spread from a shared `imgRole` object so it stays in sync with `img`.
- `image` is added to every `html-elms` `allowedRoles` list that includes `img` (`embed`, `iframe`, `object`). `svg` needs no change — it is `allowedRoles: true`, so it already accepts any role.
- The `role-img-alt` and `svg-img-alt` selectors match both roles via `:is([role='img'], [role='image'])`. Without the `svg-img-alt` change, `<svg role="image">` matched neither rule — `svg-img-alt`'s selector omitted it and `role-img-alt` is gated by `html-namespace-matches` — so axe accepted the role and then never checked for an accessible name.
- `image` is added to the `aria-roledescription` check's `supportedRoles`, so `role="image"` passes where `role="img"` does.

Because there is no synonym resolution, an unallowed `role="image"` reports **`image`** (the role the author used), not `img`.

## Tests

- `is-valid-role` / `aria-roles` integration — `image` is a valid role.
- `get-element-unallowed-roles` — `image` allowed on `svg`/`embed`/`object`/`iframe`; reported as authored when unallowed (e.g. on `hr`).
- `role-img-alt` integration + virtual-rules — matches `role="image"`.
- `svg-img-alt` integration + virtual-rules — matches `role="image"`.
- `aria-roledescription` — `role="image"` passes against the check's default `supportedRoles`.
- ACT specs `7d6734` and `23a2a8` still pass with the broadened selectors.

## Not in scope

`role-img-alt` and `svg-img-alt` metadata still name only `img` in their `description`/`help` strings. Updating them regenerates `locales/_template.json`, invalidates those translations, and needs a matching axe-rule-help change — tracked in #5272.

Closes #4656


chutchins25:
Design note: this PR implements the synonym mechanism @WilcoFiers proposed on the issue rather than special-casing `image` throughout — see https://github.com/dequelabs/axe-core/issues/4656#issuecomment-5033296404. Per the later discussion (support demonstrated by @scottaohara), `image` ships as a fully valid role, not `unsupported`. The follow-up to migrate `none`/`presentation` onto the same `synonym` field is tracked in #5249.

lib/rules/role-img-alt.json
```diff
@@ -1,7 +1,7 @@
 {
   "id": "role-img-alt",
   "impact": "serious",
-  "selector": "[role='img']:not(img, area, input, object)",
+  "selector": "[role='img']:not(img, area, input, object), [role='image']:not(img, area, input, object)",
```
straker: We can use `:is` to shorten this:


```suggestion
  "selector": ":is([role='img'], [role='image']):not(img, area, input, object)",
```

lib/standards/aria-roles.js
straker: This is our first time doing this. I wonder if we should instead make this an object that is spread in both so we don't have code duplication we have to update twice:

```js
const img = {
  // props
}

export const ariaRoles = {
  img: { ...img },
  image: {
    ..img,
    synonym: 'img'
  }
}
```

lib/commons/aria/get-explicit-role.js
```diff
@@ -1,8 +1,9 @@
 import isValidRole from './is-valid-role';
+import standards from '../../standards';
 import { getNodeFromTree, tokenList } from '../../core/utils';
 import AbstractVirtuaNode from '../../core/base/virtual-node/abstract-virtual-node';
 
-function getExplicitRole(vNode, { fallback, abstracts, dpub } = {}) {
+function getExplicitRole(vNode, { fallback, abstracts, dpub, noSynonym } = {}) {
```
straker: This should probably happen in `getImplicitRole` as well so it covers element internals using `image` as the role also can use this logic. Also we'll want tests for `getRole` that makes sure the `noSynonym` option is passed through both role functions.

```js
customElements.define(
  'my-elm',
  class MyElm extends HTMLElement {
    constructor() {
      super();
      this._internals = this.attachInternals();
      this._internals.role = 'image';
    }
  }
); 
```

lib/commons/aria/get-element-unallowed-roles.js
```diff
@@ -41,8 +42,11 @@ function getRoleSegments(vNode) {
     roles = roles.concat(nodeRoles);
   }
 
-  // filter invalid roles
-  return roles.filter(role => isValidRole(role));
+  // filter invalid roles and resolve synonym roles (e.g. `image` -> `img`)
+  // so the allowed-role check compares against the canonical role
+  return roles
+    .filter(role => isValidRole(role))
+    .map(role => standards.ariaRoles[role]?.synonym || role);
```
straker: So one thing we'll need to consider is how this is used in context and reported to the user through the final rule. 

So for example let's say an `img` is not allowed on an element (for example the `hr` element), but the user used `role=image`. This could would map the `image` role to `img` and then report that `img` is not allowed. But when we report that back to the user it'll say `img` is not allowed but the user used `image`. Instead this should make sure that the reported role to the user is what they used (`img` or `image`) and not always the `synonym`,

<img width="439" height="79" alt="screenshot of aria-allowed-role rule showing <hr role='image'> failure message saying that role 'img' is not allowed" src="https://github.com/user-attachments/assets/762c086f-64a3-4db9-9ade-104b19c4d27a" />

straker (dismissed):

straker (commented):
After look at this pr and https://github.com/dequelabs/axe-core/pull/5255, I think we should drop the whole `synonym` stuff as it can cause real issues in reporting the role back to the user (we'd have to remember that both these cases require getting the role the user used and not the one our code resolved it to). I'd rather our code was a bit messier in the `[presentation, none].includes(role)` checks in order to prevent confusing reporting.

Additionally, this pr also  needs to add `image` to the html-elms spec anywhere the `allowedRole` allows `img` (e.g. `embed`)

lib/standards/aria-roles.js
```diff
@@ -267,13 +278,13 @@ const ariaRoles = {
     accessibleNameRequired: false,
     nameFromContent: true
   },
+  // image is an ARIA 1.3 alias of img (w3c/aria#1370) with an identical
+  // mapping. Spread the shared definition so the two entries stay in sync.
```
straker: No need for the comment, was already stated at the defintion.

```suggestion
```

straker (commented):
Just a minor nit about duplicate comments

lib/rules/role-img-alt.json
chutchins25: Done — selector is now `:is([role='img'], [role='image']):not(img, area, input, object)`.

lib/standards/aria-roles.js
chutchins25: Done — `img` and `image` now both spread from a shared `imgRole` object so the two stay in sync.

lib/commons/aria/get-explicit-role.js
chutchins25: This one's moot now — per your later comment we dropped the synonym mechanism entirely, so there's no `noSynonym` option to thread through. `image` is a first-class role, so an element-internals `role="image"` resolves as a valid `image` role directly (no normalization step).

lib/commons/aria/get-element-unallowed-roles.js
chutchins25: Resolved by dropping the synonym approach: with no `image`→`img` resolution, an unallowed `role="image"` now reports `image` — the role the author actually used. Added a `get-element-unallowed-roles` test for `<hr role="image">` asserting it reports `image`.

lib/standards/aria-roles.js
chutchins25: Done — removed; the shared `imgRole` definition already documents the `img`/`image` link.

chutchins25:
Done. Dropped the synonym mechanism — `image` is now a first-class role sharing `img`'s definition, and I added `image` to every `html-elms` `allowedRoles` that allowed `img` (`embed`, `iframe`, `svg`). `role-img-alt` matches both via `:is()`. Since there's no resolution step, the authored role is what gets reported. Updated the PR description to match.

lib/standards/aria-roles.js
```diff
@@ -267,13 +278,13 @@ const ariaRoles = {
...
+  },
   img: {
-    type: 'structure',
-    // Spec difference: Aria-expanded removed in 1.2
-    allowedAttrs: ['aria-expanded'],
-    superclassRole: ['section'],
-    accessibleNameRequired: true,
-    childrenPresentational: true
+    ...imgRole
   },
```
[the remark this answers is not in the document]
chutchins25: Outdated. The synonym mechanism referenced here was dropped — `image` is now a first-class role sharing `img`'s definition via a spread `imgRole` object, with no `synonym`/`noSynonym` handling. The PR title and description have been updated to match; they no longer describe synonym normalization.

straker (commented):
I believe `svg-img-alt` also needs to be updated to select `[role=image}`, but this will require asking @WilcoFiers if the `image` role also applies to svg elements (it's a different spec).

chutchins25:
> I believe `svg-img-alt` also needs to be updated to select `[role=image}`, but this will require asking @WilcoFiers if the `image` role also applies to svg elements (it's a different spec).

Good catch. It was worse than a missing selector — `<svg role="image">` matched *neither* rule: `svg-img-alt`'s selector omitted it, and `role-img-alt` is gated by `html-namespace-matches`. Since `svg` is `allowedRoles: true`, axe reported the role as allowed and then never checked for an accessible name.

I've updated `svg-img-alt` to `:is([role='img'], [role='image'])` with integration and virtual-rule coverage. My reading of the spec question: SVG-AAM maps SVG *elements*, and role *values* come from ARIA, so `image` should be as valid on `<svg>` as `img` is. Both ACT specs (`7d6734`, `23a2a8`) still pass.

@WilcoFiers — could you confirm? If `image` should *not* apply to SVG, the fix is the opposite of what I've done: `svg` is `allowedRoles: true` today, so axe accepts `<svg role="image">` either way, and we'd need to give `svg` an explicit allowlist to reject it. Happy to flip it.

While sweeping for other role-name allowlists that mention `img`, I also found the `aria-roledescription` check's `supportedRoles` — it listed `img` but not `image`, so `role="image"` returned incomplete where `role="img"` passes. Added. (`lookup-table.js` mentions `img` too, but it's `@deprecated`.)

Also in this push: `image` coverage for `object`/`iframe` in `get-element-unallowed-roles`, and the leftover "synonym" wording dropped from the test names now that the mechanism is gone. Rule message wording is tracked separately in #5272.

straker (approved):

PR #5250, opened 2026-07-23
test(aria-allowed-role): cover button element widget roles

### LINKED ISSUE #4142: role="separator" not allowed on button via aria-allowed-role
### Product

axe-core

### Product Version

4.7.2

### Latest Version

#### Expectation
In order to achieve a simple, focusable separator I should be able to add `role="separator"` and `aria-valuenow` to a button.

#### Actual
Adding `role="separator"` to a button in order to achieve a focusable separator fails axe-core tests for aria-allowed-role as a best practice.

> To solve this problem, you need to fix the following:
> ARIA role separator is not allowed for given element

[ARIA role should be appropriate for the element](https://dequeuniversity.com/rules/axe/4.7/aria-allowed-role)

#### How to Reproduce
`<button role="separator" aria-valuenow="50" aria-label="Handle">></button>`

Sample code: [Simple focusable separator](https://output.jsbin.com/ciyisasude)

#### Additional context
Is there anything that is preventing me from overwriting the button semantics with `role="separator"` in order gain native focusability and event handling?

chutchins25:
Just adds the `aria-allowed-role` integration cases for the `<button>` element (`separator`, `slider`, `treeitem`; `gridcell` was already covered), which is all #4142 needs on top of #4398.

Closes #4142


straker (dismissed):
This does not resolve the issue. The issue was for the `button` element and not the `input type=button`. Also the only thing that needs to be update for this past the addition to the spec is an integration test for aria-allowed-role (don't need tests for the unit test or the virtual nodes)

chutchins25:
Done — reworked to just add the `aria-allowed-role` integration cases for the `<button>` element (`separator`, `slider`, `treeitem`; `gridcell` was already covered), which is all #4142 needs on top of #4398. Reverted the `input type="button"` role additions and dropped their unit/virtual-rule tests, since the issue is about the button element.

straker (approved):

PR #5254, opened 2026-07-23
fix(aria-prohibited-attr): allow aria-label/labelledby on form

### LINKED ISSUE #5244: Audit ARIA in HTML global aria-* allowances for null-role elements (e.g. form) in aria-prohibited-attr
Follow-up to #5185, which fixed `aria-prohibited-attr` wrongly flagging `aria-label`/`aria-labelledby` on an unnamed `<section>`/`<aside>` (elements that resolve to a `null` role but still allow global `aria-*` attributes per ARIA in HTML).

The same pattern affects other elements whose implicit role is name-dependent and resolves to `null` when unnamed. For example, `<form>`:

```js
// lib/commons/standards/implicit-html-roles.js
form: vNode => (hasAccessibleName(vNode) ? 'form' : null)
```

So `<form aria-labelledby="…">` with an empty/unresolved reference has the same false positive today.

## Ask

Audit the ARIA in HTML "Global `aria-*` attributes" allowances against axe's `aria-prohibited-attr` handling, and add any other elements that:

1. resolve to a `null` role when they have no accessible name, and
2. are permitted global `aria-*` attributes by the spec,

to the `elementsAllowedAriaLabel` option (or a more general mechanism if the list grows large).

Candidates to confirm against the spec: `form`, and any other name-dependent-implicit-role elements. (`header`/`footer` map to `null` only when scoped, but their allowance is name-independent — confirm during the audit.)

Related: #5185

chutchins25:
Extends the #5185 fix to the `form` element, which has the same false positive.

> **Stacked PR** — targets `chut/3410-body-label-prohibit` (#3410), which is itself stacked on #5185. GitHub will auto-retarget as the parents merge.

## What & why

Follow-up to #5244 / #5185. An unnamed `<form>` resolves to a `null` role in axe (`implicit-html-roles.js`: `hasAccessibleName(vNode) ? 'form' : null`), so `aria-prohibited-attr` wrongly prohibited `aria-label`/`aria-labelledby` on `<form aria-labelledby="…">` with an empty/unresolved reference.

Per [ARIA in HTML](https://www.w3.org/TR/html-aria/) (`#el-form`): the form element allows **Global `aria-*` attributes**, and "A `form` is not exposed as a landmark region unless it has been provided an accessible name."

## Audit result

I audited `implicit-html-roles.js` for the name-dependent-null-role pattern. Only `section`, `aside` (fixed in #5185) and **`form`** match. `header`/`footer` map to `null` based on *ancestor* (not accessible name), and a scoped `<footer>`/`<header>` is exposed as `generic` — which *prohibits* naming — so flagging those is correct and they are intentionally excluded.

## Fix
Add `form` to the `elementsAllowedAriaLabel` option (named forms already pass via the `form` landmark role).

## Tests
- `aria-prohibited-attr` check unit — form with empty `aria-labelledby`; named form.
- `aria-prohibited-attr` integration — form with empty `aria-labelledby` pass case.
- `doc/check-options.md` default updated.

Closes #5244

### DIFF
--- a/doc/check-options.md
+++ b/doc/check-options.md
@@ -102,7 +102,8 @@ All checks allow these global options:
   "applet",
   "input",
   "section",
-  "aside"
+  "aside",
+  "form"
 ]</code></pre>
         </td>
       <td align="left">List of element names that without a role, are allowed an `aria-label` and `aria-labelledby` attribute</td>
--- a/lib/checks/aria/aria-prohibited-attr.json
+++ b/lib/checks/aria/aria-prohibited-attr.json
@@ -2,7 +2,7 @@
   "id": "aria-prohibited-attr",
   "evaluate": "aria-prohibited-attr-evaluate",
   "options": {
-    "elementsAllowedAriaLabel": ["applet", "input", "section", "aside"],
+    "elementsAllowedAriaLabel": ["applet", "input", "section", "aside", "form"],
     "elementsProhibitedAriaLabel": ["body", "label"]
   },
   "metadata": {
--- a/test/checks/aria/aria-prohibited-attr.js
+++ b/test/checks/aria/aria-prohibited-attr.js
@@ -320,6 +320,21 @@ describe('aria-prohibited-attr', () => {
       );
       assert.isFalse(checkEvaluate.apply(checkContext, params));
     });
+
+    it('should allow aria-labelledby on a form without an accessible name', () => {
+      const params = checkSetup(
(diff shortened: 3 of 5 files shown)


chutchins25:
Superseded by #5259. Its general mechanism already allows naming on an unnamed `form` (the role resolves to the implicit role and `form` isn't `namingProhibited`), so the dedicated `form` addition here isn't needed. Closing — #5259 covers the #5244 case.

Issue #5217, opened 2026-07-20
Don't fail ARIA attributes reflected from an invalid to a valid element


WilcoFiers:
Found this issue while benchmarking. The following shouldn't fail `aria-allowed-attr`:

```html
<text-field aria-required="true">
  <input aria-required="true" />
</text-field>
```

I think what to do here is, before Axe fails the element for having aria-required, check if any of its descendants has the same property AND value. 

There are a few things to consider when fixing this:

1. Which attributes do we want to allow this on? All non-global, or maybe a more limited set?
2. Should we only allow this on generics like?
3. Do we want this to go to needs review?
4. What happens if multiple descendants have this prop?
5. Should Axe limit how far it searches?

PR #5259, opened 2026-07-27
feat(aria-prohibited-attr): allow many elements to be named and disallow label and body from being named

### LINKED ISSUE #3410: Three additional elements that prohibit naming from authors

### LINKED ISSUE #5185: section / aside elements should allow empty aria-labelledby

straker:
After looking over the HTML in ARIA spec, I realized that we could better handle how we determine when an attribute is prohibited. Basically if the element doesn't have an explicit role, then we use the `htmlElm` spec to determine prohibited attrs and naming, otherwise we use the `ariaRoles` spec to determine that. This should allow us to eliminate the `elementsAllowedAriaLabel` option that was [introduced to fix which elements allowed naming](https://github.com/dequelabs/axe-core/pull/2935) and then modified when we [added Chromium roles](https://github.com/dequelabs/axe-core/pull/2948) to fix conflict name resolution. However the option still has the `applet` element which isn't in our html spec so I'm leaving it for now.

Elements which allow naming now include: `section`, `blockquote`, `address`, `hgroup`, `ruby`.

This also adds enforcement for elements that are not allowed any aria-* attribute (which has never been used in our code) by changing the `noAriaAttrs` spec property to an empty `allowedAriaAttrs` so that they now throw in the rule `aria-allowed-attr`.

Closes: https://github.com/dequelabs/axe-core/issues/5185
Closes: https://github.com/dequelabs/axe-core/issues/3410


WilcoFiers (dismissed):
Nice cleanup — replacing the "no role at all means no name" inference with explicit spec data is clearly the right direction, and it fixes a batch of long-standing false positives. I built both the merge-base and this branch and diffed rule output across ~60 markup cases to see exactly what moves. Two things need a decision before merge.

Note: the failing `test_node (18)` and `test_node (20)` checks are an unrelated `undici`/`jsdom` incompatibility, not caused by this PR.

---

### 1. `del` loses naming prohibition while `ins` keeps it (blocking)

`lib/standards/html-elms.js` L215 (compare `ins` at L566)

`del` didn't get `namingProhibited: true`, but `ins` did:

```
<del aria-label="hello"></del>
  before: violation — aria-label attribute cannot be used on a del with no valid role attribute.
  after:  PASS

<ins aria-label="hello"></ins>
  before: violation
  after:  violation
```

Naming is prohibited on both — `del` maps to `deletion` and `ins` to `insertion`, and both carry `prohibitedAttrs` in `aria-roles.js`. Since neither element appears in `implicitHtmlRoles`, the role spec is no longer consulted, so `del` silently drops its prohibition.

**Suggestion:** add `namingProhibited: true` to `del`, and add a test case pairing `del` with `ins` so the two can't drift apart again.

---

### 2. `body` and `dd` report a role whose spec allows naming (blocking)

`lib/checks/aria/aria-prohibited-attr-evaluate.js` L76, with `body` at `html-elms.js` L135 and `dd` at L211

These are the only two elements given `namingProhibited: true` that also have a non-null implicit role, and two problems surface there.

**The outcome is right but the reported role is wrong.** `<body aria-label>` should fail — but axe maps `body` to `document` and `dd` to `definition`, and neither role has `prohibitedAttrs`, so the message blames a role that permits naming:

```
<body aria-label="hello">    before: PASS
                             after:  incomplete — aria-label attribute is not well supported with role "document".
<dd aria-label="hello">      before: PASS
                             after:  violation — aria-label attribute cannot be used with role "definition".
```

I confirmed `aria-allowed-attr` still passes both, so the two rules now disagree with each other. The real fix is in the mapping: `body` should be `generic` and `dd` should have no corresponding role. Worth noting `generic` doesn't exist in `lib/standards/aria-roles.js` at all today, so that correction is a prerequisite rather than a one-liner.

**`role === implicitRole` isn't a safe proxy for "no explicit role."** An author can write `<dd role="definition">` or `<p role="paragraph">` and land in the HTML-spec branch even though an explicit role was set — which is the case the ARIA role spec should govern.

**Suggestion:** branch on `getExplicitRole()` rather than comparing the resolved role to the implicit role, and fix the `body`/`dd` mappings (adding `generic` to `aria-roles.js`) so the prohibition and its message come from the same place. If the mapping fix is too large for this PR, splitting it out would at least keep `body` and `dd` from shipping with contradictory messaging in the meantime.

---

### 3. Four more elements correctly stop prohibiting names (note)

Beyond `del`, four elements flip from violation to pass, all correctly:

```
<address aria-label="hello">     before: violation   after: PASS
<blockquote aria-label="hello">  before: violation   after: PASS
<hgroup aria-label="hello">      before: violation   after: PASS
<ruby aria-label="hello">        before: violation   after: PASS
```

These were false positives from the old catch-all — `implicitHtmlRoles` only lists elements whose role axe actually needs, so every known HTML element in that gap fell through to `return ['aria-label', 'aria-labelledby']`. This is a fix, not a regression.

**Suggestion:** call these out in the PR description so they land in the release notes, and add test cases pinning the new behavior so we don't regress them in the future.

---

### 4. `doc/check-options.md` documents a stale default (non-blocking)

`doc/check-options.md` L98

The PR changes the `elementsAllowedAriaLabel` default to `["applet"]`, but the docs still advertise a twelve-element list including `label` — precisely what this PR reverses. The table was already stale, but this makes it actively contradict the fix in the PR title. Two-line change while you're in there.

---

### 5. `noAriaAttrs` branch reuses the `noRole` messages (non-blocking, separate issue)

`lib/checks/aria/aria-prohibited-attr-evaluate.js` L88

```
<col aria-invalid="true">
  violation — aria-invalid attribute cannot be used on a col with no valid role attribute.
```

The wording implies a valid role would fix it, but `col` permits no ARIA attributes at all regardless of role. Same message now reaches `map`, `picture`, `base`, `head`, `html`, `link`, `meta`, `noscript`, `param`, `script`, `slot`, `source`, `style`, `template`, `title`, and `track`. A dedicated message key would read better — worth its own issue rather than expanding this PR.

chutchins25 (dismissed):

lib/checks/aria/aria-prohibited-attr-evaluate.js
```diff
@@ -71,18 +71,29 @@ export default function ariaProhibitedAttrEvaluate(
...
+      getClosestAncestorRoleType(vNode) === 'widget'
+    ) {
+      return [];
+    }
+
+    const htmlSpec = getElementSpec(vNode);
+    if (!standards.htmlElms[nodeName] || htmlSpec.namingProhibited) {
+      return ['aria-label', 'aria-labelledby'];
+    }
+    if (htmlSpec.noAriaAttrs) {
```
straker: Instead of adding a new property `noAriaAttrs` to the standard and checking it here, we should use the `ariaAllowedAttrs` array and for every case where we added `noAriaAttrs` instead make it an empty array for `ariaAllowedAttrs`. That way the check then happens in `aria-allowed-attr-elm-evaluate` instead of here and we can remove it from here.

lib/checks/aria/aria-prohibited-attr-evaluate.js
```diff
@@ -71,18 +71,29 @@ export default function ariaProhibitedAttrEvaluate(
...
-    return roleSpec.prohibitedAttrs || [];
-  }
-  if (
-    !!role ||
-    elementsAllowedAriaLabel.includes(nodeName) ||
-    getClosestAncestorRoleType(vNode) === 'widget'
-  ) {
+  const implicitRole = getImplicitRole(vNode);
+
+  if (role === implicitRole) {
```
straker: With the change to `noAriaAttrs` we can change this if statement to only look at the explicit role as every time an element is not allowed naming, their role also is not allowed naming.

lib/checks/aria/aria-prohibited-attr-evaluate.js
```diff
@@ -71,18 +71,26 @@ export default function ariaProhibitedAttrEvaluate(
...
-    return roleSpec.prohibitedAttrs || [];
-  }
-  if (
-    !!role ||
-    elementsAllowedAriaLabel.includes(nodeName) ||
-    getClosestAncestorRoleType(vNode) === 'widget'
-  ) {
+  const explicitRole = getExplicitRole(vNode, { fallback: true });
+
+  if (!explicitRole) {
```
WilcoFiers: Return early would look a little cleaner.

```suggestion
  if (explicitRole) {
      const roleSpec = standards.ariaRoles[role];
      return roleSpec?.prohibitedAttrs || [];
  }
```

test/checks/aria/aria-prohibited-attr.js
```diff
@@ -196,6 +196,19 @@ describe('aria-prohibited-attr', () => {
     });
   });
 
+  it('should not allow aria-label on custom element without role', () => {
+    const params = checkSetup(
+      '<custom-elm id="target" aria-label="value"></custom-elm>'
```
WilcoFiers, line 201: Not keen on this change. This is going to start failing properties that are reflected too: https://github.com/dequelabs/axe-core/issues/5217

I think we should avoid failing these for now and tackle it as part of 5217.

lib/standards/html-elms.js
```diff
@@ -163,32 +170,36 @@ const htmlElms = {
     chromiumRole: 'Canvas'
   },
   caption: {
-    allowedRoles: false
+    allowedRoles: false,
+    namingProhibited: true
```
WilcoFiers, line 174: This should get a test

lib/standards/html-elms.js
```diff
@@ -163,32 +170,36 @@ const htmlElms = {
...
     contentTypes: ['phrasing', 'flow'],
-    allowedRoles: true
+    allowedRoles: true,
+    namingProhibited: true
   },
   code: {
     contentTypes: ['phrasing', 'flow'],
-    allowedRoles: true
+    allowedRoles: true,
+    namingProhibited: true
```
WilcoFiers, line 184: Same, test.

lib/standards/html-elms.js
```diff
@@ -163,32 +170,36 @@ const htmlElms = {
...
   colgroup: {
     allowedRoles: false,
-    noAriaAttrs: true
+    allowedAriaAttrs: []
   },
   data: {
     contentTypes: ['phrasing', 'flow'],
-    allowedRoles: true
+    allowedRoles: true,
+    namingProhibited: true
```
WilcoFiers, line 197: Here too. I think all of these new errors we're adding should have a test.

lib/standards/html-elms.js
```diff
@@ -198,11 +209,13 @@ const htmlElms = {
     }
   },
   dd: {
-    allowedRoles: false
+    allowedRoles: false,
+    namingProhibited: true
```
WilcoFiers, line 213: Etc.

lib/standards/html-elms.js
```diff
@@ -703,21 +725,23 @@ const htmlElms = {
...
     allowedRoles: false,
-    noAriaAttrs: true
+    allowedAriaAttrs: []
   },
   picture: {
     // Note: spec change (do not count as embedded), because browsers do not hide text inside the picture element
     contentTypes: ['phrasing', 'flow'],
     allowedRoles: false,
-    noAriaAttrs: true
+    allowedAriaAttrs: ['aria-hidden']
```
WilcoFiers, line 739: I think this needs a test too.

straker: Turns out we can't test this as `aria-prohibited-attr` ignores elements that aren't visible to screen readers

test/integration/rules/aria-allowed-attr/passes.html
```diff
@@ -2180,7 +2180,11 @@
 <div id="pass80" aria-labelledby="  ">Foo</div>
 
 <audio aria-label="value" id="pass81" controls></audio>
-<iframe aria-label="value" id="pass82" src="/test/playground.html"></iframe>
+<iframe
+  aria-label="value"
+  id="pass82"
+  srcdoc="<main><h1>hello</h1></main>"
```
straker, line 2186: Was tired of having this fail whenever I changed the playground code. Now it should always pass

WilcoFiers (approved):

PR #5262, opened 2026-07-27
feat(matches): add inSectioningContent, hasChild, and isSummaryForDetails matches

straker:
This should allow us to handle the variant matching in `htmlElms` needed for header, footer, [figure](https://github.com/dequelabs/axe-core/issues/3443), and [summary](https://github.com/dequelabs/axe-core/issues/3911). 

```js
{
  header: {
    variant: {
      sectioned: {
        matches: {
          inSectioningContent: true
        }
      }
    } 
  },

  figure: {
    variant: {
      figcaption: {
        matches: {
          hasChild: 'figcaption'
        }
      }
    } 
  },

  summary: {
    variant: {
      details: {
        matches: {
          isSummaryForDetails: true
        }
      }
    } 
  }
}
```


lib/commons/matches/first-of-type.js
```diff
@@ -0,0 +1,25 @@
...
+ * matches.firstOfType(vNode, 'button:not([role])');
+ * ```
+ *
+ * @param {VirtualNode} vNode
+ * @param {String} selector
+ * @returns {Boolean}
+ */
+export default function firstOfType(vNode, selector) {
+  const firstMatch = vNode.parent.children.find(node =>
+    matches(node, selector)
```
WilcoFiers: I'm conflicted about this. This is a niche enough feature that I'm not sure we should have a generic method for it. The way you built it queries across shadow trees, which details/summary doesn't do. This is also not how CSS's first-of-type works. That's just a boolean that tells you if that element is the first sibling with that node name. It doesn't take a selector. This is unintuitive to me. 

I wonder whether a dedicated `isSummaryForDetails` function would be more appropriate. I tend to prefer generics where we can get them, but I see no other use case for the function you wrote, and considering it doesn't do a perfect job for the element it is supposed to handle I'm leaning more towards just getting that done correctly.

lib/commons/matches/has-ancestor.js
```diff
@@ -0,0 +1,30 @@
...
+export default function hasAncestor(vNode, selector) {
+  // wildcard selectors allowed to support more complex selectors
+  if (selector.includes('<sectioning_content>')) {
+    selector = selector.replace(
+      '<sectioning_content>',
+      getSectioningContentSelector()
+    );
+  }
+
+  return !!closest(vNode, selector);
```
WilcoFiers: This too isn't quite doing the job it needs to. This `main:not([role]), [role=main]` thing we do is a hack that breaks in a bunch of cases. Matchers was designed as a way to eventually address that problem. I think we need a solution here that actually works.

I thought about other ways of doing ancestry lookup that resolve this, but it gets complicated quickly, and risks infinite loops with acc name -> role -> acc name -> role lookup. I think that's a problem we need to tackle separately.

I think the cleanest way to do this is to just add a `isImplicitLandmark: false` matches that does the ancestry lookup we need, guarding for infinite loops. For other roles we can keep it very dumb. Non-landmarks are false, landmarks other than header/footer are always true.

lib/commons/matches/has-descendant.js
```diff
@@ -0,0 +1,21 @@
...
+ * matches.hasDescendant(vNode, 'main');
+ * matches.hasDescendant(vNode, 'button:not([role])');
+ * ```
+ *
+ * @param {VirtualNode} vNode
+ * @param {String} selector
+ * @returns {Boolean}
+ */
+export default function hasDescendant(vNode, selector) {
+  return querySelectorAllFilter(vNode, selector).length > 0;
```
WilcoFiers: This works for `figcaption`, though I'm not sure what they actually mean with "valid" in `If the figure has a valid figcaption descendant`. Sure aren't testing that. Might be worth pulling up the discussion on when this was added to figure out what the reasoning was behind this. Should this hold if the figcaption is hidden or has a role?

WilcoFiers (dismissed):

lib/commons/matches/first-of-type.js
straker: > The way you built it queries across shadow trees, which details/summary doesn't do

Hu, does that mean the [code I took this from](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/dom/visibility-methods.js#L244-L258) also needs to be updated to not cross shadow DOM trees?

WilcoFiers: Maybe? I forgot to update that comment. Sry. I did more testing and over looked that you can't attach shadow dom to details elements so it doesn't matter much. That leaves the other point though.

lib/commons/matches/has-ancestor.js
straker: Decided that the name would make more sense as `inSectioningContent` rather than `isImplicitLandmark` as the matcher couldn't test the element itself for being a valid landmark target, only what ancestors it has.

lib/commons/matches/in-sectioning-content.js
```diff
@@ -0,0 +1,68 @@
...
+  while (vNode) {
+    const { nodeName } = vNode.props;
+
+    // avoid calling into implicit role (either directly or through getRole) in order
+    // to avoid an infinite loop of calling into the html-elms spec. unfortunately
+    // this means we need to handle element internals and conflict resolution ourselves
+    let role = null;
+    if (!hasConflictResolution(vNode)) {
+      role = getExplicitRole(vNode);
+    }
```
WilcoFiers: **Critical** — conflict resolution is being applied unconditionally.

In `get-role.js`, `hasConflictResolution()` is only ever consulted *after* the explicit role has been found to be `presentation`/`none` (`resolveRole`, lines 138-154). Applying it up front discards the explicit role for *any* focusable element or any element carrying a global ARIA attribute, whatever that role happens to be.

That goes wrong in both directions:

```js
// false negative -- getRole() says "main", this says not sectioning
<div role="main" tabindex="0"><span id="target"></span></div>

// false positive -- getRole() says "button", this says sectioning
<nav role="button" tabindex="0"><span id="target"></span></nav>
```

The `returns true for sectioning element with role conflict` test uses `<main role="group" aria-label="true">`. Since `group` isn't `presentation`/`none`, conflict resolution doesn't apply and the effective role is `group`, so descendants are *not* in sectioning content. That test currently asserts `isTrue`, which locks the bug in.

Suggested shape:

```js
let role = getExplicitRole(vNode);
if (['presentation', 'none'].includes(role) && hasConflictResolution(vNode)) {
  role = null;
}
if (!role && vNode.elementInternals?.role) {
  role = vNode.elementInternals.role;
}
```

Both `getExplicitRole` and `hasConflictResolution` are node-local, so this costs nothing in re-entrancy safety.

Worth adding alongside the flipped assertion: `<div role="main" tabindex="0">` (true), `<nav role="button" tabindex="0">` (false), and `<main role="none" aria-label="x">` (true — conflict resolution restores the implicit role).

lib/commons/matches/in-sectioning-content.js
```diff
@@ -0,0 +1,68 @@
...
+ * @returns {Boolean}
+ */
+export default function inSectioningContent(vNode) {
+  // @see https://html.spec.whatwg.org/multipage/dom.html#sectioning-content
+  // main is not considered sectioning content so we need to add it
+  const sectioningElms = cache.get('sectioningElms', () =>
+    getElementsByContentType('sectioning').concat('main')
+  );
+
+  while (vNode) {
```
WilcoFiers, line 46: **Important** — the walk starts at `vNode` rather than `vNode.parent`, so an element satisfies `inSectioningContent` by being sectioning content itself. An element isn't its own ancestor, so this doesn't match what the name claims.

The shape is inherited from `closest`, which is self-inclusive on purpose because DOM's `Element.closest()` is. That doesn't carry over to an ancestor-only question, and `implicit-html-roles.js` already makes the distinction deliberately:

```js
aside: vNode => {
  if (
    closest(vNode.parent, getSectioningContentSelector()) &&
    !hasAccessibleName(vNode, { checkTitle: true })
  ) {
    return null;
  }
  return 'complementary';
},
```

`aside` passes `vNode.parent` because `<aside>` *is* sectioning content; `header` and `footer` pass `vNode` because they never are. Hard-coding self-inclusive loses that and sets a trap — migrating `aside` to this matcher later would make it match itself unconditionally and never resolve to `complementary`.

No test covers the node itself; every case places the target as a descendant, so this is currently unspecified rather than chosen. Worth adding `isFalse` cases for `<nav id="target">`, `<aside id="target">`, and `<div role="main" id="target">`.

lib/commons/matches/in-sectioning-content.js
```diff
@@ -0,0 +1,68 @@
...
+ * Note: matches.inSectioningContent(vNode) can be indirectly used through
+ * matches(vNode, { inSectioningContent: true })
+ *
+ * Example:
+ * ```js
+ * matches.inSectioningContent(vNode);
+ * ```
+ *
+ * @param {VirtualNode} vNode
+ * @param {String} selector
```
WilcoFiers: Minor JSDoc issues across the three new files:

1. `inSectioningContent` and `isSummaryForDetails` both document `@param {String} selector`, but neither takes a selector — looks copy-pasted from `has-descendant.js`.
2. `isSummaryForDetails` is described as "the first child element of a details element"; it actually checks whether the node is the first `summary` child, which is a different thing.
3. All three carry a stray `` *`` `` line where a blank ` *` belongs. It's copied from `has-accessible-name.js`, so pre-existing, but this triples it.

lib/commons/matches/in-sectioning-content.js
```diff
@@ -0,0 +1,68 @@
...
+  const sectioningElms = cache.get('sectioningElms', () =>
+    getElementsByContentType('sectioning').concat('main')
+  );
+
+  while (vNode) {
+    const { nodeName } = vNode.props;
+
+    // avoid calling into implicit role (either directly or through getRole) in order
+    // to avoid an infinite loop of calling into the html-elms spec. unfortunately
+    // this means we need to handle element internals and conflict resolution ourselves
```
WilcoFiers: The reasoning here is right, but the comment undersells the constraint — "an infinite loop of calling into the html-elms spec" reads as though the ancestor walk is the hazard. It isn't; the walk is strictly upward and terminates.

The path that actually matters is the `<section>` branch of `implicitHtmlRoles`. `<section>` is sectioning content, so an ancestor walk hits one routinely, and resolving its role runs the local `hasAccessibleName` helper → `arialabelledbyText` → full accname → `getElementSpec`. Today that's held shut by a shape check in `get-element-spec.js:33-37` which only bails on a top-level `hasAccessibleName` key — a variant keyed on `inSectioningContent` goes straight past it:

```html
<section aria-labelledby="h">
  <header id="h">text</header>
</section>
```

`getRole(section)` → accname of `#h` → `getElementSpec(header#h, { noMatchAccessibleName: true })` → guard doesn't fire → `inSectioningContent(header#h)` → walks up to the `<section>` → `getRole(section)`. Unbounded, since the idref can point anywhere, and not fixed by starting the walk at the parent.

That's point 4 of #5263 ceasing to hold once relational matchers land, which is what makes that issue a hard prerequisite rather than a cleanup.

Could the comment name the `<section>` accname path specifically and reference #5263 as the removal condition? Otherwise this reads like defensive coding someone could later "simplify" back to `getRole`.

lib/commons/matches/is-summary-for-details.js
```diff
@@ -0,0 +1,25 @@
...
+ * Example:
+ * ```js
+ * matches.isSummaryForDetails(vNode);
+ * ```
+ *
+ * @param {VirtualNode} vNode
+ * @param {String} selector
+ * @returns {Boolean}
+ */
+export default function isSummaryForDetails(vNode) {
```
WilcoFiers: **Important** — this and `inSectioningContent` both ignore their second argument.

`fromDefinition` passes the definition value through as the matcher and returns the result directly:

```js
const matcher = definition[matcherName];
return matchMethod(vNode, matcher);
```

So the value written in the definition is never compared against:

```js
matches(vNode, { inSectioningContent: false }); // true when the node IS in sectioning content
matches(vNode, { isSummaryForDetails: false }); // same
```

This blocks the PR's own motivating case. `header`/`footer` map to `banner`/`contentinfo` only when they are *not* in sectioning content, so the variant needs `inSectioningContent: false` — and today that would silently match the opposite set of elements. `hasAccessibleName` routes through `fromPrimative` for exactly this reason:

```js
return fromPrimative(result, matcher);
```

The `from-definition` tests only exercise `true`, so this passes unnoticed — worth adding `false` cases for both.

lib/commons/matches/has-descendant.js
WilcoFiers: **Important** — following up on my earlier comment about what "valid" means in *"If the figure has a valid figcaption descendant"*: I think this is it.

`querySelectorAllFilter` matches at any depth, so a nested figure's caption satisfies the outer figure:

```html
<figure id="outer">
  <figure><figcaption>inner caption</figcaption></figure>
</figure>
```

`hasDescendant: 'figcaption'` returns true for `#outer`, but that caption belongs to the inner figure.

We already solve this for figure *naming*, and the helper's comment names the identical hazard:

```js
function descendantText(nodeName, { actualNode }, context) {
  nodeName = nodeName.toLowerCase();
  // Prevent accidently getting the nested element, like:
  // fieldset > fielset > legend (1st fieldset has no legend)
  const nodeNames = [nodeName, actualNode.nodeName.toLowerCase()].join(',');
  const candidate = actualNode.querySelector(nodeNames);

  if (!candidate || candidate.nodeName.toLowerCase() !== nodeName) {
    return '';
  }
  return accessibleText(candidate, context);
}
```

It queries for the caption *and* the container node name together and takes whichever comes first in document order — if a nested `figure` wins, there's no caption for the outer one. `figureText`, `tableCaptionText`, and `fieldsetLegendText` all rely on it. As written, `hasDescendant` would make `allowedRoles` disagree with `namingMethods` on the same element.

Either carry that idiom into the matcher (take a boundary node name, stop at the nearest matching container), or scope it to direct children — which HTML's content model requires for `figcaption`, `caption`, and `legend` anyway. Either way, `<figure><figure><figcaption>` wants a test asserting the outer figure doesn't match.

test/commons/matches/in-sectioning-content.js
```diff
@@ -0,0 +1,190 @@
...
+  const sectioningElms = axe.commons.standards
+    .getElementsByContentType('sectioning')
+    .concat('main');
+  const sectioningRoles = [
+    'article',
+    'complementary',
+    'main',
+    'navigation',
+    'region'
+  ];
```
WilcoFiers: Deriving `sectioningElms` from `getElementsByContentType('sectioning')` and re-declaring `sectioningRoles` with the same literal list the implementation uses means both loops restate the implementation rather than pinning down expected behaviour. If `htmlElms` gains or loses a sectioning element, these tests silently follow along instead of failing.

Hard-coding the expected list (`article`, `aside`, `nav`, `section`, `main`) would surface such a change as a failure that has to be consciously accepted.

WilcoFiers (dismissed):
Nice direction overall — expressing the `header`/`footer`, `figure`, and `summary` variants as matchers is the right shape, and hand-rolling the role lookup instead of calling `getRole` is the correct call given #5263.

Two correctness issues need resolving before this can be the foundation for the `htmlElms` work, plus a semantics question on the ancestor walk.

**Critical**

- `inSectioningContent` applies conflict resolution unconditionally, so any focusable or globally-labelled element loses its explicit role. One test currently asserts the incorrect result.

**Important**

- `inSectioningContent` and `isSummaryForDetails` ignore the matcher argument, so `{ inSectioningContent: false }` can't be expressed — which is exactly what the `header`/`footer` banner variant needs.
- `hasDescendant` counts a nested figure's caption for the outer figure.
- The ancestor walk is self-inclusive, so sectioning elements match themselves.

**Suggestions**

- JSDoc corrections, test independence from the standards data, and expanding the loop-avoidance comment to name #5263 as its removal condition.

Details inline.

lib/commons/matches/is-summary-for-details.js
```diff
@@ -0,0 +1,28 @@
...
+ * ```
+ *
+ * @param {VirtualNode} vNode
+ * @param {Object} matcher
+ * @returns {Boolean}
+ */
+export default function isSummaryForDetails(vNode, matcher) {
+  if (!vNode.parent || vNode.parent.props.nodeName !== 'details') {
+    return false;
+  }
```
chutchins25, line 28: **Important** — the `fromPrimative` fix landed on the return path but not on the early
return, so the matcher argument is still ignored on the bail-out branch:

```js
if (!vNode.parent || vNode.parent.props.nodeName !== 'details') {
  return false;   // <- never consults `matcher`
}
```

`fromPrimative(false, false)` is `true`, so `matches(vNode, { isSummaryForDetails: false })`
on a `<summary>` inside a `<div>` returns `false` when it should return `true` — the
negated form silently fails to match exactly the elements it is meant to. `inSectioningContent`
got both paths; this one got only the happy path.

The `allows passing false` test only exercises the positive case, so it passes.

```suggestion
  if (!vNode.parent || vNode.parent.props.nodeName !== 'details') {
    return fromPrimative(false, matcher);
  }
```

Worth adding alongside: `<div><summary id="target"></summary></div>` with `false` → `isTrue`.

lib/commons/matches/has-child.js
```diff
@@ -0,0 +1,21 @@
...
+ * matches.hasChild(vNode, 'button:not([role])');
+ * ```
+ *
+ * @param {VirtualNode} vNode
+ * @param {String} selector
+ * @returns {Boolean}
+ */
+export default function hasChild(vNode, selector) {
+  return vNode.children.some(node => matches(node, selector));
+}
```
chutchins25, line 21: **Important** — `vNode.children` has no default, so this throws on a `SerialVirtualNode`
that never had children assigned. `AbstractVirtualNode`'s constructor sets only `parent`:

```js
class AbstractVirtualNode {
  constructor() {
    this.parent = undefined;
  }
```

So `axe.runVirtualRule('figure', { nodeName: 'figure' })` — with a `figure` variant keyed
on `hasChild: 'figcaption'`, which is this PR's motivating case — gets
`TypeError: Cannot read properties of undefined (reading 'some')` out of `getElementSpec`.
The test only covers the case where `children` was assigned by hand.

There's an established idiom for this; five call sites guard explicitly —
`commons/aria/get-owned-virtual.js:11-13`, `commons/dom/has-lang-text.js:13`,
`checks/tables/same-caption-summary-evaluate.js:6`, `checks/lists/invalid-children-evaluate.js:11`,
`checks/lists/structured-dlitems-evaluate.js:3`. `detailsHidden` in `visibility-methods.js:250`
does `vNode.parent.children.find(...)` unguarded, but that only runs on the real-DOM
visibility path where `children` is always an array.

```suggestion
export default function hasChild(vNode, selector) {
  if (typeof vNode.children === 'undefined') {
    throw new TypeError('Cannot resolve children for non-DOM nodes');
  }

  return vNode.children.some(node => matches(node, selector));
}
```

`isSummaryForDetails` has the same exposure through `vNode.parent.children`. I've asked
separately (on `in-sectioning-content.js`) what the intended contract is when the subtree
can't be resolved — whichever way that lands should apply to both of these, so treat the
`throw` above as one option rather than a recommendation.

A `SerialVirtualNode`-without-children case in both test files would pin it either way.

lib/commons/matches/in-sectioning-content.js
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    // the top node of the tree will have parent === null, so a undefined parent means
+    // we are in a disconnected tree
+    if (typeof vNode.parent === 'undefined') {
+      throw new TypeError('Cannot resolve parent for non-DOM nodes');
+    }
+
+    vNode = vNode.parent;
+  }
```
chutchins25: **Important** — this guard is unreachable in the shallowest case. Now that the walk starts
at `virtualNode.parent`, a node whose own parent is `undefined` never enters the loop:

```js
let vNode = virtualNode.parent;
while (vNode) {           // undefined -> body never runs
  ...
  if (typeof vNode.parent === 'undefined') {
    throw new TypeError('Cannot resolve parent for non-DOM nodes');
  }
  vNode = vNode.parent;
}
return fromPrimative(false, matcher);   // silently "not in sectioning content"
```

So `axe.runVirtualRule('header', { nodeName: 'header' })` reports *not* in sectioning
content — resolving `header` to `banner` — while a node two levels deep in the same
unresolvable tree throws. On the previous revision the loop started at `vNode`, so the
guard fired on the first iteration; moving the start to `.parent` opened the gap.

Which answer is intended for virtual rules? If it's `false`, the throw is doing more harm
than good, since it only fires at depth ≥ 2. If it's throw, it wants hoisting above the loop
so it covers the node's own parent.

The `throws if in incomplete tree` test only covers the deep shape (`child2`, whose
grandparent's parent is `undefined`), so the single-node case isn't pinned either way.

lib/commons/matches/in-sectioning-content.js
```diff
@@ -0,0 +1,60 @@
...
+  // @see https://html.spec.whatwg.org/multipage/dom.html#sectioning-content
+  // main is not considered sectioning content so we need to add it
+  const sectioningElms = cache.get('sectioningElms', () =>
+    getElementsByContentType('sectioning').concat('main')
+  );
+
+  let vNode = virtualNode.parent;
+  while (vNode) {
+    const { nodeName } = vNode.props;
+    const role = getRole(vNode);
```
chutchins25: `getRole` is the right fix for the conflict-resolution bug — it gets `<main role="none" aria-label>`
and the element-internals case correct where the hand-rolled version didn't. But 97c339e9
also deleted the comment that recorded *why* the lookup was hand-rolled, and the re-entrancy
path is back with nothing in the file noting it:

```diff
-    // avoid calling into implicit role (either directly or through getRole) in order
-    // to avoid an infinite loop of calling into the html-elms spec. unfortunately
-    // this means we need to handle element internals and conflict resolution ourselves
+    const role = getRole(vNode);
```

`<section>`'s implicit role runs the local `hasAccessibleName` → `arialabelledbyText` →
full accname on the referenced element → `getElementSpec`. The guard at
`get-element-spec.js:33-37` only bails on a top-level `hasAccessibleName` key:

```js
for (let i = 0; i < matchProperties.length && noMatchAccessibleName; i++) {
  if (matchProperties[i].hasOwnProperty('hasAccessibleName')) {
    return standard;
  }
}
```

An `inSectioningContent`-keyed variant walks straight past it, so once `header`'s variant
lands:

```html
<section aria-labelledby="h">
  <header id="h">text</header>
</section>
```

`getRole(section)` → accname of `#h` → `getElementSpec(header#h, { noMatchAccessibleName: true })`
→ guard doesn't fire → `inSectioningContent(header#h)` → walks up to the `<section>` →
`getRole(section)`. Unbounded, since the idref can point anywhere, and not fixed by
starting the walk at the parent.

Harmless today — nothing in `htmlElms` consumes the matcher yet. Was dropping the
hand-rolled lookup a deliberate trade (correctness now, re-entrancy deferred to #5263)?
If so, could that be recorded — a comment naming the `<section>` accname path and #5263 as
the removal condition, or a note on #5263 that it now blocks relational variants in
`htmlElms`? Otherwise the ordering is easy to lose when someone wires up the
`header`/`footer` variant.

chutchins25 (commented):
Nice work — replacing `hasAncestor`/`hasDescendant`/`firstOfType` with these three
matchers is the right shape, and switching `inSectioningContent` to `getRole` is the
correct fix for the conflict-resolution bug from the last round (it gets
`<main role="none" aria-label>` and the element-internals case right, where the
hand-rolled version didn't). Confirmed resolved on `ef121b9e`: the self-inclusive
walk, the JSDoc `@param`s, the tests deriving their expectations from the standards
data, and the nested-`figcaption` problem (`hasChild` being direct-children-only
sidesteps it entirely).

Four things, two of them concrete bugs:

1. `isSummaryForDetails` got the `fromPrimative` fix on its return path but not its
   early return, so the negated form still reports the wrong answer.
2. `hasChild` throws a `TypeError` on a `SerialVirtualNode` with no `children` — the
   virtual-rule path this is being built for.
3. `inSectioningContent`'s disconnected-tree guard is unreachable for a node whose own
   parent is `undefined`, so it throws at depth ≥ 2 but silently returns `false` at
   depth 1. Which answer is intended?
4. Dropping the hand-rolled role lookup reopens the `<section>` accname re-entrancy
   path, and the comment that recorded why it was hand-rolled went with it. Nothing
   misbehaves today, but it makes #5263 a prerequisite for the `htmlElms` work rather
   than a cleanup — worth writing that down somewhere.

Details inline.

lib/commons/matches/has-child.js
```diff
@@ -0,0 +1,25 @@
...
+ * @param {VirtualNode} vNode
+ * @param {String} selector
+ * @returns {Boolean}
+ */
+export default function hasChild(vNode, selector) {
+  if (!vNode.children) {
+    throw new TypeError('Cannot resolve children for non-DOM nodes');
+  }
+
+  return vNode.children.some(node => matches(node, selector));
```
WilcoFiers, line 20: Why did you pivot to testing child nodes? The spec on figure says "valid figcaption descendants".

lib/commons/matches/has-child.js
```diff
@@ -0,0 +1,25 @@
...
+ * matches.hasChild(vNode, 'button:not([role])');
+ * ```
+ *
+ * @param {VirtualNode} vNode
+ * @param {String} selector
+ * @returns {Boolean}
+ */
+export default function hasChild(vNode, selector) {
+  if (!vNode.children) {
+    throw new TypeError('Cannot resolve children for non-DOM nodes');
```
WilcoFiers: I don't think this needs a guard, and would prefer we don't make a habit of guarding the .children / .parent prop on vNode if all we do is throw an error anyway. Am I missing something?

lib/commons/matches/is-summary-for-details.js
```diff
@@ -1,25 +1,38 @@
...
+    throw new TypeError('Cannot resolve parent for non-DOM nodes');
+  }
+
   if (!vNode.parent || vNode.parent.props.nodeName !== 'details') {
-    return false;
+    return fromPrimative(false, matcher);
+  }
+
+  if (!vNode.parent.children) {
+    throw new TypeError('Cannot resolve children for non-DOM nodes');
```
WilcoFiers: Don't think we need this guard

lib/standards/html-elms.js
```diff
@@ -317,6 +317,17 @@ const htmlElms = {
     noAriaAttrs: true
   },
   header: {
+    variant: {
+      sectioned: {
+        matches: {
+          inSectioningContent: true
+        },
+        role: 'generic'
```
WilcoFiers: Axe-core doesn't support `generic` yet. We won't know what to do with this role.

lib/standards/html-elms.js
WilcoFiers, line 1: Did you mean to lave this in? I assumed this would be a separate PR. If not, put all of the new ones in and add tests.

WilcoFiers (dismissed):

lib/commons/matches/has-child.js
straker: https://github.com/dequelabs/axe-core/pull/5262#discussion_r3682107768

> Either carry that idiom into the matcher (take a boundary node name, stop at the nearest matching container), __or scope it to direct children__ — which HTML's content model requires for figcaption, caption, and legend anyway. Either way, <figure><figure><figcaption> wants a test asserting the outer figure doesn't match.

lib/standards/html-elms.js
straker: Sorry, that wasn't suppose to be in. Was testing the infinite loop

WilcoFiers (approved):

PR #5273, opened 2026-07-31
fix(get-selector): escape control characters in attribute selectors

### LINKED ISSUE #5204: generateSelector builds an invalid attribute selector when an attribute value contains a control character (e.g. form feed U+000C) — entire axe.run throws
## Summary

When an element's attribute value contains a CSS-significant control character (observed: form feed, `U+000C`), the selector generator emits the raw character into an attribute selector. The resulting selector is invalid, `Element.matches` throws inside `getNthChildString`, and the whole `axe.run()` call fails — one bad attribute value on a page kills the entire audit.

axe-core 4.11.3 via @axe-core/playwright 4.11.2 (Chromium 147).

## Reproduction

```html
<img src="x.png" alt=" &#12;¼">  <!-- alt contains a raw form feed (U+000C) -->
```

Run `new AxeBuilder({ page }).withTags(["wcag2a","wcag2aa"]).analyze()` on a page containing that element:

```
frame.evaluate: SyntaxError: Failed to execute 'matches' on 'Element': 'img[alt=" ¼"]' is not a valid selector.
    at getNthChildString (…:10864:42)
    at generateSelector (…:10952:23)
    at _getShadowSelector (…:10739:16)
    at get selector (…:11210:40)
    at Object.toJSON (…:11223:26)
```

Minimal browser-console proof that the selector is invalid and that `CSS.escape` fixes it:

```js
document.body.matches('img[alt=" \f¼"]')
// SyntaxError: … 'img[alt=" \f¼"]' is not a valid selector

document.body.matches(`img[alt=" ${CSS.escape('\f')}¼"]`)
// false (valid selector, no throw)
```

## Expected

Attribute values interpolated into generated selectors should be CSS-escaped (`CSS.escape` or equivalent) so a hostile/garbage attribute value cannot make the audit throw; and/or a selector-generation failure for one node should degrade to a fallback selector instead of failing the whole run.

## Impact

Found by an automated post-deploy a11y crawl: a single listing card whose `alt` text contained mojibake with a form feed made axe unable to audit the page at all, silently reducing a11y coverage.

Related (older, closed) reports of the same class with newlines/quotes in attribute values: #1137, #1165.

chutchins25:
Fixes #5204. When an element's attribute value contains a C0 control character (observed: form feed `U+000C`), `generateSelector` emitted it raw into an attribute selector. In CSS, form feed is a string-terminating newline, so `img[alt=" \f¼"]` is **invalid**, `Element.matches` throws inside `getNthChildString`, and the **entire `axe.run()` fails** — one bad attribute value on a page kills the whole audit.

## Fix

`escapeAttribute` (in `lib/core/utils/get-selector.js`) previously escaped only `\`, `"`, and CR/LF newlines — missing form feed and every other control character. It now escapes **all C0 control characters (`U+0000–U+001F`) and DEL (`U+007F`)** as CSS numeric escapes (`\<hex> `), which subsumes the existing newline handling (`U+000A` → `\a `, `U+000C` → `\c `). This matches the issue's minimal proof that `CSS.escape('\f')` produces a valid selector.

## Compatibility

No behavior change for existing cases: newline (`U+000A`) still escapes to `\a `. CR and CRLF now escape to `\d `/`\d \a ` (previously both collapsed to `\a `) — more correct round-tripping, and not covered by any existing test. All existing `get-selector` escape tests still pass.

## Test

`test/core/utils/get-selector.js` — an attribute value containing form feed, vertical tab, and `U+001F` now produces a valid selector that both matches the expected escaped string and passes `matchesSelector` (which threw on the pre-fix code).

Confirmed by @straker on the issue.

Closes #5204

### DIFF
--- a/lib/core/utils/get-selector.js
+++ b/lib/core/utils/get-selector.js
@@ -30,7 +30,11 @@ const ignoredAttributes = [
 
 const MAXATTRIBUTELENGTH = 31;
 const attrCharsRegex = /([\\"])/g;
-const newlineChars = /(\r\n|\r|\n)/g;
+// C0 control characters (U+0000–U+001F) and DEL (U+007F) cannot appear raw
+// inside a CSS string. In particular the CSS newlines U+000A, U+000C (form
+// feed) and U+000D terminate the string, producing an invalid selector that
+// throws in `Element.matches`. Escape each as a CSS numeric escape (`\<hex> `).
+const controlCharsRegex = /[\u0000-\u001f\u007f]/g;
 
 /**
  * Escape an attribute selector string.
@@ -43,7 +47,10 @@ function escapeAttribute(str) {
       // @see https://www.py4u.net/discuss/286669
       .replace(attrCharsRegex, '\\$1')
       // @see https://stackoverflow.com/a/20354013/2124254
-      .replace(newlineChars, '\\a ')
+      .replace(
+        controlCharsRegex,
+        char => '\\' + char.charCodeAt(0).toString(16) + ' '
+      )
   );
 }
 
--- a/test/core/utils/get-selector.js
+++ b/test/core/utils/get-selector.js
@@ -545,6 +545,46 @@ describe('axe.utils.getSelector', () => {
     assert.isTrue(axe.utils.matchesSelector(div2, expected));
   });
 
+  it('should escape control characters in attribute (issue 5204)', () => {
+    const div1 = document.createElement('div');
+    div1.setAttribute('data-thing', 'foobar');
+
+    const div2 = document.createElement('div');
+    // Form feed (U+000C), vertical tab (U+000B) and U+001F previously passed
+    // through raw, producing an invalid selector that threw in matches().
+    div2.setAttribute(
+      'data-thing',
+      ' ' + String.fromCharCode(0x0c, 0x0b, 0x1f)
+    );
+
+    const expected = 'div[data-thing=" \\c \\b \\1f "]';
+    fixtureSetup([div1, div2]);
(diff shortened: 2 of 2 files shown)


test/core/utils/get-selector.js
```diff
@@ -545,6 +545,24 @@ describe('axe.utils.getSelector', () => {
...
+    div2.setAttribute(
+      'data-thing',
+      ' ' + String.fromCharCode(0x0c, 0x0b, 0x1f)
+    );
+
+    const expected = 'div[data-thing=" \\c \\b \\1f "]';
+    fixtureSetup([div1, div2]);
+    assert.equal(axe.utils.getSelector(div2), expected);
+    assert.isTrue(axe.utils.matchesSelector(div2, expected));
+  });
```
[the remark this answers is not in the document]
chutchins25: Added. New test "should escape CR and CRLF in attribute" asserts CR → `\d ` and CRLF → `\d \a `, locking in the newline behavior this change subsumes.

lib/core/utils/get-selector.js
```diff
@@ -39,11 +43,14 @@ const newlineChars = /(\r\n|\r|\n)/g;
  */
 function escapeAttribute(str) {
   return (
-    str
+    String(str)
```
straker: Why the call to cast as String?

chutchins25: Good call — it's not needed. Every caller passes an attribute value (`at.value` / `getAttribute()`), which is always a string, so I've dropped the `String()` cast.

straker (approved):

PR #5285, opened 2026-08-05
fix(aria-prohibited-attr): visible aria-labelledby requires review only

### LINKED ISSUE #4118: Should aria-labelledby be prohibited when its reference is invalid or visible?
I think `aria-prohibited-attr` is a little too strict for a rule that tests for WCAG. First up, just like how we don't fail `aria-prohibited-attr` when `aria-label` is empty, I think we should pass when `aria-labelledby` results in an empty string:

```html
<div aria-labelledby="d1"><!-- Pass this --></div>
<div id="d1" aria-labelledby="nowhere"><!-- Pass this --></div>
```

Closely related, I think that if the referenced element exists and is in the accessibility tree axe should report that as incomplete, similar to how it reports incomplete when the element has content. 

```html
<div aria-labelledby="h1"><!-- Incomplete this, h1 is visible --></div>
<h1 id="h1">Hello world</h1>
<div aria-labelledby="h2"><!-- Fail this, h2 is hidden --></div>
<h2 id="h2" aria-hidden="true">Hello world</h2>
```

## Optional

When we look at the subtree of the element to decide if its a fail or incomplete, we could check if the text of the label exists in the subtree, in which case instead of reporting it as incomplete we could quite reasonably pass this. That would cut down on the amount of incompletes this rule generates.

### LINKED ISSUE #5180: Let aria-prohibited-attrs incomplete on visible aria-labelledby references
The following code should be incomplete (needs review) for `aria-prohivited-attr`:

```html
<div aria-labelledby="hd">
  <h1 id="hd">Hello world</h1>
</div>
```

When an element uses a prohibited `aria-labelledby` attribute, AND at least one of its references is visible to screen readers, report this as needs review instead of as a violation.

This is part of the `aria-prohibited-attr-evaluate` function.

WilcoFiers:
If the only prohibited attribute is aria-labelledby, it doesn't reference hidden elements, mark the element as needs review. Ignore aria-label as a prohibited attribute if this is the case, as the acc name fallback would have skipped it anyway.

Closes: https://github.com/dequelabs/axe-core/issues/5180, closes https://github.com/dequelabs/axe-core/issues/4118

### DIFF
--- a/test/checks/aria/aria-prohibited-attr.js
+++ b/test/checks/aria/aria-prohibited-attr.js
@@ -307,7 +307,7 @@ describe('aria-prohibited-attr', () => {
       assert.deepEqual(checkContext._data, {
         nodeName: 'div',
         role: null,
-        messageKey: 'noRoleSingular',
+        messageKey: 'visibleLabelSingular',
         prohibited: ['aria-labelledby']
       });
     });
@@ -341,7 +341,7 @@ describe('aria-prohibited-attr', () => {
       assert.deepEqual(checkContext._data, {
         nodeName: 'code',
         role: null,
-        messageKey: 'noRolePlural',
+        messageKey: 'visibleLabelPlural',
         prohibited: ['aria-label', 'aria-labelledby']
       });
     });
@@ -355,7 +355,7 @@ describe('aria-prohibited-attr', () => {
       assert.deepEqual(checkContext._data, {
         nodeName: 'div',
         role: 'code',
-        messageKey: 'hasRoleSingular',
+        messageKey: 'visibleLabelSingular',
         prohibited: ['aria-labelledby']
       });
     });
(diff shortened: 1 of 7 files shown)


test/checks/aria/aria-prohibited-attr.js
```diff
@@ -294,4 +296,205 @@ describe('aria-prohibited-attr', () => {
...
+
+    it('should return undefined for a visible reference on a custom element', () => {
+      const params = checkSetup(html`
+        <h1 id="hd">Hello world</h1>
+        <my-element id="target" role="code" aria-labelledby="hd"></my-element>
+      `);
+      assert.isUndefined(checkEvaluate.apply(checkContext, params));
+    });
+
+    it('should return undefined for a visible reference inside shadow DOM', () => {
```
straker, line 488: Should we add a test case for the element referencing across shadow DOM boundaries? It would be the same as missing reference, but would make sure `aria-labelledby` isn't crossing boundaries in the prohibited lookup

straker (approved):
Small question, but doesn't block the pr.

Issue #4678, opened 2025-01-23
label-content-name-mismatch wrongly includes invisible text


jribbens:
### Product

axe-core

### Product Version

4.10.2

### Latest Version

- [x] I have tested the issue with the latest version of the product

#### Expectation
Given the following HTML code, axe _should not_ report an label-content-name-mismatch violation:
```html
<a aria-label="deque are great" href="https://www.deque.com/">
  deque <span class="visually-hidden">labs</span> are great
</a>
```
Given the following HTML code, axe _should_ report an label-content-name-mismatch violation:
```html
<a href="https://www.deque.com/">
  deque <span class="visually-hidden">labs</span> are great
</a>
```

#### Actual
It emits a diagnostic on the code which is correct, and does not emit one on the code which is incorrect.

#### How to Reproduce
See above. Or also: https://unequivocal.eu/label-content-name-mismatch.html

#### Additional context
The rule is supposed to be comparing the accessible name with the _visible_ text of the item, but it seems to be comparing all of the text. Also it only checks elements with an `aria-label` or `aria-labelledby`, which is incorrect as the rule can fail even without an ARIA label (i.e. the second example above). See https://www.w3.org/WAI/WCAG21/Techniques/failures/F96 for example.

dylanb:
@jribbens what is the CSS for the `visually-hidden` class in both cases?

WilcoFiers:
Thank you for reporting this @jribbens. This is an interesting case. I certainly agree the first examples shouldn't fail. I'm not entirely sure we can reliably fail the second scenario. We'll have to think bout that when we're looking at how to solve for this.

WilcoFiers:
For later reference. I've done some testing of this today, and at least MacOS's Voice Control seems to require announcing the full accessible name. So hidden texts are kind of a problem.

chutchins25:
The false positive here — the rule including the visually-hidden text — was fixed in #5096, which switched the check to compare against `visibleVirtual` (visible-on-screen text) instead of `subtreeText`. Verified on axe 4.12.1: the first example (`aria-label="deque are great"` with a `.visually-hidden` "labs") now passes. #5291 adds the regression tests that #5096 didn't include, so this can close.

[...]

Issue #5203, opened 2026-07-13
Get label-content-name-mismatch consistent with ACT's equivalent


WilcoFiers:
Axe-core is currently set to partially consistent for this ACT rule:

https://www.w3.org/WAI/standards-guidelines/act/rules/2ee8b8/proposed/

Part of that comes from us not having updated in a while. That needs to be sorted. But after sorting that failed example 4 isn't failed by Axe-core.

chutchins25:
Cross-referencing #4678, which includes a name-from-content case axe currently doesn't evaluate at all:

```html
<a href="#">deque <span class="visually-hidden">labs</span> are great</a>
```

With no `aria-label`/`aria-labelledby`, `label-content-name-mismatch-matches` bails, so axe never fails it — even though the visible label ("deque are great") isn't contained in the accessible name ("deque labs are great", since the SR-visible hidden text is part of the name). That's the same shape as "failed example 4 isn't failed by Axe-core": axe's applicability is narrower than 2ee8b8. Flagging it here so it's covered when the rule is brought in line with the ACT rule.

straker:
`<a href="#">deque <span class="visually-hidden">labs</span> are great</a>` should never fail according to the ACT rule. The rule's applicability section states (emphasis mine):

> This rule applies to any element for which all the following is true:

> * The element has a [semantic role](https://www.w3.org/WAI/standards-guidelines/act/rules/2ee8b8/proposed/#semantic-role) that is a [widget](https://www.w3.org/TR/wai-aria-1.2/#widget_roles) that [supports name from content](https://www.w3.org/TR/wai-aria-1.2/#namefromcontent); and
> * The element has [visible text content](https://www.w3.org/WAI/standards-guidelines/act/rules/2ee8b8/proposed/#visible-text-content); and
> * __The element has an aria-label or aria-labelledby attribute; and__
> * Neither the element’s accessible name nor its visible label contains any abbreviations; and
> * Every word that appears in both the element’s accessible name and its visible label uses the same spelling and hyphenation in both places.

Since the element does not have an `aria-label(ledby)` attribute we don't run the rule against it. So the example provided should just be an ignored case rather than something we need to cover. The matcher for the rule already has tests to cover missing attributes so I don't think there's anything we need to do to handle this.

straker:
[...]

PR #5291, opened 2026-08-06
test(label-content-name-mismatch): cover visually-hidden text exclusion

chutchins25:
## Summary

The false positive in #4678 — the rule "wrongly includes invisible text" — was fixed in #5096, which switched the check from `subtreeText` (all text) to `visibleVirtual` (visible-on-screen text). That fix shipped without regression coverage. This PR adds it, at the integration level.

## Tests added

Three hidden-text techniques, each hiding the word `labs` inside the visible label:

| Fixture | Technique | Expected |
| --- | --- | --- |
| `#pass10` | Bootstrap `.visually-hidden` — `position:absolute`, 1px, `overflow:hidden`, `clip` | pass |
| `#pass11` | `clip` only, normal dimensions — isolates `clipHidden` from `overflowHidden` | pass |
| `#pass12` | `clip-path: inset(50%)` — what Bootstrap 5 and Tailwind `sr-only` ship today | pass |
| `#fail9` | hidden text *interrupting* the visible label | violation |

**These are discriminating.** Reverting the check to its pre-#5096 state (`subtreeText` for `visibleVirtual`) and rebuilding turns all four red.

## `#fail9` is a genuine fail

@scottmries settled the open question from the last round:

> `#fail8` as a **fail** is right — don't switch it to incomplete. Visible label `deque are great` isn't a contiguous run within `deque labs are great`, which is a genuine 2.5.3 concern and what ACT 2ee8b8 expects. It also stays a fail under #5302's word-level matching, so the two PRs agree.

That last clause is no longer a prediction — word-level matching landed on `develop` in #5345, and the fixture still reports a violation.

## Rebased and renumbered

Rebased onto `develop`. #5345 took `#fail8` for its hyphenation case, so this PR's fail fixture is now **`#fail9`**.

## Scope

#4678 was closed as completed by @straker on 2 September. Its second defect — name-from-content elements with no `aria-label`/`aria-labelledby` never reaching the check — is tracked separately as #5339.

Refs #4678


test/integration/rules/label-content-name-mismatch/label-content-name-mismatch.html
```diff
@@ -26,6 +26,20 @@
...
+      position: absolute;
+      width: 1px;
+      height: 1px;
+      overflow: hidden;
+      clip: rect(0px, 0px, 0px, 0px);
+    "
+    >labs</span
+  >
+  are great
+</a>
```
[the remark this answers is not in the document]
chutchins25: False positive. That's Prettier's HTML formatting, not a stray `>` — whitespace before a tag's closing `>` is valid per the HTML parser and adds no rendered whitespace, so the span content is exactly `labs`. The `#pass10` integration case parses and passes green. Collapsing it to one line would just be reverted by the pre-commit `fmt` hook.

test/checks/label/label-content-name-mismatch.js
```diff
@@ -224,4 +225,35 @@ describe('label-content-name-mismatch tests', () => {
...
+  it('returns true when a visually-hidden child inside shadow DOM is excluded from the visible text', () => {
+    const vNode = queryShadowFixture(
+      '<div id="shadow"></div>',
+      '<a id="target" href="#" aria-label="deque are great">' +
+        'deque <span style="position:absolute;width:1px;height:1px;overflow:hidden;clip:rect(0px,0px,0px,0px);">labs</span> are great' +
+        '</a>'
+    );
+    const actual = check.evaluate(vNode.actualNode, options, vNode);
+    assert.isTrue(actual);
+  });
```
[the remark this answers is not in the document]
chutchins25: Applied in bb057609. Extracted the span into a shared `visuallyHiddenSpan` constant used by all three tests, so they now differ only by their surrounding text.

test/checks/label/label-content-name-mismatch.js
```diff
@@ -224,4 +227,41 @@ describe('label-content-name-mismatch tests', () => {
...
+        'deque ' +
+        visuallyHiddenSpan +
+        ' are truly great' +
+        '</a>'
+    );
+    const actual = check.evaluate(vNode.actualNode, options, vNode);
+    assert.isFalse(actual);
+  });
+
+  it('returns true when a visually-hidden child inside shadow DOM is excluded from the visible text', () => {
```
straker: Can remove this test as it's not crossing shadow DOM boundaries.

straker (dismissed):

test/checks/label/label-content-name-mismatch.js
chutchins25: Agreed — removed in 3bf48f1b. The whole element was inside one shadow root, so it wasn't crossing a boundary; the light-DOM cases already cover this.

test/checks/label/label-content-name-mismatch.js
```diff
@@ -224,4 +226,28 @@ describe('label-content-name-mismatch tests', () => {
       assert.isFalse(actual);
     }
   );
+
+  it('returns true when a visually-hidden child is excluded from the visible text', () => {
```
scottmries: **Important** — shadow DOM coverage was dropped rather than corrected.

straker was right that the removed test never crossed a boundary, but the fix was to make it cross, not delete it. `visibleVirtual` recurses over `vNode.children` — the flattened tree — so shadow content is exactly the path that can regress here, and `CLAUDE.md` §2 asks for an open Shadow DOM case on relevant check changes.

Putting the hidden span inside an open shadow root on a *child* of the labelled element is a genuine guard:

```js
fixture.innerHTML =
  '<a id="target" href="#" aria-label="deque are great">deque <span id="host"></span> are great</a>';
fixture.querySelector('#host').attachShadow({ mode: 'open' }).innerHTML = visuallyHiddenSpan;
```

I verified this returns `true` on this branch and `false` on pre-#5096 code.

test/checks/label/label-content-name-mismatch.js
```diff
@@ -224,4 +226,28 @@ describe('label-content-name-mismatch tests', () => {
...
+        'deque ' +
+        visuallyHiddenSpan +
+        ' are great' +
+        '</a>'
+    );
+    const actual = check.evaluate(vNode.actualNode, options, vNode);
+    assert.isTrue(actual);
+  });
+
+  it('returns false when the visible text alone is not contained in the accessible name', () => {
```
scottmries: **Suggestion** — this one isn't a regression guard. It passes on pre-#5096 code too: with the hidden text included the subtree is `"deque labs are truly great"`, which also fails containment, so the assertion can't distinguish visible-only from subtree comparison.

The PR body doesn't claim otherwise, but the name implies a guarantee the test doesn't provide. Either rename it as a mismatch control, or strengthen it with the markup from my comment on `#pass10`, which does distinguish the two implementations.

test/integration/rules/label-content-name-mismatch/label-content-name-mismatch.html
```diff
@@ -26,6 +26,20 @@
   </svg>
   Hello World
 </a>
+<a id="pass10" href="#" aria-label="deque are great">
```
scottmries, line 47: **Important** — #5096 also flipped a sibling of this pattern from pass to violation, and it's left unpinned.

The check compares by contiguous substring (`accText.includes(visibleText)`), so visually-hidden text in the *middle* of the visible text now breaks the match whenever the accessible name includes it:

```html
<a aria-label="deque labs are great">deque <span class="visually-hidden">labs</span> are great</a>
```

Verified: `false` (violation) on this branch, `true` on pre-#5096 code. That's plausible markup — e.g. `<a aria-label="Deque Labs home">Deque <span class="visually-hidden">Labs</span> home</a>`.

Since this is the PR that closes #4678, worth either a fixture pinning the behavior (`#fail`, or `#incomplete` if it's considered undecided) or a line in "Out of scope" with a follow-up, the way #5203 was deferred.

test/integration/rules/label-content-name-mismatch/label-content-name-mismatch.html
```diff
@@ -26,6 +26,20 @@
...
+  <span
+    style="
+      position: absolute;
+      width: 1px;
+      height: 1px;
+      overflow: hidden;
+      clip: rect(0px, 0px, 0px, 0px);
+    "
+    >labs</span
+  >
```
scottmries: **Suggestion** — Prettier expands this to eight lines, which buries the point of the fixture (hidden text between two visible words).

A `.visually-hidden` class in a `<style>` block matches what `target-size.html`, `link-in-text-block.html`, and `color-contrast/text-shadows.html` already do, reads closer to the Bootstrap repro in #4678, and avoids the formatting noise that drew the false-positive comment above.

scottmries (dismissed):
Built the branch and ran the check unit suite (26/26) and the generated integration suite (`#pass10` green), plus `prettier --check` and `eslint` — all clean. I also reverted `label-content-name-mismatch-evaluate.js` to `3a012a14~1` to confirm which of the new tests actually guard the #5096 fix; the inline comments reference that.

Two things to address, both small additions rather than rework:

1. Shadow DOM coverage was dropped rather than corrected — the code path this PR exists to protect (`visibleVirtual` recursing the flattened tree) is now untested.
2. #5096 flipped a sibling of the reported pattern from pass to violation, and this PR is what closes #4678 without pinning or noting it.

Plus two suggestions on test naming and fixture readability. Details inline.

test/checks/label/label-content-name-mismatch.js
chutchins25: Restored — added a cross-boundary case using your pattern: the hidden span sits in an open shadow root on a child of the labelled element, so it exercises `visibleVirtual`'s flattened-tree recursion. Returns `true` here and would return `false` on pre-#5096 (subtree) code.

test/checks/label/label-content-name-mismatch.js
chutchins25: Rewrote it to the mid-hidden case (`aria-label="deque labs are great"` with `labs` visually hidden): visible-only returns `false`, subtree would return `true`, so it now distinguishes the two implementations rather than passing under both.

test/integration/rules/label-content-name-mismatch/label-content-name-mismatch.html
chutchins25: Pinned as `#fail8` (the `Deque Labs home` shape). Went with `#fail` since it's the current behavior and a genuine 2.5.3 concern — a speech user seeing "deque are great" can't match the announced "deque labs are great". Happy to switch it to `#incomplete` if we'd rather treat it as undecided.

test/integration/rules/label-content-name-mismatch/label-content-name-mismatch.html
chutchins25: Moved to a `.visually-hidden` class in a `<style>` block; `#pass10` and `#fail8` both use it now, so the fixtures read as hidden-text-between-two-visible-words without the inline formatting noise.

scottmries (dismissed):
All items from my last round are addressed, and I confirmed the tests do what they claim rather than taking it on trust. Reverted `label-content-name-mismatch-evaluate.js` to `3a012a14~1` (pre-#5096), rebuilt, and re-ran: all three new unit tests fail there and pass here — so each is a real guard, including the rewritten mid-hidden case, which previously passed under both implementations. The shadow DOM case crosses a boundary now (`queryShadowFixture` attaches the root to the `<span id="shadow">` child; `#target` stays in light DOM), so `visibleVirtual`'s flattened-tree recursion is what's exercised. Check unit and generated integration suites both green.

On your open question: `#fail8` as a **fail** is right — don't switch it to incomplete. Visible label `deque are great` isn't a contiguous run within `deque labs are great`, which is a genuine 2.5.3 concern and what ACT 2ee8b8 expects. It also stays a fail under #5302's word-level matching, so the two PRs agree.

Since these are in flight together: #5291 and #5302 both touch all three of these files and conflict textually (the `const` block and the tail of the check test file). I merged them locally — the conflicts are adjacent additions, and `#pass10`, `#fail8` and `#incomplete18` all stay green under the combined code, along with the full 30-test check suite. Rebase, not a rethink.

One suggestion: the description is now behind the branch. It doesn't mention `#fail8` at all — which is the most consequential thing here, since it pins a pass→violation flip from #5096 that nothing else records — still describes the third unit test as "the mismatch counterpart", omits the `.visually-hidden` refactor, and cites axe 4.12.1 when develop is 4.13.0. Worth refreshing, since this PR is the durable record of what closing #4678 covered.

Garbee (dismissed):
The tests here are good and genuinely discriminating — each of the three new unit cases flips on pre-#5096 code, and the integration pair is correctly classified. Four things to resolve before merge: one about what this PR closes, three about what the tests actually pin.

### Don't close #4678 yet

Issue #4678 reports two defects. The first — hidden text wrongly included — is what this PR covers. The second is the reporter's `<a href="#">deque <span class="visually-hidden">labs</span> are great</a>` with no `aria-label`, which still isn't reported: `lib/rules/label-content-name-mismatch-matches.js:33-38` returns `false` when both `aria-label` and `aria-labelledby` are empty, so the node never reaches the check.

The body defers that to #5203, but #5203 is scoped to ACT rule 2ee8b8 "failed example 4" and general ACT drift — it never names the `aria-label`/`aria-labelledby` applicability gate. So as written, merging auto-closes a report whose second half has no home.

Either change the footer to `Refs #4678` and leave the issue open, or add a comment to #5203 quoting the reporter's second repro verbatim and naming the matches-level gate as the cause — then `Closes` is honest. Second option is cleaner if @straker agrees #5203 is the right home; worth a quick check with them since it's their rule.

### The shadow DOM test doesn't pin shadow traversal

`test/checks/label/label-content-name-mismatch.js:243` passes under two mutually exclusive conditions: `visibleVirtual` descends into the shadow root and the span is excluded (intended), or it never visits the shadow subtree at all (broken). Either way `visibleText` is `"deque are great"` and the check returns `true`. Break flattened-tree recursion in `lib/commons/text/visible-virtual.js` and this test stays green, and there's no shadow case in `test/commons/text/visible-virtual.js` to backstop it.

It does still guard the original #5096 regression, so it isn't dead weight — but it doesn't deliver the cross-boundary guard @scottmries asked for. The fix is one negative counterpart: same markup, but a **visible** `<span>labs</span>` in the shadow root against `aria-label="deque are great"`. Traversal working gives `visibleText` = `"deque labs are great"`, which isn't contained in the name → `isFalse`. Traversal broken gives `true` and the test fails loudly. Adding that turns the existing `isTrue` case into a real guard, since the pair can only both pass if the boundary is genuinely crossed.

### `#fail8` pins behavior without recording whether it's intended

`isStringContained` does a contiguous `includes()`, so hidden text interrupting the visible label mid-string can never match. Whether SC 2.5.3 intends a fail here, or this is a known contiguous-matching limitation, isn't recorded anywhere retrievable — and you offered upthread to switch it to `#incomplete`, which tells me it's genuinely undecided.

Whichever way it lands, the decision needs to survive outside this diff, because #5203's ACT-alignment work will hit this exact case and see a bare `#fail8` under a plain `<!-- Fail -->` heading. Get a ruling from @straker / @scottmries, then encode it: if it's intended, rename the unit test to name the reason (something like `returns false when the accessible name includes text present only in a visually-hidden child`) and add a line to the PR body cross-referencing #5203; if it's a limitation, `#incomplete` plus the same naming. Please don't settle this with a code comment — repo convention is to keep it in the name and the changelog/PR record.

### The fixtures don't exercise `clip` — they're hidden by overflow

Both the unit constant and the `.visually-hidden` class are excluded by `overflowHidden`, not `clipHidden`. `hiddenMethods` in `lib/commons/dom/is-visible-on-screen.js:13-19` runs `overflowHidden` before `clipHidden` and short-circuits; `getOverflowHiddenAncestors` includes the node itself, the `position: absolute` escape hatch needs the overflow ancestor to be `static` (here the ancestor *is* the absolute span), and `nodeRect.width < 2` fires on the 1px width. `clipHidden` is never reached.

I confirmed this in Chrome rather than by reading: running the PR's exact fixture with and without the `clip` declaration produces an identical result, and a fixture with only `position:absolute;width:1px;height:1px;overflow:hidden` and no `clip` at all is still excluded. So the `clip` value in both files is inert, and the clip branch of the visibility stack is unpinned for this rule despite the fixtures being named for it.

Not a correctness bug — the coverage claim just reads broader than it is. Cheapest fix is one more variant that isolates the branch: `position:absolute; clip:rect(0px,0px,0px,0px)` with normal dimensions and no `overflow:hidden`. I verified that one is excluded on its own, so it does pin `clipHidden`. Worth considering a `clip-path: inset(50%)` sibling in the same pass, since that's what Bootstrap 5 and Tailwind `sr-only` ship today and it's a separate branch in the same function — but that one's optional.

chutchins25:
Thanks @Garbee — all four addressed, one pending a ruling.

**1. Don't close #4678 yet** — agreed. Changed the footer to `Refs #4678`; this PR only covers the first defect (the FP). The second (name-from-content, no `aria-label`/`aria-labelledby`, gated out at `label-content-name-mismatch-matches.js:33-38`) is quoted verbatim and named on #5203 ([comment](https://github.com/dequelabs/axe-core/issues/5203#issuecomment-5208704617)). @straker — flagging to confirm #5203 is the right home for that half, or whether it wants its own tracking.

**2. Shadow test doesn't pin traversal** — added the negative counterpart you described (commit 55794f07): a *visible* `<span>labs</span>` in the shadow root against `aria-label="deque are great"` → `isFalse`. Paired with the existing `isTrue` case, the two can only both pass when the boundary is genuinely crossed.

**4. Fixtures don't exercise `clip`** — added two variants that isolate `clipHidden`: `position:absolute; clip:rect(0,0,0,0)` (normal dimensions, no `overflow`) and a `clip-path: inset(50%)` sibling for the modern `sr-only`. Both are excluded via the clip branch rather than `overflowHidden`.

**3. `#fail8` records no decision** — this is the open one. The mid-hidden case (`<a aria-label="deque labs are great">deque <span class="visually-hidden">labs</span> are great</a>`) currently fails: the visible "deque are great" isn't a contiguous part of the accessible name "deque labs are great". @straker / @scottmries — is that a genuine SC 2.5.3 failure (a speech user seeing "deque are great" can't match the announced "deque labs are great"), or a contiguous-matching limitation that should be `incomplete`? Whichever way it lands I'll rename the test to name the reason and record the decision in the PR body; leaving `#fail8` as-is until you weigh in.

straker (approved):

Issue #5207, opened 2026-07-14
ACT rule "Visible label is part of accessible name" (2ee8b8) changes break nightly


straker:
Latest ACT update for our rule 2ee8b8 "Visible label is part of accessible name" has [added more examples](https://github.com/w3c/wcag-act-rules/blob/main/content/rules/2ee8b8/proposed.md) which are [breaking our nightly tests run](https://github.com/dequelabs/axe-core/actions/runs/29303327772/job/86991534641). The changes are non-trival as they follow a [new algorithm](https://github.com/w3c/wcag-act-rules/blob/main/content/rules/2ee8b8/proposed.md#label-in-name-algorithm-label-in-name-algorithm) which we don't implement. So things like removing all content between round brackets and maintaining space between spans are currently failing. [Our code currently uses visibleVirtual](https://github.com/dequelabs/axe-core/blob/develop/lib/checks/label/label-content-name-mismatch-evaluate.js#L45-L49) to generate the visible content, which we'll need to figure out if we want to continue to use that function or create a new function just for this algorithm.

This issue is linked to a `skipTests` entry in `test/act-rules/visible-label-in-accessible-name-2ee8b8.spec.js`.

chutchins25:
Picking this up. Step 1 (parenthetical removal) is implemented in #5360 — `Passed Example 14` now passes unskipped, at 35 passing / 3 pending / 0 failing against `wcag-act-rules#main`.

The `visibleVirtual` question you raised is the one that needs your call, and it's sharper than it looks. `Passed Example 11` is this:

```html
<a aria-label="Download specification"
  ><span>Download</span><span id="space"> </span><span>specification</span></a
>
```

`visibleVirtual` recurses per child and ends each recursive call with `sanitize()`, which trims. So the whitespace-only `<span> </span>` returns `''`, the `.join('')` produces `downloadspecification`, and no amount of tokenising recovers the word boundary. It's `lib/commons/text/visible-virtual.js:63`, not the algorithm.

That leaves two options:

1. **Fix `visibleVirtual`.** Correct, but it feeds `color-contrast`, `target-size` via `create-grid`, modal detection and several matchers, so it changes what "visible text" means for every rule that reads it — the same blast radius that got #3883 de-scoped from this epic.
2. **A rule-local text function.** Contained, but it duplicates flattened-tree traversal and means `label-content-name-mismatch` stops agreeing with the rest of axe about what text is visible.

[...]

Issue #4821, opened 2025-07-02
Investigate label-content-name-mismatch when text contains short name forms of days


straker:
This was reported to us recently. In the following example, the visual text uses a short form for the day name (Wed) while the accessible name of the element uses the full name (Wednesday). This gives a violation for `label-content-name-mismatch`. We'll need to investigate if this is a violation of the WCAG SC, or if not how we would fix it so it doesn't report a problem.

```html
<button aria-label="Wednesday the 14th">Wed the 14th</button>
```

chutchins25:
Investigated this for #5226. Still reports a violation on 4.13.0:

```html
<button aria-label="Wednesday the 14th">Wed the 14th</button>
```

Visible label "Wed the 14th" vs accessible name "Wednesday the 14th" — "Wed" abbreviates "Wednesday", so the visible label isn't contained in the accessible name.

Findings:

1. **Abbreviation cases are an explicit assumption of ACT rule 2ee8b8** — it assumes the visible label doesn't use abbreviations/different spelling from the accessible name. So flagging this is consistent with the rule as written; the abbreviation scenario is a known gap in the underlying SC/ACT rule, not an axe-specific bug.
2. **It's arguably a genuine 2.5.3 failure.** If a speech-input user says the visible "Wed the 14th", it won't match a control whose accessible name is "Wednesday the 14th" — assistive tech keying off the accessible name needs the full form (per @WilcoFiers's Voice Control findings in #5228). Under that lens the violation is correct.
3. **Not reliably fixable in the rule.** Detecting that "Wed" abbreviates "Wednesday" (vs. an unrelated word) needs an abbreviation dictionary and would introduce false negatives — no safe heuristic.

Recommendation: treat as working-as-intended (a true label-in-name concern) rather than suppressing abbreviation mismatches. Open to going more lenient if preferred.

chutchins25:
cc @straker for the call on this one — my recommendation is working-as-intended (details above).

WilcoFiers:
Going to close this issue. Similar problems have been discussed by the W3C and they've concluded that abbreviations like this wouldn't meet the criterion.

PR #5302, opened 2026-08-07
fix(label-content-name-mismatch): compare visible label to name by word

### LINKED ISSUE #4311: label-content-name-mismatch should not treat hyphenated labels as matching non-hyphenated content

chutchins25:
## Summary

Fixes `label-content-name-mismatch` for hyphenated labels, per ACT rule 2ee8b8. Two parts:

1. **Word-level matching.** The visible label's words must appear as a contiguous run within the accessible name — tokenizing on non-text characters (replaced with a space, not deleted) rather than matching as a raw character substring. This correctly flags cases like `Discover It` vs `Discover Italy` that substring matching missed.
2. **Hyphenation is inapplicable.** Current ACT treats a hyphenation-only difference as inapplicable, not a failure. When the label is contained in the name once hyphens are removed, the check now returns `undefined` (needs review) instead of a violation — so `non-standard` vs `nonstandard` no longer reports a violation.

Supporting changes: a backward-compatible `replaceWith` option on `removeUnicode` (explicit unicode ranges — no `\p{…}` escape — to keep Safari 7+/IE11 support), and splitting label tokens on all whitespace.

## ACT 2ee8b8

The `act` nightly installs `wcag-act-rules#main` (38 2ee8b8 testcases), not the pinned dep (15). Verified against `main`: **33 passing / 5 pending / 0 failing**. `Inapplicable Example 6` (`non-standard`) now passes via the hyphenation handling.

Two categories remain skipped, tracked separately:

- **#5207** — whitespace between inline elements (`Passed Example 11`) and parenthetical-content removal (`Passed Example 14`). Its skip entries are left untouched here.
- **#4821** — abbreviation differences (`University Ave.` / `University Avenue`), which ACT also treats as inapplicable but cannot be reliably auto-detected. This PR's word-level matching surfaces one such case, so it is skipped.

## Related

This advances **#5203** (consistency with ACT 2ee8b8) but does not close it — the `wcag-act-rules` dependency bump and the remaining skipped cases (#5207, #4821) are tracked there and in their own issues.

Closes #4311


lib/checks/label/label-content-name-mismatch-evaluate.js
```diff
@@ -7,34 +7,58 @@ import {
...
+function curateTokens(str) {
+  const separated = removeUnicode(str, {
     emoji: true,
     nonBmp: true,
-    punctuations: true
+    punctuations: true,
+    replaceWith: ' '
   });
-  return sanitize(noUnicodeStr);
+  return sanitize(separated).split(' ').filter(Boolean);
```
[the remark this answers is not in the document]
chutchins25: Valid. `sanitize()` only collapses runs of two or more whitespace characters, so a lone `\n`/`\t` survived and `.split(' ')` would keep two words in one token. Fixed in ca8cde3c by splitting on `/\s+/`, with a regression test for a lone newline in the visible text.

test/act-rules/visible-label-in-accessible-name-2ee8b8.spec.js
```diff
@@ -2,13 +2,5 @@ require('./act-runner.js')({
...
-  skipTests: [
-    // See: https://github.com/dequelabs/axe-core/issues/4311
-    'e9bbdbec137223e2973c6d2896050770c84c26e5',
-    // See: https://github.com/dequelabs/axe-core/issues/5207
-    'fab659b02c1edb4f2c8f0bda524b1076abab7df6',
-    '94a7ce7aea9dbfaa375c459c26d3a5923de84e7a',
-    'e117393d6711d6bdf32821005219c9d9474dfeb8',
-    'f5c9811c984987443476760a1c5b91b1067f7e19'
-  ]
+  skipTests: []
```
scottmries: **Critical** — the four `#5207` entries aren't stale, and removing them turns the nightly red.

They're absent from the pinned dep, but all four exist in `wcag-act-rules#main`, which is what [`nightly-tests.yml:64`](https://github.com/dequelabs/axe-core/blob/ca8cde3c6c6ce12a60b70a7427ceefbba343efd0/.github/workflows/nightly-tests.yml#L64) installs (`main` has 38 testcases for 2ee8b8 vs 15 pinned):

| skip ID | title in `main` | under this PR |
| --- | --- | --- |
| `fab659b0` | Passed Example 11 | still fails — `<span>` whitespace |
| `94a7ce7a` | Passed Example 14 | still fails — needs parentheses removal |
| `e117393d` | Failed Example 3 | **now passes** ✅ |
| `f5c9811c` | Failed Example 15 | **now passes** ✅ |

Passed 11 and Passed 14 are the two parts of #5207 this PR doesn't implement, and the PR body agrees #5207 stays open — so its skips are still load-bearing.

Suggest keeping all four and dropping only the `#4311` entry, subject to my comment on `#fail8`.

test/integration/rules/label-content-name-mismatch/label-content-name-mismatch.html
```diff
@@ -46,6 +46,7 @@
   </svg>
   Hello Deque Systems
 </a>
+<a id="fail8" href="#" aria-label="non-standard">nonstandard</a>
```
scottmries: **Critical** — ACT reversed this expectation, so this fixture pins a decision upstream has undone.

Testcase `e9bbdbec1372…` is byte-identical between the pinned dep and `main` except for its `<title>`:

- pinned: `<title>Failed Example 4</title>` → expected `failed`
- `main`: `<title>Inapplicable Example 6</title>` → expected `inapplicable`

`main` added two clauses to the rule's **Applicability**:

> - Neither the element's accessible name nor its visible label contains any abbreviations; and
> - Every word that appears in both the element's accessible name and its visible label uses the same spelling and hyphenation in both places.

So current ACT treats hyphenation and abbreviation differences as out of scope rather than failures. #4311 was filed against the old expectation, so closing it by making this markup fail implements the reverted decision — and it's why `Inapplicable Example 5` (`University Ave.` / `University Avenue`) and `Inapplicable Example 6` newly fail against `main`.

This also answers the open question to @straker: `"e-mail"`/`"email"` and `"setting"`/`"settings"` are exactly what those two exclusions carve out. ACT's answer is "inapplicable", not "fail" — so the tokenization change needs the applicability narrowing alongside it (in `label-content-name-mismatch-matches`, or by returning `undefined` when the only difference is hyphenation/abbreviation). That clears both `Inapplicable` failures and resolves #4311 accurately.

If you'd rather ship the tokenization on its own first, dropping `#fail8` and keeping the `#4311` skip would make this landable without asserting the reverted expectation.

lib/checks/label/label-content-name-mismatch-evaluate.js
```diff
@@ -7,34 +7,58 @@ import {
 } from '../../commons/text';
 
 /**
- * Check if a given text exists in another
+ * Check whether the words of `compare` appear as a contiguous run of words
+ * within `compareWith`, following ACT rule 2ee8b8's "label in name" algorithm:
+ * non-letter/non-digit characters are treated as word separators and the
+ * comparison is done on whole words, not raw substrings.
```
scottmries: **Important** — this and the `curateTokens` block both say the code follows ACT's label-in-name algorithm, but two of its five tokenize steps aren't implemented:

1. *"Remove parentheses … and all characters that are between a left and right parenthesis"* — not done. This is what Passed Example 14 (`Search by date (YYYY-MM-DD)`) needs, and what #5207 names first.
2. *"Do Unicode case folding on the string then convert it to normalization form KD"* — only `.toLowerCase()` upstream, no NFKD.

Step 5's *"an empty list is a contiguous subsequence of any list"* also differs, since `isStringContained` returns `false` for empty token lists — though I confirmed that's unreachable, because `isHumanInterpretable` already gates those inputs to `undefined`.

Suggest scoping the comments to what's implemented (tokenize on non-text characters, then contiguous word-run match) and noting parentheses/NFKD as not yet handled, with a reference to #5207.

lib/checks/label/label-content-name-mismatch-evaluate.js
```diff
@@ -7,34 +7,58 @@ import {
...
-  const curatedCompareWith = curateString(compareWith);
-  const curatedCompare = curateString(compare);
-  if (!curatedCompareWith || !curatedCompare) {
+  const compareTokens = curateTokens(compare);
+  const compareWithTokens = curateTokens(compareWith);
+  if (!compareTokens.length || !compareWithTokens.length) {
     return false;
   }
-  return curatedCompareWith.includes(curatedCompare);
+  return isContiguousSubsequence(compareWithTokens, compareTokens);
```
scottmries: **Suggestion** — `isStringContained` no longer compares strings; it tokenizes and matches word runs, so the name now misdescribes it. Its params are also `(compare, compareWith)` while the call flips to `isContiguousSubsequence(compareWithTokens, compareTokens)`, which is easy to misread later.

Renaming to something like `isLabelContainedInName(label, name)` and aligning the helper's parameter order — or naming the helper's params `(haystack, needle)` — would make the flip obvious.

test/commons/text/unicode.js
```diff
@@ -270,4 +270,12 @@ describe('text.removeUnicode', () => {
     });
     assert.equal(actual, 'Hello World');
   });
+
+  it('substitutes matched characters with replaceWith when provided', () => {
```
scottmries: **Suggestion** — `replaceWith` now threads through five `replace` calls (`emojiRegexText`, `getUnicodeNonBmpRegExp`, `getSupplementaryPrivateUseRegExp`, `getCategoryFormatRegExp`, `getPunctuationRegExp`) but only the punctuation branch is covered.

The emoji and `nonBmp` paths matter for the new tokenization: an emoji mid-label now creates a word boundary where it previously didn't. Worth a `replaceWith: ' '` case for each.

scottmries (dismissed):
The word-level contiguous-subsequence change is the right algorithm — it matches the spec's own wording, and against latest ACT it fixes two testcases the PR body doesn't claim (Failed Example 3 `Discover It`/`Discover Italy`, Failed Example 15 `1`/`1a`). `tsc`, `prettier`, `eslint`, the check unit suite, the `removeUnicode` tests and the rule's integration suite are all green locally, and ACT 2ee8b8 against the **pinned** dep is 15/15 as claimed.

The problem is what happens against `wcag-act-rules#main`, which is what the nightly `act` job installs. I ran the 2ee8b8 suite against `main` twice — once with develop's lib code, once with this branch's — both with the empty skip list:

| Testcase | develop | this PR |
| --- | --- | --- |
| Passed 11 — `<span>` whitespace | fail | fail |
| Passed 14 — `Search by date (YYYY-MM-DD)` | fail | fail |
| Failed 3 — `Discover It` / `Discover Italy` | fail | **pass** ✅ |
| Failed 15 — `1` / `1a` | fail | **pass** ✅ |
| Inapplicable 5 — `University Ave.` / `University Avenue` | pass | **fail** ⛔ |
| Inapplicable 6 — `nonstandard` / `non-standard` | pass | **fail** ⛔ |

Both end at 34 passing / 4 failing. Today the nightly is green only because all four failures sit behind skips, so emptying the list turns it red. And the two new failures are there because ACT has since moved hyphenation and abbreviation differences *out of the rule's scope* — the opposite of what this PR implements. Details in the two inline comments on those lines; three smaller notes below them.

test/act-rules/visible-label-in-accessible-name-2ee8b8.spec.js
chutchins25: Confirmed against `main` myself — you're right, they're load-bearing (I'd only checked the pinned dep). Restored #5207's four skips untouched, removed only the #4311 entry (`e9bbdbec`, now passing), and added the abbreviation case (`University Ave.`, Inapplicable Example 5) under #4821. 2ee8b8 vs `main` is now 33 passing / 5 pending / 0 failing.

test/integration/rules/label-content-name-mismatch/label-content-name-mismatch.html
chutchins25: Confirmed — ACT reversed this to Inapplicable Example 6. Implemented the applicability narrowing you pointed to: a hyphenation-only difference now returns `undefined`, so `non-standard`/`nonstandard` no longer reports a violation and Inapplicable Example 6 passes against `main`. Dropped `#fail8`; #4311 now closes correctly as inapplicable. The abbreviation case (`University Ave.`) isn't reliably auto-detectable, so it stays skipped under #4821.

lib/checks/label/label-content-name-mismatch-evaluate.js
chutchins25: Scoped the comments to what's implemented (tokenize on non-text characters, then contiguous word-run match) and noted parenthetical-content removal and NFKD as not yet handled, with a #5207 reference.

lib/checks/label/label-content-name-mismatch-evaluate.js
chutchins25: Renamed to `isLabelContainedInName(label, name)` and made the helper's params `(haystack, needle)` so the flipped order reads clearly.

test/commons/text/unicode.js
chutchins25: Added `replaceWith: ' '` cases for emoji and non-BMP input.

chutchins25:
@scottmries — thanks for the thorough review, especially catching the `main`-vs-pinned framing. You were right on both counts: the four `#5207` entries are load-bearing against `wcag-act-rules#main` (I'd only checked the pinned dep), and ACT reversed hyphenation/abbreviation to *inapplicable*.

Reworked accordingly (per-point details in the inline replies):
- Implemented the applicability narrowing you pointed to — a hyphenation-only difference now returns `undefined`, so `non-standard`/`nonstandard` no longer fails and `Inapplicable Example 6` passes.
- Restored `#5207`'s four skips untouched; removed only the `#4311` entry (now passing); added the `University Ave.` abbreviation case under `#4821`.

Verified against `main`: 2ee8b8 is now **33 passing / 5 pending / 0 failing**. Ready for another look when you have a moment.

lib/checks/label/label-content-name-mismatch-evaluate.js
```diff
@@ -7,34 +7,72 @@ import {
...
+  const separated = removeUnicode(str, {
     emoji: true,
     nonBmp: true,
-    punctuations: true
+    punctuations: true,
+    replaceWith: ' '
   });
-  return sanitize(noUnicodeStr);
+  return sanitize(separated).split(/\s+/).filter(Boolean);
+}
```
scottmries: **Critical** — zero-width characters can't be word boundaries, so strip them before tokenizing. The `nonBmp` pass then finds nothing left to replace, and the soft-hyphen case resolves to pass without `removeHyphens` needing to cover it.

```js
import { getCategoryFormatRegExp } from '../../commons/text/unicode';

function curateTokens(str) {
  // Zero-width format characters are invisible, so they can't be word boundaries.
  const separated = removeUnicode(str.replace(getCategoryFormatRegExp(), ''), {
    emoji: true,
    nonBmp: true,
    punctuations: true,
    replaceWith: ' '
  });
  return sanitize(separated).split(/\s+/).filter(Boolean);
}
```

lib/checks/label/label-content-name-mismatch-evaluate.js
```diff
@@ -7,34 +7,72 @@ import {
...
+ * Remove hyphens so a hyphenated word collapses into a single word (e.g.
+ * "non-standard" becomes "nonstandard"). Used to detect when the only
+ * difference between the label and the name is hyphenation.
+ *
+ * @param {String} str
+ * @returns {String}
+ */
+function removeHyphens(str) {
+  return str.replace(/[-‐‑]/g, '');
 }
```
scottmries: **Suggestion** — this covers U+002D/2010/2011, but `getPunctuationRegExp` also splits on the rest of the dash family, which `removeHyphens` then won't rejoin. With `aria-label="email"`, on develop vs here: `e-mail`/`e‐mail`/`e‑mail` go pass → incomplete, but `e‒mail` (U+2012), `e–mail` (U+2013) and `e—mail` (U+2014) go pass → violation. Rare in practice, but the three-way split is hard to reason about later. `/[\u002D\u2010-\u2015\u2212]/g` would land them all on the same outcome.

lib/checks/label/label-content-name-mismatch-evaluate.js
```diff
@@ -59,7 +97,21 @@ function labelContentNameMismatchEvaluate(node, options, virtualNode) {
...
+  }
+
+  // ACT rule 2ee8b8 treats hyphenation differences as inapplicable. When the
+  // label is contained in the name once hyphens are removed rather than treated
+  // as word separators, the only difference is hyphenation, so return undefined
+  // (needs review) instead of a violation.
+  if (
+    isLabelContainedInName(removeHyphens(visibleText), removeHyphens(accText))
+  ) {
+    return undefined;
```
scottmries, line 114: **Suggestion** — the check has no `incomplete` message, so this surfaces as "axe couldn't tell the reason. Time to break out the element inspector!" — same text as the `isHumanInterpretable` path, though here axe knows exactly why. A `this.data()` call plus a templated `incomplete` message would say "differ only in hyphenation".

lib/checks/label/label-content-name-mismatch-evaluate.js
```diff
@@ -59,7 +97,21 @@ function labelContentNameMismatchEvaluate(node, options, virtualNode) {
...
 
-  return isStringContained(visibleText, accText);
+  if (isLabelContainedInName(visibleText, accText)) {
+    return true;
+  }
+
+  // ACT rule 2ee8b8 treats hyphenation differences as inapplicable. When the
+  // label is contained in the name once hyphens are removed rather than treated
+  // as word separators, the only difference is hyphenation, so return undefined
+  // (needs review) instead of a violation.
```
scottmries: **Suggestion** — worth a line on #5203 that the dep bump should revisit this branch. `undefined` is load-bearing only while the pinned dep is stale; once it tracks `main`, whether these should be incomplete forever or stop matching the rule is a live question.

scottmries (dismissed):
Re-reviewed against `main` — the three items from last round are resolved, and I reproduce your numbers. Locally on node 24: `tsc`, `prettier`, `eslint` clean; the checks, commons and integration unit suites green; 2ee8b8 is 15/15 against the pinned dep and **33 passing / 5 pending / 0 failing** against `wcag-act-rules#main`.

Worth recording somewhere: `undefined` is the only value that satisfies both dependencies at once — pinned `e9bbdbec` is `Failed Example 4` (expects `failed`, and act-runner accepts incomplete for that), `main` is `Inapplicable Example 6` (expects zero violations). `true` would break the pinned run.

One blocker. `replaceWith: ' '` also lands on `getCategoryFormatRegExp()` — Unicode category Cf, which is zero-width. Those characters can't create a visible word boundary, but they now split one visible word into two tokens. I built both `71c46831` and this branch and ran the same fixtures:

| visible text | accessible name | develop | this PR |
| --- | --- | --- | --- |
| `non&shy;standard` (U+00AD) | `nonstandard` | pass | **violation** ⛔ |
| `time\u200Bline` (ZWSP) | `timeline` | pass | **violation** ⛔ |
| `mi\u200Cshavad` (ZWNJ) | `mishavad` | pass | **violation** ⛔ |
| `sea\u200Erch` (LRM) | `search` | pass | **violation** ⛔ |
| `sub\uFEFFmit` (BOM) | `submit` | pass | **violation** ⛔ |

ZWNJ is grammatically required in Persian and Indic scripts, `&shy;` is the standard long-word hyphenation hint, and bidi marks are pervasive in RTL content — so these are new false positives on a `serious` rule. Fix inline; three smaller notes below it.

lib/checks/label/label-content-name-mismatch-evaluate.js
chutchins25: Done — `curateTokens` now strips `getCategoryFormatRegExp()` characters before the `replaceWith: ' '` pass, per your snippet, so zero-width/soft-hyphen characters no longer create a word boundary. Added a regression test (`non­standard` vs `nonstandard` → pass).

lib/checks/label/label-content-name-mismatch-evaluate.js
chutchins25: Widened `removeHyphens` to `/[-‐-―−]/g`, so the whole dash family lands on the same outcome as the tokenizer rather than the three-way split.

lib/checks/label/label-content-name-mismatch-evaluate.js
chutchins25: Added a templated `incomplete` message: `this.data({ messageKey: 'hyphenation' })` on this branch plus an `incomplete` object in the check JSON (`default` for the `isHumanInterpretable` path, `hyphenation` here). The unit tests assert the messageKey, and `doc/rule-descriptions.md` now reflects the "needs review" outcome.

lib/checks/label/label-content-name-mismatch-evaluate.js
chutchins25: Added a `TODO(#5203)` at the hyphenation branch noting the `undefined` result is load-bearing only while the pinned dep is stale, and that the dep bump should revisit whether these stay incomplete or stop matching. The PR body's Related section also references #5203.

lib/checks/label/label-content-name-mismatch-evaluate.js
```diff
@@ -5,36 +5,80 @@ import {
...
+ * "non-standard" becomes "nonstandard"). Used to detect when the only
+ * difference between the label and the name is hyphenation.
+ *
+ * @param {String} str
+ * @returns {String}
+ */
+function removeHyphens(str) {
+  // The whole dash family, so it stays consistent with the dashes that
+  // `getPunctuationRegExp` treats as word separators during tokenizing.
+  return str.replace(/[\u002D\u2010-\u2015\u2212]/g, '');
```
scottmries: **Suggestion** — the widened class is untested: every hyphenation test uses ASCII `-`, so nothing pins `\u2010-\u2015\u2212`. I confirmed all six dash variants return `undefined`, but a future narrowing would pass CI silently. One case with `aria-label="email"` / `e–mail` would cover it.

Also, U+2212 isn't matched by `getPunctuationRegExp` — it's in `\u2200-\u22FF`, i.e. `getUnicodeNonBmpRegExp`. Both feed the tokenizer so the reasoning holds; "the dashes the tokenizer treats as word separators" would be accurate.

scottmries (dismissed):
`0c949242` addresses all four items. I re-ran the fixtures rather than reading the diff:

- **Zero-width (the blocker):** soft hyphen, ZWSP, ZWNJ, LRM and BOM all return `true` again, matching develop.
- **Dash family:** all six visible dash variants now return `undefined` consistently, and soft hyphen returns `true`. The three-way split is gone.

`tsc`, `prettier`, `eslint` clean; checks, commons and integration suites exit 0; `test-locales` 20/20 with the new key; 2ee8b8 15/15 against the pinned dep. The `messageKey` plumbing is right — `process-message.js` reads `data.messageKey`, `MockCheckContext` is wired with an `afterEach` reset, and `_template.json` is regenerated and committed.

Two things worth a line in the PR body, since both are user-visible and neither is mentioned: the `doc/rule-descriptions.md` flip to "failure, needs review" is a **fix** (the rule already returned `undefined` on the `isHumanInterpretable` path, so the table was wrong), and the new `default` message replaces the generic "axe couldn't tell the reason" fallback for the 17 pre-existing incomplete cases.

Since #5291 is in flight against the same three files: the two conflict textually. I merged them locally — conflicts are adjacent additions, and `#pass10`, `#fail8` and `#incomplete18` all stay green under the combined code. Rebase, not a rethink.

One suggestion inline.

lib/checks/label/label-content-name-mismatch.json
```diff
@@ -9,7 +9,11 @@
     "impact": "serious",
     "messages": {
       "pass": "Element contains visible text as part of it's accessible name",
-      "fail": "Text inside the element is not included in the accessible name"
+      "fail": "Text inside the element is not included in the accessible name",
+      "incomplete": {
+        "default": "Unable to determine if the visible text is part of the accessible name",
+        "hyphenation": "The visible text and the accessible name differ only in hyphenation"
```
Garbee: **Blocker:** this new `messageKey` will throw a `TypeError` in every localized bundle.

`build/configure.mjs:103` replaces `check.metadata.messages` wholesale with the locale entry when building a localized bundle:

```js
check.metadata.messages = locale.checks[check.id];
```

All 18 non-English `locales/*.json` have only `pass`/`fail` for `label-content-name-mismatch` — no `incomplete`. `publish-metadata.js:62` then evaluates:

```js
} else if (checkData && checkData.messageKey) {
  return messages.incomplete[checkData.messageKey];   // undefined['hyphenation']
}
```

This branch is **newly reachable** for this check: before this PR the check never called `this.data({ messageKey })`, so every incomplete result took the guarded `getDefaultMsg` path. Now any `axe.<lang>.js` bundle hard-throws on a page containing a hyphenation-only mismatch.

Note the runtime path (`axe.configure({ locale })`) is fine — `mergeCheckLocale` in `lib/core/base/audit.js` preserves `a.messages.incomplete` when it is an object. Only build-time localized bundles break.

**Suggested fix** — guard the access in `lib/core/utils/publish-metadata.js`, which protects every check rather than just this one:

```js
} else if (checkData && checkData.messageKey) {
  return messages.incomplete?.[checkData.messageKey] ?? getDefaultMsg(messages);
}
```

Please add a regression test that resolves this check's messages under a locale lacking `incomplete`.

lib/checks/label/label-content-name-mismatch-evaluate.js
```diff
@@ -5,36 +5,80 @@ import {
...
+  const separated = removeUnicode(str.replace(getCategoryFormatRegExp(), ''), {
     emoji: true,
     nonBmp: true,
-    punctuations: true
+    punctuations: true,
+    replaceWith: ' '
   });
-  return sanitize(noUnicodeStr);
+  return sanitize(separated).split(/\s+/).filter(Boolean);
+}
```
Garbee: **Blocker:** punctuation other than hyphens silently regresses to hard violations.

Punctuation used to be **deleted** before comparison (`curateString` passed `punctuations: true` with the old `''` replacement). It is now a **word separator**, and only hyphens get the `removeHyphens` escape hatch. Running the new tokenizer:

| visible text | accessible name | before | after |
|---|---|---|---|
| `it's` | `its book` | pass | **fail** |
| `don't save` | `dont save now` | pass | **fail** |
| `u.s.a` | `usa` | pass | **fail** |
| `e-mail` | `email` | fail | incomplete ✅ |

Apostrophes in contractions and possessives are everyday UI copy, so this is a meaningful new false-positive class on a WCAG 2.5.3 rule — and nothing in the diff tests or documents it either way.

**Suggested fix** — pick one and make it explicit:

1. Generalize the carve-out: instead of `removeHyphens`, strip *all* characters that `curateTokens` treats as separators before the second comparison, so any punctuation-only difference lands in `incomplete` rather than `false`. That also makes the `hyphenation` message key a more accurate `punctuation` key.
2. Or keep the hyphen-only behavior deliberately, and add tests asserting `isFalse` for an apostrophe case and a period case, so the asymmetry vs. hyphens is a recorded decision rather than a side effect of the tokenizer change.

Option 1 matches ACT 2ee8b8's treatment of punctuation differences more closely; either way the tests should pin it.

lib/commons/text/remove-unicode.js
```diff
@@ -17,22 +17,23 @@ import { emojiRegexText } from '../../core/imports';
...
-      .replace(getUnicodeNonBmpRegExp(), '')
-      .replace(getSupplementaryPrivateUseRegExp(), '')
-      .replace(getCategoryFormatRegExp(), '');
+      .replace(getUnicodeNonBmpRegExp(), replaceWith)
+      .replace(getSupplementaryPrivateUseRegExp(), replaceWith)
+      .replace(getCategoryFormatRegExp(), replaceWith);
   }
   if (punctuations) {
-    str = str.replace(getPunctuationRegExp(), '');
+    str = str.replace(getPunctuationRegExp(), replaceWith);
```
Garbee: `replaceWith` is documented as "string to substitute", but as a string second argument to `String.prototype.replace` it is a *replacement pattern* — `$&`, `` $` ``, `$'`, `$$` (and `$1`…`$9`) all have special meaning at these five call sites.

`removeUnicode` is on the public `axe.commons.text` surface, so this becomes an observable guarantee once released:

- `removeUnicode(str, { emoji: true, replaceWith: '$&' })` removes nothing at all.
- `removeUnicode(str, { emoji: true, replaceWith: "$'" })` duplicates the string tail on every match — quadratic blowup on emoji-dense text.
- `$1` is inert today only because none of the three regexes (including vendored `emoji-regex@10.6.0`) has a capture group. A future emoji-regex release adding one changes the semantics of every consumer silently.

Separately, `replaceWith: null` yields the literal string `"null"` rather than the documented `''` default, because destructuring defaults only fire on `undefined`.

**Suggested fix** — the option is unreleased, so closing this is free now and breaking later. Use a replacer function so the value is always a literal and the capture-group coupling disappears:

```js
function removeUnicode(str, options) {
  const { emoji, nonBmp, punctuations, replaceWith = '' } = options;
  const replacer = () => replaceWith;

  if (emoji) {
    str = str.replace(emojiRegexText(), replacer);
  }
  // …same for the nonBmp and punctuations replacements
}
```

If the `$` semantics are actually intended, please say so explicitly in the JSDoc instead.

lib/checks/label/label-content-name-mismatch-evaluate.js
Garbee: Two things on this line and its comment.

**1. The comment misattributes U+2212.** It says the set "stays consistent with the dashes that `getPunctuationRegExp` treats as word separators", but U+2212 MINUS SIGN is not in `getPunctuationRegExp` — it is matched by `getUnicodeNonBmpRegExp` (`∀-⋿`, Mathematical Operators). The behavior is right, because `curateTokens` passes both `nonBmp: true` and `punctuations: true`; only the explanation sends a reader to the wrong function. Suggested rewording:

```js
// The whole dash family: U+002D and U+2010–U+2015 come from
// getPunctuationRegExp, U+2212 from getUnicodeNonBmpRegExp. Together
// they cover every dash curateTokens treats as a word separator.
```

**2. The range is untested beyond ASCII `-`.** Both new hyphenation tests use a plain `-`, so narrowing this regex to `/-/g` leaves the entire suite green. Please add a case at each end of the intended set — U+2013 EN DASH and U+2212 MINUS SIGN — asserting `isUndefined` plus `messageKey: 'hyphenation'`. En dashes in particular show up in real labels.

test/checks/label/label-content-name-mismatch.js
```diff
@@ -224,4 +237,56 @@ describe('label-content-name-mismatch tests', () => {
...
+    );
+    const actual = check.evaluate.call(
+      checkContext,
+      vNode.actualNode,
+      options,
+      vNode
+    );
+    assert.isUndefined(actual);
+    assert.deepEqual(checkContext._data, { messageKey: 'hyphenation' });
+  });
```
Garbee: This file has no open Shadow DOM case (`queryShadowFixture`). That gap predates the PR, but this change makes it worth closing here: the check now depends on how `visibleVirtual`'s output is *split into tokens*, and shadow boundaries are exactly where text gets concatenated without whitespace — e.g. a shadow-hosted button whose shadow text and slotted light-DOM text abut and collapse two words into one, or where a boundary introduces a split the old substring comparison was immune to.

Repo convention (CLAUDE.md §2) asks for an open Shadow DOM case for every relevant check. Suggested addition alongside these tests:

```js
it('matches across an open shadow DOM boundary', () => {
  const vNode = queryShadowFixture(
    '<div id="shadow"></div>',
    '<button id="target" aria-label="save changes">save <slot></slot></button>',
    'changes'
  );
  assert.isTrue(check.evaluate(vNode.actualNode, options, vNode));
});
```

(adjust to this repo's `queryShadowFixture` signature — the point is one case where the label text spans the boundary).

Garbee (dismissed):
The move from substring matching to word-token matching is the right direction for 2ee8b8, and the hyphenation incomplete branch is a nice touch. Two blockers before this can go in, plus three smaller items — all inline.

**Blockers**

1. **Localized bundles will throw on the new `hyphenation` message key.** `build/configure.mjs` replaces `check.metadata.messages` wholesale with the locale entry, and none of the 18 non-English locales define `incomplete` for this check, so `publish-metadata.js`'s unguarded `messages.incomplete[checkData.messageKey]` becomes `undefined['hyphenation']`. This branch is newly reachable — previously the check never set a `messageKey`, so incompletes always took the guarded default path. (The runtime `axe.configure({ locale })` path is unaffected; `mergeCheckLocale` preserves the base `incomplete` object.)

2. **Non-hyphen punctuation silently regresses to hard violations.** Punctuation used to be deleted before comparison and is now a word separator, but only hyphens get an escape hatch. `it's` vs `its book`, `don't save` vs `dont save now`, and `u.s.a` vs `usa` all flip from pass to fail. Apostrophes are everyday UI copy; this needs to be either generalized to all punctuation-only differences or deliberately pinned with tests.

**Also worth addressing**

3. `replaceWith` is a `String.prototype.replace` *pattern*, not a literal — `$&`, `` $` ``, `$'`, `$$` are live on a newly public `axe.commons.text` option. Cheap to close now, breaking to close after release.
4. `removeHyphens`'s dash range is untested beyond ASCII `-`; narrowing it to `/-/g` leaves the suite green.
5. The `removeHyphens` comment attributes U+2212 to `getPunctuationRegExp`, which doesn't match it (it comes from `getUnicodeNonBmpRegExp`).
6. No open Shadow DOM case — pre-existing, but this change makes tokenization sensitive to boundary text concatenation.

For transparency: I reviewed this statically and by executing the tokenizer and the locale/ACT lookups in isolation — I did not run the karma or ACT suites, so the test-related points above are reasoned from the code rather than observed failures.

lib/checks/label/label-content-name-mismatch.json
chutchins25: Good catch, and thanks for the depth here. One correction on the mechanism: `getIncompleteReason` is only reached when `typeof messages.incomplete === 'object'` (publish-metadata.js:87), so with `incomplete` absent entirely — the localized-bundle case — it isn't called and line 62 doesn't throw. The actually-reachable failure is a level up: line 94 sets `data.message = messages.incomplete`, i.e. `undefined`, leaving the incomplete result with no message. Fixed there (`?? incompleteFallbackMessage()`), and I also hardened line 62 as you suggested since it's real for a *partial* `incomplete` object (present but missing the key). Added the regression test resolving this check's messages under a locale lacking `incomplete`.

lib/checks/label/label-content-name-mismatch-evaluate.js
chutchins25: Went with your option 1 (generalize). The carve-out now removes *all* non-text characters before the second comparison rather than only hyphens, so any punctuation-only difference — `it's`/`its book`, `u.s.a`/`usa`, `e-mail`/`email` — lands in `incomplete` instead of `false`. `removeHyphens` is gone; messageKey is now `punctuation`. Added tests for apostrophe, periods and an en dash.

lib/commons/text/remove-unicode.js
chutchins25: Fixed — `removeUnicode` now uses a replacer function (`() => replaceWith`) at all five call sites, so `replaceWith` is always inserted literally and the `$&`/`$'`/`$1` coupling is gone. Added a test asserting `replaceWith: '$&'` inserts the literal string.

lib/checks/label/label-content-name-mismatch-evaluate.js
chutchins25: Both resolved by generalizing the carve-out (comment 2): `removeHyphens` and its dash regex are removed entirely, so the U+2212 misattribution and the untested-dash-range concern go away. The en dash / minus cases now flow through the general `joinNonText` path, and I added an en-dash test.

test/checks/label/label-content-name-mismatch.js
chutchins25: Added an open-shadow-DOM case where the visible label spans the boundary (shadow text `save` + light-DOM ` changes` vs `aria-label="save changes"` → `isTrue`), so token concatenation across the boundary is now covered.

chutchins25:
@Garbee — thorough review, thank you. Both blockers and the three smaller items are addressed in the latest push (per-point details in the inline replies).

You flagged that you ran the tokenizer and the locale/ACT lookups in isolation but held off on the full karma/ACT suites, so for completeness I ran those against the changes — check unit (incl. the new apostrophe/period/en-dash `incomplete` cases and the shadow-boundary case), `removeUnicode`, the `publish-metadata` regression test, the rule's integration suite, and ACT 2ee8b8 against `wcag-act-rules#main` (33 passing / 5 pending / 0 failing). All green.

One mechanism correction worth surfacing (also on the `publish-metadata` thread): the localized-bundle path doesn't actually throw — the `typeof messages.incomplete === 'object'` guard means `getIncompleteReason` isn't reached when `incomplete` is absent. The reachable failure was a *missing* message; fixed at that path, and I hardened your line-62 suggestion too since it's real for a partial `incomplete` object.

lib/core/utils/publish-metadata.js
```diff
@@ -59,7 +59,11 @@ function getIncompleteReason(checkData, messages) {
       }
     }
   } else if (checkData && checkData.messageKey) {
-    return messages.incomplete[checkData.messageKey];
+    // Localized bundles replace `messages` with a locale entry that may omit
+    // `incomplete`, so fall back to the default rather than throwing.
+    return (
+      messages.incomplete?.[checkData.messageKey] ?? getDefaultMsg(messages)
```
scottmries, line 65: **Important** — this is core message resolution for *every* check, in a PR scoped to one rule (plus the `removeUnicode` replacer change). CLAUDE.md §3 is "one change per PR."

The commit message justifies it as preventing a throw. That doesn't reproduce — I tested both routes against `0c949242`:

- `axe.configure({locale})` with a bundle lacking `incomplete` (the shape shipped in `de`/`fr`/`ja`) merges cleanly: no throw, correct message.
- `axe.configure({checks})` replacing the metadata wholesale: no throw either, just `message: undefined`.

So the real symptom is an undefined message, not a crash, and the new test reaches it via a synthetic `axe._load` rather than a public path. Worth fixing — in its own PR, with the justification corrected.

lib/checks/label/label-content-name-mismatch-evaluate.js
```diff
@@ -59,7 +96,25 @@ function labelContentNameMismatchEvaluate(node, options, virtualNode) {
...
+  // ACT rule 2ee8b8 treats punctuation-only differences (e.g. hyphenation) as
+  // inapplicable. When the label is contained in the name once non-text
+  // characters are removed rather than treated as word separators, the only
+  // difference is punctuation, so return undefined (needs review) instead of a
+  // violation.
+  //
+  // TODO(#5203): the incomplete result here is load-bearing only while the
+  // pinned wcag-act-rules dep is stale. When the dep bump tracks `main`,
+  // revisit whether these should stay incomplete or stop matching the rule.
+  if (isLabelContainedInName(visibleText, accText, { joinNonText: true })) {
```
scottmries, line 112: **Important** — ACT 2ee8b8's applicability carves out **abbreviations and hyphenation**, not punctuation generally. `joinNonText: true` strips all punctuation, emoji and symbols, so genuine violations become `undefined`. Verified `0c949242` → this commit:

| visible / name | before | after |
| --- | --- | --- |
| `it's` / `its book` | violation | **incomplete** |
| `us` / `u.s. army` | violation | **incomplete** |
| `1000` / `1,000 items` | violation | **incomplete** |

The abbreviation case is arguably right. `1,000 items` is a plain punctuation difference ACT doesn't exempt — a speech user saying "1000" wouldn't match "1,000 items", which is exactly what 2.5.3 is about.

ACT stays green either way, so conformance isn't what's driving this. Suggest narrowing back to the hyphenation-scoped carve-out (plus abbreviations if you want to pick up #4821), or making the case in the PR body for why punctuation-wide is the right reading.

scottmries (dismissed):
Reopening — my approval was for `0c949242` (+150/-29); this is +257/-35 and changes behavior I'd signed off on. ACT is still clean (15/15 pinned, 33 passing / 5 pending / 0 failing vs `main`), and `tsc`/`prettier`/`eslint` plus the checks, commons, core and integration suites all pass. Two concerns with the new commit, both inline.

To be clear about what I actually asked for last round: a test pinning the `\u2010-\u2015\u2212` range. This commit deleted the range instead and generalized the carve-out — that's a bigger change than the note warranted.

chutchins25:
Surfacing a scope question where two reviews point different directions, so we settle it before I recode. cc @scottmries @Garbee @straker.

**The carve-out currently generalizes** (this commit): a difference that is only non-text characters — hyphenation, apostrophes, periods, commas — returns `incomplete` instead of a violation.

**@Garbee** asked for exactly this: apostrophe/period cases like `it's`/`its book` and `u.s.a`/`usa` were flipping to hard violations, a false-positive class on everyday copy.

**@scottmries** (reopening) reads it as too broad: 2ee8b8 carves out **abbreviation + hyphenation**, not punctuation generally, so genuine violations now return `incomplete` — e.g. `1000` / `1,000 items` (a speech user saying "1000" wouldn't match "1,000 items", which is what 2.5.3 is about).

The crux is the apostrophe/comma cases:
- `non-standard` / `nonstandard` — hyphenation → `incomplete` (agreed; #4311).
- `u.s.a` / `usa`, `University Ave.` / `Avenue` — abbreviation → exempt under 2ee8b8, but not reliably auto-detectable (#4821).
- `it's` / `its` — a spelling difference (2ee8b8's clause is "same spelling and hyphenation"), so arguably **not** exempt.
- `1,000 items` / `1000` — a punctuation difference 2ee8b8 doesn't exempt.

So the deciding question: should the carve-out be **hyphenation-only** (+ abbreviation via #4821), matching 2ee8b8's exemptions, or **punctuation-wide** to suppress the apostrophe/period false-positives @Garbee flagged? It's your rule @straker — what's the intended scope? I'll implement whichever you land on.

Separately, @scottmries — agreed on the `publish-metadata` scope point; I'll pull that change into its own PR (justification corrected to "undefined message," not a throw) regardless of the scope outcome.

scottmries:
I defer to @straker on the design questions, LGTM otherwise.

chutchins25:
Closing this in favour of #5345.

@straker set out on #5203 how the rule should reach ACT consistency:

> Just to be clear, getting consistent means our act test passes on the new examples. This means implementing the new algorithm as defined in [the label in name algorithm](https://www.w3.org/WAI/standards-guidelines/act/rules/2ee8b8/proposed/#label-in-name-algorithm). Our current code just does a substring comparison of the accessible name and the visible text. We'll need to add the other steps to that part of the code.
>
> We are accepting that Inapplicable 5 & 6 will fail instead of incomplete and should be skipped. See act-rules/act-rules.github.io#2443

That settles the scope question I raised on 17 August, and it lands outside the range this PR was arguing over — the `incomplete` branch goes away entirely rather than being narrowed to hyphenation. Since the only surviving content is the tokenizer from the first commit, a fresh PR against #5203 was cleaner than reworking five commits of a rejected approach.

#5345 carries over the word-level contiguous-subsequence comparison, the `replaceWith` option on `removeUnicode`, the zero-width handling, and @scottmries' requested en-dash case — now asserting a violation rather than an incomplete.

It drops the punctuation `incomplete` branch with its check messages and locale entries, and the `publish-metadata.js` change, which was out of scope as @scottmries said. That defect is now tracked as #5344.

Against `wcag-act-rules#main` it runs 34 passing / 4 pending / 0 failing, up from 33/5 here.

Thanks @scottmries and @Garbee — the tokenizer survived because of these reviews.

lib/checks/label/label-content-name-mismatch-evaluate.js
chutchins25: You were right on both points, and the code they applied to is gone — `removeHyphens` and the widened dash class went with the whole carve-out.

The U+2212 correction stands on the record: it's matched by `getUnicodeNonBmpRegExp` via `\u2200-\u22FF`, not `getPunctuationRegExp`. Your suggested case survives in #5345 as `aria-label="email"` / `e–mail`, now asserting a violation.

lib/core/utils/publish-metadata.js
chutchins25: Agreed and reverted — it's out of #5345 entirely.

Your correction to the justification was right too: it's an undefined message, not a throw. I've filed it as #5344 with that framing, and the repro there goes through `axe.run` on a real fixture rather than a synthetic `axe._load` — `label-content-name-mismatch` has no `incomplete` message today, so `failureSummary` renders as `"Fix any of the following:\n  "` on `develop`. Seven checks are in that state.

lib/checks/label/label-content-name-mismatch-evaluate.js
chutchins25: Your read was the closer one, and @straker went further still — see the ruling on #5203. There's no carve-out at all in #5345: hyphenation and abbreviation differences report violations, and ACT `Inapplicable Example 5` and `6` are skipped as accepted divergence.

So all three rows in your table are violations again, including `1,000 items`.

PR #5303, opened 2026-08-10
fix(dq-element): remove memoize wrapper around constructor

### LINKED ISSUE #5298: Remove memoize of DqElement
The `memoize` around `DQElement` is actually hurting performance and making it take 20s more to process `new DqElement` when running the reporter. Removing the `memoize` wrapper greater reduces the time to instantiate it. 

However this should also consider the reason [why we memoized DqElement to being with](https://github.com/dequelabs/axe-core/pull/4452). We'll need to investigate if we can still resolve both perf concerns.

shoaibyazdani:
Closes #5298

### What
Removes the `memoize` wrapper around the DqElement constructor.

### Why
Closes #5298. The `memoize` wrapper was hurting performance and adding ~20s to reporter initialization. Each `new DqElement` call had to unwrap the memoize wrapper, which is more expensive than direct instantiation.

### Changes
- Removed `import memoize from './memoize';` from `lib/core/utils/dq-element.js`
- Changed `const DqElement = memoize(function DqElementMemoized(...))` to `const DqElement = function DqElement(...)`
- Renamed inner function name to match the variable
- 1 file changed, 2 insertions(+), 3 deletions(-)

Closes:

### DIFF
--- a/lib/core/utils/dq-element.js
+++ b/lib/core/utils/dq-element.js
@@ -4,7 +4,6 @@ import getXpath from './get-xpath';
 import getNodeFromTree from './get-node-from-tree';
 import AbstractVirtualNode from '../base/virtual-node/abstract-virtual-node';
 import cache from '../base/cache';
-import memoize from './memoize';
 import getSource from './get-element-source';
 
 const CACHE_KEY = 'DqElm.RunOptions';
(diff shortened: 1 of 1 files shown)


straker:
Thanks for the pr. This wasn't quite ready to pick up though as it needs to consider the reason why we [memoized in the first place](https://github.com/dequelabs/axe-core/pull/4452). Instead of removing it completely we'll need to investigate if we can still resolve both perf issues.

PR #5304, opened 2026-08-10
fix(audit): remove undocumented tagExclude option (closes #620)

### LINKED ISSUE #620: Remove the undocumented tagExclude option
There is currently no documentation in the API docs describing how it works or what the default setting is

shoaibyazdani:
## What
Removes the undocumented `tagExclude` option from the public API.

## Why
Closes #620. The `tagExclude` option was undocumented and provided no real value beyond filtering rules by internal tags. Removing it simplifies the public API and removes a maintenance burden.

## Changes
- Removed `this.tagExclude = ['experimental', 'deprecated']` default in audit.js
- Removed `spec.tagExclude` handling in configure.js
- Removed `tagExclude` filtering in get-rules.js
- Removed `defaultExclude` logic in rule-should-run.js
- Removed/updated tests that depended on tagExclude
- 9 files changed, 4 insertions(+), 304 deletions(-)

Closes: #620


straker:
Thanks for the pr. Unfortunately while the feature is undocumented it is a necessary part of the code. Without it rules marked as `experimental` and `deprecated` will be run by default which is not what we want to do. Before we can remove it we need a different way to prevent those rules from running by default.

Issue #5315, opened 2026-08-18
Customizable <select>'s button and selectedcontent are not treated as inert, causing false button-name violations


ItsMeSousuke:
### Product

axe-core

### Product Version

4.13.0

### Latest Version

- [x] I have tested the issue with the latest version of the product

#### Expectation

`axe.run()` should not report a `button-name` violation for the internal `<button>` used by a customizable `<select>` (`<select><button>…</button><option>…</option></select>`). Per the HTML spec, that `<button>` and its `<selectedcontent>` child are inert — excluded from the accessibility tree and not focusable — so `button-name` (and any other rule relying on accessibility-tree visibility or focusability) should treat them as inapplicable, not as a naming violation.

#### Actual

axe-core has no awareness of this select-specific inert rule. `lib/commons/dom/is-inert.js` only accounts for the `inert` attribute and elements outside an open modal `<dialog>`. As a result, the scan sees a `<button>` with no text, no `aria-label`, and no wrapping `<label>`, and reports a `critical` `button-name` violation on it. The same gap can also affect any other rule that depends on `isVisibleToScreenReaders`, `isFocusable`, or `isInert` for this subtree (e.g. `nested-interactive`, `color-contrast`).

#### How to Reproduce

```js
axe.run(document.body).then(results => {
  console.log(results.violations);
});
```

against this fixture:

```html
<select>
  <button><selectedcontent></selectedcontent></button>
  <option>Apple</option>
  <option>Banana</option>
</select>
```

`results.violations` includes a `button-name` violation for the `<button>`, even though it is spec-inert and never exposed to assistive technology.

#### Additional context

Suggested fix: teach `isInert` (`lib/commons/dom/is-inert.js`) that any `<button>` that is a direct child of a `<select>`, and any `<selectedcontent>` element, are inert per [the HTML spec's `select` element definition](https://html.spec.whatwg.org/multipage/form-elements.html#the-select-element). This should apply regardless of the select's rendered `appearance` (`base-select` vs. the default UA widget) — inertness comes from being a `select`'s button/selectedcontent, not from how it's styled. This propagates automatically to `isVisibleToScreenReaders` and `focusDisabled`/`isFocusable`, which already delegate to `isInert`. `axe.runVirtualRule('button-name', …)` bypasses the gather-time visibility filter, so `button-name`'s `matches` function should also be updated to skip inert nodes directly.

[...]

PR #5316, opened 2026-08-18
fix(commons/dom): treat a select's button and selectedcontent as inert

ItsMeSousuke:
The HTML spec makes the button and selectedcontent inside a select inert: they are excluded from the accessibility tree and cannot be focused. Without this, the customizable-select (`appearance: base-select`) markup reports a false-positive `button-name` violation.

#### Describe the changes

- `isInert` (`lib/commons/dom/is-inert.js`) now treats any `<button>` that is a direct child of a `<select>`, and any `<selectedcontent>` element, as inert, per the [HTML spec's `select` element definition](https://html.spec.whatwg.org/multipage/form-elements.html#the-select-element). This is unconditional — it doesn't check the select's rendered `appearance` — since inertness is an HTML behavior, not a CSS one. This change propagates to `isVisibleToScreenReaders` and `focusDisabled`/`isFocusable`, which already delegate to `isInert`, so `nested-interactive`, `color-contrast`, and any other rule relying on those utilities also benefit.
- Added `lib/rules/button-name-matches.js`, wrapping `no-explicit-name-required-matches` with an `isInert` check, and pointed `button-name.json`'s `matches` at it. This is needed because `axe.runVirtualRule` forces `excludeHidden: false` and bypasses the gather-time visibility filter that `axe.run` applies.
- Added unit tests covering the select button/`selectedcontent` cases (including Shadow DOM) in `test/commons/dom/is-inert.js`, `is-visible-to-screenreader.js`, `focus-disabled.js`, and `is-focusable.js`.
- Added `test/rule-matches/button-name-matches.js` and a `SerialVirtualNode` case in `test/integration/virtual-rules/button-name.js` to cover the `matches` function directly.
- Added an inapplicable select-button fixture to `test/integration/rules/button-name/button-name.html`.

#### Closes:

Closes issue #5315

### DIFF
--- a/lib/commons/dom/is-inert.js
+++ b/lib/commons/dom/is-inert.js
@@ -17,10 +17,25 @@ export default function isInert(vNode, { skipAncestors, isAncestor } = {}) {
 }
 
 /**
+ * Determine if a node is a select element's button or selectedcontent.
+ * These are inert per the HTML spec, regardless of the select's rendered
+ * appearance (e.g. `appearance: base-select`).
+ * @see https://html.spec.whatwg.org/multipage/form-elements.html#the-select-element
+ */
+function isInertSelectContent(vNode) {
+  const { nodeName } = vNode.props;
+  if (nodeName === 'selectedcontent') {
+    return true;
+  }
+
+  return nodeName === 'button' && vNode.parent?.props.nodeName === 'select';
+}
+
+/**
  * Check the element for inert
  */
 const isInertSelf = memoize(function isInertSelfMemoized(vNode, isAncestor) {
-  if (vNode.hasAttr('inert')) {
+  if (vNode.hasAttr('inert') || isInertSelectContent(vNode)) {
     return true;
   }
 
--- a/lib/rules/button-name-matches.js
+++ b/lib/rules/button-name-matches.js
@@ -1,0 +1,9 @@
+import noExplicitNameRequired from './no-explicit-name-required-matches';
+import { isInert } from '../commons/dom';
+
+export default function buttonNameMatches(node, virtualNode) {
+  if (isInert(virtualNode)) {
+    return false;
+  }
+  return noExplicitNameRequired(node, virtualNode);
+}
--- a/lib/rules/button-name.json
+++ b/lib/rules/button-name.json
@@ -2,7 +2,7 @@
   "id": "button-name",
   "impact": "critical",
   "selector": "button",
-  "matches": "no-explicit-name-required-matches",
+  "matches": "button-name-matches",
   "tags": [
     "cat.name-role-value",
     "wcag2a",
--- a/test/commons/dom/focus-disabled.js
+++ b/test/commons/dom/focus-disabled.js
@@ -74,6 +74,14 @@ describe('dom.focus-disabled', () => {
     assert.isTrue(focusDisabled(vNode));
   });
 
+  it('returns true for a select button', () => {
+    const vNode = queryFixture(
+      '<select><button id="target"><selectedcontent></selectedcontent></button><option>a</option></select>'
+    );
+
+    assert.isTrue(focusDisabled(vNode));
+  });
+
   describe('SerialVirtualNode', () => {
     it('returns false if element is hidden for everyone', () => {
       const vNode = new axe.SerialVirtualNode({
--- a/test/commons/dom/is-focusable.js
+++ b/test/commons/dom/is-focusable.js
@@ -26,6 +26,15 @@ describe('dom.isFocusable', () => {
     assert.isTrue(axe.commons.dom.isFocusable(el));
   });
 
+  it('should return false for a select button', () => {
+    fixture.innerHTML =
+      '<select><button id="target"><selectedcontent></selectedcontent></button><option>a</option></select>';
+    const el = document.getElementById('target');
+    flatTreeSetup(fixture);
+
+    assert.isFalse(axe.commons.dom.isFocusable(el));
+  });
+
   it('should return true for visible, enabled, non-hidden inputs', () => {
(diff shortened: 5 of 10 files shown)


straker:
Thanks for the pr. This is a good first pass, but I think needs some more work before we can merge it in. I tested to see how `<button>` is treated inside a `<select>` element, and it seems that regardless of if the button is the first child of a select the button itself is never rendered in any browser and no AT can navigate to them. In the following example neither of the 3 buttons are visible / navigatable.

```html
<select>
  <button>foo</button>
  <button>bar</button>
  <option>Hello</option>
  <span><button>bar</button></span>
</select>
```

This to me suggests that a `<button>` should return `true` for the function `isHiddenForEveryone` when it is a descendant of a `<select>`. An interesting thing though is that when the `<button>` is a child of an `<option>` it is visible but treated as the `<option>` rather than as a separate button, so I think it's fine if we treat the button as hidden still.

I think the way we want to handle this is to add a new [visibility method](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/dom/visibility-methods.js) for buttons that are descendants of selects (something like `selectButtonHidden`) and then use that in the `isHiddenForEveryone` function.

straker:
@ItsMeSousuke Would you have time to update the pr, or would you like us to take it over?

ItsMeSousuke:
> @ItsMeSousuke Would you have time to update the pr, or would you like us to take it over?

I've updated this PR, please check it after that.

lib/commons/dom/visibility-methods.js
```diff
@@ -257,6 +257,31 @@ export function detailsHidden(vNode) {
...
+  }
+
+  let node = vNode.parent;
+  while (node) {
+    if (node.props.nodeName === 'select') {
+      return true;
+    }
+
+    node = node.parent;
+  }
```
straker: Doing an ancestor look up on the entire ancestor tree for every button on the page is going to be a huge performance problem. Mostly because most buttons are not going to be descendants of selects so this will always navigate up to the root element for every button. 

To solve for these types of problems where we need to know if a specific ancestor exists we use `memoize` on a recursive function. That way every node on the tree only gets asked the question once (is there a select parent) and if two buttons share the same ancestor, that shared ancestor already knows the answer.

An example of this kind of ancestry lookup is [is-fixed-position](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/dom/is-fixed-position.js). Lastly we make sure to use [primitive mode for memoize](https://github.com/medikoo/memoizee#primitive-mode) to speed up performance even further.

```js
export function selectButtonHidden(vNode) {
  if (vNode.props.nodeName !== 'button') {
    return false;
  }

  if (!vNode.parent) {
    return false;
  }

  return hasSelectAncestor(vNode.parent);
}

const hasSelectAncestor = memoize(function hasSelectAncestorMemoized(vNode) {
  if (vNode.props.nodeName === 'select') {
    return true;
  }

  if (!vNode.parent) {
    return false;
  }

  return hasSelectAncestor(vNode.parent);
}, { primitive: true });
```

straker: Instead of a loop here we can use `closest` to find the first `select` ancesotr

```suggestion
  if (closest(node, 'select')) {
    return true;
  }
```

test/commons/dom/is-hidden-for-everyone.js
```diff
@@ -395,5 +395,81 @@ describe('dom.isHiddenForEveryone', () => {
       const actual = isHiddenForEveryone(vNode);
       assert.isFalse(actual);
     });
+
+    it('should return true for a button inside a select', () => {
```
straker, line 399: These tests should be go in their own describe block and not the `SerialVirtualNode` describe, similar to the other display tests. Can add it right after the `details` describe block.

lib/commons/dom/visibility-methods.js
straker, line 280: Instead of a loop here we should use `closest`


```suggestion
  if (closest(vNode, 'select')) {
    return true;
  }
```

test/commons/dom/is-hidden-for-everyone.js
```diff
@@ -395,5 +395,81 @@ describe('dom.isHiddenForEveryone', () => {
...
+    });
+
+    it('should return true for a select button inside an open shadow DOM', () => {
+      const vNode = queryShadowFixture(
+        '<div id="shadow"></div>',
+        '<select><button id="target"></button><option>a</option></select>'
+      );
+
+      assert.isTrue(isHiddenForEveryone(vNode));
+    });
```
straker, line 473: A shadow DOM test isn't needed unless the test in question needs to test if the content can cross shadow DOM boundaries. Select elements can't be shadow roots so we don't need to add a test for it.

straker (changes requested):
Alright, so I dug into this a bit more and found I made a mistake before. I didn't realize that you needed to use the CSS style `select, ::picker(select) { appearance: base-select; }` in order to get the custom select to work. The CSS appearance changes the entire answer to when a `button` is visible or not.

When a select is not `appearance: base-select` (e.g `auto`) no button will ever show, which is what my original answer was. However when `appearance: base-select` is set then the browser will not render only the first button as the spec says, but any other button will be rendered. Also the first child element must be a button in order for it not to render, if the button is wrapped in an element then it will still render, and if the 2nd child is also a button it will still render even if the first did.

So what that means is that the `selectButtonHidden` needs to do a few things:

1) find the `select` element and check it's `appearance` CSS value (use `vNode.getComputedStylePropertyValue('appearance')`
2) if the `appearance` is not `base-select`, then any button is hidden
3) if the `appearance` is `base-select`, then only if the button is the first direct child of the select is it hidden, all others are visible. You can copy how `detailsHidden` did that check.

PR #5321, opened 2026-08-20
perf(create-grid): memoize findScrollRegionParent

### LINKED ISSUE #5296: Memoize `findScrollRegionParent` in `create-grid.js`
The majority of time spent creating the grid in `create-grid.js` is spent in the internal function `findScrollRegionParent`. `findScrollRegionParent` is not memoized and does an entire ancestor tree traversal so memoizing that using primitive mode results in huge performance increase. Make sure to use primitive mode.

chutchins25:
Closes #5296

Was stacked on #5319, which has now merged; this targets `develop` and contains a single commit touching one file.

## What

Replaces the hand-rolled `_scrollRegionParent` cache in `findScrollRegionParent` with a memoized, recursive function, per @straker's suggestion on the issue. Net -8 lines.

## Why

The hand-rolled cache had a hole: it stored `null` for nodes with no scroll-region ancestor, and the short-circuit tested truthiness —

```js
if (parentVNode._scrollRegionParent) {   // null is falsy
```

so on pages without scroll regions the cache never hit and **every node re-walked to the document root**. Memoizee caches a `null` return like any other value, so recursing through the memoized function covers that case for free.

Measured ancestor-walk iterations per call:

| Fixture | Iterations/call |
|---|---|
| 12,000 elements, 120 scroll regions | 1.08 |
| 1,362 elements, 40 scroll regions | 1.06 |
| **12,002 elements, 40 deep, no scroll regions** | **25.91** |

## Results

`createGrid()` in isolation, medians of 24 runs:

| Fixture | Before | After | Change |
|---|---|---|---|
| 12k elements, 40 deep, no scroll regions | 343.5 ms | 201.2 ms | **-41.4%** |
| 12k elements, 120 scroll regions | 191.9 ms | 183.6 ms | -4.3% |
| 1.4k elements, 40 scroll regions | 19.9 ms | 20.4 ms | neutral |

Results are identical on every fixture (same violations, passes, incomplete and relatedNodes counts). Full unit suite: 484 test files passing.

## Note on the call site

`findScrollRegionParent` previously took `(vNode, parentVNode)` and handled a null `parentVNode` via `while (parentVNode)`. The memoized version dereferences `parentVNode.actualNode` immediately, so the root case is guarded at the call site instead:

```js
const scrollRegionParent = parentVNode
  ? findScrollRegionParent(parentVNode)
  : null;
```

The `vNode` argument is gone — it only existed to seed the `checkedNodes` list for the old cache.

### DIFF
--- a/lib/commons/dom/create-grid.js
+++ b/lib/commons/dom/create-grid.js
@@ -2,6 +2,7 @@ import isVisibleOnScreen from './is-visible-on-screen';
 import { getBoundingRect } from '../math/get-bounding-rect';
 import { isPointInRect } from '../math/is-point-in-rect';
 import VirtualNode from '../../core/base/virtual-node/virtual-node';
+import memoize from '../../core/utils/memoize';
 import { getNodeFromTree, getScroll, isShadowRoot } from '../../core/utils';
 import constants from '../../core/constants';
 import cache from '../../core/base/cache';
@@ -86,7 +87,9 @@ export default function createGrid(
 
     vNode._stackingOrder = createStackingOrder(vNode, parentVNode, nodeIndex++);
 
-    const scrollRegionParent = findScrollRegionParent(vNode, parentVNode);
+    const scrollRegionParent = parentVNode
+      ? findScrollRegionParent(parentVNode)
+      : null;
     const grid = scrollRegionParent ? scrollRegionParent._subGrid : rootGrid;
 
     if (getScroll(vNode.actualNode)) {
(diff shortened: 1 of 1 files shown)


lib/commons/dom/create-grid.js
```diff
@@ -355,38 +358,27 @@ function getRealZIndex(vNode, parentVNode) {
...
+ * @param {VirtualNode} parentVNode
  * @return {VirtualNode|null}
  */
-function findScrollRegionParent(vNode, parentVNode) {
-  let scrollRegionParent = null;
-  const checkedNodes = [vNode];
-
-  while (parentVNode) {
+const findScrollRegionParent = memoize(
+  function findScrollRegionParentMemoized(parentVNode) {
```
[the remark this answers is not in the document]
chutchins25: Stale — this reviewed `b320217f`, which was the pre-rebase head.

The extra files were phantoms, not scope creep. #5319 was squash-merged, so its individual commits never landed on `develop`; this branch still carried them even after GitHub auto-retargeted the base. Replayed the single real commit with `git rebase --onto origin/develop`, now at `4bc0900a`.

Current diff:

```
lib/commons/dom/create-grid.js | 42 +++++++++++++-----------------------
1 file changed, 17 insertions(+), 25 deletions(-)
```

One commit, one file. The title and description match that scope.

straker, line 364: We can simplify this a bit if we test for `null` first 

```js
function findScrollRegionParentMemoized(parentVNode) {
  if (!parentVNode) {
    return null;
  } 
```

this would allow us to not ternary the call to `findScrollRegionParent` and have to check for `grandParentVNode` before recusing the call.

straker (changes requested):

lib/commons/dom/create-grid.js
chutchins25: Done in `1530b41a` — that reads much better. The call site is back to a plain `findScrollRegionParent(parentVNode)` and the `grandParentVNode` variable is gone entirely, since the recursive call can pass `getNodeFromTree(...)` straight through.

`createGrid` on the no-scroll-region fixture is unchanged at 209.5ms vs 343.5ms on develop (24-run medians), full suite green.

straker (approved):

PR #5323, opened 2026-08-20
perf(color): replace private class members in Color

### LINKED ISSUE #5295: Don't use private class properties and functions in Color class
Babel downcompiles any private functions and properties in a class to be a lookup function that takes a lot of overhead when run many times. This becomes a problem in color contrast as every time we parse a color string we have to access the private color properties as well as instantiate a new Color object which also suffers a lot of overhead due to the private properties. This can be fixed by not using private properties and functions in the [Color class](https://github.com/dequelabs/axe-core/blob/develop/lib/commons/color/color.js).

Unfortunately our version of [color.js](https://github.com/color-js/color.js) uses private properties that are also accessed through `Color.parseString`, so we can't avoid that at the moment. However our version is an older version and they may have removed the private properties in a newer version. We would have to investigate if that's true and also if we can update to latest version as it was pinned due to a [bug with prototype.js](https://github.com/dequelabs/axe-core/issues/4428).

For now though we can cache the return value of `color = new Colorjs(colorString)` to save on the amount of times we call colorjs:

```js
parseString(colorString) {
  let color;
  const parseStringMap = cache.get('parseStringMap', () => { return {}; });

  if (parseStringMap[colorString]) {
    color = parseStringMap[colorString];
  }
  else {
    // rest of code

    color = new Colorjs(colorString)
      .toGamut({
        space: 'srgb',
        method: 'clip'
      })
      .to('srgb');
    parseStringMap[colorString] = color;
  }

  this.r = color.r;
  this.g = color.g;
  this.b = color.b;
  // color.alpha is a Number object so convert it to a number
  this.alpha = +color.alpha;

  return this;
}
```

chutchins25:
Closes #5295

Two commits, covering both halves of the issue:

- **`769c85d4`** — remove the private class members from `Color`.
- **`c37e5e1d`** — stop calling colorjs repeatedly for the same colour string.

## What

**Private members.** The six private channel fields `#r #g #b #red #green #blue` become `_r _g _b _red _green _blue`, and the private method `#add(value)` becomes a module level `addToChannels(color, value)` helper. The getter/setter pairs and the 0-1 ↔ 0-255 synchronisation are unchanged.

**Parse cache.** `parseString` now memoises its result for the run in a `Map` held in the axe cache, keyed on the colour string. The lookup happens before the `hsl` rewrite so a hit skips that too, and the entry is only written on success, so unparseable colours still throw and record `incompleteData` exactly as before.

## Why

Babel compiles private fields and methods into per-instance helper calls. A CPU profile of a 12,000 element page attributed **852ms of 5,323ms to `_classPrivateFieldInitSpec` alone**, every caller being `Color`:

```
303ms  Color <- createStackingContext <- addToStackingContext
 86ms  Color <- getOwnBackgroundColor <- _getBackgroundColor
 66ms  Color <- getOwnBackgroundColor <- _getStackingContext
 65ms  Color <- createStackingContext <- _stackingContextToColor
```

Separately, every `parseString` call built a fresh `Colorjs` instance and ran it through `toGamut()` and a colour space conversion. A page reuses a small palette across thousands of elements, so nearly all of that work was repeated. Note that `parseString`'s own self time is only ~27ms — the cost is almost entirely in its callees, which is why caching it is worth far more than a self-time reading suggests.

## Results

Chrome, medians of 12 runs:

| Fixture | Before | After private members | After both | Total change |
|---|---|---|---|---|
| 12,000 elements | 5393 ms | 3465 ms | **2772 ms** | **-48.6%** |
| 4,000 paragraphs, 8×4 colour palette | 1862 ms | — | **1472 ms** | **-21.0%** |

`_classPrivateFieldInitSpec` drops from 935ms to 40ms; the residue is the bundled colorjs, which still uses private fields internally and is pinned by #4428. Results are identical throughout — same violations, passes, incomplete and relatedNodes counts.

## Observable changes

Private fields are not merely an implementation detail, so this is worth stating plainly rather than calling it a no-op:

1. **The channel storage is now enumerable.** `Object.keys(color)` and `{ ...color }` include the six underscore properties, where previously `alpha` was the only own enumerable property. `JSON.stringify` is unaffected, because `toJSON()` already controls serialisation.
2. **The channel storage is now writable from outside.** Assigning `color._r` directly no longer updates `_red`, so a caller can put a `Color` into an inconsistent state. Going through the documented `r` / `red` setters still maintains the pair, exactly as before.

This matches how every other public class in axe-core stores its internals — `VirtualNode._cache` / `_isHidden` / `_isXHTML` / `_type`, `SerialVirtualNode._props` / `_attrs`, and `DqElement._element` / `_virtualNode` / `_options` are all enumerable and writable.

Deriving one representation from the other, so the invariant could not be broken at all, is not equivalent: the setters are deliberately asymmetric. `set r` rounds (`Math.round(clamp(value, 0, 1) * 255)`) while `set red` does not (`clamp(value, 0, 255)`), so `new Color(64.5, …)` preserves `red === 64.5` today. Deriving would round it and change behaviour.

## Testing

This also strengthens the colour suites incidentally: because the channel values were previously private, the ~117 `assert.deepEqual(colorA, colorB)` assertions across `test/commons/color/` were comparing only `alpha`. They now compare all six channels, and still pass.

### DIFF
--- a/lib/commons/color/color.js
+++ b/lib/commons/color/color.js
@@ -124,70 +124,63 @@ export default class Color {
    */
   parseString(colorString) {
     // parsing is by far the most expensive part of Color, and a page reuses
-    // the same handful of color strings across thousands of elements
-    const parsedColors = cache.get('parsedColors', () => new Map());
-    const parsed = parsedColors.get(colorString);
-    if (parsed) {
-      this.r = parsed.r;
-      this.g = parsed.g;
-      this.b = parsed.b;
-      this.alpha = parsed.alpha;
-      return this;
-    }
-    const cacheKey = colorString;
+    // the same handful of color strings across thousands of elements.
+    // null prototype so a colour string like `constructor` can't resolve to
+    // an inherited property
+    const parsedColors = cache.get('parsedColors', () => Object.create(null));
+    let color = parsedColors[colorString];
 
-    // Colorjs <v0.5.0 does not support rad or turn angle values
-    // @see https://github.com/LeaVerou/color.js/issues/311
-    colorString = colorString.replace(hslRegex, (match, angle, unit) => {
-      const value = angle + unit;
+    if (!color) {
+      const cacheKey = colorString;
 
-      switch (unit) {
-        case 'rad':
-          return match.replace(value, radToDeg(angle));
-        case 'turn':
-          return match.replace(value, turnToDeg(angle));
+      // Colorjs <v0.5.0 does not support rad or turn angle values
+      // @see https://github.com/LeaVerou/color.js/issues/311
+      colorString = colorString.replace(hslRegex, (match, angle, unit) => {
+        const value = angle + unit;
+
+        switch (unit) {
+          case 'rad':
+            return match.replace(value, radToDeg(angle));
+          case 'turn':
+            return match.replace(value, turnToDeg(angle));
+        }
+      });
+
+      try {
+        // revert prototype.js override of Array.from
+        // in order to get color-contrast working
+        // @see https://github.com/dequelabs/axe-core/issues/4428
+        let prototypeArrayFrom;
+        if ('Prototype' in window && 'Version' in window.Prototype) {
+          prototypeArrayFrom = Array.from;
+          Array.from = ArrayFrom;
+        }
+
+        // srgb values are between 0 and 1
+        color = new Colorjs(colorString)
+          .toGamut({
+            space: 'srgb',
+            method: 'clip'
+          })
+          .to('srgb');
+
+        if (prototypeArrayFrom) {
+          Array.from = prototypeArrayFrom;
+          prototypeArrayFrom = null;
+        }
+      } catch {
+        incompleteData.set('colorParse', colorString);
+        throw new Error(`Unable to parse color "${colorString}"`);
       }
-    });
 
-    try {
-      // revert prototype.js override of Array.from
-      // in order to get color-contrast working
-      // @see https://github.com/dequelabs/axe-core/issues/4428
-      let prototypeArrayFrom;
-      if ('Prototype' in window && 'Version' in window.Prototype) {
-        prototypeArrayFrom = Array.from;
-        Array.from = ArrayFrom;
-      }
-
-      // srgb values are between 0 and 1
-      const color = new Colorjs(colorString)
-        .toGamut({
-          space: 'srgb',
-          method: 'clip'
-        })
-        .to('srgb');
-
-      if (prototypeArrayFrom) {
-        Array.from = prototypeArrayFrom;
-        prototypeArrayFrom = null;
-      }
-
-      this.r = color.r;
-      this.g = color.g;
-      this.b = color.b;
-      // color.alpha is a Number object so convert it to a number
-      this.alpha = +color.alpha;
-    } catch {
-      incompleteData.set('colorParse', colorString);
-      throw new Error(`Unable to parse color "${colorString}"`);
+      parsedColors[cacheKey] = color;
     }
 
-    parsedColors.set(cacheKey, {
-      r: this.r,
-      g: this.g,
-      b: this.b,
-      alpha: this.alpha
-    });
+    this.r = color.r;
+    this.g = color.g;
+    this.b = color.b;
+    // color.alpha is a Number object so convert it to a number
+    this.alpha = +color.alpha;
 
     return this;
   }
--- a/lib/commons/color/color.js
+++ b/lib/commons/color/color.js
@@ -1,5 +1,6 @@
 import { Colorjs, ArrayFrom } from '../../core/imports';
 import incompleteData from './incomplete-data';
+import cache from '../../core/base/cache';
 
 const hexRegex = /^#[0-9a-f]{3,8}$/i;
 const hslRegex = /hsl\(\s*([-\d.]+)(rad|turn)/;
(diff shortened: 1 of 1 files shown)


chutchins25:
Valid on the facts, declined on the remedy.

Correct that this is observable: `_r`/`_red` are enumerable and writable, so `Object.keys(color)` and spread now include them, and assigning `color._r` directly no longer updates `_red`. The PR body claimed "no API change" — that was wrong and has been corrected with an explicit "Observable changes" section.

Not preserving the prior property shape, for two reasons.

**It matches existing convention rather than introducing a new hazard.** Every comparably public class in axe-core already stores mutable, enumerable internals under an underscore: `VirtualNode._cache` / `_isHidden` / `_isXHTML` / `_type`, `SerialVirtualNode._props` / `_attrs`, `DqElement._element` / `_virtualNode` / `_options`. Assigning `virtualNode._cache` desynchronises that object exactly as readily.

**Making the invariant structural would change behaviour.** Deriving `red` from `_r` so the pair cannot disagree is not equivalent, because the setters are deliberately asymmetric:

```js
set r(value)   { _r = value;       _red = Math.round(clamp(value, 0, 1) * 255); }  // rounds
set red(value) { _r = value / 255; _red = clamp(value, 0, 255); }                  // does not
```

`new Color(64.5, …)` preserves `red === 64.5` today; a derived getter would round it.

Restoring true privacy via per-instance `Object.defineProperty` would reintroduce exactly the per-instance construction cost this PR removes — 852ms of 5,323ms on a 12,000 element page — and would still not prevent mutation, only hide the keys.

straker (dismissed):
This only partially closes the ticket. The other half of the ticket is to not call `colorjs` repeatedly which contains private functions / properties.

chutchins25:
You were right, and my estimate of the size was badly off.

I'd read `parseString`'s self time — 27ms of 5,300ms — and concluded the colorjs half was worth under 1%. That was the wrong number to look at: the cost is almost entirely in its callees (`new Colorjs()`, `toGamut()`, `to('srgb')`), which caching skips as well.

Added in `c37e5e1d`: a run-scoped `Map` in the axe cache keyed on the colour string. The lookup sits before the `hsl` rewrite so a hit skips that too, and the entry is only written on success, so unparseable colours still throw and record `incompleteData` as before. Used a `Map` rather than a plain object so a colour string like `constructor` can't collide with `Object.prototype`.

On a 12,000 element page, medians of 12 runs:

| | Before | After private members | After both |
|---|---|---|---|
| Total `axe.run` | 5393 ms | 3465 ms | **2772 ms** |

That's **-48.6% combined**, and -21.0% on a colour-heavy fixture (4,000 paragraphs over an 8×4 palette). Results identical, full suite green.

lib/commons/color/color.js
```diff
@@ -131,6 +123,19 @@ export default class Color {
    * @instance
    */
   parseString(colorString) {
+    // parsing is by far the most expensive part of Color, and a page reuses
+    // the same handful of color strings across thousands of elements
+    const parsedColors = cache.get('parsedColors', () => new Map());
```
straker: Minor nit: `colorString` is a string so we can just use a plain object here rather than a Map (I tend to only use Map if the key is an object).

lib/commons/color/color.js
```diff
@@ -131,6 +123,19 @@ export default class Color {
...
+    const parsed = parsedColors.get(colorString);
+    if (parsed) {
+      this.r = parsed.r;
```
straker: This duplicates the logic of setting the rgba values 3 times. You can simplify this by setting the value of the cache as the return value of `new Colorjs(colorString).toGamut().to('srgb')` and then always setting the rgba values outside the try/catch/if block.

straker (dismissed):

lib/commons/color/color.js
chutchins25: Switched in `4ed44281`, with one adjustment: it's `Object.create(null)` rather than `{}`.

With a plain object literal, `parsedColors['constructor']` resolves to `Object` through the prototype chain — truthy — so a colour string of `constructor`, `toString` or `hasOwnProperty` would skip parsing entirely and produce a `Color` with `undefined` channels instead of throwing. Verified against a build: with the null prototype all three throw `Unable to parse color "..."` as they should.

Happy to use a bare `{}` if you'd rather; it just seemed worth the one extra token to keep the failure mode.

lib/commons/color/color.js
chutchins25: Done in `4ed44281`. The cache now holds the `Colorjs` result, and the channel assignment happens once after the `if` block:

```js
let color = parsedColors[colorString];

if (!color) {
  // hsl rewrite + parse, then
  parsedColors[cacheKey] = color;
}

this.r = color.r;
this.g = color.g;
this.b = color.b;
this.alpha = +color.alpha;
```

Total run time on a 12,000 element page is unchanged at 2869ms vs 4892ms on develop (12-run medians), full suite green.

lib/commons/color/color.js
```diff
@@ -131,52 +123,65 @@ export default class Color {
...
-      switch (unit) {
-        case 'rad':
-          return match.replace(value, radToDeg(angle));
-        case 'turn':
-          return match.replace(value, turnToDeg(angle));
+    // parsing is by far the most expensive part of Color, and a page reuses
+    // the same handful of color strings across thousands of elements.
+    // null prototype so a colour string like `constructor` can't resolve to
+    // an inherited property
+    const parsedColors = cache.get('parsedColors', () => Object.create(null));
```
straker: We don't need to create an object without a prototype, returning an empty object is fine.


```suggestion
    const parsedColors = cache.get('parsedColors', () => ({}));
```

straker (dismissed):

lib/commons/color/color.js
chutchins25: Applied in `8f8285c6`.

Agreed it doesn't matter here — every call site feeds `parseString` a value from `getPropertyValue(...)`, and a browser will never hand back `constructor` as a computed colour, so the case isn't reachable through axe's own paths.

straker (approved):

Issue #5294, opened 2026-08-06
Use primitive mode for memoizee


straker:
Memoizee has a [primitive mode](https://github.com/medikoo/memoizee#primitive-mode) which performs much faster than the default mode. The one requirement is that the arguments being passed into the function must be stringifiable and produce unique strings. Most of our functions that we memoize take a virtual node, so if we add a `toString` function to virtual node that returns a unique string (such as the `nodeIndex`) then we could use primitive mode for most of our functions. Testing shows this improves functions like `isVisibleToScreenReaders` significantly by between 70%-90%.

We should update all uses of `memoize` to use primitive mode when passed a virtual node. Some functions do not take a virtual node and only take an HTML element. For these elements we can create a passthrough function that gets the virtual node from the node, then calls a memoized function that does the work. For example `getVisibleChildTextRects` could turn into:

```js
function getVisibleChildTextRects(node) {
  const vNode = getNodeFromTree(node);
  return getVisibleChildTextRectsVirtual(vNode);
}
const getVisibleChildTextRectsVirtual = memoize(
  function getVisibleChildTextRectsMemoized(vNode) {
    const node = vNode.actualNode;
   // rest of function
```

straker:
Updating the function to have a passthrough for domNode to virtual node that we can memoize already has precedent in https://github.com/dequelabs/axe-core/blob/develop/lib/commons/dom/is-visible-to-screenreader.js

straker:
Just a note, the linked pr only resolved half the this issue. The other half was about using primitive mode even for functions that take DOM nodes, which requires a pass through function. So the issue is still left open until we do that.

chutchins25:
Picked up the passthrough half in #5324. Three of the five memoized DOM-node functions convert cleanly — `getVisibleChildTextRects` (your example), `toGrid`, and `getPseudoElementArea`. Two don't, and one of them is a design question for you rather than something I wanted to decide.

**`getScroll` — the passthrough approach conflicts with the `core/utils` boundary.** Adding one there fails our own eslint rule:

> Utils is meant for utility functions that work independently of axe's state; utilities that require the virtual tree to be set up should go in commons, not utils.

The rule has an `ignores` list, but it's described as files with "known uses of virtual node that are legacy before this rule was enforced", so adding `get-scroll.js` to it would be working against the rule rather than qualifying for it. As far as I can tell the options are:

[...]

PR #5324, opened 2026-08-21
perf(commons): use primitive mode for DOM node memoizations

chutchins25:
Refs #5294 — the passthrough half, following on from #5319.

## The criterion

Primitive mode builds its cache key by calling `String()` on each argument up to `fn.length`, so a function is convertible only if **every** keyed argument stringifies uniquely. After #5319 a virtual node does, via `AbstractVirtualNode#toString`; strings, numbers and booleans always have. Objects do not — every `HTMLTableCellElement` stringifies to the same `[object HTMLTableCellElement]`.

Applying that to the five memoized functions that take a DOM node:

| Function | Arguments | Convertible |
|---|---|---|
| `getVisibleChildTextRects` | `(node)` | ✅ single node → virtual node |
| `toGrid` | `(node)` | ✅ single node → virtual node |
| `getPseudoElementArea` | `(node, pseudo)` | ✅ node → virtual node, `pseudo` is `':before'` / `':after'` |
| `getScroll` | `(elm, buffer)` | ⛔ keys fine, but blocked by an architectural rule — see below |
| `getCellPosition` | `(cell, tableGrid)` | ⛔ second argument is an array of cells |

## What changed

The three convertible functions now resolve the node to a virtual node and delegate to a primitive-mode memoized function, so the cache keys on a string rather than object identity:

- **`getVisibleChildTextRects`** — the example given in the issue. It already called `getNodeFromTree` internally; that now happens in the passthrough.
- **`toGrid`** — same shape, plus an uncached path for tables outside the virtual tree (see below).
- **`getPseudoElementArea`** — no passthrough needed: both call sites already hold the virtual node and were passing `vNode.actualNode`, so it now takes the virtual node directly.

## Honest note on the numbers

**This is a consistency change, not a measured win.** Neither function is hot enough for the cache-key lookup to register:

| | develop | this branch |
|---|---|---|
| `getVisibleChildTextRects` (4,000 paragraph page) | 0 ms | 0 ms |
| `toGrid` (60 tables × 26 × 8) | 2 ms | 0 ms |

Whole-run timing is flat in both directions across two rounds of 12, with overlapping ranges — no gain and no regression. The value is that these no longer scale as O(N²) in cache lookups if they ever do become hot, and that the codebase applies one pattern consistently.

## A note on the tests

`toGrid` is called from tests that set `fixture.innerHTML` without building a virtual tree, so `getNodeFromTree` returned `null` once it started resolving the node. Since any function in `commons` can expect the tree to be set up, the fix is in the tests: `test/commons/table/to-grid.js` and `test/checks/tables/caption-faked.js` now call `axe.testUtils.flatTreeSetup(fixture.firstElementChild)`, matching `test/commons/table/get-cell-position.js`.

## Why the other two are not converted

**`getScroll` — blocked by the `core/utils` boundary, not by its arguments.** `(elm, buffer)` would key perfectly well as a virtual node plus a number. The problem is that adding a passthrough means touching a virtual node inside `lib/core/utils/`, which our own eslint rule rejects:

> Utils is meant for utility functions that work independently of axe's state; utilities that require the virtual tree to be set up should go in commons, not utils.

There is an `ignores` list, but it is explicitly for files with "known uses of virtual node that are legacy before this rule was enforced", so adding `get-scroll.js` would be working around the rule rather than qualifying for it. The options are to move `getScroll` to `commons` — it is public API as `axe.utils.getScroll` — to widen the exemption, or to leave it. That is a design call, so it is raised on #5294 rather than decided here. It applies to any other `core/utils` function we would want to convert this way.

**`getCellPosition` — the second argument defeats the key.** It takes `(cell, tableGrid)` where `tableGrid` is an array of cells. The primitive key would be `String(tableGrid)`, and since every cell stringifies to `[object HTMLTableCellElement]`, two structurally identical tables produce an identical key — the second table would silently receive the first table's positions.

Keying on `cell` alone would be correct today: the position is a function of the cell, and all three callers pass `toGrid(findUp(cell, 'table'))`, the cell's own grid. But that makes correctness depend on no future caller passing a foreign grid, so it is left in default mode pending a decision on #5294.

## Testing

Full unit suite: 484 test files passing. Results identical on every benchmark fixture.

### DIFF
--- a/lib/commons/table/to-grid.js
+++ b/lib/commons/table/to-grid.js
@@ -9,16 +9,11 @@ import { getNodeFromTree, memoize } from '../../core/utils';
  * @return {Array<HTMLTableCellElement>} Array of HTMLTableCellElements
  */
 function toGrid(node) {
-  const vNode = getNodeFromTree(node);
-  // tables outside the virtual tree have no key to cache against
-  return vNode ? toGridVirtual(vNode) : computeGrid(node);
+  return toGridVirtual(getNodeFromTree(node));
 }
 
 function toGridMemoized(vNode) {
-  return computeGrid(vNode.actualNode);
-}
-
-function computeGrid(node) {
+  const node = vNode.actualNode;
   const table = [];
   const rows = node.rows;
   for (let i = 0, rowLength = rows.length; i < rowLength; i++) {
--- a/lib/commons/table/to-grid.js
+++ b/lib/commons/table/to-grid.js
@@ -1,4 +1,4 @@
-import { memoize } from '../../core/utils';
+import { getNodeFromTree, memoize } from '../../core/utils';
 
 /**
  * Converts a table to an Array of arrays, normalized for row and column spans
@@ -9,6 +9,16 @@ import { memoize } from '../../core/utils';
  * @return {Array<HTMLTableCellElement>} Array of HTMLTableCellElements
  */
 function toGrid(node) {
+  const vNode = getNodeFromTree(node);
+  // tables outside the virtual tree have no key to cache against
+  return vNode ? toGridVirtual(vNode) : computeGrid(node);
+}
+
+function toGridMemoized(vNode) {
+  return computeGrid(vNode.actualNode);
+}
+
+function computeGrid(node) {
   const table = [];
   const rows = node.rows;
   for (let i = 0, rowLength = rows.length; i < rowLength; i++) {
@@ -45,4 +55,6 @@ function toGrid(node) {
   return table;
 }
 
-export default memoize(toGrid);
+const toGridVirtual = memoize(toGridMemoized, { primitive: true });
+
+export default toGrid;
--- a/lib/checks/color/color-contrast-evaluate.js
+++ b/lib/checks/color/color-contrast-evaluate.js
@@ -192,8 +192,8 @@ function findPseudoElement(
   const rect = vNode.boundingClientRect;
   const minimumSize = rect.width * rect.height * pseudoSizeThreshold;
   do {
-    const beforeSize = getPseudoElementArea(vNode.actualNode, ':before');
-    const afterSize = getPseudoElementArea(vNode.actualNode, ':after');
+    const beforeSize = getPseudoElementArea(vNode, ':before');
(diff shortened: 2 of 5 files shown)


lib/commons/table/to-grid.js
```diff
@@ -9,6 +9,16 @@ import { memoize } from '../../core/utils';
  * @return {Array<HTMLTableCellElement>} Array of HTMLTableCellElements
  */
 function toGrid(node) {
+  const vNode = getNodeFromTree(node);
+  // tables outside the virtual tree have no key to cache against
+  return vNode ? toGridVirtual(vNode) : computeGrid(node);
```
straker: So this might have happened if one of the tests that ends up calling `toGrid` didn't set up the virtual tree first. This is legacy and any function inside `commons` can be expected to have set up the virtual tree before calling. 

Instead of having a branching path here, we should update the test files that end up calling `toGrid` to set up the virtual tree first, either by setting up the test code using `queryFixture` or `checkSetup`, or if that isn't easy to use `flatTreeSetup` after the test code is created.

straker (dismissed):

lib/commons/table/to-grid.js
chutchins25: You're right — the failing tests were the bug, not `toGrid`. Fixed in `f0f30141`.

The branching path is gone; `toGrid` is back to a single line:

```js
function toGrid(node) {
  return toGridVirtual(getNodeFromTree(node));
}
```

Two files were setting `fixture.innerHTML` without building a tree — `test/commons/table/to-grid.js` and `test/checks/tables/caption-faked.js`, six tests each. Both now call `axe.testUtils.flatTreeSetup(fixture.firstElementChild)` after the fixture, matching what `test/commons/table/get-cell-position.js` already does next door. Full suite green, 484 files.

I didn't copy that file's `axe._tree = undefined` teardown, since CLAUDE.md says `fixture` and `axe._tree` are auto-cleared and only `checkContext` needs resetting — let me know if you'd rather it be explicit.

I also removed the section of the PR description that justified the uncached path.

straker (approved):

PR #5326, opened 2026-08-24
perf(aria): cache ARIA values on the virtual node

### LINKED ISSUE #5300: Investigate caching/memoizing `getAriaValue`, `hasAriaValue`, and `getResolvedRefs`
We discussed caching `get/hasAriaValue` on the virtual node (similar to `tabbableElements`, etc.). We should investigate if we should cache each individual property or just a select high use ones (e.g. `aria-sort` is probably only ever looked up once, while `aria-hidden` would be looked up a lot). While we're at it we should also investigate caching `getResolvedRefs`.

chutchins25:
Closes #5300

## What

`getAriaValue`, `hasAriaValue` and `getResolvedRefs` now cache their result on the virtual node's existing `_cache` object, keyed by attribute name plus whichever option changes the answer — `lowercase` for `getAriaValue`, `self` for `getResolvedRefs`. Serialized virtual nodes have no `_cache` and fall through uncached.

## On the question in the issue

The issue asked whether to cache every property or only high-use ones. Instrumenting a run over an ARIA-heavy page (1,500 groups with labels, descriptions, states and idrefs) gives 112,524 calls, of which 13.3% would be repeats. Your split was right — a handful of attributes are read several times per node, and many are read exactly once:

| Call | Repeat rate |
|---|---|
| `getResolvedRefs:aria-labelledby` | 50% |
| `hasAriaValue:aria-labelledby` | 37% |
| `getAriaValue:aria-hidden` | 27% |
| `getAriaValue:aria-label` | 25% |
| `hasAriaValue:aria-owns` | 20% |
| `aria-disabled`, `aria-errormessage`, `aria-invalid`, `aria-level`, `aria-checked` | 0% |

This caches all of them rather than a selected list, since the cache is a property write on an object the virtual node already carries and a cold entry costs one missed key lookup.

## Honest note on the numbers

**I measure this as neutral, and I would rather say so than dress it up.** Three alternating rounds of ten runs on the ARIA-heavy fixture:

|---|---|---|
| 1 | 3982 ms | 4036 ms |
| 2 | 4026 ms | 3965 ms |
| 3 | 3719 ms | 3809 ms |

The sign flips between rounds, so it is inside the run-to-run noise on my machine. A 12,000 element page is likewise unchanged (4997 ms vs 4951 ms). An earlier prototype appeared to show -13.6%, but that did not survive repeated rounds — it was drift.

So the case for this rests on the call counts rather than a wall-clock win I can demonstrate. If you know of page shapes where these lookups dominate, they would be worth pointing the benchmark at; my fixtures may simply not be the right ones.

## The trade-off worth reviewing

Caching means an ARIA value is read once per node per run instead of on every call, so mutating an attribute part way through a run no longer changes the result.

That crosses a line the code currently draws deliberately: `VirtualNode` caches `props` and `attrNames` in `_cache`, but `attr()` reads live through `getAttribute`. Two tests in `test/checks/label/title-only.js` depended on the live read — they evaluated a node, set `aria-label` on it, and evaluated it again. They now use two separate elements and assert the same two behaviours without relying on a mid-run mutation. Nothing else in the suite was affected.

The pattern is reachable through public API — `axe.setup()` followed by manual check evaluation — so it is worth a deliberate decision rather than my assumption.

## Testing

Full unit suite: 484 test files passing. Results identical on every benchmark fixture.

### DIFF
--- a/lib/commons/aria/get-aria-value.js
+++ b/lib/commons/aria/get-aria-value.js
@@ -39,6 +39,14 @@ export default function getAriaValue(node, attrName, options = {}) {
   const { vNode } = nodeLookup(node);
   const { lowercase } = options;
 
+  // ARIA values are read repeatedly for the same node during a run, so cache
+  // them on the virtual node. Serialized nodes have no cache and skip this.
+  const cacheKey = `getAriaValue:${attrName}:${!!lowercase}`;
+  const nodeCache = vNode?._cache;
+  if (nodeCache && cacheKey in nodeCache) {
+    return nodeCache[cacheKey];
+  }
+
   for (const { source, getValue } of sources) {
     let value = getValue(vNode, attrStandard, attrName);
     if (value === null || value === undefined) {
@@ -64,10 +72,17 @@ export default function getAriaValue(node, attrName, options = {}) {
         ']';
     }
 
-    return { value, source };
+    return cacheValue(nodeCache, cacheKey, { value, source });
   }
 
-  return { value: null, source: null };
+  return cacheValue(nodeCache, cacheKey, { value: null, source: null });
+}
+
+function cacheValue(nodeCache, cacheKey, value) {
+  if (nodeCache) {
+    nodeCache[cacheKey] = value;
+  }
+  return value;
 }
 
 function getAttributeValue(vNode, attrStandard, attrName) {
--- a/lib/commons/aria/has-aria-value.js
+++ b/lib/commons/aria/has-aria-value.js
@@ -17,11 +17,24 @@ export default function hasAriaValue(node, attrName) {
   }
 
   const { vNode } = nodeLookup(node);
-  return (
+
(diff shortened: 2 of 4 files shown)


straker:
I did some further testing. In all of my benchmark sites, not a single one of these 3 functions ever takes more than 50ms in total across all calls in axe-core for a single run. The functions are called tens of thousands of times. For example, on one site the `getAriaValue` function is called 47,303 times just for the value of `aria-hidden` and 33,755 times for the value of `aria-disabled`, but across all calls only took 41.49ms. In my largest benchmark site, `getAriaValue` was called 252,143 times for `aria-hidden` and 152,366 times for `aria-disabled` and took a total of 120.60ms across all calls.

So with @chutchins25's testing and mine I think we are safe to not worry about caching these functions. I'll pull these comments into the issue and close it.

PR #5328, opened 2026-08-25
perf(color-contrast): memoize findPseudoElement

### LINKED ISSUE #5297: Investigate memoizing `findPseudoElement` in `colorContrastEvaluate` using primitive mode
The function isn't currently written in a way that makes it easily memoizeable, but if we can rewrite it to work that way it will help improve the speed of the ancestor lookup.

chutchins25:
Closes #5297

## What

`findPseudoElement` walked the ancestor chain in a loop, which meant it could not benefit from memoization: two sibling elements sharing thirty ancestors each rewalked all thirty. It is now split in two.

```js
function findPseudoElement(vNode, pseudoSizeThreshold) {
  pseudoSizeThreshold ??= 0.25;
  const rect = vNode.boundingClientRect;
  const minimumSize = rect.width * rect.height * pseudoSizeThreshold;
  return findPseudoElementAbove(vNode, minimumSize);
}

const findPseudoElementAbove = memoize(
  function findPseudoElementAboveMemoized(vNode, minimumSize) { ... },
  { primitive: true }
);
```

The outer function resolves the minimum size once; the inner one recurses, so **each ancestor is cached in its own right** and descendants resolving to the same minimum size reuse the answers for the ancestors they share. Thanks @straker — that was the missing piece, and it is the same shape that made #5296 worthwhile.

Two supporting changes let it be memoized at all:

- `ignorePseudo` moves to the call site, so a call that would return immediately never reaches the function
- `minimumSize` is carried down as a number rather than an options object, which also puts it in the cache key

`pseudoSizeThreshold ??= 0.25` is assigned in the body rather than as a default parameter, because default parameters are excluded from `fn.length` and memoizee sizes its key from that — as a default it would silently drop out of the key, exactly as `buffer` did in `getScroll`.

## Honest note on the numbers

**This measures neutral, and the win the ticket was after has already been collected by #5324.**

When the ticket was sized, profiling memoizee's key resolution attributed **49ms** on a 12,000 element page to the `findPseudoElement → colorContrastEvaluate` path. That was not `findPseudoElement` itself — it was `getPseudoElementArea`'s cache lookup, and #5324 converted that function to primitive mode. On current develop, `findPseudoElement` measures 0–2ms of self time and `getPseudoElementArea` measures 0ms.

End-to-end after the recursive rewrite, medians of 10 runs:

| Fixture | develop | this branch |
|---|---|---|
| 120 branches × 30 deep, `::before`/`::after` on every level | 1069 ms | 1056 ms |
| same, second round | 1058 ms | 1058 ms |
| 12,000 elements | 2186 ms | 2183 ms |

The deep fixture exists specifically to stress this: deep ancestor chains, real pseudo content at every level, contrast-checked text at every leaf. The recursion removes the repeated traversal, but the traversal was already cheap — `getPseudoElementArea` has been memoized per ancestor since #5324, so every level was resolving from cache regardless.

So this is better-shaped code that does not move the clock. Happy for it to be closed unmerged like #5326 if you would rather not carry it, and equally happy to be wrong if your benchmark sites show otherwise.

## Test

`should evaluate the same element separately per pseudoSizeThreshold` covers the key that Copilot flagged: one element whose pseudo content covers 22% reports `pseudoContent` at a `0.2` threshold and does not at `0.3`. Verified against the defect rather than just observed passing — setting `{ length: 1 }` on the memoize call, which is what an arity mistake looks like, fails it with `expected 'pseudoContent' to not equal 'pseudoContent'`.

## Testing

### DIFF
--- a/lib/checks/color/color-contrast-evaluate.js
+++ b/lib/checks/color/color-contrast-evaluate.js
@@ -181,24 +181,34 @@ export default function colorContrastEvaluate(node, options, virtualNode) {
   return isValid;
 }
 
-const findPseudoElement = memoize(
-  function findPseudoElementMemoized(vNode, pseudoSizeThreshold) {
-    // assigned in the body rather than as a default parameter: default
-    // parameters are excluded from fn.length, which memoize uses to decide how
-    // many arguments to cache on
-    pseudoSizeThreshold ??= 0.25;
+function findPseudoElement(vNode, pseudoSizeThreshold) {
+  // assigned in the body rather than as a default parameter: default
+  // parameters are excluded from fn.length, which memoize uses to decide how
+  // many arguments to cache on
+  pseudoSizeThreshold ??= 0.25;
 
-    const rect = vNode.boundingClientRect;
-    const minimumSize = rect.width * rect.height * pseudoSizeThreshold;
-    let ancestor = vNode;
-    do {
-      const beforeSize = getPseudoElementArea(ancestor, ':before');
-      const afterSize = getPseudoElementArea(ancestor, ':after');
-      if (beforeSize + afterSize > minimumSize) {
-        // Combined area of before and after exceeds the minimum size
-        return ancestor;
-      }
-    } while ((ancestor = ancestor.parent));
+  const rect = vNode.boundingClientRect;
+  const minimumSize = rect.width * rect.height * pseudoSizeThreshold;
+  return findPseudoElementAbove(vNode, minimumSize);
+}
+
+// recursive rather than a loop so that each ancestor is cached in its own
+// right: descendants that resolve to the same minimum size reuse the answer
+// for the ancestors they share, instead of rewalking the chain
+const findPseudoElementAbove = memoize(
+  function findPseudoElementAboveMemoized(vNode, minimumSize) {
+    if (!vNode) {
+      return undefined;
+    }
+
+    const beforeSize = getPseudoElementArea(vNode, ':before');
+    const afterSize = getPseudoElementArea(vNode, ':after');
+    if (beforeSize + afterSize > minimumSize) {
+      // Combined area of before and after exceeds the minimum size
+      return vNode;
+    }
+
+    return findPseudoElementAbove(vNode.parent, minimumSize);
   },
   { primitive: true }
 );
--- a/lib/checks/color/color-contrast-evaluate.js
+++ b/lib/checks/color/color-contrast-evaluate.js
@@ -55,10 +55,9 @@ export default function colorContrastEvaluate(node, options, virtualNode) {
 
   // if element or a parent has pseudo content then we need to mark
   // as needs review
-  const pseudoElm = findPseudoElement(virtualNode, {
-    ignorePseudo,
-    pseudoSizeThreshold
-  });
+  const pseudoElm = ignorePseudo
+    ? undefined
+    : findPseudoElement(virtualNode, pseudoSizeThreshold);
   if (pseudoElm) {
     this.data({
       fontSize: `${((fontSize * 72) / 96).toFixed(1)}pt (${fontSize}px)`,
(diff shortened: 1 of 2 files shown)


lib/checks/color/color-contrast-evaluate.js
```diff
@@ -182,23 +181,27 @@ export default function colorContrastEvaluate(node, options, virtualNode) {
...
+    const minimumSize = rect.width * rect.height * pseudoSizeThreshold;
+    let ancestor = vNode;
+    do {
+      const beforeSize = getPseudoElementArea(ancestor, ':before');
+      const afterSize = getPseudoElementArea(ancestor, ':after');
+      if (beforeSize + afterSize > minimumSize) {
+        // Combined area of before and after exceeds the minimum size
+        return ancestor;
+      }
+    } while ((ancestor = ancestor.parent));
```
straker: This inner loop should be rewritten to a recursive function in order to take advantage of memoize. Otherwise even if we traverse the same ancestor nodes multiple times using different descendant nodes we have to do this loop every time and get no cache benefit.

chutchins25: You're right, and it's the same point that turned #5296 from "no benefit" into a real win — I should have carried it across. Rewritten in `078991c7`.

The loop is now a recursive `findPseudoElementAbove(vNode, minimumSize)`, memoized, so each ancestor is cached in its own right rather than only the starting node. The minimum size is computed once in the outer function and carried down, which also puts it in the cache key so elements resolving to different thresholds don't share an answer.

Being straight about the measurement though: it is still neutral. deep-ancestor fixture (120 branches × 30 deep, `::before`/`::after` on every level, contrast-checked text at every leaf), medians of 10:

|---|---|---|
| 1 | 1069 ms | 1056 ms |
| 2 | 1058 ms | 1058 ms |

12,000 element page: 2186 ms vs 2183 ms.

My read on why: `getPseudoElementArea` is already memoized per ancestor since #5324, so the walk was resolving from cache at every level already — the recursion removes the traversal, but the traversal was the cheap part. The restructure is better code and I'm happy to keep it, but I don't want to claim a gain the numbers don't show. If it looks different on your benchmark sites, the branch is here.

lib/checks/color/color-contrast-evaluate.js
```diff
@@ -182,23 +181,27 @@ export default function colorContrastEvaluate(node, options, virtualNode) {
...
-    if (beforeSize + afterSize > minimumSize) {
-      return vNode; // Combined area of before and after exceeds the minimum size
-    }
-  } while ((vNode = vNode.parent));
-}
+const findPseudoElement = memoize(
+  function findPseudoElementMemoized(vNode, pseudoSizeThreshold) {
+    // assigned in the body rather than as a default parameter: default
+    // parameters are excluded from fn.length, which memoize uses to decide how
+    // many arguments to cache on
```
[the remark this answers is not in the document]
chutchins25: Valid. Added in `078991c7`.

`should evaluate the same element separately per pseudoSizeThreshold` evaluates one element twice: a pseudo element covering 22% exceeds a `0.2` threshold and reports `pseudoContent`, and the same element under `0.3` does not.

Checked that it actually catches the regression it describes rather than just passing — setting `{ length: 1 }` on the memoize call, which is what an arity mistake would look like, fails it with `expected 'pseudoContent' to not equal 'pseudoContent'`.

Worth noting the concern was well aimed: this exact class of bug has bitten this codebase three times now, most recently `getScroll`, where a default parameter silently dropped `buffer` out of the key.

straker:
Additional testing. Using my benchmark pages the function never takes longer than 300ms. On my largest page it's called 22,126 times and takes 247.70ms total across then entire run. So again we can close this as not needing to do. Memoizing the child functions seems to be doing enough.

Issue #5327, opened 2026-08-25
Cache common element computations on vNode


WilcoFiers:
There are a few node computations Axe-core does over and over again. Things like reading the accessible name, or the role. We should look at caching these on the virtual node to see if it helps with performance:

```js
class AbstractVirtualNode {
  // Pseudo code
  get role                  => memoize(this, getRole(this))
  get roleType              => memoize(this, getRoleType(this))
  get implicitRole          => memoize(this, getImplicitRole(this))
  get explicitRole          => memoize(this, getExplicitRole(this))
  get accessibleText        => memoize(this, accessibleTextVirtual(this))
  vNode.aria                => memoize([this, arg[0]], getAriaValue(this, 'aria-' + arg[0]).value)
  vNode.ariaSource          => memoize([this, arg[0]], getAriaValue(this, 'aria-' + arg[0]).source)
  vNode.ref                 => memoize([this, arg[0]], getResolvedRef(this, arg[0]))
}
```

straker:
@chutchins25 opened a pr to test the role portion of this ticket and found caching them didn't result in any meaningful time savings:

| Round | develop | this branch |
|---|---|---|
| 1 | 2228 ms | 2278 ms |
| 2 | 2269 ms | 2260 ms |
| 3 | 2254 ms | 2278 ms |


I did some further testing. In all my benchmarks sites, not a single one of these 3 functions (`getRole`, `getExplicitRole`, and `implicitRole`) ever takes more than 300ms. On my largest benchmark page, `getRole` is called 154,246 times and takes a total of 289.30ms across the entire run, `getExplicitRole` is called 402,454 times and takes a total of 127.20ms, and `implicitRole` is called 153,813 times and takes 100.00ms. 

So for the role functions at least I don't think we need to worry about caching them.

chutchins25:
Took a stab at worlds where the modified other role functions exist in case you wanted to run metrics on them anyhow. I have each function as separate PRs so we can pick and choose which to keep and which to close and ignore.

- [x] getRole
- [x] getRoleType
- [x] getImplicitRole
- [x] getExplicitRole
- [x] accessibleTextVirtual
- [x] vNode.aria
- [x] vNode.ariaSource
- [x] vNode.ref

chutchins25:
Progress on the rest of the codeblock. Two of the remaining items measure as dead ends, so I want to put the numbers up before spending more time on them.

**`explicitRole` — implemented, measured negative, not opening a PR.**

[...]

PR #5329, opened 2026-08-25
perf(aria): cache getRole per node

chutchins25:
Refs #5327 — the `getRole` part of it. `accessibleText` and the `aria`/`ariaSource`/`ref` accessors are deliberately not included; reasons at the bottom.

## What

`getRole` caches its result on the virtual node's `_cache`. The cache lives inside `getRole` rather than as a `vNode.role` getter, so all 33 call sites benefit without any of them changing.

Three details that are load-bearing:

**The options are part of the key.** The pseudo-code in the issue is `get role => memoize(this, getRole(this))`, but `getRole` takes six booleans that change the answer, and instrumenting a run shows six distinct shapes actually in use:

| Key | Calls (ARIA-heavy page) |
|---|---|
| no options | 82,518 |
| `{ dpub, fallback }` | 15,000 |
| `{ fallback }` | 9,000 |
| `{ noPresentational }` | 4,500 |
| `{ noImplicit }` | 3,000 |
| `{ noPresentational, chromium }` | 1,500 |

A single cached value would return the wrong role for five of those.

**The entry is written after the role resolves, not before.** That keeps a throw from `getInheritedRole` out of the cache, and it means role resolution can still re-enter for a node while that node is being resolved, exactly as it does today.

**The virtual node is resolved once per call.** `getRole` needs it to reach `_cache`, and `resolveRole` used to look it up again, so a cache miss paid for two lookups where it previously paid for one. The resolved node is now passed down (second commit).

## Results

**The cache hits often and saves nothing measurable.**

| | ARIA-heavy | 12,000 elements |
|---|---|---|
| `getRole` calls | 115,518 | 60,626 |
| Cache hits | 73,514 (**63.6%**) | 42,740 (**70.5%**) |

Three alternating rounds of ten runs on the ARIA-heavy page, against a `develop` build that includes #5325:

|---|---|---|
| 1 | 2228 ms | 2278 ms |
| 2 | 2269 ms | 2260 ms |
| 3 | 2254 ms | 2278 ms |

12,000 element page: 2182 ms vs 2165 ms. The sign flips between rounds, so it is inside the run-to-run noise here.

So the hit rate is high — far higher than the 13% measured for the ARIA value accessors in #5300 — and the wall clock does not move. The conclusion I draw is that `getRole` is simply cheap: removing 70% of the calls is not worth anything because the calls were not costing anything. Results are identical throughout and the full suite passes.

Worth having in front of you as code rather than as a claim, since my measurements have been the less reliable half of this epic. If your benchmark sites disagree, the branch is here to run.

## Not included

**`accessibleText`.** `accessibleTextVirtual(virtualNode, context)` takes a context that carries `alreadyProcessed`, which is the guard against the role/accessible-name re-entrancy cycle. Caching on the node alone would ignore it, and getting that wrong is a correctness bug rather than a neutral result. Worth deciding deliberately.

**`aria`, `ariaSource` and `ref`.** These are `getAriaValue` and `getResolvedRefs`, which #5300 closed as not worth caching, with your own benchmark data showing 252,143 calls costing 120ms.

**`roleType`** is worth a note if this idea is taken further: 10 of its 11 call sites pass a role *string* rather than a node (`getRoleType(role)`), and only `is-in-text-block.js:84` passes a node. A `vNode.roleType` getter would therefore only serve that one call site.

## Testing

### DIFF
--- a/lib/commons/aria/get-role.js
+++ b/lib/commons/aria/get-role.js
@@ -117,14 +117,13 @@ export function hasConflictResolution(vNode) {
 /**
  *
  * @method resolveRole
- * @param {Element|VirtualNode} node
+ * @param {VirtualNode} vNode
  * @param {Object} options
  * @see getRole for option details
  * @returns {string|null} Role or null
  * @deprecated noImplicit option is deprecated. Use aria.getExplicitRole instead.
  */
-function resolveRole(node, { noImplicit, ...roleOptions } = {}) {
-  const { vNode } = nodeLookup(node);
+function resolveRole(vNode, { noImplicit, ...roleOptions } = {}) {
   if (vNode.props.nodeType !== 1) {
     return null;
   }
@@ -184,7 +183,7 @@ function getRole(node, options = {}) {
   // still re-enter for the same node while it is being computed.
   const nodeCache = vNode?._cache;
   if (!nodeCache) {
-    return resolveRoleWithOptions(node, options);
+    return resolveRoleWithOptions(vNode, options);
   }
 
   const { noImplicit, fallback, abstracts, dpub, noPresentational, chromium } =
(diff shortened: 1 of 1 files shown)


straker:
Additional testing. In all my benchmarks sites, not a single one of these 3 functions (`getRole`, `getExplicitRole`, and `implicitRole`) eve takes more than 300ms. On my largest benchmark page, `getRole` is called 154,246 times and takes a total of 289.30ms across the entire run, `getExplicitRole` is called 402,454 times and takes a total of 127.20ms, and `implicitRole` is called 153,813 times and takes 100.00ms. 

So with @chutchins25's testing and mine I think we are safe to not worry about caching these functions.

Issue #5293, opened 2026-08-06
[EPIC] Improve axe-core performance


straker:
Improve the run time of axe-core to make it better, faster, stronger. We have a rough estimate that doing these changes will result in a reduction of about 60% run time overall.

straker:
Work completed

PR #5330, opened 2026-08-27
perf(aria): memoize getRole

chutchins25:
Second of the items in Wilco's codeblock on #5327.

Reworked after @straker's review: this branch previously memoized `getRoleType`, and now memoizes `getRole` instead, per his finding that the cost sits in the `getRole` calls inside `getRoleType` rather than in the type lookup.

### What this does

`getRole` resolves a node's role from scratch on every call, and it is called tens of thousands of times in a run across 45 call sites. This memoizes it on the virtual node.

The wrinkle is the options. All six of `noImplicit`, `fallback`, `abstracts`, `dpub`, `noPresentational` and `chromium` change the answer, so all six belong in the cache key — but memoizee's primitive mode builds its key from `String(argument)`, and every options object stringifies to the same `[object Object]`. Memoizing the function as-is would merge six distinct answers into one. So the options are flattened into a six-character primitive first, and the virtual node is keyed on directly (its `toString()` is a unique `_uid`). A node outside the tree has no virtual node to key on and resolves uncached.

The key is built by concatenation rather than `map().join()`: it runs on every `getRole` call, including the many that never repeat, so it must not allocate.

### Relationship to #5329

#5329 cached `getRole` per node and was closed unmerged after measuring no wall-clock change, despite a 63–70% hit rate. That conclusion was wrong, and the reason is worth recording: it was benchmarked on ARIA-heavy and large-DOM pages, which is where the repeats *aren't*. Re-measured on a fixture that actually exhibits the repeat pattern, the same idea is worth about 4%.

### Numbers

Both builds from this branch's merge base (`c54d1aa2`), checksums confirmed different, identical results on every fixture (violation, pass, incomplete and related-node counts all match).

Four variants, benchmarked interleaved in a single session on 60 paragraphs × 20 links — the fixture where `isInTextBlock` walks every sibling once per link:

| variant | medians (4 rounds) | mean | Δ |
| --- | --- | --- | --- |
| base | 390.1 / 397.6 / 389.0 / 391.5 | 392.1 ms | — |
| both caches | 369.3 / 371.3 / 374.9 / 374.4 | 372.5 ms | −5.0% |
| memoize `getRoleType` | 376.7 / 376.3 / 372.1 / 366.3 | 372.9 ms | −4.9% |
| **memoize `getRole` (this branch)** | 377.3 / 382.7 / 372.6 / 372.9 | **376.4 ms** | **−4.0%** |

Worth being straight about the tradeoff: memoizing `getRoleType` measured slightly better than memoizing `getRole`, and doing both added nothing over `getRoleType` alone (0.4 ms apart, ranges fully overlapping). `getRole` is the more general lever — 45 call sites rather than 6 — which is why this branch takes that route, but on this fixture it is the smaller of the two wins.

Pages without the repeat pattern are unaffected. Three rounds each, rounds crossing in both directions:

| fixture | base | this branch |
| --- | --- | --- |
| aria-heavy | mean 1510.9 ms | mean 1520.0 ms |
| large DOM | mean 1337.7 ms | mean 1328.1 ms |
| dense small targets | mean 134.1 ms | mean 132.8 ms |

So the per-call cost of building the key and going through memoizee does not show up where the cache never hits.

### Tests

One test: the same node queried with and without `dpub` must not share a cache entry. It fails against a naive `memoize(getRole, { primitive: true })` and passes here, so it isn't vacuous. The existing `dpub` tests can't catch it — they use a fresh node per test, and the collision needs one node queried twice.

Refs issue #5327

### DIFF
--- a/lib/commons/aria/get-role-type.js
+++ b/lib/commons/aria/get-role-type.js
@@ -1,5 +1,6 @@
 import standards from '../../standards';
 import AbstractVirtualNode from '../../core/base/virtual-node/abstract-virtual-node';
+import { nodeLookup } from '../../core/utils';
 
 /**
  * Get the "type" of role; either widget, composite, abstract, landmark or `null`
@@ -10,12 +11,30 @@ import AbstractVirtualNode from '../../core/base/virtual-node/abstract-virtual-n
  * @return {Mixed} String if a matching role and its type are found, otherwise `null`
  */
 function getRoleType(role) {
-  if (
+  if (!(
     role instanceof AbstractVirtualNode ||
     (window?.Node && role instanceof window.Node)
-  ) {
-    role = axe.commons.aria.getRole(role);
+  )) {
+    return roleTypeOf(role);
   }
+
+  // resolving the role of a node is the expensive half of this, and the same
+  // node is asked for its role type repeatedly during a run, so cache it on
+  // the node. Written after resolving, so a throw is not cached.
+  const { vNode } = nodeLookup(role);
+  const nodeCache = vNode?._cache;
+  if (nodeCache && 'getRoleType' in nodeCache) {
+    return nodeCache.getRoleType;
+  }
+
+  const roleType = roleTypeOf(axe.commons.aria.getRole(role));
+  if (nodeCache) {
+    nodeCache.getRoleType = roleType;
+  }
+  return roleType;
+}
+
+function roleTypeOf(role) {
   const roleDef = standards.ariaRoles[role];
   return roleDef?.type || null;
 }
--- a/test/commons/aria/get-role-type.js
+++ b/test/commons/aria/get-role-type.js
@@ -46,4 +46,25 @@ describe('aria.getRoleType', () => {
     ).actualNode;
     assert.equal(getRoleType(domNode), 'stuff');
   });
+
+  it('resolves the role of a node only once', () => {
+    const vNode = queryFixture('<span id="target" role="cats"></span>');
+    assert.equal(getRoleType(vNode), 'stuff');
+
+    // a second lookup must not re-resolve the role
+    vNode.actualNode.setAttribute('role', 'dogs');
+    assert.equal(getRoleType(vNode), 'stuff');
+  });
+
+  it('caches the type on the virtual node, not the DOM node', () => {
+    const vNode = queryFixture('<span id="target" role="cats"></span>');
+    assert.equal(getRoleType(vNode.actualNode), 'stuff');
+    assert.equal(vNode._cache.getRoleType, 'stuff');
+  });
+
+  it('does not cache the type of a role passed as a string', () => {
+    const vNode = queryFixture('<span id="target" role="cats"></span>');
+    assert.equal(getRoleType('cats'), 'stuff');
+    assert.isFalse('getRoleType' in vNode._cache);
+  });
 });


straker (dismissed):
Testing on my benchmark sites shows that for the most part this function isn't a problem, but certain types of pages do hit this a lot and it does create a noticeable problem. For example, https://web.archive.org/web/20190613132353/https://giveawaylisting.com/ shows that with `target-size` enabled, this function is hit 74,1711 times and takes 1,983.30ms of time. The entire run takes 17,931.9ms, so 2s out of that is significant. So caching this would be worth it for those cases. Testing shows that on the same page, caching would result in a function time of 162.7ms and a total run time of 16,396.3ms.

However, instead of a custom cache we should just memoize the function. The `role` is already a string so we can use `{ primitive: true }`. Probably don't need any tests for that (I don't think we've added tests for any of the other memoized functions).

chutchins25:
Reworked to use `memoize` — thanks, that's cleaner and it keeps the win.

One thing I hit applying it: `getRoleType` takes a node as well as a role string, and six of the eleven call sites pass one — including `target-size-evaluate`, the path in your giveawaylisting.com numbers. memoizee's primitive mode keys on `String(argument)`, so `memoize(getRoleType, { primitive: true })` on its own collapses every same-tag element onto one entry:

```
no toString:     button button -> calls: 1  COLLISION
unique toString: button link   -> calls: 2  ok
```

So the node path resolves through `nodeLookup` first and the memoized fn only ever sees a virtual node, whose `toString()` is a unique `_uid` — the same shape `isVisibleOnScreen` and `toGrid` use.

I kept one test for that case: it's the only one in the file that fails against the un-normalized version. Dropped the three `_cache` tests as you suggested.

Re-benchmarked after the rework (old numbers were against the custom cache): −3.2% on the link-heavy fixture, within noise on the three 0%-hit fixtures.

lib/commons/aria/get-role-type.js
```diff
@@ -14,8 +16,28 @@ function getRoleType(role) {
     role instanceof AbstractVirtualNode ||
     (window?.Node && role instanceof window.Node)
   ) {
-    role = axe.commons.aria.getRole(role);
+    // Resolving the role of a node is the expensive half of this, and the same
+    // node is asked for its role type repeatedly during a run. Memoize on the
+    // virtual node rather than on `role` itself: memoizee's primitive mode
+    // keys off `String(argument)`, and only a virtual node has a unique string
+    // form — every span element stringifies to `[object HTMLSpanElement]`.
```
straker, line 23: When I said we should memoize I didn't realize that `role` could be a node or vNode. This makes me wonder if the actual expensive part of this function is the role lookup rather than the type lookup. I'll need to do some testing to figure that out. It may be that we should memoize `getRole` rather than this function.

straker: Confirmed that it's actually `getRole` calls inside `getRoleType` that are causing the issue. So we don't need to cache `getRoleType` but actually `getRole`.

chutchins25:
Switched it over to `getRole` — you were right that that's where the cost is, and right to be suspicious of where I'd put the cache.

Two things worth reporting back.

**#5329 was closed on a bad benchmark.** It cached `getRole` per node, measured a 63–70% hit rate and no wall-clock change, so I dropped it. But I'd benchmarked it on ARIA-heavy and large-DOM pages, which is exactly where the repeats aren't. Re-run on a fixture with the `isInTextBlock` pattern, the same idea is worth about 4%. So your instinct was better than my measurement.

**The options have to be flattened into the cache key.** All six of them change the answer, and memoizee's primitive mode keys on `String(argument)`, so every options object collapses onto `[object Object]`. A plain `memoize(getRole, { primitive: true })` returns the no-options answer for every option shape — `getRole(node)` then `getRole(node, { dpub: true })` gives `null` twice. There's a test for that case now; it's the only one in the file that catches it, since the existing `dpub` tests use a fresh node each.

One thing you should see before this goes further. I measured four variants interleaved in one session on the link-heavy fixture:

| variant | mean | Δ |
| --- | --- | --- |
| base | 392.1 ms | — |
| both caches | 372.5 ms | −5.0% |
| memoize `getRoleType` | 372.9 ms | −4.9% |
| memoize `getRole` | 376.4 ms | −4.0% |

Memoizing `getRoleType` edged out memoizing `getRole` here, and doing both added nothing over `getRoleType` alone. I've gone with `getRole` since it's the more general lever — 45 call sites rather than 6 — but if you'd rather have the larger number on this shape of page, the other branch is a revert away. Your call.

Fixture is 60 paragraphs × 20 links each; happy to share it if you want to run it against your sites.

lib/commons/aria/get-role.js
```diff
@@ -175,8 +175,65 @@ function resolveRole(node, { noImplicit, ...roleOptions } = {}) {
...
-function getRole(node, { noPresentational, ...options } = {}) {
-  const role = resolveRole(node, options);
+function getRole(node, options = {}) {
+  const { vNode } = nodeLookup(node);
+
+  // The role of a node is resolved many times over during a run, so memoize
+  // it. A node outside the tree has no virtual node to key on, so it resolves
+  // uncached.
+  if (!vNode) {
+    return resolveRoleWithOptions(vNode, options);
```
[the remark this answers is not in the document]
chutchins25: Partly valid — the comment was wrong, the behaviour is not.

Confirmed the branch throws on that path. It also throws on `develop`, identically:

```
develop  → TypeError: Cannot read properties of undefined (reading 'props')
this PR  → TypeError: Cannot read properties of undefined (reading 'props')
```

(detached element, with and without options.) `resolveRole` dereferenced `vNode.props` after its own `nodeLookup` before this change too, so the out-of-tree path is unchanged.

The defect was the comment claiming it "resolves uncached". Corrected to say what actually happens, and added a test pinning the throw — it passes against `develop`'s `get-role.js` as well, which is the point: it characterises existing behaviour rather than new behaviour.

Not taking the "preserve the original element for an uncached resolution path" half of the suggestion: making out-of-tree elements resolve where they currently throw is a behaviour change, and not one a perf PR should carry.

Fixed in b8698e8f.

lib/commons/aria/get-role-type.js
chutchins25: Done — switched the cache from `getRoleType` to `getRole` in 8c4f6eb2, per your confirmation that the cost is in the `getRole` calls. This file is out of the diff now.

The four-way benchmark is in the PR comment above. Short version: your read was right, and #5329 was closed on a benchmark that never hit the repeat pattern.

straker:
Alright. This took me way too long to figure out. The performance numbers I was seeing while testing on the site were just not making sense, where caching things resulted in 0 performance gains. I finally figured out that the site loads a script that overrides Function prototypes (`wombat.js`) which was drastically slowing down the page during the run. Blocking that script resulted in the numbers making sense again. So here's the results.

Looking at the page as it stands with `target-size` enabled, here is the call time for `getRole`, `getRoleType`, `getImplicitRole`, and `getExplicitRole`:

getExplicitRoleTotal: 1,470.6
getRoleTotal: 3,013.6
getRoleTypeTotal: 2,552.7
implicitRoleTotal: 184.1

The results are still that `getRole` dominates the time for `getRoleType`, but we can also see that `getExplicitRole` takes half that time. Testing a cache on `getExplicitRole` can reduce that number in half:

getExplicitRoleTotal: 671.5
getRoleTotal: 2,200.3
getRoleTypeTotal: 1,905.7
implicitRoleTotal: 182.0

The main problem is that trying to cache our `getRole` functions is difficult as we allow many options to be passed that changes the answer to what an elements role is. For example, just to cache `getExplicitRole` required creating a new function that caches processing the explicit roles and then uses the options to filter that to an answer.

```js
export default function getExplicitRole(node, { fallback, abstracts, dpub } = {}) {
  const { vNode } = nodeLookup(node);
  if (vNode.props.nodeType !== 1) {
    return null;
  }
  const roleMap = buildRoleMap(vNode);
  
  // if fallback is not supported the entire role is treated as a single value
  // so multiple roles would never be valid
  if (!fallback && roleMap.length > 1) {
    return null;
  }

  const role = roleMap.find(role => {
    return (
      abstracts && role.type === 'abstract' ||
      dpub && role.type === 'dpub' ||
      !['abstract', 'dpub', null].includes(role.type)
    );
  });

  return role ? role.value : null;
}

const buildRoleMap = memoize(function buildRoleMapMemozied(vNode) {
  const roleAttr = (vNode.attr('role') || '').trim().toLowerCase();
  return tokenList(roleAttr).map(role => {
    const roleDefinition = standards.ariaRoles[role];
    if (roleDefinition) {
      return {
        value: role,
        type: role.substr(0, 4) === 'doc-' ? 'dpub' : roleDefinition.type
      };
    }

    return {
      value: role,
      type: null
    };
  });
}, { primitive: true });
```

Doing this for `getRole` won't be nearly as easy though since it also allows passing options through to `getExplicitRole`. 

Regardless, I think this work should fall under #5389 instead of #5293 as it's `target-size` that is causing this problem and not the current rules.

Issue #5301, opened 2026-08-07
Track axe-core performance over time


straker:
We should start tracking the run time performance of axe-core over time, that way we know ahead of time when [the performance degrades](https://github.com/dequelabs/axe-core/issues/5298) and could get notified about it. The simplest way to do that would be to have a script run axe-core against a benchmark page a few times, take the average time, then append the result and date to a file in axe-core.

In terms of running it, we wouldn't want to run it for every pr, but we would want to track performance regularly and be able to tie changes to performance to commits so we know what changed and when.

straker:
Completed via perf reports for [4.11.0](https://github.com/dequelabs/axe-core/pull/5378), [4.12.0](https://github.com/dequelabs/axe-core/pull/5377), [4.13.0](https://github.com/dequelabs/axe-core/pull/5376), and [latest develop](https://github.com/dequelabs/axe-core/pull/5379).

padmavemulapati:
No QA Required

PR #5333, opened 2026-08-27
perf(get-selector): greatly improve get-selector performance

### LINKED ISSUE #4008: Analysis never finishes
### Product

axe-core

### Product Version

4.7.0

### Latest Version

#### Expectation
Describe what you expected the product to do.

#### Actual
Describe what the product actually does.

#### How to Reproduce
On some pages like: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html-single/package_manifest/index#revision_history

The analysis never finishes. I am injecting axe in this way:

```
const script = document.createElement("script");
script.src = "https://cdnjs.cloudflare.com/ajax/libs/axe-core/4.7.0/axe.min.js";
document.head.appendChild(script);
```

And then I just run `axe.run().then(console.log)` which it never ends.

Is there a way to get axe-core to abort once a certain time has lapsed? This would help with error handling if axe can't finish or takes too long. 

Replication: https://codesandbox.io/p/sandbox/aged-field-v7c7mh?file=%2Frun-puppeteer.js

### LINKED ISSUE #4273: Super-linear runtime degradation with respect to number of DOM nodes
### Product

axe-core

### Product Version

4.8.2

### Latest Version

#### Expectation

I would expect runtime performance to correspond at most linearly to the number of DOM nodes present.

#### Actual

Performance appears to degrade more than linearly with the number of DOM nodes in play. That is: if I double the number of DOM nodes, performance gets ~8x worse, whereas I would expect it to at most get 2x slower.

#### How to Reproduce

https://github.com/nwalters512/axe-core-performance-repro

Run `yarn` to install dependencies, then run `node index.js`. With `OPTION_COUNT = 500`, Axe runs in ~11 seconds. When `OPTION_COUNT` is doubled to 1000, runtime increases to 90 seconds

#### Additional context

Running this code with the Node performance profiler, I was able to identify that the vast majority of time is spent in [`processAggregate`](https://github.com/dequelabs/axe-core/blob/af0b548a61d7156f5e71194f8f2172ce6d5521fa/lib/core/reporters/helpers/process-aggregate.js#L29), and specifically in [`trimElementSpec`](https://github.com/dequelabs/axe-core/blob/af0b548a61d7156f5e71194f8f2172ce6d5521fa/lib/core/reporters/helpers/process-aggregate.js#L85) inside that. I wasn't able to dig significantly deeper than that, though it appears as though [`getNthChildString`](https://github.com/dequelabs/axe-core/blob/af0b548a61d7156f5e71194f8f2172ce6d5521fa/lib/core/utils/get-selector.js#L351) is the slowest thing that runs as part of `trimElementSpec`. Here's some more detail from the profiler:

Screenshot 2023-12-12 at 16 36 24

### LINKED ISSUE #4427: `axe.runPartial` performance on page with 100k nodes
We recently ran into an issue where a page had 108,000 `option` nodes in a single `select` element. Axe-core took a bit to just process that many nodes, but the real problem was processing selectors for that many nodes. 

Typically our answer to that is [to use `resultTypes`](https://github.com/dequelabs/axe-core/blob/develop/doc/API.md#section-4-performance) to filter out most nodes which allows us to just generate selectors for things like violations. However when using `runPartial` the `resultTypes` option is not used which means the at the end of the `runPartial` all 108,000 nodes will try to generate a selector even if they will be removed after running `finishRun`.

One thing we'll have to figure out is how to deal with rules that have after methods which may change the result of a node.

### LINKED ISSUE #5299: Create a faster get-selector algorithm
The slowest part of a run in axe-core is currently generating all the unique selectors for each reported node which can easily account for at least 50% of the overall run time. We can significantly improve the run time of axe-core by changing how the selectors are generated. However, changing the final selector is a breaking change so the new algorithm will only be available as a run/configure option to axe-core. I'm not sure what to call the option or what values it should have, but an example could look somethig like:

```js
axe.run({
  generateSelectors: 'default' | 'fast' | 'both'
});
```

### LINKED ISSUE #5332: Remove memoize from `getSelector`
The comment says it was added as the same node could be used by multiple DqElements, but we already "memoize" DqElement so that wouldn't be true. Additional as we've found out any uses of memoize on DOM nodes is probably super slow on very large DOM pages as it has to traverse all the calls to know if it's a cache miss.

straker:
## Why

Selector uniqueness is the hot path in `axe.run` reporting. Before this change, the algorithm called `doc.querySelectorAll` for every chain link it evaluated and `matchesSelector` for every sibling in the nth-child check — on large pages this dominated the run at ~85s.

## What

Keeps feature-selection (`getElmId`, `getThreeLeastCommonFeatures`, `getNthChildString`) identical so selector output does not change. Replaces the CSS-engine-backed uniqueness check with an in-memory match against per-node bitset indexes built once by `getSelectorData`:

- tag / class / attr bitsets keyed by their `getBaseSelector`-form
- per-root scope bitsets (shadow DOM elements are scoped by intersecting with `_rootBits`, not a separate code path)
- per-root id counts (skips the `doc.querySelectorAll('#id')` call)
- flat `parentOf` Int32Array for O(1) parent lookup

Level-0 match: intersect the bitsets, filter by nth-child + suffix-attr, enumerate.
Level-1+: advance each candidate's ancestor pointer through `parentOf` and single-bit-test it against the next fragment's precomputed bitset.

Additional caches: `computeFragment` per element (nodeIdx-keyed plain array), `bitsForFragment` / `indicesForFragment` per (rootId, fragment string) sharing one `_matchCache` entry, and a per-root chain-suffix cache so targets sharing an intermediate chain reuse each other's filtered candidate set. That last cache is what unlocks the win on repetitive pages like the Large text page.

Suffix-form attribute selectors (`[href$="X"]` / `[src$="X"]`) use CSS suffix semantics that atnv-exact matching can't express, so raw href / src values are stored per node and `endsWith`-filtered post-intersection. Without this, level-0 candidate counts under-count against `querySelectorAll` and produce shorter-than-develop selectors.

Also drops the `memoize(getSelector)` wrapper — see [Why the memoize wrapper was removed](#why-the-memoize-wrapper-was-removed) at the bottom for the reasoning and measurements.

### Not-in-tree callers

`generateSelector` runs only on nodes that were in the tree at `getSelectorData` time. The DOM is stable across a single `axe.run` and no rule constructs synthetic virtual nodes that end up in results, so during a normal run every caller lands on the in-memory path. A caller that hands `getSelector` a node the tree never walked (the "detached DocumentFragment" unit test does this) falls through to a small safety net: we verify id uniqueness with a targeted `doc.querySelectorAll(id)` and otherwise return `getBaseSelector(elm)`. Unit tests exercise this path and its edge cases (id shared with a tree element, post-run shadow root attach, detached DocumentFragment).

## Verification (selectors byte-identical to develop)

All 57 unit tests in `test/core/utils/get-selector.js` pass on both this PR and on develop — 51 pre-existing, plus 6 I/O regressions added here (four covering the reviewer's shadow-DOM drop-site rounds, one for prototype pollution, one for `$` in attr name). None coupled to internal shape.

Selector output diffed against develop on 8 pages, snapshotted locally with a `script-src 'none'` CSP so both algorithms see identical DOM state — **0 raw diff lines on every page**:

| Page | Rows | Raw diff |
|---|---:|---:|
| [Stack Overflow](https://stackoverflow.com/questions/11832914/how-to-round-to-at-most-2-decimal-places-if-necessary)* | 561 | 0 |
| [webtestingcourse](https://webtestingcourse.dequecloud.com/) | 2,433 | 0 |
| [broken-workshop](https://broken-workshop.dequelabs.com/) | 4,993 | 0 |
| [DQ Dream Destination](https://dequeuniversity.com/demo/dream) | 9,769 | 0 |
| [DQ Mars Commuter](https://dequeuniversity.com/demo/mars/) | 14,433 | 0 |
| [MDN ElementInternals](https://developer.mozilla.org/en-US/docs/Web/API/ElementInternals) | 23,945 | 0 |
| [TensorFlow.js API](https://js.tensorflow.org/api/latest/) | 617,249 | 0 |
| Large text page (>152,366 text nodes) | 2,480,657 | 0 |

*The Stack Overflow snapshot is the Cloudflare "Performing security verification" page — headless bypass didn't work. The row confirms both algorithms agree on that DOM, not on real Stack Overflow content.

## Perf

Measured on 8 pages, sorted by page size. All numbers are wall-clock median across N samples (headless Chrome, warmed-up JIT, single Selenium session). N varies from 20 on small pages to 3 on the Large text page (each fast sample there takes ~22s). p95 alongside p50 to catch tail-variance regressions; `cold` = the first sample after page load / axe injection so any first-run penalty from unwarmed JIT or unprimed caches shows up.

### Full `axe.run` time (end-to-end)

| Page | Nodes | Develop p50 | This PR p50 | Speedup | Develop p95 | This PR p95 |
|---|---:|---:|---:|---:|---:|---:|
| [Stack Overflow (CF page)](https://stackoverflow.com/questions/11832914/how-to-round-to-at-most-2-decimal-places-if-necessary) | 48 | 17ms | 22ms | 0.8x | 22ms | 25ms |
| [webtestingcourse](https://webtestingcourse.dequecloud.com/) | 141 | 37ms | 38ms | 1.0x | 42ms | 41ms |
| [broken-workshop](https://broken-workshop.dequelabs.com/) | 268 | 59ms | 58ms | 1.0x | 63ms | 63ms |
| [DQ Mars Commuter](https://dequeuniversity.com/demo/mars/) | 1,141 | 118ms | 106ms | 1.1x | 128ms | 112ms |
| [DQ Dream Destination](https://dequeuniversity.com/demo/dream) | 2,087 | 519ms | 506ms | 1.0x | 527ms | 512ms |
| [MDN ElementInternals](https://developer.mozilla.org/en-US/docs/Web/API/ElementInternals) | 10,323 | 1.79s | 839ms | 2.1x | 1.82s | 881ms |
| [TensorFlow.js API](https://js.tensorflow.org/api/latest/) | 35,133 | 13.5s | 5.56s | 2.4x | 13.6s | 5.66s |
| Large text page | 152,366 | 170.5s | 23.4s | 7.3x | 170.5s | 23.4s |

Small pages (< 3k nodes) don't shift much — selector generation is a small fraction of `axe.run` on those pages, so replacing it doesn't move the total by much. The 2–7× wins land on pages large enough (10k+ nodes) for develop's CSS-engine-backed selector loop to dominate. On the Large text page the full run drops from 170s to 23s; the remaining time is other rule work, unaffected by this PR.

### Isolated selector-generation time (`reporter` measure)

`axe.run`'s `reporter` phase on develop is dominated by selector generation, so this isolates the algorithm change:

| Page | Develop | This PR | Speedup |
|---|---:|---:|---:|
| Stack Overflow (CF page) | 1ms | 1ms | 1.0x |
| webtestingcourse | 2ms | 1ms | 2.0x |
| broken-workshop | 6ms | 2ms | 3.0x |
| DQ Mars Commuter | 16ms | 4ms | 4.0x |
| DQ Dream Destination | 18ms | 3ms | 6.0x |
| MDN ElementInternals | 1.04s | 52ms | 20.0x |
| TensorFlow.js API | 8.25s | 299ms | 27.6x |
| Large text page | 149.9s | 1.90s | 79.1x |

### `axe.setup` time + cold-start

| Page | Setup dev | Setup PR | Cold `axe.run` dev | Cold `axe.run` PR |
|---|---:|---:|---:|---:|
| Stack Overflow (CF page) | 1ms | 2ms | 28ms | 28ms |
| webtestingcourse | 3ms | 3ms | 44ms | 44ms |
| broken-workshop | 5ms | 5ms | 327ms | 191ms |
| DQ Mars Commuter | 16ms | 17ms | 143ms | 134ms |
| DQ Dream Destination | 27ms | 30ms | 536ms | 514ms |
| MDN ElementInternals | 82ms | 91ms | 1.90s | 970ms |
| TensorFlow.js API | 196ms | 213ms | 13.7s | 5.88s |
| Large text page | 748ms | 890ms | 167.2s | 23.3s |

`axe.setup` builds the precomputed model. Growth versus develop is 10–20% on medium/large pages (the extra work is building the bitmap indexes plus one `querySelectorAll('*')` per root for the phantom-seed pass) but absolute cost stays under 1s even on 150k nodes. Cold-start tracks p50 across all pages — no unusual first-run penalty from the new indexes.

### `_selectorData` memory footprint (post-run)

| Page | Develop | This PR | Δ |
|---|---:|---:|---:|
| Stack Overflow (CF page) | 2.1 KB | 24.2 KB | +22.1 KB |
| webtestingcourse | 2.7 KB | 71.2 KB | +68.5 KB |
| broken-workshop | 4.0 KB | 160.5 KB | +156.5 KB |
| DQ Mars Commuter | 25.7 KB | 763.2 KB | +737.5 KB |
| DQ Dream Destination | 65.3 KB | 1.4 MB | +1.4 MB |
| MDN ElementInternals | 75.9 KB | 9.8 MB | +9.7 MB |
| TensorFlow.js API | 71.5 KB | 38.6 MB | +38.5 MB |
| Large text page | 1.7 KB | 127.7 MB | +127.7 MB |

Sized after `axe.run` completes, captured through a `defineProperty` hook so teardown doesn't zero it before the sample. Growth is the price of the perf: bitmap indexes, `_perNode` records, and query-time caches all live here. The delta on the Large text page looks disproportionate because develop's `_selectorData` on that snapshot happens to be unusually tiny (0 unique classes, 19 tags, 48 attrs); on tensorflow the ratio is ~540×, on MDN ~130× — those are more representative.

### Drift over 50 back-to-back runs (MDN page)

`axe.run` called 50 times in the same tab without page reload, to catch any leak that would show as monotonic upward drift. Both algorithms show a slightly negative slope (times drift down as JIT settles) — no leak signal on either side.

|  | Develop | This PR |
|---|---:|---:|
| First-10 mean | 1,906ms | 871ms |
| Last-10 mean | 1,891ms | 865ms |
| Slope | −0.53ms/iter | +0.04ms/iter |
| Min / Max | 1,843ms / 2,020ms | 841ms / 975ms |

#### Memory tradeoff — impact on consumers

The table above shows peak `_selectorData` at the top of the range (128 MB on the Large text page). What this actually means for someone running axe-core:

**Transient, not resident.** `axe.run`'s internal teardown frees `_selectorData` at the end of every run. This is a spike-then-drop on the memory graph, not a floor that persists. An app calling `axe.run` in a loop doesn't accumulate — steady-state cost is zero. Peak is what matters for OOM, and only during the run itself.

**Distribution across real pages.**

| Page shape | Typical `_selectorData` peak |
|---|---:|
| Small (< 1k nodes) — most in-page a11y checks | < 25 KB |
| Medium (1–10k nodes) — typical SPA screen | 100 KB – 10 MB |
| Large (10–50k nodes) — dense docs, dashboards | 10–40 MB |
| Extreme (100k+ nodes) — Large text page, generated content | 100+ MB |

The vast majority of `axe.run` invocations in practice land in Small or Medium — <10 MB peak. Extreme is the case this PR was built for and where the transient cost is highest. In every band the cost scales with the DOM cost the consumer is already paying (a 152k-node DOM tree uses several hundred MB of Chrome heap on its own).

**Where it could bite.**
- Low-memory mobile devices auditing extreme-scale pages (rare intersection in practice).
- Highly parallel CI — e.g. 20 workers × 100k-node pages ≈ 2.5 GB aggregate transient. Fine on 16 GB CI runners; could OOM inside a memory-capped Docker container that's tight on develop already.
- In-page RUM / live user-session profiling on very large documents.

Develop already has a comparable cost profile on the extreme cases — it just holds the state inside the CSS engine (opaque to `performance.memory`) rather than in a JS object we can size directly, so it doesn't show up in a memory table.

If we notice that memory becomes a problem there are things we can do later to reduce it, such as capping `_chainCache` and `_matchCache` sizes, or skipping the precomputation below a node-count threshold.

### Synthetic fixtures

Two shapes the canonical 8 don't hit:

**`perf/sites/synth-small`** — 43-element hand-authored page, verifies no small-page regression from setup overhead:
- Full `axe.run`: 25ms → 26ms
- `_selectorData`: 1.3 KB → 18 KB

**`perf/sites/synth-shadow`** — 50 custom elements with open shadow roots, verifies `_rootMap` scaling and phantom-seed pass:
- Full `axe.run`: 74ms → 70ms
- `_selectorData`: 2 KB → 171 KB (`_rootMap` populated with 51 roots)

Both confirm no regression on shapes that aren't in the primary perf table.

## Why the memoize wrapper was removed

The default export of `getSelector` was memoized before this PR. #4611 added it in early 2025 to fight three problems that were catastrophic on repetitive-DOM pages under the QSA-per-fragment algorithm:

1. Same-element selector recomputation across DqElement passes.
2. Same-fragment `doc.querySelectorAll` firing over and over.
3. Per-candidate `.matches` calls when QSA returned N duplicates.

Memoize was the right fix at the time — the perf table in #4611 shows a 5–20× improvement on their 12k-button test.

This PR replaces the whole algorithm. All three of #4611's problems now have finer-grained fixes built directly into the new machinery:

| #4611 problem | How this PR handles it |
|---|---|
| Same-element selector recomputation | DqElement's `_dqElementCache` (added by #5325) already dedupes per element. `.selector` fires only 2–4 times per unique DqElement regardless of what memoize does at this layer. |
| Same-fragment queries | `_matchCache` caches `(rootId, fragment string) → {bits, indices}`. Hash-map lookup, O(1) per query. Every fragment ask after the first is a cache hit. |
| Per-candidate `.matches` | Gone. Bitmap AND replaces `.matches` per candidate — one 32-bit AND covers 32 candidates. |

Plus the new `_chainCache`: when two targets pass through the same ancestor chain suffix, the second reuses the first's filtered candidate list. A level of caching #4611's memoize never provided.

The memoize wrapper on top of all that saves nothing the algorithm doesn't already cache. It still costs, though: `memoizee` on non-primitive args (Elements) resolves cache keys with a linear scan of prior calls — the exact bottleneck #5325 solved for DqElement. On huge pages that's O(N) per lookup and dominates.

### Measured impact of removing the memoize

A/B/C run holding everything else in this PR constant, varying only the memoize wrapper:

**MDN ElementInternals** (10,323 nodes)

| Variant | Full `axe.run` p50 | `reporter` p50 |
|---|---:|---:|
| (a) memoize kept | 822ms | 61ms |
| (b) memoize removed — this PR | 801ms | 49ms |
| (c) memoize removed + DqElement caches `.selector` on instance | 821ms | 48ms |

**Large text page** (152,366 nodes)

| Variant | Full `axe.run` p50 | `reporter` p50 |
|---|---:|---:|
| (a) memoize kept | 26.8s | 6.11s |
| (b) memoize removed — this PR | 22.1s | **1.65s** |
| (c) memoize removed + DqElement caches `.selector` on instance | 22.3s | 1.35s |

Removing the memoize (variant (b)) drops the Large text page's `reporter` phase from 6.1s to 1.65s — a further **3.7× speedup on top of the algorithm change**, and 17% off the full `axe.run` time. On medium pages like MDN the effect is smaller (~20% off reporter) but still positive.

Variant (c) is a natural follow-up: cache the selector on the DqElement instance itself (parallel to the fix #5325 made for DqElement construction). Small additional win on top of (b); belongs in `dq-element.js`, not this PR.

## Test plan

- [x] `pnpm test:unit -- --files 'test/core/utils/get-selector.js'` (57/57 on this PR; 51 pre-existing + 6 new I/O regressions, all 6 also pass on develop)
- [x] Selectors byte-identical to develop on 8 varied pages
- [ ] Full CI test run
- [ ] Test on additional real-world sites of your choice 

## Notes

- Not a behavior change from an axe-user perspective. `axe._selectorData` gains new `_`-prefixed internal fields; the existing `.classes`, `.tags`, `.attributes` maps keep their exact shape.
- `constants.selectorSimilarFilterLimit` is no longer read after the legacy QSA path was removed. Left in `constants.js` untouched by this PR — cleaning it up belongs in a follow-up.
- The `perf/` scripts I wrote to iterate on this (`quick.js`, `capture.js`, `snapshot.js`, `verify-sites.sh`, `all-perf.sh`, `measure-total.js`, `measure-all.sh`, `measure-drift.js`) are not part of this PR — they live only in my working tree.

Closes: #5332
Closes: #5299
Closes: https://github.com/dequelabs/axe-core/issues/4008
Closes: https://github.com/dequelabs/axe-core/issues/4273
Closes: https://github.com/dequelabs/axe-core/issues/4427


lib/core/utils/get-selector.js
```diff
@@ -1,11 +1,12 @@
+/* eslint-disable no-bitwise -- bitset intersection and bit-testing are
+   inherent to the fast uniqueness check; disabling scoped to this file
+   keeps the algorithm readable without per-line pragmas. */
```
WilcoFiers: Can we start breaking up this file? 1200 lines is quite a lot to manage.

lib/core/utils/get-selector.js
```diff
@@ -99,66 +100,272 @@ function filterAttributes(at) {
...
+      // if and only if the element's attribute is byte-for-byte the same. We
+      // record which nodes have this exact string so we can look them up
+      // directly later.
+      matchAttrsForNode.push(atnv);
+      let list = attrLists.get(atnv);
+      if (!list) {
+        list = [];
+        attrLists.set(atnv, list);
+      }
+      list.push(nodeIdx);
```
chutchins25: **Index memory scales as `distinct attribute values x nodes / 8`, and most of the entries can never be used.**

This branch runs for *every* attribute — `filterAttributes` only gates `data.attributes` and `attrsForNode` further down. So `id`, `class`, `style`, `tabindex`, the ignored `aria-*` list, and every value 31 chars or longer each get a full-page bitset, even though none of them can ever be picked as a feature and therefore nothing can ever query their bitset.

On the Large text page, 49 of 754 attribute bitsets are usable. On a synthetic 80k-element page with a unique `id` per element, 2 of 84,098 are — 802 MB of `Uint32Array` for two usable entries.

An exact `atnv` implies the same name and the same value, hence the same `filterAttributes` verdict, so gating here can't lose a node that could have matched a fragment:

```suggestion
      // Only index attrs that can actually be picked as a feature. An exact
      // atnv implies the same name+value, hence the same filterAttributes
      // verdict, so no node that could match a fragment is lost here.
      if (filterAttributes(at)) {
        matchAttrsForNode.push(atnv);
        let list = attrLists.get(atnv);
        if (!list) {
          list = [];
          attrLists.set(atnv, list);
        }
        list.push(nodeIdx);
      }
```

Measured with that applied: 802 MB → 0.1 MB on the synthetic page, 14.1 MB → 1.3 MB on the Large text page, 6.3 MB → 3.5 MB on Stack Overflow. 0 selector diffs against develop on all 8 pages, reporter time unchanged.

One trap worth flagging: `urlAttrs` has to stay **outside** this gate. I first tried gating the whole loop body and it changed 6 MDN selectors (`li:nth-child(5) > a[href$="plus"]` collapsed to `a[href$="plus"]`). Suffix matching is many-to-one — a node whose raw `href` is too long to be a feature itself can still match *another* node's `[href$="..."]` fragment, so dropping it from `urlAttrs` under-counts and produces a shorter-than-develop selector. Only the exact-attr branch is safe to gate.

lib/core/utils/get-selector.js
```diff
@@ -99,66 +100,272 @@ function filterAttributes(at) {
...
+      let parentIdx = -1;
+      let nthChild = 0;
+      if (domParent) {
+        const pIdx = data._elmToIdx.get(domParent);
+        if (pIdx !== undefined) {
+          parentIdx = pIdx;
+        }
+        if (domParent.children) {
+          nthChild = 1 + Array.prototype.indexOf.call(domParent.children, node);
+        }
```
chutchins25: **`indexOf` per node makes this O(n²) in sibling count.**

Every element pays a linear scan of its parent's `children`, so a parent with N element children costs O(N²) across the walk. develop never paid this: it computed nth-child lazily inside `getNthChildString`, only for the elements it actually generated a selector for, and only when a sibling matched.

On a synthetic page whose `body` has 40k children (80k elements total), that's roughly 800M operations — `axe.setup()` goes from **277ms on develop to 20,039ms here**, a 72x regression. It doesn't show up in the PR's perf table because `getSelectorData` runs before the `reporter` measure starts.

Building the index once per parent instead makes it linear overall:

```suggestion
        if (domParent.children) {
          let nthMap = nthIndexCache.get(domParent);
          if (!nthMap) {
            nthMap = new Map();
            const kids = domParent.children;
            for (let k = 0; k < kids.length; k++) {
              nthMap.set(kids[k], k + 1);
            }
            nthIndexCache.set(domParent, nthMap);
          }
          nthChild = nthMap.get(node) || 0;
        }
```

That needs one more declaration alongside the other list accumulators around line 241, which the suggestion block can't reach:

```js
  const nthIndexCache = new Map();
```

Measured with both applied: `axe.setup()` 20,039ms → 384ms on the synthetic page, and 0 selector diffs against develop on all 8 pages.

chutchins25 (commented):
Nice work — this reproduces, and the win is real. A few questions for you.

I rebuilt `d50cdf6e` (develop) and `aabfba8d` (this PR) into separate bundles, re-snapshotted all 8 pages with `script-src 'none'`, and re-measured the `reporter` measure in headless Chrome.

**Timings reproduce.**

| Page | Nodes | develop | This PR | My speedup | Your table |
|---|---:|---:|---:|---:|---:|
| webtestingcourse | 142 | 1.4ms | 0.8ms | 1.8x | 1.5x |
| broken-workshop | 269 | 3.8ms | 1.2ms | 3.2x | 3.1x |
| Mars Commuter | 1,142 | 12.6ms | 4.2ms | 3.0x | 3.6x |
| Dream Destination | 2,088 | 4.9ms | 1.7ms | 2.9x | 4.3x |
| MDN ElementInternals* | 1,355 | 18.4ms | 3.1ms | 5.9x | 17.2x |
| Stack Overflow† | 11,873 | 592.9ms | 66.0ms | 9.0x | 1.1x |
| TensorFlow.js API | 35,134 | 4,317ms | 311ms | 13.9x | 14.4x |
| Large text page | 152,368 | 56,945ms | 4,641ms | **12.3x** | 13.7x |

**Selectors reproduce too: 0 diff lines on all 8 pages, across 396,448 compared selector rows** (310,133 on the Large text page alone), including `relatedNodes`.

Six of eight node counts match your table to within one node (Large text page exactly: 152,366). *My MDN snapshot came out at 1,355 nodes against your 10,323, so that row isn't comparable. †My Stack Overflow snapshot captured real content rather than the Cloudflare page, so that row is a data point your table doesn't actually have — and it lands at 9.0x.

CI is green across the board on `aabfba8d`, so the "Full CI test run" box looks tickable.

### The measurement window is what I'd push on

`reporter` doesn't include `getSelectorData` — that runs in `run-rules`, before the reporter phase — so all the new index-building work lands outside the window your table measures. Measuring `axe.setup()` directly:

| Page | develop setup | This PR setup | Index memory |
|---|---:|---:|---:|
| Stack Overflow | 58.8ms | 80.8ms | 6.3 MB |
| Large text page | 408ms | 546ms | 14.1 MB |
| 80k elements, unique ids | **277ms** | **20,039ms** | **802 MB** |

That last row is synthetic — 40k `<div id="row-N">` each wrapping a `<span id="c-N">`. Two independent causes, both inline below (`collectNodeAttrs` and the `nthChild` computation). I built both fixes locally and verified them: synthetic setup 20,039ms → 384ms, memory 802MB → 0.1MB, Large text page index 14.1MB → 1.3MB, reporter time unchanged (Large text page 4,554ms vs 4,641ms), and **0 selector diffs on all 8 pages including Large text page**.

The other two comments are edge cases where the new tree-scoped uniqueness check disagrees with develop's DOM-scoped one.

On `constants.selectorSimilarFilterLimit` — agree that cleanup belongs in a follow-up. Worth noting `test/core/constants.js:38` still asserts it equals `700`, so that assertion travels with it.

lib/core/utils/get-selector.js
```diff
@@ -337,66 +584,617 @@ function getThreeLeastCommonFeatures(elm, selectorData) {
...
-    let features = getElmId(elm);
-    if (!features) {
-      features = getThreeLeastCommonFeatures(elm, axe._selectorData);
-      features += getNthChildString(elm, features);
+
+  const perNode = selectorData._perNode;
+  const elmToIdx = selectorData._elmToIdx;
+  const targetIdx = elmToIdx.get(elm);
+  if (targetIdx === undefined) {
+    // A normal `axe.run` never gets here — every element that ends up in
```
chutchins25: Is the "every element that ends up in results was walked at `getSelectorData` time" assumption meant to hold for elements that are in the DOM but *not* in axe's flattened tree? That's a wider set than nodes added mid-run, and it changes what uniqueness means.

develop answered uniqueness with `doc.querySelectorAll` / `matchesSelector`, so the CSS engine saw the whole DOM. This branch answers it from the bitsets, which only contain nodes the flat tree walked. Feature *selection* was always tree-based, so that part is unchanged — but uniqueness *verification* was not, and anything the tree skips is now invisible to it.

**Reachable from a normal `axe.run`** — unslotted light-DOM children of a shadow host are dropped from the flattened tree but are still real DOM siblings:

```js
host.attachShadow({ mode: 'open' }).innerHTML = '<slot name="only"></slot>';
host.innerHTML = '<p slot="only" class="zz">a</p><p class="zz">b</p>';
// selector for the slotted <p>:
// develop -> "p:nth-child(1)"   (1 match)
// this PR -> "p"                (2 matches)
```

`b` isn't in `_elmToIdx`, so neither the sibling scan in `computeFragment` nor the candidate count sees it, and the `:nth-child` that develop added is dropped.

**Worse, but on a more forgiving path** — `axe.setup(node)` with a subtree. Uniqueness is then only checked within that subtree, while develop still queried the full document:

```js
document.body.innerHTML =
  '<div id="outside"><p id="dupe">A</p></div>' +
  '<div id="inside"><p id="dupe">B</p></div>';
axe.setup(document.querySelector('#inside'));
// develop -> "#inside > p"  (1 match, correct element)
// this PR -> "#dupe"        (2 matches, resolves to the WRONG element)
```

`getElmId` reads `_idCounts[rootId]`, which only counted ids inside the walked subtree, so a page-duplicated id now looks unique. I know `axe.setup(node)` is documented as "may result in incomplete results", so I'd rate that second case lower — but returning a selector that resolves to a *different* element is a sharper failure than an incomplete one, and it's worth deciding deliberately rather than inheriting it.

Both of these sit behind the "not a behavior change from an axe-user perspective" line in the description, which is why I'd rather ask than assume.

lib/core/utils/get-selector.js
```diff
@@ -337,66 +584,617 @@ function getThreeLeastCommonFeatures(elm, selectorData) {
...
+  // doesn't have to keep scanning for `$="` on every check.
+  const classes = [];
+  let exactAttrs = null;
+  let suffixAttrs = null;
+  for (const f of features) {
+    if (f.species === 'class') {
+      classes.push(f.name);
+      continue;
+    }
+    const sep = f.name.indexOf('$="');
```
chutchins25: Not a blocker, just checking the intent here — is sniffing for `$="` in the assembled `atnv` string meant to be the classifier, rather than a flag carried out of `getAttributeNameValue`?

It misfires when the attribute *name* ends in `$`. `escapeSelector('data-x$')` produces `data-x\$`, so the atnv is `data-x\$="q1"`, which contains `$="`. That routes it to `suffixAttrs` with `name: 'data-x\'` (everything before the separator), while `collectNodeAttrs` keyed `urlAttrs` by the raw `at.name`, `'data-x$'`. The lookup in `matchesSuffixAttrs` then misses, so the element stops matching its own fragment, the candidate set empties, and the walk falls through to the `:root` return.

```html
<div id="w">
  <span data-x$="q1"></span>
  <span data-x$="q2"></span>
</div>
```

develop returns `span[data-x\$="q1"]`; this branch returns `:root`. It fails silently — no throw, just a selector that no longer identifies the element.

Narrow, but not purely theoretical: Polymer-style `attr$="..."` binding syntax left unprocessed in serialized markup produces exactly these attribute names. Having `getAttributeNameValue` report which form it built would remove the guesswork at both this call site and the matching one in `collectNodeAttrs` (line 145).

lib/core/utils/get-selector/generate.js
```diff
@@ -0,0 +1,354 @@
...
+    const cached = rootChain.get(selector);
+    if (cached !== undefined) {
+      // Another target has already reached this exact chain string —
+      // reuse its filtered candidate list. Treat it as if we just did
+      // level 0: `anc` gets rebuilt from `indices` on the next filter.
+      indices = cached;
+      anc = null;
+      candCount = cached.length;
+    } else if (indices === null) {
+      indices = indicesForFragment(
```
chutchins25, line 295: **The `:nth-child` half of my last comment is fixed, but the candidate count is still tree-only — so a reported target can resolve to a different element.**

`indicesForFragment` here (and `bitsForFragment` on the ancestor steps) answers "how many match?" from the bitsets, which only hold nodes the flat tree walked. Unslotted light-DOM children of a shadow host are real DOM elements the flattened tree drops, so they're invisible here in a way they never were to develop's `findSimilar` → `doc.querySelectorAll`.

Reachable from a plain `axe.run`; unslotted host first in document order:

```html
<x-b id="hb"><p><img src="y.png"></p></x-b>  <!-- shadow: <slot name="nope">; img unslotted, not in tree -->
<x-a id="ha"><p><img src="x.png"></p></x-a>  <!-- shadow: <slot>; img slotted, flagged -->
```

`await axe.run(document, { runOnly: ['image-alt'] })`:

```
develop  → "#ha > p > img"   1 match   → the flagged <img>
this PR  → "img"             2 matches → the OTHER <img> (unrendered)
```

Reverse the hosts and the target still matches 2; it just lands on the right one first. This also puts pressure on "Not a behavior change from an axe-user perspective" in the description.

Cheapest fix that keeps the fast path: when `candCount === 1`, confirm with one scoped `root.querySelectorAll(selector)` before returning — one query per target, only on the terminating path. Seeding the bitsets from `root.querySelectorAll('*')` instead of the flat tree also works, at the cost of a bigger index.

lib/core/utils/get-selector/bit-index.js
```diff
@@ -0,0 +1,445 @@
...
+ */
+function getMatchCacheEntry(selectorData, rootId, key) {
+  let cache = selectorData._matchCache;
+  if (!cache) {
+    cache = selectorData._matchCache = [];
+  }
+  let rootCache = cache[rootId];
+  if (!rootCache) {
+    rootCache = cache[rootId] = {};
+  }
```
chutchins25, line 223: **`rootCache` inherits from `Object.prototype`, so a bare-tag fragment key of `constructor` shares an entry with the global `Object`.**

Fragment strings are prefixed almost everywhere (`.cls`, `[a="b"]`, `#id`) — a bare tag name isn't, and `<constructor>` parses as an unknown element. `rootCache['constructor']` returns the `Object` constructor, which is truthy, so the `if (!entry)` never fires and `entry.bits = bits` writes onto `Object`. Diffed against develop:

```
① cross-root bleed  — shadow <constructor> reads root 0's bits
     develop ["#hh","constructor"]   →  this PR ["#hh",":root"]
② stale across runs — run 1's bits reused in run 2 after the DOM changed
     develop "div > constructor"     →  this PR "constructor" (2 matches)
③ Object.bits survives axe.teardown(), retaining a Uint32Array for the page's life
```

② is the one that bothers me — the second run gets an answer computed against the previous run's `_perNode` indices. `_chainCache` uses a `Map`, which is why this only shows up here.

```suggestion
  let rootCache = cache[rootId];
  if (!rootCache) {
    // Null prototype: fragment strings are untrusted keys — a bare tag name
    // can be `constructor` — and an inherited hit would hand one root's
    // bitmap to another, then strand it on `Object` for the life of the page.
    rootCache = cache[rootId] = Object.create(null);
  }
```

lib/core/utils/get-selector/generate.js
```diff
@@ -0,0 +1,354 @@
...
+    // (`parentElement`) because measuring showed that's actually faster
+    // than looking up `parentOf[curIdx]` and then converting back to an
+    // Element via `perNode[idx].elm`. `curIdx` is kept in sync so
+    // `computeFragment`'s per-nodeIdx cache stays hot.
+    curElm = curElm.parentElement;
+    curIdx = curElm ? elmToIdx.get(curElm) : -1;
+    if (curIdx === undefined) {
+      curIdx = -1;
+    }
+    // Node type 11 is DocumentFragment (the parent of `<html>` in a
```
chutchins25: Not a blocker, just two comments that don't match the code — flagging them because the docs on this PR are good enough that the next reader will trust them.

- **Here:** the parent of `<html>` is the Document (nodeType 9), and `documentElement.parentElement` is `null`. `curElm` comes from `.parentElement` on line 337, so it's only ever an Element or `null` — `nodeType !== 11` can never be false. The guard is carried over from develop (`get-selector.js:394`); it's the new rationale that isn't the rationale.
- **`precompute.js:81`:** `@param {Map} cache  WeakMap<Element, ...>` — it's a plain `Map` (`:264`). Right choice for the usage, but the annotation contradicts itself.

chutchins25 (dismissed):
Re-reviewed on `a33f767d` (last round was `aabfba8d`). Nice turnaround on those four — the `filterAttributes` gate, the per-parent nth-child map and `_suffixInfo` all landed as proposed, and the split reads far better than the 1200-line version. I rebuilt develop (`78d90849`) and this branch into separate bundles and re-ran each case in headless Chrome.

The threads were resolved without replies, so here's my read of where they landed — correct me where I'm wrong:

| Thread | Now in | Status |
|---|---|---|
| @WilcoFiers — break up the file | 5 modules under `get-selector/` | ✅ Entry point is 20 lines |
| Attr bitsets for every attribute | `precompute.js:171-187` | ✅ `urlAttrs` correctly outside the gate |
| `indexOf` per node → O(n²) | `precompute.js:84-95`, `:367` | ✅ |
| `$="` sniff on names ending in `$` | `attr-format.js:101` | ✅ `span[data-x\$="q1"]`, matches develop |
| Tree- vs DOM-scoped uniqueness | `generate.js:170-176` | ⚠️ `:nth-child` half fixed; **count half still reproduces** — inline on `generate.js:303` |

One process item: `9d7966bd` changed behavior in four places and added no tests — its `--stat` touches only `get-selector.js`, and all 179 added test lines came in `aabfba8d`. Each fix already has an exact repro in the thread it answered; the unslotted-sibling and `$`-suffix ones are correctness fixes with two-line HTML repros, so they're straightforward regression tests. The two perf ones can assert shape rather than timing (`_attrBits` key count bounded, `nthChild` correct on a wide parent) to avoid flakiness.

Two smaller notes inline.

straker:
Round 2 addressed in `493c1a41`:

## Correctness

- **Prototype pollution** (`bit-index.js`): `Object.create(null)` for the per-root `_matchCache` bucket. Regression test lands a `<constructor>` fragment and asserts `Object.bits` never appears.
- **Unslotted-host uniqueness** (`generate.js:303`): phantom-records pass at the end of `precompute.js`. Every element in a shadow host's light-DOM that flatten-tree dropped now gets a full `_perNode` record so `_tagBits` / `_classBits` / `_attrBits` / `_rootBits` and the `data.tags` / `data.classes` / `data.attributes` counts all see the true DOM shape. Phantoms stay out of `_elmToIdx` and `_idToIdx` so `getSelector` / `getElmId` never resolve to one — they exist purely to make bitmap counting DOM-aware. Your exact repro: `img` (2 matches, wrong element) → `img[src="x.png"]` (1 match, correct element).

## Why not verify-with-querySelectorAll at candCount === 1

I prototyped the "cheap" `root.querySelectorAll(selector)`-at-terminate approach and measured. Reporter phase vs current algorithm:

| Page | Nodes | Current | + QSA verify |
|---|---:|---:|---:|
| MDN ElementInternals | 10.3k | 51ms | 1,969ms (38× slower) |
| TensorFlow.js API | 35k | 305ms | 30.5s (100× slower) |
| Large text page | 152k | 1.66s | script timeout at 600s |

The old algorithm could absorb one more QSA at the terminating path because QSA was already the dominant cost — one more was noise. Bitmap intersection replaced that entirely; adding one QSA back per target on the successful path puts the cost right back into the reporter phase at the same order of magnitude the PR was set up to remove. On the Large text page that's ~310k targets × one QSA each.

The phantom pass moves the correctness fix to setup instead: one-time cost proportional to unslotted content (0 on all 8 canonical pages), no query-time cost. Full `axe.run` delta across all 8 pages after applying: −24ms to +156ms (inside noise), setup +5% at worst on the Large text page, memory unchanged everywhere.

Alternative I considered: seeding bitsets from `root.querySelectorAll('*')` per root instead of the flat tree, which the review suggested. It also fixes correctness without a query-time cost, but changes what the walk indexes (all DOM elements, not just those the flat tree emitted) — a bigger blast radius. Phantoms are a narrower fix scoped exactly to the delta between the flat tree and the DOM.

## Nits + process

- Removed the dead `nodeType !== 11` guard in `generate.js` — you were right, `curElm.parentElement` only ever returns Element or null, so it never fires.
- Fixed the `WeakMap` → `Map` JSDoc typo on `precompute.js:81`.
- Added six regression tests covering the four round-1 fixes plus the two round-2 correctness fixes. Shape assertions for the perf-shape ones (`_attrBits` gate, `nthChild` on a wide parent) so they don't flake on timing; input/output for the correctness ones (unslotted-host, `$`-suffix, prototype-pollution). 56/56 pass.

Selectors remain byte-identical to develop on all 8 canonical pages.

lib/core/utils/get-selector/precompute.js
```diff
@@ -0,0 +1,590 @@
...
+  // why. For each shadow host we visited in the flat-tree walk, iterate
+  // its real DOM `.children`. Any child NOT in `_elmToIdx` was dropped
+  // by flatten-tree (unslotted content, or content that only matched a
+  // <slot> we're not walking) and needs a phantom record so its
+  // existence shows up in bitmap counts.
+  for (let h = 0; h < shadowHostList.length; h++) {
+    const hostIdx = shadowHostList[h];
+    const hostElm = perNode[hostIdx].elm;
+    const hostRootId = perNode[hostIdx].rootId;
+    const domKids = hostElm.children;
```
chutchins25: **This pass covers one of three ways flatten-tree drops DOM elements. A second is covered incidentally; the third still resolves to the wrong element.**

| Dropped by flatten-tree | Covered by |
|---|---|
| Unslotted light-DOM children of a host | ✅ this pass |
| `<slot>` elements themselves (`get-flattened-tree.js:145`) | ✅ incidentally — they're siblings, so the sibling scan's `matchesSelector` fallback catches them (verified) |
| **Slot fallback content when the slot has assigned nodes** (`:146-149`) | ❌ nothing |

Fallback content sits *inside* the `<slot>`, in the shadow root — not a host light-DOM child (this loop misses it), not a sibling of shadow-root targets (the sibling fallback misses it):

```js
host.attachShadow({ mode: 'open' }).innerHTML =
  '<slot><span class="zz">fallback</span></slot>' +
  '<div><span class="zz">target</span></div>';
// host light DOM: <em>assigned</em> — so the fallback is dropped
```

```
develop → ["#hf","div > span"]   1 match   → correct element
this PR → ["#hf","span"]         2 matches → resolves to the FALLBACK span
```

Probed from `_selectorData`: `_elmToIdx.has(fallbackSpan) === false`, `data.tags.SPAN === 1`, one bit set in `_tagBits.span` — against `shadowRoot.querySelectorAll('span').length === 2`. `<slot name="icon"><svg …></slot>` is common, and it only misbehaves when consumers *do* pass content. (Closed roots are fine — `shadowRoot` is null, so the light DOM is walked normally and the closed content is invisible to `querySelectorAll` too.)

Smallest fix in this pass's shape: also walk each shadow root's `<slot>` elements and phantom-record their element children when `assignedNodes().length > 0`. The v0 `<content>` branch at `:137-143` is the same class if you want to be exhaustive.

Longer term, enumerating drop-sites means revisiting this pass whenever flatten-tree changes what it emits, with no test that would tell you a new one appeared. The `root.querySelectorAll('*')` seeding you set aside is immune by construction — and per-root-at-setup is a different cost profile from the per-target-at-query one you measured, so it may survive where that didn't. Your call; I just don't think they were the same proposal.

chutchins25 (commented):
Round 3 on `b95c8419`. Phantom records are the right shape, and rejecting my `querySelectorAll`-at-terminate suggestion was the right call — 100× on TensorFlow and a timeout on the Large text page settles it. CI green. Rebuilt develop (`78d90849`) and this head and re-ran every case:

| Round-2 finding | Status |
|---|---|
| Tree-scoped candidate count | ◐ My repro is fixed (`img` 2 matches → `img[src="x.png"]` 1 match, correct element). One more drop-site still reproduces — inline on `precompute.js:441` |
| `Object.prototype` cache key | ✅ Resolved; rewriting the test away from asserting `Object.bits` was the right move |
| No tests for round-1 fixes | ◐ See below |
| `nodeType !== 11` / `WeakMap` JSDoc | ✅ Both resolved. Deleting the guard is safe — `curElm` only comes from `.parentElement` |

### On the dropped tests

`941fa3d5` leaves the two round-1 perf fixes — the 802 MB index blowup and the 20-second `axe.setup()` — with no regression coverage. Both are invisible to I/O tests, so if the `filterAttributes` gate or the per-parent nth-child map comes back out, nothing catches it.

I take the point about not coupling to branch-only structures, though this file already reads `axe._selectorData` at lines 44, 51 and 632. Would guarding them (`if (!axe._selectorData._attrBits) return;`) give you develop-compatibility while keeping them load-bearing here? Fine if the answer is that the perf tables cover it — I'd just rather it be a decision than a side effect.

Also: `b95c8419` stripped the "why" comments from eight stress tests, which recorded the failure mode each one guards; the assertions don't carry that alone. And "Added six regression tests… 56/56" was true at `493c1a41`, but three landed by `b95c8419`.

straker:
> Fine if the answer is that the perf tables cover it — I'd just rather it be a decision than a side effect. 

When we implement https://github.com/dequelabs/axe-core/issues/5301, these types of performance problems will be caught and reported. This will be the better way to guard against bad performance problems rather than unit tests.

> The v0 <content> branch at :137-143 is the same class if you want to be exhaustive.

Our docs say [we do not support v0 shadow DOM](https://github.com/dequelabs/axe-core/blob/develop/README.md#supported-browsers) so I'm fine not supporting correct selectors for it

> Longer term, enumerating drop-sites means revisiting this pass whenever flatten-tree changes what it emits, with no test that would tell you a new one appeared. The root.querySelectorAll('*') seeding you set aside is immune by construction

Ya this is a problem with the current code. Longer term my thinking is to combine the 2 selector caches we have in `selector-cache` and this one. That would solve the problem of QSA and this code missing stuff when we change flatten tree (since selector cache is generated at flatten tree time).

lib/core/utils/get-selector/precompute.js
```diff
@@ -0,0 +1,641 @@
...
+          // element inside the shadow tree if we walked it, else the
+          // host (flatten-tree grafts top-of-shadow content onto the
+          // host as its virtual parent).
+          const walkedParentIdx =
+            slot.parentElement && data._elmToIdx.get(slot.parentElement);
+          const slotParentIdx =
+            walkedParentIdx !== undefined ? walkedParentIdx : hostIdx;
+          const fallbackKids = slot.children;
+          for (let k = 0; k < fallbackKids.length; k++) {
+            phantomWalk(fallbackKids[k], slotParentIdx, shadowRootId);
```
chutchins25: **The fallback children get phantom records but the `<slot>` that parents them doesn't, so `parentIdx` lands one level too high and the ancestor filter drops the phantom.**

Same failure class as the round-2 finding — develop is correct, this branch resolves to a different element:

```js
host.attachShadow({ mode: 'open' }).innerHTML =
  '<main>' +
    '<section><slot class="wrap"><span>fallback</span></slot></section>' +
    '<div class="wrap"><span>target</span></div>' +
    '<div></div><div></div><div></div>' +
  '</main>';
host.innerHTML = '<em>assigned</em>';
```

```
develop → "main > .wrap > span"   1 match   → target
this PR → ".wrap > span"          2 matches → resolves to the FALLBACK span
```

Probed from `_selectorData`: no phantom record exists for the `<slot>`, and the fallback span's `parentIdx` is the `<section>`. At level 1 the fragment is `.wrap`; `<section>` isn't `.wrap`, so the phantom drops, `candCount` hits 1, and `.wrap > span` ships — even though the slot *is* `.wrap` and the fallback really does match it.

It needs the slot to not be a sibling of the target's ancestor. When it is a sibling, `computeFragment`'s `matchesSelector` fallback adds `:nth-child` and saves it — which is why the test you added passes.

**Second defect in the same expression**, which the fix below makes moot:

```js
const walkedParentIdx = slot.parentElement && data._elmToIdx.get(slot.parentElement);
const slotParentIdx = walkedParentIdx !== undefined ? walkedParentIdx : hostIdx;
```

When the slot is a direct child of the shadow root — the common case, including your new test — `slot.parentElement` is `null`, so `walkedParentIdx` is `null`, and `null !== undefined` is true. The `hostIdx` fallback the comment describes is unreachable, and `parentIdx: null` coerces to `0` in the `_parentOf` copy, grafting the phantom onto whatever `nodeIdx 0` is. Measured on your test's own markup: `parentIdx` is `null`, `_parentOf` reads `0`, `hostIdx` was `2`.

Suggestion: phantom-record the `<slot>` itself — parented at its walked parent, else `hostIdx` — and parent the fallback children at the slot's phantom index. That restores the missing chain level, puts the slot in the bitmaps (drop-site #3, currently only covered incidentally by the sibling scan), and removes the `null`/`undefined` question.

Minor, same block: `if (hostElm.shadowRoot)` at `:465` can never be false — `shadowHostList` only receives indices where `node.shadowRoot` was truthy. Same shape as the `nodeType !== 11` guard you deleted last round.

chutchins25 (commented):
Round 4 on `59cb0b36`. Checked your references — #5301 is open and is the right home for perf guarding, and the README does list v0 shadow DOM as unsupported. Both answers work for me; the test question is settled as far as I'm concerned. Merging this with `selector-cache` at flatten-tree time is a better direction than either thing I suggested.

Rebuilt develop (`78d90849`) and this head:

| Round-3 item | Status |
|---|---|
| Slot fallback drop-site | ◐ My repro is fixed (`["#hf","div > span"]`, identical to develop). The fix is one ancestor level short though — inline on `precompute.js:493` |
| Dropped perf tests | ✅ Answered — deferred to #5301 |
| v0 `<content>` | ✅ Answered — unsupported per README |
| Enumeration vs construction | ✅ Acknowledged, with a plan |

Also confirmed no regression: ordinary pages and shadow roots give byte-identical selectors to develop.

lib/core/utils/get-selector/precompute.js
```diff
@@ -0,0 +1,665 @@
...
+    if (shadowRootId === undefined) {
+      continue;
+    }
+    const slots = hostElm.shadowRoot.querySelectorAll('slot');
+    for (let s = 0; s < slots.length; s++) {
+      const slot = slots[s];
+      if (typeof slot.assignedNodes !== 'function') {
+        continue;
+      }
+      if (slot.assignedNodes().length === 0) {
```
chutchins25: **This skip leaves the more common of the two slot shapes: a slot nobody filled.** Same markup as my round-4 repro with the host's light DOM empty:

```js
host.attachShadow({ mode: 'open' }).innerHTML =
  '<main>' +
    '<section><slot class="wrap"><span>fallback</span></slot></section>' +
    '<div class="wrap"><span>target</span></div>' +
    '<div></div><div></div><div></div>' +
  '</main>';
// no host.innerHTML — the slot has no assigned nodes
```

```
develop → "main > .wrap > span"   1 match   → target
this PR → ".wrap > span"          2 matches → resolves to the FALLBACK span
```

The comment is right that flatten-tree walks the fallback children in the slot's place. Two things follow from that which the skip doesn't account for:

1. The **slot itself** still gets no record, so drop-site #2 is covered only for slots that *have* assigned nodes — "drop-site #2 is now covered explicitly" holds for half the cases.
2. In the **main walk** those fallback children are real nodes, but `node.parentNode` is the `<slot>`, which isn't in `_elmToIdx`, so `parentIdx` stays `-1`. Probed: `fallbackIsRealNode: true`, `fallbackParentIdx: "-1"`, `slotHasPhantom: false`, `assignedNodes: 0`.

Then `bit-index.js:446` drops them:

```js
const parent = parentOf[source[i]];
if (parent < 0) {
  continue;
}
```

The fallback span leaves every candidate set at the first ancestor step regardless of fragment, `candCount` hits 1 early, and `.wrap > span` ships.

**Narrow fix:** phantom-record slots regardless of `assignedNodes().length`, and for empty ones re-point their real children's `parentIdx` at the slot's phantom index. The `_parentOf` copy runs after the phantom pass, so a fix-up lands before anything reads it.

**The one I'd actually argue for:** `parent < 0 → continue` is what turns every missing-ancestor case into a silently **too low** count, and too low is the only direction that produces a wrong element — too high just produces a longer selector. Making that branch conservative (keep the candidate, or verify it with `matchesSelector`) bounds this whole class whether or not the drop-site enumeration is complete.

This is the third round running where the enumeration came up one case short, which I think is really an argument for your `selector-cache`-at-flatten-tree-time plan rather than against this PR. A conservative `-1` in the meantime would stop the enumeration from being load-bearing for correctness while that lands.

chutchins25 (commented):
Round 5 on `83c6ef12`. Handing the slot to `phantomWalk` as a unit was the right call — it fixed both round-4 defects with less code than the version it replaced, and `hostIdx` is actually reachable now.

| Round-4 item | Status |
|---|---|
| Phantom chain skipped the `<slot>` level | ✅ Resolved — my repro returns `main > .wrap > span`, identical to develop |
| `null` sneaking past `!== undefined` | ✅ Resolved |
| Dead `if (hostElm.shadowRoot)` guard | ✅ Removed |

No regressions — ordinary pages and shadow roots still match develop exactly. CI green.

One case left, inline on `precompute.js:498`: the same repro with the slot *unfilled*, which is the more common of the two shapes.

lib/core/utils/get-selector/precompute.js
straker: Went with doing the QSA step for each root to catch all these instead of continually adding to each edge-case.

chutchins25 (dismissed):
Approving on `2c514c16`.

Replacing the drop-site enumeration with a `root.querySelectorAll('*')` seed per root is the right call, and it came out ~30 lines *shorter* than the enumeration it replaced. Six rounds in, this is the first version where correctness doesn't depend on the drop-site list being complete — which is the property that was actually missing.

Rebuilt develop (`78d90849`), `83c6ef12`, and this head, and re-ran everything:

| Prior finding | On `2c514c16` |
|---|---|
| R2 unslotted light-DOM children (via real `axe.run`) | ✅ `img[src="x.png"]`, 1 match, correct element |
| R3 slot fallback with assigned nodes | ✅ `div > span`, identical to develop |
| R4 phantom chain skipping the `<slot>` level | ✅ `main > .wrap > span`, identical to develop |
| R5 empty-slot fallback | ✅ `main > .wrap > span`, identical to develop |
| R2 `axe.setup(subtree)` duplicate id — I'd dropped this in round 2 | ✅ Incidentally fixed: `#inside > p` on both |

I also checked the two load-bearing claims in the commit message rather than taking them on faith, and both hold:

- **"Setup slightly faster"** — three-way on a 60k-element page: develop 166ms, `83c6ef12` 241ms, this head 239ms. The QSA seed is free relative to the enumeration it replaced; the delta against develop is the pre-existing index build already documented in the body.
- **"Every DOM sibling is in `_elmToIdx`"** — swept every element in every root looking for one that wasn't, found none. Removing the `matchesSelector` fallback in the sibling scan is safe.

Regression sweep on ordinary pages and shadow roots: selectors byte-identical to develop.

Nice work sticking with this through six rounds — the perf win is large and the correctness story is in much better shape than where it started.

WilcoFiers (approved):

PR #5335, opened 2026-08-30
feat(new-rule): flag image alternatives that are a file name, path or URL

### LINKED ISSUE #3685: Warn alt attributes that are files / paths
### Product

axe-core

### Feature Description

The latest [Web Almanac section on accessibility](https://almanac.httparchive.org/en/2022/accessibility#conclusion) has this nugget at the end:

> …we see that 2.2% of alt attributes across websites have a file extension in them…

From memory, I used to encounter these frequently enough in accessibility reviews that I knew Axe does not flag those for review. While 2.2% may be a small percentage, with 8,360,179 sites in the June 2022 data set, that suggests up to 183,924 sites could have this construct (all those bad alts could live on one site for all I can tell from the report).

Obviously there are cases where this may be valid, such as for downloads where the `alt` represents either text of the image, file links in general, or when the file name is incidental to other content in the `alt`.

I built a test: https://codepen.io/aardrian/pen/jOxYZXd

The code:
```
<p>
<img src="https://assets.codepen.io/2729/steamed-hams_08.jpg" alt="https://assets.codepen.io/2729/steamed-hams_08.jpg">
</p>

<p>
<img src="https://assets.codepen.io/2729/steamed-hams_08.png" alt="steamed-hams_08.jpg">
</p>
```

One of those is arguably a 1.1.1 issue and the other arguably a 1.4.5 issue. But a human would need to confirm that.

My suggestion is that Axe flags for review images with `alt` attributes that can be file names/paths. Whether looking at extensions, URL patterns, slashes, and so on, I defer to the multitude of regexes already out there.

Currently WAVE and ARC both flag those two constructs for review.

I do not know what editors / CMSes / tools still default to the file name as the `alt` value. [Hidde indicated](https://twitter.com/hdv/status/1574743520346509314) that Perch does. I know some WordPress plug-ins did (maybe still do), but I cannot recall which. Either way, it is not clear how much of that 2.2% is legacy content and how much is new.

dylanpulver:
Adds `image-alt-file-name`, a best-practice rule that reports an image as **needs review** when its accessible name is the file name, path or URL of the image. Never a violation — as the issue says, whether a file name is an adequate alternative is a human call.

The scoping is the part worth reviewing. The check fires only when the *whole* accessible name is a file reference: it ends in one of the configurable image extensions, or it is a URL. I measured the alternatives over 11,888 non-empty `alt` values from 555 pages across 99 sites:

- **as shipped** — 54 hits (0.45%), 32 unique, all genuinely file names or bare URLs
- **plus a bare `/` or `\` path clause** — also flags `NASA/JPL`, a photo credit, so that clause is out
- **plus a no-whitespace constraint** — drops 8 of the 54, e.g. `[Hi-Res Image] Saul Williams by Sam Lee.jpeg`, so that is out too

Deliberately not flagged: non-image extensions (`Annual report.pdf` — a file link's name can be the right label), and a file name incidental to surrounding text (`Download photo.jpg from the archive`). Known limitation, disclosed: an alt that *ends* on an incidental file name, e.g. `Download logo.png`, is flagged.

Nothing is compared against `src`, so this is not a revival of ACT 9eb3f6 / #1504.

Full suite, integration, virtual-rules, examples, jsdom, node, eslint and prettier all green locally. Happy to ship it `enabled: false` if you would rather it be opt-in.

Closes: [#3685](https://github.com/dequelabs/axe-core/issues/3685)


straker:
Thanks for the pr! From a code standpoint this looks fine, but I'd like @WilcoFiers to review it as well.

lib/rules/image-alt-file-name.json
```diff
@@ -0,0 +1,14 @@
+{
+  "id": "image-alt-file-name",
+  "impact": "minor",
+  "selector": "img",
+  "tags": ["cat.text-alternatives", "best-practice"],
```
WilcoFiers: I'm leaning that this should be a critical issue under SC 1.1.1, not a minor best practice issue. This very likely is a real issue if Axe hits it, and you're already avoiding false positives because of reviewOnFail.

WilcoFiers (changes requested):
Solid work. I'd make this a WCAG rule rather than a best practice.

WilcoFiers (approved):

straker:
@dylanpulver would be able able to sign the CLA? Once you sign it we can merge this in.

PR #5338, opened 2026-09-01
perf: track axe-core perf metrics

straker:
Adds perf tracking script and logs to track axe-core performance over time and a github action that can be triggered to run perf against any version of axe-core. This will let us do @WilcoFiers request after this is merged.

The initial site is a built version of the [open source MDN docs](https://github.com/mdn/content) for the ElementInternals page (which I have been using as a benchmark site). They apparently load _a lot_ of javascript files which are needed as without the files the page can run axe-core much quicker than the live site. I combined all the CSS and JS files into a single file in order to keep the number of files down. Before develops perf changes axe-core would run in ~3s, and without the js files it would run in ~600ms.

The `very-large-page` is a basically my largest benchmark page of just text, but using lorium ipsum.

Using the log to alert us of performance problems during development will be done in separate prs.

For reviewing, the main files are `perf/report.js` which runs all the sites and reports their performance metrics, `.github/workflows/perf-report.yml` which is the github action script that can run the perf report for any axe-core version, `package.json` which adds a package script to run the report, and `.eslint.config.js` which ignores the `perf/sites` directory. Anything in `perf/sites` is from 3rd party content.

Ref: #5301


stephenmathieson (dismissed):
Does all of this need to be versioned in this repository? This PR contains hundreds of generated files (eg `perf/sites/mdn/static/client/3255.ffeced2a4f21cd01.js`) that IMO should be added to `.gitignore` and added to our build process rather than versioning them in Git.

straker:
That won't be possible. The built repo file is just a JSON file that get's used in their server to process the page. Just getting a static HTML file with all the scripts took some work that wouldn't be possible in a build step.

straker:
Sorry, not that it wouldn't be possible. But it requires opening up a local server to see the page, downloading it and then finding all assets and resolving them. Not really a build process IMO.

WilcoFiers (dismissed):
Before you close that ticket I wouldn't mind actually seeing a chart that shows performance of the last 5 minor versions in a chart. Can we make that happen? Doesn't need to be this PR, probably best not considering how big it is. But maybe just don't close this one?

As far as checking these files in, I'm fine with it.

straker:
Added a github action to generate the pef logs for specific versions of axe-core so we can get the last 5 versions data. Also added my largest benchmark site as it shows the largest perf wins.

package.json
```diff
@@ -113,9 +113,9 @@
     "rule-gen": "node build/rule-generator.mjs",
     "sri-update": "pnpm run build && node build/sri-update.mjs && git add sri-history.json",
     "sri-validate": "node build/sri-update.mjs --validate",
-    "codemod:modernize-tests": "node build/codemods/modernize-test-files.mjs",
```
straker, line 116: No longer used/needed

chutchins25 (approved):
Looks good to me. Had Claude go over this a few times and nothing major was found. Reviewed some of this with Steve as well. Thankfully too as the change set can look pretty scary at first glance.

WilcoFiers (approved):

PR #5342, opened 2026-09-03
chore: Readability tweaks get-selector rewrite

WilcoFiers:
A readability pass over the `get-selector` rewrite. Mostly renames and
function extraction, plus two behavior changes called out below.

## Renames

The old names leaned on abbreviations that didn't survive re-reading:

| Before | After |
| --- | --- |
| `_perNode` / `rec` | `_nodeRecords` / `nodeRecord` |
| `feature.species` | `feature.kind` |
| `rec.attrs` / `rec.matchAttrs` | `featureAttrs` / `exactAttrs` |
| `at` | `attr` |
| `advanceAndFilter` | `matchingParents` |
| `getMatchCacheEntry` | `getCachedMatch` |
| `data` (in precompute) | `selectorData` |
| `{ selector }` from `getThreeLeastCommonFeatures` | `{ selectorString }` |

The inline shape comments are now JSDoc typedefs — `SelectorData`,
`NodeRecord`, `IndexState`, `Fragment`, `Feature`, `CachedMatch` — so
the fields are documented once and referenced everywhere else.

## Refactors

**`precompute.js`** — `getSelectorData` was one long function with a nested closure. It's now five named steps: `createIndexState`, `walkFlatTree`, `seedPhantomsFromRoots`, `repairMissingParents`, `buildIndexes`. The flat-tree walk and the phantom pass had two near-identical copies of the "read the element, update the tables" logic; they now share `indexElement`, differing only in how the parent is resolved and whether the element feeds the frequency counts.

**`generate.js`** — `computeFragment` split into `buildFragment`, `classifyFeatures` and `withNthChildIfNeeded`, with the cache lookup left in `computeFragment`. The candidate-list bookkeeping in `generateSelector` (three interacting variables: `indices`, `anc`, `candCount`) collapsed into one `candidatesForChain` call per step.

**`bit-index.js`** — `matchingParents` now owns the bitmap lookup and the working buffer instead of taking them as arguments; the buffer lives on `selectorData._candidateBuf` and is reused across walks, with the chain cache storing copies.

`generateSelector`'s `options` argument is now optional.

## Behavior changes

**Phantoms no longer feed the public frequency counts.** Selectors are compared across axe runs, so the counts that drive feature selection have to reflect the same set of elements axe reports on. Letting unslotted light DOM or slot fallback markup into `selectorData.classes` shifts which class gets picked and rewrites the selector string. 
Uniqueness is a separate question, still answered by the bitmaps and id tables, which phantoms do feed.

**Phantoms are now in `_idToIdx`.** They used to only bump `_idCounts`, which meant a `<slot id="uniq">` couldn't be named by `#uniq` even though `querySelectorAll` sees it perfectly well. `getElmId` now resolves through `_idToIdx` (a hit is proof of uniqueness, since `pruneDuplicateIds` drops everything else) and falls back to a live DOM query for elements the walk never recorded.

## Tests

Every selector test that only asserted "matches exactly one element" now also asserts the exact selector string, so a refactor that silently changes the output fails loudly instead of passing. New cases cover slots colliding with the target on a class or an id, slot fallback content under a slot with a unique id, and a `feature count stability` block that pins the frequency counts for unslotted light DOM, slot elements, and assigned-slot fallback.

### DIFF
--- a/lib/core/utils/get-selector/attr-format.js
+++ b/lib/core/utils/get-selector/attr-format.js
@@ -92,14 +92,14 @@ function escapeAttribute(str) {
  * introduces a `\$="` that trips the same check. Reporting the form
  * explicitly avoids the guesswork.
  *
- * @param {Element} node		The element that has the attribute
- * @param {Attribute} at		The attribute
+ * @param {Element} node   The element that has the attribute
+ * @param {Attr}    attr   The attribute
  * @returns {{atnv: String, isSuffix: Boolean, rawName: String, suffix: String}}
  *   `rawName` is the original attribute name (not CSS-escaped).
  *   `suffix` is set only when `isSuffix` is true.
  */
-export function getAttributeNameValue(node, at) {
-  const name = at.name;
+export function getAttributeNameValue(node, attr) {
+  const name = attr.name;
   const escName = escapeSelector(name);
 
   // Substring match (not equality) catches related URL-carrying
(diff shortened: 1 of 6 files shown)


straker (approved):
All the changes LGTM.

straker (approved):

PR #5343, opened 2026-09-06
test(target-size): add regression coverage for wrapped-span false positive

### LINKED ISSUE #4295: Incorrect reporting of target size violation from incorrect offset calculation
### Product

axe-core

### Product Version

4.8.3

### Latest Version

#### Expectation

I would expect no target size violations to occur in the example.

#### Actual

The link labelled "Incorrectly flagged as target size violation" is in violation of target size, claiming it has 0 offset from both of the the other links.

IMG_2406

#### How to Reproduce

Unfortunately I was not able to reproduce this in jsfiddle (or the like), so the following html will need to be saved and opened as a file in your browser. All instructions to reproduce are in the comment in the code snippet.

```html
<!DOCTYPE html>
<html lang="en-US">
	<head>
		<title>Target Size</title>
	</head>
​
	<body>
		<div style="max-width:290px">
			<p>
				<a href="#">lorem ipsum dolor</a>
				<a href="#">amet</a>
				<span>et dolore magna aliqua</span>
			</p>
			<a href="#" style="display: inline-block;">Incorrectly flagged as target size violation</a>
		</div>
	</body>
​
	<!--
​
	In your browser, do the following (could not replicate in jsfiddle or the like)
​
	1. Open this HTML file
	2. Open the devtools console on that page - the rest of the instruction will take place there.
	3. Paste in axe source to make the global `axe` available on the page
​
	   For example: https://www.unpkg.com/axe-core@4.8.3/axe.min.js
​
	4. Configure axe to ONLY report on target size violations
​
		axe.configure({
			rules: [{id: 'target-size', enabled: true}],
			disableOtherRules: true
		})
​
	5. Run axe and print out the violations
​
		axe.run().then(r => console.log(r.violations));
​
	At this point you will observe that the last link is flagged as being in violation
	of target size. Specifically, it states that it has 0 offset from the other 2 links
	on the page, despite clearly being far away from them.
​
	This can be replicated in Chrome, Firefox, Safari
​
	-->
</html>
```

#### Additional context

The test case above is very carefully constructed. Here is what I found to be important about it specifically (and what is not)

- the div must be sized such that the paragraph wraps to multiple lines
- re: the first paragraph
  - The 2 links in here must be close enough to each other that they _should_ be flagged as violating offset limitations _if_ the were inline-block (its my understanding inline elements are not caught by axe present day for target-size violations?)
  - The text after the links _must_ be wrapped in a span
  - This span _must_ break in the middle as part of the wrapping forced by the div
- The final link outside the paragraph  _must_ be `display: inline-block;`

arpitbharadwaj1:
## Background

#4295 reported that a link far away from a paragraph was incorrectly flagged as a `target-size` violation, claiming "0 offset" from two other links inside the paragraph. The maintainer's own debug notes traced it to `get-target-rects.js`: a `<span>` wrapping across two lines was being treated as **one merged obscuring rect** (spanning the full vertical extent of both wrapped lines) instead of one rect per line, so `split-rects` found no non-overlapping space and returned an empty array — which then made the far-away link look fully obscured.

## What I found

I couldn't reproduce the bug. Loaded the exact reported repro in a real browser and confirmed:
- `span.getClientRects()` natively returns one rect **per wrapped line** (4 rects in my test), not one merged rect — so the raw browser API isn't the issue.
- Running current axe-core (4.10.2) against the exact repro via CDN: `target-size` passes with zero violations.

This looks like it's already fixed — most likely as a side effect of some other `get-target-rects`/`split-rects` change since the issue was filed — but there's no test locking this scenario in, so it could regress silently.

## Change

Adds a regression test reproducing the original report exactly (wrapped inline `<span>` inside a narrow container, plus an unrelated link elsewhere on the page) to `test/integration/rules/target-size/`, expecting a pass. No source changes — this is test-only, confirming and protecting behavior that already works.

Closes #4295

### DIFF
--- a/test/integration/rules/target-size/target-size.html
+++ b/test/integration/rules/target-size/target-size.html
@@ -235,6 +235,20 @@ this should not effect their outcome -->
   </a>
 </p>
 
+<!-- https://github.com/dequelabs/axe-core/issues/4295
+An inline <span> wrapping across multiple lines must not be treated as a
+single merged obscuring rect that fully overlaps everything near it -->
+<div style="max-width: 290px">
+  <p>
+    <a href="#" id="pass-wrapped-span-1">lorem ipsum dolor</a>
+    <a href="#" id="pass-wrapped-span-2">amet</a>
+    <span>et dolore magna aliqua</span>
+  </p>
+  <a href="#" id="pass-wrapped-span-3" style="display: inline-block"
+    >Not obscured by the wrapped span above</a
+  >
+</div>
+
 <!-- Inapplicable -->
 <p>
   The quick <a href="#" id="inapplicable1">brown</a><br />
--- a/test/integration/rules/target-size/target-size.html
+++ b/test/integration/rules/target-size/target-size.html
@@ -240,11 +240,11 @@ An inline <span> wrapping across multiple lines must not be treated as a
 single merged obscuring rect that fully overlaps everything near it -->
 <div style="max-width: 290px">
   <p>
-    <a href="#" id="pass-wrapped-span-1">lorem ipsum dolor</a>
-    <a href="#" id="pass-wrapped-span-2">amet</a>
+    <a href="#" id="ignored-inline-1">lorem ipsum dolor</a>
+    <a href="#" id="ignored-inline-2">amet</a>
     <span>et dolore magna aliqua</span>
   </p>
-  <a href="#" id="pass-wrapped-span-3" style="display: inline-block"
+  <a href="#" id="pass-adjacent-inline" style="display: inline-block"
     >Not obscured by the wrapped span above</a
   >
 </div>
--- a/test/integration/rules/target-size/target-size.json
+++ b/test/integration/rules/target-size/target-size.json
@@ -39,7 +39,7 @@
     ["#pass25"],
     ["#pass-adjacent"],
     ["#pass-fixed"],
-    ["#pass-wrapped-span-3"]
+    ["#pass-adjacent-inline"]
   ],
   "incomplete": [
     ["#incomplete1"],
--- a/test/integration/rules/target-size/target-size.json
+++ b/test/integration/rules/target-size/target-size.json
@@ -38,7 +38,8 @@
     ["#pass24"],
     ["#pass25"],
     ["#pass-adjacent"],
-    ["#pass-fixed"]
+    ["#pass-fixed"],
+    ["#pass-wrapped-span-3"]
   ],
   "incomplete": [
     ["#incomplete1"],


test/integration/rules/target-size/target-size.html
```diff
@@ -235,6 +235,20 @@
...
+<div style="max-width: 290px">
+  <p>
+    <a href="#" id="pass-wrapped-span-1">lorem ipsum dolor</a>
+    <a href="#" id="pass-wrapped-span-2">amet</a>
+    <span>et dolore magna aliqua</span>
+  </p>
+  <a href="#" id="pass-wrapped-span-3" style="display: inline-block"
+    >Not obscured by the wrapped span above</a
+  >
+</div>
```
straker, line 250: When an element doesn't match the rule we typically id these as such so we know just in the file and don't have to look at the json to verify. Also the elements are anchors not spans.


```suggestion
<div style="max-width: 290px">
  <p>
    <a href="#" id="ignored-inline-1">lorem ipsum dolor</a>
    <a href="#" id="ignored-inline-2">amet</a>
    <span>et dolore magna aliqua</span>
  </p>
  <a href="#" id="pass-adjacent-inline" style="display: inline-block"
    >Not obscured by the wrapped span above</a
  >
</div>
```

test/integration/rules/target-size/target-size.json
```diff
@@ -38,7 +38,8 @@
     ["#pass24"],
     ["#pass25"],
     ["#pass-adjacent"],
-    ["#pass-fixed"]
+    ["#pass-fixed"],
+    ["#pass-wrapped-span-3"]
```
straker:
```suggestion
    ["#pass-adjacent-inline"]
```

straker (changes requested):
Thanks for the pr. Seems this was fixed with the changes in https://github.com/dequelabs/axe-core/pull/5000 so can be closed. Just a small change.

arpitbharadwaj1:
Thanks for confirming and for the precise suggestions — applied both exactly as suggested (renamed the exempted-inline ids to `ignored-inline-1`/`ignored-inline-2`, and `pass-wrapped-span-3` to `pass-adjacent-inline` in both the fixture and expected-results json). No behavior change, just the clearer naming.

straker (approved):

straker:
Reviewed for security.

straker:
@arpitbharadwaj1 once you sign the CLA we can merge this in.

straker:
@arpitbharadwaj1 Ping on signing the CLA

Issue #5344, opened 2026-09-08
Incomplete results have no message when a check omits `metadata.messages.incomplete`


chutchins25:
## Description

When a check returns `undefined` (incomplete) and its JSON has no `metadata.messages.incomplete` entry, the node gets no message at all — `check.message` is `undefined` and `failureSummary` renders an empty bullet.

`lib/core/utils/publish-metadata.js` has two paths for incomplete messages and neither covers this case:

- The doT-template path is gated on `typeof messages.incomplete === 'object'`, which is `false` when the key is absent.
- The fallback then assigns `data.message = messages.incomplete`, i.e. `undefined`.

`incompleteFallbackMessage()` is imported in the file and used by `getDefaultMsg`, but `getDefaultMsg` is only reachable from inside `getIncompleteReason` — which the gate above stops from ever being called.

## Repro

`label-content-name-mismatch` has `pass` and `fail` messages but no `incomplete` message, and returns `undefined` for text that isn't human-interpretable:

```html
<button aria-label="comet">☄️</button>
```

Running that rule on `develop` (`540ecae9`):

```
incomplete count:    1
check.message:       undefined
node.failureSummary: "Fix any of the following:\n  "
```

## Expected

Fall back to `incompleteFallbackMessage()` so the node carries a usable message.

## Scope

Seven checks currently have a literal `return undefined` and no `incomplete` message: `label-content-name-mismatch`, `listitem`, `only-dlitems`, `only-listitems`, `important-letter-spacing`, `important-line-height`, `important-word-spacing`.

## Notes

Surfaced while reworking #5302, where a check-level `incomplete` message was added and then reverted as out of scope for that PR.

chutchins25:
Fixed in #5367 (`4887c3a8`): `publishMetaData` now falls back to `incompleteFallbackMessage()` when a check declares no `metadata.messages.incomplete`, so those nodes carry a usable message instead of rendering an empty bullet.

@straker filed #5375 as the follow-up — stopping checks from returning a state they have no message for in the first place, by moving check tests onto `testutils.getCheckEvaluate`. The seven checks named here are the ones that issue covers.

Closing manually — the `Closes issue #5344` footer on #5367 didn't fire because the PR targeted `develop` rather than the default branch.

Issue #5308, opened 2026-08-15
label-content-name-mismatch: punctuation normalization is language-dependent — CJK / full-width punctuation is not removed, causing false positives on Chinese, Japanese and Korean pages


doinsane:
### Product

axe-core

### Product Version

 4.13.0

### Latest Version

- [x] I have tested the issue with the latest version of the product

## Summary

`isStringContained()` normalizes both strings with `removeUnicode(str, { punctuations: true })`
before comparing, so a visible label and an accessible name that differ **only in punctuation**
are treated as a match.

`getPunctuationRegExp()` only covers ASCII/Latin punctuation. It does not cover CJK punctuation
(U+3000–U+303F) or the full-width forms (U+FF01–U+FF65) that are used in essentially all
Chinese, Japanese and Korean content.

The result: **the same authoring pattern passes in English and fails in Chinese.**

## Environment

- axe-core **4.13.0** (also reproduced on **4.10.3** — long-standing, not a regression)
- Chromium 145.0.7632.6 (Playwright 1.58.2); also reproducible in Chrome DevTools
- Rule: `label-content-name-mismatch` — WCAG 2.5.3 Label in Name, **Level A**

## Steps to reproduce

```html
<!DOCTYPE html>
<html lang="en">
<head><meta charset="utf-8"><title>label-content-name-mismatch: CJK punctuation</title></head>
<body>
  <!-- English: visible text and accessible name differ only in punctuation -->
  <a id="en" href="/a" aria-label="Example Corp, new window">Example Corp (new window)</a>

  <!-- Chinese: identical relationship, full-width punctuation -->
  <a id="zh" href="/b" aria-label="範例公司官網，另開新視窗">範例公司官網（另開新視窗）</a>
</body>
</html>
```

```js
await axe.run(document, { runOnly: { type: 'rule', values: ['label-content-name-mismatch'] } });
```

### Expected

Both pass. In both cases the accessible name contains the visible label; only the punctuation
separating the supplementary phrase differs (`(...)` / `, ...` vs `（...）` / `，...`).

### Actual

```json
{ "violations": ["#zh"], "passes": ["#en"] }
```

## Root cause

`curateString()` → `removeUnicode(str, { emoji: true, nonBmp: true, punctuations: true })`
→ `getPunctuationRegExp()`:

```js
function getPunctuationRegExp() {
[...]

PR #5345, opened 2026-09-08
fix(label-content-name-mismatch): compare visible label to name by word

chutchins25:
## Summary

Implements the comparison at the core of ACT rule 2ee8b8's [label in name algorithm](https://www.w3.org/WAI/standards-guidelines/act/rules/2ee8b8/proposed/#label-in-name-algorithm). The check compared the accessible name and the visible label as raw character substrings, so `Discover It` matched `Discover Italy`. Both strings are now tokenized into words at non-text characters — replaced with a space rather than deleted — and the label's words must form a **contiguous run** within the name.

Supporting change: a backward-compatible `replaceWith` option on `removeUnicode`, built from axe's explicit unicode ranges rather than a `\p{…}` property escape, to keep Safari 7+ and IE11 working.

## Scope

Per @straker on #5203, abbreviation and hyphenation differences are an **accepted divergence** from ACT:

> We are accepting that Inapplicable 5 & 6 will fail instead of incomplete and should be skipped. See act-rules/act-rules.github.io#2443

So the check reports those as violations and the two testcases stay skipped. There is no `incomplete` branch for punctuation.

Three algorithm steps remain unimplemented:

1. Parenthetical-content removal — tracked on #5207, which also covers the `Passed Example 11` whitespace case
2. Unicode case folding + NFKD normalization — the CJK and full-width half of this is tracked on #5308
3. Language-aware word segmentation (this splits on whitespace) — not tracked

The `wcag-act-rules` dependency bump called out on #5203 is tracked as #5346.

## ACT 2ee8b8 results

The `act` nightly installs `wcag-act-rules#main` (38 2ee8b8 testcases), not the pinned dependency (15). Verified against `main`: **34 passing / 4 pending / 0 failing**.

| Skipped testcase | Reason |
| --- | --- |
| Inapplicable 5 — `University Ave.` | Accepted divergence (abbreviation) |
| Inapplicable 6 — `non-standard` | Accepted divergence (hyphenation) |
| Passed 11 — `<span>Download</span><span> </span><span>specification</span>` | #5207. `visibleVirtual` recurses per child and `sanitize()`s each result, so a whitespace-only inline element is trimmed to `''` and the words join into `downloadspecification`. A `commons/text` fix, not a rule fix. |
| Passed 14 — `Search by date (YYYY-MM-DD)` | #5207. Needs the parenthetical-removal step. |

`Failed Example 3` and `Failed Example 15` were previously skipped under #5207 and are fixed by word-level matching, so their skips are removed. I confirmed the two accepted-divergence skips are load-bearing: unskipping them against `main` produces exactly two failures.

## Notes

Supersedes #5302, which was opened against #4311 before @straker closed that as a duplicate of #5203. The tokenizer here is the part of #5302 that survived review; the punctuation `incomplete` branch it grew is dropped per the ruling above.

The `publish-metadata.js` change from #5302 is not carried over — it is out of scope for a rule PR, and the underlying defect is now tracked as #5344.

Ref #5203

### DIFF
--- a/lib/checks/label/label-content-name-mismatch-evaluate.js
+++ b/lib/checks/label/label-content-name-mismatch-evaluate.js
@@ -9,13 +9,7 @@ import { getCategoryFormatRegExp } from '../../commons/text/unicode';
 
 /**
  * Check whether the visible label's words appear as a contiguous run of words
- * within the accessible name. This implements the comparison at the core of ACT
- * rule 2ee8b8's "label in name" algorithm: tokenize on non-text characters, then
- * match whole words rather than raw substrings.
- *
- * Note: the algorithm's parenthetical-content removal and case-folding/NFKD
- * normalization steps are not implemented yet
- * (see https://github.com/dequelabs/axe-core/issues/5203).
+ * within the accessible name.
  *
  * @param {String} label visible label text
  * @param {String} name accessible name
--- a/lib/checks/label/label-content-name-mismatch-evaluate.js
+++ b/lib/checks/label/label-content-name-mismatch-evaluate.js
@@ -5,36 +5,67 @@ import {
   sanitize,
   visibleVirtual
 } from '../../commons/text';
+import { getCategoryFormatRegExp } from '../../commons/text/unicode';
 
 /**
- * Check if a given text exists in another
+ * Check whether the visible label's words appear as a contiguous run of words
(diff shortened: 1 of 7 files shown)


straker (commented):
Looks good

straker (approved):

PR #5348, opened 2026-09-08
fix(p-as-heading): ignore paragraphs with heading roles

### LINKED ISSUE #5283: False positive: p-as-heading fires on <p role="heading" aria-level="n">
### Product

axe-core

### Product Version

4.12.1

### Latest Version

#### Expectation
p-as-heading should not report a violation for a `<p>` element that carries an explicit role="heading" with aria-level. Such an element is exposed as a heading in the accessibility tree, and the pattern is [ARIA12](https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA12), a documented sufficient technique for SC 1.3.1. The element should not be matched by the rule at all (expected outcome: inapplicable).

#### Actual
The element is reported as a serious violation — `"Heading elements should be used instead of styled <p> elements"` — regardless of the role or aria-level present on it.

#### How to Reproduce

```html
<div id="root">
  <p role="heading" aria-level="1" style="font-size: 2em; font-weight: 700;">Section title</p>
  <p>Body copy that follows the heading.</p>
</div>
```

```js
axe.run('#root', { runOnly: ['p-as-heading'] })
   .then(results => console.log(results.violations));
```

Result: one violation, `target #root > p:nth-child(1)`.

#### Additional context
lib/rules/p-as-heading-matches.js checks only three things — non-empty text, fewer than two sentence terminators, and the presence of a following non-empty `<p>` sibling. Neither the matches function nor p-as-heading-evaluate calls getRole / getExplicitRole, so an explicit role="heading" can never exclude the element. Evaluation then proceeds purely on computed style (weight / italic / size ratio against the next paragraph).

dvd233:
Fix the false positive where `p-as-heading` evaluates a paragraph that already resolves to an ARIA heading role.

- use axe-core's `getRole` resolution instead of parsing the `role` attribute directly
- make resolved headings inapplicable before the visual heading heuristics run
- add focused rule-match and rule integration coverage

The new unit and integration cases both fail against the current `develop` implementation and pass with this change.

## Testing

- `npm exec --yes -- pnpm@11.17.0 run build`
- `npm exec --yes -- pnpm@11.17.0 run test:tsc`
- all 44 rule-match test files, passed via explicit file paths
- focused `p-as-heading-matches` unit test
- generated focused `p-as-heading` integration test
- changed-file ESLint and Prettier checks

### Windows test-runner note

The repository aggregate scripts that contain POSIX single-quoted globs do not enumerate files correctly from this Windows/MSYS checkout (for example, `test/core/**/*.js` is passed as a literal pattern). I therefore ran the affected test and the complete rule-match group with explicit file paths; upstream CI can exercise the aggregate Linux commands.

Closes: #5283

### DIFF
--- a/lib/rules/p-as-heading-matches.js
+++ b/lib/rules/p-as-heading-matches.js
@@ -1,4 +1,10 @@
-function pAsHeadingMatches(node) {
+import { getRole } from '../commons/aria';
+
+export default function pAsHeadingMatches(node, virtualNode) {
+  if (virtualNode && getRole(virtualNode) === 'heading') {
+    return false;
+  }
+
   const children = Array.from(node.parentNode.childNodes);
   const nodeText = node.textContent.trim();
   const isSentence = /[.!?:;](?![.!?:;])/g;
@@ -17,5 +23,3 @@ function pAsHeadingMatches(node) {
 
   return siblingsAfter.length !== 0;
 }
-
-export default pAsHeadingMatches;
--- a/test/integration/rules/p-as-heading/p-as-heading.html
+++ b/test/integration/rules/p-as-heading/p-as-heading.html
@@ -22,6 +22,11 @@
   <p>Some text</p>
 </div>
 
+<div>
+  <p id="inapplicable5" role="heading" aria-level="1">Some text</p>
+  <p>A paragraph!</p>
+</div>
+
 <!-- Passes -->
 <div>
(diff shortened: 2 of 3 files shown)


WilcoFiers (approved):
LGTM. Thank you @dvd233! Would you mind signing the CLA?

For the books: Reviewed for security.

PR #5353, opened 2026-09-10
fix: Update de.json

timea1304:
spelling mistake "bieteen" should be "bieten"

Typo Mistake in German de.json

Closes:


straker (approved):
Thanks for the pr. 

Approved for security.

Issue #3739, opened 2022-10-21
target-size rule should be applicable to nodes in the body element


WilcoFiers:
When you put a link directly in the body of a page, target-size considers it inapplicable:

```
<html>
  <a href="#">link</a>
</html>
```

Weirdly, this doesn't seem to happen with `button`. It looks like this has something to do with the `isInTextBlock` method.

padmavemulapati:
Validated with the latest build,
for the test snippet:
```
<a role="link" aria-label="play" tabindex="0">
    <button tabindex="-1" style="margin:1px;">Play</button>
  </a>
```
![image](https://user-images.githubusercontent.com/47409579/197195016-40ff4b81-4ff5-4f7f-9a09-8944d4789914.png)


Also one more observation:
If we wrap this `href` within `div` it will be resulting `incomplete`
```
<div> <a role="link" aria-label="play" tabindex="0">
    <button tabindex="-1" style="margin:1px;">Play</button>
  </a> </div>
```
![image](https://user-images.githubusercontent.com/47409579/197195301-5ee93c59-bc3f-40b8-82ee-c5f11215d13f.png)

arpitbharadwaj1:
Checked this against current axe-core (4.10.2) with the exact repro:

```html
<a href="#">link</a>
<button>button</button>
```

Both elements are now flagged as `target-size` **violations** — the asymmetry is gone. (Also just found and fixed a related already-resolved-but-untested `target-size` false positive in #4295 → #5343, so this rule area seems to have had real fixes land since these were filed, likely as part of the ongoing target-size epic in #5225.)

Since both elements now get the same, correct treatment, this looks resolved. Given the fix already ships, I don't think a dedicated regression test is needed here the way #4295 warranted one — the current behavior is just ordinary rule output, not a special edge case — but happy to add one if you'd still like coverage locking in the anchor-directly-in-body case specifically.

straker:
This took a bit of time to debug. When using just the provided sample code and then loading axe-core on the page from devtools, running target-size produces a passes result. However when loading the code and adding a script tag as well to run axe, the results turn in to inapplicable.

```html
<html>
  <a href="#">link</a>
  <script src="/axe.js"></script>
  <script>
  (async () => {
    const res = await axe.run({ runOnly: 'target-size' })
    console.log(res);
    // inapplicable
  })();
  </script>
</html>
```

[...]

PR #5354, opened 2026-09-10
fix(commons): exclude script/style/noscript/template text from isInTextBlock

arpitbharadwaj1:
## Current behavior

`isInTextBlock()` (`lib/commons/dom/is-in-text-block.js`) walks all descendant text nodes of an element's nearest block-level ancestor to decide whether the element sits inside a run of surrounding text. The walk does not skip `<script>`, `<style>`, `<noscript>` or `<template>` elements, so their non-rendered content (e.g. a `<script>` tag's inline JS source, which is technically a text-node child) gets counted as if it were visible surrounding text.

This can flip the result incorrectly. For example, a lone `<a>` link that is otherwise the only content in its block should return `false` (not "in a text block"), but adding a `<script>` (or `<style>`) sibling with enough text content flips it to `true`, purely because of the non-rendered text length.

Found while investigating #3739 — `@straker` traced the `inapplicable` result for `target-size` back to exactly this function when a `<script>` tag was present in the test page ([comment](https://github.com/dequelabs/axe-core/issues/3739#issuecomment-3496671050)).

## Verification

Reproduced live against the current published build (axe-core 4.10.2) in a browser:
- `<div><a href="#">link</a></div>` → `isInTextBlock(link)` is `false` (correct).
- Same markup + an appended `<script>` with non-empty `textContent` → flips to `true`.
- Same, but with an *empty* `<script>` → stays `false` — isolating the cause to the swept-in text content, not the element's mere presence.
- Same test with `<style>` instead of `<script>` reproduces identically.

## Fix

Added `SCRIPT`/`STYLE`/`NOSCRIPT`/`TEMPLATE` to the set of nodes the walk treats as "not displayed on screen" (same branch that already excludes `display:none`, `overflow:hidden`, floated and absolutely/fixed-positioned content), so their content is skipped entirely rather than being walked into.

This mirrors the existing `nativelyHidden()` element list already used elsewhere in the codebase (`lib/commons/dom/visibility-methods.js`) for the same class of non-rendered elements, so it's consistent with how axe-core already treats this content elsewhere.

## Tests

Added two regression tests (`ignores script content`, `ignores style content`) next to the existing `ignores hidden content` test in `test/commons/dom/is-in-text-block.js`.

Related to #3739

### DIFF
--- a/lib/commons/dom/is-in-text-block.js
+++ b/lib/commons/dom/is-in-text-block.js
@@ -76,7 +76,11 @@ function isInTextBlock(
       currNode.style.display === 'none' ||
       currNode.style.overflow === 'hidden' ||
       !['', null, 'none'].includes(currNode.style.float) ||
-      !['', null, 'relative'].includes(currNode.style.position)
+      !['', null, 'relative'].includes(currNode.style.position) ||
+      // Don't count content that isn't rendered/visible text, e.g. a
+      // <script> tag's source code is a text node child but isn't text
+      // content a sighted user would perceive as part of the block
+      ['SCRIPT', 'STYLE', 'NOSCRIPT', 'TEMPLATE'].includes(nodeName)
     ) {
       return false;
 
--- a/test/commons/dom/is-in-text-block.js
+++ b/test/commons/dom/is-in-text-block.js
@@ -65,6 +65,28 @@ describe('dom.isInTextBlock', () => {
     assert.isFalse(isInTextBlock(link));
   });
 
+  it('ignores script content', () => {
+    fixtureSetup(html`
+      <p>
+        <a href="" id="link">link</a>
+        <script>var someScriptText = 'not real text content';</script>
+      </p>
+    `);
+    const link = document.getElementById('link');
+    assert.isFalse(isInTextBlock(link));
+  });
+
+  it('ignores style content', () => {
+    fixtureSetup(html`
+      <p>
+        <a href="" id="link">link</a>
+        <style>.some-style-text { color: red; }</style>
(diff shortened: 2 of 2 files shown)


lib/commons/dom/is-in-text-block.js
```diff
@@ -76,7 +76,11 @@ function isInTextBlock(
       currNode.style.display === 'none' ||
       currNode.style.overflow === 'hidden' ||
       !['', null, 'none'].includes(currNode.style.float) ||
-      !['', null, 'relative'].includes(currNode.style.position)
+      !['', null, 'relative'].includes(currNode.style.position) ||
+      // Don't count content that isn't rendered/visible text, e.g. a
+      // <script> tag's source code is a text node child but isn't text
+      // content a sighted user would perceive as part of the block
+      ['SCRIPT', 'STYLE', 'NOSCRIPT', 'TEMPLATE'].includes(nodeName)
     ) {
```
straker, line 84: I think instead of adding the list of elements to this if statement, we should swap out the logic to use `isVisibleOnScreen` which covers most of what this if statement is trying to do and gains us the list of hidden by default nodes.

```js
    if (
      nodeIsBlock || !isVisibleOnScreen(currNode)
    ) {
```

straker (changes requested):
Thanks for the pr.

lib/commons/dom/is-in-text-block.js
straker: @arpitbharadwaj1 Would you have time to update the pr, or would you like us to take it over?

PR #5355, opened 2026-09-10
fix(label-content-name-mismatch): mark value-text roles incomplete

### LINKED ISSUE #4561: label-content-name-mismatch in ARIA switch compares value text with name
### Product

axe-core

### Product Version

4.10.0

### Latest Version

#### Expectation
label-content-name-mismatch should not compare the content of the switch component with the aria-labelledby text.  

#### Actual
 label-content-name-mismatch is comparing the contents of the switch component with the aria-labelledby text.

#### How to Reproduce
The following code will produce the violation - "Ensure that elements labelled through their content must have their visible text as part of their accessible name"

    <label id="theswitch">Notifications</label>
    <div aria-labelledby="theswitch" role="switch" aria-checked="true" tabindex="0" >
	    <span>[On]</span>
	    <span >Off</span>
    </div>

<label id="theswitch">Notifications</label>
<div aria-labelledby="theswitch" role="switch" aria-checked="true" tabindex="0" >
	<span>[On]</span>
	<span >Off</span>
</div>
`
#### Additional context
The text contained in the switch is value text of sorts even if the real value is controlled by the aria-checked.  it's not a label so it shouldn't be compared for purposes of 2.5.3 Label in Name

chutchins25:
## Summary

`switch`, `checkbox`, `radio`, `menuitemcheckbox` and `menuitemradio` are all `nameFromContent` widget roles, so they reach this check today — but their visible content typically includes **value text** alongside the label. The reporter's example fails for that reason:

```html
<label id="theswitch">Notifications</label>
<div aria-labelledby="theswitch" role="switch" aria-checked="true" tabindex="0">
  <span>[On]</span>
  <span>Off</span>
</div>
```

Per @straker on #4561, these roles now return **incomplete** rather than a violation when the visible text isn't contained in the accessible name:

> We feel that instead of failing this example, we should mark it as incomplete for roles which normally wouldn't expect to have all content be part of the name. This includes `switch`, `checkbox`, `radio`, `menuitemcheckbox`, and `menuitemradio` roles. For those roles if we would normally return a failure we instead change it to an incomplete. We'll add the roles as a check option.

The role list is exposed as a new `valueTextRoles` check option, so it can be narrowed or extended without a code change. A pass is still a pass — the new branch only intercepts what would otherwise be a failure.

## Also in this PR

Two knock-on changes, both worth a look:

- **`incomplete` messages are new for this check.** It has never had any, so today its 17 pre-existing incomplete results (the `isHumanInterpretable` path) render with **no message at all** — `failureSummary` comes out as `"Fix any of the following:\n  "`. Adding `incomplete.default` fixes that for this rule as a side effect. The general defect is #5344.
- **`doc/rule-descriptions.md` flips to `failure, needs review`.** Generated. Arguably a correction rather than a change, since the rule already returned `undefined` on the `isHumanInterpretable` path.

## Testing

Unit coverage for all five roles, plus a matching-label pass, a role outside the list, and an empty option list. The reporter's exact repro is added as integration `#incomplete18`.

I mutation-tested the new branch rather than trusting a green run: deleting it and rebuilding turns all five role tests red with `expected false to equal undefined`.

## Stacked

Based on `chut/5203-label-in-name-word-match` (#5345), which is awaiting a final review. GitHub will retarget this to `develop` once that lands.

Closes #4561


test/checks/label/label-content-name-mismatch.js
```diff
@@ -281,4 +295,63 @@ describe('label-content-name-mismatch tests', () => {
     const actual = check.evaluate(vNode.actualNode, options, vNode);
     assert.isTrue(actual);
   });
+  valueTextOptions.valueTextRoles.forEach(role => {
```
straker: We should put these tests inside a `describe` block

```js
describe('options.valueTextRoles`, () => {});
```

lib/checks/label/label-content-name-mismatch.json
```diff
@@ -3,13 +3,24 @@
...
   },
   "metadata": {
     "impact": "serious",
     "messages": {
       "pass": "Element contains visible text as part of it's accessible name",
-      "fail": "Text inside the element is not included in the accessible name"
+      "fail": "Text inside the element is not included in the accessible name",
+      "incomplete": {
+        "default": "Unable to determine if the visible text is part of the accessible name",
+        "valueText": "Text inside the element may be value text rather than a label; value text does not need to be part of the accessible name"
```
straker: I think we could shorten this to just the reason rather than the explanation (we'll need to rebuild the _template.json as well)

```suggestion
        "valueText": "Text inside the element may be value text rather than a label"
```

straker (dismissed):

test/checks/label/label-content-name-mismatch.js
chutchins25: Done — the option's tests are now under `describe('options.valueTextRoles', …)`. It also picked up an open shadow DOM case for the role gate, so there are nine tests in the block.

lib/checks/label/label-content-name-mismatch.json
chutchins25: Applied your suggestion verbatim, and rebuilt `locales/_template.json`.

lib/checks/label/label-content-name-mismatch.json
```diff
@@ -3,13 +3,24 @@
...
+      "checkbox",
+      "radio",
+      "menuitemcheckbox",
+      "menuitemradio"
+    ]
   },
   "metadata": {
     "impact": "serious",
     "messages": {
       "pass": "Element contains visible text as part of it's accessible name",
```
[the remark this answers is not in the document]
chutchins25: Fixed in `7b1edd93`. Now reads "Element contains visible text as part of its accessible name". The translated locales carry their own text for this check, so none of them needed updating — only the English string and the generated template changed.

locales/_template.json
```diff
@@ -817,7 +817,11 @@
     },
     "label-content-name-mismatch": {
       "pass": "Element contains visible text as part of it's accessible name",
```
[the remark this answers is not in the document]
chutchins25: Fixed in `7b1edd93`. `locales/_template.json` is generated from the check metadata, so it picked the correction up on rebuild.

test/checks/label/label-content-name-mismatch.js
```diff
@@ -330,4 +344,81 @@ describe('label-content-name-mismatch tests', () => {
...
+      const actual = check.evaluate.call(
+        checkContext,
+        vNode.actualNode,
+        valueTextOptions,
+        vNode
+      );
+      assert.isFalse(actual);
+    });
+
+    it('returns false for a switch when the roles option is empty', () => {
```
straker, line 409: Minor nit: Because you said `roles` I looked at the element's `role` and saw it wasn't empty so was confused. This should help clarify you meant the option for the describe block `valueTextRoles`. But it's fine and can be merged as is. Up to you.

```suggestion
    it('returns false for a switch when the option is empty', () => {
```

straker (approved):

test/checks/label/label-content-name-mismatch.js
chutchins25: Merging as-is — taking you up on "fine as is" so the approval doesn't get dismissed by another push. You're right that `roles` reads as the element's role; I'll fold the rename into #5308, which touches this same test file.

Issue #4612, opened 2024-10-16
Input-button-name rule does not run on non-lowercase types


straker:
The following code returns `inapplicable` when running `input-button-name`. This is because the [rules selector](https://github.com/dequelabs/axe-core/blob/develop/lib/rules/input-button-name.json#L4) is case sensitive so does not match when running against non-lowercase types.

```html
<input type="SUBMIT" />
<input type="ImAGE" />
<input type="ResET" />
<input type="buTTON" />
```

PR #5358, opened 2026-09-11
fix(utils/matches): match HTML input type case-insensitively

cpruijsen:
The selector engine now treats HTML `input[type=…]` value matching as case-insensitive. XHTML and `type` on other elements (for example `ol`) stay case-sensitive.

`input-button-name` was inapplicable for mixed-case types such as `<input type="SUBMIT">` / `type="buTTON"`. HTML input `type` is an ASCII case-insensitive enumerated attribute; native `document.querySelector('input[type="submit"]')` matches those elements, but axe's virtual selector compared `getAttribute('type')` with exact case.

## Decision

Engine-level case folding for HTML input `type` rather than widening `input-button-name` to `input` and filtering in a matches function via `virtualNode.props.type` (the `label-matches.js` pattern). Native CSS already matches mixed-case input types, and `doc/developer-guide.md` says axe QSA should work seamlessly with `document.querySelectorAll`. The same path also covers `input-image-alt` (`type="ImAGE"` in the issue sample) and other `input[type=…]` selectors.

Can switch to a per-rule matches filter if that is preferred.

## Test plan

- [x] `input[type="submit"]` matches `<input type="SUBMIT">` (HTML)
- [x] Mixed-case `button` / `reset` / `image` types match
- [x] Open Shadow DOM mixed-case type matches
- [x] XHTML input type stays case-sensitive
- [x] `ol[type="A"]` does not match `ol[type="a"]`
- [x] Unrelated attributes stay case-sensitive
- [x] `input-button-name` integration: mixed-case fail/pass
- [x] Virtual rule: mixed-case `button` fails, `SUBMIT` passes

## Checklist

- [x] Default export at the top of the file
- [x] Return early
- [x] DocBlock on the new helper
- [x] No `console.log`
- [x] No hardcoded ARIA/HTML lists
- [x] Import restrictions (`core/utils`, no commons import)
- [x] Unit tests for the new paths
- [x] Integration HTML + JSON updated for the rule
- [x] Virtual-rules tests updated
- [x] Shadow DOM case added
- [x] No message / locale changes
- [N/A] API.md / axe.d.ts
- [N/A] CHANGELOG (generated from commits)

Closes issue #4612


lib/core/utils/matches.js
```diff
@@ -30,11 +30,41 @@ function matchesAttributes(vNode, exp) {
...
     exp.attributes.every(att => {
       const nodeAtt = vNode.attr(att.key);
-      return nodeAtt !== null && att.test(nodeAtt);
+      if (nodeAtt === null) {
+        return false;
+      }
+      if (att.test(nodeAtt)) {
+        return true;
+      }
+      return matchesHtmlInputType(vNode, att, nodeAtt);
```
WilcoFiers, line 39: I'm going to bring this up for discussion internally. Two things that I'm not sure about this:

1. This creates a discrepancy between how Axe selectors work and how regular CSS selectors do. I'm not sure about that. One way to avoid that could be to introduce `[type="button" i]` as case insensitive syntax instead
2. We might actually want all attribute values to be treated case insensitive. The only other attr value Axe ever queries are roles, and I think those are fully case insensitive too these days We'd need to do a little testing to confirm that

Issue #4260, opened 2023-11-29
color-contrast false-positives on empty text inputs


dbjorge:
### Product

axe-core

### Product Version

4.8.2

### Latest Version

- [X] I have tested the issue with the latest version of the product

#### Expectation

We saw an example today of a customer site using a text input that used event handlers to apply different classes to the input element in empty vs nonempty states. The empty state used a `color` with < 4.5:1 contrast, but the non-empty state was fine. This shouldn't be a color contrast violation.

#### Actual

This caused a `color-contrast` false positive because `axe-core` evaluates the text input contrast styling even when it contains no text:

https://github.com/dequelabs/axe-core/blob/f1af598cf3a9d0d99c05ae2e06fd203ae4936b00/lib/rules/color-contrast-matches.js#L71-L72

This gives extra-confusing results when the text input uses a `placeholder` - the contrast rule will currently ignore the placeholder text styling and instead evaluate as if the input contained a non-placeholder value, and any resulting violations will appear as if it's evaluated the placeholder text with a usually-incorrect foreground color.

Placeholder text *is* generally subject to contrast requirements, so ideally we'd evaluate the placeholder (but using its foreground color, not the input's). But that can be a separate issue vs this false positive; for the purposes of this bug, either of "incomplete" or "not matched" would be an improvement over the current behavior.

#### How to Reproduce

Suggested new `color-contrast.html` integration test cases:

```html
<!-- currently fails, should be either incomplete (with some new message format suggesting testing it with a value set) or unmatched -->
<input
  id="ignore13"
  type="text"
  style="background: #fff; color: #eee"
/>

<!-- currently fails, should be either pass (ideal eventually), incomplete (ideal for this bug), or unmatched (better than false positive at least) -->
<style>
.placeholder-000::placeholder {
  color: #000
}
</style>
<input
  id="tbd"
  type="text"
  placeholder="Placeholder"
  style="background: #fff; color: #eee"
  class="placeholder-000"
/>

<!-- currently fails, but against #eee foreground color. Should either fail with #ddd foreground color data (ideal eventually), incomplete (ideal for this bug), or unmatched (false negative, but better than the above case being a false positive) -->
<style>
.placeholder-ddd::placeholder {
  color: #ddd
}
</style>
<input
  id="tbd"
  type="text"
  placeholder="Placeholder"
  style="background: #fff; color: #eee"
  class="placeholder-ddd"
/>
```

#### Additional context

[...]

PR #5359, opened 2026-09-11
fix(color-contrast): treat empty low-contrast text inputs as incomplete

cpruijsen:
The check now returns incomplete with `emptyValue` when a native text-entry control has no sanitised value and contrast would fail, asking testers to re-run with a value entered. Filled fields with the same colors still fail. Empty fields that already meet the ratio still pass.

Empty text inputs can apply a low-contrast `color` in the empty state and a passing color once a value is entered. color-contrast still matches those fields (intentionally, since PR #2130). The fail was on the empty-state foreground, including when a placeholder was visible.

## Decision

Incomplete-on-fail for empty native textboxes (and password), same shape as `shortTextContent`. The issue called incomplete ideal for this bug and asked for a message that suggests testing with a value set. Matching empty form fields was left in place on purpose in PR #2130, and incomplete-all-empty would flood reports with blank search boxes that already meet 4.5:1.

Alternative: stop matching empty inputs, or incomplete every empty text input including those that already pass. Can switch to unmatched or to incomplete-all-empty.

Placeholder contrast using `::placeholder` color is not implemented here; those cases are incomplete rather than a false fail against the input `color`.

Closes issue #4260

### DIFF
--- a/test/checks/color/color-contrast.js
+++ b/test/checks/color/color-contrast.js
@@ -946,6 +946,126 @@ x
     });
   });
 
+  describe('with empty text entry', () => {
+    it('should return undefined for an empty text input with insufficient contrast', () => {
+      const params = checkSetup(html`
+        <input
+          id="target"
+          type="text"
+          style="background-color: #fff; color: #eee"
+        />
+      `);
+
+      const actual = contrastEvaluate.apply(checkContext, params);
+      assert.isUndefined(actual);
+      assert.equal(checkContext._data.messageKey, 'emptyValue');
+    });
+
+    it('should return undefined for an empty textarea with insufficient contrast', () => {
+      const params = checkSetup(
+        '<textarea id="target" style="background-color: #fff; color: #eee"></textarea>'
+      );
+
+      const actual = contrastEvaluate.apply(checkContext, params);
+      assert.isUndefined(actual);
+      assert.equal(checkContext._data.messageKey, 'emptyValue');
+    });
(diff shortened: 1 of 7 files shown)


test/checks/color/color-contrast.js
```diff
@@ -946,6 +946,126 @@ x
...
+          id="target"
+          type="text"
+          value="hello"
+          style="background-color: #fff; color: #eee"
+        />
+      `);
+
+      const actual = contrastEvaluate.apply(checkContext, params);
+      assert.isFalse(actual);
+    });
```
WilcoFiers, line 1048: Optional guard: this example sets `value` from the HTML attribute, which is enough today because `isEmptyTextEntry` reads `virtualNode.props.value` (`actualNode.value`), not `getAttribute('value')`. A script-set property would catch a later switch to the attribute.

```suggestion
    });

    it('should return false for a text input with a script-set value and insufficient contrast', () => {
      const input = document.createElement('input');
      input.id = 'target';
      input.type = 'text';
      input.setAttribute('style', 'background-color: #fff; color: #eee');
      input.value = 'hello';
      const params = checkSetup(input);

      const actual = contrastEvaluate.apply(checkContext, params);
      assert.isFalse(actual);
    });
```

WilcoFiers (approved):

WilcoFiers:
@cpruijsen Thank you very much for the contribution!

Issue #5311, opened 2026-08-18
Add tags for EN 301 549 v4 to target-size


WilcoFiers:
EN 301 549 version 4.1.1 is releasing soon. Once it does we want to add tags to axe-core. There's only one rule that changes, so it's a fairly simple change. All we need to do is add `EN-301-549v4` and `EN-9-2.5.8` to the `target-size` rule, and than update our docs for the tags. Anyone who wants to run EN 301 549 v3 needs to just run the `EN-301-549` tag. Anyone who wants v4 needs to also run `EN-301-549v4`.

straker:
Closed via #5361

padmavemulapati:
### QA validation — tags correct, docs look outstanding

Validated on `develop` @ `<commit>` (axe `<version>`).

**Tags — correct**
- `target-size` carries `EN-301-549v4` and `EN-9.2.5.8`, with `cat.sensory-and-visual-cues`, `wcag22aa` and `wcag258` intact.
- It does **not** carry bare `EN-301-549`, so v3 consumers don't pick it up — which is the behaviour described in the issue.
- Both new tags are unique to this rule; no other rule carries either.
- The sub-criterion tag shipped as `EN-9.2.5.8` rather than the `EN-9-2.5.8` written in the description. The dotted form matches the existing convention (`EN-9.2.4.4`, `EN-9.4.1.2`), so I've taken the description as a typo rather than the spec — flag it if that's wrong.

**Selection — works as intended**
- `runOnly: {type: 'tag', values: ['EN-301-549']}` does not select `target-size`.
- Adding `EN-301-549v4` does, and the combined v3 + v4 tag run is the union of the two.
- The rule produces a genuine violation on undersized adjacent targets when run through the new tag.

**Outstanding**
`doc/rule-descriptions.md` doesn't appear to have been regenerated — the `target-size` row still lists only `cat.sensory-and-visual-cues, wcag22aa, wcag258`, and the file contains no occurrence of `EN-301-549v4`. Since the issue calls out documentation as part of the change, I'd like that regenerated before this moves to Done.



<img width="1713" height="966" alt="Image" src="https://github.com/user-attachments/assets/1b0cff61-c59e-4e72-a8af-0abff2898a85" />

PR #5361, opened 2026-09-13
feat: add EN 301 549 v4 tags to target-size

modusensus:
## Summary

- Add `EN-301-549v4` and `EN-9.2.5.8` tags to the `target-size` rule. EN 301 549 v4.1.1 (published 2026-09-02) aligns clause 9 with WCAG 2.2, so target-size (`wcag258`) is now required by the standard. The existing `EN-301-549` tag keeps mapping to v3.2.1, which is WCAG 2.1-based and has no SC 2.5.8 equivalent.
- Extend the rule tag validation in `build/validate-files.mjs`: the EN 301 549 standard tag regex now also accepts `EN-301-549v4`, and the WCAG level restriction moves into the EN-specific check so that v4-tagged rules may sit on WCAG 2.2 levels while rules carrying the v3 `EN-301-549` tag stay restricted to WCAG 2.0/2.1 levels.
- Document the new tag in `doc/API.md`. `doc/rule-descriptions.md` is regenerated by `pnpm run build` (prettier also re-aligns the affected tables).

Audit behavior stays incremental: v3 audits keep running just the `EN-301-549` tag (unchanged), while v4 audits run `EN-301-549v4` in addition, which brings in `target-size`.

Closes issue #5311

## Test plan

- `pnpm run validate` passes with the new tags; sanity checks confirm a v3-tagged rule on `wcag22aa` is still rejected.
- `pnpm run eslint` and `pnpm run build` pass; `doc/rule-descriptions.md` regenerates cleanly.
- Unit suites (`test:unit:*`) run locally against a version-matched ChromeDriver; the only failures reproduce identically on a pristine `develop` checkout in the same environment (Chrome 147 rendering/media-related), so they are unrelated to this change.


WilcoFiers (approved):
@modusensus Thank you very much!

For the books. Reviewed for security.

PR #5364, opened 2026-09-14
fix(scrollable-region-focusable): negative tabindex content needs review instead of failing outright

### LINKED ISSUE #5363: False positive: scrollable-region-focusable on menus with programmatic keyboard focus
### Product

axe-core

### Product Version

4.13.0 (develop)

### Latest Version

#### Expectation

`scrollable-region-focusable` should pass. The list is keyboard accessible: activating the Products button (or ArrowDown on it) moves focus to the first item, and arrow keys move between items.

#### Actual

axe reports a violation on `#menu` (`Element should have focusable content` / `Element should be focusable`).

#### How to Reproduce

```html
<style>
  #menu {
    max-height: 6em;
    overflow-y: auto;
    width: 12em;
    margin: 0;
    padding: 0;
    border: 1px solid;
    list-style: none;
  }
  #menu button {
    display: block;
    width: 100%;
    text-align: left;
  }
</style>

<button id="menubutton" aria-expanded="true" aria-controls="menu">
  Products
</button>
<ul id="menu" aria-labelledby="menubutton" tabindex="-1">
  <li><button tabindex="-1">Keyboards</button></li>
  <li><button tabindex="-1">Mice</button></li>
  <li><button tabindex="-1">Monitors</button></li>
  <li><button tabindex="-1">Headsets</button></li>
  <li><button tabindex="-1">Webcams</button></li>
  <li><button tabindex="-1">Microphones</button></li>
  <li><button tabindex="-1">Docking stations</button></li>
  <li><button tabindex="-1">Cables</button></li>
</ul>

<script>
  const menubutton = document.getElementById('menubutton');
  const menu = document.getElementById('menu');
  const menuitems = [...menu.querySelectorAll('button')];

  function setExpanded(isExpanded) {
    menubutton.setAttribute('aria-expanded', String(isExpanded));
    menu.hidden = !isExpanded;
  }

  menubutton.addEventListener('click', () => {
    const isExpanded = menubutton.getAttribute('aria-expanded') === 'true';
    setExpanded(!isExpanded);
    if (!isExpanded) {
      menuitems[0].focus();
    }
  });

  menubutton.addEventListener('keydown', event => {
    if (event.key !== 'ArrowDown') {
      return;
    }
    event.preventDefault();
    setExpanded(true);
    menuitems[0].focus();
  });

  menu.addEventListener('keydown', event => {
    const index = menuitems.indexOf(event.target);
    const offset = { ArrowDown: 1, ArrowUp: -1 }[event.key];
    if (index === -1 || !offset) {
      return;
    }
    event.preventDefault();
    menuitems[(index + offset + menuitems.length) % menuitems.length].focus();
  });
</script>
```

#### Additional context

Customer mega menu. Items stay `tabindex="-1"` and are focused by script. `tabindex="-1"` on the `ul` is only there to stop Chrome from focusing the scroll container.

WilcoFiers:
If a scrollable region is / has no content in focus navigation report it as incomplete (needs review) if it does have elements that have negative tabindex, because a script might be present that gives these elements focus to allow the scroll.

Closes: https://github.com/dequelabs/axe-core/issues/5363


WilcoFiers (commented):
## Review

Scrollable regions with no tabbable content now go to needs review when they still have visible `[tabindex]` descendants, which is the right direction for menus that move focus by script ([#5363](https://github.com/dequelabs/axe-core/issues/5363)). Tests, incomplete message, and `locales/_template.json` are in place.

Two follow-ups:

### 1. Important — only incomplete for programmatically focusable tabindex content

[`focusable-content-evaluate.js`](https://github.com/dequelabs/axe-core/blob/76943859bd34a74770044d3706ff96626bfab19a/lib/checks/keyboard/focusable-content-evaluate.js#L19-L21) currently treats any visible `[tabindex]` descendant as incomplete. Disabled, inert, or otherwise non-focusable controls cannot be focused by script to make the region scrollable, so those should still **fail**.

Suggestion: also require `vNode.isFocusable` (after the tabbable early-return this is effectively negative tabindex + focusable). Please add a unit test such as a `disabled` input with `tabindex="-1"` still returning `false`.

### 2. Suggestion — incomplete message

[`focusable-content.json`](https://github.com/dequelabs/axe-core/blob/76943859bd34a74770044d3706ff96626bfab19a/lib/checks/keyboard/focusable-content.json#L8): `Check if content with tabindex is keyboard accessible` can read as “put these back in the tab order.” For this rule, reviewers should check whether scripted focus still lets keyboard users scroll the region.

Something like: `Check that elements with a negative tabindex can receive keyboard focus so the region can be scrolled`. Please update `locales/_template.json` in the same change.

lib/checks/keyboard/focusable-content-evaluate.js
```diff
@@ -1,20 +1,30 @@
...
+  if (tabbableContentElements.length > 0) {
+    return true;
   }
 
-  // remove thyself from tabbable elements (if exists)
-  const tabbableContentElements = tabbableElements.filter(
-    el => el !== virtualNode
+  // Elements with a tabindex can be given focus by script, which cannot be
+  // determined automatically
+  const tabindexContent = querySelectorAll(virtualNode, '[tabindex]').filter(
```
straker, line 18: Changes look fine, but this double `querySelectorAll` on the node keeps irking me the wrong way (`vNode.tabbableElements` does a `querySelectorAll` as well). I'd be better if it was a single query selector that looks at both tabbable and tabindex, but I know doing so would 1) make the code way more complicated as we don't have a `isTabbable` function and 2) negate the `vNode.tabbableElements` cache. So in the end it may not matter either way.

test/checks/keyboard/focusable-content.js
```diff
@@ -28,16 +30,61 @@ describe('focusable-content tests', () => {
     assert.isFalse(actual);
   });
 
-  it('returns false when content element is taken out of focusable order (tabindex = -1)', () => {
+  it('returns undefined when content has a tabindex but is not in the tab order', () => {
```
straker, line 33: We should add a test that returns `true` when there is both a tabbable element and a negative tabindex element to lock in the return order. At the moment swapping the returns around (return `undefined` if negative tabindex first) would not break the tests.

straker (changes requested):

PR #5365, opened 2026-09-14
perf(get-scroll): remove memoize

straker:
In the same vein as https://github.com/dequelabs/axe-core/issues/5332 and  https://github.com/dequelabs/axe-core/issues/5298, the memoize around the function causes a ton of slow down on larger sites. On my largest benchmark site using latest develop, axe-core runs in 20s. Without the memoize on `getScroll` it runs in 10s.


chutchins25 (approved):
Looks good to me.

WilcoFiers (approved):

PR #5367, opened 2026-09-15
fix(core): use the fallback message when a check has no incomplete message

chutchins25:
## Summary

A check that returns `undefined` without a `metadata.messages.incomplete` entry
produced no message at all: `check.message` was `undefined` and `failureSummary`
rendered an empty bullet.

`publish-metadata.js` already imports `incompleteFallbackMessage`, but it was
unreachable — `getDefaultMsg` runs only from inside `getIncompleteReason`, which
the `typeof messages.incomplete === 'object'` gate skips when the key is absent.
This falls back to it directly.

## Repro

`label-content-name-mismatch` has `pass` and `fail` messages but no `incomplete`
message, and returns `undefined` for text that isn't human-interpretable:

```html
<button aria-label="comet">☄️</button>
```

| | `check.message` | `node.failureSummary` |
| --- | --- | --- |
| `develop` (`388132db`) | `undefined` | `"Fix any of the following:\n  "` |
| this branch | `"axe couldn't tell the reason…"` | `"Fix any of the following:\n  axe couldn't tell the reason…"` |

## Scope

Seven checks return `undefined` with no incomplete message:
`label-content-name-mismatch`, `listitem`, `only-dlitems`, `only-listitems`,
`important-letter-spacing`, `important-line-height`, `important-word-spacing`.

This fixes the generic fallback only; it does not author per-check incomplete
messages (#5355 adds real ones for `label-content-name-mismatch`). The fallback
string is already localizable via `incompleteFallbackMessage`, so no message
strings are added and `locales/_template.json` is unchanged.

## Testing

New unit test in `test/core/utils/publish-metadata.js`, confirmed red on
`develop` before the fix. Local unit suites (core, checks, commons,
rule-matches, integration, virtual-rules, api, gather-internals) plus `test:tsc`,
`test:locales` and `eslint` all green.

Closes issue #5344


straker (approved):
Good find. However I think this reveals a larger problem in that we have rules which are allowed to return `undefined` but do not specify a message. We probably shouldn't be able to do that since the message is not helpful and the user won't know 1) why it returned as incomplete and 2) how to fix it. I'll create a follow up issue that focuses on preventing that from happening.

_EDIT:_ https://github.com/dequelabs/axe-core/issues/5375

PR #5368, opened 2026-09-15
fix(commons/text/visibleVirtual): sanitize visible text once, not at every level

chutchins25:
## Summary

`visibleVirtual` ends every recursive call with `sanitize()`, which trims. A
whitespace-only child therefore collapses to `''` before the parent joins its
children, gluing together the words on either side:

```html
<a aria-label="Download specification"
  ><span>Download</span><span id="space"> </span><span>specification</span></a
>
```

produced `downloadspecification`, which no downstream tokenizing can recover.

Per the decision on #5207 — keep using `visibleVirtual` rather than writing a
rule-specific text function — nested calls now return their text as-is and only
the outermost call sanitizes. The recursion passes an internal `skipSanitize`
option, so the public signature and every existing call site are unchanged.

## Effect

ACT `2ee8b8` against `wcag-act-rules#main`, which is what the nightly installs:
**36 passing / 2 pending / 0 failing**, up from 35/3/0. `Passed Example 11`
passes with its `skipTests` entry removed; the two remaining pendings are the
accepted abbreviation/hyphenation divergence
(act-rules/act-rules.github.io#2443). Against the pinned dependency: 14 passing
/ 1 pending, unchanged.

Whitespace at the edge of a child element is now also kept as a word separator —
`<span>Hello </span><span>world</span>` returns `Hello world` where it
previously returned `Helloworld`. That is the same defect, and it is what the
algorithm's "maintaining space between spans" requires.

## Testing

Three unit tests in `test/commons/text/visible-virtual.js`, each confirmed red
before the change.

`visibleVirtual` feeds `color-contrast` (`noRecursing: true`, unaffected),
`duplicate-img-label`, `form-control-value`, `aria.labelVirtual`,
`text.labelVirtual` and `label-content-name-mismatch`. Locally green: every unit
suite, the 136-page selenium integration run, `test:node`, `test:jsdom`,
`test:locales`, `test:tsc` and `eslint`, with no existing test needing a change.

Closes issue #5207


straker (changes requested):
Visible Virtual tests use the same tests twice: once for screen readers and the other for visible on screen. We'll need to add the same set of tests to both parts. Additionally we should add a test similar to `should trim the result` which has multiple whitespace characters in a row of various types to ensure that all whitespace between elements is collapsed to a single whitespace (e.g. `<span>Hello   &nbsp;\u00A0</span><span>  &nbsp;\r\n   </span><span>     \n \n &nbsp; \nHi</span>` should turn into `Hello Hi`)

chutchins25:
Thanks @straker — both done in fc7fa877.

Mirrored all three cases into the `screen reader` block, placed after its own `should trim the result`, and replaced my nested-sanitize case with `should collapse whitespace between elements` using your fixture:

```js
fixture.innerHTML =
  '<span>Hello   &nbsp; </span><span>  &nbsp;\r\n   </span>' +
  '<span>     \n \n &nbsp; \nHi</span>';
assert.equal(visibleVirtual(tree[0]), 'Hello Hi');
```

All six are discriminating: reverting `lib/commons/text/visible-virtual.js` to `388132db` and rebuilding turns every one of them red.

straker (approved):

PR #5369, opened 2026-09-15
chore: fix perf-report action (again)

straker:
So apparently setting `persist-credentials: false` had a secondary affect where doing that meant there was no credentials for the token to use so it tries to [prompt for the username when trying to create the pr](https://github.com/dequelabs/axe-core/actions/runs/34983814839/job/104430657590)? Didn't expect that at all so removing it.


chutchins25 (approved):
Looks good.

WilcoFiers (approved):

PR #5371, opened 2026-09-15
chore: fix perf reporter log capture for axe-core <=4.12.0

straker:
axe-core 4.12.0 and prior did not have `axe._setLogger` so have to override the console log.

Additional, when I tried to run against `@next` the [logs looked wrong](https://github.com/dequelabs/axe-core/pull/5372/changes) and the performance reporter changes didn't look like it got run. I'm not entirely sure why it didn't work, but it could be a stale metadata tag from pnpm cache (the `+ axe-core 4.13.0-canary.ee015fa (4.13.0 is available)` might mean that). It could also be a problem with installing a version of axe-core in the axe-core package itself.

Either way, getting the tarball directly and using that should (🤞) fix the problem.

_EDIT:_ 

Running the action from this branch shows working as expcted:

- using `@next` tag https://github.com/dequelabs/axe-core/pull/5373
- using `4.11.0` tag https://github.com/dequelabs/axe-core/pull/5374


.github/workflows/perf-report.yml
```diff
@@ -21,16 +21,35 @@ jobs:
...
+          # If the input starts with a digit, treat as a pinned semver; otherwise
+          # resolve as a dist-tag by hitting the registry directly. Going through
+          # `pnpm add` can use a stale tag from pnpm's cache, so fetch the tarball
+          # straight from the registry instead of installing
+          if [[ "$Req" =~ ^[0-9] ]]; then
+            Requested="$Req"
+          else
+            Requested=$(curl -sSf https://registry.npmjs.org/-/package/axe-core/dist-tags | jq -re ".\"$Req\"")
+          fi
+          echo "Resolved axe-core@$Req -> $Requested"
```
chutchins25: Since this line is being rewritten anyway — worth passing the input through `env:` rather than expanding `${{ github.event.inputs.version }}` into the script body? GitHub substitutes it textually before bash parses the line, so a `'` in the version string breaks out of the quotes. The Open PR step below already uses the `env:` form. Triggering needs write access, so this is hardening rather than an open door.

Same idea for the dist-tag lookup — `jq --arg` passes the tag as data instead of splicing it into the filter text.

```suggestion
        env:
          REQ: ${{ github.event.inputs.version }}
        run: |
          # If the input starts with a digit, treat as a pinned semver; otherwise
          # resolve as a dist-tag by hitting the registry directly. Going through
          # `pnpm add` can use a stale tag from pnpm's cache, so fetch the tarball
          # straight from the registry instead of installing
          if [[ "$REQ" =~ ^[0-9] ]]; then
            Requested="$REQ"
          else
            Requested=$(curl -sSf https://registry.npmjs.org/-/package/axe-core/dist-tags | jq -re --arg tag "$REQ" '.[$tag]')
          fi
          echo "Resolved axe-core@$REQ -> $Requested"
```

chutchins25 (commented):
Nice work! The core fix checks out — I verified the `_setLogger` boundary against the published tarballs rather than the tags (absent in `4.12.0`, present in `4.12.1` and `4.13.0`, so the comment is accurate), and confirmed `4.12.0`'s bundle reads `console.log` at call time via `Function.prototype.apply.call(console.log, console, arguments)` — which is what makes the fallback work. Your v4.11.0 run in #5374 produced a full report with real timings, so it's verified end to end.

Dropping `git checkout -- package.json pnpm-lock.yaml` is right too, now that nothing mutates those.

One hardening question inline, plus a nit: `window.console.log = captureLogs` on line 182 of `perf/report.js` is missing its semicolon — it's inside a template literal, so `pnpm run fmt` won't pick it up.

chutchins25 (approved):
Thanks for the changes! Looks good to me now.

PR #5383, opened 2026-09-18
ci: run perf-compare on same runner

straker:
[Testing the compare action](https://github.com/dequelabs/axe-core/pull/5382) revealed a problem. What should have been a 10-20s difference resulted in a 1s difference between head and base branches. Digging into it I noticed the CPUs were different between the base and head reports, which would account for the major difference in performance. Older model CPU runs ~2x slower, so the page normally runs in 10s locally would take 20s on base, while a newer CPU could run just like my local machine so the additional 10s run time resulted in a 20s total time. This is why the compare passed when it shouldn't have.

The fix is to run both reports in the same runner so we get the same CPU and can actually compare apples-to-apples.

Unfortunately each of the historical reports in `perf/reports` also suffer from different CPUs so we can't compare the times there like I hoped we would be able to.


chutchins25 (approved):
Looks good to me.

Issue #5179, opened 2026-06-18
link-in-text-block: violation message does not explain why the link was matched as in-text-block


blaumohn:
#### Expectation

When `link-in-text-block` reports a violation, the message should explain
why the link was considered to be inside a text block — i.e. what context
triggered the rule to apply at all.

#### Actual

The violation message only states:
> "Ensure links are distinguished from surrounding text in a way that does not rely on color."

It does not explain why the link matched the rule in the first place.
The `matches` function (`isInTextBlock`) uses a length comparison:
`parentText.length > widgetText.length`. This threshold is invisible in the
output, making the result appear inconsistent or even random to the user.

Example: two HTML files with structurally identical `<footer>` elements — same CSS,
same `<a>` — produce different results solely because the surrounding
`<span>` text is one word longer in one file than in the other.
The violation fires when `parentText` (62 chars) exceeds `LINK_HREF` (50 chars),
and does not fire when `parentText` (48 chars) is shorter.
Nothing in the reported message hints at this.

#### How to Reproduce

1. Install dependencies:

   ```
   npm install @axe-core/playwright @playwright/test playwright
   ```

2. Save the script below as `repro.mjs` and run:

   ```
   node repro.mjs
   ```

3. Expected output:

   ```
   FAIL — parentText (span+pipe) longer than link text
   span text length : 62
   link text length : 50
   violation        : link-in-text-block

   PASS — parentText (span+pipe) shorter than link text
   span text length : 48
   link text length : 50
   violation        : none
   ```

```js
import AxeBuilder from '@axe-core/playwright';
import { chromium } from '@playwright/test';

const CSS = `
  footer a { color: #555; text-decoration: none; }
  footer a:hover { text-decoration: underline; }
`;

const LINK_HREF = 'https://github.com/blaumohn/lebenslauf-web-vorlage';

const cases = [
  {
    label: 'FAIL — parentText (span+pipe) longer than link text',
    span: '2026-06-18 Dani Y. | PHP-Lebenslauf-App inkl. CI/CD, Vollstack',
  },
  {
    label: 'PASS — parentText (span+pipe) shorter than link text',
    span: '2026-06-18 Alex B. | Preview der App von Dani Y.',
  },
];

function buildPage(spanText) {
  return `<!DOCTYPE html><html lang="de"><head>
    <meta charset="utf-8">
    <style>${CSS}</style>
  </head><body>
    <main><h1>Test</h1></main>
    <footer>
[...]

PR #5384, opened 2026-09-19
fix(link-in-text-block): explain the text length comparison in messages

modusensus:
## Summary

- Surface the text-length comparison that makes `link-in-text-block` apply. The rule only runs when the surrounding block text is longer than the link text (`parentText.length > widgetText.length`), but nothing in the report said so — the same markup could pass on one page and fail on another with no explanation (#5179).
- Added an additive `returnLengths` option to `isInTextBlock`. Default behaviour is unchanged for the existing callers (`link-in-text-block-matches`, `widget-not-inline-matches`).
- Appended the lengths to the `fgContrast` and `bgContrast` failure messages, and to the `link-in-text-block-style` failure message, so both checks of the rule explain themselves consistently.
- Regenerated `locales/_template.json`.

## Notes on the shape

`isInTextBlock` is called from the rule's `matches`, and the check `evaluate` runs separately afterwards, so those lengths aren't available in `this.data()` for free. The evaluates call `isInTextBlock(node, { returnLengths: true })` on the failure path only, so passing nodes pay nothing extra. `link-in-text-block-evaluate` already re-walks the block parent itself, so this fits the file's existing style. Happy to move it into `matches` with a cache instead if that shape is preferred.

## Test plan

- New `options.returnLengths` unit tests in `test/commons/dom/is-in-text-block.js`.
- Updated the existing `checkContext._data` assertions in `test/checks/color/link-in-text-block.js` and `test/checks/color/link-in-text-block-style.js`.
- New end-to-end assertion that the rendered `failureSummary` explains the comparison.
- Ran `test/commons`, `test/rule-matches` and `test/checks` under Web Test Runner in headless Chrome, plus `tsc`, eslint and prettier.

Closes issue #5179


straker:
Thanks for taking on this pr. We're currently trying to figure out the message as right now it's getting kind of long. 

I'm also not sure if the additional information is useful since the aim of the message is to give succinct information to fix the problem, and I don't foresee anyone updating the text length in order to fix the violation. However the premise of the ticket is valid. I'll get with @WilcoFiers and see if we can figure something out.

